Issue #173, opened 2015-05-15
Handle user log out


caseypt:
Depends on #162.

Currently the user is directed to a logout page.

Instead we should tell them they logged out, update the header, and keep them on the current page if it doesn't require permissions, or redirect them to the home page if it does, or something along those lines.

To reproduce:

- Log in
- Log out. You will be redirected to the logout page.

PR #201, opened 2015-05-21
AJAX User Logout

rajadain:
Connects #173 

I'm looking for feedback for work done in this PR, especially in 9e00729 and 5adc73d. I've worked along the current patterns, but perhaps we should rethink some of the organization. For example, would it be better for the User model to have all API interaction code, so that other places are just calling a method on that? We defer requiring App in HeaderView since core views are a dependency of App and we don't want circular references, but if we need access to App methods in multiple places we have to `require` it each time, is that okay?

I expect this PR to go through some revision before being merge-able.


src/mmw/apps/user/views.py
```diff
@@ -47,6 +47,20 @@ def ajax_login(request):
     return Response(data=response_data, status=status)
 
 
+@decorators.api_view(['GET'])
+@decorators.permission_classes((AllowAny, ))
+def ajax_logout(request):
+    response_data = {}
+    status = 200
```
lliss: I think 200 is the default so, unless you need to send an error code, you should be able to leave this out.

src/mmw/js/src/app.js
```diff
@@ -64,6 +65,34 @@ var App = new Marionette.Application({
...
+    userLogout: function() {
+        var csrftoken = helpers.Cookie.get('csrftoken');
+
+        $.ajax({
+            url: '/user/ajaxlogout',
+            type: 'GET',
+            beforeSend: function(xhr) {
+                xhr.setRequestHeader('X-CSRFToken', csrftoken);
+            }
+        }).always(function() {
```
lliss: If the request fails I think this will still set the ui to show the user as logged out. That could put things in an inconsistent state. I could imagine thinking I'm logged out and then refreshing the page and finding that I'm logged back in.

src/mmw/js/src/core/templates/header.ejs
```diff
@@ -13,7 +13,7 @@
                     </a>
                     <ul class="dropdown-menu">
                         <li><a href="#">Profile</a></li>
-                        <li><a href="accounts/logout">Logout</a></li>
+                        <li><a href="javascript:void(0);" class="user-logout">Logout</a></li>
```
lliss: I hate this javascript void stuff. You don't need an href to make a link. I defer to @mmcfarland but I don't see any purpose to this and I think it looks ugly.

kdeloach: IMO this should always link to the logout page so someone could right click on it and select "Copy URL" or maybe open it in a new tab. It's pretty easy to disable client-side if need be.

src/mmw/js/src/app.js
```diff
@@ -64,6 +65,34 @@ var App = new Marionette.Application({
 
     hideLoginModal: function() {
         $('#login').modal('hide');
+    },
+
+    userLogout: function() {
```
lliss: It feels like most of this should be in the user module.

src/mmw/js/src/core/templates/header.ejs
lliss: Good point @kdeloach. There's certainly something to be said for being able to actually log out without js.

src/mmw/js/src/core/views.js
```diff
@@ -59,6 +61,10 @@ var HeaderView = Marionette.ItemView.extend({
         // Defer requiring app until needed as it is not defined when
         // core.views are initialized (they are required in app.js)
         require('../app').showLoginModal();
+    },
+
+    userLogout: function() {
+        require('../app').userLogout();
```
kdeloach: Since this gets required more than once in this module, consider moving it with the rest of the require statements and updating these references.

rajadain: Can't move it with the rest of the require statements since `app` requires this file and would create circular dependencies. Probably should refactor and move this code away from this area.

src/mmw/js/src/app.js
lliss: App has a reference to user already so I think you'd probably want to do something like call user.logout() since there may be other areas of the application later that also cause a logout to occur.

src/mmw/js/src/core/views.js
lliss: Maybe this is a good case for Backbone.Wreqr which I think is already part of Marionette. We might be able to handle this through the radio feature:
https://github.com/marionettejs/backbone.wreqr

rajadain:
Should there be a dedicated logout page that is user visible, or should the user be redirected to the home page instead (and potentially be blasted with a login modal since they are no longer logged in)?

lliss:
I think there should be a dedicated logout page. And I don't really see the harm in a dedicated login page too. Password keeps like lastpass work so much better when there is a dedicated page and I think we should be conscious about supporting tools like these that improve security on the web (especially since many of us use them). But I think clicking that link should probably not send you to that page but do it approximately the way you have it. But in a pinch, it's nice to go to /logout and know that it will work.

src/mmw/js/src/core/views.js
kdeloach: Actually, I think we just need to pass a reference to App in the constructor of this view. It's not a best practice to have require statements at "run-time" like this.

mmcfarland: I think we're doing pretty well without introducing event based communication, I would like to try and just resolve the initialization dependence through restructuring somehow.  Maybe passing in App would be the simplist.

src/mmw/js/src/app.js
```diff
@@ -64,6 +65,34 @@ var App = new Marionette.Application({
...
+            // We silence the first change because we don't want to send two
+            // events. The event generated by the second change is enough.
+            App.user.unset('username', {silent: true});
+            App.user.set('guest', true);
+
+            // Let the user stay at `/` or `/analyze`, but other pages require
+            // an account so in those cases redirect to `/`
+            var router = require('./router').router;
+            var currentRoute = Backbone.history.getFragment();
+            if (currentRoute === '' || currentRoute === 'analyze') {
```
rajadain: Do we have a mechanism for identifying views that are logged-in user only? This if-condition seems to be a really flimsy place to list them.

rajadain:
Thank you all for your input, this has really helped me solve this better. I had to create a new PR because the old repo which made this one had to be blown away. The new one is available at #216

Issue #176, opened 2015-05-15
Allow selection of Area of Interest from map layer


mmcfarland:
Currently Congressional Districts are enumerated in the "select area" draw tool.  The new workflow has that dropdown being a list of available feature layers (CD, HUC, etc) that when selected add a polygon layer to the map.  Hovering, via UTFGrid, should present the name of the feature as a tooltip and clicking should fetch the geojson, load it as the AoI and reset the layer selector to empty.

PR #202, opened 2015-05-22
Predefined shape outlines.

jamesmcclain:
Instead of giving a selection of predefined shapes in the dropdown, give a selection of types of shapes (congressional districts, hucs, &c). When a type is chosen, the outlines of those shapes will appear on the map, allowing the user choose that way.

This patch assumes the existence of two environment variables on the `app` VM: `MMW_TILER_HOST` and `MMW_TILER_PORT`.  If they do not exist, the corresponding variables within Django are given the defaults `localhost` and `4000`, respectively.

Also, the name of the column containing the polygon in the modeling_district table has been renamed from `polygon` to `geom`.

Also, `src/mmw/mmw/settings/base.py`, `src/mmw/apps/modeling/views.py` and `src/tiler/server.js` have all been updated to support adding new geometry-bearing tables more easily.

**To Test:**
   1. Run migrations
   2. `reload_dev_data`
   3. You may have to ssh into the tiler VM and do a `sudo services mmw-tiler restart` after you apply this patch to make the tiler respond to the changes.
   4. If you have not done so recently, you may need to reprovision the app VM (changes were made to environment variables recently which will prevent the tiler from getting the correct services VM IP address unless you do this).
   5. If you are getting `500`s on the `boundary-layer` endpoint, try deleting your *.pyc files.
   6. I also recommend boosting your tiler RAM to 1024.

Connects to #176
Connects to #210


src/tiler/server.js
```diff
@@ -24,7 +24,7 @@ var config = {
         }
     },
     renderCache: {
-        ttl: 60000, // seconds
```
mmcfarland: This seems like a pretty big change to make without adding an explanation of why it was necessary in the commit message, might have some implications worth discussing.

jamesmcclain: You are right.  @lliss and I were testing yesterday, and that did not get changed back.

mmcfarland:
This is looking better, but I think there are some improvements we can make that will allow adding additional layers in the future easier and with less code.  Lets you, @lliss and I try to meet up briefly this afternoon to discuss the ideas.

jamesmcclain:
okay, sounds good

jamesmcclain:
This patch has been updated to allow new tables to be added and removed by only changing `settings/base.py` and `server.js`, which was the principal constraint that I have noted down from the meeting.  The `polygon` column has also been renamed to `geom` in the congressional district table.

jamesmcclain:
Also `modeling/views.py` has to be changed when a new table is added because it was not clear to me how to dynamically map from a table name or id to a model object without having to change that file anyway (willing to take direction on that).

src/mmw/apps/modeling/views.py
```diff
@@ -102,18 +105,32 @@ def _initiate_tr55_job_chain(model_input, job_id):
...
+                  'geometry': {'type': 'Polygon',
+                               'coordinates': coordinates}}
+    return Response(dictionary)
+
+
+@decorators.api_view(['GET'])
+@decorators.permission_classes((AllowAny, ))
+def boundary_layers(request, id=None, state=None):
+    tiler_prefix = '//'
+    tiler_host = getattr(settings, 'MMW_TILER_HOST')
```
kdeloach: Consider using `settings.MMW_TILER_HOST`.

src/mmw/js/src/app.js
```diff
@@ -62,16 +62,17 @@ var App = new Marionette.Application({
 
 function RestAPI() {
     return {
-        getPredefinedShapes: function() {
+        getPredefinedShapeTypes: function() {
             return $.ajax({
-                'url': '/api/modeling/congressional_districts',
+                'url': '/api/modeling/boundary_layers/',
```
kdeloach: We should prefer hyphens instead of underscores in urls.

src/mmw/mmw/settings/base.py
```diff
@@ -252,6 +252,13 @@ def get_env_setting(setting):
 # END URL CONFIGURATION
 
 
+# TILER
+MMW_TILER_HOST = environ.get('MMW_TILER_HOST', environ.get('MMW_TILER_IP', 'localhost'))  # noqa
```
kdeloach: Consider removing the default value or just having it default to localhost.

src/mmw/apps/modeling/views.py
```diff
@@ -102,18 +105,32 @@ def _initiate_tr55_job_chain(model_input, job_id):
...
-                                   'coordinates': coordinates}}
-        return Response(dictionary)
-    else:  # provide list of all ids
-        shapes = District.objects.order_by('state_fips', 'district_fips')
-        shapes = [{'id': shape.id, 'name': shape.name()} for shape in shapes]
-        dictionary = {'shapes': shapes}
-        return Response(dictionary)
+    district = get_object_or_404(District, id=id)
+    coordinates = []
+    for polygon in district.geom.coords:
```
kdeloach: Consider separating this out into a utility function to explain what's going on here.

src/mmw/js/src/draw/views.js
```diff
@@ -176,7 +169,55 @@ var PlaceMarkerView = Marionette.ItemView.extend({
     }
 });
 
-function clearLayer(layer) {
+function changeOutlineLayer(endpoint,select) {
+    var map = App.getLeafletMap(),
+        ol = App.outlineLayer,
+        grid = App.utfGrid;
```
kdeloach: Unclear why we are attaching things to `App`. We should try to keep these layers localized to this module / view.

src/mmw/js/src/draw/views.js
```diff
@@ -176,7 +169,55 @@ var PlaceMarkerView = Marionette.ItemView.extend({
     }
 });
 
-function clearLayer(layer) {
+function changeOutlineLayer(endpoint,select) {
```
kdeloach: Missing space in args list.

src/mmw/js/src/draw/views.js
```diff
@@ -176,7 +169,55 @@ var PlaceMarkerView = Marionette.ItemView.extend({
...
+        map.addLayer(ol);
+        grid = App.utfGrid = new L.UtfGrid(endpoint + '.grid.json?callback={cb}', { resolution: 4, maxZoom: 18 });
+        map.addLayer(grid);
+
+        grid.on('click', function (e) {
+            var shapeId = e.data ? e.data.id : null,
+                revertLayer = clearLayer();
+
+            if (select) {
+                select.disableTools();
```
kdeloach: Consider renaming `select` to `model` or `selectModel` for context.

mmcfarland:
Re: modifying the views file, I think we can avoid this by implementing a URL pattern similar to this:

`GET /bounday-layers` will return a list of display name & tiler endpoints for the layers
`GET /bounday-layers/<table id>/<feature id>` will return the GeoJSON for the feature identified by the table/layer

This way, we don't need to construct URLs for every new boundary layer and can remove all `congressional-district` URLs

jamesmcclain:
Okay, the issue that I was having with modifying the views file, is that at present an object called `District` is used to access the table.  What I am missing is a mechanism to dynamically map a name to one of these objects.  If that is not requested or required, then I know where to go from here.

mmcfarland:
We can access the DB with [raw sql] (https://docs.djangoproject.com/en/1.8/topics/db/sql/) still, however using the aliased table names and cross referencing from the list in settings.py for any http requests.  Essentially, the main view function will construct the appropriate query based on the url parameters and the mapping provided in settings.py, which is also the single definition of the layers available.  We can sketch it out again real quick if you want to.

src/mmw/apps/modeling/views.py
kdeloach: I replaced this with Django geojson serialization to support both polygons and multipolygons so there will probably be a conflict.

src/mmw/mmw/settings/base.py
kdeloach: The reason being that these environmental variables should be set during provisioning and it isn't worth it to make these too robust.

src/mmw/apps/modeling/views.py
jamesmcclain: I noticed.  I am working on fulfilling all of the requests, then I will rebase on top of develop and deal with the conflicts.

jamesmcclain:
The latest commit attempt to respond to all comments so far.

It appears to work well *except* for the following bug: after the shadow mask has been applied, it blocks click events inteded for the utf layer, preventing the click-to-select-district functionality from working (unless you happen to click in the unshaded portion).

The standard expedient of moving the utfgrid layer to the top upon creation does not seem to work, so I will need some consultation on this issue and/or to make a separate card for this bug.

kdeloach:
I can pair on this with you tomorrow morning.

mmcfarland:
The utf-grid and the boundary layers shouldn't conflict because they're not on at the same time.  After selecting an area of interest via a particular feature from the layer, the boundary layer should be removed and the geojson of the multipolygon continues as in other AoI selections, which is when the mask should appear.

mmcfarland:
Following theses steps:
* Provision app & tiler
* Run migrations
* Run reload_dev_data

Requests to tiler fail with `Cannot GET /0/10/298/385.png` and no interesting output in `/var/log/mmw-tiler.log both through the browser or when SSH'd into tiler and using `curl` against localhost:4000

Any additional steps needed?

jamesmcclain:
I am not sure, when I switch from master to this branch, I frequently revert the services VM to an earlier snapshot then rerun the migrations and reload the development data.

jamesmcclain:
I just remembered: you should probably do a `sudo servers mmw-tiler restart` on the tiler VM.

src/mmw/js/src/draw/views.js
```diff
@@ -148,7 +140,45 @@ var PlaceMarkerView = Marionette.ItemView.extend({
     }
 });
 
-function clearLayer(layer) {
+function changeOutlineLayer(endpoint, tableId, select) {
+    var map = App.getLeafletMap();
+
+    if (endpoint && tableId != undefined) {
+        var ol = App.outlineLayer = new L.TileLayer(endpoint + '.png', {maxZoom: 18});
```
lliss: Why is maxZoom set to 18? Is that the same as elsewhere in the app? We might want to consider changing this to a setting.

jamesmcclain:
Regarding the question of the utfgrid and the boundary layers being in conflict that was raised above:

The issue that I was seeing was with the utfgrid and the mask layer, not the boundary layer.  As things are at present, the mask layer can be on-screen while the user is trying to select a polygon.  If that is true, then the mask layer prevents click events from getting to the utfgrid.

jamesmcclain:
Thank you @kdeloach for the chat.  This latest patch includes changes so that the outline layers are cleared when the toolbar is disabled.

I think that I would prefer to leave the issue of the mask persisting after the user has gone backwards to a separate card and/or pull request.

src/mmw/apps/modeling/views.py
```diff
@@ -202,13 +207,29 @@ def _initiate_tr55_job_chain(model_input, job_id):
 
 @decorators.api_view(['GET'])
 @decorators.permission_classes((AllowAny, ))
-def district(request, id=None, state=None):
-    if id:  # query by unique id
-        district = get_object_or_404(District, id=id)
-        geojson = json.loads(district.polygon.geojson)
+def boundary_layers(request, table_id=None, obj_id=None):
+    if not table_id or not obj_id:
```
mmcfarland: Seems like we require `table_id` for either conditional, should this be an `and` and then another conditional to return a bad request response if it's not provided?

src/mmw/js/src/draw/views.js
```diff
@@ -148,7 +167,41 @@ var PlaceMarkerView = Marionette.ItemView.extend({
     }
 });
 
-function clearLayer(layer) {
+function changeOutlineLayer(endpoint, tableId, model) {
+    var map = App.getLeafletMap();
+
+    if (endpoint && tableId != undefined) {
+        var ol = App.outlineLayer = new L.TileLayer(endpoint + '.png');
```
kdeloach: This might be an artifact from an older version of the code, but it looks like something is still assigned to `App` here and on the next line and it doesn't get user anywhere.

src/mmw/js/src/draw/tests.js
```diff
@@ -78,7 +79,7 @@ describe('Draw', function() {
 
             var $li = $($el.find('#select-area-region li a').get(0));
             $li.trigger('click');
-
+            App._mapView._leafletMap.fireEvent('click', {'latlng': [29.979, 31.134]});
```
kdeloach: Consider replacing with `App.getLeafletMap`. By convention, properties with underscore prefix aren't meant to be accessed outside of their modules.

src/mmw/js/src/draw/views.js
```diff
@@ -19,13 +19,30 @@ var ToolbarView = Marionette.LayoutView.extend({
...
         placeMarkerRegion: '#place-marker-region'
     },
 
+    modelEvents: {
+        'change:toolsEnabled': 'manageOutlineLayer'
+    },
+
+    initialize: function() {
+        var map = App.getLeafletMap();
+        var ofg = this.model.outlineFeatureGroup = L.featureGroup();
```
kdeloach: I don't know if anyone else has strong feelings about this assignment shorthand, but I find it's generally less readable than just having 1 assignment per line.

mmcfarland: It is, but also model attributes should be set using the `.set` method and not applied directly to the object.

src/mmw/apps/modeling/views.py
```diff
@@ -202,13 +207,29 @@ def _initiate_tr55_job_chain(model_input, job_id):
 
 @decorators.api_view(['GET'])
 @decorators.permission_classes((AllowAny, ))
-def district(request, id=None, state=None):
-    if id:  # query by unique id
-        district = get_object_or_404(District, id=id)
-        geojson = json.loads(district.polygon.geojson)
+def boundary_layers(request, table_id=None, obj_id=None):
```
kdeloach: Consider adding some additional boundary condition checks. Poking around with different URLs I got:

1. KeyError with "/api/modeling/boundary-layers/1/0/"
2. TypeError with "/api/modeling/boundary-layers/0/0/"

src/mmw/js/shim/leaflet.utfgrid.js
```diff
@@ -0,0 +1,359 @@
+L.Util.ajax = function (url, cb) {
```
caseypt: Since we're shimming this in, we should add the license here. https://github.com/danzel/Leaflet.utfgrid/blob/master/MIT-LICENCE.txt

src/mmw/js/src/draw/views.js
```diff
@@ -34,12 +51,22 @@ var ToolbarView = Marionette.LayoutView.extend({
         this.placeMarkerRegion.show(new PlaceMarkerView({
             model: this.model
         }));
+    },
+
+    manageOutlineLayer: function() {
+        var enabled = this.model.get('toolsEnabled');
+        var ofg = this.model.outlineFeatureGroup;
```
caseypt: App style is one `var` statement with commas

src/mmw/apps/modeling/views.py
```diff
@@ -202,13 +207,29 @@ def _initiate_tr55_job_chain(model_input, job_id):
...
+            retval['endpoint'] = urljoin(tiler_base, index + tiler_postfix)
+            return retval
+
+        layers = [augment(i, d)
+                  for i, d in settings.BOUNDARY_LAYERS.items()]
+        return Response(layers)
+    else:
+        table_name = settings.BOUNDARY_LAYERS[table_id]['table_name']
+        cursor = connection.cursor()
+        query = 'SELECT geom FROM %s WHERE id = %s' % (table_name, obj_id)
```
kdeloach: Consider passing `obj_id` as a parameter to the query to avoid any possibility of a SQL injection. Ex. `cursor.execute(query, (obj_id,))` instead of string interpolation. The `table_name` value comes from a whitelist so I don't see any risk there.

src/mmw/js/src/draw/views.js
kdeloach: Yeah, this feature group is a special case. It isn't an attribute that needs to be persisted to the model necessarily... but I guess it wouldn't hurt to make it an attribute either. I'm still thinking of a good design pattern for handling extra layers outside of the map view.

kdeloach:
After running `reload_dev_data` and reprovisioning my app and tiler VMs I'm able to see the congressional district boundary layer.

kdeloach:
A lot of nice changes here!

kdeloach:
Consider retroactively applying the `+` label to #210, assigning it a point value, and moving it to the correct milestone.

src/mmw/js/src/draw/views.js
mmcfarland: Yeah, it would be good to consolidate them.  Since `leafletMap` is accessible to anyone, it makes sense that other modules would be interested in turning on/off layer display.

mmcfarland:
Agree - looking good, and the boundary layers are loading and working for me.  I think just this last round of small cleanups.

src/mmw/apps/modeling/migrations/0004_merge.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
```
mmcfarland: Not sure why 003/4 migrations were generated - they look like no-ops.  Can they be removed?

mmcfarland:
+1

Issue #186, opened 2015-05-19
Add data for new land uses


rajadain:
Add in data for `chaparral`, `shortGrass`, `tallGrass`, and `desert` types, and update `src/mmw/apps/water_balance/static/data/water_balance/model.csv` with it. See #185 for more detailed explanation.

Issue #185, opened 2015-05-19
Update model data for mini-app, fix data issues


rajadain:
The new mini-app uses a CSV file of data taken from the old Flash app, which is stored in `src/mmw/apps/water_balance/static/data/water_balance/model.csv`. The structure of this data is as follows:

```
P,land,soil,ET,I,R
```

Where `P` is precipitation and has values in the set `[0.5, 1.0, 2.0, 3.2, 8.0]`, `land` is the [NLCD code of different land uses](http://www.mrlc.gov/nlcd11_leg.php), `soil` has the following mapping:

```
0 = sand
1 = loam
2 = sandyClay
3 = clayLoam
```

And `ET`, `I`, and `R` are floating point values corresponding to Evapotranspiration, Infiltration, and Runoff respectively.

**Incongruous Land Code Interpretation**

The Flash app interpreted the `land` values as follows:

```actionscript
	var landCodes = {
		21: LandCovers.LOW_INTENSITY_RESIDENTIAL,
		22: LandCovers.HIGH_INTENSITY_RESIDENTIAL,
		23: LandCovers.COMMERCIAL_INDUSTRIAL,
		43: LandCovers.FOREST,
		85: LandCovers.TURF_GRASS,
		81: LandCovers.XYZ,
		71: LandCovers.GRASSLAND,
		82: LandCovers.ROW_CROPS,
		91: LandCovers.WETLANDS
	}
```

But the official definition of these codes is different: http://www.mrlc.gov/nlcd11_leg.php

This official description is the one used by TR-55: https://github.com/WikiWatershed/tr-55/blob/develop/tr55/tables.py#L387-L414, and also by the new mini-app:

```javascript
    landMap: {
        '21': 'turfGrass',
        '22': 'lir',
        '23': 'hir',
        '24': 'commercial',
        '41': 'forest',
        '71': 'grassland',
        '81': 'pasture',
        '82': 'rowCrops',
        // TODO The following are missing NLCD identifiers.
        // Replace X? with actual numbers when that data is available.
        'X0': 'chaparral',
        'X1': 'tallGrass',
        'X2': 'shortGrass',
        'X3': 'desert'
    },
```

**Missing Data**

As can be seen in the JavaScript definition above, we use the code `24` for `commercial` land use, but our current data set is missing data for that code. Also, we have added four new land uses: `chaparral`, `tallGrass`, `shortGrass`, and `desert`, data for which is also not available.

**Faulty Data**

The data model implies:

[...]

PR #206, opened 2015-05-22
Add missing data to mini-app.

jamesmcclain:
Missing data have been added to the mini-app.

"Short grass" is of type 81 [1], "chaparral" is of type 52 [2], "desert"
is of type 31 [2], and "tall grass" is of type 85 [1].

To test, it is import to copy the `model.csv` file after the patch has been applied so that the newly-modified version of `model.csv` to the client.  Type:

`vagrant ssh app`

Once there, type:
`cp /opt/app/apps/water_balance/static/data/water_balance/model.csv /var/www/mmw/static/data/water_balance/model.csv`

This copying process is done automatically by the provisioner, but you must copy this file by hand unless you wish to re-provision.

[1] http://www.asprs.org/a/publications/pers/2003journal/december/2003_dec_1387-1397.pdf p. 1389
[2] http://www.mrlc.gov/nlcd11_leg.php

Connects to WikiWatershed/model-my-watershed#185
Connects to WikiWatershed/model-my-watershed#186


src/mmw/bundle.sh
```diff
@@ -113,6 +118,7 @@ done
 VAGRANT_COMMAND="cd /opt/app && \
     $ENSURE_DIRS_EXIST && { \
     $COPY_IMAGES_COMMAND &
+    $COPY_SIMPLE_APP_DATA_COMMAND &
```
rajadain: We actually don't need any changes to `bundle.sh`. Because `model.csv` lives under the `static` directory in a Django app, it is automatically packaged and deployed by the `collectstatic` command:

```bash
./scripts/bundle.sh collectstatic --noinput
```

This command is run during provisioning, thus we don't need the additional commands in `bundle.sh`.

src/mmw/js/src/water_balance/models.js
```diff
@@ -2,20 +2,17 @@
...
+        '22': ['lir'],
+        '23': ['hir'],
+        '24': ['commercial'],
+        '41': ['forest'],
+        '71': ['grassland'],
+        '81': ['pasture', 'shortGrass'],
+        '82': ['rowCrops'],
+        '52': ['chaparral'],
+        '31': ['desert'],
+        '85': ['tallGrass']
```
rajadain: Consider sorting this by ascending NLCD code to ease reading.

src/mmw/bundle.sh
jamesmcclain: That command does not appear to do anything.  I delete /var/www/.../model.csv on the app vm, ran that command, and model.csv has not returned.

rajadain:
Well done! The mini app now works perfectly. The following was performed as part of this review:

 * Ran all tests successfully
 * Tried a variety of combinations and they all rendered
 * Height of Water Column stays constant, and correctly represents the component share
 * Numbers are consistent with the "under-the-hood" behavior of the old flash app, although not with the apparent behavior, since the old flash app interpreted the NLCD codes incorrectly as described in #185. This is expected and correct.

src/mmw/bundle.sh
rajadain: Sorry, my mistake. The command that I was looking for is

```
./scripts/manage.sh collectstatic --noinput
```

It collects all static assets to the Django static directory. See here for details: https://docs.djangoproject.com/en/1.8/ref/contrib/staticfiles/

jamesmcclain:
Okay, I removed the change to `bundle.sh` and changed the testing instructions above.  I will remember to put the correct testing instructions in the commit message when I squash this.

rajadain:
Splendid. +1

jamesmcclain:
Thank you for the information about the provisioning process, I was unaware.  Thank you for reviewing.

src/mmw/bundle.sh
kdeloach: Sorry if this is out of context, but I just wanted to point out that the bundle script was designed to generate output directly to `DJANGO_STATIC_ROOT` to eliminate the need to run `collectstatic`.

PR #207, opened 2015-05-22
Add AnsibleGalaxyHelper to remove noise from Vagrantfile

hectcastro:
I broke out the Ansible Galaxy related cruft into a Ruby module that gets required in the main `Vagrantfile`. The goal here is just to clean things up before the fold so that people can get to the things they're probably more interested in quicker.

Need to look into getting this into a Vagrant plugin or something at some point.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -2,47 +2,11 @@
 # vi: set ft=ruby :
 
 Vagrant.require_version ">= 1.6"
-require "yaml"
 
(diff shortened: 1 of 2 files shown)


mmcfarland:
:+1: Preformed successful `vagrant reload` and a successful provision of app VM after manually deleting the `azavea.*` ansible roles.  All were fetched and the readability of the Vagrantfile is improved.

kdeloach:
Neat.

Issue #163, opened 2015-05-15
Refactor shared provisioning tasks and attributes across VMs


hectcastro:
Currently there is a significant amount of duplication in Ansible attributes and tasks across virtual machines. Some more thought should be given to how we can make use of all the attributes and tasks without as much repetition.

Some examples:

- `app_config` (including ITSI portal attributes)
- Python dependency installs
- Geospatial library installs
- `envdir` setup

Issue #110, opened 2015-04-29
Missing template variables for Django static and media root


hectcastro:
The `envdir` files for the Django static and media roots on at least the `worker` virtual machine are invalid because the variable they're attempting to template does not exist:

```
TASK: [model-my-watershed.celery-worker | Configure application] **************
ok: [worker] => (item={'content': u'mmw', 'file': 'MMW_DB_NAME'})
ok: [worker] => (item={'content': u'mmw', 'file': 'MMW_DB_USER'})
ok: [worker] => (item={'content': u'mmw', 'file': 'MMW_DB_PASSWORD'})
ok: [worker] => (item={'content': u'33.33.34.30', 'file': 'MMW_DB_HOST'})
ok: [worker] => (item={'content': u'5432', 'file': 'MMW_DB_PORT'})
ok: [worker] => (item={'content': u'test_mmw', 'file': 'MMW_TEST_DB_NAME'})
ok: [worker] => (item={'content': u'33.33.34.30', 'file': 'MMW_STATSD_HOST'})
ok: [worker] => (item={'content': u'33.33.34.30', 'file': 'MMW_CACHE_HOST'})
ok: [worker] => (item={'content': u'6379', 'file': 'MMW_CACHE_PORT'})
ok: [worker] => (item={'content': u'mmw.settings.development', 'file': 'DJANGO_SETTINGS_MODULE'})
ok: [worker] => (item={'content': u'{# app_static_root #}', 'file': 'DJANGO_STATIC_ROOT'})
ok: [worker] => (item={'content': u'{# app_media_root #}', 'file': 'DJANGO_MEDIA_ROOT'})
ok: [worker] => (item={'content': u'2.1.3', 'file': 'DJANGO_POSTGIS_VERSION'})
ok: [worker] => (item={'content': u'03448aceb9a9825188243881ae783577', 'file': 'DJANGO_SECRET_KEY'})
```

Note the value of `DJANGO_STATIC_ROOT` and `DJANGO_MEDIA_ROOT`.

PR #208, opened 2015-05-22
Refactor shared provisioning tasks into a base role

hectcastro:
At a high level, this moves most things that are shared by more than one type of server into a shared base role that all server types depend on. Types of things bundled into the base role include:

- General configuration settings (database, cache, etc.)
- Vagrant image configuration management cruft removal (Chef, Puppet)
- Service accounts, account permissions
- Core geospatial library dependencies
- PostgreSQL client support (`psql`, etc.)

Connects #110
Connects #163

---

**Testing**

I fully expect this pull request to break things. Right now everyone's static asset generation occurs as `root`. This pull request tries to change that, but it won't be able to overwrite the current files.

Along those lines, a lot of core things have been moved around, so I tried a bunch of tests to ensure that everything is still working. Kibana still appears to be receiving logs, Graphite receiving metrics, and the application interactions in the browser all seems good. Please test things you anticipate would be broken by these changes.


deployment/ansible/roles/model-my-watershed.app/defaults/main.yml
```diff
@@ -1,26 +1,17 @@
...
-  - { file: "MMW_STATSD_HOST", content: "{{ statsite_host }}" }
-  - { file: "MMW_CACHE_HOST", content: "{{ redis_host }}" }
-  - { file: "MMW_CACHE_PORT", content: "{{ redis_port }}" }
-  - { file: "DJANGO_SETTINGS_MODULE", content: "{{ django_settings_module }}" }
-  - { file: "DJANGO_STATIC_ROOT", content: "{{ app_static_root }}" }
-  - { file: "DJANGO_MEDIA_ROOT", content: "{{ app_media_root }}" }
-  - { file: "DJANGO_POSTGIS_VERSION", content: "{{ app_postgis_version }}" }
-  - { file: "DJANGO_SECRET_KEY", content: "{{ app_secret_key }}" }
-  - { file: "ITSI_CLIENT_ID", content: "{{ itsi_client_id }}" }
-  - { file: "ITSI_SECRET_KEY", content: "{{ itsi_secret_key }}" }
```
mmcfarland: The ITSI values did not get converted to the new data structure and seem to be the only ones left out.  Are they handled differently now? I don't see them addressed in other commits.

hectcastro: I missed them in the conversion. Fixed in https://github.com/WikiWatershed/model-my-watershed/commit/892d54c4a00aa4d9aa01a0cb8ca1ab1be6f176cf.

mmcfarland:
+1 tested.  App features all work, including ITSI oAuth2 integration.  I started with a `destroy` to avoid any file permission errors, which I suspect is the reason for the failing PR builder.  Ancillary services all behave well, too.

PR #211, opened 2015-05-22
Remove initial zoom when page loads

kdeloach:
This code was a proof of concept to show how we could restore saved state
when the page loads. It was supposed to be temporary but never got removed...
until now.


hectcastro:
@azavea-bot rebuild

lewfish:
It now zooms to an area around northern DE and central PA. +1, tested.

Issue #193, opened 2015-05-20
Remove sbo-selenium


mmcfarland:
We shouldn't need this plugin for MMW, it relates largely to sauce labs integration.

https://github.com/azavea/model-my-watershed/blob/develop/src/mmw/requirements/test.txt#L7

PR #212, opened 2015-05-22
Remove sbo-selenium. 

jamesmcclain:
All occurrences of the string sbo-selenium have been removed.  The tests
continue to run.


mmcfarland:
+1 Tests pass.

@hectcastro - don't know if you needed to do anything after this was removed, or if you were just hoping it'd get resolved.

jamesmcclain:
I'll wait for Hector's response.

hectcastro:
Looks good to me. It will probably remain installed on the Jenkins machines until we destroy them, but I don't see reason why we need to do that right now. Maybe we can destroy them at the end of the day so that it doesn't slow everyone down on the next build.

hectcastro:
Connects #193

jamesmcclain:
Okay, sounds good.  I will go ahead and merge.

Issue #197, opened 2015-05-21
Bake CSRF token in all Backbone requests


rajadain:
When we make AJAX requests to Django from Backbone, especially when a user is logged in, each request must specify an X-CSRFToken header. This has been implemented during user login as follows:

```javascript
    getUserOrShowLogin: function() {
        var csrftoken = helpers.Cookie.get('csrftoken');

        this.user.fetch({
            beforeSend: function(xhr) {
                xhr.setRequestHeader('X-CSRFToken', csrftoken);
            }
        });
    },
```

So that we don't have to specify this `beforeSend` function and have code duplication for every `.get()` or `.fetch()` or `.save()` on every model and collection, we should bake this mechanism into [`Backbone.sync`](http://backbonejs.org/#Sync).  This is one example of how to do it: https://gist.github.com/gcollazo/1240683

We should probably do this in `src/mmw/js/shim/backbone.js` so it is always imported.

Once this fix is available, please update all occurrences of the above code since we don't need a second `X-CSRFToken` header.

It should be noted that even if we override `Backbone.sync` to include this header, any `$.ajax()` requests will still require specifying the token. There exists even wider implementations which bake in the token to all `$.ajax()` requests, such as https://gist.github.com/justinperkins/3960219, and this will work for both `$.ajax()` and all Backbone requests since `Backbone.sync` uses `$.ajax()` under the hood. However, if we make any `$.get()` requests to third party domains, we will expose the CSRF token. If we are certain that will not be the case, we can consider this second option as well.

kdeloach:
Here's the solution used on NYC: https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/js/src/csrf.js

rajadain:
Great, that seems to be akin to the second option mentioned in the Issue description, one that overrides `$.ajax()` across the app. The drawback being that it could potentially expose the CSRF Token to 3rd party sites if any cross domain requests are made, but perhaps [this line](https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/js/src/csrf.js#L32) explicitly forbids those? Perhaps @jwalgran could confirm this?

In this case we should definitely adopt the same approach here.

PR #213, opened 2015-05-22
Bake CSRF token in all ajax requests

kdeloach:
Extends jQuery ajax options to include the `X-CSRFToken` header in each
ajax request.

Connects #197


mmcfarland:
Does this behave well in the case of a non-authenticated user visiting the site and not having this cookie value available?

kdeloach:
Good point. We should test that this allows authenticated and non-authenticated users to queue jobs.

rajadain:
Consider adding `$.ajaxSetup()` to the new `csrf` helper file, so that for any module that needs it can simply `require` it and not have to also initialize it. Is there a use case where we require it but don't initialize it?

kdeloach:
The CSRF token actually changes after logging in. Since a page refresh doesn't happen, we need to update the CSRF token attached to all ajax requests. Requiring the csrf module a second time would have no effect.

rajadain:
If we moved this line:

```
var csrftoken = getCookie('csrftoken');
```

into the `beforeSend` function, it would always get the current CSRF token, and then requiring `csrf` would be enough.

kdeloach:
Great suggestion, will update.

rajadain:
Tested with both logged in and not logged in users. GET requests consistently don't get CSRF token headers and still work in both cases, and POST requests consistently get CSRF token headers and also work correctly in both cases.

It should be noted that CSRF Tokens are available for all users, and are not a function of a users authentication status. They do not verify if a user is logged in, but instead that the JavaScript request is coming from a page that the served, and is not being surreptitiously accessed by a third party. They prevent Cross Site Request Forgery.

Logged in users have a `sessionid` cookie set. The `csrftoken` cookie is set for all users.

Consider deleting the `helpers.js` file since this subsumes it.

rajadain:
Can we further reduce redundancy by requiring this in some universal dependency like the `shim/backbone.js` so that we don't have to `require` it in all modules? Once the behavior of `$.ajaxSetup` is defined it should just apply to all `$.ajax()` calls after it, right?

kdeloach:
Updated based on your suggestions.

Since main.js is the entry point to this application, attaching the csrf ajax options there makes the most sense to me. Note that this is a universal change that affects all ajax calls throughout the app.

rajadain:
:+1: looks really great! I've noted your comment about rebasing #216 on top of this, will edit the content there to work with this.

Issue #215, opened 2015-05-22
Which front-end pages require authentication?


rajadain:
The issue #173 asks for users to be redirected to the home page if they are on a page which requires authentication. How do we figure out which ones are they? Or if the current page requires authentication?

mmcfarland:
This is probably highly nuanced, but as a starting point:

* "My Projects" - listing of a users saved projects
* Private, saved project page
    * Either "model" or "compare"
    * It's further tricky because Public projects can be viewed without auth, but are read only and should have modification tools available.
* Account Profile pages (change password, details, etc)

It may be worth simply having an edict that logging out takes you to the root map page, perhaps retaining an area of interest.

rajadain:
By "retaining an area of interest" do you mean that the polygon should still be there? That can be potentially problematic since there is no button that the user can press to go from having the polygon to the `analyze` view. This is done automatically in the usual case, when the user is forwarded to the `analyze` view as soon as the polygon is drawn. However, if the user is on the home page with a drawn polygon, the only way for them to get to the analyze page is to draw another polygon, thus losing the work they have currently.

mmcfarland:
Correct, but we do have an issue for that already so it's in the works:
https://github.com/WikiWatershed/model-my-watershed/issues/121

rajadain:
Added b3aa25f on #216 which uses the simple edict to redirect all logouts to the home page while retaining the area of interest. Thank you for the clarifications.

Issue #214, opened 2015-05-22
Logging out should redirect user to homepage and open login prompt.


rajadain:
We have a static log out page, made for #173, which needs to be designed. It will be available at [`/user/logout`](http://localhost:8000/user/logout).

caseypt:
According to the latest mock ups by @ctaylo37, logging out redirects the user to the homepage. https://marvelapp.com/ih686e#6443208

caseypt:
Sorry, wrong button. Updating title.

ctaylo37:
Yes, logout should redirect to the homepage.

rajadain:
Very well, I will make the changes once #225 is merged in.

PR #216, opened 2015-05-22
AJAX User Logout Part 2

rajadain:
Connects #173 

Revises #201 

I'm creating a new PR because my old private repo was blown away as a part of the transition. For reference, the old PR is #201.

**Testing Notes**
Login as a user. Click name in the header and logout. Verify that the username now has changed to a Login button. Click the login button to make the login modal show up. Refresh the page without logging in, and it should identify you as not logged in and pop up the login modal.
Login as a user. Right click the logout link and open in new tab. Refresh the first tab, and it should identify you as not logged in and pop up the login modal.

**Further Work**
 * This PR has an unstyled log out page. This probably needs some design attention, and input from @mmcfarland and @ctaylo37 for how best to handle. Created #214 for this.
 * The original issue asks for the user to be redirected the home page in case they are at a page which is for logged-in users only. I don't have a clear idea of which pages this would apply to, and once we have that we will have to figure out how to mark those pages as requiring authentication within Backbone / Marionette. Created #215 for this.

### DIFF
--- a/src/mmw/apps/user/templates/user/logout.html
+++ b/src/mmw/apps/user/templates/user/logout.html
@@ -1,0 +1,6 @@
+{% extends 'base.html' %}
+
+{% block content %}
+    <h1>You have been logged out.</h1>
+
+{% endblock content %}
--- a/src/mmw/apps/user/views.py
+++ b/src/mmw/apps/user/views.py
@@ -1,8 +1,8 @@
 from rest_framework.response import Response
-from django.contrib.auth import authenticate, login
+from django.contrib.auth import authenticate, login, logout
 from django.contrib.auth.models import User
 from django.core.urlresolvers import reverse
-from django.shortcuts import redirect
+from django.shortcuts import redirect, render_to_response
 
 from rest_framework import decorators
 from rest_framework.permissions import AllowAny
@@ -45,6 +45,18 @@ def ajax_login(request):
         status = 400
 
     return Response(data=response_data, status=status)
+
+
+@decorators.api_view(['GET'])
+@decorators.permission_classes((AllowAny, ))
+def user_logout(request):
+    logout(request)
+
+    if request.is_ajax():
+        response_data = {'guest': True, 'result': 'success'}
+        return Response(data=response_data)
+    else:
+        return render_to_response('user/logout.html')
 
 
 itsi = ItsiService()
--- a/src/mmw/js/src/core/templates/header.ejs
+++ b/src/mmw/js/src/core/templates/header.ejs
@@ -13,7 +13,7 @@
                     </a>
                     <ul class="dropdown-menu">
                         <li><a href="#">Profile</a></li>
-                        <li><a href="/user/logout">Logout</a></li>
+                        <li><a href="/user/logout" class="user-logout">Logout</a></li>
                     </ul>
                 </li>
             <% } %>
--- a/src/mmw/js/src/core/templates/header.ejs
+++ b/src/mmw/js/src/core/templates/header.ejs
@@ -13,7 +13,7 @@
                     </a>
                     <ul class="dropdown-menu">
                         <li><a href="#">Profile</a></li>
-                        <li><a href="accounts/logout">Logout</a></li>
+                        <li><a href="/user/logout">Logout</a></li>
                     </ul>
                 </li>
             <% } %>
--- a/src/mmw/js/src/user/models.js
+++ b/src/mmw/js/src/user/models.js
@@ -7,7 +7,39 @@ var UserModel = Backbone.Model.extend({
         guest: true
     },
 
-    url: 'user/ajaxlogin'
+    url: 'user/ajaxlogin',
+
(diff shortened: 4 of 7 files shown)


rajadain:
Connects #173

src/mmw/apps/user/templates/user/logout.html
```diff
@@ -0,0 +1,6 @@
+{% extends 'base.html' %}
+
+{% block content %}
+    <h1>You have been logged out.</h1>
+
+{% endblock content %}
```
kdeloach: I would get an opinion from @ctaylo37 to see if we should implement a separate logout page. In other projects we redirect to the login page with `?next_url=` set to whatever page you were just looking at.

src/mmw/apps/user/views.py
```diff
@@ -47,6 +47,18 @@ def ajax_login(request):
...
+@decorators.permission_classes((AllowAny, ))
+def user_logout(request):
+    logout(request)
+
+    if request.is_ajax():
+        response_data = {'guest': True, 'result': 'success'}
+        return Response(data=response_data)
+    else:
+        return render_to_response('user/logout.html')
+
```
kdeloach: I like this much better than having to maintain separate urls/views for ajax vs non-ajax requests. We should consider doing the same for "ajaxlogin" in the future.

src/mmw/js/src/user/models.js
```diff
@@ -1,13 +1,54 @@
...
+    // be specified upon usage. They both update the user model, so any event
+    // listeners that subscribe to the `sync` event will be triggered.
+    login: function(attrs) {
+        return this.fetch({
+            type: 'POST',
+            url: 'user/ajaxlogin',
+            data: attrs,
+            beforeSend: function(xhr) {
+                var csrftoken = helpers.Cookie.get('csrftoken');
+                xhr.setRequestHeader('X-CSRFToken', csrftoken);
```
kdeloach: Consider rebasing on top of #213 to prevent merge conflicts.

kdeloach:
+1

kdeloach:
I merged #213 so you should be able to rebase on top of develop now.

rajadain:
Rebased on top of @kdeloach's work on #213 by removing inline CSRF references, now that they have been configured globally.

lliss:
Is it done elsewhere in the app? It seems fine to me.

lliss:
+1. Tested.

caseypt:
I think we should avoid using `window.MMW.*` in our code. The purpose was to have it there for debugging. I don't believe we are doing this else where in the application.

We should consider refactoring our `main.js`, `app.js`, etc to be able to get access to the global objects we need through `require` and avoid have circular imports.

mmcfarland:
I agree about accessing off of `window` for the same reason we avoid other global variables.  Refactoring to support requiring seems like the way to go.

rajadain:
I've added two options:

 1. d96f969 simply requires `router` in `core/views`
 2. 0634bb0 is a more elaborate refactoring, in which we move `router` from `main` to `App`, thus giving everyone who has access to `App` to be able to use the `router`. Additionally, we inject `App` into `HeaderView` at initialization, allowing it to call methods defined in `App` without having to `require` it inline.

Honestly I'm unsure which is the better one. The first one is easy and simple, but is it too simplistic? The second one does a slightly better job of reducing `require` statements, since instead of doing it in two places `main` and `core/views` we now do it in only one place `App`, and reference it or inject it. But does this cause unnecessary coupling between these modules? Is the best way somewhere in between?

lliss:
I prefer the first solution.

src/mmw/apps/user/views.py
lliss: I agree. This is nice.

rajadain:
Very well, I've squashed the first solution into the main branch and discarded the alternative. Will merge after a final okay.

src/mmw/js/src/core/templates/header.ejs
```diff
@@ -13,7 +13,7 @@
                     </a>
                     <ul class="dropdown-menu">
                         <li><a href="#">Profile</a></li>
-                        <li><a href="accounts/logout">Logout</a></li>
+                        <li><a href="/user/logout" class="user-logout" onclick="return false;">Logout</a></li>
```
lliss: I don't think this really bothers anyone else, but I prefer keep the html pure and then call ```preventDefault``` in the handler.

lliss:
+1. Tested.

src/mmw/js/src/core/templates/header.ejs
mmcfarland: Agreed.  I think event handlers should be implemented in js files, not html/templates, so it's easier to find and read the behavior.

rajadain:
Switched to using `preventDefault` in the handler in 4d287d1. Since that was the only issue and it is now fixed, seems like this is ready to merge!

lliss:
Great. Nice work.

PR #217, opened 2015-05-23
Faster bundle

kdeloach:
I've added a few flags to the bundle script which should make rebuilding a lot faster.

1. The vendor bundle rarely changes so it is no longer generated by default. Use the `--vendor` flag instead.
2. Test bundles are no longer generated by default. Use the `--tests` flag instead.
3. Created a vendor bundle for tests.
4. Moved d3 to the vendor bundle.

**Note:** You'll need to reprovision the app VM to use testem since a vendor bundle has been added to the test harness.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ After pulling in new commits, you may need to run the following two commands:
 ```bash
 $ ./scripts/manage.sh migrate
 $ ./scripts/manage.sh reload_dev_data
+$ ./scripts/bundle.sh
 ```
 
 See debug messages from the web app server:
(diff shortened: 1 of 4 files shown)


jamesmcclain:
Perhaps this should include an update to the README file, as well.

kdeloach:
Good idea, will update the README.

mmcfarland:
Not sure if it would push us over the reasonable number of flags for this script, but the water_balance JS will probably, for the rest of the project, only need to be rebundled very occasionally as well.

kdeloach:
Updated the README.

Since the water balance app is so small (only 2 files) I'd like to hold off on adding a flag just for that.

mmcfarland:
Testing now

lewfish:
Sorry, I didn't realize I was assigned to this until now.

mmcfarland:
+1 Much faster for regular workflow, tests still run and app functions normally.

Issue #180, opened 2015-05-15
Add front-end views for ITSI registration, error


rajadain:
The work done for #179 expects a front-end registration for new users signing in using ITSI to be available at `/register`, and in case of errors with ITSI portal to redirect to a view at `/error-itsi`. In case these views end up having different names, please update the references in `apps.user.views.itsi_register` and `apps.user.views.itsi_auth`.

**Register**
A user has signed in with ITSI, but we don't have an account corresponding to that login. This page, which is expected to be available at `/register`, should be similar to the regular user sign up which is being created for #152, but instead POST data to `/user/itsi/register`, which will link it to the already authenticated ITSI account. Currently the back-end view `apps.user.views.itsi_register` expects three parameters: `username`, `first_name` and `last_name`. If adding more, or changing these, please update the view as well.

**Error**
In case there is an error in ITSI interaction, we should notify the user. Dumping them back to the home page is not helpful. Thus, there should be a `/error-itsi` route which takes them to the home page, but also displays a modal explaining the error and suggesting alternatives.

The route `/error-itsi` was chosen over the friendlier `/itsi-error` in the expectation of many error routes to come in the future, and having them all start with the same `error-` prefix would make managing them easier. If this assumption does not apply, feel free to use a more appropriate name for the route.

Issue #152, opened 2015-05-13
Implement sign-up and forgot username/password


caseypt:
See mockups.

Will need to implement front-end views and back-end views to handle this over AJAX instead of through the Django registration forms.

PR #220, opened 2015-05-26
ITSI Login Error

rajadain:
This PR presents a tentative approach for route based error handling, with the immediate application of handling route based errors due to ITSI OAuth failures. It can be outlined as follows:

 * All `/error[.*]` routes will be handled by Django to render the main Backbone app
 * The Backbone app will pass these routes to an `ErrorController`
 * The `ErrorController` has a list of error `types` from which it looks up a specific `ErrorHandler` and fires it.

The following questions are raised by this work:

 * Currently the `ErrorHandler` for ITSI just pops up an `alert` message, which is rather inelegant. We should probably have design guidelines for showing error messages. Created #224 for this.
 * What should happen if the `ErrorController` is not passed in any error `type`, e.g. `/error` or `/error/`? Currently this is handled by `ErrorHandler.generic` and pops up a bland `alert` box.
 * What should happen if the `ErrorController` cannot find a matching error `type`, e.g. `/error/xyz`? Currently this is handled by `ErrorHandler.unknown` and silently logs the error to the console.

If we find `ErrorController` growing into having models and views, we might consider moving it out of the generic `controllers.js` into its own folder.

This PR addresses part of #180, namely the error view. I will wait for work on #152 to finish before adding a registration view.

Connects #180

### DIFF
--- a/src/mmw/apps/home/urls.py
+++ b/src/mmw/apps/home/urls.py
@@ -12,4 +12,5 @@ urlpatterns = patterns(
     url(r'^analyze$', home_page, name='home_page'),
     url(r'^model$', home_page, name='home_page'),
     url(r'^compare$', compare, name='compare'),
+    url(r'^error', home_page, name='home_page'),
 )
--- a/src/mmw/apps/user/views.py
+++ b/src/mmw/apps/user/views.py
@@ -63,14 +63,14 @@ def itsi_auth(request):
 
     # Basic validation
     if code is None:
-        return redirect('/error-itsi')  # TODO Make this Backbone route
+        return redirect('/error/itsi')
 
     try:
         session = itsi.get_session_from_code(code)
         itsi_user = session.get_user()
     except:
         # In case we are unable to reach ITSI and get an unexpected response
-        return redirect('/error-itsi')   # TODO Make this Backbone route
+        return redirect('/error/itsi')
 
     user = authenticate(itsi_id=itsi_user['id'])
     if user is not None and user.is_active:
--- a/src/mmw/js/src/core/error/controllers.js
+++ b/src/mmw/js/src/core/error/controllers.js
@@ -1,0 +1,19 @@
+"use strict";
+
+var ErrorHandlers = require('./handlers').ErrorHandlers;
+
+var ErrorController = {
+    error: function(type) {
+        switch(type) {
+            case 'itsi':
+                ErrorHandlers.itsi();
+                break;
+            default:
+                ErrorHandlers.generic(type);
+        }
+    }
+};
+
+module.exports = {
+    ErrorController: ErrorController
+};
--- a/src/mmw/js/src/core/error/handlers.js
+++ b/src/mmw/js/src/core/error/handlers.js
@@ -1,0 +1,25 @@
+"use strict";
+
+var router = require('../../router').router,
+    App = require('../../app');
+
+var ErrorHandlers = {
+    itsi: function() {
+        router.navigate('');
+        alert("We're sorry, but there was an error logging you in with your" +
+            " ITSI credentials. Please log in using another method or" +
+            " continue as a guest."
+        );
+        App.showLoginModal();
+    },
+
+    generic: function(type) {
+        router.navigate('');
+        alert("We're sorry, but an error occurred in the application.");
+        console.log("[MMW] An unknown error occurred: " + type);
+    }
+};
+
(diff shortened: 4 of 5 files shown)


src/mmw/js/src/controllers.js
```diff
@@ -26,6 +27,34 @@ var AppController = {
     }
 };
 
+var ErrorHandlers = {
+    itsi: function() {
+        router.navigate('');
+        alert("We're sorry, but there was an error logging you in with your" +
```
mmcfarland: As part of this PR, can you create an issue for @ctaylo37 to give us some markup for a general purpose error modal?

rajadain: Just added #224.

caseypt:
When you get in this morning, can you walk me through this PR? I'm a little out-of-date on the ITSI stuff.

src/mmw/js/src/controllers.js
```diff
@@ -26,6 +27,44 @@ var AppController = {
...
+        } else {
+            ErrorHandlers.generic();
+        }
+    }
+};
+
 module.exports = {
-    AppController: AppController
+    AppController: AppController,
+    ErrorController: ErrorController
```
caseypt: I was hoping to phase out this top-level controller file (we've been moving the page routes to their respective apps as they've been created). Maybe we could move this to `core/errorController.js`? Curious what others think.

caseypt: Just saw your comment about moving this to it's own folder. I would consider `errorViews`, `errorModels`, `errorController`, etc in `core` before moving it to it's own app. But again, open to input.

rajadain: I agree, cross-cutting concerns like error handling and logging should be in core, not in separate apps. I do like the simplicity of one-word non-CamelCase file names we have so far though, so perhaps we could do `core/error/views.js` and `/core/error/controllers.js` instead?

caseypt:
Here's what I'm seeing. Is this the correct behavior?

- Go to localhost:8000
- Navigate to user/itsi/login
- Login on ITSI using the `tstudent2` credentials.
- Get redirected back to localhost:8000
- An alert pops up with the following message: "We're sorry, but there was an error logging you in with your ITSI credentials. Please log in using another method or continue as a guest."

rajadain:
Well, you were redirected to `/error/itsi` and it showed you the popup message, so that part of the PR works, which is good. But no, since authentication was successful you should've been re-routed to `/register` instead.

Can you try running migrations to make sure that the `user_itsiuser` table is setup?

rajadain:
Also please make sure you have setup the `MMW_ITSI_SECRET_KEY` environment variable in the `app` VM, as described in the README. That could cause errors when negotiating an OAuth token.

src/mmw/js/src/controllers.js
caseypt: I'm cool with this. Any thoughts @mmcfarland?

mmcfarland: Yeah, agree that `core` is the right place, the sub-dir `error` seems like a good layout to me.

caseypt:
Thanks. The env var fixed it. I was successfully redirected to `/register`. 

Without the env var, I inadvertently tested that the error handling does work.

+1

rajadain:
As a side note, to test successful user authentication with ITSI account `tstudent2`, please do the following in the app shell `scripty manage shell`:

```python
from django.contrib.auth.models import User
from apps.user.models import ItsiUser
user = User.objects.create_user(username="tstudent2", email=None, password=None)
itsi_user = ItsiUser.objects.create_itsi_user(user, 214)
user.save()
itsi_user.save()
```

caseypt:
Cool. Ran that script, logged in, and was successfully redirected to the index page instead of register.

kdeloach:
+1

rajadain:
I've pushed some new commits to this PR, based on feedback from @caseypt, @kdeloach and @mmcfarland. Changes are as follows:

 * Moved `ErrorController` and `ErrorHandlers` out of `src/controllers.js` and into `src/core/error/controllers.js` and `src/core/error/handlers.js` respectively.
 * Combined `generic` and `unknown` error handlers into one
 * Removed awkward code for capturing URL before `Backbone.history.start` since that is handled natively by Backbone

Resulting code is much sleeker, thanks for all your input. Will merge in shortly.

PR #221, opened 2015-05-26
Add area of interest mask layer

kdeloach:
This adds a shadow to the map with a cutout of the selected area of
interest.

The conformMapToShape logic is no longer needed since we now return
properly serialized MultiPolygon geojson.

Connects #222

### DIFF
--- a/src/mmw/js/src/app.js
+++ b/src/mmw/js/src/app.js
@@ -62,12 +62,12 @@ var App = new Marionette.Application({
 
 function RestAPI() {
     return {
-        getPredefinedShapes: function() {
+        getPredefinedShapes: _.memoize(function() {
             return $.ajax({
                 'url': '/api/modeling/congressional_districts',
                 'type': 'GET'
             });
-        },
+        }),
 
         getPolygon: function(args) {
             return $.ajax({
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -151,7 +151,7 @@ var MapView = Marionette.ItemView.extend({
             try {
                 var layer = new L.GeoJSON(areaOfInterest);
                 this._areaOfInterestLayer.addLayer(layer);
-                this._leafletMap.fitBounds(this._areaOfInterestLayer.getBounds());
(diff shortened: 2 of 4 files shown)


src/mmw/js/src/app.js
```diff
@@ -62,12 +62,12 @@ var App = new Marionette.Application({
 
 function RestAPI() {
     return {
-        getPredefinedShapes: function() {
+        getPredefinedShapes: _.memoize(function() {
```
kdeloach: I didn't mean to commit this but it shouldn't hurt.

src/mmw/js/src/core/views.js
```diff
@@ -167,12 +167,65 @@ var MapView = Marionette.ItemView.extend({
...
+        if (areaOfInterest) {
+            var layer = new L.GeoJSON(areaOfInterest);
+            this._leafletMap.fitBounds(layer.getBounds(), { reset: true });
         }
     }
 });
 
+// Apply a mask over the entire map excluding bounds/shape specified.
+function applyMask(featureGroup, shapeLayer) {
+    var worldBounds = L.latLngBounds([-90, -360], [90, 360]),
```
jamesmcclain: To the best of my knowledge, latitude ranges from -180 to +180 degrees.  Is there some special reason for the double wrap?

jamesmcclain:
You create the shadow effect by making a transparent sheet with the shape cut out of it.  This is what I see when I zoom all the way out:
screenshot from 2015-05-26 17 25 14

jamesmcclain:
Other than the question about the double-wrap of the glob and the issue just above, it looks good to me.  I will reassign to you, for now.  Please feel free to assign back to me if/when you want to.

src/mmw/js/src/core/views.js
kdeloach: I believe this is so the shadow covers more of the planet when you are zoomed really far out.

mmcfarland:
I think we'll put some reasonable map bounds on at some point, since this is designed only for the conterminous US.  The wrap effect shouldn't be a problem.

mmcfarland:
+1 Tested.

Issue #223, opened 2015-05-26
Consider removing 'forgot username' option


lewfish:
The wireframe for the "Retrieve account info" modal shows checkboxes for "Forgot username" and "Forgot Password". However, we are using the Django authorization module which only provides a password reset mechanism (which sends the username and allows the user to reset their password). For now, these checkboxes don't do anything, but we should either remove them, or modify the functionality so that it's possible to only email the user their username. The latter option doesn't seem like it would be very useful.

![screen shot 2015-05-26 at 3 59 22 pm](https://cloud.githubusercontent.com/assets/1896461/7821949/3b88ceb8-03c0-11e5-80ca-465e751c65a8.png)

lewfish:
Wow, that's a big screenshot.

mmcfarland:
You Mac users and your retina screens!

caseypt:
I think we could remove both the checkboxes and just have a the email field.

jfrankl:
Yeah let's remove the checkboxes.

ajrobbins:
Addressed with redesign in #2254

PR #225, opened 2015-05-26
Add sign up and forgot password views

lewfish:
To test:

Try logging in with invalid username/password. The forgot link should pop up. Enter email address. Look at the console for the email. It isn't sent for real because we are using develop settings in Django. Follow the activation link. 

Try signing up for a new account. Follow the link in the console.

Try using input that has errors caught on the front-end (not filling required fields) and back-end errors (eg. passwords don't match, username is already in use, etc.).

I've noticed that the analyze view is unable to get the data to make the bar charts. I don't think this commit introduced the error. I might just need to run some secret command to get it to work (I tried destroying the VMs).

See https://github.com/WikiWatershed/model-my-watershed/issues/223

Connects #152

### DIFF
--- a/src/mmw/apps/user/views.py
+++ b/src/mmw/apps/user/views.py
@@ -1,4 +1,6 @@
-from django.contrib.auth import authenticate, logout, login as auth_login
+from django.contrib.auth import (authenticate,
+                                 logout as auth_logout,
+                                 login as auth_login)
 from django.contrib.auth.models import User
 from django.core.urlresolvers import reverse
 from django.shortcuts import redirect, render_to_response
@@ -50,8 +52,8 @@ def login(request):
 
 @decorators.api_view(['GET'])
 @decorators.permission_classes((AllowAny, ))
-def user_logout(request):
-    logout(request)
+def logout(request):
+    auth_logout(request)
 
     if request.is_ajax():
         response_data = {'guest': True, 'result': 'success'}
--- a/src/mmw/apps/user/views.py
+++ b/src/mmw/apps/user/views.py
@@ -1,10 +1,14 @@
-from rest_framework.response import Response
-from django.contrib.auth import authenticate, login, logout
+from django.contrib.auth import authenticate, logout, login as auth_login
 from django.contrib.auth.models import User
 from django.core.urlresolvers import reverse
 from django.shortcuts import redirect, render_to_response
+from django.contrib.auth.forms import PasswordResetForm
 
-from rest_framework import decorators
+from registration.forms import RegistrationFormUniqueEmail
+from registration.backends.default.views import RegistrationView
+
+from rest_framework import decorators, status
+from rest_framework.response import Response
 from rest_framework.permissions import AllowAny
 
 from apps.user.models import ItsiUser
@@ -13,22 +17,22 @@ from apps.user.itsi import ItsiService
 
 @decorators.api_view(['POST', 'GET'])
 @decorators.permission_classes((AllowAny, ))
-def ajax_login(request):
+def login(request):
     response_data = {}
-    status = 200
+    status_code = status.HTTP_200_OK
 
     if request.method == 'POST':
-            user = authenticate(username=request.REQUEST.get('username'),
-                                password=request.REQUEST.get('password'))
+        user = authenticate(username=request.REQUEST.get('username'),
+                            password=request.REQUEST.get('password'))
 
-            if user is not None and user.is_active:
-                login(request, user)
-                response_data['result'] = 'success'
-                response_data['username'] = user.username
-                response_data['guest'] = False
-            else:
-                response_data['result'] = 'error'
-                status = 400
+        if user is not None and user.is_active:
+            auth_login(request, user)
+            response_data['result'] = 'success'
+            response_data['username'] = user.username
+            response_data['guest'] = False
+        else:
+            response_data['result'] = 'error'
+            status_code = status.HTTP_400_BAD_REQUEST
     elif request.method == 'GET':
         user = request.user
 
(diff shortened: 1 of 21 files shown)


caseypt:
Re: the analyze page not getting the data, that is due to the URL not having a trailing slash. I fixed it in the issue I'm working on.

caseypt:
The form allows me to submit to the server without filling in the email field. I do get an "Email is invalid or already in use" message when the server responds, but we should prevent submitting the form if any of the fields are blank.

caseypt:
Submitting mismatched passwords generates an error server side, but no message is displayed to the user.

caseypt:
I was able to do the following:

- Successfully sign-up when all the fields are filled out correctly.
- Get an error when the username is already in use.

caseypt:
If you don't fill in any of the fields, you get an error message for each field. However, if you fill in the username field, you then only get messages for the password fields. You should still get an error for the email field.

caseypt:
Good start. I'll take another look once these issues are addressed.

lewfish:
Thanks for all the feedback. I believe I've addressed all the concerns.

lewfish:
I just pulled develop, rebased on top of develop, and there's still a commit from Matt in this PR. I'm not sure why.

kdeloach:
I think you can safely remove that commit (bbdccd0c23a26b8e561dab69f1ecabcb3b4a704c) with interactive rebase.

lewfish:
@caseypt Ok, this is good to go.

lewfish:
@azavea-bot rebuild

rajadain:
You may have to rebase over latest `develop` because I just merged #216 which touches some files that you also changed.

caseypt:
Yeah, I get a few conflicts when pulling this down. When you fix those up, I'll take a look.

lewfish:
Assuming it builds on Jenkins, this should be ready to test.

src/mmw/apps/user/views.py
```diff
@@ -114,6 +114,57 @@ def itsi_register(request):
...
+    form = RegistrationFormUniqueEmail(request.POST)
+    if form.is_valid():
+        view.register(request, **form.cleaned_data)
+    else:
+        if 'username' not in form.cleaned_data:
+            response_data['username_valid'] = False
+        if 'password1' not in form.cleaned_data or \
+           'password2' not in form.cleaned_data or \
+           form.cleaned_data['password1'] != form.cleaned_data['password2']:
+            response_data['password_valid'] = False
```
caseypt: Indentation is off here.

caseypt:
Looks good. Tested signing up and retrieving a password. Verified that all of the bugs I reported previously have been fixed.

A few things to consider:

- I think we should redirect to the home page when logging out.
- If you click "Sign Up", then dismiss the modal, then click "Login", the Sign Up modal opens (and preserves and displays error messages if there were any) instead of the Login modal. I think clicking "Login" should always open the Login modal. Likewise, if you successfully sign up or fill out the lost password form, when you click "Login", you get the last modal from each of those workflows.
- You might be able to squash a few of these commits together.

rajadain:
After I signed up, I got the "Confirm your email address" modal. After clicking that link in the email, I was taken to a new page that activated my new account. Back on the first page, the "Confirm your email address" modal stayed on. When I dismissed it and clicked "Login", the same "Confirm your email address" modal came up. But when I refreshed the page so that I could log in, I saw that I was already logged in as the new user.

To add to @caseypt's point above, I think that when clicking "Login" if the user has already been logged in then only that area should update to reflect that the user is logged in (just like clicking "Logout" does not refresh the page, simply updates that area). If the user is not already logged in, then they should see the Login modal.

lewfish:
I updated according to the latest feedback.

lewfish:
Sorry, I need to fix some unit tests first.

lewfish:
To test the latest additions:

Sign up a new user and dismiss the modal, but don't go to the activation page. Click Login. This should show the login modal and should not say 'Continue as Guest'. 

Now activate the new account in a new tab, but don't click a link to login or go back to MMW. Switch back to MMW and click Login. It should switch from Login to the name of the user without having to go through the login modal.

rajadain:
Good work. After dismissing the post-sign-up modal and clicking "Login" again, I was shown the Login modal and not the "Continue as Guest" modal. When I tried to login with the new credentials, I was shown the error "Invalid username or password", which I _guess_ is fine, although we could also have a message that says that the account hasn't been activated yet which is more meaningful for the user. How do others feel about this?

Also consider redirecting to the home page after account has been activated. Although I guess this could be potentially confusing if the user already has work in progress in the first tab.

mmcfarland:
I agree for a new message about needing to activate the account.

For redirecting, we're supporting a workflow for anonymous users who have work in progress and decide to register at that point.  We should leave the register window completion modal up with a button for login available (or flip directly to the login) and in the activation tab, there should be instructions similar to "if you had work in progress, visit the previous tab to continue and save your work".  Activating an account should not log you in.

mmcfarland:
re: my comments above, I don't think they need to be addressed in this PR.  We can incorporate them into a task about making the not-logged-in-save workflow functional.

rajadain:
@lewfish to not auto-login a user when they activate their account I think you can change this one line to `False`: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/base.py#L305

caseypt:
+1. Tested. Works as described. Agree that other changes can come in a different issue.

If there isn't an issue consider creating one for the workflow described above as well as one for the custom modal transitions.

lewfish:
Some new issues related to this PR:

#258 
#259 
#260

Issue #196, opened 2015-05-21
Update project models to sync to DRF endpoints.


mmcfarland:
This includes the frontend updates to allow proper saving and loading of projects and scenarios.

PR #230, opened 2015-05-27
Save and update projects

caseypt:
Connects to #196

This PR adds the ability to create new projects and update existing projects through the front-end. However, none of the controls are wired up, so testing will be done through the console.

To test:

- Run migrations.
- You'll need the [Marionette Inspector](https://github.com/marionettejs/marionette.inspector) to fully test this.
- Launch the app, draw an area or select one of the existing areas.
- Navigate to the modeling page.
- In the Marionette Inspector, load the project model in the console. This is done by clicking the download icon (selection_028) which appears when you hover a model in the data tab. In my case, the project was always `c34`.

selection_027

- You should now be able to make modifications to the project in the console through the `mn1` variable.
- Give the project a new name: `mn1.set('name', 'Test Project');`. The name of the project should be updated in the project drop down menu.
- Add a scenario by clicking the `+` button.
- Give the new scenario a new name: `mn1.get('scenarios').at(1).set('name', 'Green Roof');` The scenario name should be updated on the scenario tab.
- Save the project `mn1.saveAll();`. You should see save messages in the console and the URL should be updated with the project ID.
- Refresh the page. The project should be loaded with the modifications you made and the project AoI should be added to the map.
- Make some more changes (change the project name again, add a few more scenarios, etc). 
- Save the project and refresh. Again, the project should be loaded with your changes.

Some issues:

https://github.com/WikiWatershed/model-my-watershed/issues/233 What does a read-only/public project look like?

### DIFF
--- a/src/mmw/apps/home/views.py
+++ b/src/mmw/apps/home/views.py
@@ -4,9 +4,10 @@ from __future__ import unicode_literals
 from __future__ import division
 
 from django.contrib.auth.models import User
-from django.shortcuts import render_to_response
+from django.shortcuts import render_to_response, get_object_or_404
 from django.template.context_processors import csrf
 from rest_framework import serializers, viewsets
+from apps.modeling.models import Project
 
 
 # Serializers define the API representation.
@@ -25,6 +26,23 @@ class UserViewSet(viewsets.ModelViewSet):
 def home_page(request):
     csrf_token = {}
     csrf_token.update(csrf(request))
+
+    return render_to_response('home/home.html', csrf_token)
+
+
+def model(request, proj_id=None):
+    """
+    If proj_id was specified, check that the user owns
+    the project. It not, return a 404. Otherwise, just
+    load the index template and the let the front-end
+    handle the route and request the project through
+    the API.
+    """
+    csrf_token = {}
+    csrf_token.update(csrf(request))
+
+    if proj_id:
+        get_object_or_404(Project, user=request.user.id, id=proj_id)
 
     return render_to_response('home/home.html', csrf_token)
 
--- a/src/mmw/apps/modeling/models.py
+++ b/src/mmw/apps/modeling/models.py
@@ -79,7 +79,7 @@ class Scenario(models.Model):
     name = models.CharField(
         max_length=255)
     project = models.ForeignKey(Project, related_name='scenarios')
-    current_condition = models.BooleanField(
+    is_current_conditions = models.BooleanField(
         default=False,
         help_text='A special type of scenario without modification abilities')
     modifications = models.TextField(
--- a/src/mmw/apps/modeling/views.py
+++ b/src/mmw/apps/modeling/views.py
@@ -17,7 +17,9 @@ from apps.core.models import Job
 from apps.core.task_helpers import save_job_error, save_job_result
 from apps.modeling import tasks
 from apps.modeling.models import District, Project, Scenario
-from apps.modeling.serializers import ProjectSerializer, ScenarioSerializer
+from apps.modeling.serializers import (ProjectSerializer,
+                                       ProjectUpdateSerializer,
+                                       ScenarioSerializer)
 
 
 @decorators.api_view(['GET', 'POST'])
@@ -57,7 +59,8 @@ def project(request, proj_id):
 
     elif request.method == 'PUT':
         ctx = {'request': request}
-        serializer = ProjectSerializer(project, data=request.data, context=ctx)
+        serializer = ProjectUpdateSerializer(project, data=request.data,
+                                             context=ctx)
         if serializer.is_valid():
             serializer.save()
             return Response(serializer.data)
@@ -95,10 +98,7 @@ def scenarios(request):
 @decorators.permission_classes((IsAuthenticated, ))
 def scenario(request, scen_id):
     """Retrieve, update or delete a scenario"""
-    try:
-        scenario = Scenario.objects.filter(user=request.user, id=scen_id)
-    except Scenario.DoesNotExist:
-        return Response(status=status.HTTP_404_NOT_FOUND)
+    scenario = get_object_or_404(Scenario, id=scen_id)
 
     if request.method == 'GET':
         serializer = ScenarioSerializer(scenario)
--- a/src/mmw/js/src/modeling/controllers.js
+++ b/src/mmw/js/src/modeling/controllers.js
@@ -9,66 +9,63 @@ var $ = require('jquery'),
     models = require('./models');
 
 var ModelingController = {
-    modelPrepare: function() {
-        if (!App.map.get('areaOfInterest')) {
+    modelPrepare: function(projectId) {
+        if (!projectId && !App.map.get('areaOfInterest')) {
             router.navigate('', { trigger: true });
             return false;
         }
     },
 
-    model: function() {
-        // TODO: If you are coming to the page with a
-        // project ID in the URL, we need to load an
-        // existing  project here.
-        // i.e. var project = new models.ProjectModel({
-        //  id: XX
-        //  }).fetch
-        //  .done(init everything else);
+    model: function(projectId) {
+        var project;
 
-        // TODO: For a new project, users will eventually
-        // be able to choose which modeling package
-        // they want to use in their project. For
-        // now, the only option is TR55, so it is
-        // hard-coded here.
+        if (projectId) {
+            project = new models.ProjectModel({
+                id: projectId
+            });
 
-        // Init models
-        var taskModel = new models.Tr55TaskModel(),
-            currentConditions = new models.ScenarioModel({
-                name: 'Current',
-                currentConditions: true
-            }),
-            scenario1 = new models.ScenarioModel({
-                name: 'Scenario 1'
-            }),
-            scenario2 = new models.ScenarioModel({
-                name: 'Flood Scenario'
-            }),
+            project
+                .fetch()
+                .done(function(data) {
+                    var scenarios = new models.ScenariosCollection();
+
+                    _.each(data.scenarios, function(scenario) {
+                        scenarios.add(new models.ScenarioModel(scenario));
+                    });
+
+                    project.set('scenarios', scenarios);
+                    project.set('active_scenario_slug', 'current-conditions');
+
+                    initViews(project);
+                });
+        } else {
+            var taskModel = new models.Tr55TaskModel(),
+                currentConditions = new models.ScenarioModel({
+                    name: 'Current Conditions',
+                    is_current_conditions: true
+                });
+
             project = new models.ProjectModel({
                 name: 'My Project',
-                createdAt: Date.now(),
-                areaOfInterest: App.map.get('areaOfInterest'),
-                activeScenarioSlug: 'scenario-1',
-                modelPackage: new models.ModelPackageModel({
+                created_at: Date.now(),
+                area_of_interest: App.map.get('areaOfInterest'),
+                active_scenario_slug: 'current-conditions',
+                model_package: new models.ModelPackageModel({
+                    // TODO: For a new project, users will eventually
+                    // be able to choose which modeling package
+                    // they want to use in their project. For
+                    // now, the only option is TR55, so it is
+                    // hard-coded here.
                     name: 'TR-55',
                     taskModel: taskModel
                 }),
                 scenarios: new models.ScenariosCollection([
-                    currentConditions,
-                    scenario1,
-                    scenario2
+                    currentConditions
                 ])
             });
 
-        // Init views
-        var modelingResultsWindow = new views.ModelingResultsWindow({
-                model: project
-            }),
-            modelingHeader = new views.ModelingHeaderView({
-                model: project
-            });
-
-        App.rootView.subHeaderRegion.show(modelingHeader);
-        App.rootView.footerRegion.show(modelingResultsWindow);
+            initViews(project);
+        }
     },
 
     modelCleanUp: function() {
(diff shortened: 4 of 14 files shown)


caseypt:
Connects to #196

caseypt:
Address initial comments from @mmcfarland re: doc string for `model` view and fixing serialize import statements.

Unaddressed comment here on outdated commit: https://github.com/caseypt/model-my-watershed/commit/b70d977b36a5457d6a8fe019c02fac9ef70f5a4e#commitcomment-11395150

kdeloach:
I get an error after selecting an area from the districts dropdown and clicking "Model".
image

src/mmw/js/src/modeling/models.js
```diff
@@ -15,9 +15,75 @@ var ResultCollection = Backbone.Collection.extend({
...
+        // save each one once we have an ID for the project
+        if (this.isNew) {
+            var tempScenarios = this.get('scenarios');
+        }
+
+        this.save()
+            .done(function(resp) {
+                console.log('Saved project');
+
+                var scenarios = tempScenarios ? tempScenarios : this.get('scenarios');
```
kdeloach: Should `this.get('scenarios')` be `self.get('scenarios');`?

src/mmw/apps/home/views.py
```diff
@@ -29,5 +30,22 @@ def home_page(request):
...
+def model(request, proj_id=None):
+    """
+    If proj_id was specified, check that the user owns
+    the project. It not, return a 404. Otherwise, just
+    load the index template and the let the front-end
+    handle the route and request the project through
+    the API.
+    """
+    csrf_token = {}
+    csrf_token.update(csrf(request))
```
kdeloach: Do we really need to pass the csrf token to the client like this? We attach the csrf token as a header to each ajax call now.

src/mmw/apps/modeling/models.py
```diff
@@ -79,7 +79,7 @@ class Meta:
     name = models.CharField(
         max_length=255)
     project = models.ForeignKey(Project, related_name='scenarios')
-    current_condition = models.BooleanField(
+    is_current_conditions = models.BooleanField(
```
kdeloach: Can we replace this boolean with something more descriptive? For example, the name of the model simulation to run for this scenario? If it's null we can infer that the scenario is the "current condition".

src/mmw/apps/modeling/views.py
```diff
@@ -96,7 +99,7 @@ def scenarios(request):
 def scenario(request, scen_id):
     """Retrieve, update or delete a scenario"""
     try:
-        scenario = Scenario.objects.filter(user=request.user, id=scen_id)
+        scenario = Scenario.objects.get(id=scen_id)
```
kdeloach: Consider replacing try/catch with `get_object_or_404`.

caseypt:
@kdeloach sorry. fixed.

src/mmw/apps/modeling/models.py
caseypt: I'm not sure I understand. The "current conditions" scenario is a special scenario that can't be modified. The model simulation to run depends on the project it is a part of. All scenarios in a project get run through the same model.

src/mmw/apps/home/views.py
caseypt: To be honest, I don't know. @lliss and I were having trouble with this before when implementing the login stuff. I can try removing it from this view and see if I have any trouble.

It seems like we should never have to return it since we have the CSRFMiddleware app turned on.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -9,66 +9,54 @@ var $ = require('jquery'),
...
-        // hard-coded here.
+        if (projectId) {
+            project = new models.ProjectModel({
+                id: projectId
+            });
+
+            project
+                .fetch()
+                .done(function(data) {
+                    App.map.set('areaOfInterest', project.get('area_of_interest'));
```
kdeloach: I wonder if we should consider using https://github.com/vbabiy/djangorestframework-camel-case?

src/mmw/js/src/modeling/models.js
caseypt: Yeah, good find.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -9,66 +9,54 @@ var $ = require('jquery'),
...
                 name: 'My Project',
-                createdAt: Date.now(),
-                areaOfInterest: App.map.get('areaOfInterest'),
-                activeScenarioSlug: 'scenario-1',
-                modelPackage: new models.ModelPackageModel({
-                    name: 'TR-55',
-                    taskModel: taskModel
-                }),
+                created_at: Date.now(),
+                area_of_interest: aoi,
```
kdeloach: It would be great if we could avoid mixing styles. I know that's easier said than done...

src/mmw/js/src/modeling/controllers.js
caseypt: I'm not opposed to this.

mmcfarland: I found that too, but the project doesn't look super active.  Worried about updates to DRF breaking it.

src/mmw/js/src/modeling/models.js
```diff
@@ -15,9 +15,75 @@ var ResultCollection = Backbone.Collection.extend({
...
+                            console.log('Failed to save scenario');
+                        });
+                });
+            })
+            .fail(function(resp) {
+                console.log('Failed to save project');
+            });
+    },
+
+    parse: function(response, options) {
```
kdeloach: Can we use this parse function to convert underscore data to camel case?

src/mmw/js/src/modeling/models.js
```diff
@@ -15,9 +15,75 @@ var ResultCollection = Backbone.Collection.extend({
...
+
+    saveAll: function() {
+        var self = this;
+
+        // If we haven't saved the project before,
+        // hold on to the scenarios so that we can
+        // save each one once we have an ID for the project
+        if (this.isNew) {
+            var tempScenarios = this.get('scenarios');
+        }
```
kdeloach: I don't fully understand what `tempScenarios` is for. At a glance, it doesn't seem necessary.

src/mmw/js/src/modeling/models.js
caseypt: I'd be more open to using the DRF add-on then doing that. Seems more fragile and that we would have to implement for every model that needs to be synced.

src/mmw/js/src/modeling/controllers.js
kdeloach: I'm open to just using underscores for all Backbone model attributes.

src/mmw/js/src/modeling/models.js
caseypt: Before we can save the scenarios, we have to save the project so that we know the ID of the project. If we save the project without temporarily storing the scenarios, we'll lose them when we refresh our project model from the server response because the scenarios are independent models on the server and not yet saved. Once we have the project ID, we can iterate through the scenarios and save them.

src/mmw/js/src/modeling/controllers.js
lliss: That certainly seems easier than anything else.

caseypt: I think I like that solution best.

src/mmw/apps/home/views.py
kdeloach: Maybe we can leave this for now and investigate later.

src/mmw/js/src/modeling/models.js
kdeloach: Can we save project & scenarios at the same time in a single request?

src/mmw/apps/modeling/models.py
kdeloach: I thought it would make the app a little more flexible in the future. We could easily add the ability to run a different models per scenario in the same project if we just specify the model simulation name here. For "current condition" we could just have a null or noop model simulation?

src/mmw/js/src/modeling/models.js
caseypt: It's not implemented by default in DRF (saving nested models through a serializer http://www.django-rest-framework.org/api-guide/serializers/#dealing-with-nested-objects) so we had to implement it there or on the front-end. I chose here, but I can implement it on the back-end if you think that makes more sense.

src/mmw/apps/modeling/views.py
caseypt: Will do.

src/mmw/apps/modeling/models.py
mmcfarland: The format and type of the modifications and inputs for the different models are so different that running them on existing scenarios would be prohibitive.  For example, modifications to tr55 can be a geometry, where in gwlf-e, they'll likely be declarative "add 5m riparian buffer around all stream networks".  That's why we chose to associate the model package with the project instead.

src/mmw/js/src/modeling/models.js
kdeloach: I would favor doing it server-side. Seems like creating/updating nested structures is something we'll need to do a lot and it would be good to establish some conventions around that.

mmcfarland: Having project/scenario endpoints is useful for more atomic updates when editing (changing a modification on a single scenario), which is more common than an initial project save of existing scenarios.  Rather than having multiple types of endpoints to maintain, I kind of like having just the flat `/projects/` and `/scenarios/` endpoints to handle both use cases.

kdeloach: Good point. Do we save projects/scenarios individually? Or do we always save them in bulk? It would be great if DRF could take care of both cases for us.

src/mmw/apps/modeling/models.py
kdeloach: Conceptually, I don't see "current conditions" as different from other model simulations. Maybe this will become more clear as we add alternative model simulations. Thanks for explaining.

kdeloach:
+1 The address bar URL updated after saving a project and renaming scenarios updated the UI correctly.  These changes persisted when reloading the page.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -9,66 +9,58 @@ var $ = require('jquery'),
...
+            // on whether the shape was drawn or selected.
+            var aoi = App.map.get('areaOfInterest').geometry ?
+                      App.map.get('areaOfInterest').geometry :
+                      App.map.get('areaOfInterest');
+
+            // The AoI must always be a MultiPolygon
+            if (aoi.type === 'Polygon') {
+                aoi.type = 'MultiPolygon';
+                aoi.coordinates = [aoi.coordinates];
+            }
```
kdeloach: This throws an error for me when I try to load this page with a district selected. Consider fixing this upstream instead of mutating AOI client-side.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -9,66 +9,58 @@ var $ = require('jquery'),
...
+                .done(function(data) {
+                    App.map.set('areaOfInterest', project.get('area_of_interest'));
+                    initViews(project);
+                });
+        } else {
+            // The structure of the AoI is slightly different depending
+            // on whether the shape was drawn or selected.
+            var aoi = App.map.get('areaOfInterest').geometry ?
+                      App.map.get('areaOfInterest').geometry :
+                      App.map.get('areaOfInterest');
```
kdeloach: I don't know if this has any impact on this, but I recently modified the GeoJSON that gets populated for selected shapes (https://github.com/WikiWatershed/model-my-watershed/commit/ce884f6e5b3fe9dc1453c1d2fc8f535449aa3223#diff-059395c30679e53cc84d53e77c16f93dL107).

src/mmw/js/src/modeling/controllers.js
caseypt: https://github.com/WikiWatershed/model-my-watershed/issues/237

src/mmw/js/src/modeling/models.js
caseypt: https://github.com/WikiWatershed/model-my-watershed/issues/238

caseypt: Scenarios are saved both in bulk (through their individual endpoints) and individually

src/mmw/js/src/modeling/controllers.js
caseypt: Okay. Thought I fixed it for both drawn and selected shapes, but I'll take another look and try to modify it further upstream.

src/mmw/js/src/modeling/models.js
caseypt: Fixed.

src/mmw/apps/modeling/views.py
caseypt: Fixed.

src/mmw/js/src/modeling/controllers.js
caseypt: Fixed and moved upstream.

src/mmw/apps/home/views.py
caseypt: https://github.com/WikiWatershed/model-my-watershed/issues/239

caseypt:
Updated with some small fixes. Issues have been created for bigger suggestions/changes.

src/mmw/js/src/modeling/models.js
```diff
@@ -15,9 +15,75 @@ var ResultCollection = Backbone.Collection.extend({
...
+    },
+
+    parse: function(response, options) {
+        var scenariosCollection = new ScenariosCollection();
+
+        _.each(response.scenarios, function(scenario) {
+            scenariosCollection.add(new ScenarioModel(scenario));
+        });
+
+        response.scenarios = scenariosCollection;
```
kdeloach: Consider using `reset` here (http://backbonejs.org/#Collection-reset). You may have to specify a `model` attribute on `ScenariosCollection` if there isn't one already.

kdeloach:
Looks good.

src/mmw/js/src/modeling/models.js
caseypt: Thanks. Good call. Updated

PR #231, opened 2015-05-27
Feature/lol/scenario dropdown

lliss:
This is currently a preview. I'll connect it properly and submit for for formal review when #196 us done.

### DIFF
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -21,6 +21,9 @@ var _ = require('lodash'),
     chart = require('../core/chart.js'),
     barChartTmpl = require('../core/templates/barChart.ejs');
 
+var ENTER_KEYCODE = 13,
+    ESCAPE_KEYCODE = 27;
+
 // The entire modeling header.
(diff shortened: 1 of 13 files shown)


src/mmw/js/src/modeling/views.js
```diff
@@ -98,15 +100,58 @@ var ScenariosView = Marionette.LayoutView.extend({
...
+        } else {
+            return 'New Scenario ' + this.scenarioCounter();
+        }
+    },
+
     setActiveScenario: function(e) {
-        var scenarioSlug = $(e.target).data('scenario-slug');
+        var scenarioSlug = $(e.currentTarget).data('scenario-slug');
+
+        _.each(this.collection.models, function(model) {
```
caseypt: There is an `activeScenarioSlug` attribute on `project` (now `active-scenario-slug` in #230). I did it that way so we didn't have to worry about keeping that state on all the scenarios and we only have to update it in one place. If possible, consider using that attribute instead of `scenario.active`.

lliss:
As a note, the functionality was based on the way that google spreadsheets work. I tried to mimic it as closely as possible.

Issue #166, opened 2015-05-15
Implement model modification tools


caseypt:
Depends on #165

These are the tools used to modify the Area of Interest and to generate model input data.

The results of the modifications (i.e. drawing a shape) need to be stored on a scenario in some form (likely a GeoJSON blob of all modifications).

It's probably best to blow away the current set of modifications for a scenario and reset with the latest each time new modifications are made or existing modifications are edited. 

We'll need to preserve the GeoJSON so we can redraw the modifications when a project is open, and we'll also need to provide the model with the results of the modifications.

Ex.

![selection_017](https://cloud.githubusercontent.com/assets/1042475/7655983/8ec7ceae-faf5-11e4-83fe-8c24d0123978.png)

Issue #226, opened 2015-05-27
Save on Edit for projects and scenarios


mmcfarland:
Initial auto-save functionality should be to save project/scenarios on attribute changes.  Keep the update as narrow as possible, ie, when a scenario name is changed, don't update all project scenarios or the project.  This workflow applies only to logged in users.

caseypt:
- Remove controls for public projects that don't belong to the user.
- Don't try to fire save requests when non-owner is viewing project.

lliss:
- if logged in - wait until a project attribute is changed or a scenario is added
 - if logged out, the login should trigger a save, probably immediately

PR #235, opened 2015-05-28
Add scenario modification drawing tools

kdeloach:
This adds the ability to add and remove features to the map.

Connects #166

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -18,6 +18,15 @@ var ResultCollection = Backbone.Collection.extend({
 var ProjectModel = Backbone.Model.extend({
     urlRoot: '/api/modeling/projects/',
 
+    defaults: {
+        name: '',
+        created_at: null,          // Date
+        area_of_interest: null,    // GeoJSON
+        model_package: null,       // ModelPackageModel
+        taskModel: null,           // TaskModel
+        scenarios: null            // ScenariosCollection
+    },
+
     initialize: function() {
         // TODO: For a new project, users will eventually
         // be able to choose which modeling package
@@ -88,13 +97,33 @@ var ProjectModel = Backbone.Model.extend({
     }
 });
 
+var ModificationModel = Backbone.Model.extend({
+    defaults: {
+        name: '',
+        type: '',
+        geojson: null
+    }
+});
+
+var ModificationsCollection = Backbone.Collection.extend({
+    model: ModificationModel
+});
+
 var ScenarioModel = Backbone.Model.extend({
     urlRoot: '/api/modeling/scenarios/',
 
     defaults: {
         name: '',
         is_current_conditions: false,
+        modifications: null,      // ModificationsCollection
+        controlsEnabled: true,
         active: false
+    },
+
+    initialize: function() {
+        // TODO: When fetching model or populating from a preloaded source
+        // this attribute should not be initialized as empty.
+        this.set('modifications', new ModificationsCollection());
     },
 
     getSlug: function() {
@@ -103,6 +132,14 @@ var ScenarioModel = Backbone.Model.extend({
                        .replace(/ /g, '-') // Spaces to hyphens
                        .replace(/[^\w-]/g, ''); // Remove non-alphanumeric characters
         return slug;
+    },
+
+    enableControls: function() {
+        this.set('controlsEnabled', true);
+    },
+
+    disableControls: function() {
+        this.set('controlsEnabled', false);
     }
 });
 
@@ -123,6 +160,8 @@ module.exports = {
     ModelPackageModel: ModelPackageModel,
     Tr55TaskModel: Tr55TaskModel,
     ProjectModel: ProjectModel,
+    ModificationModel: ModificationModel,
+    ModificationsCollection: ModificationsCollection,
     ScenarioModel: ScenarioModel,
     ScenariosCollection: ScenariosCollection
 };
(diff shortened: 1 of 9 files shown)


kdeloach:
I'm in the process of reverting the changes I made regarding "current conditions".

kdeloach:
Rebased and ready for testing.

kdeloach:
Rebased onto develop. There's one bug that I'm fixing now. Since each scenario tab has the same slug, switching tabs won't quite work as expected.

caseypt:
Checking it out now.

caseypt:
Not sure if this needs to happen as part of this PR, but we can persist and reload the modifications now.

caseypt:
Another thing that may not be a part of this PR: Clicking on a modification to get info, delete it, etc.

mmcfarland:
Already got a separate issue for persisting, unless tossing it in helps to test that things are working.  Though it might be good to focus on whatever nuances come up.  https://github.com/WikiWatershed/model-my-watershed/issues/226

caseypt:
I created three scenarios in addition to current conditions, and renamed them so that the tabs work. 

When I draw on the 3rd scenario, the new shape, plus any previous shapes show up on the map. When I switch back to the first scenario, all of the shapes are still on the map. Then, when I switch back to the 3rd scenario, none of the shapes are there.

Let me know if that makes sense.

caseypt:
@kdeloach I can still reproduce the above if I update the scenario name through the console, and ensure that the slug has been updated too. However, your method of testing was better, so I may be seeing an unrelated issue.

caseypt:
If you think of any tests to add, add them to the list: https://github.com/WikiWatershed/model-my-watershed/issues/242

kdeloach:
I'll try to reproduce this issue with the Marionette inspector.

src/mmw/js/src/modeling/views.js
```diff
@@ -173,16 +179,69 @@ var ScenarioDropDownMenuView = Marionette.CompositeView.extend({
...
+
+    startDrawing: function(e) {
+        var self = this,
+            $el = $(e.currentTarget),
+            modificationsColl = this.model.get('modifications'),
+            featureName = $el.data('feature-name'),
+            featureType = $el.data('feature-type'),
+            map = App.getLeafletMap();
+        this.model.disableControls();
+        drawUtils.drawPolygon(map).then(function(geojson) {
```
mmcfarland: Still really like this promise syntax for drawing, nice.

src/mmw/js/src/modeling/views.js
```diff
@@ -199,6 +258,8 @@ var ToolbarTabContentsView = Marionette.CollectionView.extend({
     childView: ToolbarTabContentView,
 
     onRender: function() {
+        // TODO: Move this logic to the template (active tab should
+        // be stored in model state somewhere).
```
caseypt: I think I made this comment already, but I can't find it. So here it is again: There's a attribute on project called `active-scenario-slug` that is used to set the active tab. I did it that way since it required less state. If it's more of a pain to have it there, we should add it to scenario, then remove `active-scenario-slug`.

kdeloach: Sorry, it got rebased away. Definitely sounds like "initial scenario" and "active scenario slug" could be combined. I'll think it through a little more.

caseypt: Yeah `initialScenario` is `activeScenarioSlug`, but I thought the name `initialScenario` made more sense as it got passed down and the view was rendered initially. Open to whatever changes though.

kdeloach:
Fixed some rebase errors, updated model attributes to use snake case in some places, and removed  obsolete code related to enabling/disabling controls while drawing.

kdeloach:
@caseypt Updating these models via Marionette Inspector won't work because we don't presently re-render the tab panels when the scenario model is updated. I'll try to fix this as part of this PR.

See how the tab panels all have the same ID even after I altered the name/slug of each scenario.

image

kdeloach:
Fixed all the tab problems (that I'm aware of). Everything should be working now.

caseypt:
+1. Previously reported issue re: tab switching and persisting shapes has been fixed. I can give it another run through when you rebase it on develop.

kdeloach:
Rebased.

caseypt:
Tested again. Looks good.

src/mmw/js/src/modeling/models.js
```diff
@@ -18,6 +18,16 @@ var ResultCollection = Backbone.Collection.extend({
...
     urlRoot: '/api/modeling/projects/',
 
+    defaults: {
+        name: '',
+        created_at: null,          // Date
+        area_of_interest: null,    // GeoJSON
+        model_package: null,       // ModelPackageModel
+        taskModel: null,           // TaskModel
+        scenarios: null,           // ScenariosCollection
+        active_scenario_slug: ''
```
caseypt: I think we can kill this now that we have `active` on scenario?

kdeloach: Removed.

src/mmw/js/src/modeling/views.js
```diff
@@ -228,6 +272,10 @@ var ModelingResultsWindow = Marionette.LayoutView.extend({
     tagName: 'div',
     template: resultsWindowTmpl,
 
+    modelEvents: {
```
caseypt: Sorry, here too.

kdeloach: Thanks, will remove in a follow up PR.

PR #241, opened 2015-05-28
Add debugtiler.sh

kdeloach:
May be usuful for troubleshooting tiler issues. Copied over from NYC
TreesCount!

### DIFF
--- a/src/tiler/package.json
+++ b/src/tiler/package.json
@@ -6,8 +6,12 @@
   "dependencies": {
     "windshaft": "0.36.0"
   },
+  "devDependencies": {
+    "supervisor": "^0.6.0"
+  },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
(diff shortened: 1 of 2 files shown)


mmcfarland:
+1 Tested.  Great output, too.

src/tiler/package.json
```diff
@@ -6,8 +6,12 @@
   "dependencies": {
     "windshaft": "0.36.0"
   },
+  "devDependencies": {
+    "supervisor": "^0.6.0"
```
hectcastro: According to [this project's `README`](https://github.com/isaacs/node-supervisor/blob/master/README.md), it is no longer being maintained. Do we have any better alternatives?

kdeloach: Open to suggestions, but this isn't mission critical (more of a convenience).

hectcastro: A thing I've wanted to evaluate to try and solve this problem in a more general way, but haven't had time yet is [Watchman](https://facebook.github.io/watchman/).

kdeloach: Added an issue https://github.com/WikiWatershed/model-my-watershed/issues/270

Issue #243, opened 2015-05-29
One failing unit test


jamesmcclain:
There is currently a failing unit test:

```
ok 21 Firefox 38.0 - Draw ToolbarView enables/disables toolbar controls when the model enableTools/disableTools methods are called
not ok 22 Firefox 38.0 - Draw ToolbarView adds a layer to the map when an area of interest is chosen from the select area dropdown
    ---
        message: >
            expected { Object (type, coordinates) } to equal { Object (type, geometry) }
    ...
ok 23 Firefox 38.0 - Geocoder SearchBoxView renders no list items if there no suggestions
```

and

```
ok 60 PhantomJS 1.9 - Draw ToolbarView enables/disables toolbar controls when the model enableTools/disableTools methods are called
not ok 61 PhantomJS 1.9 - Draw ToolbarView adds a layer to the map when an area of interest is chosen from the select area dropdown
    ---
        message: >
            expected { Object (type, coordinates) } to equal { Object (type, geometry) }
        stack: >
            AssertionError: expected { Object (type, coordinates) } to equal { Object (type, geometry) }
                at http://localhost:7357/5046/js/test.vendor.js:5414
                at http://localhost:7357/5046/js/test.vendor.js:6973
                at http://localhost:7357/5046/js/test.bundle.js:2284
                at http://localhost:7357/testem/mocha.js:4212
                at http://localhost:7357/testem/mocha.js:4584
                at http://localhost:7357/testem/mocha.js:4643
                at next (http://localhost:7357/testem/mocha.js:4510)
                at http://localhost:7357/testem/mocha.js:4519
                at next (http://localhost:7357/testem/mocha.js:4463)
                at http://localhost:7357/testem/mocha.js:4487
                at timeslice (http://localhost:7357/testem/mocha.js:5480)
    ...
ok 62 PhantomJS 1.9 - Geocoder SearchBoxView renders no list items if there no suggestions
```

It was not present in commit `a2d911244` but is present thereafter.  It is possible that this was not detected before because Jenkins is not rebuilding the tests (possibly due to changes to the `bundle.sh` script).

kdeloach:
Good catch.

caseypt:
Fixed in #244

PR #244, opened 2015-05-29
Build test bundle when provisioning production environment

kdeloach:
This should fix tests not running on Jenkins but does not fix any presently failing tests.

Connects #243


jamesmcclain:
+1

I deleted `/var/www/mmw/static/js/test.(bundle|vendor).js` on the `app` VM, re-provisioned, and they  had reappeared.

Perhaps we should wait until there is a fix for the unit test before closing issue #243.

kdeloach:
Agreed.

jamesmcclain:
I saw that you pushed a fix for the test, as well.  I tested it and it worked, +1

Issue #168, opened 2015-05-15
Implement scenario drop down functionality


caseypt:
![selection_019](https://cloud.githubusercontent.com/assets/1042475/7656721/60b6ba52-fafa-11e4-8d86-9fe8f2f6b1ef.png)

PR #245, opened 2015-05-29
Feature/lol/scenario dropdown

lliss:
Connects #168 

This PR adds real functionality to the dropdowns and creates abstracted modals that can be used anywhere.

To test:

1. Draw an area and enter modeling mode.
2. Click the "+" twice to create two new scenarios.
3. Use the dropdown to rename a scenario.
4. Use the dropdown to delete a scenario, you should be confronted with a modal confirmation box.
5. Try sharing and printing (similar functionality).
6. Attempt to create two scenarios with the same name. Try to trick the UI to create scenarios with the same slug by making names like "-A-" and "+A+".


lliss:
@azaveaci rebuild

lliss:
@azaveaci rebuild

src/mmw/js/src/core/views.js
```diff
@@ -238,9 +240,72 @@ function getLatLngs(boundsOrShape) {
...
 
+var DeleteModal = Marionette.ItemView.extend({
+    el: '#modal-basic-target',
+
+    viewOptions: ['objToDelete'],
+
+    initialize: function(options) {
+        var self = this;
+        this.mergeOptions(options, this.viewOptions);
+        this.$el.on('hide.bs.modal', function(e) {
```
mmcfarland: Unused var `e`

src/mmw/js/src/core/views.js
```diff
@@ -238,9 +240,72 @@ function getLatLngs(boundsOrShape) {
...
+    ui: {
+        deleteConfirmed: '.delete',
+    },
+
+    events: {
+        'click @ui.deleteConfirmed': 'deleteModel',
+    },
+
+    deleteModel: function() {
+        this.objToDelete.destroy();
```
mmcfarland: Seems like we may want the delete modal to just confirm the delete and let the calling controller to the actual deleting.  At the very least, I would want to the jqXHR from this `destroy` returned so the caller could handle success and failure situations.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -38,7 +38,8 @@ var ModelingController = {
                 scenarios: new models.ScenariosCollection([
                     new models.ScenarioModel({
                         name: 'Current Conditions',
-                        is_current_conditions: true
+                        is_current_conditions: true,
+                        active: true
```
mmcfarland: I think there's already some discussion about the active scenario tab going on in some work by @caseypt and @kdeloach

mmcfarland:
`Current Conditions` scenario should not be modifiable, right now I can still change the name of the tab.

mmcfarland:
Selecting `rename` on a scenario and clicking on the header gives an error the the name is already in use, when it's not.

screenshot from 2015-05-29 16 19 47

mmcfarland:
Selecting a dropdown on a scenario tab should also active that tab as if you're clicking on it.

screenshot from 2015-05-29 16 20 53

src/mmw/js/src/modeling/views.js
```diff
@@ -102,15 +106,58 @@ var ScenariosView = Marionette.LayoutView.extend({
...
+        } else {
+            return 'New Scenario ' + this.scenarioCounter();
+        }
+    },
+
     setActiveScenario: function(e) {
-        var scenarioSlug = $(e.target).data('scenario-slug');
+        var scenarioSlug = $(e.currentTarget).data('scenario-slug');
 
+        _.each(this.collection.models, function(model) {
```
mmcfarland: Backbone Collections already have all the underscore collection methods on them, you can just use:

`this.collection.each(function(model) ....)`

or in this case, more succinctly, `this.collection.invoke('set', 'active', false);`

src/mmw/js/src/modeling/views.js
```diff
@@ -102,15 +106,58 @@ var ScenariosView = Marionette.LayoutView.extend({
...
+            model.set('active', false);
+        }, this);
+
+        _.find(this.collection.models, function(model) {
+            return model.get('slug') === scenarioSlug;
+        }, this).set('active', true);
+
+        // we are trying to active a different tab so need to switch tabs here.
+        this.panelsRegion.$el.find('a[data-scenario-slug=' + scenarioSlug + ']').tab('show');
```
mmcfarland: I'd try to sync up with the existing work of keeping track of the active scenario here.  Also not sure why you need to show a tab explicitly in this function now, was it not previously working?

src/mmw/js/src/modeling/views.js
```diff
@@ -130,6 +177,64 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
...
+        'click @ui.rename': 'renameScenario',
+        'click @ui.destroyConfirm': 'destroyConfirm',
+        'click @ui.print': function() {
+            window.print();
+        },
+        'click @ui.share': 'showShareModal'
+    },
+    renameScenario: function() {
+        var self = this,
+            $nameField = this.$el.find('.tab-name');
```
mmcfarland: This could be added to `ui` as well.

src/mmw/js/src/modeling/views.js
```diff
@@ -130,6 +177,64 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
...
+            objToDelete: this.model,
+            model: new Backbone.Model({ deleteLabel: 'scenario' })
+        });
+        del.render();
+        del.$el.modal('show');
+    },
+
+    showShareModal: function() {
+        var share = new coreViews.ShareModal({
+            model: new Backbone.Model({ url: window.location.href })
```
mmcfarland: If we're generating a link to the active scenario, that slug should be in the URL.  However, if this model isn't persisted yet (we're not logged in) we don't have a URL to generate so we should prevent this behavior.

src/mmw/js/src/modeling/views.js
```diff
@@ -142,6 +247,20 @@ var ScenarioTabPanelsView = Marionette.CollectionView.extend({
     },
 
     childView: ScenarioTabPanelView,
+    childEvents: {
+        'rename:scenario': function(view, model, newName) {
+            var match = _.find(this.collection.models, function(model) {
```
mmcfarland: I think all of this logic could be on a `setName` method on the model, in case we need to rename scenarios from other places.  Should return success so the view can handle the necessary UI

src/mmw/js/src/modeling/views.js
```diff
@@ -130,6 +177,64 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
...
+                self.render();
+            }
+        });
+
+        $nameField.on('keypress', function(e) {
+            var keycode = (e.keyCode ? e.keyCode : e.which);
+            if (keycode == ENTER_KEYCODE) {
+                // Don't add line returns to the text.
+                e.preventDefault();
+                self.triggerMethod('rename:scenario', self.model, $(this).text());
```
mmcfarland: Is there a way to directly call these methods instead of trigger events?

src/mmw/js/src/modeling/templates/scenarioTabPanel.ejs
```diff
@@ -1,15 +1,15 @@
-<a href="#<%= slug %>" data-scenario-slug="<%= slug %>" aria-controls="home" role="tab" data-toggle="tab"><%= name %></a>
+<a class="tab-name" data-scenario-slug="<%= slug %>" aria-controls="home" role="tab" data-toggle="tab"><span><%= name %></span></a>
```
kdeloach: Is there any opposition to keeping the href so the boostrap tabs "just work"?

lliss: No. I'll add it back.

src/mmw/js/src/modeling/views.js
```diff
@@ -102,15 +106,58 @@ var ScenariosView = Marionette.LayoutView.extend({
...
+            if (slug.match(regEx) !== null && slug.match(regEx).length === 1) {
+                return parseInt(model.get('slug').replace(/^new-scenario-/g, ''));
+            }
+        }), undefined);
+        if (!_.isEmpty(numbers)) {
+            var max = _.max(numbers) + 1;
+            return 'New Scenario ' + max;
+        } else {
+            return 'New Scenario ' + this.scenarioCounter();
+        }
```
kdeloach: Consider making simpler? `i = 0; while (this.collection.findWhere({ name: 'Foo ' + i++  })); return 'Foo ' + i;` etc.

src/mmw/js/src/modeling/views.js
```diff
@@ -130,6 +177,64 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
...
+            if (keycode == ENTER_KEYCODE) {
+                // Don't add line returns to the text.
+                e.preventDefault();
+                self.triggerMethod('rename:scenario', self.model, $(this).text());
+            }
+        });
+
+        $nameField.on('blur', function(e) {
+           self.triggerMethod('rename:scenario', self.model, $(this).text());
+        });
```
kdeloach: If you are aching to try out promises you may be able to abstract away some of this into a utility function. Ex. `util.makeContentEditable($el).done(renameScenario).fail(revertScenario)` etc.

src/mmw/js/src/modeling/views.js
lliss: Good idea. Fixed this.

src/mmw/js/src/modeling/views.js
lliss: This doesn't work without actually triggering it manually.

I'm going to tie the active_scnario_slug to the model's active property but we need the model for a scenario to know if it is active to it can add appropriate classes at render time.

lliss:
@azaveaci rebuild

lliss:
I'm going to need to clean up the branch history but this should address the issues discussed above.

lliss:
@azaveaci rebuild

lliss:
It's finally green! @mmcfarland You want to take another look or want me to reassign?

src/mmw/js/src/modeling/models.js
```diff
@@ -40,6 +40,17 @@ var ProjectModel = Backbone.Model.extend({
         }
     },
 
+    set: function(attributes, options) {
+        if (attributes === 'active_scenario_slug') {
```
caseypt: `active_scenario_slug` is gone as of #249. There is also a custom method for setting the active scenario, `setActiveScenario`.

lliss: Yes. Rebase fallout... Fixing.

src/mmw/js/src/app.js
```diff
@@ -50,13 +51,17 @@ var App = new Marionette.Application({
...
 
     showLoginModal: function() {
-        $('#login').find('#id_username').val("");
-        $('#login').find('#id_password').val("");
-        $('#login').modal('show');
+        var userView = new userViews.LoginModalView({
+            model: new userModels.LoginFormModel({}),
+            app: this
+        }).render();
+        userView.$el.find('#id_username').val('');
```
kdeloach: Consider creating some helper methods so we don't have to manipulate `$el` outside of its view.

src/mmw/js/src/core/templates/shareModal.html
```diff
@@ -0,0 +1,19 @@
+<div class="modal-dialog modal-dialog-sm">
+    <div class="modal-content">
+        <div class="modal-header">
+            <h2>Share Project</h2>
+        </div>
+        <div class="modal-body">
+            <div class="custom-input-group">
+              <input type="text" required value="<%- url %>">
```
kdeloach: Old template syntax here.

src/mmw/js/src/modeling/models.js
```diff
@@ -85,6 +85,57 @@ var ProjectModel = Backbone.Model.extend({
...
+
+    updateScenarioName: function(model, newName) {
+        // Bail early if the name actually didn't change.
+        if (model.get('name') === newName) {
+            return;
+        }
+
+        var match = _.find(this.get('scenarios').models, function(model) {
+            return model.get('name') === newName;
+        });
```
kdeloach: Consider using Collection [findWhere](http://backbonejs.org/#Collection-findWhere).

src/mmw/js/src/modeling/views.js
```diff
@@ -74,12 +88,17 @@ var ScenariosView = Marionette.LayoutView.extend({
         'click @ui.tab': 'onScenarioTabClicked'
     },
 
+   childEvents: {
+        'rename:scenario': function(view, model, newName) {
+            this.triggerMethod('rename:scenario', model, newName);
```
kdeloach: We could get rid of all this event bubbling code if we put `updateScenarioName` and `createNewScenario` on `ScenariosCollection`.

src/mmw/js/src/modeling/views.js
```diff
@@ -91,11 +110,7 @@ var ScenariosView = Marionette.LayoutView.extend({
     },
 
     addScenario: function() {
-        var scenario = new models.ScenarioModel({
-                name: 'New Scenario'
-            });
-        this.collection.add(scenario);
-        this.collection.setActiveScenario(scenario.cid);
+        this.triggerMethod('create:scenario');
```
kdeloach: Same comment as above, we don't need to add a layer of indirection if the new helper methods are on the collection instead of the project.

src/mmw/js/src/modeling/views.js
```diff
@@ -114,6 +129,12 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
         role: 'presentation'
     },
 
+    onDomRefresh: function() {
+        if (this.model.get('active')) {
+            this.$el.addClass('active');
+        }
+    },
```
kdeloach: I think this is no longer needed because of https://github.com/lliss/model-my-watershed/blob/feature/lol/scenario-dropdown/src/mmw/js/src/modeling/views.js#L138.

src/mmw/js/src/modeling/views.js
```diff
@@ -126,6 +147,77 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
 
     onRender: function() {
         this.$el.toggleClass('active', this.model.get('active'));
+    },
+
+    updateSlug: function() {
+        this.model.slugifyName();
+        this.render();
+    },
```
kdeloach: This may be obsolete (doesn't appear to be called by anything).

mmcfarland:
Does this need another look after a rebase?

lliss:
@azaveaci rebuild

src/mmw/js/src/app.js
```diff
@@ -56,14 +56,15 @@ var App = new Marionette.Application({
     },
 
     showLoginModal: function() {
-        new userViews.LoginModalView({
+        this.loginModal = new userViews.LoginModalView({
```
mmcfarland: Consider keeping the naming convention consistent and calling this `loginView` or `loginModalView`, then adding a `show` and `hide` method on it so we don't have to manually manipulate the `$el.modal`

src/mmw/js/src/core/views.js
mmcfarland: Still think that if this modal view is doing the actual destroying, you should pass the return of `destroy` so caller can handle success/failure.

http://backbonejs.org/#Model-destroy

src/mmw/js/src/modeling/views.js
```diff
@@ -138,7 +220,8 @@ var ScenarioTabPanelsView = Marionette.CollectionView.extend({
     attributes: {
         role: 'tablist'
     },
-    childView: ScenarioTabPanelView
+
+    childView: ScenarioTabPanelView,
```
mmcfarland: Trailing comma.

mmcfarland:
Two last issues I've found:

* Rename function allows an empty string as input, which squishes the tab:
screenshot from 2015-06-02 12 46 17

* Delete function appears to leave the scenario header in a blank state, should activate the nearest younger sibling instead.
screenshot from 2015-06-02 12 46 31

mmcfarland:
Issues appear fixed, +1.  I think @kdeloach comments were good about collection methods, so I'm going to put some issues in for later.

lliss:
My feeling was that the project has the responsibility of maintaining unique scenario names since that is the part of the application that cares that the names are unique. I wasn't fond of adding logic to the collection itself.

PR #246, opened 2015-05-29
Updated style to simple model.

ctaylo37:
Thumbnails on the sidebar were stacking to four rows, so I increased the sidebar width.


rajadain:
:+1: Tested on large monitor as well as my 13" Retina screen. Icons are consistently in a grid of 4x3 instead of 3x4. Good work!

mmcfarland:
+1 Looks good on landscape orientation on iPad.  Portait is cutoff (at least on emulator).

ctaylo37:
Yea we need to lock this down to landscape orientation only.

PR #247, opened 2015-05-29
Add livereload

kdeloach:
Livereload can be enabled with the following command (browser extension
is required):

> ./scripts/npm.sh run livereload

You'll need to run `vagrant reload app` because there's an additional forwarded port for the livereload server.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -128,7 +128,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       guest: 80,
       host: 8000
     }.merge(VAGRANT_NETWORK_OPTIONS)
-    # Livereload server (for gulp watch)
+    # Livereload server
     app.vm.network "forwarded_port", {
       guest: 35729,
       host: 35729,
--- a/src/mmw/package.json
+++ b/src/mmw/package.json
@@ -3,7 +3,8 @@
   "version": "0.0.1",
   "description": "",
   "scripts": {
-    "test": "test"
+    "test": "test",
+    "livereload": "node_modules/.bin/livereloadify /var/www/mmw/static --glob '**/*.*'"
   },
   "repository": {
     "type": "git",
@@ -28,6 +29,7 @@
     "jstify": "0.9.0",
     "leaflet": "0.7.3",
     "leaflet-draw": "0.2.3",
+    "livereloadify": "2.0.0",
     "lodash": "3.6.0",
     "mocha": "2.0.1",
     "node-sass": "3.1.1",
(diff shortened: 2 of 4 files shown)


Vagrantfile
```diff
@@ -138,6 +138,11 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       guest: 7357,
       host: 7357
     }.merge(VAGRANT_NETWORK_OPTIONS)
+    # Livereload
+    app.vm.network "forwarded_port", {
```
lliss: https://github.com/kdeloach/model-my-watershed/blob/feature/livereload/Vagrantfile#L133
is the same as
https://github.com/kdeloach/model-my-watershed/blob/feature/livereload/Vagrantfile#L143

kdeloach: That was smart... reverting.

kdeloach: Updated.

kdeloach:
Rebased.

caseypt:
Is this the correct extension: https://chrome.google.com/webstore/detail/livereload/jnihajbhpnppcggbcgedagnkighmdlei?hl=en?

kdeloach:
Yup.

caseypt:
Cool. Consider linking to that in the README.

src/mmw/package.json
```diff
@@ -28,6 +29,7 @@
     "jstify": "0.9.0",
     "leaflet": "0.7.3",
     "leaflet-draw": "0.2.3",
+    "livereloadify": "2.0.0",
```
rajadain: Should this be added to `devDependencies` so that browserify doesn't bundle it? Or is that taken care of separately?

rajadain: Ah I see we explicitly specify dependencies in `bundle.sh`. Please ignore the previous comment.

lliss:
Yeah would be useful to know the proper browser extension and to note that it's a chrome only thing right now.

+1 Tested.

kdeloach:
You should be able to install the livereload extension for FF (have not tested it though).

http://livereload.com/extensions/

src/mmw/package.json
kdeloach: This probably should go in `devDependencies` but I think that distinction matters more for npm modules and less for standalone applications like this.

kdeloach: Browserify should only bundle dependencies referenced in `require` statements in your target JS files.

rajadain: Ah, also true. Thanks for clarifying.

kdeloach:
Updated the README with a link to the livereload browser extension page.

PR #249, opened 2015-05-30
Fix tab switching issues

### LINKED ISSUE #164: Fix scenario drop-down menu
It doesn't quite work in the prototypes.

It's currently activating the correct scenario toolbar, but it 1) doesn't activate the scenario tab and 2) doesn't clear the active class from a menu item when another menu item is selected. 

selection_016

kdeloach:
Replaces usage of `slug` with `cid` throughout. This fixes an edge case
where tabs could fall out of sync with their tab panels when the `slug`
changed.

Ex.
image

This also defines a standard way of setting the active tab by using
a method on the `ScenariosCollection` called `setActiveScenario`.

Fixes #164

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -27,8 +27,17 @@ var ProjectModel = Backbone.Model.extend({
         // itself in the future, with more modification
         // and input control data.
         this.set('model_package', 'tr-55');
-        this.set('active_scenario_slug', 'current-conditions');
         this.set('taskModel', new Tr55TaskModel());
+
+        this.makeFirstScenarioActive();
+    },
+
+    makeFirstScenarioActive: function() {
+        var scenariosColl = this.get('scenarios'),
+            scenario = scenariosColl.first();
+        if (scenario) {
+            scenariosColl.setActiveScenario(scenario.cid);
+        }
     },
 
     saveAll: function() {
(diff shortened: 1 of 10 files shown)


caseypt:
Nice work. Going to test this out now.

We should also remove this line: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/controllers.js#L37

kdeloach:
Do you think it makes more sense to have "area of interest" on the Project model or the Map model? I was leaning towards Project.

caseypt:
Yeah, I think it now makes more sense to have this on  project. Though that would involve us creating the Project model sooner. But I think that makes more sense too.

src/mmw/js/src/modeling/models.js
```diff
@@ -82,27 +91,30 @@ var ProjectModel = Backbone.Model.extend({
...
+        return slug;
     }
 });
 
 var ScenariosCollection = Backbone.Collection.extend({
     model: ScenarioModel,
-    comparator: 'created_at'
+    comparator: 'created_at',
+
+    setActiveScenario: function(cid) {
```
caseypt: +1 to custom collection methods for stuff like this.

caseypt:
+1. Tested. Works. I was able to generate a bunch of scenarios and switch between them, as well as rename some of the scenarios and have the tabs still work.

mmcfarland:
+1 Looks good.  I discussed a similar implementation with @lliss on friday, so you may want to check this out.

PR #250, opened 2015-05-30
Add bundle --minify flag

### LINKED ISSUE #68: minify vendor.js bundle
Minifying the vendor bundle should reduce the amount of time it takes to load the page over a slow network connection. I tried to do this as part of #60 but had trouble. I added uglifyify as a dependence in package.json, and changed the `bundle.sh` command 

```
    $BROWSERIFY -r backbone \
        -r backbone.marionette -r bootstrap -r bootstrap-select \
        -r leaflet -r leaflet-draw -r lodash -r underscore -r chai \
        -o ${STATIC_JS_DIR}vendor.js $EXTRA_ARGS & \
```

to have the flag `-g uglifyify`. This resulted in minification when running `bundle.sh` on the VM directly, but resulted in a file of 0 bytes when running it via `scripts/bundle.sh` from my host machine.

kdeloach:
Use the `--minify` flag to generate minified bundles. Note that this is
a slow operation and also disables source maps.

Could be used to minify *just* the vendor bundle for development and
testing over slow network connections.

    ./scripts/bundle.sh --vendor --minify

Fixes #68

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,9 @@ Test bundles are not created unless the `--tests` flag is used.
 In general, you should be able to combine `--vendor`, `--tests`, `--debug`,
 and `--watch` and have it behave as you would expect.
 
+You can also minify bundles by using the `--minify` flag. This operation is
+not fast, and also disables source maps.
+
 The `--list` flag displays module dependencies and does not actually generate
 any bundles. It doesn't make sense to combine this with `--watch`.
 This flag is for troubleshooting purposes only.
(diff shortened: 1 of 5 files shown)


kdeloach:
Rebased.

rajadain:
+1. Tested with `bundle.sh --minify` and `bundle.sh --vendor --minify`: the first minified only the app files, and the second only the vendor files. `main.js` goes from 160 KB to 95.4 KB, and `vendor.js` goes from 1.7 MB to 646 KB, which is great.

lewfish:
I tested as Terrence did, and also tried to bundle the tests with and without minify which worked as expected. I also ran `vagrant provision app` which didn't causes any errors. I'm not sure how to switch ansible into development vs. production mode though.

kdeloach:
In bash you can provision as production like this `VAGRANT_ENV=PRODUCTION vagrant provision app`.

lewfish:
It might be worth putting a comment in `static-files.yml` essentially saying what you just said.

kdeloach:
Good idea, will update.

Issue #79, opened 2015-04-21
Make numeric formatter available in templates


caseypt:
I think it would be cool if there was always functions available to do formatting on numeric values in templates. Or, maybe if you knew you were going to need them, it should be simple to bring them into a particular template.

Maybe use [numeral.js](http://numeraljs.com/)? Or, if we switch to [Nunjucks](http://mozilla.github.io/nunjucks), some of this stuff looks like it's [built in](http://mozilla.github.io/nunjucks/templating.html#math).

PR #251, opened 2015-05-31
Add nunjucks

kdeloach:
Connects #79


rajadain:
+1 I like that the syntax is now basically identical to the Django template syntax, making context switching much easier. From what I can read, it seems like the performance impact is minimal and it supports async function calls which is great going forward.

mmcfarland:
+1 Templates all appear to be working.  Strange to see pythonesque code in the frontend templates, but only because I've somehow convinced myself that it's not strange to have javascriptesque code otherwise.  Luckily looks like a pretty small change overall.

kdeloach:
I also verified that filters, blocks, and including other templates works.

caseypt:
I can still take a look at this if need be after it's rebased.

kdeloach:
Rebased.

caseypt:
+1. Tested.

PR #263, opened 2015-06-02
Use absolute path to user endpoints

mmcfarland:
Login checks and requests were not sent to the correct URL from pages
other than index such as the `model/<id>` page.


caseypt:
+1. Tested. I'm now shown as logged in when visiting the `model` or `analyze` directly.

PR #265, opened 2015-06-02
Set current conditions scenario as active

caseypt:
* Fixes an issue where no scenario was set as active when
creating a new project.


kdeloach:
We should also consider removing this code if it no longer works https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/models.js#L44

caseypt:
@kdeloach That code works when loading an existing project.

kdeloach:
Ah, thanks. Haven't tested creating new projects yet.

caseypt:
Closing this. Fixed in #245

Issue #151, opened 2015-05-13
Login modal design tasks


caseycesari:
- Need design for state when we are waiting for the server to return whether or not the login worked.
- Need design for validation errors, i.e. someone hitting enter without entering a username.
- Need sass for "incorrect username or password" as shown in mockups.

caseypt:
This is still an issue. The most obvious issue may be the unstyled validation messages:

![image](https://cloud.githubusercontent.com/assets/1042475/19930138/6826d936-a0dd-11e6-95a8-296257e44768.png)

ajrobbins:
Addressing with redesign in #2254

PR #268, opened 2015-06-02
ITSI Registration front-end

rajadain:
Adds a login and registration front-end for users logging in with ITSI.

**Before You Begin**

Please check if the `MMW_ITSI_SECRET_KEY` variable is set on your `app` VM:

    vagrant ssh app -c 'cat /etc/mmw.d/env/MMW_ITSI_SECRET_KEY'

If it isn't, you must reprovision the `app` VM with the `MMW_ITSI_SECRET_KEY` variable set, like so:

    MMW_ITSI_SECRET_KEY=*** vagrant provision app

The secret key is available in the wiki.

**Testing Instructions**

 1. Open [http://localhost:8000/](http://localhost:8000/). You should see a new login page that matches the latest mockups.
 2. Click on "ITSI". You should be taken to the ITSI portal.
 3. Log in with either the `tstudent2` or `mmwterence` account to ITSI (credentials are available in the wiki). You should be redirected to the registration front-end, and the `username`, `first_name`, and `last_name` fields should be pre-populated.
 4. Check "I agree to something" and click "Sign Up". If there aren't any validation errors, the registration modal should dismiss and you should be logged in, apparent from the username in the top right hand corner.
 5. Refresh the page to check that you are still logged in as this newly created user.
 6. Log out of this account. Click "Login", and again click "ITSI". You will be redirected to the ITSI portal again, but since you recently logged in there you should be redirected right back to our app which should automatically log you in.

**Alternate Paths**

 * Try going to [`/register`](http://localhost:8000/register) directly and registering.
 * Try dismissing the modal during registration.
 * Try signing up as an already existing username.

**Notes**

 * The mockup has a line from edge to edge behind the "Or continue as" text which I don't know how to add:  
screen shot 2015-06-02 at 4 28 32 pm
 * We pass back the username, first and last names received from ITSI to the registration modal via the query string. Is this a security / aesthetic issue? Should we use cookies or some other "under the hood" mechanism which is uglier in code but cleaner in the UI?
 * Unlike regular sign-up / login, the ITSI login requires full page refreshes and is not an XHR-only operation. This should be kept in mind when creating mechanisms to save temporary work in case the user decides to create an account using ITSI mid-way through an unsaved project.
 * We now have a handful of modal views that are very similar in basic operation. Perhaps we should refactor the commonalities into a utility class and derive from that instead.
 * I needed to add some styling for the login modal updates. I decided to add it under `sass/base/_login.scss` and not `sass/pages/_login.scss` because the login is a modal, not a page, and would be available throughout the app, not just in a particular view. If this was not the right place to add it, please let me know and I will correct it.

Connects #180

### DIFF
--- a/src/mmw/apps/home/urls.py
+++ b/src/mmw/apps/home/urls.py
@@ -15,4 +15,5 @@ urlpatterns = patterns(
     url(r'^analyze$', home_page, name='analyze'),
     url(r'^compare$', compare, name='compare'),
     url(r'^error', home_page, name='error'),
+    url(r'^sign-up', home_page, name='sign_up'),
 )
--- a/src/mmw/apps/user/views.py
+++ b/src/mmw/apps/user/views.py
@@ -33,7 +33,8 @@ def login(request):
             response_data['username'] = user.username
             response_data['guest'] = False
         else:
-            response_data['result'] = 'error'
+            response_data['errors'] = ['Invalid username or password']
+            response_data['guest'] = True
             status_code = status.HTTP_400_BAD_REQUEST
     elif request.method == 'GET':
         user = request.user
@@ -124,49 +125,55 @@ def itsi_register(request):
 @decorators.api_view(['POST'])
 @decorators.permission_classes((AllowAny, ))
 def sign_up(request):
-    response_data = {
-        'username_valid': True,
-        'email_valid': True,
-        'password_valid': True,
-    }
-    status_code = status.HTTP_200_OK
     view = RegistrationView()
     form = RegistrationFormUniqueEmail(request.POST)
-    if form.is_valid():
-        view.register(request, **form.cleaned_data)
-    else:
-        if 'username' not in form.cleaned_data:
-            response_data['username_valid'] = False
-        if 'password1' not in form.cleaned_data or \
-           'password2' not in form.cleaned_data or \
-           form.cleaned_data['password1'] != form.cleaned_data['password2']:
-            response_data['password_valid'] = False
-        if 'email' not in form.cleaned_data:
-            response_data['email_valid'] = False
-        status_code = status.HTTP_400_BAD_REQUEST
 
-    return Response(data=response_data, status=status_code)
+    if form.is_valid():
+        user = view.register(request, **form.cleaned_data)
+        response_data = {'result': 'success',
+                         'username': user.username,
+                         'guest': False}
+        return Response(data=response_data,
+                        status=status.HTTP_200_OK)
+    else:
+        errors = []
+        if 'username' not in form.cleaned_data:
+            errors.append("Username is invalid or already in use")
+        if 'password1' not in form.cleaned_data:
+            errors.append("Password must be specified")
+        if 'password2' not in form.cleaned_data or \
+           form.cleaned_data['password1'] != form.cleaned_data['password2']:
+            errors.append("Passwords do not match")
+        if 'email' not in form.cleaned_data:
+            errors.append("Email is invalid or already in use")
+
+        if len(errors) == 0:
+            errors.append("Invalid data submitted")
+
+        response_data = {"errors": errors}
+        return Response(data=response_data,
+                        status=status.HTTP_400_BAD_REQUEST)
 
 
 @decorators.api_view(['POST'])
 @decorators.permission_classes((AllowAny, ))
 def forgot(request):
-    response_data = {
-        'email_valid': True
-    }
-    status_code = status.HTTP_200_OK
     form = PasswordResetForm(request.POST)
+
     if form.is_valid():
         email = form.cleaned_data['email']
         try:
             # If there are active user(s) that match email
             next(form.get_users(email))
             form.save(request=request)
+            response_data = {'result': 'success',
+                             'guest': True}
+            status_code = status.HTTP_200_OK
         except StopIteration:
-            response_data['email_valid'] = False
+            response_data = {'errors': ["Email cannot be found"]}
             status_code = status.HTTP_400_BAD_REQUEST
     else:
-        response_data['email_valid'] = False
+        response_data = {'errors': ["Email is invalid"]}
         status_code = status.HTTP_400_BAD_REQUEST
 
     return Response(data=response_data, status=status_code)
--- a/src/mmw/apps/user/views.py
+++ b/src/mmw/apps/user/views.py
@@ -94,10 +94,38 @@ def itsi_auth(request):
     else:
         # User did not authenticate. Save their ITSI ID and send to /register
         request.session['itsi_id'] = itsi_user['id']
-        return redirect('/register')  # TODO Make this Backbone route
+        return redirect(
+            '/sign-up/itsi/{username}/{first_name}/{last_name}'.format(
+                **itsi_user['extra']
+            )
+        )
 
 
-def itsi_register(request):
+@decorators.api_view(['POST'])
+@decorators.permission_classes((AllowAny, ))
+def itsi_sign_up(request):
+    # Validate request
+    errors = []
+    if 'itsi_id' not in request.session:
+        errors.append("There was an error in authenticating you with ITSI")
+
+    if 'username' not in request.POST or not request.POST.get('username'):
+        errors.append("Username must be specified")
+    elif User.objects.filter(username=request.POST.get('username')).exists():
+        errors.append("Username already exists")
+
+    if 'first_name' not in request.POST or not request.POST.get('first_name'):
+        errors.append("First name must be specified")
+    if 'last_name' not in request.POST or not request.POST.get('last_name'):
+        errors.append("Last name must be specified")
+    if 'agreed' not in request.POST or not request.POST.get('agreed'):
+        errors.append("You must agree to the terms")
+
+    if len(errors) > 0:
+        response_data = {"errors": errors}
+        return Response(data=response_data,
+                        status=status.HTTP_400_BAD_REQUEST)
+
     itsi_id = request.session['itsi_id']
 
     # Create new user with given details and no email address or password
@@ -119,7 +147,11 @@ def itsi_register(request):
     user = authenticate(itsi_id=itsi_id)
     auth_login(request, user)
 
-    return redirect('/')
+    response_data = {'result': 'success',
+                     'username': user.username,
+                     'guest': False}
+    return Response(data=response_data,
+                    status=status.HTTP_200_OK)
 
 
 @decorators.api_view(['POST'])
--- a/src/mmw/js/src/main.js
+++ b/src/mmw/js/src/main.js
@@ -39,6 +39,13 @@ App.on('start', function() {
         router.navigate($(this).data('url'), { trigger: true });
     });
     Backbone.history.start({ pushState: true });
+
+    // Show login modal only if landing on home page
+    if (Backbone.history.getFragment() === "") {
+        this.getUserOrShowLogin();
+    } else {
+        this.user.fetch();
+    }
 });
 
 App.start();
--- a/src/mmw/js/src/user/models.js
+++ b/src/mmw/js/src/user/models.js
@@ -175,9 +175,54 @@ var ForgotFormModel = ModalBaseModel.extend({
     }
 });
 
+var ItsiSignUpFormModel = ModalBaseModel.extend({
+    defaults: {
+        username: null,
+        first_name: null,
+        last_name: null,
+        agreed: false
+    },
+
+    url: '/user/itsi/sign_up',
+
+    validate: function(attrs) {
+        var errors = [];
+
+        if (!attrs.username) {
+            errors.push('Please enter a username');
+        }
+
+        if (!attrs.first_name) {
+            errors.push('Please enter a first name');
+        }
+
+        if (!attrs.last_name) {
+            errors.push('Please enter a last name');
+        }
+
+        if (!attrs.agreed) {
+            errors.push('Please check the agreement');
+        }
+
+        if (errors.length) {
+            this.set({
+                'client_errors': errors,
+                'server_errors': null
+            });
+            return errors;
+        } else {
+            this.set({
+                'client_errors': null,
+                'server_errors': null
+            })
+        }
+    }
+});
+
 module.exports = {
     UserModel: UserModel,
     LoginFormModel: LoginFormModel,
     SignUpFormModel: SignUpFormModel,
-    ForgotFormModel: ForgotFormModel
+    ForgotFormModel: ForgotFormModel,
+    ItsiSignUpFormModel: ItsiSignUpFormModel
 };
--- a/src/mmw/js/src/user/models.js
+++ b/src/mmw/js/src/user/models.js
@@ -16,7 +16,10 @@ var UserModel = Backbone.Model.extend({
         return this.fetch({
             type: 'POST',
             url: '/user/login',
-            data: attrs
+            data: {
+                'username': attrs.username,
+                'password': attrs.password
+            }
         });
     },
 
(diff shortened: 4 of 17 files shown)


src/mmw/apps/home/urls.py
```diff
@@ -15,4 +15,5 @@
     url(r'^analyze$', home_page, name='analyze'),
     url(r'^compare$', compare, name='compare'),
     url(r'^error', home_page, name='error'),
+    url(r'^register', home_page, name='register'),
```
rajadain: This changeset demonstrates why trailing commas are useful. Had the previous line not had a trailing comma, it would also show up as changed.

src/mmw/apps/user/views.py
```diff
@@ -93,10 +93,36 @@ def itsi_auth(request):
     else:
         # User did not authenticate. Save their ITSI ID and send to /register
         request.session['itsi_id'] = itsi_user['id']
-        return redirect('/register')  # TODO Make this Backbone route
+        return redirect('/register/{username}/{first_name}/{last_name}'.format(
+            **itsi_user['extra']
```
rajadain: The `itsi_user` object has the following structure:

```json
{
	"provider":"",
	"id":"",
	"info": {
		"email":""
	},
	"extra": {
		"first_name":"",
		"last_name":"",
		"full_name":"",
		"username":"",
		"user_id":999,
		"domain":""
	}
}
```

src/mmw/apps/user/views.py
```diff
@@ -93,10 +93,36 @@ def itsi_auth(request):
...
+    if 'last_name' not in request.POST or not request.POST.get('last_name'):
+        errors.append("Last name must be specified")
+    if 'agreed' not in request.POST or not request.POST.get('agreed'):
+        errors.append("You must agree to the terms")
+
+    if len(errors) > 0:
+        response_data = {"errors": errors}
+        return Response(data=response_data,
+                        status=status.HTTP_400_BAD_REQUEST)
+
```
rajadain: If this style is too verbose, of if there is a more Pythonic way of doing this, please let me know.

src/mmw/js/src/main.js
```diff
@@ -39,6 +39,13 @@ App.on('start', function() {
         router.navigate($(this).data('url'), { trigger: true });
     });
     Backbone.history.start({ pushState: true });
+
+    // Show login modal only if landing on home page
+    if (Backbone.history.getFragment() === "") {
+        this.getUserOrShowLogin();
+    } else {
+        this.user.fetch();
```
rajadain: In case this is not the home page, still try to automatically log user in.

mmcfarland:
I realize you were following the mockups, but I think the register modal needs an "ITSI" button as well.  If the user visits `/register` they are not presented with an option and may incorrectly sign up as a regular user.  @ctaylo37

rajadain:
Actually, `/register` right now is exclusively for ITSI users. Perhaps I should change the routing so that `/register` brings up the regular sign-up modal (which currently has no route), but also has an ITSI button, and that would return the user to `/register/itsi` which would be the modal in this PR. How does that sound?

mmcfarland:
I think that proposed rerouting is good/fine, but I think if a user visits the itsi registration page without having been redirected from the portal, we should not offer registration screen - since they can't use it - but redirect them somewhere that they can use, maybe the base login view?

mmcfarland:
Really good stuff, appreciate the logical commits and instructions.  I think refactoring the common registration code is worthwhile now, but otherwise looks good.

src/mmw/apps/home/urls.py
caseypt: Never considered that before. Pretty cool.

src/mmw/apps/user/views.py
```diff
@@ -93,10 +93,36 @@ def itsi_auth(request):
...
 
+@decorators.api_view(['POST'])
+@decorators.permission_classes((AllowAny, ))
 def itsi_register(request):
+    # Validate request
+    errors = []
+    if 'itsi_id' not in request.session:
+        errors.append("There was an error in authenticating you with ITSI")
+
+    if 'username' not in request.POST or not request.POST.get('username'):
```
kdeloach: May be able to simplify with `if not request.POST.get('username', None)`?

rajadain: YES thank you that's what I was looking for. This stuff is impossible to phrase into a Google query, but easily caught by those more experienced with the language's idioms.

src/mmw/js/src/user/models.js
```diff
@@ -197,9 +197,52 @@ var ForgotFormModel = Backbone.Model.extend({
...
 });
 
+var ItsiRegisterModel = Backbone.Model.extend({
+    defaults: {
+        username: null,
+        first_name: null,
+        last_name: null,
+        agreed: false,
+        clientErrors: null,
+        serverErrors: null
```
kdeloach: We discussed using snake case for model attributes from now on (for parity with DRF).

src/mmw/js/src/user/views.js
```diff
@@ -279,8 +281,93 @@ var ForgotModalView = Marionette.ItemView.extend({
     }
 });
 
+var ItsiRegisterModalView = Marionette.ItemView.extend({
```
kdeloach: Consider adding a model property here to designate which model this view is expecting. I've found this helpful but I'm not sure what others think.

rajadain: Can you link me an example?

src/mmw/js/src/user/views.js
```diff
@@ -279,8 +281,93 @@ var ForgotModalView = Marionette.ItemView.extend({
...
+        'registerModal': '#itsi-register-modal'
+    },
+
+    events: {
+        'click @ui.registerButton': 'validate',
+        'keyup @ui.registerModal': 'handleKeyUpEvent'
+    },
+
+    initialize: function(options) {
+        this.app = options.app;
```
kdeloach: Marionette views have a helper function called [mergeOptions](http://marionettejs.com/docs/v2.4.1/marionette.view.html#viewmergeoptions) that we may be able to utilize here.

src/mmw/apps/user/views.py
kdeloach: Is this an ITSI-ism? Seems strange to have username, first name, and last name in a URL.

kdeloach:
+1 Good design pattern for modal workflow.

src/mmw/apps/user/views.py
rajadain: Yeah. We get these values from ITSI and we'd like to populate them in the form on the client side. We can either send this via the URL, or cache it in the server and have the client do a second AJAX request, or set a cookie and have the client read that and destroy it once its done. This seemed like the easiest way to do it.

src/mmw/js/src/user/views.js
kdeloach: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/views.js#L33

rajadain: And this model property gets overwritten with the actual model when instantiating the view? I can imagine that being a little confusing when looking at the code. If all we want to do is annotate it, then we could do it in a comment.

kdeloach: I could definitely imagine some people being confused.

rajadain:
As I consolidate the modals, I think that we should have consistent naming for all user-facing things. Currently, all UI elements for signing up say "Sign Up", but the route is `/register`. Should I change that to `/signup` (or `/sign-up`, which is easier to read, or `/sign_up` which is consistent with internal API naming) to match?

mmcfarland:
No real preference, but I agree with consistency.  Same with Sign In/Log in/Login.  I suppose if @ctaylo37 named the buttons "Sign Up" we should go with that nomenclature.  I have a slight preference for `/sign-up` in that case, personally.

rajadain:
All user modals now inherit from `baseModal.html` template, `ModalBaseModel` model, and `ModalBaseView` view. The routing has been updated to be the following:

 * `/` Will try to automatically log user in, if not then will show the user login modal
 * `/sign-up` Will open the regular user sign-up modal
 * `/sign-up/itsi(/username)(/first_name)(/last_name)` Will open the ITSI user sign-up modal, pre-populated with relevant values

Most of the later work for this PR has been code refactoring, so I would appreciate any comments there. The diffs might be hard to read, and I would recommend looking at the final files, especially the front-end templates, views, and models, where the extension should be more clearly apparent.

Since I've refactored a lot of his work here, I would appreciate some testing and feedback from @lewfish to make sure nothing is broken.

This code does make certain assumptions that will have to be altered in the future. Currently we navigate to `/` as soon as a sign-up modal is dismissed, but once we allow anonymous users to have projects and give them an option to save them by creating an account, that would have to be changed to return them to the same page they were at when they started the sign-up process.

src/mmw/js/src/user/models.js
```diff
@@ -42,157 +45,140 @@ var UserModel = Backbone.Model.extend({
...
-        }
-
-        if (errors.length >= 1) {
-            this.set('clientErrors', errors);
+        if (!attrs.username) errors.push('Please enter a username');
+        if (!attrs.email) errors.push('Please enter an email address');
+        if (!attrs.password1) errors.push('Please enter a password');
+        if (!attrs.password2) errors.push('Please repeat the password');
+        if (!(attrs.password1 === attrs.password2))
+            errors.push('Passwords do not match');
```
kdeloach: It might seem silly, but it's a good habit to get into to use curly braces even for conditionals with only 1 statement.

src/mmw/js/src/user/models.js
```diff
@@ -42,157 +45,140 @@ var UserModel = Backbone.Model.extend({
...
-            errors.push('Please enter a password');
-        }
-
-        if (!attrs.password2) {
-            errors.push('Please repeat the password');
-        }
-
-        if (!attrs.agreed) {
-            errors.push('Please check the agreement');
-        }
```
kdeloach: I prefer this style over the new condensed version.

src/mmw/js/src/user/templates/baseModal.html
```diff
@@ -0,0 +1,59 @@
...
+                <h2 class="light">Model My Watershed</h2>
+            {% endblock %}
+        </div>
+
+        <div class="modal-body">
+            {% block preamble %}
+            {% endblock %}
+
+            {% block errors %}
+                {% if client_errors %}
```
kdeloach: In Jinja, `for` loops don't execute if the iterable is empty. I'd be curious to see if it's the same with nunjucks. If so, we can eliminate this conditional.

src/mmw/js/src/user/views.js
```diff
@@ -51,42 +55,113 @@ var LoginModalView = Marionette.ItemView.extend({
...
+        forgot:   '.forgot'
+    },
+
+    events: {
+        'click @ui.signUp': 'signUp',
+        'click @ui.forgot': 'forgot'
+    },
+
+    initialize: function(options) {
+        ModalBaseView.prototype.initialize(this, options);
```
kdeloach: Interesting technique. Typically inheritance chains will call super methods with `call`/`apply`. Ex. `ModalBaseView.prototype.initialize.call(this, options);`

src/mmw/js/src/user/templates/baseModal.html
rajadain: Removing the `if` would leave the stray `ul` tag. Will see if that shows up in the UI, and if not will remove the `if`s and combine both client and server errors into one list.

kdeloach: No, you're right. Fine as-is.

src/mmw/js/src/user/views.js
rajadain: Ah good point. Will fix.

rajadain: Done 004260247214167ba692cdd052f684cac8d91b4e

rajadain: Ah I moved too quickly. Simply adding `.call` actually breaks things and implementing it properly might be a larger refactor that might not even be required, given that things are working as is. Also, given that we only have one layer of inheritance after Marionette, things are simple enough to not require more complicated techniques.

kdeloach: Agreed.

lewfish:
Taking a look now.

lewfish:
In the sign-up modal, if you enter something that isn't an email address (eg. "c") into the email field, the "Email" label gets greyed out and overlaps with the input. If you change the type of the field to 'text' it should fix it.

lewfish:
Nice job refactoring. I tested the non-itsi signin/signup/forgot functionality and it all seems to work. This isn't due to this PR, but I've noticed that the user drop-down on the upper-right of the screen goes off the edge of the screen. Is there already an issue made for this? I didn't see one.
screen shot 2015-06-08 at 11 12 31 am 2

mmcfarland:
@lewfish Yes: https://github.com/WikiWatershed/model-my-watershed/issues/151

rajadain:
@lewfish regarding the email field, I was initially using it for accessibility purposes, but seems like it harmed more than helped. Fixed in aaeed1e.

The reason that happens is because of the `input:valid` selector here: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/sass/components/_inputs.scss#L73, which doesn't work when an invalid email is specified in that field.

Alternatively, we could add an `input:invalid` selector and highlight that field with red or something to communicate the invalidity to the user. If @ctaylo37 can get me a color for that, I can add that in.

rajadain:
Whoops seems like there's some tests that need to be looked at. Give me a few minutes ...

ctaylo37:
#e0412d

image

ctaylo37:
I would add the color to the variables.scss file

$danger:    #e0412d;

And use $danger for the css.

rajadain:
Unfortunately using the `:invalid` pseudo-selector isn't as straightforward as I'd hoped. It just highlights everything:

image

I think we're good with using just a text field for now, and can add in-line validation at a later stage. Thanks for the help, @ctaylo37.

rajadain:
Tests have been updated and are now fixed. I see we only have tests for the login modal. Testing the others is complicated since there is so much outside interaction: emails for the sign-up / forgot modals, and portal interaction for logging in with ITSI. Perhaps once we have our staging environment setup we can write some integration tests that allow for that. If there is a way to mock these dependencies please let me know and I will attempt to add tests.

rajadain:
Also, since this PR has almost 20 commits now, I will do some aggressive squashing and rebasing before merge to streamline history.

mmcfarland:
I can do a quick scan after the big rebase, just drop a note.

rajadain:
Reduced to 4 commits, rebased on top of develop. Testing instructions repeated again:

**Testing Instructions**

 1. Open [http://localhost:8000/](http://localhost:8000/). You should see a new login page that matches the latest mockups.
 2. Click on "ITSI". You should be taken to the ITSI portal.
 3. Log in with either the `tstudent2` or `mmwterence` account to ITSI (credentials are available in the wiki). You should be redirected to the sign-up front-end, and the `username`, `first_name`, and `last_name` fields should be pre-populated.
 4. Check "I agree to something" and click "Sign Up". If there aren't any validation errors, the registration modal should dismiss and you should be logged in, apparent from the username in the top right hand corner.
 5. Refresh the page to check that you are still logged in as this newly created user.
 6. Log out of this account. Click "Login", and again click "ITSI". You will be redirected to the ITSI portal again, but since you recently logged in there you should be redirected right back to our app which should automatically log you in.

mmcfarland:
Verified that the ITSI registration and log in still work.  :+1:

Issue #228, opened 2015-05-27
Duplicate Scenario


mmcfarland:
New option on the Scenario dropdown to duplicate the current Scenario.  This copies the entire model instance, including input and modifications.  Initial name should be "<Old Scenario Name> (n)" where `n` is an incremented integer.  Place the tab name in edit mode so it can be renamed immediately.

Issue #242, opened 2015-05-29
Add tests to model app


caseypt:
- Scenario tabs are rendered correctly with 0, 1, and many scenarios.
- Clicking the + button adds a new scenario.
- ~~The scenario identified by `Project.active_scenario_slug` is the active tab.~~
- The scenario with `active === true` is the active tab.
- Projects can be saved and retrieved.
- The modifications count in the toolbar is correct.
- The model output results are updated when a different scenario is selected.
- Existing modifications are added to the map when a scenario is selected.
- The project AoI is added to the map when the project is loaded.

PR #269, opened 2015-06-02
Duplicate scenario

caseypt:
Adds a "Duplicate" option to the scenario tab context menu that duplicates the selected scenario.

**To Test** (EDITED):
- Create a new project.
- Add a scenario. Keep the default name.
- Draw some modifications for that scenario.
- On that scenario, select "Duplicate". A new scenario should be added with the name "Copy of New Scenario 1". It should also have the modifications from "New Scenario 1".
- Duplicate "New Scenario 1" again. A new scenario should be added with the name "Copy of New Scenario 1 1" (a la Google Sheets). It should also have the modifications from "New Scenario 1".

- Create a new project.
- Add a scenario. Rename it to something with a trailing number.
- Perform the steps from above. Make sure the names are unique and the modifications get carried over.

Connects to #228

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -194,9 +194,14 @@ var ScenarioModel = Backbone.Model.extend({
     },
 
     initialize: function() {
-        // TODO: When fetching model or populating from a preloaded source
-        // this attribute should not be initialized as empty.
-        this.set('modifications', new ModificationsCollection());
+        // TODO: When #278 is resolved, we should be doing something like this
+        // to setup or restore modifications.
+        // var mods = this.get('modifications') === null ? [] : JSON.parse(this.get('modifications'));
+        var mods = this.get('modifications');
+
+        if (mods === null || typeof mods === 'string') {
+            this.set('modifications', new ModificationsCollection());
+        }
     },
 
     getSlug: function() {
@@ -221,6 +226,36 @@ var ScenariosCollection = Backbone.Collection.extend({
             }
             model.set('active', active);
         });
+    },
+
+    duplicateScenario: function(cid) {
+        var source = this.get(cid),
+            sourceMods = source.get('modifications').models,
+            newModel = new ScenarioModel({
+                is_current_conditions: false,
+                name: this.makeNewScenarioName('Copy of ' + source.get('name')),
+                modifications: new ModificationsCollection(sourceMods)
+            });
+
+        this.add(newModel);
+        this.setActiveScenario(newModel.cid);
+    },
+
+    // Generate a unique scenario name based off the baseName.
+    // Assumes a basic structure of "baseName X", where X is
+    // iterated as new scenarios with the same baseName are created.
+    // The first duplicate will not have an iterated X in the name.
+    // The second duplicate will be "baseName 1".
+    makeNewScenarioName: function(baseName) {
+        var existingNames = this.pluck('name');
+
+        if (!_.contains(existingNames, baseName)) {
+            return baseName;
+        }
+
+        for (var i=1; _.contains(existingNames, baseName + ' ' + i); i++);
+
+        return baseName + ' ' + i;
     }
 });
 
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -143,7 +143,7 @@ var ProjectModel = Backbone.Model.extend({
             return model.get('name') === newName;
         });
         if (match) {
-            alert('This name is already in use.');
+            window.alert('This name is already in use.');
             return;
         } else if (model.get('name') !== newName) {
             model.set('name', newName);
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -217,7 +217,7 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
 
     showShareModal: function() {
         if (App.user.get('guest')) {
-            alert('Guest cannot share scenarios. Please log in or register.');
+            window.alert('Guest cannot share scenarios. Please log in or register.');
             return;
         }
         var share = new coreViews.ShareModal({
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -153,20 +153,24 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
     },
 
     ui: {
-        share: '.share',
-        destroyConfirm: '.delete',
-        rename: '.rename',
-        print: '.print',
+        share: '[data-action="share"]',
+        destroyConfirm: '[data-action="delete"]',
+        rename: '[data-action="rename"]',
+        print: '[data-action="print"]',
+        duplicate: '[data-action="duplicate"]',
         nameField: '.tab-name'
     },
+
     events: {
         'click @ui.rename': 'renameScenario',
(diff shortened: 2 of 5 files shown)


rajadain:
When I make multiple duplicates, I find the names to be:

 * New Scenario 1
 * Copy of New Scenario 1
 * Copy of Copy of New Scenario 1
rather than that described:

image

I actually prefer this, but I am unfamiliar with the Google Sheet way of doing things, and what effect diverging from their convention may have.

rajadain:
Found that if there were too many long-named scenarios, I couldn't edit / rename / delete the later ones since their ▼ button is inaccessible:

image

Made #274 for this.

caseypt:
@rajadain Thanks. That is the correct behavior. My instructions are wrong. If you make multiple duplicates of "New Scenario 1", you should get "Copy of New Scenario 1", "Copy of New Scenario 1 1", etc.

rajadain:
Ah! Indeed and that does work as advertised:

image

rajadain:
I see that duplicates, even of earlier scenarios, are always added to the very end:

image

Not as part of this PR, but perhaps in the future we might consider making the tabs re-orderable, in case users want duplicates to be next to each other.

rajadain:
+1 Works very well. I've added issues for my comments which can be tackled at a later date.

src/mmw/js/src/modeling/models.js
```diff
@@ -221,6 +221,54 @@ var ScenariosCollection = Backbone.Collection.extend({
...
+        var source = this.get(cid),
+            sourceMods = source.get('modifications').models,
+            newModel = new ScenarioModel({
+                is_current_conditions: false,
+                name: this.makeNewScenarioName('Copy of ' + source.get('name'))
+            });
+
+        newModel.set('modifications', new ModificationsCollection(sourceMods, {
+            model: ModificationModel
+        }));
```
kdeloach: Consider merging this logic with the initialization of `newModel` above. Also, we shouldn't need the second argument.

src/mmw/js/src/modeling/models.js
```diff
@@ -221,6 +221,54 @@ var ScenariosCollection = Backbone.Collection.extend({
...
+
+        this.add(newModel);
+        this.setActiveScenario(newModel.cid);
+    },
+
+    // Generate a unique scenario name based off the baseName.
+    // Assumes a basic structure of "baseName X", where X is
+    // iterated as new scenarios with the same baseName are created.
+    // The first duplicate will not have an interated X in the name.
```
kdeloach: Typo "integrated".

src/mmw/js/src/modeling/models.js
```diff
@@ -221,6 +221,54 @@ var ScenariosCollection = Backbone.Collection.extend({
...
+    // The second duplicate will be "baseName 1".
+    makeNewScenarioName: function(baseName) {
```
kdeloach: Consider extracting this out into a helper function and using it in both places where we generate new scenario names. We could use [Collection-pluck](http://backbonejs.org/#Collection-pluck) to grab a list of existing scenario names to pass to this helper function.

src/mmw/js/src/modeling/models.js
```diff
@@ -221,6 +221,54 @@ var ScenariosCollection = Backbone.Collection.extend({
...
+                }
+            }
+        }), undefined);
+
+        if (!_.isEmpty(numbers)) {
+            var max = _.max(numbers) + 1;
+            return baseName + ' ' + max;
+        } else {
+            return baseName;
+        }
```
kdeloach: I think regexp is overkill for this. Could be made simpler with something like...

    var result = 'Copy of ' + name;
    if (!_.contains(existingNames, result)) return result;
    for (var i = 1; _.contains(existingNames, result + ' ' + i); i++);
    return result + ' ' + i;

src/mmw/js/src/modeling/views.js
```diff
@@ -225,6 +229,10 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
         });
         share.render();
         share.$el.modal('show');
+    },
+
+    duplicateScenario: function() {
+        this.model.collection.duplicateScenario(this.model.cid);
```
kdeloach: Never realized we could access a model's collection like this! Will definitely simplify code I write in the future.

kdeloach:
+1 Unit tests pass, I was able to duplicate scenarios without naming collisions, and modifications copied over correctly.

src/mmw/js/src/modeling/models.js
caseypt: I'm working on #242 now and moved the logic from `Project` to here (it proved hard to test in it's current configuration because of the events), so now there is only once instance of this function.

src/mmw/js/src/modeling/models.js
caseypt: Thanks.

src/mmw/js/src/modeling/views.js
caseypt: Yeah, pretty handy.

src/mmw/js/src/modeling/models.js
caseypt: Unfortunately, because of some side-effect of nested models/collections not being fully supported, the collection is reset on initialization and the modifications are lost. I think if each model in `sourceMods` were being initialized as new models, we would be okay, but I'm not totally sure. Anyway, this was the simple work around. I'll remove the second argument and add a comment.

kdeloach: Would removing this code help? https://github.com/caseypt/model-my-watershed/blob/0bb927a9d4ce81f3d7e10ba4fc3768363bee5609/src/mmw/js/src/modeling/models.js#L199

src/mmw/js/src/modeling/models.js
caseypt: Nice work. Implemented and tested. Works well. Thanks.

src/mmw/js/src/modeling/models.js
caseypt: Oh interested. Didn't see that. It probably will. Let me see.

src/mmw/js/src/modeling/models.js
```diff
@@ -194,9 +194,9 @@ var ScenarioModel = Backbone.Model.extend({
     },
 
     initialize: function() {
-        // TODO: When fetching model or populating from a preloaded source
-        // this attribute should not be initialized as empty.
-        this.set('modifications', new ModificationsCollection());
+        if (this.get('modifications') === null || "[]") {
```
caseypt: @kdeloach Wrapping that line in a conditional fixed the issue. Everything else appears to be working as before.

kdeloach: Nice, now I'm just curious how we could get rid of checking for "[]". I'll look into Backbone best practices with recursive data structures.

caseypt: Actually, gotta make another change.

caseypt: On second thought, I'm going to file an issue. There's currently a problem with the way the modifications are saved.

caseypt: https://github.com/WikiWatershed/model-my-watershed/issues/278

caseypt: Updated so that we can duplicate scenarios and carry along the modifications. However, loading saved modifications is currently not working: #278. Updated so that the code doesn't throw an error.

Issue #167, opened 2015-05-15
Implement model input tools


caseypt:
Depends on #165

The model input tools provide input parameters to the model, such as the amount of rainfall.

There needs to be a two-way relationship between these tools and the scenarios. For example, when setting the rainfall value to 2", this value gets pushed to the scenario. Then, we reloading this project, the slider should be set to 2" based on the stored value in the scenario.

Ex (though I think this is being converted to a simple slider).

![selection_018](https://cloud.githubusercontent.com/assets/1042475/7656009/baae08da-faf5-11e4-89ea-614e56f34cdf.png)

mmcfarland:
Correct, replace the drop down with a slider.  We'll have to see what range and increments are appropriate, so for now, lets just do .05 to 6 with half inch tics.

mmcfarland:
Reminder that this input tool is associated with the tr55 model package.  The UI description will be client side only as part of that backbone model, but the value should be stored as `inputs` - which doesn't exist yet - on the scenario django model as serialized json, similar to `modifications`.

Issue #165, opened 2015-05-15
Define model packages


caseypt:
Model packages contain the following:

- A name (ex. TR-55)
- A list of modification tools (ex. land cover drawing tool, conservation practice drawing tool)
- A list of input tools (ex. rainfall parameter selector)

@mmcfarland and I briefly discussed this. There will likely be nothing stored in the database regarding the package.

Consider a JSON structure that then can be used to construct UI tool components with Backbone models and views for the modeling page.

mmcfarland:
This has been implemented through various commits linked to here.  There may ultimately be more work, but closing in favor of new issues when necessary.

PR #273, opened 2015-06-03
Implement model input tools

kdeloach:
Serializes input UI values to client-side models.

To test:
1. Add scenario and change precipitation value
2. Create another scenario and add modifications / change precipitation value
3. Switching between scenarios should show correct user selected values

Connects #167

### DIFF
--- a/src/mmw/js/src/modeling/controls.js
+++ b/src/mmw/js/src/modeling/controls.js
@@ -1,0 +1,124 @@
+"use strict";
+
+var _ = require('lodash'),
+    Backbone = require('../../shim/backbone'),
+    Marionette = require('../../shim/backbone.marionette'),
+    App = require('../app'),
+    drawUtils = require('../draw/utils'),
+    models = require('./models'),
+    landCoverTmpl = require('./templates/controls/landCover.html'),
+    conservationPracticeTmpl = require('./templates/controls/conservationPractice.html'),
+    precipitationTmpl = require('./templates/controls/precipitation.html');
+
+// Simulation input controls base class.
+var ControlView = Marionette.ItemView.extend({
+    model: models.ModelPackageControlModel,
+    className: 'inline',
+
+    initialize: function(options) {
+        this.mergeOptions(options, [
+            'modificationModel',
+            'addModification',
+            'addOrReplaceModification'
+        ]);
+    },
+
+    getControlName: function() {
+        throw 'Not implemented';
+    }
+});
+
+// Drawing control base class.
+var DrawControlView = ControlView.extend({
+    ui: {
+        drawControl: '[data-value]'
+    },
+
+    events: {
+        'click @ui.drawControl': 'startDrawing'
+    },
+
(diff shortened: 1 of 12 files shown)


kdeloach:
Rebased onto develop.

kdeloach:
The controls are loaded dynamically now and may need design assistance.
image

src/mmw/js/src/modeling/controls.js
```diff
@@ -0,0 +1,124 @@
...
+
+var LandCoverView = DrawControlView.extend({
+    template: landCoverTmpl,
+
+    getControlName: function() {
+        return 'landcover';
+    }
+});
+
+var ConservationPractiveView = DrawControlView.extend({
```
rajadain: Typo, v -> c, should be `ConservationPracticeView`

kdeloach:
@caseypt  Just noticed my changes don't strictly follow #165. I designed it such that there is no distinction between drawing/input modifications (other than 1 has a shape, and the other doesn't). This feature may need to be reworked.

caseypt:
#165 was just a rough sketch, so it may not matter. The only difference between the two was that that inputs have a two-way relationship with scenarios, and modifications don't.

kdeloach:
Rebased.

lliss:
Missing a require statement for jQuery.

lliss:
+1. Tested.

There's a lot of nesting of models and collections which makes things a little tricky to follow. But we've discussed that and I don't think we'd benefit from trying to do it another way.

caseypt:
I'm not able to save models because `model_package` is invalid:

`{"model_package":["\"{u'name': u'tr-55', u'controls': [{u'name': u'landcover'}, {u'name': u'conservation_practice'}, {u'name': u'precipitation'}]}\" is not a valid choice."]}`

Under the current configuration, the backend only needs to know that it is `tr-55`.

caseypt:
When loading saved projects, I get the following error:

selection_045

This is the line:

selection_046

caseypt:
These errors are related. When `project.model_package`, when it comes back from the server is just a string equal to `tr-55`, which is why the error noted above occurs.

caseypt:
These two tests are failing:

selection_047

Rebuilt the test bundle to make sure.

caseypt:
@kdeloach There are some instructions for using the inspector to save a project in this PR comment: https://github.com/WikiWatershed/model-my-watershed/pull/230

kdeloach:
Thanks, although I tested that saving works in #230, I have not made an effort to test things that have yet to be wired up to the UI. I do not really like using Marionette Inspector because of the double refresh problem.

kdeloach:
Rebased.

caseypt:
@kdeloach Is this ready for another run through?

kdeloach:
Tests should be working now. Saving may be fixed by #278.

lliss:
All right. Tried it out. All tests passing again.

PR #279, opened 2015-06-04
Prevent multiply-stacked outline layers.

### LINKED ISSUE #253: Select by Boundary drop down adds layer each time item is selected.
It should instead turn off and on the layer.

To reproduce:

- Select "Congressional Districts" from the "Select by Boundary" drop-down.
- Select it again. Note the increase in saturation of the district outlines.

jamesmcclain:
This patch prevents there ever being more than one outline layer visible on the screen at once.

To test: click on `Select by Boundary` and choose `Congressional Districts`.  Do the same again.  Previously, the saturation on the outlines would have increased, now it does not.

Fixes #253

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -172,13 +172,15 @@ var PlaceMarkerView = Marionette.ItemView.extend({
 });
 
 function changeOutlineLayer(endpoint, tableId, model) {
-    var map = App.getLeafletMap();
+    var map = App.getLeafletMap(),
+        ofg = model.get('outlineFeatureGroup');
 
+    // Go about the business of adding the ouline and UTFgrid layers.
     if (endpoint && tableId != undefined) {
         var ol = new L.TileLayer(endpoint + '.png'),
-            grid = new L.UtfGrid(endpoint + '.grid.json?callback={cb}', { resolution: 4 }),
-            ofg = model.get('outlineFeatureGroup');
+            grid = new L.UtfGrid(endpoint + '.grid.json?callback={cb}', { resolution: 4 });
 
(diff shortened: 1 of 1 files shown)


src/mmw/js/src/draw/views.js
```diff
@@ -172,12 +172,18 @@ var PlaceMarkerView = Marionette.ItemView.extend({
 });
 
 function changeOutlineLayer(endpoint, tableId, model) {
-    var map = App.getLeafletMap();
+    var map = App.getLeafletMap(),
+        ofg = model.get('outlineFeatureGroup');
 
-    if (endpoint && tableId != undefined) {
+    // If the outlines are already present, there is nothing more to do.
+    if (ofg.getLayers().length !== 0) {
```
caseypt: Should we also check that the selected layer is the layer that is already on the map? (I know we only have one layer at the moment).

jamesmcclain: I do not have a firm understanding of the intended workflow, some possible variations that I can think of would preclude that as problem.  The other alternative is to simply clear the `outlineFeatureGroup` and re-add.  That actually probably makes more sense.

jamesmcclain: What I meant by "firm understanding of the workflow" is the following: we have an acknowledged bug wherein the shadow mask prevents click events from reaching the UTF layer.  This is supposed to eventually go away (I understand) from changes to the workflow.

caseypt: Yeah, that might be a safer option for now until we get some more layers and have a better idea of how that control will work.

caseypt:
+1 Tested.

jamesmcclain:
Many thanks.

Issue #169, opened 2015-05-15
Implement project drop-down functionality


caseypt:
May want to start with some of the basic stuff first, then move on to *My Projects* and *New Project*.

![selection_020](https://cloud.githubusercontent.com/assets/1042475/7656730/705386d4-fafa-11e4-9848-61e9ee190fa7.png)

mmcfarland:
Let's just implement Rename and Delete for this issue.  Share and Print are not ready yet, and Add Tags will be its own thing.

PR #284, opened 2015-06-04
Implement Project UI dropdown

mmcfarland:
* Add an input model to be use for renaming projects
* Consolidate common methods for modal views
* Implement project operations: print, share, rename and delete

Connects #169 

Testing:
* Try renaming, sharing, printing and deleting Projects
* Both logged in and logged out will have different behavior
* Saved Models will actually be deleted.
* Can log in via the share modal

### DIFF
--- a/src/mmw/js/src/core/templates/shareModal.html
+++ b/src/mmw/js/src/core/templates/shareModal.html
@@ -1,19 +1,28 @@
 <div class="modal-dialog modal-dialog-sm">
     <div class="modal-content">
         <div class="modal-header">
-            <h2>Share Project</h2>
+            <h2>Share {{ text }}</h2>
         </div>
         <div class="modal-body">
-            <div class="custom-input-group">
-              <input type="text" required value="{{ url }}">
-              <span class="highlight"></span>
-              <span class="bar"></span>
-              <label>URL</label>
-            </div>
+            {% if guest %}
+                <p>Sorry, only saved {{ text }}s are able to be shared.  Please log in to share your work.</p>
+            {% else %}
+                <div class="custom-input-group">
+                  <input type="text" required value="{{ url }}">
+                  <span class="highlight"></span>
+                  <span class="bar"></span>
+                  <label>URL</label>
+                </div>
+            {% endif %}
         </div>
         <div class="modal-footer">
-            <button type="button" class="btn btn-md btn-default" data-dismiss="modal">Cancel</button>
-            <button type="button" class="btn btn-md btn-default" data-dismiss="modal">Copy Link</button>
+            {% if guest %}
+                <button type="button" class="btn btn-md btn-default" data-dismiss="modal">Close</button>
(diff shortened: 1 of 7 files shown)


mmcfarland:
The share modal doesn't actually copy the url yet, and `enter` key doesn't close the rename box, but I've got an issue for one and some general utility code that i'm working on for the other.

src/mmw/js/src/core/templates/shareModal.html
```diff
@@ -1,19 +1,28 @@
...
         </div>
         <div class="modal-body">
-            <div class="custom-input-group">
-              <input type="text" required value="{{ url }}">
-              <span class="highlight"></span>
-              <span class="bar"></span>
-              <label>URL</label>
-            </div>
+            {% if guest %}
+                <p>Sorry, only saved {{ text }}s are able to be shared.  Please log in to share your work.</p>
```
caseypt: Verified that this works, though we should make a note an issue that we should prompt a save before sharing once they are logged in.

src/mmw/js/src/modeling/views.js
```diff
@@ -66,10 +67,83 @@ var ModelingHeaderView = Marionette.LayoutView.extend({
 // The drop down containing the project name
 // and projects options drop down.
 var ProjectMenuView = Marionette.ItemView.extend({
-    model: models.ProjectModel,
+    ui: {
+        rename: '#rename-project',
+        share: '#share-project',
+        remove: '#delete-project',
+        print: '#print-project'
```
caseypt: I've started to use `data-action="rename"` for these types of elements. Not sure if that's the best way to go, but just FYI.

src/mmw/js/src/core/templates/shareModal.html
mmcfarland: Yeah, I think login in from the project page should always trigger an initial save immediately.

caseypt:
The rename modal currently allows me to save empty strings for the project name.

src/mmw/js/src/modeling/views.js
mmcfarland: I like IDs when we can use them because they're the fastest selector, I think attribute selectors are one of the slowest.  What's the case for a data attribute?

caseypt:
It doesn't look like it'll help us out at the moment, but we may be able to use Behaviors at some point for the modals: http://marionettejs.com/docs/v2.4.1/marionette.behavior.html http://marionettejs.com/docs/v2.4.1/marionette.behaviors.html

src/mmw/js/src/modeling/views.js
caseypt: I think it can be nice to see in the DOM which elements are bound to events and what they do, which is not always easy. Though in this case it may be because of the ID names.

caseypt:
+1 besides the empty string comment. I was able to delete, rename, share and print.

mmcfarland:
Updated with 1819b04e81705d5063eca71ef82a15704c680796 for Project Name validation.

caseypt:
Tested. Fixed.

PR #288, opened 2015-06-04
Add support for creating a VPC and Route 53 private hosted zone

hectcastro:
Make use of `majorkirby` to thread together the launching of an Amazon VPC, along with an associated Route 53 private hosted zone, with Boto.

For more details instructions, see the `README` at the root of the `deployment` subdirectory.

Connects #26. Also, tagging @kdeloach and @lliss.

Issue #285, opened 2015-06-04
Set default NAT AMI to most recent version


hectcastro:
Currently, the most recent NAT AMI is hardcoded into the Troposphere template. If not set in the deployment configuration file, this should default to the most recent NAT AMI ID.

PR #304, opened 2015-06-05
Dynamically set NAT AMI if none is provided

hectcastro:
If no NAT AMI is provided, use the most recent 64bit NAT AMI with HVM support.

Connects #285 and builds on top of #288. Also, tagging @kdeloach and @lliss.

### DIFF
--- a/deployment/cfn/utils/cfn.py
+++ b/deployment/cfn/utils/cfn.py
@@ -51,6 +51,17 @@ def get_subnet_cidr_block():
         current += 1
 
 
+def get_recent_ami(aws_profile, ami_name, owner="self"):
+    conn = boto.connect_ec2(profile_name=aws_profile)
+    images = conn.get_all_images(owners=owner, filters={
+        'name': ami_name
+    })
+
+    return sorted(filter(lambda i: True if 'beta' not in i.name else False,
+                         images),
+                  key=lambda i: i.name, reverse=True)[0].id
+
+
 def read_file(file_name):
     """Reads an entire file and returns it as a string
 
--- a/deployment/cfn/vpc.py
+++ b/deployment/cfn/vpc.py
@@ -7,7 +7,12 @@ from troposphere import (
     Join
 )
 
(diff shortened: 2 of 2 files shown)


deployment/default.yml.example
```diff
@@ -0,0 +1,11 @@
...
+Region: 'us-east-1'
+StackType: 'Staging'
+KeyName: 'stg'
+IPAccess: '0.0.0.0/0'
+NATInstanceType: 't2.micro'
+NATInstanceAMI: ''
+NATAvailabilityZones: ['us-east-1a', 'us-east-1b']
+BastionHostInstanceType: 't2.micro'
+BastionHostAMI: ''
+PrivateHostedZoneName: 'mmw.internal'
```
kdeloach: Does this replace environmental variables?

deployment/cfn/stacks.py
```diff
@@ -0,0 +1,41 @@
...
+
+    Arguments
+    :param mmw_config_path: Path to the config file
+    :param profile: Config profile to read
+    """
+    mmw_config = ConfigParser.ConfigParser()
+    mmw_config.optionxform = str
+    mmw_config.read(mmw_config_path)
+
+    return {k: v.strip('"').strip("'") for k, v in mmw_config.items(profile)}
```
kdeloach: Consider adding some config validation here. `default.yml.example` may fall out of sync with reality over time.

kdeloach:
Is there a test plan? Or are we waiting for credentials and a staging environment to begin testing?

deployment/default.yml.example
hectcastro: Yep.

deployment/cfn/stacks.py
hectcastro: You mean like validating the attributes in this file against a schema?

kdeloach: Possibly, maybe majorkirby already invalid configs gracefully. Will test.

kdeloach:
+1 Tested

Issue #300, opened 2015-06-04
Mock delineate watershed 


mmcfarland:
Draw a shape for AoI on any of the Delineate watershed draw tool

PR #305, opened 2015-06-05
Mock delineate watershed

caseypt:
* When the user selects one of the delineate watershed methods,
generate a random polygon centered where the user clicked. The
shape occasionally looks vaguely like a watershed. This is
temporary until there is support for generating watershed boundaries.

**To test:**
- Select one of the delineate watershed methods from the dropdown.
- Click a point.
- A random polygon should be added to the map.

Connects to #300.

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -4,6 +4,8 @@ var _ = require('lodash'),
     $ = require('jquery'),
     L = require('leaflet'),
     Marionette = require('../../shim/backbone.marionette'),
+    turfRandom = require('turf-random'),
+    turfBuffer = require('turf-buffer'),
     router = require('../router').router,
     App = require('../app'),
     models = require('./models'),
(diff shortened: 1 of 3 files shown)


caseypt:
I followed the @kdeloach Method for Updating NPM Shrinkwrap :copyright:. Whenever I update it, it seems to switch the `from` attribute from a version reference to a tar file on npmjs.org. Not sure why.

src/mmw/js/src/draw/views.js
```diff
@@ -153,16 +155,28 @@ var PlaceMarkerView = Marionette.ItemView.extend({
...
-                addLayer(shape);
-                navigateToAnalyze();
-            }).fail(function() {
-                revertLayer();
-            });
+            // TODO: This is temporary until we have
+            // endpoints that can actually delienate
+            // watersheds.
+            var point = {
+                  "type": "Feature", "properties": {}, "geometry": {
```
lliss: Ah, yet another reason I dislike the single var statement. It messes up the nesting for objects like this.

src/mmw/package.json
```diff
@@ -40,6 +40,8 @@
     "sinon": "1.14.1",
     "testem": "0.5.15",
     "turf-area": "1.1.1",
+    "turf-buffer": "^1.0.4",
+    "turf-random": "^1.0.2",
```
lliss: Do we want to lock the version numbers? Seems like we've done that with all the other packages.

lliss:
This is a fun feature. Looks good +1. Tested it on all three possible ways to delineate an area.

lliss:
We should add a note or a task issue to remove turf random from the project when this becomes real.

lliss:
Not sure what's up with the shrikwrap file. I'm at the point of feeling we should stop fighting with it.

src/mmw/package.json
caseypt: Good call. Forgot about that.

caseypt:
Created an issue to remove turf-random: https://github.com/WikiWatershed/model-my-watershed/issues/307

caseypt:
Updated `package.json` with pinned versions for the turf libraries.

caseypt:
@azaveaci rebuild.

Issue #275, opened 2015-06-03
Model Output window toggles


rajadain:
The amount of controls and windows in the `/model` view leaves little room for the map, even on large screens, and makes drawing details difficult:

![image](https://cloud.githubusercontent.com/assets/1430060/7961242/44411d2a-09d4-11e5-82c1-7498a91b174a.png)

Consider ways to make the map bigger when users are drawing modifications, by minimizing other panels.

mmcfarland:
There is some workflow with the model output.  It's meant to start out minimized and icons on the output tabs indicate that models are running or complete.  The user can then toggle the pane up or down.  We've also got plans to consolidate the top app header with the project/scenario header.

rajadain:
Oh! Good to know. Should I close this issue or leave it open until those tasks are done?

mmcfarland:
Nope, it's good to leave, we can use it as a waffle to implement the feature when we are ready for it.

caseypt:
I think we should update the title `Model Output` to be `{{ active scenario }} Model Output`

caseypt:
Add an arrow

PR #309, opened 2015-06-05
Toggle Results Window

rajadain:
Adds a button to toggle model results window.

**Testing Instructions**

 * Draw a shape, click the green arrow button to go to the model view
 * On the right side of the tab-title row, you should see a toggle button
 * Clicking it should hide the contents, but keep the tab-title row visible
 * Clicking it again should restore the view to before

**Alternate Testing**

 * Try in different sized screens / orientations / browser resolutions

Connects #275

### DIFF
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -464,6 +464,14 @@ var ModelingResultsWindow = Marionette.LayoutView.extend({
         detailsRegion: '#modeling-details-region'
     },
 
+    ui: {
+        'toggle': '.tab-content-toggle'
+    },
+
+    events: {
+        'click @ui.toggle': 'toggleResultsWindow'
+    },
+
     onShow: function() {
         this.showDetailsRegion();
     },
@@ -494,19 +502,37 @@ var ModelingResultsWindow = Marionette.LayoutView.extend({
     animateIn: function() {
         var self = this;
 
-        this.$el.animate({ height: '55%' }, 400, function() {
+        this.$el.animate({ height: '55%', 'min-height': '300px' }, 200, function() {
             self.trigger('animateIn');
             App.map.set('halfSize', true);
+            $(self.ui.toggle.selector).blur()
+                .find('i')
+                    .removeClass('fa-angle-up')
+                    .addClass('fa-angle-down');
         });
     },
 
     animateOut: function() {
         var self = this;
 
-        this.$el.animate({ height: '0%' }, 100, function() {
+        // Change map to full size first so there isn't empty space when
+        // results window animates out
+        App.map.set('halfSize', false);
+        this.$el.animate({ height: '0%', 'min-height': '50px' }, 200, function() {
             self.trigger('animateOut');
-            App.map.set('halfSize', false);
+            $(self.ui.toggle.selector).blur()
+                .find('i')
+                    .removeClass('fa-angle-down')
+                    .addClass('fa-angle-up');
         });
+    },
+
+    toggleResultsWindow: function() {
+        if (this.$el.css('height') === '50px') {
+            this.animateIn();
+        } else {
+            this.animateOut();
+        }
     }
 });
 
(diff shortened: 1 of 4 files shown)


src/mmw/js/src/modeling/views.js
```diff
@@ -494,19 +502,37 @@ var ModelingResultsWindow = Marionette.LayoutView.extend({
...
             self.trigger('animateOut');
-            App.map.set('halfSize', false);
+            $(self.ui.toggle.selector).blur()
+                .find('i')
+                    .removeClass('fa-angle-down')
+                    .addClass('fa-angle-up');
         });
+    },
+
+    toggleResultsWindow: function() {
```
caseypt: Nice reuse of the existing animation functions.

src/mmw/js/src/modeling/views.js
```diff
@@ -494,19 +502,37 @@ var ModelingResultsWindow = Marionette.LayoutView.extend({
     animateIn: function() {
         var self = this;
 
-        this.$el.animate({ height: '55%' }, 400, function() {
+        this.$el.animate({ height: '55%', 'min-height': '300px' }, 200, function() {
```
rajadain: Setting both `animateIn` and `animateOut` to 200 gives a more consistent feel than the previous 400 for in and 100 for out.

src/mmw/js/src/modeling/views.js
```diff
@@ -494,19 +502,37 @@ var ModelingResultsWindow = Marionette.LayoutView.extend({
...
+                    .addClass('fa-angle-down');
         });
     },
 
     animateOut: function() {
         var self = this;
 
-        this.$el.animate({ height: '0%' }, 100, function() {
+        // Change map to full size first so there isn't empty space when
+        // results window animates out
```
caseypt: This makes sense. The tricky part is Leaflet wants you to call `map.invalidateSize` when the map size has changed. This is currently being done after `.half` is removed from the map container: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/views.js#L220.

We may need to call it here in the animation callback directly or through another method on `App.map`

rajadain: Ah, good to know. I will see if I can find a way.

caseypt:
You may want to show @ctaylo37 to get his input.

src/mmw/js/src/modeling/views.js
caseypt: It's in this section of the docs: http://leafletjs.com/reference.html#map-class

ctaylo37:
Looks good to me.

src/mmw/js/src/modeling/views.js
rajadain: `invalidateSize` does get called when we change the map container size in the next line: https://github.com/WikiWatershed/model-my-watershed/pull/309/files#diff-416520c38e4c168e3f58d3f5089772c4R520. Would calling it again after the other animation has ended be of any value?

src/mmw/js/src/modeling/views.js
```diff
@@ -494,19 +502,37 @@ var ModelingResultsWindow = Marionette.LayoutView.extend({
...
+        App.map.set('halfSize', false);
+        this.$el.animate({ height: '0%', 'min-height': '50px' }, 200, function() {
             self.trigger('animateOut');
-            App.map.set('halfSize', false);
+            $(self.ui.toggle.selector).blur()
```
rajadain: The `.blur()` is necessary otherwise button retains `:focus` after animation is complete and looks dark.

src/mmw/js/src/modeling/views.js
caseypt: As long as it's being called when the map is at it's full size, I think we are okay.

rajadain: It is, since the results window is on top of the map, not next to it, the window's size doesn't affect the maps size.

caseypt:
+1. Works well.

jamesmcclain:
Sorry for not seeing this earlier.  The code looks good, I will test it first thing on Monday morning.  It looks as though it as already passed muster, though.

jamesmcclain:
I just tested: the toggle button works as advertised and seems to always be visible whenever the title row is visible, regardless of size/shape of the browser window.  +1

Issue #264, opened 2015-06-02
Display actual AoI size


mmcfarland:
We don't yet have a good way of coming up with an accurate label for the AoI Analyze bar, so for now, let's have it read "Selected Area" instead of "Philadelphia":

```
Selected Area
[Actual Area size]
```
![screenshot from 2015-06-02 13 59 25](https://cloud.githubusercontent.com/assets/1014341/7942996/c73c60fe-092f-11e5-95dd-c8d6d9729c37.png)

PR #310, opened 2015-06-05
Calculate area of Area of Interest.

caseypt:
* Display the area of the Area of Interest in
the Analyze View header.
* Replace the default value of the analyze place
attribute with "Selected Area"
* Move area conversion method from
Modeling.ModificationModel to Core.GeoModel, and
subclass Modeling.ModificationModel and
Analyze.AnalyzeModel from Core.GeoModel.
* Switch to use metric system instead of English
system in area calculation and display per @chitchens.
* Move ModificationModel tests from Modeling
tests to Core tests.
* Add test to Modeling tests to ensure
Modeling.ModificationModel is inheriting
properly from Core.GeoModel.

**To test**:

- Draw an area to analyze.
- Note that, in the analyze window header, the area is now in sq m or sq km and that the name is "Selected Area" and not Philadelphia.
- Try the other AoI methods and verify that the analyze header changes when a new area is selected.
- Run the tests.

Connects to #264

### DIFF
--- a/src/mmw/js/src/analyze/controllers.js
+++ b/src/mmw/js/src/analyze/controllers.js
@@ -8,10 +8,6 @@ var $ = require('jquery'),
     views = require('./views'),
     models = require('./models');
 
-var createTaskModel = _.memoize(function(aoi) {
-    return new models.AnalyzeTaskModel();
-});
-
 var AnalyzeController = {
     analyzePrepare: function() {
         if (!App.map.get('areaOfInterest')) {
@@ -21,12 +17,11 @@ var AnalyzeController = {
     },
 
     analyze: function() {
-        var aoi = JSON.stringify(App.map.get('areaOfInterest'));
-
-        var analyzeWindow = new views.AnalyzeWindow({
-            id: 'analyze-output-wrapper',
-            model: createTaskModel(aoi)
-        });
+        var aoi = JSON.stringify(App.map.get('areaOfInterest')),
+            analyzeWindow = new views.AnalyzeWindow({
+                id: 'analyze-output-wrapper',
+                model: createTaskModel(aoi)
+            });
 
         App.rootView.footerRegion.show(analyzeWindow);
     },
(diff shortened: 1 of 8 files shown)


rajadain:
Consider using standard SI symbols **m&sup2;** and **km&sup2;** instead of **sq. m.** and **sq. km.** [Wikipedia](http://en.wikipedia.org/wiki/Square_kilometre) [NIST](http://physics.nist.gov/cuu/Units/units.html)

While I would favor using the HTML escape code `&sup2;`, Wikipedia and other sites use the more explicit `<sup>2</sup>` since some older default fonts might not have the [Superscript Two](http://graphemica.com/%C2%B2) character.

rajadain:
After I draw an area, the draw tools don't go away:

image

And if I draw a new area using these tools, the calculated area is not updated.

rajadain:
Is the terminology we're using under the hood, **Area of Interest**, a common GIS term? If so, we should consider using that instead of **Selected Area**. /cc @mmcfarland

mmcfarland:
"Area of Interest" was just an expedient term we came up with discussing design elements initially, not sure if it is casual enough to use here, I like the generic-ness of "Selected Area" until we can figure out a good way to compute a place name (though I suspect we may never).

caseypt:
@rajadain I'm not seeing this behavior (stale area and keeping the tools on). I'll have to check it out on your computer when I get in.

caseypt:
@rajadain Good call on the symbols. Rebased and updated to use them.

caseypt:
Fixing the tests.

caseypt:
@azaveaci rebuild

rajadain:
+1 Works perfectly now. Did the following testing:

 * Tested range of values, from very small 43 m&sup2; to very large 596,615,995 km&sup2; and it calculates correctly
 * The behavior of draw controls being visible in analyze view now no longer happens. It may have been an artifact of faulty testing on my part, either faulty bundling or accidental caching of scripts. It is not there anymore.
 * Going back from analyze to draw view and then drawing a new area calculates area correctly
 * Going forward from analyze to model view and then going back shows the calculated area correctly
 * Units are now shown in more standard form
 * Tested using manually drawn area and congressional districts
 * Tested on Chrome and Firefox

rajadain:
I was actually able to reproduce the accidentally visible draw tools on both your branch and on `develop`, so it is not an artifact of the work done here. Created #314 for it.

mmcfarland:
+1 Tested and looks good.

Issue #299, opened 2015-06-04
Bars are not varying in analyze mode


lliss:
All bars seem to be the same width.
Reported in Chrome but not firefox.

PR #311, opened 2015-06-05
Fix bar chart in Chrome

rajadain:
Width is always specified in SVG. Firefox prefers SVG width over CSS, so the chart looks correct there, but Chrome prefers CSS width thus making all the bars in the chart have the same 9% width.

If there is history to the 9% number I would love to be filled in. It seems to have always been there since the file was originally added to the repository.

**Testing Notes**

Open the `/analyze` view in different browsers and make sure the bar charts look reasonable.

Connects #299

### DIFF
--- a/src/mmw/sass/components/_charts.scss
+++ b/src/mmw/sass/components/_charts.scss
@@ -16,7 +16,6 @@
   vertical-align: bottom;
 
   .bar{
-    width: 9%;
     margin: 0 2%;
     display: inline-block;
     fill: $brand-primary;


rajadain:
Connects #299

caseypt:
I think the 9% number is from the original mock ups, when it was a vertical bar chart that was rendered statically in HTML. Now that it's horizontal, the 9% width is restricting what is now the bar height. 

Just FYI, I haven't checked this out yet, but the bar charts were almost always being rendered with varying width in Chrome for me. Current version is 41.0.2272.76 Ubuntu 14.04 (64-bit). Not sure why that is.

rajadain:
I'm running on Version 43.0.2357.81 (64-bit) on Mac OS X Yosemite, and this fixes the bar widths on my machine. Seems like this was a known issue in WebKit and only fixed last year: http://stackoverflow.com/a/12631326/2053314. This may have been a regression. Here's a bug created 4 days ago about this exact issue: https://code.google.com/p/chromium/issues/detail?id=493681

rajadain:
Perhaps @lewfish could also take a look at this fix since he was also experiencing this bug.

lewfish:
Checking it out now.

lewfish:
This fix worked for me on Chrome Version 43.0.2357.81 (64-bit) on OS X.

mmcfarland:
+1 This fixed the persistent problem in Chrome (43.0.2357.65) for me on linux.

Issue #289, opened 2015-06-04
Get tests running again on Jenkins


caseypt:
@kdeloach suggested it may have something do with this PR: https://github.com/WikiWatershed/model-my-watershed/pull/250, which lines up with the first unstable build: http://civicci01.internal.azavea.com/job/model-my-watershed-develop/113/changes

PR #312, opened 2015-06-05
Fix tests not running on Jenkins

kdeloach:
The tests were failing because of a missing dependency related to
sinon. It appears that because of the way that sinon is written, its
internal utility functions are optimized away during minification.

Fixed this by removing minification of staging/production bundles.

Test that this works by running `VAGRANT_ENV=TEST vagrant provision
app` and verifiying that all the JS tests pass via
`./scripts/testem.sh`. You could also just inspect the PR builder
output on Jenkins to confirm that JS tests were run.

Connects #289

### DIFF
--- a/deployment/ansible/roles/model-my-watershed.app/tasks/static-files.yml
+++ b/deployment/ansible/roles/model-my-watershed.app/tasks/static-files.yml
@@ -16,16 +16,8 @@
     chdir: "{{ app_home }}"
   sudo: False
 
-- name: Create minified JS vendor bundle (staging/production)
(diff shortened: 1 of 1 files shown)


caseypt:
Strange problem.

+1. Tests are back: http://civicci01.internal.azavea.com/job/model-my-watershed-pull-requests/400/console

The pull-request job was missing the regexp check for `Tests 0`. It's now added so that the PR job will become unstable too if this problem comes back.

kdeloach:
Cool, thanks.

Issue #121, opened 2015-05-01
Move forward to Analyze


mmcfarland:
After selecting an AoI, you can hit "back" to go to the draw screen but we don't (and shouldn't) delete the shape.  We should however allow you to go "forward" to the analyze screen.  Maybe a little popup or tooltip on the shape?  Typically the analyze window is activated by completing a drawing.
![screenshot from 2015-05-01 15 10 43](https://cloud.githubusercontent.com/assets/1014341/7435797/068aa06a-f015-11e4-9dce-95c37d580731.png)

caseypt:
Now that we are selecting predrawn AoIs from a map layer, should analyze always be kicked off from an "Analyze" button in the feature popup?

Issue #294, opened 2015-06-04
Cannot cancel out of drawing


lliss:
On start of site, if you start drawing an area, you cannot cancel out to do something else.

lliss:
See https://github.com/WikiWatershed/model-my-watershed/issues/120

caseypt:
- Remove AoI
- Cancel drawing
- Remove districts

Issue #237, opened 2015-05-28
Use snake_case for all Backbone attribute names


caseypt:
For the purposes of matching up with DRF. This applies to Backbone models that aren't represented server-side as well.

See https://github.com/WikiWatershed/model-my-watershed/pull/230#discussion_r31261045

mmcfarland:
This guideline has been implemented and enforced through code reviews.

PR #315, opened 2015-06-08
Show area of interest bar in Draw

mmcfarland:
The area of interest header bar, previously only in Analyze now will
show in Draw, but only when an Area of Interest is defined on the map.
This typically should only happen when moving backwards from Analyze.

* Show calculated area of AoI
* Forward link goes back to Analyze
* No option to go further "back" from Draw
* Mask layer is removed when showing AoI on Draw, but applied when
  returning to analyze
* Selecting a new Boundary Layer will remove the current AoI

Connects #121


mmcfarland:
This PR is working but I'm going to wait until #264 is merged so I can rebase and move around some previously Analyze only views/styles to `core`.  Will notify when the changes are made.

rajadain:
Instead of the placeholder "10 acres", I see `[object Object]` instead:

image
but I imagine that is just waiting for #310 to be merged in.

mmcfarland:
@rajadain Correct.

rajadain:
For #309, I adjusted the animation so that when sliding down / animating out, the map is resized _before_ the animation instead of after, so we don't see a blank white space. This hardly happens in this case since the slide is so fast, but we might consider that.

Also in #309, the animation speed was adjusted to be 200ms both ways, rather than 400ms in and 100ms out. If that feels alright we can consider applying that here as well.

rajadain:
+1 works very well. After going back to draw I can finally go forward to analyze and back and forth. Can't go further back from draw. Went all the way to model and came all the way back. Drawing new area automatically jumps me to analyze. Mask is only applied from analyze onwards, and I see the highlighted AoI in draw. Also, I can now select Congressional Districts and click outside the previous AoI which couldn't be done before because of the mask.

I wonder if while drawing a new AoI, what should happen to the old area value in the black bar? Should it be hidden until drawing is complete, or visible and retain the old value?

mmcfarland:
Good question - I think I should hide the bar as the new shape is being drawn.  I'll work it in here or perhaps in a similar task in #294.

mmcfarland:
That is, once the current AoI is gone, you shouldn't be able to see a "forward button" at all.

mmcfarland:
Updated based on #264 being merged.  Addressed window animation points from @rajadain, but am going to address the footer existing while redrawing in a forthcoming PR related to #294.

src/mmw/js/src/core/models.js
```diff
@@ -10,11 +11,16 @@ var MapModel = Backbone.Model.extend({
         lat: 0,
         lng: 0,
         zoom: 0,
-        // TODO: Delete (already exists on project model)
         areaOfInterest: null,           // GeoJSON
```
kdeloach: I'm still planning to eliminate this.

src/mmw/js/src/core/models.js
```diff
@@ -113,8 +119,14 @@ var GeoModel = Backbone.Model.extend({
     }
 });
 
+var AreaOfInterestModel = GeoModel.extend({ defaults: _.extend({
```
kdeloach: Consider adding a line break here.

src/mmw/js/src/analyze/views.js
```diff
@@ -60,9 +61,12 @@ var AnalyzeWindow = Marionette.LayoutView.extend({
     },
 
     showHeaderRegion: function() {
-        this.headerRegion.show(new HeaderView({
-            model: new models.AnalyzeModel({
-                shape: this.model.get('area_of_interest')
+        this.headerRegion.show(new coreViews.AreaOfInterestView({
+            model: new coreModels.AreaOfInterestModel({
+                shape: this.model.get('area_of_interest'),
+                goBack: true,
```
kdeloach: Consider renaming to `canGoBack`?

src/mmw/js/src/draw/views.js
```diff
@@ -88,6 +89,7 @@ var SelectAreaView = Marionette.ItemView.extend({
             endpoint = $el.data('endpoint'),
             tableId = $el.data('tableid');
 
+        App.map.set('areaOfInterest', null);
```
kdeloach: Ideally, this view would follow the design pattern of the other drawing tools, so that the revert layer logic is similarly self-contained.

src/mmw/js/src/draw/views.js
```diff
@@ -9,6 +9,7 @@ var _ = require('lodash'),
     router = require('../router').router,
     App = require('../app'),
     models = require('./models'),
+    filters = require('../filters'),
```
kdeloach: I don't think this import was used in this file.

kdeloach:
+1

src/mmw/js/src/draw/views.js
mmcfarland: Will fix.

src/mmw/js/src/analyze/views.js
mmcfarland: Done.

mmcfarland:
I should have put a note that the nunjucks Environment object behaves strangely.  @caseypt and I tried to track it down, but the essential behavior is that, depending on what view file is called first, the `filters.js`  module would need to be required before template filters could be available.  Having it required in `main` seemed to have no effect, so the safest thing I could do was require it in all the view files so that it would be available no matter which URL was used to access a view first.

mmcfarland:
Updated with cleanup and code to hide the draw bar when the Area of Interest is removed.

kdeloach:
I think we should configure the nunjucks environment in `core/setup.js` and perhaps require template filters from there. Could be done in a future PR.

mmcfarland:
I'm not opposed to trying, but doing the work in `main.js` didn't seem to do the trick.  I'll create an issue so we don't lose sight of it though.

caseypt:
Looking at this now.

caseypt:
When I go back to draw after being forwarded to analyze, the mask is removed and the shape outline is restored, but the analyze header doesn't show up. 

@mmcfarland Could you try rebaseing this on `develop` and see if you get the same behavior?

mmcfarland:
Yes, but try to kill your bundle or re-save `_base.scss` to make sure it's building your css, happens to me.

caseypt:
Thanks. Resaving `_base.scss` did it.

src/mmw/js/src/analyze/views.js
```diff
@@ -60,9 +61,13 @@ var AnalyzeWindow = Marionette.LayoutView.extend({
...
 
     showHeaderRegion: function() {
-        this.headerRegion.show(new HeaderView({
-            model: new models.AnalyzeModel({
-                shape: this.model.get('area_of_interest')
+        this.headerRegion.show(new coreViews.AreaOfInterestView({
+            App: App,
+            model: new coreModels.AreaOfInterestModel({
+                shape: this.model.get('area_of_interest'),
+                canGoBack: true,
```
caseypt: Do we want to use snake_case for only those models that are persisted on the server or all Backbone models?

kdeloach:
I noticed the node-sass watch has been acting up lately (not rebuilding changed files). Created a PR to reduce the number of watched files and hopefully that helps.

src/mmw/js/src/analyze/views.js
mmcfarland: Good question, probably confusing to use both.  Open to suggestions.

caseypt: Looks like we decided to use it for all models: https://github.com/WikiWatershed/model-my-watershed/issues/237. The comment thread is gone, unfortunately. I'd still vote for that.

rajadain: I think we could use snake_case for all model attributes. They are almost never accessed directly in JavaScript, unlike properties and methods like `this.headerRegion.show`, and almost always accessed as string arguments to Backbone.Model.get. Thus using snake_case for all model attributes would be predictable without breaking JavaScript convention.

kdeloach: +1 snake case

mmcfarland:
snake_case_ified.

caseypt:
+1. Tested.

PR #313, opened 2015-06-05
Add support for creating application server AMIs

hectcastro:
Packer is used to drive the EBS backed AMI generation process, making use of our existing Ansible roles and playbooks. The same credential setup process for launching stacks is reused for AMI creation.

Connects #293 and builds on top of #288. Also, tagging @kdeloach and @lliss.

Issue #292, opened 2015-06-04
Create tile server AMI


hectcastro:
Add a Packer definition to take the tile server playbook and emit an AMI.

PR #317, opened 2015-06-08
Add support for creating tile server AMIs

hectcastro:

Connects #292 and builds on top of #313. Also, tagging @kdeloach and @lliss.


kdeloach:
+1 I was able to generate a tiler AMI.

PR #318, opened 2015-06-08
Prototype REST service

jamesmcclain:
**Script to produce smaller GeoTiffs.**

A script has been added which produces compressed GeoTiffs of reduced color depth.  Useful for soil data, and possibly land-use, as well.  This script also makes the test soil rasters usable with the code (for testing).

Connects #546

**REST service.**

A spray-based REST service has been created.  The user POSTs a polygon stored as a GeoJSON file, and receives a count of land-use, soil-type pairs in the intersection of the query polygon and the test area.

Connects #545

The goals discussed in #544 can be done by sending a `GeometryCollection` instead of a `Polygon` or `MultiPolygon.  Those results can then be woven into a census of the type expected by the TR-55 code.

Connects #544

Big thanks to @echeipesh , @lossyrob , and @jwalgran .

**To Test**
   * Install spark (I used version 1.3.1)
   * Set the `SPARK_HOME` environment variable the name of the directory where spark is located
   * Set the `SPARK_VERSION` environment variable appropriately, in my case "1.3.1"
   * Apply this patch
   * Find the directory containing the three new tiff files that this patch creates; It should be `./src/geop/processing/src/test/data/` relative to the repository directory, set the MMW_DATA environment variable to the full name of that directory
   * Go to the `src/geop` directory (relative to the project root)
   * Type `sbt compile test` to run tests
   * Type `sbt 'project services' compile assembly` to combile the REST service code
   * Type `$SPARK_HOME/bin/spark-submit ./services/target/scala-2.10/mmw-services-assembly-0.0.1.jar` to run the service
   * Try to access the service on localhost:8107.  Anything other than POST should give you a 400.  Posting anything other than a valid GeoJSON polygon should give you a 400.
   * Sending a valid GeoJSON polygon should cause the service to return a JSON list of lists containing land-use X soil-type tuples and their counts (in the intersection of the query polygon and the small part of Delaware that we have data for)

Some example polygons follow.

This polygon does not intersect with Delaware, you should see no output (but no error).
```javascript
{
    "type": "Feature",
    "geometry": {
        "type": "Polygon",
        "coordinates": [
            [
                [31.137728, 29.975284],
                [31.128097, 29.972484],
                [31.130860, 29.976007],
                [31.134046, 29.979194],
                [31.137728, 29.975284]
            ]
        ]
    }
}
```

This polygon is a subset of the covered area.  You should see some output.
```javascript
{
    "type": "Feature",
    "geometry": {
        "type": "Polygon",
        "coordinates": [
            [
                [-75.72533952081164, 39.76],
                [-75.72533952081164, 39.8024743318554],
                [-75.58521226772196, 39.77],
                [-75.58521226772196, 39.7342308109474],
                [-75.72533952081164, 39.76]
            ]
        ]
    }
}
```

Here it is again as a multipolygon:
```javascript
{
    "type": "Feature",
    "geometry": {
        "type": "MultiPolygon",
        "coordinates": [
            [
                [
                    [-75.72533952081164, 39.76],
                    [-75.72533952081164, 39.8024743318554],
                    [-75.58521226772196, 39.77],
                    [-75.58521226772196, 39.7342308109474],
                    [-75.72533952081164, 39.76]
                ]
            ]
        ]
    }
}
```
Here it is again as a `GeometryCollection`.  You will get the same output as before, but wrapped in another list.
```javascript
{
    "type": "GeometryCollection",
    "geometries": [{
        "type": "Polygon",
        "coordinates": [
            [
                [-75.72533952081164, 39.76],
                [-75.72533952081164, 39.8024743318554],
                [-75.58521226772196, 39.77],
                [-75.58521226772196, 39.7342308109474],
                [-75.72533952081164, 39.76]
            ]
        ]
    }]
}
```

This is an invalid file, you should get a 400.
```javascript
{
    "type": "Feature",
    "properties": {
        "name": 6
    },
    "geometry": {
        "type": "Polygon",
        "coordinates": [
            [
                [-75.72533952081164, 39.76],
                [-75.72533952081164, 39.8024743318554],
                [-75.58521226772196, 39.77],
                [-75.58521226772196, 39.7342308109474],
                [-75.72533952081164, 39.76]
            ]
        ]
    }
```

You can POST files to the service by typing something like this `wget localhost:8080 --post=file=foo.json -O -`


lossyrob:
It would be good to give the PR a descriptive title, and even mention some things in the body about what it includes. Also, smiley faces make for bad commit messages :)

jamesmcclain:
I should and will also add a test for a polygon which intersects the extent of the raster but is not a proper subset.

jamesmcclain:
Also, I removed the "test of the test" unit test because the comparisons of dimensions no longer made sense after I integrated integrated Eugene's changes.

src/geop/processing/src/main/scala/com/azavea/mmw/TR55.scala
```diff
@@ -0,0 +1,395 @@
...
+
+
+import org.apache.spark.SparkContext._
+import scala.collection.mutable
+import spire.syntax.cfor._
+
+object TR55 {
+
+  /** maps a 32-bit word to a soil type */
+  val wordToSoil = collection.immutable.HashMap(
```
lossyrob: This is a painfully long enough val that I would suggest pulling it out of this object. Perhaps have a `Constants` object, in it's own file, with an object that has this?

src/geop/processing/src/main/scala/com/azavea/mmw/TR55.scala
```diff
@@ -0,0 +1,395 @@
...
+                  val soilType = wordToSoil.get(soilTile.get(col,row))
+                  if (soilType != None) {
+                    val c = (nlcdTile.get(col, row), soilType)
+                    if(!localHistogram.contains(c)) { localHistogram(c) = 0 }
+                    localHistogram(c) += 1
+                  }
+                }
+              }
+            )
+          case unexpected =>
```
lossyrob: Should handle all the cases, without a catchall. You can use the `|` operator to catch a set of them:

```scala
clippedPolygon match {
  case PolygonResult(p) => //something
  case MultiPolygonResult(mp) | LineResult(l) => // something else
  case NoResult => // do nothing   
```

src/geop/processing/src/main/scala/com/azavea/mmw/TR55.scala
```diff
@@ -0,0 +1,395 @@
...
+  /** Returns a map of values to counts */
+  def combinations(nlcd: RasterRDD[SpatialKey], soil: RasterRDD[SpatialKey], polygon: Polygon): Map[(Int, Option[String]), Int] = {
+    val mapTransform = nlcd.metaData.mapTransform
+    val nlcdAndSoil = nlcd.join(soil)
+    val histogram = nlcdAndSoil
+      .map { case (key, (nlcdTile, soilTile)) =>
+        val extent = mapTransform(key) // transform spatial key to map extent
+        val rasterExtent = RasterExtent(extent, nlcdTile.cols, nlcdTile.rows) // transform extent to raster extent
+        val clippedPolygon = polygon & extent
+        val localHistogram = mutable.Map[(Int, Option[String]), Int]()
```
lossyrob: Seems like calling this a histogram is a misnomer, since we have a `Histogram` type and this is definitely not that.

jamesmcclain:
Those are all great suggestions, thank you.  I will implement them tomorrow morning.  I also plan to implement a (prototype) web service in the next few days which will appear here.

lossyrob:
I'm actually making some changes and am going to make a PR against your branch...I can go over the changes with you tomorrow, but stopped commenting when I started writing a bunch of code that I thought would be better as a PR then a github comment.

jamesmcclain:
Okay, that sound great.  See you this afternoon.

jamesmcclain:
@jwalgran was interested in seeing this.

mmcfarland:
General comments:
* bad json request should result in a `400` with a meaningful error (these are not externally available services) that can help in debugging
* Naming seems off, things like `val histogram ... =  NonHistorgram....` are confusing, I think we should stick with a standard nomenclature.  Also `UnattractiveButTemporary` indicates well that it's a temporary hack, but it's not descriptive enough to deduce it's role when I'm reading the code.  I'd prefer a real name with a TODO + Comment.
* The endpoints produce reasonable looking results for inputs, but I just want to verify that the tests are using `naiveHistogram` as a cell count/value test to be sure that we are getting the expected results.

Ultimately, I'd really appreciate some review time from @echeipesh or @lossyrob to really go over this approach.

mmcfarland:
Given the overhead of HTTP and loading the rasters, I think it makes sense (with regard to #546) to have an endpoint that accepts a list of polygons, for which histograms will be developed, with the assumption that they will all be subsets of a single extent, and there's not need to reload the rasters from the catalog.  In the case of a large amount of modification areas, this also keeps the number of HTTP requests to a constant 1.  What do you think?

Not sure it makes sense (from the client's perspective) to just return results in a list indexed against the order of the input, or to require an id and return an object keyed from that.

jamesmcclain:
> bad json request should result in a 400 with a meaningful error (these are not externally available 
> services) that can help in debugging

Fixed.

> Naming seems off, things like val histogram ... = NonHistorgram.... are confusing, I think we 
> should stick with a standard nomenclature. Also UnattractiveButTemporary indicates well that it's a 
> temporary hack, but it's not descriptive enough to deduce it's role when I'm reading the code. I'd 
> prefer a real name with a TODO + Comment.

Fixed.

> The endpoints produce reasonable looking results for inputs, but I just want to verify that the tests 
> are using naiveHistogram as a cell count/value test to be sure that we are getting the expected 
> results.

I certify that this is the case.

> Given the overhead of HTTP and loading the rasters, I think it makes sense (with regard to #546) 
> to have an endpoint that accepts a list of polygons, for which histograms will be developed, with 
> the assumption that they will all be subsets of a single extent, and there's not need to reload the 
> rasters from the catalog. In the case of a large amount of modification areas, this also keeps the 
> number of HTTP requests to a constant 1. What do you think?

The endpoint now accepts `Polygons`, `MultiPolygons`, and `GeometryCollections`.  The last can be used to send over multiple polygons or multipolygons in one go.

src/geop/sbt
```diff
@@ -0,0 +1,430 @@
...
+  residual_args=( "${residual_args[@]}" "$1" )
+}
+addResolver () {
+  addSbt "set resolvers in ThisBuild += $1"
+}
+addDebugger () {
+  addJava "-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=$1"
+}
+get_jvm_opts () {
+  # echo "${JAVA_OPTS:-$default_jvm_opts}"
```
lliss: Left over debug code?

jamesmcclain: Most probably.  I haven't done anything to `sbt` except remove that one comment.

mmcfarland:
Appears to be producing valid results for the test area.  I'm inclined to merge this and keep iterating with feedback from GT and @hectcastro, specifically regarding #6.

jamesmcclain:
Special dispensation to merge.

Issue #272, opened 2015-06-03
Replace precipitation dropdown with slider


kdeloach:
The precipitation dropdown should be replaced with a slider.

See https://github.com/WikiWatershed/model-my-watershed/issues/167 for more details.

PR #319, opened 2015-06-08
Replace precipitation dropdown with slider

kdeloach:
How to test:
1. Rebuild JS bundle
2. Verify slider control works on scenario tabs
3. Verify duplicating scenarios retains slider value

Connects #272


kdeloach:
I ran into some styling issues with the slider tooltip. May need help from design with this.
image

mmcfarland:
We're using the native input slider on the /simple app - consider using it here in favor of the bootstrap version.  I suspect it'll be an easy replacement code-wise.

kdeloach:
Will do.

kdeloach:
Replaced bootstrap slider with native slider.

lewfish:
Tested and works as expected, except that (I think) there should be a read-out of the precipitation level as there is in the wireframes.

lewfish:
Also, the 24 hour rain drop-down isn't there, although it is in the wireframes. I'm looking at 2015_03_10_Visual_Design_v2.pdf. Is there a more recent version somewhere?

mmcfarland:
There may not be a more recent version, but it is actually correct to have just the slider and not the dropdown for this model package.

kdeloach:
@ctaylo37 Here's what it looks like now. I tried but failed to style it according to the wireframe.
image

src/mmw/js/src/modeling/templates/controls/precipitation.html
```diff
@@ -1,5 +1,4 @@
-<select class="selectpicker" data-style="btn btn-sm btn-primary" data-size="3"> <!-- Weather Tools -->
-    <option value="24_hours">24 Hour</option>
-    <option value="historical">Historical</option>
-    <option value="custom">Custom</option>
-</select> <!-- End Weather Tools -->
+<label>
+    Rain Level
+    <input type="range" min="0" max="100" step="1" />
```
mmcfarland: That's a hell of a rainstorm.  Can we change the steps to be from, say, 1 to 10 but with fractional output?  Ie, 0.1, 0.2....10.0

mmcfarland:
Great idea with `setup.js`.

src/mmw/js/src/modeling/templates/controls/precipitation.html
kdeloach: Good idea.

kdeloach:
Updated slider min, max, and step.

mmcfarland:
+1, awesome that the step doesn't have to be an `int`.

lewfish:
I would prefer it if the rain read-out was there initially and updated as the value changed (but before the mouse was released). +1 otherwise.

kdeloach:
Good suggestion. Updated to display slider value while dragging.

kdeloach:
I also set the min slider value to 0 so rain fall filter could be disabled.

Issue #278, opened 2015-06-03
Scenario.modifications are serialized with unicode u


caseypt:
Scenario.modifications are sent to the server as an object. The field is a `TextField`, so DRF serializes the data to store it. As part of that process, it includes the unicode `u` before every string.

![selection_044](https://cloud.githubusercontent.com/assets/1042475/7966980/56f63560-09f5-11e5-8627-47c811ed44c1.png)

 When this value is retrieved from the server and is parsed on the client-side, you get the following error:

```
Uncaught SyntaxError: Unexpected token u
```

We need to modify the serialize to store and/or retrieve the value and not encode in unicode or not include the `u`.

PR #320, opened 2015-06-08
Custom serializer for modifications.

jamesmcclain:
The standard serializer prepends `u` to the field names when the object
is transformed into a string.  This patch prevents that from happening.

**To test:**
   * Follow the instructions given in https://github.com/WikiWatershed/model-my-watershed/pull/230, except now you must type `mn1.set('model_package', 'tr-55')` after typing `mn1.get('scenarios').at(1).set('name', 'TEST SCENARIO')` and before typing `mn1.saveAll()`.
   * In "developer tools", click on `network`, then `xhr`, then on the item on the left with `/api/modeling/projects` in it, then on the `response` tab.
   * Paste the response into `pro.jslint.com`.
   * Copy the value associated with the modifications key to the clipboard.  Type `JSON.parse("")` into the Javascript console and paste the modification information between the pair of double quotes.  When you press enter, you should get a valid Javascript object.

Fixes WikiWatershed/model-my-watershed#278

### DIFF
--- a/src/mmw/apps/modeling/serializers.py
+++ b/src/mmw/apps/modeling/serializers.py
@@ -8,6 +8,8 @@ from rest_framework_gis import serializers as gis_serializers
 
 from apps.modeling.models import Project, Scenario
 
+import json
+
 
 class UserSerializer(serializers.ModelSerializer):
     class Meta:
@@ -15,9 +17,20 @@ class UserSerializer(serializers.ModelSerializer):
         fields = ('id', 'username', 'email')
 
 
+class ModificationsField(serializers.BaseSerializer):
+
+    def to_representation(self, obj):
+        return json.loads(obj)
+
+    def to_internal_value(self, obj):
+        return json.dumps(obj)
+
+
 class ScenarioSerializer(serializers.ModelSerializer):
     class Meta:
         model = Scenario
+
+    modifications = ModificationsField()
 
 
 class ProjectSerializer(gis_serializers.GeoModelSerializer):


kdeloach:
Couple of notes regarding these instructions:
1. `model_package` is presently defined as an instance of a `ModelPackageModel` not a string (See https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/models.js#L51); That will probably change soon.
2. You can rename scenarios (scenario dropdown) and save the project (project dropdown) through the UI now.

jamesmcclain:
Okay, thank you for the update on `model package` and the UI.

src/mmw/apps/modeling/serializers.py
```diff
@@ -15,10 +17,21 @@ class Meta:
         fields = ('id', 'username', 'email')
 
 
+class ModificationsField(serializers.BaseSerializer):
+
+    def to_representation(self, obj):
+        return obj
```
mmcfarland: I think we should be returning a json object here, so it's deserialized appropriately for the client.  Consider returning `json.loads(obj)` here instead and making sure the client is not specifically expecting a string.

mmcfarland:
Tested that the modifications are now serialized to a JSON parsable string now.  Nice work, good to know we can easily override particular `ModelSerializer` fields when we need to.

jamesmcclain:
Thank you for the nice complement.  I have made the requested change.

kdeloach:
+1

caseypt:
Checking this out now.

caseypt:
+1. Looks good.

We still can't load saved scenarios without error, but that may be due to #331.

jamesmcclain:
Thank you.

Issue #229, opened 2015-05-27
Allow multiple "Untitled Projects"


mmcfarland:
When creating a new project, the default name should read "Untitled Project" a la Google Docs.  We don't force the user to rename a project, but a project must be unique within a user account.  We should make accommodations for multiple untitled projects, presumably by not storing that actual name, but otherwise indicating on the model that it is yet unnamed and having the displayed name be a UI consideration only.

caseypt:
Instead, remove the unique constraint from project name per user

PR #322, opened 2015-06-08
Allow multiple projects with same name

rajadain:
Removes unique constraint between user and project name.

**Testing Instructions**

 * Checkout branch and run all migrations.
 * Login as a user.
 * Select an area, click forward from analyze and land in the model view.
 * Use the Marionette.Inspector as instructed in #230, find the Project object, download it into the console (usually automatically declared as `mn1`) and run the following:

```javascript
mn1.set('model_package', 'tr-55');
```

> That last step is necessary because the front-end currently stores `model_package` as a complex object, but the back-end expects a simple string, so we manually fix it.

 * Save the project, using either the UI or just `mn1.saveAll();` in the console. Refresh the page to make sure it saved.
 * Create a second project and follow the same steps, make sure to save it as the exact same name as the first one.

Connects #229

### DIFF
--- a/src/mmw/apps/modeling/migrations/0005_auto_20150608_1502.py
+++ b/src/mmw/apps/modeling/migrations/0005_auto_20150608_1502.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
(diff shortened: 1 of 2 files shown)


mmcfarland:
+1 Saved multiple projects.

kdeloach:
+1 Tested

Issue #291, opened 2015-06-04
Create worker AMI


hectcastro:
Add a Packer definition to take the worker playbook and emit an AMI.

PR #323, opened 2015-06-08
Add support for creating worker AMIs

hectcastro:

Connects #291 and builds on top of #313. Also, tagging @kdeloach and @lliss.

### DIFF
--- a/deployment/packer/template.js
+++ b/deployment/packer/template.js
@@ -27,6 +27,27 @@
                 "Environment": "{{user `stack_type`}}"
             },
             "associate_public_ip_address": true
+        },
+        {
+            "name": "mmw-worker",
+            "type": "amazon-ebs",
+            "region": "{{user `aws_region`}}",
+            "source_ami": "{{user `aws_ubuntu_ami`}}",
+            "instance_type": "m3.large",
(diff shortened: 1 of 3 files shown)


deployment/packer/template.js
```diff
@@ -0,0 +1,82 @@
...
+            "associate_public_ip_address": true
+        },
+        {
+            "name": "mmw-worker",
+            "type": "amazon-ebs",
+            "region": "{{user `aws_region`}}",
+            "source_ami": "{{user `aws_ubuntu_ami`}}",
+            "instance_type": "m3.large",
+            "ssh_username": "ubuntu",
+            "ami_name": "mmw-app-{{timestamp}}",
```
kdeloach: AMI name suffix should be worker.

kdeloach:
+1 I was able to create a worker AMI.

deployment/packer/template.js
hectcastro: Fixed in https://github.com/WikiWatershed/model-my-watershed/commit/2d179f02e9f514fc69591ad2d01424ecd0e74aa7.

lliss:
Ditto. +1

PR #324, opened 2015-06-08
Revert model_package back to a string

kdeloach:
This corrects an earlier mistake I made when I changed this property
from a string to an instance of a Backbone model. There's no point in
defining static controls on the project model because they cannot be modified
and should never be persisted.

To test:
1. Save a project.
2. The project and all its scenarios should save, and the address bar
should update to contain the saved project ID.
3. Reloading the page should display your saved data.


mmcfarland:
+1 Tested.

jamesmcclain:
I also tested, +1

Issue #282, opened 2015-06-04
Add zeroclipboard for share feature


mmcfarland:
Share has a copy url to clipboard which is not implemented.

https://github.com/zeroclipboard/zeroclipboard

mmcfarland:
Shares a point with #229

PR #325, opened 2015-06-08
Add ZeroClipboard for copying share URL

rajadain:
**Testing Instructions**

 * Login as a user
 * Draw a shape and move forward to model page
 * Save the project, so that you have a URL available akin to [http://localhost:8000/model/1/](http://localhost:8000/model/1/)
 * In the Project drop-down menu, click "Share"
 * Click "Copy Link"
 * Try pasting it anywhere. It should paste the URL of the project, namely `http://localhost:8000/model/1/`

**Notes**

The way ZeroClipboard works is by placing an invisible SWF file stretched to the size of the action button. Unfortunately this prevents our UI hover states, and thus while the cancel button does have a hover state and animation, the copy link button does not. I don't know how to solve this.

Connects #282

### DIFF
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -6,6 +6,7 @@ var $ = require('jquery'),
     router = require('../router.js').router,
     Marionette = require('../../shim/backbone.marionette'),
     TransitionRegion = require('../../shim/marionette.transition-region'),
+    ZeroClipboard = require('zeroclipboard'),
     drawUtils = require('../draw/utils'),
     headerTmpl = require('./templates/header.html'),
     modalConfirmTmpl = require('./templates/confirmModal.html'),
@@ -352,13 +353,23 @@ var ShareModal = BaseModal.extend({
     },
 
     events: {
-        'click @ui.signin': 'signIn',
-        'click @ui.copy': 'copyLinkToClipboard'
+        'click @ui.signin': 'signIn'
     },
 
-    copyLinkToClipboard: function() {
-        var url = this.ui.input.val();
-        // TODO https://github.com/WikiWatershed/model-my-watershed/issues/282
+    // Override to initialize ZeroClipboard
+    initialize: function() {
+        BaseModal.prototype.initialize.call(this);
+        this.zc = new ZeroClipboard();
+    },
+
+    // Override to attach ZeroClipboard to ui.copy button
+    onRender: function() {
+        var self = this;
+        this.$el.on('shown.bs.modal', function() {
+            self.zc.clip(self.$el.find(self.ui.copy.selector));
+        });
+
+        this.$el.modal('show');
     },
 
     signIn: function() {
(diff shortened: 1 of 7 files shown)


src/mmw/package.json
```diff
@@ -43,6 +43,7 @@
     "turf-buffer": "1.0.4",
     "turf-random": "1.0.2",
     "underscore": "1.8.3",
-    "watchify": "3.1.0"
+    "watchify": "3.1.0",
+    "zeroclipboard": "^2.2.0"
```
lliss: We've been pegging the versions.

src/mmw/js/src/core/views.js
```diff
@@ -347,18 +348,28 @@ var ShareModal = BaseModal.extend({
 
     ui: {
         'signin': '.signin',
-        'copy': '.copy',
+        'copy': '#copy-link-button',
```
lliss: Any reason this was made into an id rather than a class?

lliss:
You might be able to solve the hover issue by making an invisible button (or a very small one) and then having the click on our ui button just trigger a click on the other one. Not idea but it would solve the hover issue.

lliss:
I'd take another quick pass and see if you can solve the hover issue. Otherwise looks fine. I will test later today.

kdeloach:
ZeroClipboard has documentation regarding preserving hover effects (See https://github.com/zeroclipboard/zeroclipboard/blob/master/docs/instructions.md).

kdeloach:
+1

src/mmw/js/src/main.js
```diff
@@ -22,6 +22,12 @@ L.Icon.Default.imagePath = '/static/images/';
 var csrf = require('./csrf');
 $.ajaxSetup(csrf.jqueryAjaxSetupOptions);
 
+var ZeroClipboard = require('zeroclipboard');
+ZeroClipboard.config({
+    hoverClass: 'focus',
+    activeClass: 'active'
+});
+
```
rajadain: Will rebase on top of #319 after it has been merged and move this to new `setup.js`

src/mmw/package.json
rajadain: Ah good catch! Fixed now.

kdeloach:
Merged #319.

src/mmw/js/src/core/views.js
rajadain: I was experimenting with ZeroClipboard for a while and needed to individually target it. Since the template didn't have a `copy` class on the button, I went ahead and gave it an ID. But should've switched back before committing, to minimize number of changes required to introduce this functionality. Reverted now. Thanks!

rajadain:
Thanks for the pointer on ZeroClipboard hover state, added it in. Rebased on top of develop, pushed new consolidated commits.

lliss:
Looks good. Tested. Hover state works. +1

Issue #290, opened 2015-06-04
Create bastion/monitoring AMI


hectcastro:
Add a Packer definition to take the monitoring server playbook and emit an AMI.

PR #327, opened 2015-06-09
Add support for creating monitoring server AMIs

hectcastro:

In addition, this pull request includes a commit for determining the Git SHA and branch when not provided by environment variables.

Connects #290. Also, tagging @lliss and @kdeloach.


kdeloach:
+1

Issue #301, opened 2015-06-04
Respect privacy setting of project


caseypt:
- Add UI for setting project privacy
- Check privacy when loading project

PR #328, opened 2015-06-09
Implement project privacy access and UI

caseypt:
Implements the access rights for private projects and a UI modal to change the settings. See commit messages for details.

Here's the basic rule structure for accessing projects:

- Project owners can GET/PUT/DELETE owned public and private projects
- Logged in users can GET public projects.
- Logged in users can't PUT/DELETE public or private projects.
- Anonymous users can GET public projects.
- Anonymous users can't PUT/DELETE public or private projects.

**To Test:**
- Create a new project or load an existing one.
- Use the "Privacy" option in the project drop down menu to toggle the project privacy settings.
- Try accessing the project when logged in with a different user and when not being logged in. See the basic access rules above for expected behavior.
- Run the tests

Connects to #301.

### DIFF
--- a/src/mmw/apps/home/views.py
+++ b/src/mmw/apps/home/views.py
@@ -3,6 +3,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import division
 
+from django.http import Http404
 from django.contrib.auth.models import User
 from django.shortcuts import render_to_response, get_object_or_404
 from django.template.context_processors import csrf
@@ -33,16 +34,19 @@ def home_page(request):
 def model(request, proj_id=None):
     """
     If proj_id was specified, check that the user owns
-    the project. It not, return a 404. Otherwise, just
-    load the index template and the let the front-end
-    handle the route and request the project through
-    the API.
+    the project or if the project is public.
+    If not, return a 404. Otherwise, just load the index
+    template and the let the front-end handle the route
+    and request the project through the API.
     """
     csrf_token = {}
     csrf_token.update(csrf(request))
 
     if proj_id:
-        get_object_or_404(Project, user=request.user.id, id=proj_id)
+        project = get_object_or_404(Project, id=proj_id)
+
+        if project.user != request.user and project.is_private:
+            raise Http404
 
     return render_to_response('home/home.html', csrf_token)
 
--- a/src/mmw/apps/home/tests.py
+++ b/src/mmw/apps/home/tests.py
@@ -3,4 +3,108 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import division
 
-# Create your tests here.
+from django.contrib.auth.models import User
+
+from django.test import TestCase
+
+from rest_framework.test import APIClient
+
+
+class RouteAccessTestCase(TestCase):
+    def setUp(self):
+        self.c = APIClient()
+
+        self.test_user = User.objects.create_user(username='test',
+                                                  email='test@azavea.com',
+                                                  password='test')
+
+        self.another_user = User.objects.create_user(username='foo',
+                                                     email='test@azavea.com',
+                                                     password='bar')
+        self.project = {
+            "area_of_interest": ("MULTIPOLYGON (((30 20, 45 40, 10 40, 30 20))"
+                                 ",((15 5, 40 10, 10 20, 5 10, 15 5)))"),
+            "name": "My Project",
+            "model_package": "tr-55"
+        }
+
+        self.scenario = {
+            "name": "Current Conditions",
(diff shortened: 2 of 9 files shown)


src/mmw/apps/home/views.py
```diff
@@ -42,9 +44,12 @@ def model(request, proj_id=None):
     csrf_token.update(csrf(request))
 
     if proj_id:
-        get_object_or_404(Project, user=request.user.id, id=proj_id)
+        project = get_object_or_404(Project, id=proj_id)
 
-    return render_to_response('home/home.html', csrf_token)
+    if project.user.id == request.user.id or project.is_private is False:
```
rajadain: I'm new to Python and its idioms so I might be wrong, but would `or not project.is_private` be a better expression than `or project.is_private is False`?

rajadain:
I made a project as a logged in user and made it public. When I tried to access it using an incognito window in which I was not logged in (thus as an anonymous user), I just see the home page (the draw view), although with the URL of the model, and looking at the Network Panel in Developer Tools I see a 403 response from this URL: `http://localhost:8000/api/modeling/projects/1` with the following detail: `Authentication credentials were not provided.`

image

src/mmw/apps/home/views.py
kdeloach: In this case, I think `not project.is_private` is correct. If `is_private` was nullable, you would need to use the `is` operator (if you wanted to distinguish null from `False`).

kdeloach: Consider something like:

    if project.user != request.user and project.is_private:
        raise Http404

mmcfarland:
I'm able to create a public project that others (logged in or not) can view, but not modify and save.  Private projects are not visible to other users.  Tests run, and look like they have good coverage for such an important feature.

mmcfarland:
I don't know if this is where the UI for this project will ultimately live (I think it's fine for now) but it'd be good to see a visual indication of public/private without having to attempt to change it.  Maybe a fa-icon on the menu item, or simply the reversible text "Make Private/Make Public"?  Maybe an icon and tooltip on the project header bar?  Not sure.

kdeloach:
+1 Nice tests.

rajadain:
I too am able to see public projects anonymously now. Not sure what was going on before. Sorry about that.

rajadain:
If I go to [http://localhost:8000/model/](http://localhost:8000/model/) I get an Error 500.

src/mmw/apps/home/views.py
rajadain: Also note that `project` does not always exist here. If no `proj_id` is specified, as in [http://localhost:8000/model/](http://localhost:8000/model/), this line causes a crash.

caseypt: Thanks. I'll fix up and add some tests.

caseypt:
@rajadain Updated the view and added some tests. Should have fixed the problems you were seeing with `/model`.

@mmcfarland went with changing the menu option name based on the current value of `is_private` until @ctaylo37 has a chance to come up with something. Probably worth another run through with the UI.

src/mmw/apps/home/views.py
caseypt: Updated.

rajadain:
Making a project private without saving it first saves the project. Is this the correct behavior?

The "Make Private" button implies that the project was public by default. The earlier implementation gave me the impression that projects were private by default. Do we have a policy for this?

`/model` works correctly now, great idea adding a test for that.

caseypt:
@rajadain Good catch. Projects are private be default. Forgot to handle the new project case. Will fix.

Re: change the setting saving the project, I think I'm going to change it so that the menu item doesn't appear to the until project is saved initially. This way changing that property doesn't save the project.

mmcfarland:
The manual save processes is temporary, so I think it's fine that the project saves.  Once the auto save is set, we should just be able to set the property.  Projects should be private by default.

mmcfarland:
@caseypt can you make a block of 'only if saved' options in that list?  Share, delete, rename?

caseypt:
@mmcfarland Yeah, will do.

caseypt:
Added a new commit to limit the options that are available for unsaved projects.

rajadain:
Confirmed that I can only see "Rename", "Print" and "Save" as options before saving and they all work as expected (although I'm not sure how Print is supposed to work, since the preview doesn't have the map attached to it, but that is another issue). Also confirmed that projects are private by default.

+1 great work.

mmcfarland:
:+1:

caseypt:
Rebased on develop. @azaveaci rebuild.

PR #329, opened 2015-06-09
Add turf-buffer and turf-random to vendor bundle

kdeloach:
Should reduce the primary bundle size from 1.9MB to 600KB.


caseypt:
Thanks. Should have done that when I added them.

rajadain:
I should probably do this for #325 as well.

kdeloach:
Rebased.

caseypt:
+1. Verified that turf-random and turf-buffer are in the vendor bundle and not the main bundle.

PR #333, opened 2015-06-09
Add reset button to Draw

mmcfarland:
* Remove in-progress drawings
* Remove AoI
* Remove Boundary Layer

Connects #294

### DIFF
--- a/src/mmw/js/src/draw/templates/reset.html
+++ b/src/mmw/js/src/draw/templates/reset.html
@@ -1,0 +1,1 @@
+<button class='btn btn-md btn-active btn-search-tool'>Reset</button>
--- a/src/mmw/js/src/draw/templates/toolbar.html
+++ b/src/mmw/js/src/draw/templates/toolbar.html
@@ -1,3 +1,4 @@
 <div id="select-area-region"></div>
 <div id="draw-area-region"></div>
 <div id="place-marker-region"></div>
+<div id="reset-draw-region"></div>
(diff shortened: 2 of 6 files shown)


mmcfarland:
Would like to point out how small of a change this "trash can" reset feature was to all those PSI veterans out there.  Thanks @kdeloach for the ground work that made this trivial to implement.

kdeloach:
Nice.

rajadain:
+1 works very well. I was able to draw and reset areas of interest using all three methods. Was also able to use reset while drawing. Pressing <kbd>Escape</kbd> while drawing also works, I can't remember if it worked earlier or not. Resetting hides the AoI bar at the bottom. Drawing a new shape without resetting also hides the AoI bar. I also started with a saved shape, went all the way back to draw view and drew a new shape.

It may be nice to make the AoI bar transition away instead of just vanishing, but that can be a part of a UI polish effort in the future.

I'm not a big fan of adding `!important` declarations to CSS, but in this case I couldn't see a quick way around it. We may have to refactor the styling of all those buttons to accommodate this.

caseypt:
Consider adding a test for `ResetDrawView.reset`

mmcfarland:
@caseypt 46dc47b added with new tests.

mmcfarland:
@azavea-bot rebuild

caseypt:
Looks good. Tests pass for me.

Issue #331, opened 2015-06-09
Scenarios assume modifcations are present


mmcfarland:
`ToolBarTabConentView` assumes that a scenario has modifications which causes exceptions when switching between tabs under certain conditions (like switching between empty tabs on a saved project):

https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/views.js#L419-L428

kdeloach:
I would review the logic in the constructor (See https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/models.js#L139). Scenario modifications should never be null. If there are no modifications, this attribute should be an empty collection.

mmcfarland:
Interesting, in practice, I'm seeing the empty value be a serialized array:  `"[]"` which obviously doesn't implement `filter` and throws an exception.

PR #334, opened 2015-06-09
Draw correct map features from a loaded scenario

kdeloach:
The problem regarding modifications not being present (null) when
loading a saved project has been fixed by #320.

This fixes a related bug where map features would not be drawn for the correct
tab upon loading a saved scenario.

Connects #331

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -112,6 +112,16 @@ var ModificationModel = coreModels.GeoModel.extend({
 var ModificationsCollection = Backbone.Collection.extend({
     model: ModificationModel
 });
+// Static method to create an instance of this collection.
+// This lets us create a collection from an array of raw objects (web app)
+// or from an array of models (unit tests).
+ModificationsCollection.create = function(data) {
+    if (data instanceof ModificationsCollection) {
+        return data;
+    } else {
+        return new ModificationsCollection(data);
+    }
+};
 
 var ScenarioModel = Backbone.Model.extend({
     urlRoot: '/api/modeling/scenarios/',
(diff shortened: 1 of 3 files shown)


kdeloach:
I don't think projects saved before #320 will load correctly. I would clear out old data from the modeling project/scenario tables.

kdeloach:
Fixed tests.

kdeloach:
@azaveaci rebuild

rajadain:
+1 great work, this fixes it.

I'm not seeing any errors when loading projects anymore. I was seeing some after #320 got merged in but this fixed them. Was able to create new scenarios with modifications and able to save them as well. Also deleted existing scenarios so that none were left except for current conditions and reloaded the page to 0 errors.

caseypt:
Consider adding a test for this. It could be very similar to the tests that were fixed in this branch, but it would load a response from the server, like here: https://github.com/WikiWatershed/model-my-watershed/blob/8b0fc79cb372f4ddfc8e8d9acfd1c807cf89e65c/src/mmw/js/src/modeling/tests.js#L173

mmcfarland:
I'm getting an error on `modeling/views.js:420` that model is null.  I can duplicate this by

1.  Add scenario, save
1. Add scenario, save
1. Delete scenario from step 1
1. Receive error after hitting confirm.

kdeloach:
Last commit should fix that issue. 

Discovered a somewhat related bug where saving the project more than once has no effect. Creating an issue now.

rajadain:
Here's an issue I'm seeing:

 * Create a project, add a scenario, add another scenario, save project
 * Delete a scenario, delete another scenario, save project
 * Refresh page, the scenario's are still there

Not sure if this is related to the new issue you're creating @kdeloach or if I should make a separate one for this.

kdeloach:
Yes, that's the same issue I noticed. Created issue #341.

mmcfarland:
+1 Tested besides #341

Issue #200, opened 2015-05-21
Associate model output with scenario window


mmcfarland:
Wire up the workflow for a real TR-55 celery task which returns model output to generate the stacked bar chart.  The model run and output is associated with a single scenario.

PR #345, opened 2015-06-09
Associate model output with scenario window

lewfish:
Reprovision the worker VM. (Although running `sudo service celeryd restart` and `pip install -r src/mmw/requirements/base.txt` (to install tr55) on the worker VM will probably suffice.)

Make sure that the analyze view still loads table/chart as before.

In the model view, in Current Conditions, set the value of the precipitation slider. This should start the spinner icon on the Runoff tab. After polling ~6 times, the bar chart should display with the total level of runoff/et/inf matching the selected precipitation value, and the spinner should stop.

Try changing the precipitation value once, and then again about a second later. On the console, you should see that the first job was cancelled. If the user submits several jobs within the time it takes to run the model, there will be multiple jobs in the queue ahead of the final job, delaying the final job, potentially past the time limit set in the TaskModel. I think we should create a separate issue to deal with cancelling jobs on the backend.

Try making a new scenario, and change the precip values in both scenarios so they are polling concurrently. This should work. Adding modifications will fire off the job, but the modifications will have no effect because we don't have the Geotrellis task running on the backend yet.

Connects #200

### DIFF
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -5,10 +5,12 @@ from __future__ import absolute_import
 
 from celery import shared_task
 import logging
-import urllib2
 
 # TODO Remove this when stub task is deleted.
 import time
+
+from tr55.model import simulate_modifications, simulate_cell_day
+from tr55.tablelookup import lookup_ki
 
 logger = logging.getLogger(__name__)
 
@@ -120,34 +122,52 @@ def make_gt_service_call_task(model_input):
     """
     Call geotrellis and calculate the tile data for use in the TR55 model.
     """
-    # TODO call the real service
-    # It would be nicer to use HttpDispatchTask but there were import problems.
-    # In testing we could not make it work so we will probably need to default
-    # to urllib2.
-    # return HttpDispatchTask(url="http://gtservice.internal/endpoint",
-    #                         method="POST", input=model_input)
 
-    # TODO remove me. For testing fetch something over the web.
-    result = urllib2.urlopen('http://azavea.com')
-    return result.read()
+    # TODO actually call geotrellis with model_input
+    census = {
+        'cell_count': 8,
+        'distribution': {
+            'c:commercial': {'cell_count': 5},
+            'a:deciduous_forest': {'cell_count': 3}
+        },
+        'modifications': [
+            {
+                'bmp': 'no_till',
+                'cell_count': 1,
+                'distribution': {
+                    'a:deciduous_forest': {'cell_count': 1},
+                }
+            },
+            {
+                'reclassification': 'd:rock',
+                'cell_count': 1,
+                'distribution': {
+                    'a:deciduous_forest': {'cell_count': 1}
+                }
+            }
+        ]
+    }
+    return census
 
 
 @shared_task
-def run_tr55(model_input, landscape):
+def run_tr55(census, model_input):
     """
     A thin Celery wrapper around our TR55 implementation.
     """
-    time.sleep(5)
 
-    # TODO Get the real simpulation data.
-    # (q, et, inf) = simulate_year(landscape)
+    # TODO Remove after testing.
+    time.sleep(3)
 
-    # TODO Dummy data. Remove me.
-    q = 2
-    et = 3
-    inf = 4
-    return {
-        'q': q,
-        'et': et,
-        'inf': inf,
-    }
+    et_max = 0.207
+    precip_mod = filter(lambda mod: mod['name'] == 'precipitation',
+                        model_input['modifications'])[0]
+    precip = precip_mod['value']
+
+    def simulate_day(cell, cell_count):
+
+        (soil_type, land_use) = cell.lower().split(':')
+        et = et_max * lookup_ki(land_use)
+        return simulate_cell_day((precip, et), cell, cell_count)
+
+    return simulate_modifications(census, fn=simulate_day)
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -156,18 +156,12 @@ def run_tr55(census, model_input):
     A thin Celery wrapper around our TR55 implementation.
     """
 
-    # TODO Remove after testing.
-    time.sleep(3)
-
     et_max = 0.207
-    precip_mod = filter(lambda mod: mod['name'] == 'precipitation',
-                        model_input['modifications'])[0]
-    precip = precip_mod['value']
+    precip = model_input['precip']
 
     def simulate_day(cell, cell_count):
-
-        (soil_type, land_use) = cell.lower().split(':')
+        soil_type, land_use = cell.lower().split(':')
         et = et_max * lookup_ki(land_use)
         return simulate_cell_day((precip, et), cell, cell_count)
 
-    return simulate_modifications(census, fn=simulate_day)
+    return {'runoff': simulate_modifications(census, fn=simulate_day)}
--- a/src/mmw/js/src/core/models.js
+++ b/src/mmw/js/src/core/models.js
@@ -61,36 +61,56 @@ var TaskModel = Backbone.Model.extend({
         }
     },
 
+    // Cancels any currently running jobs. The promise returned
+    // by previous calls to pollForResults will be rejected.
+    reset: function() {
+        this.set({
+            'job': null,
+            'result': null,
+            'status': null
+        });
+    },
+
     pollForResults: function() {
         var defer = $.Deferred(),
             duration = 0,
-            self = this;
+            self = this,
+            expectedJob = self.get('job');
 
         // Check the task endpoint to see if the job is
         // completed. If it is, return the results of
         // the job. If not, check again after
         // pollInterval has elapsed.
         var getResults = function() {
-                if (duration >= self.get('timeout')) {
-                    defer.reject();
-                    return;
-                }
+            if (duration >= self.get('timeout')) {
+                defer.reject();
+                return;
+            }
 
-                self.fetch()
-                    .done(function(response) {
-                        console.log('Polling ' + self.url());
-                        if (response.status !== 'complete') {
-                            duration = duration + self.get('pollInterval');
-                            window.setTimeout(getResults, self.get('pollInterval'));
-                        } else {
-                            defer.resolve(response);
-                        }
-                    })
-                    .fail(defer.reject);
+            // If job was cancelled.
+            if (expectedJob != self.get('job')) {
+                defer.reject({
+                    reset: {
+                        job: expectedJob
+                    }
+                });
+                return;
+            }
+
+            self.fetch()
+                .done(function(response) {
+                    console.log('Polling ' + self.url());
+                    if (response.status !== 'complete') {
+                        duration = duration + self.get('pollInterval');
+                        window.setTimeout(getResults, self.get('pollInterval'));
+                    } else {
+                        defer.resolve(response);
+                    }
+                })
+                .fail(defer.reject);
         };
 
         window.setTimeout(getResults, self.get('pollInterval'));
-
         return defer.promise();
     }
 });
--- a/src/mmw/js/src/core/models.js
+++ b/src/mmw/js/src/core/models.js
@@ -71,7 +71,47 @@ var TaskModel = Backbone.Model.extend({
         });
     },
 
+    // taskHelper should be an object containing an optional object,
+    // postData, an optional function, onStart, and functions pollSuccess,
+    // pollFailure, and startFailure.
+    start: function(taskHelper) {
+        taskHelper = _.defaults(taskHelper, {
+            onStart: _.noop,
+            pollSuccess: _.noop,
+            pollFailure: _.noop,
+            startFailure: _.noop
+        });
+
+        this.reset();
+        if (taskHelper.onStart) {
+            taskHelper.onStart();
+        }
+        var self = this;
+
+        self
+            .fetch({
+                method: 'POST',
+                data: taskHelper.postData
+            })
+            .done(function() {
+                self.pollForResults()
+                    .done(taskHelper.pollSuccess)
+                    .fail(function(error) {
+                        if (error && error.cancelledJob) {
+                            console.log('Job ' + error.cancelledJob + ' was cancelled.');
+                        } else {
+                            taskHelper.pollFailure();
+                        }
+                    });
+            })
+            .fail(taskHelper.startFailure);
+    },
+
     pollForResults: function() {
+        // expectedJob is the value of this.get('job')
+        // associated with a single call to start(). If start()
+        // is called again, the values of this.get('job') and
+        // expectedJob will diverge.
         var defer = $.Deferred(),
             duration = 0,
             self = this,
@@ -89,11 +129,7 @@ var TaskModel = Backbone.Model.extend({
 
             // If job was cancelled.
             if (expectedJob != self.get('job')) {
-                defer.reject({
-                    reset: {
-                        job: expectedJob
-                    }
-                });
+                defer.reject({cancelledJob: expectedJob});
                 return;
             }
 
--- a/src/mmw/js/src/modeling/controllers.js
+++ b/src/mmw/js/src/modeling/controllers.js
@@ -1,3 +1,4 @@
+
 "use strict";
 
 var $ = require('jquery'),
@@ -24,6 +25,8 @@ var ModelingController = {
                 id: projectId
             });
 
+            App.currProject = project;
+
             project
                 .fetch()
                 .done(function(data) {
@@ -40,21 +43,24 @@ var ModelingController = {
                     } else {
                         project.get('scenarios').makeFirstScenarioActive();
                     }
-
+                    project.getResultsIfNeeded();
                 });
         } else {
             project = new models.ProjectModel({
                 name: 'Untitled Project',
                 created_at: Date.now(),
                 area_of_interest: App.map.get('areaOfInterest'),
-                scenarios: new models.ScenariosCollection([
-                    new models.ScenarioModel({
-                        name: 'Current Conditions',
-                        is_current_conditions: true,
-                        active: true
-                    })
-                ])
+                scenarios: new models.ScenariosCollection()
             });
+
+            App.currProject = project;
+
+            var currentConditionsScenario = new models.ScenarioModel({
+                name: 'Current Conditions',
+                is_current_conditions: true,
+                active: true
+            });
+            project.get('scenarios').add(currentConditionsScenario);
 
             project.on('change:id', function(model) {
                 router.navigate(project.getReferenceUrl());
@@ -62,8 +68,9 @@ var ModelingController = {
 
             initScenarioEvents(project);
             initViews(project);
-        }
 
+            project.getResultsIfNeeded();
+        }
     },
 
     modelCleanUp: function() {
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -1,7 +1,8 @@
 "use strict";
 
 var Backbone = require('../../shim/backbone'),
-    _ = require('lodash'),
+    _ = require('underscore'),
+    $ = require('jquery'),
     App = require('../app'),
     coreModels = require('../core/models');
 
@@ -17,12 +18,36 @@ var ModelPackageControlsCollection = Backbone.Collection.extend({
 
 var ModelPackageModel = Backbone.Model.extend();
 
-var Tr55TaskModel = coreModels.TaskModel.extend({});
+var Tr55TaskModel = coreModels.TaskModel.extend({
+    defaults: _.extend(
+        {
+            taskName: 'tr55',
+            taskType: 'modeling'
+        },
+        coreModels.TaskModel.prototype.defaults
+    )
+});
 
-var ResultModel = Backbone.Model.extend({});
+var ResultModel = Backbone.Model.extend({
+    defaults: {
+        polling: false
+    }
+});
 
 var ResultCollection = Backbone.Collection.extend({
-    model: ResultModel
+    model: ResultModel,
+
+    setPolling: function(polling) {
+        this.forEach(function(resultModel) {
+            resultModel.set('polling', polling);
+        });
+    },
+
+    setNullResults: function() {
+        this.forEach(function(resultModel) {
+            resultModel.set('result', null);
+        });
+    }
 });
 
 var ProjectModel = Backbone.Model.extend({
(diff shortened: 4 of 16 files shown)


lewfish:
@azaveaci rebuild

lewfish:
It builds now. I forgot to rebundle tests, leading me to think that the testem tests passed when I ran them locally.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -35,22 +36,84 @@ var ModelingController = {
...
-                    })
-                ])
+                scenarios: new models.ScenariosCollection()
+            });
+
+            // Poll for model results and set the results in the scenario when done.
+            var pollModel = function(scenario) {
+                // If scenario doesn't have its own task model yet, clone and set it from the project.
+                if (!scenario.get('taskModel')) {
+                    scenario.set('taskModel', $.extend(true, {}, project.get('taskModel')));
```
caseypt: Since every scenario will need it's own taskModel, we should assign `scenario.taskModel` on initialization.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -1,3 +1,4 @@
+
```
caseypt: Extra space here.

src/mmw/js/src/core/models.js
```diff
@@ -61,36 +61,89 @@ var TaskModel = Backbone.Model.extend({
...
+                taskHelper.startFailure();
+            });
+    },
+
     pollForResults: function() {
         var defer = $.Deferred(),
             duration = 0,
-            self = this;
+            self = this,
+            expectedJob = self.get('job');
```
caseypt: Is `expectedJob` always the job that was started most recently?

src/mmw/js/src/core/models.js
```diff
@@ -61,36 +61,89 @@ var TaskModel = Backbone.Model.extend({
...
+
+        self
+            .fetch({
+                method: 'POST',
+                data: taskHelper.postData
+            })
+            .done(function() {
+                self.pollForResults()
+                    .done(function() {
+                        taskHelper.pollSuccess();
```
caseypt: If this function is optional, we should check that it exists before calling it. Also consider making it a class method.

caseypt: Just reread the comment. Ignore the suggestion about it being optional.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -35,22 +36,84 @@ var ModelingController = {
...
+            // Poll for model results and set the results in the scenario when done.
+            var pollModel = function(scenario) {
+                // If scenario doesn't have its own task model yet, clone and set it from the project.
+                if (!scenario.get('taskModel')) {
+                    scenario.set('taskModel', $.extend(true, {}, project.get('taskModel')));
+                }
+
+                var taskModel = scenario.get('taskModel'),
+                    runoffModel = scenario.get('results').find(function(model) {
```
caseypt: Consider renaming this `runoffResultsModel`, only because `runoffModel` could be confused for TR-55.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -35,22 +36,84 @@ var ModelingController = {
...
+                    return model.get('name') == 'runoff';
```
caseypt: Indentation is off here.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -35,22 +36,84 @@ var ModelingController = {
...
+                        })
+                    },
+
+                    onStart: function() {
+                        runoffModel.set('polling', true);
+                    },
+
+                    setRunoffResult: function() {
+                        var rawResult = taskModel.get('result');
+                        var result = rawResult == "" ? null : JSON.parse(rawResult);
```
caseypt: App style is one `var` statement.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -35,22 +36,84 @@ var ModelingController = {
...
-                        name: 'Current Conditions',
-                        is_current_conditions: true,
-                        active: true
-                    })
-                ])
+                scenarios: new models.ScenariosCollection()
+            });
+
+            // Poll for model results and set the results in the scenario when done.
+            var pollModel = function(scenario) {
```
caseypt: I think we may need to re-architect this a bit to make it testable since it's a crucial component of the app. What is now `pollModel` could be reworked as a method on `ScenarioModel`.

caseypt:
Nice commit history!

src/mmw/js/src/modeling/controllers.js
lewfish: I can move pollModel into scenarioModel, as long as I can give scenarioModel a reference to project. I'll look into how to do that. I have some ideas for testing. Can you come over some time and we can discuss them?

caseypt: For sure.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -35,22 +36,84 @@ var ModelingController = {
...
+                        this.setRunoffResult();
+                    }
+                };
+
+                taskModel.start(taskHelper);
+            };
+
+            // The debounce interval is an estimate of the amount of time
+            // to start the request and
+            var debouncedPollModel = _.debounce(pollModel, 500);
```
caseypt: With the way this is setup, every scenario is sharing the debounced function. So making changes to one scenario and then making changes to another scenario disregards the results for the initial scenario. Each scenario should operate independently.

src/mmw/js/src/modeling/controllers.js
caseypt: Running models is only wired up for new projects, and not saved projects.

caseypt:
Modifying one scenario, then quickly modifying another scenario produces this in the console:

selection_052

src/mmw/apps/modeling/tasks.py
```diff
@@ -120,34 +122,51 @@ def make_gt_service_call_task(model_input):
...
-        'et': et,
-        'inf': inf,
-    }
+
+    # TODO Remove after testing.
+    time.sleep(3)
+
+    et_max = 0.207
+    precip_mod = filter(lambda mod: mod['name'] == 'precipitation',
+                        model_input['modifications'])[0]
```
kdeloach: Consider graceful exception if precipitation is missing.

src/mmw/js/src/core/models.js
```diff
@@ -61,36 +61,89 @@ var TaskModel = Backbone.Model.extend({
...
+            'job': null,
+            'result': null,
+            'status': null
+        });
+    },
+
+    // taskHelper should be an object containing an optional object,
+    // postData, an optional function, onStart, and functions pollSuccess,
+    // pollFailure, and startFailure.
+    start: function(taskHelper) {
```
kdeloach: For "options" args one good convention is to provide defaults. Consider adding:

    taskHelper = _.defaults(taskHelper, {
        onStart: _.identity,
        pollForSuccess: _.identity,
        ...
    });

src/mmw/js/src/modeling/models.js
```diff
@@ -170,6 +192,29 @@ var ScenarioModel = Backbone.Model.extend({
 
         this.on('change:project change:name', this.attemptSave, this);
         this.get('modifications').on('add remove', this.attemptSave, this);
+
+        var debouncedPoll = _.debounce(_.bind(this.pollForResults, this), 500);
+        this.get('modifications').on('add change', function() {
```
kdeloach: I could be wrong, but you may be able to pass `debouncedPoll` as the second arg directly here.

src/mmw/js/src/modeling/models.js
```diff
@@ -170,6 +192,29 @@ var ScenarioModel = Backbone.Model.extend({
...
+            debouncedPoll();
+        });
+        this.set('taskModel', $.extend(true, {}, App.currProject.get('taskModel')));
```
kdeloach: Since we can now get `taskModel` from project it seems redundant to store task model on the scenarios as well. Consider removing this property from scenarios.

src/mmw/js/src/modeling/models.js
```diff
@@ -207,8 +252,73 @@ var ScenarioModel = Backbone.Model.extend({
...
+
         return response;
-     }
+    },
+
+    // Poll the taskModel for results and reset the results collection when done.
+    // If not successful, the results collection is reset to be empty.
+    pollForResults: function() {
+        var self = this,
+            taskModel = this.get('taskModel'),
```
kdeloach: Consider `taskModel  = App.currProject.get('taskModel')`.

src/mmw/js/src/modeling/models.js
```diff
@@ -241,7 +351,8 @@ var ScenariosCollection = Backbone.Collection.extend({
 
     createNewScenario: function() {
         var scenario = new ScenarioModel({
-            name: this.makeNewScenarioName('New Scenario')
+            name: this.makeNewScenarioName('New Scenario'),
+            taskModel: $.extend(true, {}, App.currProject.get('taskModel'))
```
kdeloach: Same comment as above, this may be redundant now.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -35,26 +38,29 @@ var ModelingController = {
...
-                    new models.ScenarioModel({
-                        name: 'Current Conditions',
-                        is_current_conditions: true,
-                        active: true
-                    })
-                ])
+                scenarios: new models.ScenariosCollection()
+            });
+
+            App.currProject = project;
```
kdeloach: This isn't a bad idea, but I think we need to be careful about adding a new global var. 

Couple of thoughts:

1. Consider renaming to `currentProject`.
2. Add a default value and documentation to the `App` class.
3. We need to be careful about this becoming stale. For example, switching between views or switching projects should never contain a reference to the wrong project.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -35,26 +38,29 @@ var ModelingController = {
...
                 area_of_interest: App.map.get('areaOfInterest'),
-                scenarios: new models.ScenariosCollection([
-                    new models.ScenarioModel({
-                        name: 'Current Conditions',
-                        is_current_conditions: true,
-                        active: true
-                    })
-                ])
+                scenarios: new models.ScenariosCollection()
+            });
```
kdeloach: For readability, consider condensing this function to something like...

    var project = this.getProject(projectId);
    App.currProject = project;
    // Wire up events...
    initViews(project);

src/mmw/js/src/modeling/controllers.js
```diff
@@ -35,26 +38,29 @@ var ModelingController = {
...
 
-            var scenariosColl = project.get('scenarios'),
+            var scenarios = project.get('scenarios'),
                 mapView = App.getMapView();
-            scenariosColl.on('change:activeScenario', function(scenario) {
+            scenarios.on('change:activeScenario', function(scenario) {
                 mapView.updateModifications(scenario.get('modifications'));
             });
-
-            initViews(project);
```
kdeloach: Was there a reason we had to move this line and change how `ScenariosCollection` gets created here? Or just preference?

src/mmw/js/src/core/models.js
```diff
@@ -61,36 +61,89 @@ var TaskModel = Backbone.Model.extend({
...
-                            window.setTimeout(getResults, self.get('pollInterval'));
-                        } else {
-                            defer.resolve(response);
-                        }
-                    })
-                    .fail(defer.reject);
+            // If job was cancelled.
+            if (expectedJob != self.get('job')) {
+                defer.reject({
+                    reset: {
```
kdeloach: Consider flattening this data structure, if possible. IMO `reject({ reset: { job: ... } })` isn't more descriptive than `reject(expectedJob)` or maybe even `reject(expectedJob + ' was cancelled')`.

src/mmw/js/src/modeling/models.js
```diff
@@ -170,6 +192,29 @@ var ScenarioModel = Backbone.Model.extend({
...
+
+        var resultCollection;
+        if (App.currProject.get('model_package') == 'tr-55') {
+            resultCollection = new ResultCollection([
+                {
+                    name: 'runoff',
+                    displayName: 'Runoff',
+                    result: null
+                },
+	            {
```
kdeloach: Indentation.

src/mmw/js/src/modeling/models.js
```diff
@@ -207,8 +252,73 @@ var ScenarioModel = Backbone.Model.extend({
...
+    },
+
+    // Poll the taskModel for results and reset the results collection when done.
+    // If not successful, the results collection is reset to be empty.
+    pollForResults: function() {
+        var self = this,
+            taskModel = this.get('taskModel'),
+            setPolling = function(polling) {
+                self.get('results').forEach(function(resultModel) {
+                    resultModel.set('polling', polling);
```
kdeloach: We should add a default value and maybe a comment for all model attributes.

src/mmw/js/src/modeling/models.js
```diff
@@ -207,8 +252,73 @@ var ScenarioModel = Backbone.Model.extend({
...
+                });
+            },
+            setNullResults = function() {
```
kdeloach: Consider moving this function to the results collection class.

src/mmw/js/src/modeling/models.js
```diff
@@ -207,8 +252,73 @@ var ScenarioModel = Backbone.Model.extend({
...
+            setResults = function() {
+                var rawServerResults = taskModel.get('result');
+                if (rawServerResults == "" || rawServerResults == null) {
+                    setNullResults();
+                } else {
+                    var serverResults = JSON.parse(rawServerResults);
+                    self.get('results').forEach(function(resultModel) {
+                        resultModel.set('result', serverResults[resultModel.get('name')]);
+                    });
+                }
```
kdeloach: Instead of micromanaging results client-side, can we just populate the entire results collection with whatever the server responds with?

src/mmw/js/src/modeling/models.js
```diff
@@ -299,14 +410,19 @@ var ScenariosCollection = Backbone.Collection.extend({
...
-function getControlsForModelPackage(modelPackageName) {
-    switch (modelPackageName) {
-        case 'tr-55':
+function getControlsForModelPackage(modelPackageName, options) {
+    if (modelPackageName == 'tr-55') {
+        if (options && options.is_current_conditions) {
+            return new ModelPackageControlsCollection([
+                new ModelPackageControlModel({ name: 'precipitation' })
+            ]);
+        } else {
```
kdeloach: Nice change, not a fan of switch?

src/mmw/js/src/modeling/views.js
```diff
@@ -514,14 +515,6 @@ var ToolbarTabContentView = Marionette.CompositeView.extend({
...
             controlName = model.get('name');
         return modificationsColl.findWhere({ name: controlName });
-    },
-
-    getTemplate: function() {
-        if (this.model.get('is_current_conditions')) {
-            return currentConditionsToolbarTabContentTmpl;
-        } else {
-            return scenarioToolbarTabContentTmpl;
-        }
```
kdeloach: Glad to get rid of this, you may want to also delete the "current conditions" template as well.

src/mmw/js/src/modeling/views.js
```diff
@@ -654,6 +644,11 @@ var ResultsTabPanelView = Marionette.ItemView.extend({
     template: resultsTabPanelTmpl,
     attributes: {
         role: 'presentation'
+    },
+    initialize: function() {
+        this.listenTo(this.model, 'change:polling', function() {
```
kdeloach: I think this could just be `this.listenTo(this.model, 'change:polling', this.render)`.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -35,26 +38,29 @@ var ModelingController = {
...
 
             project.on('change:id', function(model) {
                 router.navigate('model/' + model.id);
             });
 
-            var scenariosColl = project.get('scenarios'),
+            var scenarios = project.get('scenarios'),
                 mapView = App.getMapView();
-            scenariosColl.on('change:activeScenario', function(scenario) {
+            scenarios.on('change:activeScenario', function(scenario) {
```
kdeloach: I've been using a suffix to distinguish between an array of models and a collection of models but I'd be happy to start using this naming convention if it makes the most sense to everyone else.

src/mmw/js/src/modeling/models.js
```diff
@@ -17,9 +18,21 @@ var ModelPackageControlsCollection = Backbone.Collection.extend({
 
 var ModelPackageModel = Backbone.Model.extend();
 
-var Tr55TaskModel = coreModels.TaskModel.extend({});
+var Tr55TaskModel = coreModels.TaskModel.extend({
+    defaults: _.extend(
```
kdeloach: Probably doesn't matter but `_.defaults` exists for this purpose.

src/mmw/js/src/modeling/controllers.js
lewfish: I prefer singular vs plural to distinguish models and collections because it's more succinct and matches natural language well. We're also using it in `modeling/models.js` in the attribute names `results` and `scenarios` and also in view names in `modeling/views.py`.

src/mmw/js/src/modeling/models.js
lewfish: Each scenario needs its own taskModel because the scenarios can be polling for results concurrently.

kdeloach: Now I get it. Consider adding a factory method on project to make this clearer `App.currentProject.createTaskModel()` instead of using extend. It probably doesn't make sense to have an instance of task model at the project level.

src/mmw/apps/modeling/tasks.py
```diff
@@ -120,34 +122,51 @@ def make_gt_service_call_task(model_input):
...
+    precip = precip_mod['value']
+
+    def simulate_day(cell, cell_count):
+        (soil_type, land_use) = cell.lower().split(':')
```
kdeloach: Parens not needed here.

lewfish:
@azaveaci rebuild

lewfish:
As a result of the input validation commit, you should now see an error message about quality being missing from the server response (since it's not returning water quality results).

caseypt:
I attempted to pull this down and test it out, but there are merge conflicts with the latest on `develop`.

lewfish:
Ok, I rebased.

src/mmw/apps/modeling/tasks.py
```diff
@@ -120,34 +122,51 @@ def make_gt_service_call_task(model_input):
...
-    et = 3
-    inf = 4
-    return {
-        'q': q,
-        'et': et,
-        'inf': inf,
-    }
+
+    # TODO Remove after testing.
+    time.sleep(3)
```
caseypt: Let's not forget to remove this. In testing, some of the jobs run over 5 seconds, so they timeout. After removing this, we can bump the `TaskModel.timeout` value up if the polling is still timing out.

caseypt:
Testing this out now. Things are looking good. Nice job.

I'm going to +1 Kevin's comments that I think should be done before we merge this in and create issues for anything larger.

src/mmw/apps/modeling/tasks.py
caseypt: +1

lewfish:
I addressed most of Kevin's comments. I didn't follow a few because they didn't really make sense on further reflection, but if there's anything you really think should be in there, let me know.

src/mmw/js/src/modeling/models.js
```diff
@@ -110,9 +136,15 @@ var ProjectModel = Backbone.Model.extend({
             var user_id = response.user.id,
                 scenariosCollection = this.get('scenarios'),
                 scenarios = _.map(response.scenarios, function(scenario) {
+                    // We don't want to set the results until a future
```
caseypt: Consider prefacing this with `TODO` so we can easily find it later.

caseypt:
I haven't been able to definitively reproduce this, but sometimes the model runs fail completely, even with a big timeout. It seems to happen more with the current conditions scenario. I don't think there's anything to do now, but just a heads up.

lewfish:
Hmm, there will be failures (as a result of timing out) if you put a bunch of jobs in the celery queue by rapidly changing the precipitation value, but what you're reporting might be different.

src/mmw/apps/modeling/tasks.py
caseypt: Now seeing where you did this further up the chain.

lewfish: When I do the validation in the view, I set the precipitation as a field on the model_input, but I forgot to actually use it here. I'll do this when you're done.

caseypt:
+1. Besides the above-mentioned issue, it's working well. Nice job working through this.

If you can squash some of these commits without too much hassle, considering doing it.

Also, someone else should test out running models.

caseypt:
Just wanted to note that I was able to run models for both new and restored projects.

mmcfarland:
When running a model for a new scenario, you get a polling error if you add a new modification before adjusting the slider, presumably because it doesn't start with a default value.

`Missing single precipitation modification`

lewfish:
What should we do about that? We could give the precipitation slider a default value. This would mean that when Current Conditions will be polling immediately when creating a new project.

mmcfarland:
@lewfish Yes, let's give it a non-0 default value.  Start it with 1.0" and we'll see what value seems reasonable to Stroud

mmcfarland:
That is, I think that all new scenarios should start at the default as they're created.

kdeloach:
We should just populate some default mods upon creation. This way we could default precip to 0.

mmcfarland:
We could also just have the default percip at 0, but since that will produce no runoff (presumably?) it isn't very interesting or intuitive for someone doing some experimenting (ie, having no results on your modifications).  I think having random modifications added to your scenario would be confusing as well.  1" of rain seems like a reasonable storm amount to view results.

lewfish:
When a saved scenario is loaded, do we want it to immediately start polling for the results using the saved precipitation value?

mmcfarland:
We're in a pinch about that.  In real life, after this PR, we'll have the results saved for all saved scenario, so I don't want to spend too much on a workflow that will be obsolete.  I suppose it's a good check that if the results are missing on a saved scenario, we kick off a job to calculate them, this would be sensible behavior that in the future probably won't get run much.  So yes, and always use the saved precip value.

src/mmw/js/src/modeling/views.js
```diff
@@ -706,34 +704,43 @@ var BarChartView = Marionette.ItemView.extend({
...
+                    },
+                    {
+                        type: 'Modified',
+                        inf: result['modified']['inf'],
+                        runoff: result['modified']['runoff'],
+                        et: result['modified']['et']
+                    }
+                ],
+                options = {
+                    barColors: ['#329b9c', '#4aeab3', '#4ebaea'],
```
lliss: Not sure if we'll be using these elsewhere but we might want to store these as constants and add some comment about what what color each one actually is.

PR #348, opened 2015-06-10
Feature/lol/save projects and scenarios

lliss:
Connects #226 

Adds auto save functionality to projects and scenarios.

Note: Probably want to test with the network tab open to watch the request moving.

To test:
 * Log in.
 * Create shape to model.
 * Modify project either by changing the name or by adding a scenario
 * Check DB. Both project and scenario should be saved.
 * Make modifications
 * Open URL in another window. It should restore to the same state.
 * Change Project privacy to public.
 * Log out.
 * Check project again. Look for modification based controls (there shouldn't be any).

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -34,10 +34,15 @@ var ProjectModel = Backbone.Model.extend({
         area_of_interest: null,    // GeoJSON
         model_package: '',         // Package name
         taskModel: null,           // TaskModel
-        scenarios: null            // ScenariosCollection
+        scenarios: null,            // ScenariosCollection
+        uid: null
     },
 
     initialize: function() {
+        var scenarios = this.get('scenarios');
+        if (scenarios === null || typeof scenarios === 'string') {
+            this.set('scenarios', new ScenariosCollection());
+        }
         // TODO: For a new project, users will eventually
         // be able to choose which modeling package
         // they want to use in their project. For
@@ -45,6 +50,10 @@ var ProjectModel = Backbone.Model.extend({
         // hard-coded here.
         this.set('model_package', 'tr-55');
         this.set('taskModel', new Tr55TaskModel());
+        this.set('uid', App.user.get('uid'));
+
+        this.listenTo(this.get('scenarios'), 'add', this.addIdsToScenarios, this);
+        this.on('change:name', this.saveProjectAndScenarios, this);
     },
 
     updateName: function(newName) {
@@ -53,48 +62,52 @@ var ProjectModel = Backbone.Model.extend({
         this.set('name', newName);
     },
 
-    saveAll: function() {
-        var self = this;
-
-        // If we haven't saved the project before,
-        // hold on to the scenarios so that we can
-        // save each one once we have an ID for the project
-        if (this.isNew) {
-            var tempScenarios = this.get('scenarios');
-        }
-
-        this.save()
-            .done(function(resp) {
-                console.log('Saved project');
-
-                var scenarios = tempScenarios ? tempScenarios : self.get('scenarios');
-
-                scenarios.each(function(scenario) {
-                    if (!scenario.get('project')) {
-                        scenario.set('project', self.get('id'));
-                    }
-
-                    scenario
-                        .save()
-                        .done(function(resp) {
-                            console.log('Saved scenario');
-                        })
-                        .fail(function(resp) {
-                            console.log('Failed to save scenario');
-                        });
+    saveProjectAndScenarios: function() {
+        if (!this.get('id')) {
+            // We haven't saved the project before, save the project and then
+            // set the project ID on each scenario, then reattach the senarios
+            // to the project.
+            var self = this;
+            this.save()
+                .done(function() {
+                    self.updateProjectScenarios(self.get('id'), self.get('scenarios'));
+                })
+                .fail(function() {
+                    console.log('Failed to save project');
                 });
-            })
-            .fail(function(resp) {
-                console.log('Failed to save project');
-            });
+        } else {
+            this.save()
+                .fail(function() {
+                    console.log('Failed to save project');
+                });
+        }
+    },
+
+    addIdsToScenarios: function() {
+        var projectId = this.get('id');
+        if (!projectId) {
+            this.saveProjectAndScenarios();
+        } else {
+            this.updateProjectScenarios(projectId, this.get('scenarios'));
+        }
+    },
+
+    updateProjectScenarios: function(projectId, scenarios) {
+        scenarios.each(function(scenario) {
+            if (!scenario.get('project')) {
+                scenario.set('project', projectId);
+            }
+        });
     },
 
     parse: function(response, options) {
-        var scenariosCollection = new ScenariosCollection();
+        if (response.scenarios) {
+            // If we returned scenarios (probably from a GET) then set them.
+            var scenariosCollection = this.get('scenarios');
+            scenariosCollection.reset(response.scenarios);
+        }
 
-        scenariosCollection.reset(response.scenarios);
-
-        response.scenarios = scenariosCollection;
+        // TODO: Does this hurt anything if we always set.
         response.taskModel = new Tr55TaskModel();
 
         return response;
@@ -136,6 +149,15 @@ var ScenarioModel = Backbone.Model.extend({
     initialize: function(attrs, options) {
         Backbone.Model.prototype.initialize.apply(this, arguments);
         this.set('modifications', ModificationsCollection.create(attrs.modifications));
+
+        this.on('change:project change:name', this.attemptSave, this);
+        this.get('modifications').on('add remove', this.attemptSave, this);
+    },
+
+    attemptSave: function() {
+        this.save().fail(function() {
+            console.log('Failed to save scenario');
+        });
     },
 
     getSlug: function() {
@@ -160,9 +182,12 @@ var ScenarioModel = Backbone.Model.extend({
     },
 
     parse: function(response, options) {
-        var modificationsColl = this.get('modifications');
-        modificationsColl.reset(response.modifications);
-    }
+        // Modifications are essentially write only. So if we have them on our
+        // model, we shouldn't reset them from the server. Pull them off of
+        // the response to prevent overwriting them.
+        delete response.modifications;
+        return response;
+     }
 });
 
 var ScenariosCollection = Backbone.Collection.extend({
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -63,7 +63,7 @@ var ProjectModel = Backbone.Model.extend({
     },
 
     saveProjectAndScenarios: function() {
-        if (App.user.get('id') !== this.get('user_id') || App.user.get('id') < 1) {
+        if (!App.user.loggedInUserMatch(this.get('user_id'))) {
             return;
         }
 
@@ -173,7 +173,7 @@ var ScenarioModel = Backbone.Model.extend({
     },
 
     attemptSave: function() {
-        if (App.user.get('id') !== this.get('user_id') || App.user.get('id') < 1) {
+        if (!App.user.loggedInUserMatch(this.get('user_id'))) {
             return;
         }
         this.save().fail(function() {
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -34,8 +34,8 @@ var ProjectModel = Backbone.Model.extend({
         area_of_interest: null,    // GeoJSON
         model_package: '',         // Package name
         taskModel: null,           // TaskModel
-        scenarios: null,            // ScenariosCollection
-        uid: null
+        scenarios: null,           // ScenariosCollection
+        user_id: 0                 // User that created the project
     },
 
     initialize: function() {
@@ -50,7 +50,7 @@ var ProjectModel = Backbone.Model.extend({
         // hard-coded here.
         this.set('model_package', 'tr-55');
         this.set('taskModel', new Tr55TaskModel());
-        this.set('uid', App.user.get('uid'));
+        this.set('user_id', App.user.get('id'));
 
         this.listenTo(this.get('scenarios'), 'add', this.addIdsToScenarios, this);
         this.on('change:name', this.saveProjectAndScenarios, this);
@@ -63,7 +63,7 @@ var ProjectModel = Backbone.Model.extend({
     },
 
     saveProjectAndScenarios: function() {
-        if (App.user.get('uid') !== this.get('uid') || App.user.get('uid') < 1) {
+        if (App.user.get('id') !== this.get('user_id') || App.user.get('id') < 1) {
             return;
         }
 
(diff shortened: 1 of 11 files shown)


lliss:
Currently there is a timing issue, where the logged in user sometimes returns after the view builds so the proper controls don't come back.

caseypt:
I think we can remove this line now: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/views.js#L173

mmcfarland:
Every time I updated the rain slider, it fires off two PUT requests for the scenario, one without modifications and one with.

lliss:
Hmm. That's probably because the modifications don't actually change. Once is removed and another is added. Hard to differentiate between updating (remove then add) and just removing a modification.

mmcfarland:
Loading a project with scenarios and immediately adjusting the rain slider will cause no save requests to fire.

mmcfarland:
Loading a saved project that I created removes all but the "print" menu items for Project and Scenario menus

mmcfarland:
Looks like https://github.com/lliss/model-my-watershed/commit/96ad97c603cc32ffdf21916345e03e42e0e46ef5#diff-9abeff034ba5a68b4814ee90721888ebR186 is preventing saved projects from saving correctly because `App.user.get('uid')` isn't set.

lliss:
Yeah that's part of the uid not getting properly set on the view. Looking into that now.

lliss:
I'll bet the rename control isn't there either when that happens.

src/mmw/js/src/modeling/models.js
```diff
@@ -53,48 +67,71 @@ var ProjectModel = Backbone.Model.extend({
...
                 });
-            })
-            .fail(function(resp) {
-                console.log('Failed to save project');
+        }
+    },
+
+    addIdsToScenarios: function() {
+        if (!this.get('id')) {
+            this.saveProjectAndScenarios();
```
mmcfarland: Both `addIdsToScenarios` and `saveProjectAndScenarios` can add Ids and save the project and scenarios.  Consider consolidating the methods or splitting them apart so they each are responsible for one or the other, as their names suggest.

mmcfarland:
Behavior looks good now.

lliss:
Tests are failing. I'm fixing them now.

src/mmw/js/src/modeling/models.js
```diff
@@ -130,12 +164,32 @@ var ScenarioModel = Backbone.Model.extend({
...
+        active: false,
+        uid: null
     },
 
     initialize: function(attrs, options) {
         Backbone.Model.prototype.initialize.apply(this, arguments);
+        this.set('uid', App.user.get('uid'));
         this.set('modifications', ModificationsCollection.create(attrs.modifications));
+        this.on({
+            'change:project': this.attemptSave,
```
mmcfarland: Fairly sure that this can be condensed to `'change:project change:name': this.attemptSave` and likewise for modifications below.

src/mmw/js/src/modeling/models.js
```diff
@@ -34,10 +34,15 @@ var ProjectModel = Backbone.Model.extend({
...
         model_package: '',         // Package name
         taskModel: null,           // TaskModel
-        scenarios: null            // ScenariosCollection
+        scenarios: null,            // ScenariosCollection
+        uid: null
     },
 
     initialize: function() {
+        var scenarios = this.get('scenarios');
+        if (scenarios === null || typeof scenarios === 'string') {
```
mmcfarland: Seems like we should have complete control during Project initialization.  What are the cases where this is null or a string?  If we're doing a custom parse on load, I'm not sure why this is necessary.

src/mmw/js/src/modeling/models.js
```diff
@@ -45,6 +50,11 @@ var ProjectModel = Backbone.Model.extend({
         // hard-coded here.
         this.set('model_package', 'tr-55');
         this.set('taskModel', new Tr55TaskModel());
+        this.set('uid', App.user.get('uid'));
+
+        this.get('scenarios').on('add', this.addIdsToScenarios, this);
+        this.on('change:name', this.saveProjectAndScenarios, this);
```
mmcfarland: This and the following line can get merged.

src/mmw/js/src/modeling/models.js
```diff
@@ -45,6 +50,11 @@ var ProjectModel = Backbone.Model.extend({
         // hard-coded here.
         this.set('model_package', 'tr-55');
         this.set('taskModel', new Tr55TaskModel());
+        this.set('uid', App.user.get('uid'));
+
+        this.get('scenarios').on('add', this.addIdsToScenarios, this);
```
mmcfarland: Consider using `listenTo` instead of `on` so undelegating the events is handled automatically.

src/mmw/js/src/modeling/models.js
```diff
@@ -130,12 +164,32 @@ var ScenarioModel = Backbone.Model.extend({
...
+        }, this);
+
+        this.get('modifications').on({
+            'add': this.attemptSave,
+            'remove': this.attemptSave
+        }, this);
+    },
+
+    attemptSave: function() {
+        if (App.user.get('uid') !== this.get('uid')) {
```
mmcfarland: These checks also need to account for Guest users, who don't have a `uid`.  Currently sends bad requests because `undefined !== undefined` is false.

src/mmw/js/src/modeling/models.js
lliss: Ok. Will do.

src/mmw/js/src/modeling/models.js
lliss: Makes sense. I'll test it out.

src/mmw/js/src/modeling/models.js
lliss: It may not be. @kdeloach and I were looking at it yesterday and this was based on a fix he had in place. I'll try removing it.

src/mmw/js/src/modeling/models.js
lliss: Right good. catch. I'm going to make the API endpoint return a uid of zero for guest users and then update the checks for user matching to ensure a non zero and matching uid.

src/mmw/js/src/modeling/models.js
kdeloach: I think my fix had to do with modifications not scenarios. Not sure what this is about.

src/mmw/js/src/modeling/models.js
```diff
@@ -53,48 +63,71 @@ var ProjectModel = Backbone.Model.extend({
...
     parse: function(response, options) {
-        var scenariosCollection = new ScenariosCollection();
+        if (response.scenarios) {
+            // If we returned scenarios (probably from a GET) then set them.
+            var uid = response.uid = response.user.id,
+                scenariosCollection = this.get('scenarios'),
+                scenarios = _.map(response.scenarios, function(scenario) {
+                    var modifications = _.map(scenario.modifications, function(modification) {
+                        return new ModificationModel(modification);
+                    });
```
kdeloach: You may be able to just call `scenarioModel.get('modifications').reset(scenario.modifications);` (reset should cast raw objects to models).

src/mmw/js/src/modeling/models.js
```diff
@@ -103,8 +136,8 @@ var ProjectModel = Backbone.Model.extend({
 
 var ModificationModel = coreModels.GeoModel.extend({
     defaults: _.extend({
-            name: '',
-            type: ''
+            'name': '',
+            'type': ''
```
kdeloach: I feel like the only attribute keys that should be quoted are ones that contain special chars or symbols. Since model attributes should more or less conform to variable naming conventions, I would rather not quote these.

src/mmw/js/src/modeling/models.js
```diff
@@ -130,12 +164,32 @@ var ScenarioModel = Backbone.Model.extend({
         name: '',
         is_current_conditions: false,
         modifications: null,      // ModificationsCollection
-        active: false
+        active: false,
+        uid: null
```
kdeloach: This attribute could be confused for "unique ID" when it's really for "user id".

src/mmw/js/src/modeling/views.js
```diff
@@ -40,6 +40,30 @@ var ModelingHeaderView = Marionette.LayoutView.extend({
...
+        this.scenariosRegion.show(new ScenariosView({
+            collection: this.model.get('scenarios')
+        }));
+
+        this.toolbarRegion.empty();
+        this.toolbarRegion.show(new ToolbarTabContentsView({
+            collection: this.model.get('scenarios'),
+            model_package: this.model.get('model_package')
+        }));
+    },
```
kdeloach: Looks like a copy & paste of `onShow` there's probably a way to combine these.

src/mmw/js/src/modeling/models.js
lliss: Cool. Will do.

src/mmw/js/src/modeling/models.js
```diff
@@ -53,48 +63,71 @@ var ProjectModel = Backbone.Model.extend({
...
-
-        // If we haven't saved the project before,
-        // hold on to the scenarios so that we can
-        // save each one once we have an ID for the project
-        if (this.isNew) {
-            var tempScenarios = this.get('scenarios');
+    saveProjectAndScenarios: function() {
+        if (App.user.get('uid') !== this.get('uid')) {
+            return;
         }
```
kdeloach: This is a nice check but hiding the save/delete/etc. buttons for non-owners should be sufficient. If someone tries to do it via client-side API the endpoint should fail.

src/mmw/js/src/modeling/models.js
```diff
@@ -53,48 +63,71 @@ var ProjectModel = Backbone.Model.extend({
...
-                    if (!scenario.get('project')) {
+        var self = this,
+            scenarios = this.get('scenarios');
+        if (!this.get('id')) {
+            // We haven't saved the project before, save the project and then
+            // set the project ID on each scenario, then reattach the senarios
+            // to the project.
+            this.save()
+                .done(function() {
+                    scenarios.each(function(scenario) {
```
kdeloach: If we set the project ID on scenario models in the `parse` function we can eliminate this step.

src/mmw/js/src/modeling/models.js
lliss: I'll add a comment.

src/mmw/js/src/modeling/models.js
lliss: Yes. Will update.

src/mmw/js/src/modeling/models.js
```diff
@@ -130,12 +163,32 @@ var ScenarioModel = Backbone.Model.extend({
...
+            'change:name': this.attemptSave
+        }, this);
+
+        this.get('modifications').on({
+            'add': this.attemptSave,
+            'remove': this.attemptSave
+        }, this);
+    },
+
+    attemptSave: function() {
```
kdeloach: Consider letting the server fail if a non-owner tries to save. We shouldn't display save buttons for non-owners.

src/mmw/js/src/modeling/models.js
kdeloach: Consider renaming to `user_id`?

src/mmw/js/src/modeling/models.js
lliss: Ok. I'll remove them. My editor was highlighting `type` so I quoted it to reduce visual noise.

src/mmw/js/src/modeling/models.js
lliss: unique id would be better as uuid. I think uid is pretty common. I'd prefer to leave it unless someone else feels strongly.

src/mmw/js/src/modeling/models.js
kdeloach: Ah, that's actually a mistake. `type` should be renamed to `value` I just didn't get around to fixing it yet.

src/mmw/js/src/modeling/models.js
lliss: Yeah but it creates a lot of unnecessary noise on the server and in the logs. Why let requests we know will fail go through.

src/mmw/js/src/modeling/models.js
mmcfarland: I was actually going to initially comment that I'd prefer `id` just like every other resource we have.  So now there's _three_ options.  Not sure it's necessary to have special behavior for the user, I'd prefer to have things consistent.  `user.user_id` and `user.uid` both seem a little redundant, also.  When saved on another object, I think `userId` makes the most sense and is consistent with how we would store `projectId/scenarioId` (or rather, snake_cased).

src/mmw/js/src/modeling/models.js
mmcfarland: I agree to keep the check - I think the confusion here might be that there are no "buttons" that fire this off, it's really just modifications to model attributes.  The check is so that we don't try to save it, given that it will fail, not a belt and suspenders kind of check for a hidden button.

kdeloach: Good point, I didn't account for the fact that things auto-save now.

src/mmw/js/src/modeling/models.js
kdeloach: If we associated a user ID with a scenario in the DB the column would probably be `user_id` and not `uid` that's my rationale.

src/mmw/js/src/modeling/models.js
lliss: Turns out this is needed because of the way the controller creates the view.

lliss:
Went with `user_id` when referencing the user and `id` on the user object.

lliss:
I'm going to add two tests for the UI to check that the logged out user that doesn't own a project and scenario doesn't see the controls. Otherwise, I think everything else has been addressed.

lliss:
@mmcfarland Rebased and enhanced with new tests.

rajadain:
Maybe this is a really edge use case, and perhaps it can be handled once we have the local saving for logged out users built in, but consider this:

 * Log in as a user, create a project, add a scenario with some modifications, make it public
 * Log out, go to the same project. Since it is public, you should be able to see all scenarios and modifications
 * Make changes to the modifications
 * Log back in as the first user using the login button, without leaving this page
 * You are now logged in, and the state of the scenario in your browser is different from that on the server, but it is never auto-saved. If you refresh the page, any changes made before logging in are lost. However, if you make a single change after logging in, everything will be saved.

This last part isn't very obvious. Some ideas:

 * We could auto-save as soon as user logs in. But they would lose the state saved in the last "logged in" session.
 * We could prompt the user to either save the changes or reset to the last saved state, which would reload the entire project. But this would force the user to make a decision on the spot, and perhaps they'd like to keep both?
 * We could duplicate the changes into new scenarios and then auto-save, thus preserving both work saved previously, and work done while logged out. But if there were multiple changes to multiple scenarios, this could be busy and confusing.

We may need some design input, but I think we should have some visual indication of a "dirty" state, to make clear to the user when their changes have been saved and when they have not.

Auto-save can be a double-edged sword, and must be countered with some undo mechanism. We aren't saving history yet, perhaps we should look into ways of doing that.

lliss:
I don't think the undo mechanism is very important right now considering our feature list and our timeline. I'd even favour removing the save button entirely so that you can only auto save, which might help avoid confusion. We could probably handle the edge case of the user logging in with changes but again, I'm not sure that's a good use of time right now.

mmcfarland:
Good conversation.  There's been some talk about this, but the two immediate points I think are:

* Logging in should initiate a save if the current project `isNew`.
* Non owners should not have modification controls available for public projects (non owner could be an owner that is logged out).
* We can remove the save item from the project dropdown, that was temporary until this PR.

There's going to be a likely situation that the input (as opposed to modifications) will be made available to non owners to play with, but we don't want to persist those changes.  Having the `isNew` check on login auto-save will future proof that a little bit from happening.

There's also been some discussion about having a dirty state indicator for non-logged in users to show that the work in progress is not being saved.  I think we'll put that in eventually.  Undo we've been talking about, and I'm mostly just trying not to back ourselves into a place where that is really difficult, but I'm not ready to commit to it yet.

mmcfarland:
+1, I'll move some of my comments previously into new issues.

PR #349, opened 2015-06-10
Add data plane CloudFormation stack

hectcastro:
The data plane consists of an PostgreSQL RDS database and a Redis ElastiCache cache cluster. Each data store is accompanied by CloudWatch alarms, but only the PostgreSQL RDS database has been setup with a Route 53 private DNS record. The private DNS record for Redis will come with #339.

In addition, this pull request moves the bastion/monitoring server creation into the data plane stack. The security group for the bastion/monitoring server has been updated, and public/private Route 53 DNS records get created for it.

Connects #295
Connects #338

Also, tagging @lliss and @kdeloach.

---

This pull request takes a while to test because spinning up the RDS database and cache cluster is a long process. After it completes successfully, there should be a new PostgreSQL RDS database, Redis ElastiCache cache cluster, and a handful of new Route 53 DNS records.

Issue #339, opened 2015-06-09
Create CustomActionNode to create private DNS entry for cache


hectcastro:
This is necessary because the CloudFormation resource for Redis ElastiCache instances does not provide you with an common endpoint.

PR #350, opened 2015-06-10
Add CustomActionNode to create private DNS entry for cache

hectcastro:
CloudFormation can create ElastiCache cache clusters, but it cannot retrieve a specific cache cluster node's endpoint. To bridge the gap, a `CustomActionNode` comes after the data plane stack to scoop up the cache cluster ID, use it to determine the cache cluster node's endpoint, and set it as a private `CNAME` record.

Connects #339 and builds on top of #349. Also, tagging @kdeloach and @lliss.

---

Only the last commit in this pull request is relevant. After creating all of the stacks, there should be a new private DNS record for `cache.service.mmw.internal`.


lliss:
+1. Tested on AWS.

PR #351, opened 2015-06-10
Remove stray drawtools.

### LINKED ISSUE #314: Draw tools still visible after selecting area
To reproduce:

 * Go to [http://localhost:8000/analyze](http://localhost:8000/analyze) or [http://localhost:8000/model](http://localhost:8000/model)
 * You will be redirected to the home view [http://localhost:8000/](http://localhost:8000/)
 * Draw an area using "Draw". You will be moved forward to the analyze view.

Observed behavior:

 * Draw tools are still visible. Selecting a new area using these tools does not update the analyze panel:

image

Expected behavior:

 * Draw tools are hidden after area is selected

Other notes:

 * This does not happen if one starts the session at the home view, only if they start at the analyze or model view and then are redirected to home.
 * A fix might either hide the draw tools, or add code to always update the analyze view whenever new area is drawn, whichever is more appropriate.
 * If one moves forward from analyze to model, and then comes back, if the draw tools were visible before in the analyze view they still are. If they were not visible before then they still aren't.

jamesmcclain:
Modify the router's `.execute` method to make sure that when the prepare method uses `router.navigate` and returns `false`, that that route is not navigated to.  A test has been added for this case, also.

**To Test**
   * Follow the instructions given in the bug report https://github.com/WikiWatershed/model-my-watershed/issues/314
   * The draw tools and geocode search box should no longer stay after you have chosen an AoI.

Fixes #314

### DIFF
--- a/src/mmw/js/src/router.js
+++ b/src/mmw/js/src/router.js
@@ -73,21 +73,26 @@ var AppRouter = PatchedRouter.extend({
             });
         }
 
+        var self = this;
         this._sequence = this._sequence.then(function() {
             var result = prepare.apply(null, args);
             // Assume result is a promise if an object is returned.
             if (_.isObject(result)) {
+                self._previousRouteName = routeName;
                 return result.then(function() {
                     cb.apply(null, args);
                 });
-            // Only execute the route callback if prepare
-            // did not return false.
             } else if (result !== false) {
+                // Only execute the route callback if prepare
+                // did not return false.
+                self._previousRouteName = routeName;
                 cb.apply(null, args);
+            } else {
+                // Don't execute the route function
+                // and cancel the route transition
+                return false;
             }
         });
-
-        this._previousRouteName = routeName;
     },
 
     getSuffixMethod: function(context, suffix) {
--- a/src/mmw/js/src/core/tests.js
+++ b/src/mmw/js/src/core/tests.js
@@ -329,6 +329,27 @@ describe('Core', function() {
             assert.equal(spy.callCount, 0);
         });
 
+        it('does not navigated to the route associated with a prepare function that returns false', function() {
+            var controller = getController(),
+                router = new AppRouter(),
+                otherController = getOtherController(router),
+                spy = sinon.spy(controller, 'foo'),
+                otherSpy = sinon.spy(otherController, 'bazCleanUp');
+
+            router.addRoute(/^foo/, controller, 'foo');
+            router.addRoute(/^baz/, otherController, 'baz');
+
+            Backbone.history.start();
+
+            router.navigate('baz', { trigger: true });
+            router.navigate('foo', { trigger: true });
+
+            // if bazCleanup was never run then evidently the baz
+            // route was never navigated to.
+            assert.equal(spy.callCount, 1);
+            assert.equal(otherSpy.callCount, 0);
+        });
+
         it('does not execute route A\'s cleanUp function when navigating to route B if route A\'s prepare ' +
            'function returned false (i.e. route A was never navigated to)', function() {
 
@@ -387,6 +408,24 @@ function createTransitionRegionWithAnimatedHeightView(displayHeight, hiddenHeigh
         testParentView: testParentView,
         testSubView: testSubView
     };
+}
+
+function getOtherController(router) {
+    var controller = {
+        bazPrepare: function() {
+            console.log('bazPrepare');
+            router.navigate('bar', { trigger: true });
+            return false;
+        },
+        baz: function() {
+            console.log('baz');
+        },
+        bazCleanUp: function() {
+            console.log('bazCleanUp');
+        }
+    };
+
+    return controller;
 }
 
 function getController() {


rajadain:
+1! Great work! I had no idea it would be this few lines.

Tried landing on /analyze, on /model, both as logged in and logged out users, went back and forth through the workflow many times and in different combinations, works perfectly now.

rajadain:
To be explicit for the reader who hasn't tested this themselves: the draw tools are now removed when entering the /analyze view in all cases.

jamesmcclain:
Why thank you.  Your characterization of the problem was the clue that I needed.

caseypt:
@jamesmcclain If you don't mind, I'd like to take a look at this before you merge it.

mmcfarland:
Is a more simple solution available to simply route appropriately to the root view destination specified by `/`?  The router controllers for each "page" have some non-trivial setup and cleanup code that might not be activated under this situation, even though the UI reflects an appropriate state.

jamesmcclain:
@caseypt : Yes, that was expected.

caseypt:
I was going to suggest the same thing at @mmcfarland. When we were designing the routing architecture, we planned on the setup and cleanup of each route being isolated. For example, the analyze route doesn't use the `geocodeSearchRegion`, so it ideally shouldn't have to empty that region when a user leaves `/analyze`.

jamesmcclain:
I will need some discussion if I am going to address this as part of this card.

jamesmcclain:
This looks like it is going to be a very research-intensive card, unless someone has special insight.

The present code is apparently built on the assumption that when the router's `execute` method is overloaded so that in some circumstances it does not call the callback associated with a particular route (e.g. in [router.js](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/router.js#L64)), that that is equivalent to not following that route.  It seems that that assumption is not valid.  I base that on two things:

   1. The [backbone.js documentation](http://backbonejs.org/#Router-execute) says:

   > This method is called internally within the router, whenever a route matches and its corresponding callback is about to be executed. 

   Which seems to imply that the route has already been matched and that `execute` merely an opportunity to influence the callback.
   2. The behavior of the application itself, which is already strong evidence even if the first point in unconvincing.

caseypt:
I think the next line of the documentation may indicate what's wrong:

> Return false from execute to cancel the current transition.

We are not currently doing that. We return `false` from the `prepareRoute` function if we don't want to navigate to that route. We then [don't execute the callback](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/router.js#L85-L87) associated with that route.

We should also be returning `false` in this case and [not setting](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/router.js#L90) `_previousRouteName`.

I could be mistaken, but give that a shot.

jamesmcclain:
Yeah, I tried that.  Returning false from there will leave you on the `/model` page.  You are right that what I wrote above was incorrect in that I was trying to abbreviate and left out the fact that we are getting there via the call to `router.navigate`.

(As I remember it) you actually arrive in `execute` two different times: once when on the initial load when you visit `/model` with no AoI and once as a result of the call to `.navigate`.  Returning `false` in the second instance will leave you on the model page.

jamesmcclain:
@caseypt : I just amended my previous comment.  Thank you for the suggestion, I welcome all thoughts on this.

caseypt:
Here's a modified version of `router.execute` that I think fixes the problem. I'm not sure that I tested all the edge cases, so consider it a potential solution. https://gist.github.com/caseypt/88727541cbd2e510efcb. Let me know if you want to walk through this together.

I left some log statements in there that helped me debug the problem. I believe it was a combination of not returning `false` from execute when we didn't want to transition to a route, and always setting `previousRouteName` even if we didn't execute the given route.

We should add a test to prevent a regression. None of the routes in our [test controller](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/tests.js) are calling `router.navigate` in their prepare function. We should add a controller that does and a new test to ensure that when that happens, the route associated with that prepare function doesn't get navigated to. It would be pretty similar to [this test](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/tests.js#L314-L330).

lliss:
+1. Looks good. Tests pass. Functionality works as expected.

jamesmcclain:
Thanks for reviewing, I will merge.

src/mmw/js/src/router.js
```diff
@@ -73,21 +73,26 @@ var AppRouter = PatchedRouter.extend({
...
+                self._previousRouteName = routeName;
                 cb.apply(null, args);
+            } else {
+                // Don't execute the route function
+                // and cancel the route transition
+                return false;
             }
         });
-
-        this._previousRouteName = routeName;
```
kdeloach: This actually creates subtle queuing problems. If any route is navigated to, before the current route "prepare" method has completed (for example, if there is an animated transition), the sequence will be queued out of order. This could easily happen if a user switches between 2 routes quickly.

The problem isn't obvious because none of our controllers actually return promises (sadly). However, I don't think there's an urgency to fix this until the problem manifests itself.

Issue #257, opened 2015-06-01
Associate drawn modifcations with an SVG fill pattern


mmcfarland:
Each landscape modification will be associated with an SVG pattern.  See PSI for an example implementation.

https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Patterns

![screenshot from 2015-06-01 17 44 45](https://cloud.githubusercontent.com/assets/1014341/7924022/f66bf358-0885-11e5-9a9f-1e8a98819f5a.png)
![screenshot from 2015-06-01 17 45 08](https://cloud.githubusercontent.com/assets/1014341/7924023/f66c22ec-0885-11e5-92d1-8b100608e1af.png)

caseypt:
Check w/ @ctaylo37 to see if the patterns are done.

mmcfarland:
@ctaylo37 - this is a priority for us, let us know when you've got something.

ctaylo37:
Working on getting the SVG fill patterns for this.  Come see me when you are ready to work on this.  I may need to experiment a little to get this to turn out right.

ctaylo37:
Working on this now.  Just need to test all of the patterns.

ctaylo37:
Here is the file for the SVG's:
mb://FILESHARE.internal.azavea.com/projects/Stroud_ModelMyWatershed/documents/design/SVG Fill Patterns/DrawTools.html

and Here's the file for the fill patterns:
mb://FILESHARE.internal.azavea.com/projects/Stroud_ModelMyWatershed/documents/design/SVG Fill Patterns/DrawToolsFill.html

rajadain:
Leaflet does not natively support fill patterns. Currently looking in to a recent plugin that adds some functionality: https://github.com/teastman/Leaflet.pattern

caseypt:
@rajadain Here's how we do something similar on [Stormwater Interactive](http://water.phila.gov/swexp/): 

https://github.com/azavea/pwd-stormwater-interactive/blob/29164e2e8fed85c7bb0ab2b788ee6fbaa28ad81d/stormwater/nonres/js/src/bmpMap.js#L134

https://github.com/azavea/pwd-stormwater-interactive/blob/c162c970b72b98329d09b828fa60cc94f64311da/stormwater/nonres/templates/bmp_patterns.html

rajadain:
Ah thanks for the tip! I looked at the staging site yesterday but could only draw flat colors, not patterns, didn't know where exactly to look.

PR #352, opened 2015-06-10
Modification SVG Fill Patterns

rajadain:
Adds SVG patterns made by @ctaylo37 to the `/model` view. They look like this:

image

Land Cover Thumbnails:

image

Conservation Practice Thumbnails:

image

**Testing Instructions**

 * Draw an area of interest, move on to `/model` view
 * Add a scenario, add various Land Cover and Conservation Practice modifications

**Alternate Scenarios**

 * Try saving modifications and revisit page to ensure they show up correctly

**Notes**

While I'm happy with most of the work done here, the one issue which I need help with is finding the best strategy for pattern loading. Currently it has been added to the end of `base.html`, which means we have an addition of 616 KB to every full page load. Perhaps this is acceptable because all further interaction is done via AJAX calls. Or perhaps we should somehow dynamically load this when people land on the `/model` view, which would make initial load faster, but the transition and loading of this page a little slower. We may even consider adding loading animations to the modification drop-downs until the patterns are loaded.

Another attempt might be to minify `patterns.html`. I've removed all extraneous spacing, but it is still well-indented to be human readable. We could probably save some bandwidth if we got rid of all newlines and whitespace.

Input, there and elsewhere, is welcome.

Connects #257

### DIFF
--- a/src/mmw/apps/core/templates/base.html
+++ b/src/mmw/apps/core/templates/base.html
@@ -33,4 +33,6 @@
     <script type="text/javascript" src="{% static 'js/vendor.js' %}"></script>
     <script type="text/javascript" src="{% static 'js/main.js' %}"></script>
     {% endblock javascript %}
+
+    {% include 'patterns.html' %}
 </body>
--- a/src/mmw/apps/core/templates/patterns.html
+++ b/src/mmw/apps/core/templates/patterns.html
@@ -1,0 +1,3503 @@
+<svg>
+    <defs>
+        <!-- LAND COVER TYPES -->
+        <!-- Low Intensity Residential-->
+        <pattern height="72" id="fill-lir" overflow="visible" patternUnits="userSpaceOnUse" viewBox="79.59 -151.59 72 72" width="72">
+            <g>
+                <polygon fill="#329b9c" points="79.59,-79.59 151.59,-79.59 151.59,-151.59 79.59,-151.59"/>
+                <g>
+                    <polygon fill="none" points="79.59,-79.59 151.59,-79.59 151.59,-151.59 79.59,-151.59"/>
+                    <g>
+                        <line fill="none" stroke="#231F20" stroke-linejoin="round" stroke-width="0.3" x1="144.106" x2="162.106" y1="-162.105" y2="-144.105"/>
+                        <line fill="none" stroke="#231F20" stroke-linejoin="round" stroke-width="0.3" x1="144.106" x2="168.106" y1="-168.105" y2="-144.105"/>
+                        <line fill="none" stroke="#231F20" stroke-linejoin="round" stroke-width="0.3" x1="144.106" x2="174.106" y1="-174.105" y2="-144.105"/>
+                        <line fill="none" stroke="#231F20" stroke-linejoin="round" stroke-width="0.3" x1="144.106" x2="180.106" y1="-180.105" y2="-144.105"/>
+                        <line fill="none" stroke="#231F20" stroke-linejoin="round" stroke-width="0.3" x1="144.106" x2="186.106" y1="-186.105" y2="-144.105"/>
+                        <line fill="none" stroke="#231F20" stroke-linejoin="round" stroke-width="0.3" x1="144.106" x2="192.106" y1="-192.105" y2="-144.105"/>
+                        <line fill="none" stroke="#231F20" stroke-linejoin="round" stroke-width="0.3" x1="144.106" x2="198.106" y1="-198.105" y2="-144.105"/>
+                        <line fill="none" stroke="#231F20" stroke-linejoin="round" stroke-width="0.3" x1="144.106" x2="204.106" y1="-204.105" y2="-144.105"/>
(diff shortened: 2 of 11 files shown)


ctaylo37:
I think we should definitely minify the HTML file that gets pulled on to the server, but keep a working copy for changes that isn't loaded.

rajadain:
On a side note @ctaylo37, if you'd like to play with the colors / patterns, these are the files you need to adjust: [`patterns.js`](https://github.com/rajadain/model-my-watershed/blob/feature/svg-patterns/src/mmw/js/src/core/patterns.js) for colors, stroke width, opacity etc and [`patterns.html`](https://github.com/rajadain/model-my-watershed/blob/feature/svg-patterns/src/mmw/apps/core/templates/patterns.html) for adjusting the actual patterns. Please note that `patterns.html` has slightly modified `id`s to match what was defined in the app. Based on discussion with @mmcfarland, I imagine these `id`s might change once we have a definitive list, but perhaps that can be a separate PR.

caseypt:
You could try using django `spaceless` template filter in `patterns.html` to reduce the page load: https://docs.djangoproject.com/en/1.8/ref/templates/builtins/#spaceless

caseypt:
+1 Well done. I was able to draw some modifications and see the custom pattern. I saved the project and restored it, and the modifications were redrawn with the correct pattern.

If you can think of a way to test that everything is wired up correctly, consider adding it.

selection_051

rajadain:
@caseypt Good call on removing `randomDrawOpts`. I will make sure it isn't called elsewhere before cleaning up.

Regarding `spaceless`, that's a great find, thank you! But I'm afraid we would transfer the burden of bandwidth to server CPU which must now strip out the spaces for each full page load. What ideally I'd like to do is to treat the `patterns.html` file like a static asset, that can be cached. For example, one idea might be to minify it and put it in the `static` directory, and have the minification be a part of the build step. That way, the space stripping is only happening each time we build the project, instead of each time we serve a page.

Let me start with that, and if there are more ideas of reduction we will incorporate those as well.

This would allow us to use maybe slightly faster serving, but it still wouldn't be cacheable because it won't be an independent request from the browser. A client side delayed load would be ideal, because the user doesn't need the patterns immediately (unless they are landing on a scenario ...), and it would allow the browser to cache the request, or a CDN to quickly supply it, without entanglement with our server side template generation. I'm not sure what is the best way of achieving this, and am open to ideas, as well as thoughts on if it would be worth the effort.

caseypt:
@rajadain Good points. This may be overkill, but we could have a django route that returns the patterns HTML, request it via an AJAX call, [parse it](http://api.jquery.com/jQuery.parseHTML/), then load it into the DOM.

mmcfarland:
I really like the way you laid this out, good stuff.  The 600+kb payload is a little tough to swallow, and I agree with the conversation about it so far.  I think the fills look really good and we should work towards a manageable solution.  Doing a cachable async request is pretty good idea, but I'm curious to see what minified and gzipped payload translates to on production before we try to solve something.  Also a little more concerned with potential parsing time as opposed the xfer time, especially if it ends up being cacheable - would be good to get some numbers to evaluate.  For the time being, so long as it doesn't grind on load for our first milestone, I'm fine with pushing these discussions off to a new issue.

rajadain:
Actually after doing a number of tests, it seems like I was wrong about the 600 KB number. Even though that is the size of the file, what is transferred to the browser is actually under 60 KB, and I'm not exactly sure I understand why that is:

image

One possibility could be the encoding of the characters: the source might be in unicode whereas the browser transfers ASCII. Unsure.

Here's the rough averages I've been able to achieve:

Metric | Without `spaceless` | With `spaceless`
------- | ----------- | -----------
**Time** | 200ms | 210ms
**Size** | 58 KB | 56 KB

So we don't really gain much saving for the additional stripping, thus I've decided to forego the `spaceless` tag for now. I agree with both of you that we should only revisit this if it becomes an issue in the future, rather than overengineering for an unlikely use case now.

rajadain:
I've removed `randomDrawOpts` in the latest commit: https://github.com/WikiWatershed/model-my-watershed/commit/6c2ff6d8703502f9b8ff4f718977af6a69bcd9f6#diff-96ec3a491ec24e06f1414e91dd22d2cbL66

mmcfarland:
I think that small size transfer is the size after getting gzipped from the server, you can have the Network tab report both xfer size and uncompressed size.  I was imagining that an SVG file is mostly made up of repeated characters and would compress very well.  Let's just keep an eye on xfer times and deal with issues that come up.

rajadain:
Added tests to ensure mappings between IDs and patterns is correct, and that draw options are defined for all modifications.

rajadain:
Okay, finalized test additions and fixed a flake8 issue. I've never really written tests before, so would appreciate a quick review to see if they make sense / seem idiomatic.

caseypt:
The tests look good to me. They are a quick way to check that everything is wired up correctly.

kdeloach:
+1 Nice.

PR #354, opened 2015-06-11
Fixes #190

### LINKED ISSUE #190: Improve appearance of bar chart
The following needs to be done:

 * Use colors that are consistent with the app color scheme for the stacked bar chart.
 * Fix the space between the left-most bar and the y-axis (when in vertical bar mode) so that the y-axis label doesn't get occluded.
 * Make the order of the colors/labels in the legend match the order of the colors within the bars.
 * Make x-axis labels aligned with bars. (This used to work...)

ctaylo37:
This still needs some work.  I would like to pair with a dev to see how to customize more of this.

### DIFF
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -668,7 +668,7 @@ var BarChartView = Marionette.ItemView.extend({
                 }
             ],
             options = {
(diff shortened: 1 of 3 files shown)


caseypt:
@azaveaci rebuild

ctaylo37:
I think this is good for demo.

Issue #287, opened 2015-06-04
Setup a VPC endpoint for private subnet communication to S3


hectcastro:
By default, instances in the private subnet will have to route traffic through the NAT instances in order to communicate with S3. Setting up a [VPC endpoint](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-endpoints.html) allows the traffic to flow directly to S3 from private subnets.

PR #358, opened 2015-06-11
Add CustomActionNode to create a VPC endpoint for Amazon S3

hectcastro:
By default, instances in the private subnet have to route traffic through a NAT instance in a public subnet to communicate with Amazon S3. Setting up a VPC endpoint allows the traffic to flow directly to Amazon S3 from private subnets via a special route:

  https://aws.amazon.com/blogs/aws/new-vpc-endpoint-for-amazon-s3/

This is being done in a `CustomActionNode` because CloudFormation does not have support for creating VPC endpoints. Also, many of the additional Python classes are necessary because Boto does not have
direct support for describing and creating VPC endpoints.

Connects #287. Also, tagging @kdeloach and @lliss.

---

To test, go through the process of launching a stack and check the VPC > Endpoints page. There should be an entry associated with the `MMWVPC` and the default route table for `com.amazonaws.us-east-1.s3`.

### DIFF
--- a/deployment/cfn/stacks.py
+++ b/deployment/cfn/stacks.py
@@ -1,6 +1,7 @@
 from majorkirby import GlobalConfigNode
 
 from vpc import VPC
+from s3_vpc_endpoint import S3VPCEndpoint
 from private_hosted_zone import PrivateHostedZone
 
 import ConfigParser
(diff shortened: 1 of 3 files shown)


hectcastro:
@azavea-bot rebuild

deployment/cfn/vpc.py
```diff
@@ -107,7 +107,7 @@ def set_up_stack(self):
             'BastionHostAMI', Type='String', Description='Bastion host AMI'
         ), 'BastionHostAMI')
 
-        self.create_vpc()
+        public_route_table = self.create_vpc()
 
         # TODO: Setup VPC endpoint to bypass NAT when talking to S3
```
lliss: This TODO can probably go.

deployment/cfn/stacks.py
```diff
@@ -29,13 +30,18 @@ def build_graph(mmw_config, aws_profile, **kwargs):
...
     vpc = VPC(globalconfig=global_config, aws_profile=aws_profile)
+    s3_vpc_endpoint = S3VPCEndpoint(globalconfig=global_config, VPC=vpc,
+                                    aws_profile=aws_profile)
     private_hosted_zone = PrivateHostedZone(globalconfig=global_config,
-                                            VPC=vpc, aws_profile=aws_profile)
+                                            VPC=vpc,
+                                            aws_profile=aws_profile)
 
-    return vpc, private_hosted_zone
+    return s3_vpc_endpoint, private_hosted_zone
```
lliss: So we just weren't using the vpc result. Is that still to come.

deployment/cfn/vpc.py
hectcastro: Addressed in https://github.com/WikiWatershed/model-my-watershed/commit/26133d4c0b41e14cd90b390d674d404491c7d15d.

deployment/cfn/stacks.py
hectcastro: Since `s3_vpc_endpoint` depends on `vpc`, the graph resolves without it.

lliss:
+1. I created a VPC and confirmed the s3 endpoint was created along with the route table entry.

Issue #361, opened 2015-06-11
Add ability to resend activation email


rajadain:
Currently when a user signs up, they are sent an email with a confirmation code to activate their account. In case they don't receive that email, or loose it, they should be able to resend another confirmation code.

Issue #262, opened 2015-06-02
Signup tasks


mmcfarland:
* Activating a new account should not log you in immediately.

https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/base.py#L305

* Add a login button to the activation instruction modal
![screenshot from 2015-06-02 11 26 48](https://cloud.githubusercontent.com/assets/1014341/7939700/538051b2-091a-11e5-810e-4347699d078b.png)

Issue #259, opened 2015-06-01
inactive user login message


lewfish:
New users who haven't activated their accounts yet should see a special error message (eg. 'Please activate your account.') when attempting to login.

rajadain:
shares a point with #262

PR #362, opened 2015-06-11
Post-Sign Up Login

rajadain:
Adjusts Sign Up behavior so that users are not logged in automatically upon activation, and are directed to the Login form instead. If they try to log in before activation, they are informed that they need to activate their accounts.

**Testing Instructions**

 * Stop the app and run it manually so you see the console output:

```shell
vagrant ssh app -c 'sudo stop mmw-app'
./scripts/manage.sh runserver
```

 * Make sure you are not logged in. Click the Login button. Click Sign Up
 * Fill in the form and sign up as a new user.
 * Ensure that after signing up you see a dialog which instructs you to activate the account using link in email, and that you see a Login button.
 * Click the Login button. Attempt to log in as the new user.
 * Ensure that you see a message instructing you to activate your account first.
 * Follow the activation link in your "email" (from the console output)
 * Try logging in again after activation

**Notes**

This code assumes that any account which is not currently active has received a confirmation code. If an existing account is made inactive by a superuser, then they will receive the same message (to activate their account), whereas in this case they should see an "Invalid username or password" message. I'm not sure if such user deactivation is an expected use case for this app.

There is also no mechanism to resend the confirmation code. I've added #361 for that.

Connects #262 
Connects #259

### DIFF
--- a/src/mmw/apps/user/views.py
+++ b/src/mmw/apps/user/views.py
@@ -27,17 +27,24 @@ def login(request):
         user = authenticate(username=request.REQUEST.get('username'),
                             password=request.REQUEST.get('password'))
 
-        if user is not None and user.is_active:
-            auth_login(request, user)
-            response_data['result'] = 'success'
-            response_data['username'] = user.username
-            response_data['guest'] = False
-            response_data['id'] = user.id
+        if user is not None:
+            if user.is_active:
+                auth_login(request, user)
+                response_data['result'] = 'success'
+                response_data['username'] = user.username
+                response_data['guest'] = False
+                response_data['id'] = user.id
+            else:
+                response_data['errors'] = ['Please activate your account']
+                response_data['guest'] = True
+                response_data['id'] = 0
+                status_code = status.HTTP_400_BAD_REQUEST
         else:
             response_data['errors'] = ['Invalid username or password']
             response_data['guest'] = True
             response_data['id'] = 0
             status_code = status.HTTP_400_BAD_REQUEST
+
     elif request.method == 'GET':
         user = request.user
 
--- a/src/mmw/js/src/user/controllers.js
+++ b/src/mmw/js/src/user/controllers.js
@@ -7,6 +7,7 @@ var App = require('../app'),
 var SignUpController = {
     signUp: function() {
         new views.SignUpModalView({
+            app: App,
             model: new models.SignUpFormModel({})
         }).render();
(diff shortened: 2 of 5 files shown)


mmcfarland:
+1 Code looks good, not tested.

src/mmw/js/src/user/controllers.js
```diff
@@ -7,6 +7,7 @@ var App = require('../app'),
 var SignUpController = {
     signUp: function() {
         new views.SignUpModalView({
+            app: App,
```
caseypt: I did not notice when we started passing `App` as an argument. Is this to avoid circular references?

caseypt:
+1. Tested. Works as advertised.

src/mmw/js/src/user/controllers.js
rajadain: That's correct. `app.js` imports `user/views.js`so it can instantiate a `LoginModalView`: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/app.js#L58. Also, because of their inherent nature, all user views affect or manipulate the `App.user` state, thus need access to `App`. We pass it in as a dependency to avoid circular references.

Since `app.js` only uses `user/views.js` for one specific operation, if we can refactor that away then we can simply require `App` in `user/views.js`. One such way would be to make a `/login` route which opens the `LoginModalView`, and the `app.js` simply triggers that view. But that can be complicated if one is trying to login from the `/model` page and already has unsaved work in progress that they'd like to login to save: changing the URL at that point seems hairy.

If we can think of another way of removing this dependency, that would be great.

lewfish:
+1, tested

src/mmw/apps/user/views.py
```diff
@@ -27,17 +27,24 @@ def login(request):
         user = authenticate(username=request.REQUEST.get('username'),
                             password=request.REQUEST.get('password'))
 
-        if user is not None and user.is_active:
-            auth_login(request, user)
-            response_data['result'] = 'success'
-            response_data['username'] = user.username
-            response_data['guest'] = False
-            response_data['id'] = user.id
+        if user is not None:
```
kdeloach: Doesn't need to be changed now, but more Pythonic would be `if not user`.

rajadain: Good catch! Will keep that in mind

Issue #280, opened 2015-06-04
Active scenario tab should be a part of the URL


mmcfarland:
This makes sharing via URL or by `window.location.href` easier.  I'm thinking `/project/<id>/scenario/<slug-name>` and if no scenario is provided, current conditions are displayed active.

caseypt:
Use scenario Id instead of slug

Issue #356, opened 2015-06-11
Initial save on login


mmcfarland:
Logging in should initiate a save if the current project isNew and you're on the `/model` page.  The theory being you may have been logged out, created a project with scenarios and decided to save it.

mmcfarland:
Also remove the `save` menu item from projects.  Remember to remove the view events and the public/private menu list length tests.

rajadain:
We might consider keeping the Save button for non-logged in users, clicking which would open the Login modal.

mmcfarland:
Great idea.

rajadain:
Thinking about the use case when a user is logged in but is viewing a public project owned by someone else. Should the "Save" button be visible then? If so, does it make sense to give it a behavior that would "Fork" the public project into the user's account?

Hmm, thinking about it a little more I think that overloading the "Save" button with such multiple meanings is not a good idea. If we ever decided to support forks, that button should just say "Fork" or "Clone" instead of "Save".

For this use case I will stick with removing the "Save" button for all logged in users, and only showing it for anonymous users viewing unsaved projects.

mmcfarland:
Yeah, we've bounced some ideas around about this, but to keep things simple, any implicit/explicit "save" mechanism should only be available to project owners.

PR #363, opened 2015-06-11
Update the URL with active scenario

mmcfarland:
Allow the URL to be set with the active scenario and restore that
scenario to active status allows easier sharing of public links.

Testing:
* Non-saved projects should not generate unique url structures
* Saved projects should update the URL with the scenario ID when it's activated
* Full scenario URLs should select the correct tab on reload
* Bad scenario Ids redirect to the first scenario
* Adding a new scenario will update the URL with the correct ID when the ID is received from the save operation.

Connects #280

### DIFF
--- a/src/mmw/apps/home/views.py
+++ b/src/mmw/apps/home/views.py
@@ -31,7 +31,7 @@ def home_page(request):
     return render_to_response('home/home.html', csrf_token)
 
 
-def model(request, proj_id=None):
+def model(request, proj_id=None, scenario_id=None):
     """
     If proj_id was specified, check that the user owns
     the project or if the project is public.
--- a/src/mmw/js/src/modeling/tests.js
+++ b/src/mmw/js/src/modeling/tests.js
@@ -302,6 +302,34 @@ describe('Modeling', function() {
                             });
                 });
             });
+
+            describe('#getReferenceUrl', function() {
+                var base = '/model/';
+
+                it('generates no url fragment for unsaved projects', function() {
+                    var project = new models.ProjectModel({ id: null });
+                    assert.equal(project.getReferenceUrl(), base);
+                });
+
+                it('generates a project id fragment when a project is saved and no active scenario is set', function() {
+                    var project = new models.ProjectModel({ id: 42 });
+
+                    assert.equal(project.getReferenceUrl(), base + project.id);
+                });
+
+                it('generates a project & scenario id fragment when a project is saved and an active scenario is set', function() {
+                    var project = new models.ProjectModel({ id: 42 }),
+                        scenario = new models.ScenarioModel({ id: 23 }),
+                        coll = new models.ScenariosCollection([ scenario ]);
+
+                    project.set('scenarios', coll);
+
+                    coll.setActiveScenario(scenario);
+
+                    assert.equal(project.getReferenceUrl(), base + project.id +
+                                 '/scenario/' + scenario.id);
+                });
+            });
         });
 
         describe('ModificatioModel', function() {
--- a/src/mmw/js/src/modeling/tests.js
+++ b/src/mmw/js/src/modeling/tests.js
@@ -354,7 +354,7 @@ describe('Modeling', function() {
 
                     assert.isFalse(collection.at(2).get('active'));
 
-                    collection.setActiveScenario(collection.at(2).cid);
+                    collection.setActiveScenario(collection.at(2));
 
                     assert.isTrue(collection.at(2).get('active'));
                 });
@@ -363,7 +363,7 @@ describe('Modeling', function() {
                     var collection = getTestScenarioCollection();
 
                     collection.first().set('active', true);
-                    collection.setActiveScenario(collection.at(2).cid);
+                    collection.setActiveScenario(collection.at(2));
 
                     assert.isFalse(collection.at(0).get('active'));
                     assert.isFalse(collection.at(1).get('active'));
--- a/src/mmw/apps/home/urls.py
+++ b/src/mmw/apps/home/urls.py
@@ -12,6 +12,8 @@ urlpatterns = patterns(
     url(r'^$', home_page, name='home_page'),
     url(r'^model/$', model, name='model'),
     url(r'^model/(?P<proj_id>[0-9]+)/$', model, name='model'),
+    url(r'^model/(?P<proj_id>[0-9]+)/scenario/(?P<scenario_id>[0-9]+)/$',
+        model, name='model'),
     url(r'^analyze$', home_page, name='analyze'),
     url(r'^compare$', compare, name='compare'),
     url(r'^error', home_page, name='error'),
--- a/src/mmw/js/src/modeling/controllers.js
+++ b/src/mmw/js/src/modeling/controllers.js
@@ -16,7 +16,7 @@ var ModelingController = {
         }
     },
 
-    model: function(projectId) {
+    model: function(projectId, scenarioParam) {
         var project;
 
         if (projectId) {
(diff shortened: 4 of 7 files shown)


mmcfarland:
There's some issue with being logged out and making scenarios, then logging in and having this work, but I think it's more appropriate to address that in #356.

mmcfarland:
Cleaned up some typos and unused vars in the tests.

rajadain:
Landing on a pre-existing model does not update the URL. If I go to http://localhost:8000/model/7, the URL stays the same. If I click the active "Current Conditions" the URL becomes http://localhost:8000/model/7/scenario/20. It should have been that after loading since the "Current Conditions" scenario was active.

mmcfarland:
That's intentional, but I debated about it since it didn't have a practical difference (correct tab is selected).  I mostly thought it would be gentler not to update the URL if it wasn't necessary, and only do so when state changes, as the user might expect it could.  I thought that if someone shared a link with you, it's somewhat reassuring to see that same link in the URL.  Open to counter arguments if it seems inconsistent or has a side effect I wasn't aware of.

mmcfarland:
Added cb9626f to show default scenario id in url on project load.

rajadain:
Thanks for adding the auto-load current scenario to URL behavior! It works very well.

Sharing scenarios is still a little problematic. While sharing the Project and current Scenario shows the correct URL in the Share Modal, sharing an inactive Scenario shows the incorrect URL of the active Scenario, instead of the one being shared:

image

image

In this case, the URL should have been http://localhost:8000/model/7/scenario/**20**, not http://localhost:8000/model/7/scenario/**21**

rajadain:
Successfully tested the following:

 * Ran all tests and they passed
 * Non-saved projects do not affect the URL at all
 * Saved projects are always updated with the correct scenario
 * URLs with scenario specified correctly load the scenario if found, correctly default to Current Conditions if not found, and also correctly load Current Conditions if URL specifies only project.
 * Adding a new scenario to an unsaved project correctly updates the URL with the new project and scenario ID
 * Renaming an unsaved project correctly updates the URL with the new project and Current Conditions scenario ID

Failing use cases:

 * Not specifying a scenario results in a 404: http://localhost:8000/model/7/scenario/
 * Specifying a non-numeric scenario results in a 404: http://localhost:8000/model/7/scenario/xyz

If we forgive non-matching numeric scenario URLs and redirect them back to the default scenario, we might consider doing the same for these. Or we might not, since the error here is syntactic and not semantic. Would love to hear what others think.

+1 splendid work. I've only tested the behavior, and apart from the issues listed it works very well, and makes our app easier to understand for the user. Will do a quick pass through the code tomorrow morning.

mmcfarland:
Yeah, I considered what to do with invalid scenario fragments, and from a strict point of view it made sense that if the resource you asked for didn't exist, you should get a 404 - but probably we should be more generous than that.  The main use case I'm thinking of is that of a shared scenario that gets deleted, should that link be broken or should it just go to the default project?  The only reason to consider not going there is if there's some context like, "look at how badly this new plan affects the water quality", but the default scenario is another plan that affects the quality positively - seems like it could lead to confusion.  I suppose with named scenarios, the best thing to do is just go to the project if the scenario is bad and hope for the best.

mmcfarland:
re: context menu of inactive scenario, I commented before that I think clicking the context menu icon should activate that tab, and I still think that would be the best fix for this behavior.

caseypt:
My thoughts are that even if the someone visits a URL like `model/37/scen`, we should navigate to `model/37`. We still have enough information to get them pretty close to where they wanted to go. In this case, and cases like `model/37/scenario/{{ an invalid scenario id}}`, we could open a modal saying something like "The requested scenario isn't available, but we've loaded the requested project`.

src/mmw/js/src/modeling/tests.js
```diff
@@ -302,6 +302,34 @@ describe('Modeling', function() {
...
+                });
+
+                it('generates a project & scenario id fragment when a project is saved and no active scenario is set', function() {
+                    var project = new models.ProjectModel({ id: 42 }),
+                        scenario = new models.ScenarioModel({ id: 23 }),
+                        coll = new models.ScenariosCollection([ scenario ]);
+
+                    project.set('scenarios', coll);
+
+                    coll.setActiveScenario(scenario);
```
caseypt: I think this test is checking that the URL is correct if no scenario is active, but this line is setting an active scenario. Should the test name be updated or, should this line be removed?

caseypt:
If we aren't going to use the scenario slug in the url (which I think is the right choice, since we aren't using the project slug because we allow projects to have the same name), we should consider removing the slug related code.

src/mmw/js/src/modeling/tests.js
mmcfarland: No the description of the test is suffering from some copy and paste errors, it's testing correctly.  I'll update the name.

rajadain:
I'm not sure how I feel about `model/37/scen` going to an actual page. In my mind while the IDs are not to be trusted, routes such as `model` and `scenario` are sacred. For example, while this works: https://twitter.com/rajadain/status/609156936714592256, and even this kinda works https://twitter.com/rajadain/status/609 (it takes you to a different tweet by somebody else, with the right tweet ID), this does not work at all: https://twitter.com/rajadain/stat

But just because other sites don't do it, doesn't mean we can't.

mmcfarland:
re: slugs - I'm not touching it here, but I remember people spending some time on it.  Is it not used at all in the app anymore?

caseypt:
@rajadain Good points. I guess I just feel that the scenario is nice to have, but not required. We still know what project should be loaded. But I could go either way.

@mmcfarland I don't believe so. I think we switched to `cid` everywhere. I'll create an issue.

rajadain:
re: deleted scenarios, I think it depends on the amount of public discourse we imagine people to have around them. If we imagine them to be very engaging and have a lot of discussion, then if a scenario is deleted we should show that it is deleted instead of redirecting to home. For example, when tweets are deleted they show up as 404s https://twitter.com/RepDuckworth/status/599268147284287488, and there are entire sites dedicated to preserving them http://politwoops.sunlightfoundation.com/tweet/599268147284287488

A compromise between the two extremes – a 404 in your face or a silent redirect – would be a subtle [alert](http://getbootstrap.com/javascript/#alerts), visible after being redirected, that informs that they were redirected. If we like this idea I'll make an issue for it.

rajadain:
re: context menu of inactive scenario, added #378

src/mmw/apps/home/views.py
```diff
@@ -31,7 +31,7 @@ def home_page(request):
     return render_to_response('home/home.html', csrf_token)
 
 
-def model(request, proj_id=None):
+def model(request, proj_id=None, scenario_id=None):
```
rajadain: I'm glad flake8 doesn't complain about this even though it is never used.

rajadain:
+1 code looks good!

mmcfarland:
Good discussion but I think we're venturing too far outside the scope of this issue, which is just setting the active scenario in the URL.  I'm going to leave as is (for URL resolving) and we can use log data to actually see if there is a problem with invalid scenario URLs being used often - may not be worth it to think this hard about it now.

caseypt:
Sorry, one more thought I just had while riding home. In the future, we may want to consider structuring the URL as something like: `/model/34#scenario-12`. We always will load the project, but if you provide an anchor tag or URL parameter with a scenario, we'll make that scenario the active tab.

rajadain:
I just added #379 for the redirect alert idea. It's really low priority, and we might decide to close it in the future, but this way it's recorded.

Issue #340, opened 2015-06-09
Setup custom e-mail backend for sending via SES


hectcastro:
A custom e-mail backend needs to be used within EC2 to send messages with Amazon SES via Boto. Most of this work was already done for the Treescount project.

PR #368, opened 2015-06-11
Add boto-ses email sender.

lliss:
Connects #340 

Testing this is difficult.

1. Modify the requirements file development.txt and add boto: ```boto==2.34.0```
2. Re-provision the app or otherwise install boto on the app vm.
3. On the app vm escalate to root and cd to `/etc/mmw.d/env`
4. Create two new files `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`
5. Add proper credentials (ask me in OTR or in person).
6. Change the email backend in `mmw/settings/development.py` to code show below.
7. You can use my email (as the default or add your own to the SES authorized sender list)
8. Use the forgot password form to retrieve your email.
9. Watch inbox in anticipation.
10. Read email and delight.

development.py changes:
```
DEFAULT_FROM_EMAIL = 'SOME-AUTHORIZED-EMAIL'
EMAIL_BACKEND = 'apps.core.mail.backends.boto_ses_mailer.EmailBackend'
```

### DIFF
--- a/src/mmw/apps/core/mail/backends/boto_ses_mailer.py
+++ b/src/mmw/apps/core/mail/backends/boto_ses_mailer.py
@@ -1,0 +1,110 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import division
+
+import json
+import logging
+import sys
+import traceback
+import boto.ses
+
+from django.conf import settings
+from django.core.mail.backends.base import BaseEmailBackend
+
+from django_statsd.clients import statsd
+from email.mime.text import MIMEText
+
+logger = logging.getLogger(__name__)
+
+
+class EmailBackend(BaseEmailBackend):
+
+    def __init__(self, fail_silently=False, aws_region='us-east-1', *args,
+                 **kwargs):
+        super(EmailBackend, self).__init__(fail_silently=fail_silently)
+        self.mailer = BotoMailer(aws_region)
+        self.check_quota = getattr(settings, 'EMAIL_BOTO_CHECK_QUOTA', False)
+
+    def send_messages(self, email_messages):
+        if not email_messages:
+            return
+        sent_message_count = 0
+
+        if self.check_quota:
+            remaining_quota = self.mailer.get_remaining_message_quota()
+            self._log_quota(remaining_quota)
+        else:
+            remaining_quota = sys.maxint
+
+        if len(email_messages) <= remaining_quota:
+            for email_message in email_messages:
+                if self._send_message(email_message):
+                    sent_message_count += 1
+            return sent_message_count
+        else:
+            raise SESQuotaException("Attempted to send %d messages with only "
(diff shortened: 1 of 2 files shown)


lliss:
Note. This code lifted from NYC trees almost as is. I merged the two files into one because it seemed cleaner for the time being.

mmcfarland:
Why not have the requirements file updated with `boto` for this commit?  I believe it would break the build if checked in, right?  If you're worried about removing the ability for devs to have a console email, you could set it up to provision with this backend in production only, and we could provision for the env to test.  Also, we can set those env variable on the cmd line as the app is provisioning or upping, similar to the ITSI stuff.

lliss:
We already have boto in prod requirements.
https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/requirements/production.txt#L3

src/mmw/apps/core/mail/backends/boto_ses_mailer.py
```diff
@@ -0,0 +1,110 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class EmailBackend(BaseEmailBackend):
+
+    def __init__(self, fail_silently=False, aws_region='us-east-1', *args,
+                 **kwargs):
+        super(EmailBackend, self).__init__(fail_silently=fail_silently)
+        self.mailer = BotoMailer(aws_region)
```
hectcastro: Since we're already depending on [instance metadata](http://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/ec2-instance-metadata.html#instancedata-data-categories) for authentication credentials, we can also pull the region out by trimming the availability zone.

src/mmw/mmw/settings/production.py
```diff
@@ -34,6 +34,7 @@
 # EMAIL CONFIGURATION
 EMAIL_BACKEND = 'apps.core.mail.backends.boto.EmailBackend'
 EMAIL_BOTO_CHECK_QUOTA = False
+DEFAULT_FROM_EMAIL = 'ADD-REAL-ADDRESS-HERE'
```
hectcastro: We can default this to something like `noreply@mmw.azavea.com` for now. The `mmw.azavea.com` domain will be useful in testing the Amazon SES DKIM setup.

lliss: Good call. Fixed and rebased.

src/mmw/apps/core/mail/backends/boto_ses_mailer.py
hectcastro: The Boto function to get this is [`get_instance_metadata `](https://github.com/boto/boto/blob/2.38.0/boto/utils.py#L379), but it'll be hard to parse the dictionary that comes back without testing against an EC2 instance with real metadata.

caseypt:
+1. Code looks good. I was able to set everything up and get an email when I filled out the lost password form: 

selection_057

kdeloach:
+1

PR #376, opened 2015-06-12
Do not return user's email address.

### LINKED ISSUE #344: Don't give out the project creator's email address.
Currently when fetching a saved project, we return the email address of the user as part of the user object. We should strip this out and only return the id.

jamesmcclain:
The `email` field has been removed from that serializer.

Connects #344

### DIFF
--- a/src/mmw/apps/user/serializers.py
+++ b/src/mmw/apps/user/serializers.py
@@ -9,4 +9,4 @@ from rest_framework import serializers
 class UserSerializer(serializers.ModelSerializer):
     class Meta:
         model = User
-        fields = ('id', 'username', 'email', 'is_staff')
+        fields = ('id', 'username', 'is_staff')


rajadain:
Instead of taking out the `email` at the endpoint, consider taking it out in the serializers themselves [here](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/serializers.py#L36) and [here](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/serializers.py#L46-47). I can't think of any use case where we would need the user's email when interacting with Projects.

mmcfarland:
I'm pretty sure you can just remove it as a serializer field here https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/user/serializers.py#L12

rajadain:
If we're sure we don't want the `email` field in most of the cases, I agree with @mmcfarland and it should be taken out at the source. For when we do want it, we can add it individually on a case-by-case basis.

rajadain:
+1 tested loading an existing project, did not get user's email:

image

All tests pass and nothing seems broken.

jamesmcclain:
Thanks for testing, I appreciate it.

lliss:
+1

jamesmcclain:
Is this okay to merge?

lliss:
I think if we are being formal about workflow, we should technically wait until @caseypt gives his okay or passes it off to someone else to do.

jamesmcclain:
Okay, will-do.

caseypt:
Sorry for the delay! +1

jamesmcclain:
No problem, thank you for reviewing.

Issue #298, opened 2015-06-04
Create application server stack


hectcastro:
Using the application server AMI, create an application server stack with a load balancer and auto scaling group.

PR #380, opened 2015-06-12
Add application CloudFormation stack

hectcastro:
The application servers live within an Auto Scaling Group (ASG) that is associated with an Elastic Load Balancer (ELB). Health checks executed by the ELB influence when the ASG adds or removes nodes from itself. All application servers live inside private subnets and are exposed to the internet via the ELB. In addition, SSL is terminated at the ELB.

Connects #298 and builds on top of #349 and #350. Also, tagging @lliss and @kdeloach.

---

Only the last two commits in the pull request are relevant. To test, bring up the stack and determine if all of the resources included in the template are created. As of right now, the application server AMI doesn't appear to be coming up cleanly. I haven't started debugging that yet.


deployment/cfn/application.py
```diff
@@ -0,0 +1,363 @@
...
+                SecurityGroups=[Ref(app_server_security_group)]
+            ))
+
+        app_server_auto_scaling_group_name = 'asgAppServer'
+
+        self.add_resource(
+            asg.AutoScalingGroup(
+                app_server_auto_scaling_group_name,
+                AvailabilityZones=Ref(self.availability_zones),
+                Cooldown=300,
```
lliss: In practice has this proven to be a reasonable time for the cooldown?

hectcastro: I think so. There is usually room for tuning based on application type, but this is a safe default (it is also the default Amazon uses if you don't provide a `Cooldown`).

deployment/cfn/application.py
```diff
@@ -0,0 +1,363 @@
...
+                             'load balancer',
+            VpcId=Ref(self.vpc_id),
+            SecurityGroupIngress=[
+                ec2.SecurityGroupRule(
+                    IpProtocol='tcp', CidrIp=ALLOW_ALL_CIDR, FromPort=p,
+                    ToPort=p
+                )
+                for p in [HTTP, HTTPS]
+            ],
+            SecurityGroupEgress=[
```
lliss: So ELBs aren't based on AMIs right. So we don't need this to access the outside world via 80 or 443 because it will never need to do updates or the like, right?

deployment/cfn/application.py
```diff
@@ -0,0 +1,363 @@
...
+                ec2.SecurityGroupRule(
+                    IpProtocol='udp', CidrIp=VPC_CIDR, FromPort=p, ToPort=p
+                )
+                for p in [STATSITE]
+            ] + [
+                ec2.SecurityGroupRule(
+                    IpProtocol='tcp', CidrIp=ALLOW_ALL_CIDR, FromPort=p,
+                    ToPort=p
+                )
+                for p in [HTTP]
```
lliss: If we're going to allow the app server full access to the outside on 80 we should probably add 443 since some things might be enforcing ssl everywhere.

hectcastro: You're right. I actually caught this while working on the tiler stack and accidentally committed in https://github.com/WikiWatershed/model-my-watershed/commit/2cc6d90ff2c890b83ec900ef252084062209d9b5.

deployment/cfn/application.py
hectcastro: Under the covers somewhere they might, but none of that is exposed to us. The reason why HTTP is enabled for egress here is because ELBs send health checks to the application servers via HTTP. For ingress, HTTP and HTTPS are enabled because we accept HTTP traffic, but then redirect users to the same URL over HTTPS.

lliss:
I'd support merging this in and then trying to debug why the AMI fails to properly instantiate. When I tried it on Friday with an app AMI that I made it seemed to look right.

Issue #374, opened 2015-06-12
ModalBaseView.handleServerError does not handle 500 errors properly


caseypt:
While testing out #368, I got the following response from the server:

![selection_054](https://cloud.githubusercontent.com/assets/1042475/8131485/8cf998a6-10e8-11e5-82f9-d8cf0b9ab925.png)

`ModalBaseView.handleServerError` throws an error when this response comes back. Here's the stack trace:

![selection_055](https://cloud.githubusercontent.com/assets/1042475/8131499/aa861c8c-10e8-11e5-8e81-4382843efd20.png)

And the line it's referencing:

![selection_056](https://cloud.githubusercontent.com/assets/1042475/8131512/b92bbda0-10e8-11e5-97b8-e2bb3f2dd378.png)

rajadain:
Ah I should've better tested that. I thought that `response.responseJSON.errors` would be falsy if any one of them was undefined, but it causes an error instead. We should probably replace that with:

```javascript
if (response && response.responseJSON && response.responseJSON.errors) {
    server_errors = response.responseJSON.errors;
}
```

PR #382, opened 2015-06-12
Handle cases when `response` or `responseJSON` is undefined

rajadain:
Also add a test to exercise that bit of code. In the test we simulate an error 500, but what is more relevant is the empty response string.

Connects #374


lliss:
+1

caseypt:
+1. Tests pass.

Issue #297, opened 2015-06-04
Create tile server stack


hectcastro:
Using the tile server AMI, create a tile server stack with a load balancer and auto scaling group.

PR #383, opened 2015-06-12
Add tiler CloudFormation stack

hectcastro:
The tile servers live within an Auto Scaling Group (ASG) that is associated with an Elastic Load Balancer (ELB). Health checks executed by the ELB influence when the ASG adds or removes nodes from itself. All tile servers live inside private subnets and are exposed to the internet via the ELB.

Connects #297 and builds on top of #349, #350, #380.  Also, tagging @kdeloach and @lliss.


lliss:
Tile server was allowing incoming port 80 connections from anywhere.

kdeloach:
+1

lliss:
+1

deployment/ansible/roles/model-my-watershed.tiler/tasks/app.yml
```diff
@@ -1,15 +1,18 @@
...
         owner="{{ ansible_ssh_user }}"
         group=mmw
         mode=0755
         state=directory
+  with_items:
+    - "{{ tiler_home }}"
+    - "/opt/model-my-watershed"
+
+- name: Clone application source
+  git: repo=https://github.com/azavea/model-my-watershed.git
```
lliss: Repo URL should be https://github.com/WikiWatershed/model-my-watershed.git no?

deployment/packer/template.js
```diff
@@ -48,6 +48,7 @@
                 "Service": "Tiler",
                 "Environment": "{{user `stack_type`}}"
             },
+            "subnet_id": "subnet-3781f46e",
```
lliss: Since we hard coded this, it might be nice to have a comment noting such to avoid confusion for new devs in the future.

hectcastro: JSON won't let you create inline comments.

scripts/aws/setupdb.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -e
+set -x
+
+# Export settings required to run psql non-interactively
+export PGHOST=$(cat /etc/mmw.d/env/MMW_DB_HOST)
+export PGDATABASE=$(cat /etc/mmw.d/env/MMW_DB_NAME)
+export PGUSER=$(cat /etc/mmw.d/env/MMW_DB_USER)
+export PGPASSWORD=$(cat /etc/mmw.d/env/MMW_DB_PASSWORD)
```
kdeloach: Kind of begs the question... why not just use these env var names instead? It would be nice to be able to run `envdir /etc/mmw.d/env psql`.

hectcastro: I had this thought before too. I ended up siding with consist naming in the application configuration vs. conforming to the `libpq` standard.

It is nice for PostgreSQL, but then other things like Redis don't have good names.

deployment/packer/template.js
lliss: Right. Didn't realize it was pure json since the file has a normal js extension.

scripts/aws/setupdb.sh
kdeloach: Yeah, good point.

Issue #296, opened 2015-06-04
Create worker stack


hectcastro:
Using the worker AMI, create a worker stack with an auto scaling group.

PR #384, opened 2015-06-12
Add worker CloudFormation stack

hectcastro:
The workers live within an Auto Scaling Group (ASG) that is currently using Amazon's base EC2 health check type. All workers live inside private subnets and generally only communicate with the outside world through messages passed via Redis.

Connects #296 and builds on top of #349, #350, #380, #383.


deployment/cfn/vpc.py
```diff
@@ -47,11 +43,8 @@ class VPC(StackNode):
         'Region': 'us-east-1',
         'StackType': 'Staging',
         'KeyName': 'mmw-stg',
-        'IPAccess': '0.0.0.0',
+        'AvailabilityZones': ['us-east-1b', 'us-east-1d'],
```
kdeloach: This is a stupid question, but why default to availability zones b/d instead of a/b?

kdeloach:
+1 Looks solid.

deployment/cfn/vpc.py
hectcastro: Completely arbitrary.

lliss:
+1

PR #385, opened 2015-06-12
Autosave project after login

### LINKED ISSUE #260: saving for guest users
I'm not sure I completely understand this, but quoting @mmcfarland in https://github.com/WikiWatershed/model-my-watershed/pull/225

"For redirecting, we're supporting a workflow for anonymous users who have work in progress and decide to register at that point. We should leave the register window completion modal up with a button for login available (or flip directly to the login) and in the activation tab, there should be instructions similar to "if you had work in progress, visit the previous tab to continue and save your work". Activating an account should not log you in."

### LINKED ISSUE #375: Clicking sign-up on the model page loses project, but keeps AoI
Not sure if this is the correct behavior. I think, ideally, you could be on `/model`, sign-up, activate, and then login without ever leaving `/model`. This way any work that was started is not loss. This may be tricky because the sign-up modal changes the URL to `/sign-up`.

To reproduce:

- Visit the app as an anonymous user.
- Navigate to the modeling page.
- Click login then sign up.

rajadain:
If an anonymous user has created a project that they would like to save, they must first login. This PR autosaves the project as soon as they do.

**Testing Instructions**

 * Run the app from the console to facilitate sign-up `./scripts/manage.sh runserver`
 * Make sure you're logged out. Draw a shape and create a project, add scenarios and modifications.
 * In the Project Menu, click "Save"
 * Click "Sign Up"
 * Enter details and click "Sign Up"
 * Activate the account with the activation link. Ensure you see a friendlier message instructing you to revisit the old tab in case you have unsaved work in progress (you do)
 * Go back to the original tab and click "Login"
 * Login as the new user. As soon as you do, the entire project should be saved.

**Alternate Scenarios**

 * Try creating a new project as a logged in user. Do you see the "Save" button in the Project Menu? Do you see it after the project has been saved? See e44c74e7e726282c1867e9a71696c9a3362929bf for an explanation.
 * Try creating a new project as an anonymous user. Instead of clicking "Save" in the Project Menu, click "Login" in the app header and login as an existing user. Was your project saved?
 * Try creating a new project as an anonymous user, and click "Login". Then dismiss the Login modal. Make some more changes to the project. Then login. Is the project saved correctly?

**Notes**

This does not handle the case for ITSI users, which will require a little more thought. Here's an initial outline of how that might work:

 * When a user clicks "Continue as ITSI", instead of doing a GET /user/itsi/login, do a POST and send current project data to the server. The server should serialize it and store it as a session variable.
 * When the user finally signs up after the round-trip to the ITSI portal, the server should send the project back. It will be the right one because it is associated with this user's session.
 * The `ItsiSignUpModalView.handleSuccess` method should take that project from the response, and recreate it (objects, routes, etc) on the front-end. Then it should log the user in, which will trigger the autosave implemented in c8d4ffea8d4cfdec78545e1688ac283b063c2a3b.

Connects #356 
Connects #375
Connects #260

### DIFF
--- a/src/mmw/js/src/modeling/tests.js
+++ b/src/mmw/js/src/modeling/tests.js
@@ -180,11 +180,19 @@ describe('Modeling', function() {
         describe('ProjectMenuView', function() {
             it('displays a limited list of options in the drop down menu until the project is saved', function() {
                 var project = getTestProject(),
-                    view = new views.ProjectMenuView({ model: project });
+                    view = new views.ProjectMenuView({ model: project }),
+                    preSaveMenuItems = [
+                        'Rename',
+                        'Save',
+                        'Print'
+                    ];
 
                 $('#sandbox').html(view.render().el);
 
                 assert.equal($('#sandbox li').length, 3);
+                $('#sandbox li').each(function() {
+                    assert.include(preSaveMenuItems, $(this).text());
+                });
             });
 
             it('displays all the options in the drop down menu if the project is saved', function() {
@@ -192,7 +200,15 @@ describe('Modeling', function() {
                 App.user.set('id', 1);
                 var project = getTestProject(),
                     view = new views.ProjectMenuView({ model: project }),
-                    projectResponse = '{"id":21,"user":{"id":1,"username":"test","email":"test@azavea.com"},"scenarios":[],"name":"Test Project","area_of_interest":{},"is_private":true,"model_package":"tr-55","created_at":"2015-06-03T20:09:11.988948Z","modified_at":"2015-06-03T20:09:11.988988Z"}';
+                    projectResponse = '{"id":21,"user":{"id":1,"username":"test","email":"test@azavea.com"},"scenarios":[],"name":"Test Project","area_of_interest":{},"is_private":true,"model_package":"tr-55","created_at":"2015-06-03T20:09:11.988948Z","modified_at":"2015-06-03T20:09:11.988988Z"}',
+                    postSaveMenuItems = [
+                        'Share',
+                        'Make Public',
+                        'Delete',
+                        'Add Tags',
+                        'Rename',
+                        'Print'
+                    ];
 
                 this.server.respondWith('POST', '/api/modeling/projects/',
                             [ 200, { 'Content-Type': 'application/json' }, projectResponse ]);
@@ -201,7 +217,10 @@ describe('Modeling', function() {
 
                 $('#sandbox').html(view.render().el);
 
-                assert.equal($('#sandbox li').length, 7);
+                assert.equal($('#sandbox li').length, 6);
+                $('#sandbox li').each(function() {
+                    assert.include(postSaveMenuItems, $(this).text());
+                });
             });
 
             it('displays limited options in the user does not own the project', function() {
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -43,6 +43,7 @@ var ModelingHeaderView = Marionette.LayoutView.extend({
         // If the user changes, we should refresh all the views because they
         // have user contextual controls.
         this.listenTo(App.user, 'change', this.reRender);
+        this.listenTo(App.user, 'change:guest', this.saveAfterLogin);
     },
 
     reRender: function() {
@@ -65,6 +66,17 @@ var ModelingHeaderView = Marionette.LayoutView.extend({
 
     onShow: function() {
         this.reRender();
+    },
+
+    saveAfterLogin: function(user, guest) {
+        if (!guest && this.model.isNew()) {
+            var user_id = user.get('id');
+            this.model.set('user_id', user_id);
+            this.model.get('scenarios').each(function(scenario) {
+                scenario.set('user_id', user_id);
+            });
+            this.model.saveProjectAndScenarios();
+        }
     }
 });
 
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -85,7 +85,7 @@ var ProjectMenuView = Marionette.ItemView.extend({
         'click @ui.remove': 'deleteProject',
         'click @ui.share': 'shareProject',
         'click @ui.print': 'printProject',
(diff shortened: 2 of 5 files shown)


rajadain:
Whoops forgot to adjust the tests. Hold on a bit ...

caseypt:
Just a quick heads up that we are using "Connects" or "Refs" when referencing issues so that they are not auto-closed (like "Fix" and "Closes"). This way we can manually move the issue card to the done column when we have verified that changes work in staging.

rajadain:
Ah! Thank you, will update the commit messages and PR description to follow these guidelines.

rajadain:
Fixed tests, added more specific options delineating which options should be visible in the menu pre- and post-save.

src/mmw/js/src/modeling/views.js
```diff
@@ -66,6 +67,17 @@ var ModelingHeaderView = Marionette.LayoutView.extend({
 
     onShow: function() {
         this.reRender();
+    },
+
+    saveAfterLogin: function(user, guest) {
+        if (!guest && this.model.isNew()) {
+            var user_id = user.get('id');
+            this.model.set('user_id', user_id);
```
kdeloach: How important is it to set user ID when saving models? Hopefully, we are pulling the user ID from the current session (server-side) when saving these, and not relying on POST data.

rajadain: The `user_id` seems to be required as of #348 to correctly save projects and scenarios, otherwise they are not associated with each other. But I agree, either we are overusing the `user_id` field here or we are not handling it well enough on the server side. I didn't give it much thought at the time, but will investigate this tomorrow morning.

rajadain: After doing some more investigation, I find that [modeling/models.js#L66-68](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/models.js#L66-68) checks the `user_id` before saving to ensure that users can save only their own projects, and not another user's public project. For a logged in user creating a new project, the `user_id` is set during initialization in [modeling/models.js#L53](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/models.js#L53). However, for projects created before a user is logged in, that field defaults to `0`. After logging in, the `App.user.id` is different from `model.get('user_id')` and the save would not happen, thus necessitating this bit of code.

I believe this check for `user_id` before saving reduces erroneous AJAX requests. I don't understand the code on the server side completely due to my unfamiliarity with Django Rest Framework, but I hope that [modeling/views.py#L47](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/views.py#L47) checks against the user on the server side as well.

mmcfarland: The POST there in https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/views.py#L47 will be for creating Projects, and since the view is decorated with an `Authenticated` class, it will always be created in that user's context.

For project updates, this bit of code on the backend is preventing cross-user updates:
https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/views.py#L68

I believe there is some pretty good test coverage on this functionality, too.

src/mmw/js/src/modeling/tests.js
```diff
@@ -187,7 +195,15 @@ describe('Modeling', function() {
                 App.user.set('id', 1);
                 var project = getTestProject(),
                     view = new views.ProjectMenuView({ model: project }),
-                    projectResponse = '{"id":21,"user":{"id":1,"username":"test","email":"test@azavea.com"},"scenarios":[],"name":"Test Project","area_of_interest":{},"is_private":true,"model_package":"tr-55","created_at":"2015-06-03T20:09:11.988948Z","modified_at":"2015-06-03T20:09:11.988988Z"}';
+                    projectResponse = '{"id":21,"user":{"id":1,"username":"test","email":"test@azavea.com"},"scenarios":[],"name":"Test Project","area_of_interest":{},"is_private":true,"model_package":"tr-55","created_at":"2015-06-03T20:09:11.988948Z","modified_at":"2015-06-03T20:09:11.988988Z"}',
+                    postSaveMenuItems = [
+                        'Share',
+                        'Make Public', 'Make Private',
```
lliss: We've set the privacy to private in the faked response so we know that the string should be "Make Public". Why allow the alternate case. If that showed up that would actually be an error.

rajadain: Fair enough, will fix

src/mmw/templates/registration/activation_complete.html
```diff
@@ -7,7 +7,7 @@
 <p>
     {% trans "Your account is now activated." %}
     {% if not user.is_authenticated %}
-    {% trans "You can log in." %}
+    {% trans "If you had work in progress, please switch back to that tab and login to save your work." %}
```
lliss: I'd love to do something like stash this in local storage and then rebuild the page instead of expecting users not to mess up their tabs. It seems suboptimal to me that I can't complete any process from one window. Not an issue immediately but I think we should consider dealing with this in a better way later. As it stands this feels like the bank telling me not to use the back button.

rajadain: I agree it is clunky, and I agree this is a big enough issue to be dealt with separately. The activation link opens in a new tab. If we recreate the the original state and save it, what happens to the original tab? Do we send a signal from the server indicating that the project has been updated elsewhere? We don't have the WebSocket architecture in place for something like that. Or we could tell them to only use the new tab and close the current one as soon as they sign up. But in case they want to continue working for a while before activating their account, they will no longer be able to do so.

Created #399 for this.

src/mmw/js/src/modeling/tests.js
rajadain: Fixed in 6d6fb75

mmcfarland:
Sorry, this one slipped by me - testing now.

mmcfarland:
+1 Tested.

rajadain:
Rebased on top of `develop`, consolidated commits.

Issue #369, opened 2015-06-11
Modifications drop down shows IDs instead of Names


rajadain:
We switched to using internal IDs as part of #352 6c2ff6d8703502f9b8ff4f718977af6a69bcd9f6. As a side-effect, the modifications drop down now shows the internal IDs instead of names:

![image](https://cloud.githubusercontent.com/assets/1430060/8118929/fff68794-105f-11e5-9865-5969d495386d.png)

Figure out a way to store both Names and IDs in the model, which currently only saves one, and use the new Name field in all user-facing displays, including any devised for #252.

kdeloach:
One way I solved this was to hard-code the names in the template (See https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/templates/scenarioToolbarTabContent.html#L21). We should also consider adding a label field on `ModificationModel`.

rajadain:
Currently, in the Control thumbnail macro definition https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/templates/controls/utils.html, we use the `value` parameter to populate the `data-value` attribute. But the macro also has access to a `name` parameter which is the pretty name and is used in the thumbnail label. We should be able to easily add a `data-label` attribute to the control and populate that with the `name` parameter, and just pick it up along with `data-value` when initializing modifications.

```
        <div class="thumb" data-value="{{ value }}" data-label="{{ name }}">
            ...
        </div>
        ...
```

We might also consider renaming the `name` parameter to `label` since modification names currently refer to Land Use, Conservation Practice, or Precipitation.

jamesmcclain:
Okay, thank you to everyone for the thoughts.

Issue #252, opened 2015-06-01
Add popup to modification features


caseypt:
![selection_039](https://cloud.githubusercontent.com/assets/1042475/7918442/6972f10c-085f-11e5-97a9-053844c1272e.png)

PR #391, opened 2015-06-15
Human-readable labels in modifications dropdown.

jamesmcclain:
A helper function has been created to map non-human-readable names to human-readable ones.

**To Test**
   * Choose an area of interest
   * Create a new scenario and draw some land use and/or conservation polygons
   * Click on the "modifications" button
   * You should now see human-readable labels
   * If you save and restore, you should still see human-readable labels

Connects WikiWatershed/model-my-watershed#369

### DIFF
--- a/src/mmw/js/src/modeling/controls.js
+++ b/src/mmw/js/src/modeling/controls.js
@@ -62,6 +62,12 @@ var DrawControlView = ControlView.extend({
 var LandCoverView = DrawControlView.extend({
     template: landCoverTmpl,
 
+    templateHelpers: function() {
+        return {
+            label: models.getHumanReadableLabel
+        };
+    },
+
     getControlName: function() {
         return 'landcover';
     }
@@ -69,6 +75,12 @@ var LandCoverView = DrawControlView.extend({
 
 var ConservationPracticeView = DrawControlView.extend({
     template: conservationPracticeTmpl,
+
+    templateHelpers: function() {
+        return {
+            label: models.getHumanReadableLabel
+        };
+    },
 
     getControlName: function() {
         return 'conservation_practice';
--- a/src/mmw/js/src/modeling/templates/controls/conservationPractice.html
+++ b/src/mmw/js/src/modeling/templates/controls/conservationPractice.html
@@ -7,14 +7,14 @@
     <div id="conserve-tools" class="dropdown-menu draw-tools menu-left" role="menu"> <!-- Dropdown -->
         <div class="pad-1"> <!-- Dropdown Content -->
             <ul class="row">
-                {{ utils.thumb('Rain Garden', 'rain_garden') }}
-                {{ utils.thumb('Veg Infil Basin', 'veg_infil_basin') }}
-                {{ utils.thumb('Porous Paving', 'porous_paving') }}
+                {{ utils.thumb(label, 'rain_garden') }}
+                {{ utils.thumb(label, 'veg_infil_basin') }}
+                {{ utils.thumb(label, 'porous_paving') }}
             </ul>
             <ul class="row">
-                {{ utils.thumb('Green Roof', 'green_roof') }}
-                {{ utils.thumb('No-Till Agriculture', 'no_till_agriculture') }}
-                {{ utils.thumb('Cluster Housing', 'cluster_housing') }}
+                {{ utils.thumb(label, 'green_roof') }}
+                {{ utils.thumb(label, 'no_till_agriculture') }}
+                {{ utils.thumb(label, 'cluster_housing') }}
             </ul>
         </div> <!-- End Dropdown Content -->
 
--- a/src/mmw/js/src/modeling/templates/controls/utils.html
+++ b/src/mmw/js/src/modeling/templates/controls/utils.html
@@ -1,10 +1,10 @@
-{% macro thumb(name, value, size=56) %}
+{% macro thumb(label, value, size=56) %}
     <li>
         <div class="thumb" data-value="{{ value }}">
             <svg>
                 <rect height="{{ size }}" width="{{ size }}" fill="url(#fill-{{ value }})"></rect>
             </svg>
         </div>
-        <label class="dark">{{ name }}</label>
+        <label class="dark">{{ label(value) }}</label>
     </li>
 {% endmacro %}
--- a/src/mmw/js/src/modeling/tests.js
+++ b/src/mmw/js/src/modeling/tests.js
@@ -519,7 +519,7 @@ var greaterThanOneAcrePolygon = { "type": "FeatureCollection", "features": [ { "
 
 var modificationsSample1 = {
     "name":"Land Cover",
-    "value":"LIR",
+    "value":"lir",
     "shape":{"type":"Feature","properties":{},"geometry":{"type":"Polygon","coordinates":[[[-76.00479125976562,40.19251207621169],[-76.04324340820312,40.13794057716276],[-75.95260620117188,40.136890695345905],[-75.93338012695312,40.182020964319086],[-75.96221923828125,40.199854889057676],[-76.00479125976562,40.19251207621169]]]}},
     "area":10977.041602204828,
     "units":"acres"
@@ -527,7 +527,7 @@ var modificationsSample1 = {
 
 var modificationsSample2  = {
     "name":"Conservation Practice",
-    "value":"Rain Garden",
+    "value":"rain_garden",
     "shape":{"type":"Feature","properties":{},"geometry":{"type":"Polygon","coordinates":[[[-75.53237915039062,40.18307014852534],[-75.66009521484375,40.107487419012415],[-75.50491333007812,40.10118506258701],[-75.43350219726561,40.13899044275822],[-75.42800903320312,40.1673306817866],[-75.53237915039062,40.18307014852534]]]}},
     "area":26292.18855342856,
     "units":"acres"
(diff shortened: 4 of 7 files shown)


rajadain:
It works well in the case of new modifications:

image

But for old modifications in the database that do not have a `label`, we see nothing:

image

Perhaps you were right earlier, and we should fallback to `value` in case `label` is unavailable. However, all modifications going forward _will_ have the `label` defined, so perhaps we shouldn't add that fallback? Curious to hear what others think.

You will have to update tests to match the new modification data structure, updating `value` and adding `label` [here](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/tests.js#L522) and [here](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/tests.js#L530).

Also, our Waffle convention is to use "Connects #369" rather than "Fixes ", which would automatically close the issue, and we want to manually close it after testing.

jamesmcclain:
Yeah, I forgot to run tests before I pushed, made the PR, my apologies.  Regarding the issue of whether to fall back to `values` when `labels` is missing: I think I will wait a little while to see  what the consensus is.

src/mmw/js/src/modeling/templates/controls/utils.html
```diff
@@ -1,6 +1,6 @@
 {% macro thumb(name, value, size=56) %}
     <li>
-        <div class="thumb" data-value="{{ value }}">
+        <div class="thumb" data-value="{{ value }}" data-label="{{ name }}">
```
kdeloach: Consider renaming parameter so we can say ` data-label="{{ label }}"`.

src/mmw/js/src/modeling/controls.js
```diff
@@ -48,11 +48,13 @@ var DrawControlView = ControlView.extend({
             $el = $(e.currentTarget),
             controlName = this.getControlName(),
             controlValue = $el.data('value'),
+            controlLabel = $el.data('label'),
             map = App.getLeafletMap();
         drawUtils.drawPolygon(map, patterns.getDrawOpts(controlValue)).then(function(geojson) {
             self.addModification(new models.ModificationModel({
                 name: controlName,
                 value: controlValue,
+                label: controlLabel,
```
kdeloach: Please add new model attributes to the `defaults` hash on `ModificationModel`.

caseypt: +1. This should take care of the missing labels problem described above.

kdeloach:
Although I suggested adding a label attribute, I didn't realize the impact it would have on saving/loading modifications. It's probably not a good idea to save labels in the DB for reasons mentioned above.

We should always be able to derive the label for any given modification value client-side, so consider creating a lookup table/function. We could probably use a custom template tag to convert values to labels.

lliss:
I agree, I don't think we should be relying on the stored data to let us know the label. We should derive it somehow.

mmcfarland:
At some point, we'll probably be doing some static exporting server side that will include the modification labels.  I think there's a high probability of that being a PhantomJS process, so even then, having a lookup in JS is probably a workable solution.  Can change in the future if we need to.

rajadain:
The best place to add a lookup is probably the constructor of `ModificationModel`, where we can initialize the `label` given the `value`. Unfortunately, I can't seem to find a way to reference that in the Nunjucks templates for [landCover](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/templates/controls/landCover.html) and [conservationPractice](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/templates/controls/conservationPractice.html). So we'll have some redundancy in the code, since the mapping will be defined twice. We should probably add a test to ensure they stay in sync.

jamesmcclain:
I was asking about that issue on slack.

kdeloach:
If we use a template tag to perform a lookup we could also utilize that in the landCover and conservationPractice templates to reduce redunancies.

src/mmw/js/src/modeling/controls.js
```diff
@@ -12,6 +12,32 @@ var $ = require('jquery'),
     conservationPracticeTmpl = require('./templates/controls/conservationPractice.html'),
     precipitationTmpl = require('./templates/controls/precipitation.html');
 
+function getHumanReadableLabel(value) {
```
lliss: I kind of feel as thought this helper should be part of the ModificationModel or at least in the models file.

src/mmw/js/src/modeling/templates/controls/conservationPractice.html
```diff
@@ -7,14 +7,14 @@
     <div id="conserve-tools" class="dropdown-menu draw-tools menu-left" role="menu"> <!-- Dropdown -->
         <div class="pad-1"> <!-- Dropdown Content -->
             <ul class="row">
-                {{ utils.thumb('Rain Garden', 'rain_garden') }}
-                {{ utils.thumb('Veg Infil Basin', 'veg_infil_basin') }}
-                {{ utils.thumb('Porous Paving', 'porous_paving') }}
+                {{ utils.thumb(getHumanReadableLabel('rain_garden'), 'rain_garden') }}
```
rajadain: Can we make this even DRYer by refactoring the macro to accept only one parameter, and use the `getHumanReadableLabel` function inside the macro?

```
{{ utils.thumb('rain_garden') }}
```

And use the [set](https://mozilla.github.io/nunjucks/templating.html#set) directive to initialize a `label` value inside the macro:

```
{% macro thumb(value, size=56) %}
{% set label = getHumanReadableLabel(value) %}
    ...
    <label class="dark">{{ label }}</label>
    ...
{% endmacro %}
```

lliss:
+1. Tested. Works as described.

src/mmw/js/src/modeling/controls.js
```diff
@@ -12,6 +12,32 @@ var $ = require('jquery'),
...
+        'rain_garden': 'Rain Garden',
+        'veg_infil_basin': 'Veg Infil Basin'
+    };
+
+    if (mapping[value]) {
+        return mapping[value];
+    } else {
+        throw 'Unknown Land Cover or Conservation Practice: ' + value;
+    }
+}
```
kdeloach: Nice and simple.

src/mmw/js/src/modeling/templates/controls/conservationPractice.html
rajadain: This is a purely nice-to-have though, I don't think we need to spend too much time on it if it isn't very straightforward.

jamesmcclain:
@lliss I left the helper separate from `ModificationModel` because it does not actually interact with it in anyway (since I took the label field back out).  Do you feel strongly about this?

@rajadain Good idea, I will give that a try.

jamesmcclain:
Thanks everyone for the comments.

caseypt:
I think it would be better if the helper function were on the model for two reasons:

1) It is easily testable there.
2) I need a way to get the label for the popups in #252

kdeloach:
We could create a static method or (at minimum) expose the function from that module. (I'm a fan of using plain functions when they don't need to use instance data).

caseypt:
Just talked to @jamesmcclain and a static method sounds good.

jamesmcclain:
I ended up creating a static method as well as just exporting it out of `models.py` because only doing one seems to be inconvenient.

lliss:
+1 :shipit:

src/mmw/js/src/modeling/templates/controls/conservationPractice.html
```diff
@@ -7,14 +7,14 @@
     <div id="conserve-tools" class="dropdown-menu draw-tools menu-left" role="menu"> <!-- Dropdown -->
         <div class="pad-1"> <!-- Dropdown Content -->
             <ul class="row">
-                {{ utils.thumb('Rain Garden', 'rain_garden') }}
-                {{ utils.thumb('Veg Infil Basin', 'veg_infil_basin') }}
-                {{ utils.thumb('Porous Paving', 'porous_paving') }}
+                {{ utils.thumb(label, 'rain_garden') }}
```
rajadain: Great job on the reduction here!

jamesmcclain: Thank you for the suggestion!

jamesmcclain:
Okay, I will let this hang out for a little while and merge right before I go home.

kdeloach:
> I ended up creating a static method as well as just exporting it out of models.py because only doing one seems to be inconvenient (for different usages).

For future reference, The Zen of Python is a great set of guidelines for good API design. It applies to more than just Python. In this case, the passage "There should be one-- and preferably only one --obvious way to do it" comes to mind. (Source: https://www.python.org/dev/peps/pep-0020/)

jamesmcclain:
Understood.

src/mmw/js/src/modeling/tests.js
```diff
@@ -527,7 +527,7 @@ var modificationsSample1 = {
 
 var modificationsSample2  = {
     "name":"Conservation Practice",
-    "value":"Rain Garden",
+    "value":"rain_garden",
```
lliss: It's pretty great that the tests already check for the labels. Your new stuff got a test for free.

jamesmcclain: I like free.

Issue #402, opened 2015-06-15
Roboto font only served over HTTP


hectcastro:
References to the Roboto font are being pulled over HTTP, which causes mixed content error messages when the parent page is served via HTTPS.

PR #403, opened 2015-06-16
Always serve fonts over HTTPS.

lliss:
If we are on an HTTPS connection (as we will be in production) we want our fonts
to be served over HTTPS as well. Since there is no harm in it, always serve
fonts over an HTTPS connection.

Connects #402


rajadain:
+1 both CSS and fonts are served over HTTPS:

image

lewfish:
+1, tested

Issue #397, opened 2015-06-15
Precipitation slider should display 0" if the slider is all the way to the left


caseypt:
![selection_062](https://cloud.githubusercontent.com/assets/1042475/8166497/332ebc7c-1366-11e5-8734-7aa2acb3ca98.png)

kdeloach:
This line needs to be updated: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/controls.js#L96.

Issue #357, opened 2015-06-11
Remove scenario modification tools for non-owners


mmcfarland:
Non owners should not have modification controls available for public projects (non owner could be an owner that is logged out).

Issue #393, opened 2015-06-15
Add JS tests around running models


caseypt:
- Making a change to a scenario starts a model (add modification, change an input value)
- Test `TaskModel.reset`
- Test `TaskModel.start`
- Test `TaskModel.pollForResults`
- Test `ResultCollection.setPolling`
- Test `ResultCollection.setNullResults`
- Test `ScenarioModel.initialize`
- Test `ScenarioModel.getResults`

Depends on #345

kdeloach:
Other candidates for testing:
  * Non-owners should not be able to see modification controls (input controls should still be visible).

caseypt:
Add tests for `modificationConfigUtils`.

lewfish:
Modify tests for saving projects/scenarios to account for their asynchrony.

PR #404, opened 2015-06-16
Remove scenario modification tools for non-owners

kdeloach:
Hides modifications tools for non-owners and renames a few arguments
to make things easier to understand.

How to test:
1. You should be able to see all modification controls on all owned projects.
2. You should only be able to see the precipitation slider on non-owned projects (or if you are logged out).
3. Changing the precip slider on non-owned projects should update the results panel, but should *not* save anything.

Connects #357
Connects #397

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -9,6 +9,14 @@ var Backbone = require('../../shim/backbone'),
 var ModelPackageControlModel = Backbone.Model.extend({
     defaults: {
         name: ''
+    },
+
+    // Return true if this is an input control and false if it is a
+    // modification control.
+    isInputControl: function() {
+        return _.contains([
+            'precipitation'
+        ], this.get('name'));
     }
 });
 
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -105,7 +105,7 @@ var ProjectMenuView = Marionette.ItemView.extend({
 
     templateHelpers: function() {
         return {
-            editable: App.user.userMatch(this.model.get('user_id'))
+            editable: isEditable(this.model)
         };
     },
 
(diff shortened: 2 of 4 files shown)


src/mmw/js/src/modeling/views.js
```diff
@@ -429,20 +430,21 @@ var ScenarioDropDownMenuView = Marionette.CompositeView.extend({
 var ToolbarTabContentView = Marionette.CompositeView.extend({
     model: models.ScenarioModel,
     template: scenarioToolbarTabContentTmpl,
-    collection: models.ModificationsCollection,
+    collection: models.ModelPackageControlsCollection,
```
kdeloach: This was actually a mistake before that I just noticed.

src/mmw/js/src/modeling/models.js
```diff
@@ -9,6 +9,14 @@ var Backbone = require('../../shim/backbone'),
 var ModelPackageControlModel = Backbone.Model.extend({
     defaults: {
         name: ''
+    },
+
+    // Return true if this is an input control and false if it is a
+    // modification control.
+    isInputControl: function() {
```
kdeloach: This is poorly named and will be address when I split modifications & input controls.

kdeloach:
Updated so input controls are always visible.

kdeloach:
Updated with test instructions.

src/mmw/js/src/modeling/views.js
```diff
@@ -770,6 +779,10 @@ var ResultsTabContentsView = Marionette.CollectionView.extend({
     }
 });
 
+function isEditable(scenario) {
```
caseypt: Consider making this an instance method on scenario.

caseypt:
Consider adding some tests, or adding potential tests to the running list here: https://github.com/WikiWatershed/model-my-watershed/issues/393

caseypt:
+1. Tested. I was able to edit and run scenarios on an owned, public project. Then in an incognito window I loaded that project. The modification controls were gone, but the precipitation slider was still there. I was able to change the slider and run models.

caseypt:
Also, the precipitation slider now shows `0.0"` when it is all the way to the left.

src/mmw/js/src/modeling/views.js
kdeloach: Forgot to address this before merging. I wasn't sure if it made sense to make this an instance method, considering it relies on a global variable. I could rewrite this to something like `isEditableBy(user_id)` and make it an instance method on Scenario?

caseypt: +1 to your proposal.

kdeloach: Cool, I'll fix this tomorrow.

Issue #388, opened 2015-06-15
Add Favicon


ctaylo37:

hectcastro:
Retina favicon appears to be 404ing because the `@` symbol is getting escaped.

PR #406, opened 2015-06-16
Retina favicon filename.

jamesmcclain:
Provide the correct filename for the retina favicon.

**To test**:
   * Load the site on a browser that can use a retina favicon
   * Hopefully you should see a retina favicon

Connects WikiWatershed/model-my-watershed#388

### DIFF
--- a/src/mmw/apps/core/templates/base.html
+++ b/src/mmw/apps/core/templates/base.html
@@ -7,7 +7,7 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <link rel="icon" type="image/png" href="{% static 'favicon.png' %}" sizes="16x16">
-    <link rel="icon" type="image/png" href="{% static 'favicon.png@2x' %}" sizes="32x32">
+    <link rel="icon" type="image/png" href="{% static 'favicon@2x.png' %}" sizes="32x32">
     <link rel="stylesheet" href="{% static 'css/vendor.css' %}" />
     <link rel="stylesheet" href="{% static 'css/main.css' %}" />
 </head>


jamesmcclain:
I don't think that my browser automatically requests the retina favicon, however I was able to get the retina favicon to show up after applying this change then commenting-out the line for the regular favicon.  If this does not fix the problem, I will try again.

lliss:
+1

rajadain:
+1 tested on a Retina screen, confirmed that I'm seeing the high resolution icon:

image

hectcastro:
I get a 200 on the Retina favicon now, but the `@` still appears to be getting escaped:

```
GET /static/favicon%402x.png HTTP/1.1
Host: localhost:8000
Connection: keep-alive
Pragma: no-cache
Cache-Control: no-cache
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/44.0.2403.39 Safari/537.36
Accept: */*
Referer: http://localhost:8000/
Accept-Encoding: gzip, deflate, sdch
Accept-Language: en-US,en;q=0.8,es;q=0.6
```

rajadain:
@hectcastro I don't think the original 404 was because of the escaping, but instead because the filename was specified incorrectly which has been fixed here. `@2x` should be before, not after, the `.png` extension.

hectcastro:
Yeah, you're right. I diagnosed the original issue incorrect, but still seems weird to have this symbol escaped here.

rajadain:
Both

    curl http://localhost:8000/static/favicon@2x.png

and

    curl http://localhost:8000/static/favicon%402x.png

thankfully result in the same output for me. I think it is escaped in the template pre-processing, but I could be wrong.

jamesmcclain:
What is the verdict, here?  Are we interested in trying to prevent this string from being escaped or is the present solution okay?

hectcastro:
:+1: It works.

jamesmcclain:
Thanks everyone.

PR #407, opened 2015-06-16
Add JSHint to build

kdeloach:
This adds a new npm command to lint the project and also will fail the build if any lint is present.

Run jshint with this command:

    ./scripts/npm.sh run lint

Connects #387

### DIFF
--- a/src/mmw/js/src/core/setup.js
+++ b/src/mmw/js/src/core/setup.js
@@ -7,6 +7,17 @@
 var $ = require('jquery');
 window.jQuery = window.$ = $;
 
+// Marionette attaches itself to Backbone so these must be required in
+// this order.
+// Instead of requiring both Backbone and Marionette in every module that
+// needs one or the other, explicitly require them here in the correct
+// order, so they work predictably everywhere.
+require('../../shim/backbone');
+require('../../shim/backbone.marionette');
+
+// Include template filters.
+require('../filters');
+
 require('bootstrap');
 require('bootstrap-select');
 
--- a/src/mmw/js/src/router.js
+++ b/src/mmw/js/src/router.js
@@ -9,14 +9,19 @@ var Backbone = require('../shim/backbone'),
 // Source: https://github.com/jashkenas/backbone/commit/623207831ec09687f9cd9fa3a461f88849b02578
 var PatchedRouter = Backbone.Marionette.AppRouter.extend({
     route: function(route, name, callback) {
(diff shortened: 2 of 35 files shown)


src/mmw/js/src/core/setup.js
```diff
@@ -7,6 +7,9 @@
 var $ = require('jquery');
 window.jQuery = window.$ = $;
 
+require('../../shim/backbone');
```
lliss: Are you sure this doesn't need to be added here so that other parts of the application have access to them. Didn't we have trouble with this before.

src/mmw/js/src/router.js
```diff
@@ -9,14 +9,19 @@ var Backbone = require('../shim/backbone'),
 // Source: https://github.com/jashkenas/backbone/commit/623207831ec09687f9cd9fa3a461f88849b02578
 var PatchedRouter = Backbone.Marionette.AppRouter.extend({
     route: function(route, name, callback) {
-        if (!_.isRegExp(route)) route = this._routeToRegExp(route);
+        if (!_.isRegExp(route)) {
```
lliss: Yay.

lliss:
Tests still pass so +1.

lliss:
I do agree with @mmcfarland, it would be nice if this showed as an instability rather than a hard failure.

lliss:
Also, as a general comment, it is nice to see this.

src/mmw/js/src/core/setup.js
kdeloach: The reason is subtle, but these dependencies have to be included in this order. It used to work fine, until I removed some "unused" require statements caught by jshint.

For example, `router.js` references `Backbone.Marionette` but it only requires `Backbone` (not `Marionette`). 

The problem is that `Marionette` attaches itself to `Backbone`. This means that `Backbone` always has to be required before `Marionette`. Instead of doing this in every module, I decided to explicitly include them in the correct order here.

kdeloach: I'll add a comment.

mmcfarland:
+1

kdeloach:
@azaveaci rebuild

kdeloach:
* Fixed the command to scan JS files recursively
  * Fixed (a lot) more lint
  * Updated Jenkins to mark build as unstable instead of failing if there is lint

caseypt:
The unused require flag is nice. +1

kdeloach:
@azaveaci rebuild

Issue #261, opened 2015-06-02
Add Water Quality chart to model output


mmcfarland:
`tr55` v0.1.0 was released and has output to model simple water quality measurements.  Update our requirements to bring in the new version and construct a simple bar chart and data table to display the high level measurements of the scenario model run.

lewfish:
We should wait until #200 is done to do this, since the code should be very similar.

Issue #80, opened 2015-04-21
Handle loading and error states for analyze window


caseypt:
- While the app is waiting the analyze request to come back, the window should slide in and display a loading spinner, or something similar.
- If the response comes back with an error, a nice message should be shown instead of the tables and graphs.

Consult @ctaylo37 on this.

Issue #62, opened 2015-04-15
Geocoder searching, error, and no results mock-ups


caseypt:
We need mock-ups/designs around what the the geocoder should do/look like when

- A searching is being executed
- An error occurred
- There are no results for the current query

caseypt:
Since we are using a difference service for suggestions and geocoding, we need a mockup for when you click on a result in the list and are waiting for the geocode response.

caseypt:
Also, I think we should implement this soon.

ctaylo37:
I think we should have the loading spinner in-line with the result that has been clicked:

![image](https://cloud.githubusercontent.com/assets/5880793/8161790/3c59d5da-1347-11e5-821b-1b23c97731c4.png)

mmcfarland:
This would be a nice-to-have for this sprint, but not a main priority.  Updated label.

PR #412, opened 2015-06-16
Add Water Quality chart/table to model output

lewfish:
To test:
 * Restart celery
 * Create a new project
 * Make modifications
 * The Runoff and Water Quality tabs should show spinners and then show results. The Water Quality tab should show a table and bar chart with four items.
 * Try hitting reload to make sure it works with a saved project.

Known Issues:
 * An extra space above the chart (only visible on high-contrast screens).
screen shot 2015-06-16 at 12 29 00 pm

 * The ids for the Runoff and Water Quality view els are the same for each scenario because these views do not have access to the scenario they are part of. This isn't causing a problem, but generally speaking, ids should be unique.
 * The code/functionality for the Water Quality view is similar to that of the Analyze view. In the future, we might want to refactor these to exploit commonalities. 

Connects #261

### DIFF
--- a/src/mmw/js/src/modeling/tr55/quality/views.js
+++ b/src/mmw/js/src/modeling/tr55/quality/views.js
@@ -1,0 +1,92 @@
+"use strict";
+
+var Marionette = require('../../../../shim/backbone.marionette'),
+    DataCollection = require('../../../core/models.js').DataCollection,
+    chart = require('../../../core/chart.js'),
+    barChartTmpl = require('../../../core/templates/barChart.html'),
+    resultTmpl = require('./templates/result.html'),
+    tableRowTmpl = require('./templates/tableRow.html'),
+    tableTmpl = require('./templates/table.html');
+
+var ResultView = Marionette.LayoutView.extend({
+    className: 'tab-pane',
+    id: function() {
+        return this.model.get('name');
+    },
+    template: resultTmpl,
+    attributes: {
+        role: 'tabpanel'
+    },
+    regions: {
+        tableRegion: '.quality-table-region',
+        chartRegion: '.quality-chart-region'
+    },
+    modelEvents: {
+        change: 'onShow'
+    },
+    onShow: function() {
+        this.tableRegion.reset();
+        this.chartRegion.reset();
+        if (this.model.get('result')) {
+            var dataCollection = new DataCollection(
+                this.model.get('result')
+            );
+
+            this.tableRegion.show(new TableView({
+                collection: dataCollection
+            }));
+
+            this.chartRegion.show(new ChartView({
+                model: this.model,
+                collection: dataCollection
+            }));
+        }
+    }
+});
+
+var TableRowView = Marionette.ItemView.extend({
+    tagName: 'tr',
+    template: tableRowTmpl,
+
+    templateHelpers: function() {
+        return {
+            roundedLoad: this.model.get('load').toFixed(3)
+        };
+    }
+});
+
+var TableView = Marionette.CompositeView.extend({
+    childView: TableRowView,
+    childViewContainer: 'tbody',
+    template: tableTmpl
+});
+
+var ChartView = Marionette.ItemView.extend({
+    template: barChartTmpl,
+    className: 'chart-container quality-chart-container',
+
+    onAttach: function() {
+        this.addChart();
+    },
+
+    addChart: function() {
+        var selector = '.quality-chart-container .bar-chart',
+            chartData = this.collection.map(function(model) {
+                return model.attributes;
+            }),
+            chartOptions = {
+                isPercentage: false,
+                depAxisLabel: 'Load (lbs)',
+                useHorizBars: true,
+                horizMargin: {top: 20, right: 80, bottom: 40, left: 150}
+            },
+            depVars = ['load'],
+            indVar = 'measure';
+
+        chart.makeBarChart(selector, chartData, indVar, depVars, chartOptions);
+    }
+});
+
+module.exports = {
+    ResultView: ResultView
+};
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -151,20 +151,20 @@ def make_gt_service_call_task(model_input):
 
 
 def format_quality(model_output):
-    metrics = ['Biochemical Oxygen Demand',
-               'Total Suspended Solids',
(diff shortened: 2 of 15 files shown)


rajadain:
It might be a nice idea to implement the loading animations defined in #80 here as well, but that can also be pushed into a later PR.

The numbers look really unwieldy in their current representation:

image

Consider rounding them to an acceptable number, or using scientific notation if accuracy is desired.

The "Pounds" unit label on the vertical axis is obstructed by the bar chart on small screens.

The bar chart label "Biochemical Oxygen Demand" is too wide and runs into the next "Total Suspended Solids" on small screens. Consider setting a max-width or percentage width for those labels and wrap text to ensure they don't overlap.

rajadain:
Question for people: should the results window auto-expand after results have been updated? Currently it must be done by hand:

water-quality-pr

caseypt:
@rajadain I'm in favoring of keeping it minimized after the user minimizes it. I think it could get annoying if  the user minimizes it, makes a change, it pops up, the user minimizes it, makes a change, etc.

We may want to add a better signal there are new results instead of only showing the spinner.

rajadain:
@caseypt agreed. I would also like it if the spinner icon didn't change the width of the tabs like it does now. One way to fix it would be to have a default icon there all the time, and change it to the spinner while loading, and change it back to the default icon when done. Perhaps @ctaylo37 can choose a default icon that makes sense.

Also, I would be in favor of using [`fa-circle-o-notch`](http://fortawesome.github.io/Font-Awesome/icon/circle-o-notch/) for the spinner, as specified in #62, to maintain visual consistency.

rajadain:
Added #414 as optional enhancement to the behavior here.

mmcfarland:
Reassigning because I can't test it out due to #419, which is affecting my local dev.  I'll take a pass at the code though.

rajadain:
TR-55 model is correctly run when:

 * You land on the modeling page
 * You add a new scenario
 * You change the precipitation slider
 * You add a modification

It should also be run (but is not) when:

 * You delete a modification

rajadain:
I'm not entirely sure what the numbers in the table and graph represent:

image

I know they are in Pounds, but is it mass or force? Pounds are the units (likely to change) but what metric do they describe of the different components?

Given that we may switch (or allow the user to switch on the fly) between imperial and metric units, consider using a label which describes the data being measured.

I was going to suggest to switch the variable name as well, but I see that `pounds` is actually the name of the field we receive from the API, so switching that will be a larger refactor and may be left to a future PR.

lewfish:
@mmcfarland Do you have any better ideas about what the labels on the table should be called, rather than metric and pounds?

mmcfarland:
I would call the column "Load" and have the units expressed as a row label, and "Quality Measure" for the first column.  Sort of just made that last one up, but we can get feedback and it's accurate.

```
Quality Measure                Load
Total Nitrogen                 0.234 lbs
```

lewfish:
I'm about to use the scenario.cid as a prefix on view ids so they'll be unique. But, to do so I have to pass the scenario (or at least its cid) down the chain of views. I hesitate to do this since we aren't using the ids on the views for anything. Can I just omit the id altogether? Or, is it considered a bad practice for the el of views to not have an id?

mmcfarland:
If you don't need an id for styling or selection, I would drop it.

lewfish:
@azaveaci rebuild

lewfish:
I've responded to all feedback, except:

"The bar chart label "Biochemical Oxygen Demand" is too wide and runs into the next "Total Suspended Solids" on a 13" Retina full screen. Consider setting a max-width or percentage width for those labels so they are never wider than the bar width and wrap text to ensure they don't overlap."

In the past, we decided to use horizontal bar charts when the labels of the bars are too long, so I've done that for this PR. @mmcfarland Is it worth making a new issue to add the functionality @rajadain suggested above?

mmcfarland:
No, I think at some point we're going to get some feedback from the client on the structure of the output, so I don't want to spend too much time tweaking it.  I think having a reasonable representation now is sufficient.

src/mmw/js/src/modeling/tr55/quality/views.js
```diff
@@ -0,0 +1,92 @@
+"use strict";
+
+var Marionette = require('../../../../shim/backbone.marionette'),
+    coreModels = require('../../../core/models.js'),
```
rajadain: If we're only ever using `coreModels.DataCollection` in this file, the code below could be made a little DRYer by importing:

```javascript
DataCollection = require('../../../core/models.js').DataCollection,
```

and then just using `new DataCollection({})` instead of `new coreModels.DataCollection({})`

src/mmw/js/src/modeling/tr55/quality/views.js
```diff
@@ -0,0 +1,92 @@
...
+    }
+});
+
+var TableRowView = Marionette.ItemView.extend({
+    tagName: 'tr',
+    template: tableRowTmpl,
+
+    templateHelpers: function() {
+        return {
+            roundedLoad: this.model.get('load').toFixed(2)
```
rajadain: In @mmcfarland's example, he used 3 decimal points: https://github.com/WikiWatershed/model-my-watershed/pull/412#issuecomment-112840142

I'd say we'd need three, because some values are like 0.0096 something.

rajadain:
Code changes look great apart from the two minor issues above. If you could rebase this I'll test it out.

src/mmw/js/src/modeling/tr55/quality/views.js
kdeloach: Just as a general comment we can omit the attributes argument if it's empty so `new coreModels.DataCollection()` should by valid.

rajadain: Good to know!

rajadain:
+1 works perfectly now, splendid work!

 * Chart has no occlusions or overflows anymore
 * Table and chart labels are now more meaningful, and table values are easier to consume without loosing too much detail. The smallest value, "Total Phosphorus" has a non-zero value starting at 0.2" precipitation.
 * Deleting a modification causes model rerun, in addition to adding a modification and changing the slider
 * Adding a new scenario, or duplicating an existing one cases model rerun. Deleting a scenario correctly does not cause model rerun.
 * Spinner icons are now consistent with those used elsewhere.

rajadain:
Looks pretty good on my end:

image

PR #413, opened 2015-06-16
Add popup for modifications.

caseypt:
* Create view for modifications popup and set as popup content.
* Make modifications clickable and set map mask as unclickable.
* Add modification popup template.

**To Test:**
- Load an existing or create a new model.
- Draw some modifications.
- Click on the modifications to trigger a pop-up.
- Make sure a label appears and that when you click the trash can, the modification is removed, and the count in the scenario toolbar drops.
- Run the tests.

Connects to #252

### DIFF
--- a/src/mmw/js/src/core/patterns.js
+++ b/src/mmw/js/src/core/patterns.js
@@ -25,16 +25,14 @@ var getDrawOpts = function(pattern) {
     if (!pattern || !colors[pattern]) {
         // Unknown pattern, return generic grey
         return {
-            clickable: false,
             color: '#888',
             opacity: 1,
             weight: 3,
             fillColor: '#888',
             fillOpacity: 0.74
-        }
+        };
     } else {
         return {
-            clickable: false,
             color: colors[pattern],
             opacity: 1,
             weight: 3,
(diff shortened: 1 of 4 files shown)


caseypt:
Ideally, all of this code would be in `modeling`, but the drawing of the modifications stretches across `modeling` and `core`. I placed it in `core` because it was the simplest way to do it. I tried keeping it all in `modeling`, and assigning the popup content as a attribute on the modification model, but that introduced some problems with serialization, and didn't seem right.

rajadain:
Regarding the delete / trash icon in the popup, it feels a little too small and alone:

image

The icon makes a lot of sense in the tabular compact representation of the Modifications Dropdown, but since we have more room in the popup, what do you think about using a word instead? "Delete", or "Trash" as suggested by the mockup?

caseypt:
@rajadain I'll defer to @ctaylo37 on that. It could use some design help in general.

rajadain:
+1 tested, works as advertised. Deletion works as well. Also works on old projects created long before! Great work.

ctaylo37:
@caseypt If you want to merge this in now, I can take a stab at styling tomorrow.

mmcfarland:
+1 Tested.  Nice small footprint.

caseypt:
@ctaylo37 merged.

PR #416, opened 2015-06-16
Focus primary field on modals

kdeloach:
This will focus the primary field on modals when they first appear, and
after any validation errors occur.

I altered the structure of these classes to use a more traditional
inheritance design pattern to make it easier to support overriding
methods.

### DIFF
--- a/src/mmw/js/src/user/views.js
+++ b/src/mmw/js/src/user/views.js
@@ -92,6 +92,7 @@ var ModalBaseView = Marionette.ItemView.extend({
             'client_errors': null,
             'server_errors': server_errors
         });
+        this.onValidationError();
     },
 
     // Dismiss Action can be "cancel" or "continue as guest". Noop by default.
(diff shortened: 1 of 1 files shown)


rajadain:
Successfully tested the following:

 * Focus first field upon render for:
    - Login Modal
    - Sign Up Modal (when clicking Sign Up from Login Modal)
    - Sign Up Modal (when landing on `/sign-up`)
    - Forgot Modal
    - Sign Up ITSI Modal
 * Focus first field upon client validation errors

Did not work:

 * Focus first field upon server validation errors

For example, enter the wrong username and password. The username field is not focused on.

rajadain:
Very good! I like this inheritance scheme much _much_ better! The code is now simpler and more explicit.

Regarding setting focus on server validation errors, consider adding a `onValidationError();` call at the end of `ModalBaseView.handleServerFailure`.

Stylistically, do you think it makes sense to rename `primaryAction`, `handleServerSuccess` etc to `onPrimaryAction`, `onServerSuccess` etc?

caseypt:
I haven't explored them yet, but I think this could be a good case for Marionette Behaviors: http://marionettejs.com/docs/v2.4.1/marionette.behavior.html

kdeloach:
Fixed "Focus first field upon server validation errors" bug.

rajadain:
+1 tested! Works perfectly.

Issue #364, opened 2015-06-11
New projects should create default editable scenario


mmcfarland:
New projects should create a Current Conditions scenario, but also a new editable scenario that is the active tab.  Save on Edit logic shouldn't change.  The purpose is for the user to be able to start interacting with the modifications immediately without other actions.

PR #417, opened 2015-06-16
Add editable scenario when creating a new project.

caseypt:
* When a new project is created, in addition to adding a
scenario for current conditions, add a scenario that the user
can make modifications too.

**To Test:**
- Create a new project.
- There should be a new scenario named "New Scenario" that is active and editable. 

Connects to  #364


caseypt:
It's cool that this was so easy.

mmcfarland:
Very cool, yeah.  +1 tested.

PR #421, opened 2015-06-16
Add support for building app, tiler, and worker AMIs in parallel

hectcastro:
If you don't provide the `--machine-type` argument to the `create-ami` subcommand, it will now attempt to build all AMIs (except the monitoring AMI) in parallel.


hectcastro:

lliss:
Seems reasonable. +1

Issue #367, opened 2015-06-11
Initial extent should be determined from browser


mmcfarland:
Attempt to geolocate the user via the browser and setCenter on it on app start, if no AoI or Project extent exists.

rajadain:
It might also be useful to have a "locate me" button in the home page for the user to manually trigger this, a-la Google Maps.

mmcfarland:
Yeah, we use a leaflet plugin for that on USACE nicely https://github.com/domoritz/leaflet-locatecontrol

mmcfarland:
USACE Reference implementation:
https://github.com/azavea/usace-wisdm-symbolizer/blob/1640ddaa422917b83b70c95965e936dc567e1015/src/js/map.js#L704

PR #425, opened 2015-06-16
Use Geolocation.

jamesmcclain:
The initial position is now based on the user's location.  Also, a button to allow the user to warp to his/her location has been added to the top-right of the map.

**To test**
   1. You must first npm install the new dependency, leaflet-locatecontrol
   2. You should notice a new button in the upper-right of the map.  Clicking that button should center the map on your present physical location.
   3. You should also notice that when you visit the site with no AoI or project, the view will be centered on your present physical location.
   4. Claims 2 and 3 depend on geo-location being permitted

Connects #367

### DIFF
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -2,6 +2,7 @@
 
 var $ = require('jquery'),
     L = require('leaflet'),
+    leafletLocate = require('leaflet.locatecontrol'),
     _ = require('lodash'),
     router = require('../router.js').router,
     Marionette = require('../../shim/backbone.marionette'),
@@ -82,6 +83,33 @@ var HeaderView = Marionette.ItemView.extend({
 
 });
 
+// Init the locate plugin button and add it to the map.
+function addLocateMeButton(map, maxZoom, maxAge) {
+    var locateOptions = {
+        position: 'topright',
+        metric: false,
+        drawCircle: false,
+        showPopup: false,
+        follow: true,
+        markerClass: L.marker,
+        markerStyle: {
+            opacity: 0.0,
+            clickable: false,
+            keyboard: false
+        },
+        locateOptions: {
+            maxZoom: maxZoom,
+            // Cache location response, in ms
+            maximumAge: maxAge
+        },
+        strings: {
+            title: 'Zoom to your location.'
+        }
+    };
+
+    L.control.locate(locateOptions).addTo(map);
+}
+
 // This view houses a Leaflet instance. The map container element must exist
 // in the DOM before initializing.
 var MapView = Marionette.ItemView.extend({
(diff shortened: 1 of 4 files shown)


azaveaci:
Can one of the administrators verify this patch?

src/mmw/package.json
```diff
@@ -29,6 +29,7 @@
     "jstify": "0.9.0",
     "leaflet": "0.7.3",
     "leaflet-draw": "0.2.3",
+    "leaflet.locatecontrol": "0.43.0",
```
hectcastro: In addition to updating `package.json`, can you please update the associated shrinkwrap file using the steps outlined [here](https://docs.npmjs.com/cli/shrinkwrap#building-shrinkwrapped-packages)?

kdeloach: Please add new deps to vendor bundle as well by adding an entry here: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/bundle.sh#L109

src/mmw/js/src/core/views.js
```diff
@@ -83,6 +84,33 @@ var HeaderView = Marionette.ItemView.extend({
...
+        markerClass: L.marker,
+        markerStyle: {
+            opacity: 0.0,
+            clickable: false,
+            keyboard: false
+        },
+        locateOptions: {
+            maxZoom: 9,
+            // Cache location response, in ms
+            maximumAge: 60000
```
mmcfarland: If we're using this value twice, it should probably end up in a single var.  Probably the same with `maxZoom` above, but I'm less clear as to what it does and we're using zoom 10 below.

src/mmw/package.json
jamesmcclain: Certainly, I will do both.

kdeloach: Thanks!

rajadain:
+1 Works very well when geolocation is allowed. Tested the following:

 * Received location permission request from browser upon visiting the home page. I'm not sure if anyone would be taken aback by that, especially since the map is obscured by the login modal. But they can always deny it, so it should be fine.
 * Map was centered on my location after I gave permission.
 * Tried clicking the zoom to my location button from various locations and zoom levels, worked consistently.
 * Zoom to my location button is only visible in draw and analyze views, just like the zoom in and out buttons.

When I didn't allow location permission, the map did not automatically zoom to my location. Clicking the button gave the following alert message:

image

Which I think is fine for now. Perhaps in the future if we ever have a standardized error modal we could use that instead.

Given that a user has denied geolocation permission, I wonder if we could hide that button? But on second thought perhaps we should leave it so that if ever the user changes their mind in the future they know how to fix it.

rajadain:
Also ensured that upon loading existing projects, the map is centered on the AoI and not the user's location.

rajadain:
The rendering of map tiles is actually blocked until the user gives permission or not:

image

I assume this is because the map is waiting for a location to pull tiles for, but it looks weird. Perhaps we should setup the map to the default location while we are waiting for the user to give permission? This would cause more network load, but would make for a smoother user experience.

jamesmcclain:
Okay, I'll give that a try.

hectcastro:
@azavea-bot rebuild

rajadain:
Alright! Now the tiles are loaded for the default location as soon as one lands on the page, and after allowing geolocation the map is centered. There is a slight delay, but as I understand that is just the geolocation response time and not caused by a timeout on our end. Is that correct?

geolocation

++1, works great. Good job.

jamesmcclain:
Regarding the delay: yes, I believe that to be the case.

I will leave this open until at least after lunch, then squash and merge.  Thanks!

jamesmcclain:
@azavea-bot rebuild

jamesmcclain:
@azaveaci rebuild

kdeloach:
+1

jamesmcclain:
I am still having issue getting the CI service to build things.  In this particular case, the code hasn't changed since the last build (I just squashed two commits), but I wonder whether this is something I am doing wrong?

Issue #422, opened 2015-06-16
Projects don't save on Staging


mmcfarland:
Log in and start a new project, saves fail with a 403:

"CSRF Failed: CSRF token missing or incorrect."

mmcfarland:
Specifically at `POST https://mmw.azavea.com/api/modeling/projects/ `

PR #426, opened 2015-06-16
Simplify checking of URL when making AJAX call.

caseypt:
* All in internal AJAX requests should be relative, so check if there
is the requested URL starts with a HTTP. If it does, it shouldn't
have the CSRF header set.
* Fixes an issue that prevented API calls on staging because the CSRF
header was not being set on the requests.

Connects to  #422

### DIFF
--- a/src/mmw/js/src/core/csrf.js
+++ b/src/mmw/js/src/core/csrf.js
@@ -26,19 +26,12 @@ function getCookie(name) {
     return cookieValue;
 }
 
-// Source: http://stackoverflow.com/questions/6238351/fastest-way-to-detect-external-urls
-var isExternal = (function(){
-    var domainRe = /https?:\/\/((?:[\w\d]+\.)+[\w\d]{2,})/i;
+// Internal requests are relative.
+function isExternal(url) {
+    var domainRegExp = /^http/i;
 
-    return function(url) {
-        function domain(url) {
-          var result = domainRe.exec(url) !== null ? domainRe.exec(url)[1] : null;
-          return result;
-        }
-
-        return domain(location.href) !== domain(url);
-    };
-})();
+    return domainRegExp.exec(url) !== null;
+}
 
 exports.jqueryAjaxSetupOptions = {
     beforeSend: function(xhr, settings) {
--- a/src/mmw/js/src/core/tests.js
+++ b/src/mmw/js/src/core/tests.js
@@ -424,7 +424,7 @@ describe('Core', function() {
             }
         });
 
-        it('doesn\'t attach CSRF tokens to external requests', function() {
+        it('doesn\'t attach CSRF tokens to external requests over HTTP', function() {
             var requestMethods = [
                 'GET',
                 'HEAD',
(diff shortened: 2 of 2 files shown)


caseypt:
I'm not totally sure how the existing tests didn't catch this. It may have had something to do with where the tests are run from. Any suggestions for improving [them](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/tests.js#L368-L454) are welcome.

caseypt:
**To Test**:
- Try to create a new project and load an existing project. All API calls should be permitted.
- Try searching in the geocoder.
- Run the tests

caseypt:
Also tagging @kdeloach.

caseypt:
Also curious how this has been handled in other applications.

src/mmw/js/src/core/csrf.js
```diff
@@ -26,19 +26,12 @@ function getCookie(name) {
     return cookieValue;
 }
 
-// Source: http://stackoverflow.com/questions/6238351/fastest-way-to-detect-external-urls
-var isExternal = (function(){
-    var domainRe = /https?:\/\/((?:[\w\d]+\.)+[\w\d]{2,})/i;
+// Internal requests are relative.
+function isExternal(url) {
+    var domainRegExp = /^http/i;
```
mmcfarland: Does this catch `https` as well?

caseypt:
A bigger, and probably better, solution may be to proxy requests to the ESRI suggestion service through the back-end.

src/mmw/js/src/core/csrf.js
mmcfarland: Might be good to add an https test as well, either way.

caseypt: Yes, but I will also add a test case for this.

caseypt: Test added.

src/mmw/js/src/core/csrf.js
```diff
@@ -26,19 +26,12 @@ function getCookie(name) {
...
 
-    return function(url) {
-        function domain(url) {
-          var result = domainRe.exec(url) !== null ? domainRe.exec(url)[1] : null;
-          return result;
-        }
-
-        return domain(location.href) !== domain(url);
```
kdeloach: I think the tests passed because we never mocked out  `location.href` to use a "real" domain. The old regex always returns null for localhost.

kdeloach:
+1

mmcfarland:
+1

src/mmw/js/src/core/csrf.js
caseypt: Unfortunately, I've found no way to fake the host within Sinon.

kdeloach: I think we would need to create proxy browser objects for use throughout the app.

caseypt:
I'm going to merge this to get the fix in, but continue to research a better way to test it out.

Issue #423, opened 2015-06-16
Update branding of "simple model"


mmcfarland:
The `/simple` route which houses the Simple App is being renamed to "Micro Site Storm Model" and the url should be `/micro`.  Add the full name to the header bar, along the lines of "Model My Watershed - Mirco Site Storm Model", and any other public reference as appropriate.

PR #427, opened 2015-06-16
Rebrand mini-app from "simple" to "micro"

rajadain:
* Add `metatitle` block to `base.html` so page title can be set by
   child template
 * Change mini-app URL from `/simple` to `/micro`
 * Add "Model My Watershed Micro Site Storm Model" branding to title,
   header, and information dropdown in the mini-app.

**Testing Instructions**

 * Go to [`/micro`](http://localhost:8000/micro) and ensure you see the correct branding everywhere

Connects #423

### DIFF
--- a/src/mmw/apps/water_balance/templates/home_page/index.html
+++ b/src/mmw/apps/water_balance/templates/home_page/index.html
@@ -2,11 +2,15 @@
 
 {% load staticfiles %}
 
+{% block metatitle %}
+    <title>Model My Watershed - Micro Site Storm Model</title>
+{% endblock metatitle %}
+
 {% block header %}
     <header class="water-balance">
 
         <nav id="app-header" class="navbar-watershed">
-            <div class="brand"><a href="#">Model My Watershed</a></div>
+            <div class="brand"><a href="/">Model My Watershed</a><a href="#">&nbsp;&dash;&nbsp;Micro Site Storm Model</a></div>
             <div class="navigation">
                 <ul class="main">
                     <li class="header-link">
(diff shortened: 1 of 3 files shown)


src/mmw/apps/water_balance/templates/home_page/index.html
```diff
@@ -2,11 +2,15 @@
 
 {% load staticfiles %}
 
+{% block metatitle %}
+    <title>Model My Watershed Micro Site Storm Model</title>
```
mmcfarland: I think this might look better with some sort of separator between the main title and the sub, `:`, `-`, `|` or something.  Hope those don't just look like emojis here.

rajadain: Updated to have ':' as separator. Sorry, should've done that in the first place.

jamesmcclain:
+1, I believe that I see the correct branding on the page.

rajadain:
Switched from "Model My Watershed**:** Micro Site Storm Model" to "Model My Watershed **-** Micro Site Storm Model" in accordance with @ctaylo37's #428.

mmcfarland:
+1 looks good.  Let's merge this in favor of #428 since it adds all the other necessary changes.

src/mmw/apps/water_balance/templates/home_page/index.html
```diff
@@ -2,11 +2,15 @@
...
+{% block metatitle %}
+    <title>Model My Watershed - Micro Site Storm Model</title>
+{% endblock metatitle %}
+
 {% block header %}
     <header class="water-balance">
 
         <nav id="app-header" class="navbar-watershed">
-            <div class="brand"><a href="#">Model My Watershed</a></div>
+            <div class="brand"><a href="#">Model My Watershed - Micro Site Storm Model</a></div>
```
mmcfarland: I know this is sort of sneaking this in on this PR, but could you change `#` to `/` so we got a proper "home" link that works?

mmcfarland: Sorry - realized that this was the /micro homepage, which has no other "homes" to go to, was looking for this on the main page.

rajadain: Would you like a dual link solution, which links the main title to the home page but the subtitle to nothing?

```html
<div class="brand"><a href="/">Model My Watershed</a><a href="#">&nbsp;&dash;&nbsp;Micro Site Storm Model</a></div>
```

mmcfarland: Actually, yeah - they really want to emphasize the relationship between these two apps, and we'll need to put some thoughts towards that (i've got some ideas), but for now, at least having basic navigation is a good step.  I don't think we need the second link for the micro site though, we don't have anything to wire up to it.

rajadain: There's a lot of styling associated with `.brand .a`, so removing the link on the second part changes that text, and I think it is harmless to have that noop link. Note that this subtitle is only available on the `/micro` site, so we are not polluting the base app with it.

Should the link to the main app open in a new tab? We don't persist the user state in the micro app anywhere, but there is still state that the user might not want to lose.

rajadain: Implemented in f06680421c068820a029e295f2ff3b852194d1e6

mmcfarland: Good point.  Let's defer the decision for now and consider it in a broader context of perhaps persisting user data in the future, with respect to the priorities we should/shouldn't give this app.

PR #430, opened 2015-06-17
Add documentation for applying migrations to an AWS deployment

hectcastro:
Mostly covers using SSH authentication agent forwarding.


kdeloach:
+1

PR #431, opened 2015-06-17
Add NTP role to monitoring server playbook

hectcastro:
After #208, base roles were moved into `model-my-watershed.base`, but the monitoring server playbook does not list the base role as a dependency. Of the roles in `model-my-watershed.base`, `azavea.ntp` is the only one we want on the monitoring server.


hectcastro:

kdeloach:
+1

lliss:
+1

Issue #424, opened 2015-06-16
Add remaining Land Use types to Modification tools.


mmcfarland:
There should be a 1:1 match between modification land use types and those on the microsite (`/simple`).  Ultimately, they should have their own patterns, but we can reuse/modify unless @ctaylo37 can get something out soon.

ctaylo37:
So we are just using the 12 land cover types from the simple model?  Do you need additional patterns from me?

PR #434, opened 2015-06-17
Add four new modificaton types.

jamesmcclain:
Added modification tools for four more land use type: "chaparral", "desert", "tall grass prairie", and "short grass prairie".

~~There is still one land use type that is present in the micro app but not in the modification tools: "wetland".~~
I said that backwards: "wetland" is in the tools but not in the micro-app.

**To test**
   * Go to a scenario
   * Click on "Land Cover" to bring up a selection of draw tools
   * You should notice four new draw tools
   * If you draw polygons with the new tools, those polygons should be saved and restored as expected

Connects WikiWatershed/model-my-watershed#424

### DIFF
--- a/src/mmw/js/src/core/patterns.js
+++ b/src/mmw/js/src/core/patterns.js
@@ -11,6 +11,10 @@ var colors = {
     'grassland'  : '#54d2d0',
     'row_crop'   : '#52c6dd',
     'wetland'    : '#4ebaea',
+    'chaparral'  : '#16161d',
+    'desert'     : '#16161d',
+    'sg_prairie' : '#16161d',
+    'tg_prairie' : '#16161d',
 
     // Conservation Practice
     'rain_garden'        : '#51dec2',
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -511,15 +511,20 @@ function getControlsForModelPackage(modelPackageName, options) {
 
 function getHumanReadableLabel(value) {
     var mapping = {
+        'chaparral': 'Chaparral',
         'commercial': 'Commercial',
+        'desert': 'Desert',
         'forest': 'Forest',
         'grassland': 'Grassland',
         'hir': 'HIR',
         'lir': 'LIR',
         'pasture': 'Pasture',
         'row_crop': 'Row Crop',
+        'sg_prairie': 'Short Grass Prairie',
+        'tg_prairie': 'Tall Grass Prairie',
         'turf_grass': 'Turf Grass',
         'wetland': 'Wetland',
+
         'cluster_housing': 'Cluster Housing',
         'green_roof': 'Green Roof',
         'no_till_agriculture': 'No-Till Agriculture',
--- a/src/mmw/apps/core/templates/patterns.html
+++ b/src/mmw/apps/core/templates/patterns.html
@@ -1977,6 +1977,538 @@
                 </g>
             </g>
         </pattern>
+        <!-- Chaparral-->
+        <pattern height="72" id="fill-chaparral" overflow="visible" patternUnits="userSpaceOnUse" viewBox="79.59 -151.59 72 72" width="72">
+            <g>
+                <polygon fill="#9b9b9c" points="79.59,-79.59 151.59,-79.59 151.59,-151.59 79.59,-151.59"/>
+                <g>
+                    <polygon fill="none" points="79.59,-79.59 151.59,-79.59 151.59,-151.59 79.59,-151.59"/>
+                    <g>
+                        <line fill="none" stroke="#231F20" stroke-linejoin="round" stroke-width="0.3" x1="144.106" x2="162.106" y1="-162.105" y2="-144.105"/>
+                        <line fill="none" stroke="#231F20" stroke-linejoin="round" stroke-width="0.3" x1="144.106" x2="168.106" y1="-168.105" y2="-144.105"/>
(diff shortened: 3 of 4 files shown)


src/mmw/js/src/modeling/models.js
```diff
@@ -508,15 +508,20 @@ function getControlsForModelPackage(modelPackageName, options) {
...
         'grassland': 'Grassland',
         'hir': 'HIR',
         'lir': 'LIR',
         'pasture': 'Pasture',
         'row_crop': 'Row Crop',
+        'sg_prairie': 'Short Grass Prairie',
+        'tg_prairie': 'Tall Grass Prairie',
         'turf_grass': 'Turf Grass',
         'wetland': 'Wetland',
+
```
caseypt: Extra space here.

src/mmw/js/src/core/patterns.js
```diff
@@ -11,6 +11,11 @@ var colors = {
     'grassland'  : '#54d2d0',
     'row_crop'   : '#52c6dd',
     'wetland'    : '#4ebaea',
+    'chaparral'  : '#16161d',
+    'desert'     : '#16161d',
+    'sg_prairie' : '#16161d',
+    'tg_prairie' : '#16161d',
+
```
caseypt: Extra space here.

caseypt:
We may want to consider adding another column to this control. It's a bit long now with 5 rows and I could imagine it may get clipped on smaller screens. @ctaylo37 

selection_006

azaveaci:
Can one of the administrators verify this patch?

src/mmw/js/src/modeling/models.js
jamesmcclain: Here, I was trying to separate the land uses from the BMPs, as is done in patterns.js.  I did have an extra newline in patterns.js, though (two instead of one).

jamesmcclain:
I spoke a little bit with @mmcfarland and he advised me to keep the three columns (for now).

caseypt:
Sounds good. Should we keep `wetlands`?

mmcfarland:
If we have a NLCD supported class in tr55 then I say yes, we should keep wetlands.  Or rather, anything that doesn't have a corollary in tr55 should not be supported, and everything on the micro site should be since tr55 generated the dataset.

jamesmcclain:
That was another question that I had.  The issue calls for a 1-1 correspondence between the tools and the micro-app, but wetlands is present in the former but not the latter.

mmcfarland:
@ctaylo37 would be cool to get some new patterns for new options if you have time.

ctaylo37:
I could make some more patterns today.  I also think we should go to 4 columns.  Also, should we drop the footer with description for now since it isn't wired up, or leave it to explain how it will work?

caseypt:
+1 pending the wetlands question. Nice work. I was able to draw modifications with the new types, see the right name in the pop-up, and load a saved project and have the correct pattern be selected.

selection_009

caseypt:
I can take another look after the columns change is made.

caseypt:
I'd vote for keeping the footer. I'll create an issue to wire it up.

jamesmcclain:
Here is what it looks like now:
screenshot from 2015-06-17 11 26 53

azaveaci:
@azavea-bot rebuild

caseypt:
Looks good. I hope we can drop wetlands so we don't need the extra row.

jamesmcclain:
To the wetlands question: there are actually two different types of wetlands in TR-55: Herbaceous and Woody.

hectcastro:
@azavea-bot rebuild

ctaylo37:
I'm in favor of dropping wetlands and using the 12 types from the simple model to keep things consistent.  I'm also going to try to make the fill patterns more related to the simple model colors as well.

ctaylo37:
@caseypt or @jamesmcclain can you point me to where the fill pattern files are located?

caseypt:
@ctaylo37 https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/core/templates/patterns.html

ctaylo37:
Thanks @caseypt

jamesmcclain:
Is it conclusive that we are dropping wetland?  Another possible alternative is to find three more land cover types to complete the row (although I understand that that would mean three additional patterns that need to be created).

ctaylo37:
I would check with @mmcfarland on dropping wetlands.

mmcfarland:
I think the real answer is that we'll support 100% of modified TR-55 classes.  Given that these tools do nothing at the moment, we can drop wetlands to have a nicer UI and we'll address the real types when we make them do something.  I think what is most important now is that California teachers can see types that are relevant to their environment.  So, yes, we can drop - though we'll re-add later most likely as herbaceous and woody.

ctaylo37:
I will add the new patterns after this is merged.

jamesmcclain:
Okay, sounds good.  Should I just remove wetland from the UI, or should I remove all traces of it?

jamesmcclain:
Okay, "wetland" has been removed.
screenshot from 2015-06-17 14 20 50

jamesmcclain:
@caseypt Would you mind taking another look?

caseypt:
Testing now

caseypt:
+1

jamesmcclain:
Cool beans.  Merging now.

jamesmcclain:
Thanks everyone.

PR #437, opened 2015-06-17
Fix flash of white on bar chart when running model simulation

kdeloach:
This fixes a problem where the bar chart flashes white when you run a
new simulation.

Fixed by clearing the bar chart view only when polling has ended,
instead of clearing it when polling changes.

Test by running a few model simulations. The bar chart update should
look smoother.


lliss:
+1 tested. Looks good.

PR #441, opened 2015-06-17
Add mechanism for injecting custom configuration into AMI

hectcastro:
Take select settings from the stack configuration file, and thread them through to the Ansible provisioning process. In addition, use the new `t2.large` instance type for AMI creation vs. `m3.large`.


hectcastro:

deployment/packer/template.js
```diff
@@ -4,7 +4,9 @@
         "branch": "",
         "aws_region": "",
         "aws_ubuntu_ami": "",
-        "stack_type": ""
+        "stack_type": "",
+        "postgresql_password": "",
```
lliss: Should we null this out so that it is required to create the AMI?

deployment/ansible/group_vars/packer
```diff
@@ -2,7 +2,6 @@
 django_settings_module: "mmw.settings.production"
 
 postgresql_username: "modelmywatershed"
-postgresql_password: "{{ lookup('env', 'MMW_DB_PASSWORD') | default('modelmywatershed', true) }}"
 postgresql_database: "modelmywatershed"
```
kdeloach: Will we be moving the rest of these RDS setting to the yaml config eventually?

kdeloach:
+1

deployment/ansible/group_vars/packer
hectcastro: I think just this one, because it is the only one that's sensitive enough to be outside of the group variables.

deployment/packer/template.js
hectcastro: I don't think changes here will make Packer fail if variables aren't present. If we want to do that, I think it would have go to into the CLI tool.

hectcastro: Looking at that code, it will attempt to access the `RDSPassword` key and throw an exception if it doesn't exist.

lliss: Cool.

lliss:
+1

Issue #415, opened 2015-06-16
Allow Frame-Options to be open to multiple domains


mmcfarland:
ITSI cannot display 'https://mmw.azavea.com/' in an iframe because its  'X-Frame-Options' are set to 'SAMEORIGIN'.  We'll need to make this more liberal for ITSI integration.  Initially, this may be easiest to open to all domains, but we could potentially limit it to ITSI based ones.  Not sure if the header supports wildcards.

rajadain:
I think we will have to allow it for all domains by [removing the XFrameOptionsMiddleware from our config](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/base.py#L240). The `ALLOW-FROM` directive only supports a single URI, and moreover [is not supported by Chrome or Safari at all](https://developer.mozilla.org/en-US/docs/Web/HTTP/X-Frame-Options).

mmcfarland:
That's fine, since we're targeting a view for an iframe and have public projects, I think it may be a nice bonus feature to have embeddable iframes.

PR #445, opened 2015-06-17
Remove X-Frame-Options to allow iframing from ITSI

rajadain:
X-Frame-Options prevents clickjacking, but disables iframe embeds from other domains. Since we require ITSI Portal to be able to ember the app in an iframe, we remove it.

**Testing Instructions**

 * Open the home page and look at response headers for the first request and ensure that `X-Frame-Options` header is not present
 * Try opening the page from within an iframe: http://codepen.io/anon/pen/waqLxY

Connects #415

### DIFF
--- a/src/mmw/mmw/settings/base.py
+++ b/src/mmw/mmw/settings/base.py
@@ -237,7 +237,6 @@ MIDDLEWARE_CLASSES = (
     'django.middleware.csrf.CsrfViewMiddleware',
     'django.contrib.auth.middleware.AuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
-    'django.middleware.clickjacking.XFrameOptionsMiddleware',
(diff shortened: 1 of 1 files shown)


caseypt:
Checking this out now.

caseypt:
+1. Tested. Verified that `X-Frame-Options:SAMEORIGIN` is there on `develop` but not on this branch. Codepen works for me too.

rajadain:
It's a small enough change, but I wonder if I should merge it now or leave it hanging until we've cut the release, and merge it afterwards.

caseypt:
Not sure if it's a deal breaker for ITSI integration, but if it is, seems like we should merge it.

mmcfarland:
+1, nice small change.  Verified it was not working in an iframe prior and does work after.

mmcfarland:
I say merge.  It will let ITSI start playing around with their side of our integration.

Issue #443, opened 2015-06-17
Error loading saved projects


kdeloach:
Stack trace:

Uncaught TypeError: Cannot read property 'get' of undefinedmain.js:6968 isEditablemain.js:6393 Marionette.LayoutView.extend.templateHelpersvendor.js:5070 Marionette._getValuevendor.js:6198 Marionette.View.Backbone.View.extend.mixinTemplateHelpersvendor.js:6602 Marionette.ItemView.Marionette.View.extend._renderTemplatevendor.js:6573 Marionette.ItemView.Marionette.View.extend.rendervendor.js:7464 Marionette.LayoutView.Marionette.ItemView.extend.rendervendor.js:5651 Marionette.Region.Marionette.Object.extend.showmain.js:6226 Marionette.LayoutView.extend.reRendermain.js:6238 Marionette.LayoutView.extend.onShowvendor.js:5200 (anonymous function)vendor.js:6423 Marionette.View.Backbone.View.extend.triggerMethodvendor.js:5236 Marionette.triggerMethodOnvendor.js:5695 Marionette.Region.Marionette.Object.extend.showmain.js:4439 initViewsmain.js:4369 (anonymous function)vendor.js:23945 jQuery.Callbacks.firevendor.js:24057 jQuery.Callbacks.self.fireWithvendor.js:29110 donevendor.js:29451 jQuery.ajaxTransport.send.callback

caseypt:
When loading a project with one additional scenario from the default (so there are 3 total), the project loads, but some of the scenario tabs are missing.

kdeloach:
Scenarios are saving with the incorrect project ID.

PR #448, opened 2015-06-17
Prevent double saving projects on initialization.

lliss:
We were not able to restore saved projects because our initialization of the
project caused it to save not once but twice (when actually we did not want it
to save at all). This was caused by adding two scenarios to the new project.
Adding scenarios triggers an initial save. This is expected functionality however
adding two simultaneously caused two saves on the yet unsaved project which
caused two POST events rather than a POST and PUT. The scenarios were then
taking the last returned result and in most cases were getting the wrong project
id, which resulted in an error.

This prevents an initial save by adding the scenarios to the collection with the
silent option. Further, we introduce a flag on the project which is used to see
if a save has been attempted on the project and prevents saving if it has been
but an id has not yet been returned.

Connects #443

### DIFF
--- a/src/mmw/js/src/modeling/controllers.js
+++ b/src/mmw/js/src/modeling/controllers.js
@@ -50,6 +50,7 @@ var ModelingController = {
                 scenarios: new models.ScenariosCollection()
             });
 
+            // TODO evalutate if we can remove this global by reworking this code.
             App.currProject = project;
 
             project.get('scenarios').add([
@@ -61,7 +62,12 @@ var ModelingController = {
                     name: 'New Scenario',
                     active: true
                 })
-            ]);
+                // Silent is set to true because we don't actually want to save the
+                // project without some user interaction. This initialization
+                // should set the stage but we should wait for something else to
+                // happen to save. Ideally we will move this into the project
+                // creation when we get rid of the global.
+            ], { silent: true });
 
             project.on('change:id', function() {
                 router.navigate(project.getReferenceUrl());
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -108,16 +108,23 @@ var ProjectModel = Backbone.Model.extend({
         this.set('name', newName);
     },
 
+    // Flag to prevent double POSTing of a project.
+    saveCalled: false,
+
     saveProjectAndScenarios: function() {
         if (!App.user.loggedInUserMatch(this.get('user_id'))) {
+            // Fail fast if the user can't save the project.
             return;
         }
 
-        if (!this.get('id')) {
+        if (this.isNew() && this.saveCalled) {
+            // Fail fast if we are in the middle of our first save.
+            return;
+        } else if (this.isNew() && !this.saveCalled) {
             // We haven't saved the project before, save the project and then
-            // set the project ID on each scenario, then reattach the senarios
-            // to the project.
+            // set the project ID on each scenario.
             var self = this;
+            this.saveCalled = true;
             this.save()
                 .done(function() {
                     self.updateProjectScenarios(self.get('id'), self.get('scenarios'));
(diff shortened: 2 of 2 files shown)


src/mmw/js/src/modeling/models.js
```diff
@@ -108,16 +108,23 @@ var ProjectModel = Backbone.Model.extend({
         this.set('name', newName);
     },
 
+    // Flag to prevent double POSTing of a project.
+    saveCalled: false,
```
caseypt: Does [isNew](http://backbonejs.org/#Model-isNew) not work here?

lliss: From the docs
```
isNewmodel.isNew() 
Has this model been saved to the server yet? If the model does not yet have an id, it is considered to be new.
```
In our case we don't have an id but we have a pending request. So if we save again we get the double POST, we really want to create something like a lock until the first save comes back.

To be fair though I did not try it and assumed that checking isNew was about the same as checking if this.get('id')

rajadain:
+1 seems to work well!

 * Tried creating a new project while I was logged in and it did not save until I clicked "Save" or made a change
 * Tried creating a new project while logged out and it saved when I logged in
 * The project loaded correctly after a page refresh in both cases

Should we consider generalizing this approach to act like a mutex so that we are protected against simultaneous saves in general? Or would that be overkill at this point?

lliss:
Probably overkill. This is a weird case that shouldn't happen aside for certain scenarios that we can probably avoid under most circumstances.

src/mmw/js/src/modeling/models.js
```diff
@@ -108,16 +108,23 @@ var ProjectModel = Backbone.Model.extend({
...
+
     saveProjectAndScenarios: function() {
         if (!App.user.loggedInUserMatch(this.get('user_id'))) {
+            // Fail fast if the user can't save the project.
             return;
         }
 
-        if (!this.get('id')) {
+        if (!this.get('id') && this.saveCalled) {
```
rajadain: Consider replacing `!this.get('id')` with `this.isNew()` which is more expressive

src/mmw/js/src/modeling/models.js
caseypt: Got it. Yeah, that's what it does.

src/mmw/js/src/modeling/models.js
```diff
@@ -108,16 +108,23 @@ var ProjectModel = Backbone.Model.extend({
...
+            // Fail fast if we are in the middle of our first save.
+            return;
+        } else if (!this.get('id') && !this.saveCalled) {
             // We haven't saved the project before, save the project and then
-            // set the project ID on each scenario, then reattach the senarios
-            // to the project.
+            // set the project ID on each scenario.
             var self = this;
+            this.saveCalled = true;
```
caseypt: Are we using this for the first save? For example, do we not reset this after the save is complete?

caseypt: A more direct question: are we only expecting to use `saveProjectAndScenarios` once per session?

rajadain: I think the reasoning is that once a project has been saved, all further requests will be PUTs not POSTs. The `saveProjectAndScenarios` is used multiple times per session I believe. I've added #451 for a more general case.

caseypt: Got it. I'm seeing the flow now. Thanks.

src/mmw/js/src/modeling/models.js
mmcfarland: ^ This would make the flow a bit easier to read.

mmcfarland:
+1, projects are saving correctly again.  We may want to address this in a more sophisticated way as described by #451, but I think this is a good approach for now.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -61,7 +62,12 @@ var ModelingController = {
                     name: 'New Scenario',
                     active: true
                 })
-            ]);
+            ], { silent: true });
+            // Silent is set to true because we don't actually want to save the
+            // project without some user interaction. This initialization
+            // should set the stage but we should wait for something else to
+            // happen to save. Ideally we will move this into the project
+            // creation when we get rid of the global.
```
kdeloach: Great comment but consider moving it up one line.

kdeloach:
+1 Nice job figuring this out.

lliss:
Rebased to address the comments. Only functional change should be the use of `this.isNew()` over `!this.get('id')`
Might be worth another quick check.

caseypt:
+1. Tested.

PR #449, opened 2015-06-17
Change model input based on modifications.

caseypt:
* Adding modifications will now change the input to the model in
a deterministic way. Adding more modifications will mean more
commericial area is replaced with chaparral. This is so when
demoing the app, the model responds to user input.
* Increase the number cells in the base census to support more
modifications.

**To test:**
- Create a new project
- Draw a modification. You should see a increase in infiltration and evapotransipiration and a decrease in runoff.
- Continue to add modifications. The changes noted above should continue.

Removing modifications doesn't currently trigger a model run, but it looks like that is fixed in #412.

### DIFF
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -125,10 +125,10 @@ def make_gt_service_call_task(model_input):
 
     # TODO actually call geotrellis with model_input
     census = {
-        'cell_count': 8,
+        'cell_count': 100,
         'distribution': {
-            'c:commercial': {'cell_count': 5},
-            'a:deciduous_forest': {'cell_count': 3}
+            'c:commercial': {'cell_count': 70},
+            'a:deciduous_forest': {'cell_count': 30}
         },
         'modifications': [
             {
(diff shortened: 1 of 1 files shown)


caseypt:
Rebased. You can now test that removing modifications runs the model and updates the results. Also the water quality results will change based on the modifications.

src/mmw/apps/modeling/tasks.py
```diff
@@ -176,13 +176,42 @@ def run_tr55(census, model_input):
...
+
+# TODO: For demonstration purposes.
+def get_census_modifications(model_input):
+    """
+    For each modification that was drawn, replace
+    a cell of commercial with a cell of
+    chaparral. This has the effect of improving
+    infiltration and evapotranspiration.
+    """
+    count = len(model_input['modifications'])
```
mmcfarland: I added a ` * 10` to the end of this line for a more pronounced effect - was having a hard time seeing a change otherwise.  What do you think?

rajadain:
+1 adding modifications does have an effect on both Runoff and Water Quality outputs:

tr55-modifications

I wonder if we can find a combination that makes it more apparent for the demo.

src/mmw/apps/modeling/tasks.py
mmcfarland: Didn't see the recent comment about removing mods, that may have been what I was seeing re: lack of change.

rajadain:
Also tested: Removing a modification has an effect as well, as well as adjusting the slider.

src/mmw/apps/modeling/tasks.py
caseypt: There are 70 cells of commercial that are available for replacement. As it is now, it's a 1:1 replacement. Adding `* 10` will be a 10:1 replacement which will allow for 7 modifications. I'm not sure what you are planning to show, but that could work if you keep that limit in mind. Making it `* 5` might be a good trade-off.

mmcfarland: Sounds good.

caseypt:
Updated with a 4x multiplier to have a more noticeable effect on the modified model run.

caseypt:
One more change. Forgot that precipitation was stored in the modifications. This meant there was always one modification so you wouldn't get identical results if there were no drawn modifications. I now account for precipitation.

mmcfarland:
Code update looks good.

Issue #400, opened 2015-06-15
Add mechanism to dynamically pull the tiler host


hectcastro:
Currently, the tiler host is hardcoded to `tiles.tld.com`. This needs be discovered by the application servers dynamically based on environment.

Issue #360, opened 2015-06-11
Add ability to launch stacks namespaced by color


hectcastro:
Add ability to launch parallel application, tiler, and worker stacks that are namespaced by color.

PR #452, opened 2015-06-18
Add support for colors; dynamically set tiler host

hectcastro:
The tiler stack conditionally creates an ALIAS record based on stack color connecting its load balancer to the public hosted zone:

- `blue-tiles.mmw.azavea.com`
- `green-tiles.mmw.azavea.com`

The tile delivery network stack creates two CloudFront distributions, each with one of the previous DNS records as its origin. From there, the application stack accepts both CloudFront distribution endpoints via parameters and conditionally sets a launch configuration that overwrites `/etc/mmw.d/env/MMW_TILER_HOST` via cloud-config user data:

```yaml
#cloud-config

write_files:
  - path: /etc/mmw.d/env/MMW_TILER_HOST
    permissions: 0750
    owner: root:mmw
    content: d1kh2j2mft29y2.cloudfront.net
```

The content attribute above is based on color, and is threaded through from the tile delivery network stack.

Lastly, a Python DNS library was removed from the production requirements file because it was not being used.

Connects #360 and #400. Also, tagging @lliss and @kdeloach.


hectcastro:
@kshepard, we talked about this stuff a while back for CAC I think. Not sure what you settled on, but the `cloud-config` backdoor seems like it could work.

kshepard:
@hectcastro, this is great. We currently have a manual process of baking the desired CloudFront distribution endpoint into the AMI via group_vars, but using `cloud-config` as you did here is a much better solution. I'd love to get this into CAC if we have the time. Thanks for pointing it out.

kdeloach:
In NYC, the launch stacks script would detect the color of the next stack. Is this no longer the case?

deployment/cfn/application.py
```diff
@@ -270,29 +299,75 @@ def create_load_balancer(self, app_server_lb_security_group):
...
-                app_server_launch_config_name,
+                'lcAppServerBlue',
+                Condition='BlueCondition',
                 ImageId=Ref(self.app_server_ami),
                 IamInstanceProfile=Ref(self.app_server_instance_profile),
                 InstanceType=Ref(self.app_server_instance_type),
                 KeyName=Ref(self.keyname),
-                SecurityGroups=[Ref(app_server_security_group)]
+                SecurityGroups=[Ref(app_server_security_group)],
+                UserData=Base64(
```
kdeloach: Where could I find out more about how the UserData property works? The Cloudformation docs says this property should contain shell scripts, but it looks like we do something a little different here.

hectcastro:
Good point. I think that the stack detection for automatic deployment to staging will occur in an external script that Jenkins runs.

For production deployments, running the `launch-stacks` subcommand with the active color will evaluate the graph, but not reapply anything because the stack name is derived via a SHA256 of the stack's associated tags (stack type, color, etc.). If the non-active color is used, it'll spin up new stacks namespaced by that color.

deployment/cfn/application.py
hectcastro: User data can be shell scripts, but it can also be [`cloud-init`](http://cloudinit.readthedocs.org/en/latest/index.html) directives. Some more AWS specific details can be found [here](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/user-data.html#user-data-cloud-init).

Unfortunately, `cloud-init` has a history of never really being well defined, and each distribution runs a different version. That said, I still like it better than arbitrary shell scripts. Also, CoreOS has a nice [validator](https://coreos.com/validate/), but it only really works with its flavor of `cloud-init`.

kdeloach: Thanks, that's pretty cool. Since cloud-init only runs the first time you launch an instance, does that mean we need to relaunch a new app instance if there was a tiler change?

kdeloach:
When I try to launch stacks with the color omitted it does nothing (a blue stack already exists in staging):

```bash
./mmw_stack.py launch-stacks --aws-profile=mmw-stg --mmw-config-path=default.yml --mmw-profile=staging 
[DEBUG] VPC-34fed9de992df2f38f0276d2bd45 CREATE_COMPLETE
[INFO] VPC-34fed9de992df2f38f0276d2bd45 finished
[INFO] S3VPCEndpoint-23d722307cd4c23619 finished
[INFO] PrivateHostedZone-1e8810baf4b563 finished
[DEBUG] DataPlane-3b0b2ccb6182d7efd9217b UPDATE_COMPLETE
[INFO] DataPlane-3b0b2ccb6182d7efd9217b finished
[INFO] CachePrivateDNSRecord-a53abc2abf finished
```

If I specify `--stack-color=blue` if produces a little more output (but still doesn't do much):

```bash
./mmw_stack.py launch-stacks --aws-profile=mmw-stg --mmw-config-path=default.yml --mmw-profile=staging --stack-color=blue
[DEBUG] VPC-34fed9de992df2f38f0276d2bd45 CREATE_COMPLETE
[INFO] VPC-34fed9de992df2f38f0276d2bd45 finished
[INFO] S3VPCEndpoint-23d722307cd4c23619 finished
[INFO] PrivateHostedZone-1e8810baf4b563 finished
[DEBUG] DataPlane-3b0b2ccb6182d7efd9217b UPDATE_COMPLETE
[INFO] DataPlane-3b0b2ccb6182d7efd9217b finished
[INFO] CachePrivateDNSRecord-a53abc2abf finished
[DEBUG] Tiler-f658c60706d4277483b6671c67 CREATE_COMPLETE
[INFO] Tiler-f658c60706d4277483b6671c67 finished
[DEBUG] TileDeliveryNetwork-a9c314b5583b CREATE_COMPLETE
[INFO] TileDeliveryNetwork-a9c314b5583b finished
[DEBUG] Application-8dc3393b341c8e5c5500 CREATE_COMPLETE
[INFO] Application-8dc3393b341c8e5c5500 finished
[DEBUG] Worker-fde705a1f8cff3b8e6da85d8e CREATE_COMPLETE
[INFO] Worker-fde705a1f8cff3b8e6da85d8e finished
```

hectcastro:
The idea with that behavior is that not providing a stack color means you want to launch stacks that don't have a color attribute (data plane, VPC, private hosted zones, etc.). It effectively short circuits the part of the graph with color attributes (app, tiler, worker).

The second output has more output because the provided color avoids the short circuit mentioned above. It didn't do any work because the blue stack resources are already created.

What were your expectations? Does that explanation make sense?

kdeloach:
Configuration issues I ran into:
  * Empty string for bastion, app, and worker tried to create a stack with no AMI (had to comment out the AMI properties).
  * The Availability Zones in my config didn't match the subnets Availability Zones. (the default.yml uses a/b but our subnets use b/d)

Another issue I noticed is that launch-stacks wouldn't update the stack after updating settings in default.yml (after an error occurred launching the stack the first time). I had to delete the stacks first. Not sure if this is by design.

kdeloach:
> What were your expectations? Does that explanation make sense?

I was expecting the command to echo "I don't know which color stack to launch. Aborting..." but your explanation makes perfect sense.

deployment/cfn/application.py
hectcastro: That's correct.

kdeloach:
I was able to create green and blue stacks, the tiler DNS records pointed to the correct ELB, and the app instances had the correct cloudfront location specified in `MMW_TILER_HOST`.

kdeloach:
+1

hectcastro:
Attempted to address the configuration issues with https://github.com/WikiWatershed/model-my-watershed/commit/1d63a87ff94f3321db16a63de233af1c9f835221 and https://github.com/WikiWatershed/model-my-watershed/commit/fb810f062d4868ac52f46890174b364ea095f034.

As far as being able to update the stack, that isn't really handled by `majorkirby`, but can be done interactively via the CloudFormation UI. My preference would be to stand up new stacks when we need to make changes, but for quick one-off things, updates via the UI will always be there.

kdeloach:
Thanks, I wasn't suggesting that we needed to update the default.yml (just pointing out problems I ran into) but it can't hurt.

Issue #418, opened 2015-06-16
Add HUCs to boundary selector


mmcfarland:
Add these layers with these labels:
* HUC8: USGS Subbasin unit (HUC-8)
* HUC10: USGS Watershed unit (HUC-10)
* HUC12: USGS Subwatershed unit (HUC-12)

To the boundary level control.  Add the data as a fixture similarly to the congressional districts we have.

PR #453, opened 2015-06-18
Add HUC boundary layers

mmcfarland:
* Add references to 3 HUC Watershed Boundary Datasets
* Allow for specifying the order in which layers show up in the selector
* Use descriptive table aliases for helpful debugging

This PR doesn't include the data necessary to make this work, but I have gz sql files to create the tables and data for them, ~~for the reviewer (@caseypt, @lliss both have `huc08.tar.gz` it in your `/tmp` dir).  You should just be able to pipe the file to the psql~~

~~Keeping this as a WIP until we figure out how to distribute the data appropriately and also do some additional data work, notably to simply some features to reduce the load that exists on the tiler otherwise.~~

Connects #418

### DIFF
--- a/scripts/aws/setupdb.sh
+++ b/scripts/aws/setupdb.sh
@@ -3,6 +3,31 @@
 set -e
 set -x
 
+usage="$(basename "$0") [-h] [-d] \n
+--Sets up a postgresql database for MMW \n
+\n
+where: \n
+    -h  show this help text\n
+    -d  load/reload base data\n
+"
+
+load_data=false
+
+while getopts ":hd" opt; do
+    case $opt in
+        h)
+            echo -e $usage
+            exit
+            ;;
+        d)
+            load_data=true
+            ;;
+        \?)
+            echo "invalid option: -$OPTARG"
+            exit
+            ;;
+    esac
+done
 # Export settings required to run psql non-interactively
 export PGHOST=$(cat /etc/mmw.d/env/MMW_DB_HOST)
 export PGDATABASE=$(cat /etc/mmw.d/env/MMW_DB_NAME)
@@ -14,3 +39,13 @@ psql -c "CREATE EXTENSION IF NOT EXISTS postgis;"
 
 # Run migrations
 envdir /etc/mmw.d/env /opt/app/manage.py migrate
+
+if [ "$load_data" = "true" ] ; then
+	# Fetch boundary layer sql files
+	FILE_HOST="https://s3.amazonaws.com/data.mmw.azavea.com"
+	FILES=("boundary_huc12.sql.gz" "boundary_huc10.sql.gz" "boundary_huc08.sql.gz")
+
+	for f in "${FILES[@]}"; do
+	    curl -s $FILE_HOST/$f | gunzip -q | psql --single-transaction
+	done
+fi
--- a/src/mmw/apps/modeling/views.py
+++ b/src/mmw/apps/modeling/views.py
@@ -239,29 +239,38 @@ def _construct_tr55_job_chain(model_input, job_id):
 @decorators.api_view(['GET'])
 @decorators.permission_classes((AllowAny, ))
 def boundary_layers(request, table_id=None, obj_id=None):
+    layer_list = settings.BOUNDARY_LAYERS
+    layer_ids = [layer['code'] for layer in layer_list]
+
     if not table_id and not obj_id:
         tiler_prefix = '//'
         tiler_host = settings.TILER_HOST
         tiler_postfix = '/{z}/{x}/{y}'
         tiler_base = '%s%s' % (tiler_prefix, tiler_host)
 
-        def augment(index, dictionary):
-            retval = {}
-            retval['display'] = dictionary['display']
-            retval['tableId'] = index
-            retval['endpoint'] = urljoin(tiler_base, index + tiler_postfix)
-            return retval
+        def augment(dictionary):
+            code = dictionary['code']
+            return {
+                'display': dictionary['display'],
+                'tableId': code,
+                'endpoint': urljoin(tiler_base, code + tiler_postfix)
+            }
 
-        layers = [augment(i, d)
-                  for i, d in settings.BOUNDARY_LAYERS.items()]
+        layers = [augment(layer) for layer in layer_list]
         return Response(layers)
-    elif table_id in settings.BOUNDARY_LAYERS and obj_id:
-        # obj_id = str(int(obj_id))
-        table_name = settings.BOUNDARY_LAYERS[table_id]['table_name']
+
+    elif table_id in layer_ids and obj_id:
+        layers = filter(lambda l: l['code'] == table_id, layer_list)
+        table_name = layers[0]['table_name']
+        json_field = layers[0].get('json_field', 'geom')
+
+        query = 'SELECT {field} FROM {table} WHERE id = %s'.format(
+                field=json_field, table=table_name)
+
         cursor = connection.cursor()
-        query = 'SELECT geom FROM ' + table_name + ' WHERE id = %s'
         cursor.execute(query, [int(obj_id)])
         row = cursor.fetchone()
+
         if row:
             geojson = json.loads(GEOSGeometry(row[0]).geojson)
             return Response(geojson)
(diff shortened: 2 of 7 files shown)


lliss:
Looks great:
screencap

lliss:
The tiles are pretty simple. Have we considered pre-rendering them and storing them all in S3?

caseypt:
Re: storing the data, how about uploading the zip to S3 and have a makefile/management command to download it?

src/mmw/apps/modeling/views.py
```diff
@@ -227,29 +227,35 @@ def _initiate_tr55_job_chain(model_input, job_id):
 @decorators.api_view(['GET'])
 @decorators.permission_classes((AllowAny, ))
 def boundary_layers(request, table_id=None, obj_id=None):
+    layer_list = settings.BOUNDARY_LAYERS
```
caseypt: Code looks good, but I'm not totally sure of what the arguments are and what the conditionals are checking for exactly. Consider adding a docstring and/or some comments.

lliss:
+1

mmcfarland:
Re: pre-rendering tiles:  appealing option I've been thinking about, but we also need the utfgrid stuff for each, and at each zoom level for 5-12 layers, it might become a headache in terms of storing the national scale / zoom levels we need.  Also, since we need the geojson for each of these features, we'll also need them in the database.  Not sure if pre-rendering + duplicating the data will end up as a net-win.  I suspect not, especially since we can just basically cache the tiles forever.

mmcfarland:
re: storing the archives in s3 and having a command to get them in the db is, I think, the way to go.  Don't think that should be a part of this card though, but also don't want to merge this until the data is there.

mmcfarland:
This is ready for another round of reviews.  To test:

1. Run the updated `setupdb.sh` from your `app` vm to load datafiles into the database (these are currently being served from my dev machine, but will update the link to an S3 bucket after I iterate on any changes from review)
1. See 3 new layers in the boundary list, ordered according to their appearance in the settings file
1. Load the HUC layers, see that they render and that they allow being selected as an AoI

### Known Issues
* We may want to generate a smaller dev-only extent, but the db has been pretty stable and this may just be a once in while run - open to opinions
* HUC12 zoomed out to large geographic extents can start to perform poorly, even with simplified geometry.  I'm interested to see how it performs with the staging environment and CloudFront caching before considering alternatives, which could include:
    * Further simplifiying geometry
    * Only allow display at certain zoom levels
    * Pre-warming the cache for these troubled zoom level.
* The existing Congress District names and data should be ported to this new system, taken care of in a future issue.

mmcfarland:
Reminder to swap out internal URL to a publicly accessible S3 url before merging.

scripts/aws/setupdb.sh
```diff
@@ -14,3 +14,14 @@ psql -c "CREATE EXTENSION IF NOT EXISTS postgis;"
...
 # Run migrations
 envdir /etc/mmw.d/env /opt/app/manage.py migrate
+
+# Fetch boundary layer sql files
+FILE_HOST="http://santacruz.internal.azavea.com:9999"
+curl $FILE_HOST/boundary_huc08.sql.tar.gz | tar xz
+curl $FILE_HOST/boundary_huc10.sql.tar.gz | tar xz
+curl $FILE_HOST/boundary_huc12.sql.tar.gz | tar xz
+
+psql -f boundary_huc08.sql
```
mmcfarland: It's worth noting that this script leaves the `*.sql` files around, but they are happily overwritten on future runs from `tar xz`.  Uncompressed they can be several gigs, just wasn't sure if it was worth `rm`ing them.

mmcfarland:
@hectcastro Interested in your feedback on this.

scripts/aws/setupdb.sh
hectcastro: This approach appears to require a significant amount of RAM on the box issuing the `psql` command. We will probably need to come up with a streaming solution to avoid that overhead (a streaming approach may also help us avoid letting the tarballs touch disk).

Could we easily separate the SQL and data from these SQL files? I think then we might be able to steam the data in via `STDIN` using `COPY`. Unsure what format the geometries need to be in for that to happen. Might have to be something like WKT.

Were you able to get the import to work using these steps somewhere?

mmcfarland: Yeah, we can separate the two, and the current sql is performing a `COPY` (not `INSERT`s) within the `psql` session, so it shouldn't be too hard to extract that out to stream directly to `COPY` (the geometry is WKB). 

I boosted my services VM numbers before working on this to speed up the data updating, which could be why I didn't have any issues loading.

I'll take some measurements on memory usage and give it a shot to see if it reduces.

hectcastro: This is the type of command I was thinking:

```bash
$ curl -s $FILE_HOST/boundary_huc08_data.tar.gz | gunzip | psql -f boundary_huc08.sql
```

Let me know when you get the two separated and I can help with testing.

mmcfarland:
Updated load script to use fewer system resources while loading data.  Instructions and caveats remain the same.  Was able to load datasets multiple times in a row on a standard provisioned `app` VM, RAM usage stays fairly stable throughout.

scripts/aws/setupdb.sh
```diff
@@ -14,3 +14,11 @@ psql -c "CREATE EXTENSION IF NOT EXISTS postgis;"
 
 # Run migrations
 envdir /etc/mmw.d/env /opt/app/manage.py migrate
+
+# Fetch boundary layer sql files
+FILE_HOST="http://santacruz.internal.azavea.com:9999"
+FILES=("boundary_huc12.sql.gz" "boundary_huc10.sql.gz" "boundary_huc08.sql.gz")
```
hectcastro: It looks like the `mmw` username is embedded in these but the username on RDS is `modelmywatershed` because `mmw` is too short.

scripts/aws/setupdb.sh
```diff
@@ -14,3 +14,11 @@ psql -c "CREATE EXTENSION IF NOT EXISTS postgis;"
...
+
+for f in "${FILES[@]}"; do
```
hectcastro: Not worth changing, but just wanted to mention that since they increase in steps of 2, you could also loop against `seq`:

```bash
$ seq -f "%02g" 8 2 12
08
10
12
```

mmcfarland:
Updated the sql files hosted on santacruz to not include the owner in the export.  The db user running the import should become the owner when running the import.

```bash
➜  layersql  grep -i 'mmw' huc8.sql
➜  layersql  
```

```sql
mmw=# select tableowner from pg_tables where tablename = 'boundary_huc08';
 tableowner 
------------
 mmw
```

mmcfarland:
Still unable to reproduce the reported memory errors, even on a standard `app` vm:

```bash
vagrant@app:~$ cat /proc/meminfo 
MemTotal:        1017836 kB
MemFree:          485176 kB
```

hectcastro:
Owner changes work, and the memory pressure is no longer an issue. I uploaded the 3 HUC files to S3 and applied them on staging without issue:

```
modelmywatershed=> select count(*) from boundary_huc08;
 count
-------
  2116
(1 row)

modelmywatershed=> select count(*) from boundary_huc10;
 count
-------
 15939
(1 row)

modelmywatershed=> select count(*) from boundary_huc12;
 count
-------
 86024
(1 row)
```

mmcfarland:
Updated `setupdb.sh` with:
* New S3 URLs for data files
* Flag (`-d`) for specifying to load the data, which takes a significant amount of time and won't need to be done frequently.

caseypt:
This may not be relevant anymore, but in regards to https://github.com/WikiWatershed/model-my-watershed/pull/453#issuecomment-113231373, on thing I've done in the past when I needed GeoJSON with static tiles was to store the GeoJSON as an attribute that comes back with the UTFGrid response.

I got the idea from this project: http://graphics.latimes.com/la-mayoral-maps/#12/34.0895/-118.4482

All the tiles here are pregenerated, but they are able to highlight the selected ward on hover because the GeoJSON is an attribute in the UTFGrid.

hectcastro:
Just finished running `setupdb.sh` on staging. Worked great again:

```
modelmywatershed=> select count(*) from boundary_huc08;
 count
-------
  2116
(1 row)

modelmywatershed=> select count(*) from boundary_huc10;
 count
-------
 15939
(1 row)

modelmywatershed=> select count(*) from boundary_huc12;
 count
-------
 86024
(1 row)
```

Issue #440, opened 2015-06-17
Add mechanism to prune stale staging AMIs


hectcastro:
All staging AMIs are tagged with `Environment: Staging`. We should only be keeping the last 10 or so of these. Created mechanism should be added as a post-build action to the AMI creation Jenkins jobs.

PR #457, opened 2015-06-18
Add support for EC2 AMI pruning by service type

hectcastro:
A CLI subcommand has been added to deregister AMIs and remove root volume snapshots. See `README` additions for more details.

Connects #440. Also, tagging @kdeloach and @lliss.

### DIFF
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -15,7 +15,7 @@ Next, install [Packer](https://packer.io/) using the steps detailed on Packer's 
 
 ## Create Identity and Access Management (IAM) roles
 
-TBD
+See [Create an IAM Instance Profile for Your Amazon EC2 Instances](http://docs.aws.amazon.com/codedeploy/latest/userguide/how-to-create-iam-instance-profile.html).
 
 ## Configure an AWS Profile using the AWS CLI
 
@@ -39,10 +39,18 @@ Stack launching is managed with `mmw_stack.py`. This command provides an interfa
 
 Before launching the Model My Watershed stack, AMIs for each service need to be generated:
 
-
 ```bash
 $ ./mmw_stack.py create-ami --aws-profile mmw-stg --mmw-profile staging \
                             --machine-type mmw-{app,monitoring,tiler,worker}
+```
+
+### Pruning AMIs
+
+After creating several AMIs, older ones become stale and are no longer needed. To prune them use:
+
+```bash
+$ ./mmw_stack.py prune-ami --aws-profile mmw-stg --mmw-profile staging \
+                           --keep 5 --machine-type mmw-{app,monitoring,tiler,worker}
 ```
 
 ### Launching Stacks
--- a/deployment/ec2/amis.py
+++ b/deployment/ec2/amis.py
@@ -1,0 +1,43 @@
+"""Helper functions to handle EC2 related operations with Boto"""
+
+import boto
+import logging
+
+
+LOGGER = logging.getLogger('mmw')
+LOGGER.addHandler(logging.StreamHandler())
+LOGGER.setLevel(logging.INFO)
+
+MACHINE_TYPE_MAPPING = {
+    'mmw-app': 'Application',
+    'mmw-monitoring': 'Monitoring',
+    'mmw-tiler': 'Tiler',
+    'mmw-worker': 'Worker',
+}
+
+
+def _prune_ami(ami):
+    """Actually deregister AMI and its associated snapshot"""
+    LOGGER.info('Identified that [%s] is eligible to be pruned..', ami.id)
+
+    ami.deregister(delete_snapshot=True)
+
+    LOGGER.info('Deregistered [%s] and snapshot [%s]..', ami.id,
+                ami.block_device_mapping['/dev/sda1'].snapshot_id)
+
+
+def prune(mmw_config, machine_type, keep, aws_profile):
+    """Filter owned AMIs by machine type, environment, and count"""
+    stack_type = mmw_config['StackType']
+
+    conn = boto.connect_ec2(profile_name=aws_profile)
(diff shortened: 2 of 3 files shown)


deployment/README.md
```diff
@@ -39,12 +39,20 @@ Stack launching is managed with `mmw_stack.py`. This command provides an interfa
...
                             --machine-type mmw-{app,monitoring,tiler,worker}
 ```
 
+### Pruning AMIs
+
+After creating several AMIs, older ones become stale and are no longer needed. To prune them use:
+
+```bash
+$ ./mmw_stack.py prune-ami --aws-profile mmw-stg --mmw-profile staging \
+                           --keep 5 --service {Application,Monitoring,Tiler,Worker}
```
kdeloach: For `create-ami` we specify `--machine-type=mmw-app` but for `prune-ami` we specify `--service=Application`. Any to way to make it the same for both, or is it not worth it?

deployment/ec2/amis.py
```diff
@@ -0,0 +1,35 @@
...
+    stack_type = mmw_config['StackType']
+
+    conn = boto.connect_ec2(profile_name=aws_profile)
+    images = conn.get_all_images(owners='self', filters={
+        'tag:Service': service,
+        'tag:Environment': stack_type
+    })
+
+    if len(images) > keep:
+        map(_prune_ami, reversed(sorted(images[0:len(images) - keep])))
```
kdeloach: If `keep` is 5 then this would remove the 5 oldest amis instead of keeping the 5 newest amis. May want to do something like this instead `list(reversed(sorted(images)))[0:keep]`?

deployment/README.md
hectcastro: You're right. Addressed in https://github.com/WikiWatershed/model-my-watershed/commit/4253e5264045890a03ed514a12a4310c6fd07518.

deployment/ec2/amis.py
hectcastro: Had to put this to the test:

```
ami-cdce3ca6 - mmw-worker-1434571992-d7d3da6c46f0655b700b00530fdf3ec0585e11e4
ami-17e31e7c - mmw-worker-1434643973-797e66613623850c32509d7b8cf9597b04dc2307
ami-abdd20c0 - mmw-worker-1434645344-2b4056d1c3b94138622f2f080253868da11ccc1f
ami-21da274a - mmw-worker-1434648388-dbb3d148d9242a23a4c1fc8fb480e24bab4257fd
ami-efd42984 - mmw-worker-1434649432-4d2647e62c08fd4975a3d564eed6fd3269b55825
ami-43d22f28 - mmw-worker-1434651785-9b5384dcb80ac368b59d8ce2d7a8d1cf8ab7afa4

Old:

Identified that [ami-cdce3ca6] is eligible to be pruned..
Deregistered [ami-cdce3ca6] and snapshot [snap-33ccca72]..

New:

Identified that [ami-43d22f28] is eligible to be pruned..
Deregistered [ami-43d22f28] and snapshot [snap-21b19357]..
Identified that [ami-efd42984] is eligible to be pruned..
Deregistered [ami-efd42984] and snapshot [snap-dfdd53ac]..
Identified that [ami-21da274a] is eligible to be pruned..
Deregistered [ami-21da274a] and snapshot [snap-26f08147]..
Identified that [ami-abdd20c0] is eligible to be pruned..
Deregistered [ami-abdd20c0] and snapshot [snap-bbb720ca]..
Identified that [ami-17e31e7c] is eligible to be pruned..
Deregistered [ami-17e31e7c] and snapshot [snap-bd97ebd9]..
```

Here, "old" represents:

```python
map(_prune_ami,
    list(sorted(images, key=lambda i: i.creationDate))[0:len(images) - keep]) 
```

And "new" represents:

```python
map(_prune_ami,
    list(reversed(sorted(images, key=lambda i: i.creationDate)))[0:keep])
```

So, I think there were two bugs with the original:

1. There is no need for `reversed`
2. `sorted` needs to be explicitly told to use `name` or `createdDate` to sort properly

Should be addressed in https://github.com/WikiWatershed/model-my-watershed/commit/1bab4920405a42fb9b712caf96489f0985980a23.

deployment/ec2/amis.py
```diff
@@ -0,0 +1,43 @@
...
+
+    conn = boto.connect_ec2(profile_name=aws_profile)
+    images = conn.get_all_images(owners='self', filters={
+        'tag:Service': MACHINE_TYPE_MAPPING[machine_type],
+        'tag:Environment': stack_type
+    })
+
+    if len(images) > keep:
+        map(_prune_ami,
+            list(sorted(images, key=lambda i: i.creationDate))[0:len(images) - keep])  # NOQA
```
kdeloach: I think it definitely makes more sense to sort on a key like this, instead of doing sort+reverse, but is the slicing correct?

I could be misunderstanding what `keep` should do, but here is an assertion of how I would expect it to work (line 4).

```python
In [1]: images = range(15)
In [2]: keep = 5
In [3]: result = images[0:len(images) - keep]
In [4]: assert len(result) == keep
---------------------------------------------------------------------------
AssertionError
```

hectcastro: `keep` is the number of AMIs we want to be kept, but the output list is what's getting passed to `_prune_aim` to get axed.

kdeloach: That totally makes sense. Sorry for the confusion.

kdeloach:
+1

deployment/ec2/amis.py
hectcastro: Helped catch some bugs early, so net positive in my book.

lliss:
Looks good. I'm going to try it out shortly.

lliss:
+1.

PR #458, opened 2015-06-18
Prevent leafletlocate jshint error.

jamesmcclain:
Leafletlocate is required to be included in `core/views.js`, but is never explicitly used.  This prevents `jshint` from flagging this.

### DIFF
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -2,7 +2,6 @@
 
 var $ = require('jquery'),
     L = require('leaflet'),
-    leafletLocate = require('leaflet.locatecontrol'),
     _ = require('lodash'),
     router = require('../router.js').router,
     Marionette = require('../../shim/backbone.marionette'),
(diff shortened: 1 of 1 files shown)


src/mmw/js/src/core/views.js
```diff
@@ -2,7 +2,7 @@
 
 var $ = require('jquery'),
     L = require('leaflet'),
-    leafletLocate = require('leaflet.locatecontrol'),
+    leafletLocate = require('leaflet.locatecontrol'), // jshint ignore:line
```
kdeloach: The design pattern we follow for require statements that mutate the environment (but don't return anything useful) is to require without assigning to a variable (Ex: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/setup.js#L15).

kdeloach:
+1

jamesmcclain:
Many thanks.

jamesmcclain:
@azaveaci rebuild

hectcastro:
@azavea-bot rebuild

PR #459, opened 2015-06-18
Don't clear charts when polling starts

lewfish:
To test: 
 * In modeling view, change precip level. 
 * The current chart should not clear before the new one appears.


kdeloach:
+1

jamesmcclain:
+1, tested

rajadain:
+1 tested

PR #456, opened 2015-06-18
Show loading and error messages in Analyze view

lewfish:
To test:
 * Go to the analyze view
 * Check that there is a centered spinner that says Loading.
 * Check that results appear after a few seconds
 * Repeat, except right when spinner starts, switch the network to offline. (To switch the network to offline, in Developer Tools, click the smartphone icon in the upper left corner. Then back in the browser, there should be a network dropdown on the top.)
 * Check that an error message with icon shows. 
 * With network back to normal, check that you can go to the Analyze view, wait for results to show, and then go to the modeling view, and back. The results should show immediately.

I've only had time to fix the existing unit tests for the Analyze view. I also was unable to get the new animation for the Analyze view to work due to complicated CSS issues. I'll make issues for these in a minute.

Connects #80

Issue #103, opened 2015-04-27
Bar chart is initially wrong size


lewfish:
The bar chart is initially the wrong size (defaulting to 200px). At first I thought this was because the animation was still going when the chart reads the size of the container it is in, but I no longer think this is true because it also happens when switching tabs after the animation is done.

PR #472, opened 2015-06-18
Refresh bar chart on animateIn

lewfish:
The chart in the Analyze view is initially the correct size as a benevolent side effect of https://github.com/WikiWatershed/model-my-watershed/pull/456

However, the chart is still the wrong size when hitting back from the Modeling view. This PR fixes that problem.

Connects #103

### DIFF
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -14,8 +14,7 @@ var $ = require('jquery'),
     tableRowTmpl = require('./templates/tableRow.html'),
     tabPanelTmpl = require('./templates/tabPanel.html'),
     tabContentTmpl = require('./templates/tabContent.html'),
-    barChartTmpl = require('../core/templates/barChart.html'),
-    TransitionRegion = require('../../shim/marionette.transition-region');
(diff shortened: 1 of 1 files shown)


src/mmw/js/src/analyze/views.js
```diff
@@ -24,11 +23,29 @@ var AnalyzeWindow = Marionette.LayoutView.extend({
     regions: {
         headerRegion: '#analyze-header-region',
         detailsRegion: {
-            regionClass: TransitionRegion,
             el: '#analyze-details-region'
         }
     },
 
+    /*
+     Any idea why this doesn't work?
```
lliss: It appears that items in the events hash only work on jquery events: http://backbonejs.org/#View-delegateEvents

Part of the reason this seems to not be functioning properly is that you are using the `trigger` method, which is a pure jQuery thing, whereas `triggerMethod` is the marionette method to trigger special event callbacks.

lliss:
I don't see any reason to leave in the block of commented out code, but this does work and I have tested it. I think we might want to do a quick audit at some point and figure out when and where trigger is used over triggerMethod and decide if that is what we both expect and want to have happen. +1

PR #477, opened 2015-06-19
Make all VMs default to 1024 Mb of RAM.

lliss:
To reduce OOM issues on our development stack, increase the alloted RAM on all
the VMs to 1024 Mb.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -141,6 +141,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
     app.ssh.forward_x11 = true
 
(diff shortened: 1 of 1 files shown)


caseypt:
Reprovisioning on this branch.

hectcastro:
Just a note that everyone will have to at the very least do a `vagrant reload` to see the memory allocations change on the virtual machines.

caseypt:
+1. Rebuilt my environment and every machine has 1024 mbs of RAM.

kdeloach:
+1 Thanks.

Issue #479, opened 2015-06-19
Turn off geolocation tracking


mmcfarland:
The "geolocate me" feature should just provide and initial coordinate to set the initial extent of the map.  Currently, the site tracks the user location, and after about a minute of use, the page will re-navigate to the users location, disrupting the view they've constructed.

Should be considered with #460 (and may be the same fix).

rajadain:
I can still reproduce this on Staging. Should I reopen and move this to the backlog?

mmcfarland:
Yeah, could you reopen and list the steps you use to recreate it? I haven't been able to.

rajadain:
The steps to reproduce are the following:

 1. Navigate to a far away place, like New York, and draw an area of interest
 2. Navigate to a different tab in Chrome and wait 30 - 60 seconds
 3. Navigate back to first tab, move mouse (but don't click), or hover over the navigation icon:  
  ![pasted_image_7_2_15__12_31_pm](https://cloud.githubusercontent.com/assets/1430060/8482263/8571703a-20b6-11e5-805a-444846aa6de7.png)
 4. You will be brought back to your own location

lewfish:
We think this is being caused by leaflet.locatecontrol.

mmcfarland:
Yeah, I've used the navigator function a bunch without experiencing this issue, suspecting the leaflet plugin seems reasonable.

lewfish:
I can't reproduce this, so moving on to another task for now.

rajadain:
I can reproduce this on my machine, will take a stab at it after I finish my current task.

Issue #460, opened 2015-06-18
Don't geolocate if landing on `/model`


rajadain:
Automatic geolocaiton is useful for new users just entering the system, but harmful if users land on a project with an already defined AoI and are zoomed away from it.

For example, see here: `<staging_url>/model/40/scenario/46`

The map is recentered if the results window is closed and opened, but not if users switch between tabs.

mmcfarland:
I suspect this is not actually the geolocation, but the issue brought up in #401 about fitting bounds.

rajadain:
Actually I think the "bug" of #401 is what allows the user to find the right AoI in this case. If you go to that URL, you will see that the AoI is actually around Facebook headquarters in Menlo Park, but if you land on it the map is centered around Philadelphia (if you have allowed geolocation that is).

mmcfarland:
Not sure I'm seeing that behavior.  Going to `/model/40/scenario/46` shows me an active tab and the AoI + modifications in Menlo Park, the browser is reporting that the page is tracking my location.  If you then zoom away and toggle the results window, you're brought back to centered on the AoI because of #401.  Are you seeing something different?

mmcfarland:
@rajadain showed me and verified this behavior.

caseypt:
Shares a point with #479

PR #486, opened 2015-06-19
Prevent geolocation from repositioning late

lewfish:
To test: 
 * Change the maxAge constant to 0, so the geolocator will not use a cached version of your position. We want to force it to get a new position which takes some time.
 * Go to URL for saved model
 * It shouldn't change the map position to your geolocated position. (You can verify that the geolocator is getting a position, but just not updating the map by setting a breakpoint in the `geolocation_success` function)
 * Repeat, except this time go to the home page and quickly drag the map to change the position. The geolocator shouldn't change the position.
 * Go to the home page and wait a few seconds. It should change the position to your geolocated position.
 
Connects #460 
Connects #479

### DIFF
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -147,7 +147,9 @@ var MapView = Marionette.ItemView.extend({
             areaOfInterestLayer = new L.FeatureGroup(),
             modificationsLayer = new L.FeatureGroup(),
             maxZoom = 10,
-            maxAge = 60000;
+            maxAge = 60000,
+            timeout = 30000,
+            self = this;
 
         map.addControl(new L.Control.Zoom({position: 'topright'}));
         addLocateMeButton(map, maxZoom, maxAge);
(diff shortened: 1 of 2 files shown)


src/mmw/js/src/core/views.js
```diff
@@ -164,11 +166,15 @@ var MapView = Marionette.ItemView.extend({
         this._areaOfInterestLayer = areaOfInterestLayer;
         this._modificationsLayer = modificationsLayer;
 
+        this.model.set('disable_geolocation', false);
```
rajadain: Consider renaming the attribute so the behavior is not a double negative. For example, `geolocation_enabled`, which would be easier to read in the `if` condition below. `if (geolocation_enabled)` is easier to read than `if(!disable_geolocation)`.

Also consider moving the initialization to the `defaults` in `models.js`.

rajadain:
+1 tested. Good work!

 * Tested geolocation by default which still works
 * Tested landing on a project and wasn't drawn away
 * Moved the map before geolocation and it didn't draw me away

Issue #469, opened 2015-06-18
Current Conditions should not have "modified" output


mmcfarland:
Current Conditions will never have modified content, so the two bars on the graph will always be the same.  We should remove "modified".

![screenshot from 2015-06-18 17 31 18](https://cloud.githubusercontent.com/assets/1014341/8242781/edd4a584-15df-11e5-8efd-a74735b7de22.png)

PR #488, opened 2015-06-22
Don't show modified bar if Current Conds

lewfish:
To test:
Open the modeling view and check that in the Runoff tab for Current Conditions there is only one bar with no label on the x axis. For other scenarios, there should be Original and Modified bars as before.

@mmcfarland I removed the 'Original' label for Current Conditions, since there is only one bar, but I could add it back if you want.

Connects #469


jamesmcclain:
+1 Tested.  I looked over the code exercised the program to the best of my ability.

The only potential issue that I saw as that the when there is a single stacked bar chart, it is double-width.  As long as this is understood/intentional, then I do not detect any problems.

Issue #470, opened 2015-06-18
User list is exposed via API


mmcfarland:
https://mmw.azavea.com/api/users/

We should disable the list of user names in the system from being displayable.

PR #489, opened 2015-06-22
Remove api/users route

lewfish:
To test, go to http://localhost:8000/api/users. You should get a 404 response where you used to see a list of users.

Connects #470

### DIFF
--- a/src/mmw/mmw/urls.py
+++ b/src/mmw/mmw/urls.py
@@ -5,7 +5,6 @@ from __future__ import division
 
 from django.conf.urls import patterns, include, url
 from django.contrib import admin
-from rest_framework import routers
 
 import registration.backends.default.urls
 import watchman.urls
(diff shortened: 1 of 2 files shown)


lewfish:
@azaveaci rebuild

src/mmw/mmw/urls.py
```diff
@@ -20,13 +19,9 @@
 
 admin.autodiscover()
 
-router = routers.DefaultRouter()
-router.register(r'users', apps.home.views.UserViewSet)
-
 urlpatterns = patterns(
     '',
     url(r'^', include(apps.home.urls)),
-    url(r'^api/', include(router.urls)),
```
kdeloach: Interesting how our usage of DRF evolved. Instead of using the Viewset API we now just use the DRF decorators/serializers with generic Django view functions. Maybe we'll reap the benefits of the Viewset API on the next project...

kdeloach:
+1

Issue #468, opened 2015-06-18
Register django models with Admin


mmcfarland:
All of our data models should be exposed to the Django Admin so that it's trivial to do quick data fixes during demos and tests.

PR #490, opened 2015-06-22
Register models with Admin

lewfish:
To test:
 * Create superuser account (run `./scripts/manage.sh createsuperuser`)
 * Login as super user at http://localhost:8000/admin/
 * Check that Jobs, Districts and ItsiUser models can be viewed. The label for each Job model should be its uuid, and for ItsiUser its username.

Connects #468


caseypt:
+1. Verified that all of the models are now registered in the admin and that they are being displayed correctly.

Issue #484, opened 2015-06-19
Serve /version.txt from web accessible service endpoints


hectcastro:
Thread through the Git SHA used to generate an AMI into `/version.txt`. This file probably needs to go into a consistent location across all AMI types, and then exposed via Nginx directly.

kdeloach:
Great idea.

PR #492, opened 2015-06-22
Add support for detecting AMI SHA via HTTP

hectcastro:
During the Packer provisioning steps, the version (usually a Git SHA) passed through is written to a file inside of the machine image. The application and tile server Nginx configuration files have been updated to map `/version.txt` to this file so that external requests can provide AMI version information.

Connects #484


kdeloach:
Should the version sha1 for the app and tiler endpoints always yield the same value? Or can you rebuild app without rebuilding the tiler?

hectcastro:
They will be able to have different values because some AMI builds will only happen for application code changes vs. tiler code changes. In some scenarios, they'll be the same though.

kdeloach:
+1

Issue #435, opened 2015-06-17
Wire-up modification tool footer


caseypt:
It should update when hovering over the different modification types.

![selection_010](https://cloud.githubusercontent.com/assets/1042475/8211080/26ec2174-14e3-11e5-8683-c29724248a12.png)

PR #493, opened 2015-06-23
Show summary of modification on hover

lewfish:
To test:
 * In Modeling view, in Land Uses drop-down, drag the mouse over different Land Use icons. 
 * The titles and summaries should update properly, and clear when the cursor is not directly over an icon. Check the grammar / spelling of the text.
 * Repeat with the Conservation Practices. We don't have the summaries yet (see issue https://github.com/WikiWatershed/model-my-watershed/issues/494), but for testing purposes 'Green Roof' has a summary which just says 'Green Roof Summary...'. This should be removed before merging. 
 * Check that clicking a modification allows you to draw a polygon.

Connects #435


src/mmw/js/src/modeling/templates/controls/landCover.html
```diff
@@ -7,29 +7,27 @@
...
         <div class="pad-1"> <!-- Dropdown Content -->
             <ul class="row">
-                {{ utils.thumb(label, 'lir') }}
-                {{ utils.thumb(label, 'hir') }}
-                {{ utils.thumb(label, 'commercial') }}
-                {{ utils.thumb(label, 'forest') }}
+                {{ utils.thumb(labelFn, 'lir') }}
+                {{ utils.thumb(labelFn, 'hir') }}
+                {{ utils.thumb(labelFn, 'commercial') }}
+                {{ utils.thumb(labelFn, 'forest') }}
```
rajadain: Thanks for doing this, makes it more readable.

src/mmw/js/src/modeling/modificationDescriptions.json
```diff
@@ -0,0 +1,73 @@
+{
+    chaparral: {
```
rajadain: All keys and values must be enclosed in double quotes in JSON files. http://json.org/

src/mmw/js/src/modeling/templates/controls/utils.html
```diff
@@ -1,7 +1,7 @@
 {% macro thumb(label, value, size=56) %}
     <li>
         <div class="thumb" data-value="{{ value }}">
-            <svg>
+            <svg height="{{ size }}" width="{{ size }}">
```
rajadain: Why did we need to add this? Genuinely curious.

lewfish: Without the height/width attributes I added, the svg was much wider than the rect and div (.thumb) and when the mouse was over the svg, it would propagate mouseenter events to the div, which is not what we want.

rajadain: Ah I see. Thanks for clarifying!

rajadain:
Consider moving the colors from `patterns.js` to the new `modificationDescriptions.json`, and the `getDrawOpts` function to `core/views.js` to further consolidate modification configuration.

rajadain:
From a design perspective, the leading in the description feels a little tight:

image

We might consider slightly increasing the line spacing. /cc @ctaylo37

rajadain:
Also, from a design perspective, how do people feel about not clearing the description on mouse leave? It makes transitions less jarring and is more forgiving of the user's mouse movements.

With clear upon mouse leave:

modifications-hover

Without clear upon mouse leave:

modifications-hover-fixed

Also, please note that since we have switched from `LIR` and `HIR` to `Low Intensity Residential` and `High Intensity Residential`, part of the text is clipped in the thumbnail.

lliss:
I think not clearing does look a little nicer.

ctaylo37:
I think the description text definitely needs more line height.  How does it look if you wrap it in a <P> tag, or even a <p><small> set of tags?  I also agree with @rajadain about not clearing the description on mouse leave.  Is there a specific reason why we can't use "LIR and  HIR" for the thumbnail and the full name in the description?

src/mmw/js/src/modeling/templates/controls/summary.html
```diff
@@ -0,0 +1,3 @@
+<h2 class="dark">{{label}}</h2>
+<br>
+<span>{{summary}}</span>
```
lliss: From a semantic HTML point of view I don't like using <br> tags for spacing. They're really for things like poems:

```html
<p>Sing in me, Muse, and through me tell the story<br>
of that man skilled in all ways of contending,<br>
the wanderer, harried for years on end,<br>
after he plundered the stronghold<br>
on the proud height of Troy.</p>

File: src/mmw/js/src/modeling/controls.js [hunk 5]
```diff
@@ -57,29 +59,70 @@ var DrawControlView = ControlView.extend({
...
+    events: _.defaults({
+        'mouseenter @ui.thumb': 'setThumbValue',
+        'mouseleave @ui.thumb': 'clearThumbValue'
+    }, DrawControlView.prototype.events),
+
+    templateHelpers: {
+        labelFn: models.getHumanReadableLabel
+    },
+
+    setThumbValue: function(event) {
```
lliss: I think there are some good reasons not to use the word `event` as an parameter or variable name.
https://stackoverflow.com/questions/1510094/is-event-a-reserved-word-in-javascript
I think we generally go with `e` or `evnt`.
Others may feel differently though.

src/mmw/js/src/modeling/controls.js
```diff
@@ -57,29 +59,70 @@ var DrawControlView = ControlView.extend({
...
+        this.model.set('thumbValue', $(event.currentTarget).data('value'));
     },
 
+    clearThumbValue: function() {
+        this.model.set('thumbValue', null);
```
lliss: As discussed, since this is far away from the model, I'd consider making a setter on the model and explaining in a comment what the value is used for.

src/mmw/js/src/modeling/controls.js
caseypt: Yeah I'd recommend staying away from it. I usually go with `e`..

src/mmw/js/src/modeling/templates/controls/summary.html
caseypt: +1

lliss:
Also, support more line-height on the descriptions as mentioned by @rajadain .

lewfish:
I've responded to all feedback. The styling of the summary might still need some work. Maybe the text should be a little bigger and not grey? And maybe there shouldn't be as big a space above the name of the modification (the name above the summary, not below the thumbnail)?

rajadain:
+1 looks much nicer now! Spacing to the title can be adjusted via CSS (`padding-bottom` or something), but can probably be left to a later task. I think it looks good enough and works well for now, and @ctaylo37 can work his magic and add more polish in time.

Also, good job on adding Short Names and Long Names to the modifications. We definitely needed the variations. I like how it falls back to Long Name if no Short Name is found, so we don't have to repeat ourselves for cases like "Desert".

src/mmw/js/src/modeling/modificationConfig.json
```diff
@@ -0,0 +1,98 @@
...
+        "summary": "This category includes areas where grasses and weedy (herbaceous) plants cover more than 80% of the ground area. These areas are not intensively managed but may be used for grazing animals.",
+        "color": "#54d2d0"
+    },
+    "hir": {
+        "name": "High Intensity Residential",
+        "shortName": "HIR",
+        "summary": "This category includes areas with apartment complexes or row houses. The amount of area that is covered by hard surfaces like roads, driveways and roofs is between 80% and 100%. No more than 20% of the area is covered with vegetation.",
+        "color": "#39afa1"
+    },
+    "lir": {
```
rajadain: I like how almost all the modification configuration is now present in one file. Maybe someday we'll be able to get the patterns in here too. But this is great for now.

src/mmw/js/src/modeling/modificationConfigUtils.js
```diff
@@ -0,0 +1,61 @@
+"use strict";
+
+var modificationConfig = require('./modificationConfig.json');
+
+// modKey should be a key in modificationsConfig (eg. 'turf_grass').
+function getHumanReadableName(modKey) {
```
rajadain: `modKey` is indeed more explicit than `value`, good thinking

lliss:
+1. Tested. Looks much better.

Issue #487, opened 2015-06-20
Handle AnonymousUser in project listing


hectcastro:
Relevant traceback from staging:

```
Internal Server Error: /api/modeling/projects/
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/views.py", line 452, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/views.py", line 449, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/decorators.py", line 50, in handler
    return func(*args, **kwargs)
  File "/opt/app/apps/modeling/views.py", line 39, in projects
    projects = Project.objects.filter(user=request.user)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/manager.py", line 127, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 679, in filter
    return self._filter_or_exclude(False, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 697, in _filter_or_exclude
    clone.query.add_q(Q(*args, **kwargs))
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/sql/query.py", line 1304, in add_q
    clause, require_inner = self._add_q(where_part, self.used_aliases)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/sql/query.py", line 1332, in _add_q
    allow_joins=allow_joins, split_subq=split_subq,
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/sql/query.py", line 1194, in build_filter
    lookups, value)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/fields/related.py", line 1740, in get_lookup_constraint
[...]

PR #496, opened 2015-06-24
Disallow unauthenticated requests to project listing

mmcfarland:
The app can still support public project sharing via the readonly DRF
decorator on the `/project` endpoint, but `/projects` is only meaningful
to authenticated users.  This will prevent `500` errors for anonymous
access (via bots in the case we were seeing) and instead give a `403`.

Public projects are still accessible.

Connects #487 

To test:

* Browse to `/api/modeling/projects` when not logged in, you should receive a `403` and not a `500`.
* Access a public project, which should still be available to browse.


caseypt:
+1. Tested.

PR #499, opened 2015-06-24
Fix geocodeing issue.

lliss:
It turns out that way back in PR #69 we forgot to add the key to our geocoder
endpoint. Python-Omgeo works by accepting query text and a key in the PlaceQuery
object. The `key` is passed on to the ESRI geocoder as magicKey. We had simply
omitted this and our results were thus not the result of doing a geocode on the
key but only on the query text. This resulted in ambiguous results.

Connects #485

### DIFF
--- a/src/mmw/apps/geocode/views.py
+++ b/src/mmw/apps/geocode/views.py
@@ -18,11 +18,12 @@ geocoder = Geocoder(sources=settings.OMGEO_SETTINGS)
 @decorators.permission_classes((AllowAny, ))
 def geocode(request, format=None):
     query = request.REQUEST.get('search', None)
+    key = request.REQUEST.get('key', None)
     if (not query):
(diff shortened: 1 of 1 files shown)


lliss:
To test:
Do searches for things with ambiguous names and pick something interesting from the list.
"new york"  --> First result sent you to NY Hotel in Las Vegas
"the white house" --> pick the result in TN (sent you to San Diego)
"olive st philadelphia" -->  (sent you to the olive garden on Chestnut st)

You should end up at the right place now. Test against mmw.azavea.com as a comparison.

rajadain:
+1 tested the three prescribed ones, also a few others, everything works predictably now.

kdeloach:
+1

jamesmcclain:
:+1: Tested

Issue #471, opened 2015-06-18
DRF endpoints show GUI in production


mmcfarland:
I had assumed that the DRF GUI was associated with the Django `DEBUG` settings.  If you are logged in, this will give you a nice screen of your projects:

https://mmw.azavea.com/api/modeling/projects/

Is this something we should turn off?  Presumably the endpoint would continue to give json output, and one can still get to it via the browser (https://mmw.azavea.com/api/modeling/projects/?format=json).

rajadain:
Yeah we should definitely turn that off. Especially since it isn't just a read-only view, it actually allows users to submit POST requests in a form at the bottom.

lewfish:
So, we just want to turn this off in production right, since it could be useful during development?

mmcfarland:
Correct - and just the GUI part, it'll obviously still return whatever data it's meant to.

PR #505, opened 2015-06-25
Turn off DRF GUI in production and test

lewfish:
To test:
 * Go to localhost:8000/api/modeling/projects/. You should see the DRF GUI.
 * Switch to test mode by running `VAGRANT_ENV=TEST vagrant provision` (This will take ~10 min or so.)
 * Go to the same URL. This time you should just see some raw JSON.
 * The same setting used to turn off the GUI in test mode is in the production config, but there's no easy way to test production mode locally. 

Connects #471


hectcastro:
@azavea-bot rebuild

kdeloach:
+1 I verified that the DRF GUI appears in dev mode, and switching to prod/test mode with `VAGRANT_ENV=TEST vagrant provision app` disables the default DRF GUI.

PR #507, opened 2015-06-26
Remove udev rules; just restart Celery always

hectcastro:
Shell provisioning steps are supposed to happen late in the provisioning process. Here, a shell provisioner is added that always runs. Its job is to attempt to restart Celery in case it didn't start because the Vagrant synced folder wasn't mounted yet.

---

**Testing**

- [x] Destroy `worker` VM so that `udev` rules are removed
- [x] Recreate the `worker` VM
- [x] Attempt to `reload` it several times to see if there are any issues in the Vagrant pre-provisioning process

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -114,6 +114,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       ansible.groups = ANSIBLE_GROUPS.merge(ANSIBLE_ENV_GROUPS)
       ansible.raw_arguments = ["--timeout=60"]
     end
+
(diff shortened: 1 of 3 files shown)


lliss:
Tried to rebuild and got
`==> worker: stdin: is not a tty`

hectcastro:
Hacks exist to remove the tty message, but I would like to try and avoid them. Despite that message in the output, the inline script executes correctly.

See also: https://github.com/mitchellh/vagrant/issues/1673#issuecomment-17237843

kdeloach:
+1

Issue #343, opened 2015-06-09
Make enter key work on modals


lliss:
I'd like to use the keyboard to control modals. Doesn't seem to work right now.

rajadain:
Also, the primary field should be focused by default. While #416 addresses this for all user modals, core modals including input and share still require mouse focusing and clicking.

PR #509, opened 2015-06-29
Make modals more intuitive and keyboard-friendly

rajadain:
Pre-select text in Share and Input modals so one may begin typing new values immediately, and trigger the primary action upon pressing <kbd>Enter</kbd>.

This does not work for the Confirm modal because there is not text input to receive `keyup` events. Since confirmation should require a higher degree of user attention, perhaps it is fitting that they must use the mouse to select the option they want. However, if this behavior is deemed inconsistent, there are hacks we can use to force the modal div to receive keyboard input.

**Testing Instructions**

 * Try renaming a project to test the input modal.
 * Try deleting a project to test the confirm modal.
 * Try sharing a project to test the share modal.
 * Try closing the modals by pressing <kbd>Esc</kbd>

Connects #343

### DIFF
--- a/src/mmw/js/src/core/templates/confirmModal.html
+++ b/src/mmw/js/src/core/templates/confirmModal.html
@@ -5,7 +5,7 @@
         </div>
         <div class="modal-footer">
             <button type="button" class="btn btn-md btn-default" data-dismiss="modal">{{ cancelLabel }}</button>
-            <button type="button" class="btn btn-md btn-default confirm" data-dismiss="modal">{{ confirmLabel }}</button>
+            <button type="button" class="btn btn-md btn-active confirm" data-dismiss="modal">{{ confirmLabel }}</button>
         </div>
     </div>
 </div>
--- a/src/mmw/js/src/core/templates/inputModal.html
+++ b/src/mmw/js/src/core/templates/inputModal.html
@@ -14,7 +14,7 @@
         </div>
         <div class="modal-footer">
             <button type="button" class="btn btn-md btn-default" data-dismiss="modal">Cancel</button>
-            <button type="button" class="btn btn-md btn-default save">Save</button>
+            <button type="button" class="btn btn-md btn-active save">Save</button>
         </div>
     </div>
 </div>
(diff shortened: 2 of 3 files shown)


caseypt:
+1 Tested. I agree with you re: the confirm modal.

kdeloach:
+1

lewfish:
I noticed that hitting enter doesn't work on the Share modal. Why is that?

rajadain:
Clipboard access is [limited](https://github.com/zeroclipboard/zeroclipboard/blob/master/docs/instructions.md#limitations) from JavaScript to prevent [clipboard poisoning](http://www.computerworld.com/article/2533706/security0/adobe-patches-flash-clickjacking-and-clipboard-poisoning-bugs.html). Copying to the clipboard requires the user explicitly click the button, and there is no way for us to simulate that click with an <kbd>Enter</kbd> press in the Share modal.

lewfish:
+1, tested. See if there's an easy way to get the escape button working across all the modals.

rajadain:
I switched to using the `events` hash instead of `jQuery.on` or `Backbone.listenTo` because this will prevent memory leaks which might happen upon unloading the view if the listeners are not removed, and `listenTo` can only be used with Backbone models and collections, not with DOM objects. Available in b8120c8

rajadain:
By adding a `tabindex="-1"` attribute on all modals, we ensure that they always receive keyboard input. This allows the user to <kbd>Escape</kbd> out of any modal, including Confirm, but also press <kbd>Enter</kbd> to perform the primary action, including Delete. Available in bf4b9da

Since this is potentially dangerous, with the user now able to delete a project forever with a keystroke, we now highlight the button that will be pressed upon pressing <kbd>Enter</kbd>. This is true for Confirm and Input modals, but not for the Share modal because of [reasons described above](https://github.com/WikiWatershed/model-my-watershed/pull/509#issuecomment-116839274). Available in bc098f8

caseypt:
@azavea-bot rebuild

lewfish:
Nice job on the updates. Everything works except that when pressing tab to switch focus, the initially highlighted button stays highlighted even though it shouldn't.

rajadain:
Yeah, that's because we use the `btn-active` class which is designed to stay on all the time. But when "Cancel" is focused, there is plenty of visual feedback for this not to be a dealbreaker. For example, in Chrome:

image

And in Firefox:

image

rajadain:
@azavea-bot rebuild

rajadain:
Squashed commits to make them cleaner and easier to understand, with more detailed commit messages. Will merge shortly.

Issue #378, opened 2015-06-12
Clicking a Scenario context menu button ▾ should make scenario active


rajadain:
While sharing the Project and current Scenario shows the correct URL in the Share Modal, sharing an inactive Scenario shows the incorrect URL of the active Scenario, instead of the one being shared:

![image](https://cloud.githubusercontent.com/assets/1430060/8119697/7b593788-1065-11e5-8047-542e998cb73a.png)

![image](https://cloud.githubusercontent.com/assets/1430060/8119709/8aa0e010-1065-11e5-8d21-127a10ebc589.png)

In this case, the URL should have been http://localhost:8000/model/7/scenario/**20**, not http://localhost:8000/model/7/scenario/**21**

Clicking the ▾ button should make that scenario active first, before opening the drop-down menu.

kdeloach:
I think this is easy to remedy by using `cid` from the click event, to find the correct model in the collection, instead of always using the active scenario.

However, disabling the dropdown for non-active tabs would increase parity with Google Sheets, which isn't a bad thing.

Issue #377, opened 2015-06-12
Remove scenario slug from app


caseypt:
I don't believe we are using it anywhere anymore, but verify before removing. We transitioned to `cid` for referencing unsaved scenarios, and since we allow for projects to have the same name, we are not using a project slug in the URL.

https://github.com/WikiWatershed/model-my-watershed/search?utf8=%E2%9C%93&q=slug

caseypt:
Shares a point with #377

Issue #365, opened 2015-06-11
Change `model` url path to `project`


mmcfarland:
`Model` has too many contexts and this URL doesn't refer to any of them specifically, but rather is a resource of type `Project`.  It will also read better as `/project/<id>/scenario/<id>`

PR #511, opened 2015-06-29
Hide dropdown button on inactive scenario tabs

rajadain:
I could think of three approaches to solve the problem of sharing the incorrect scenario URL:

 1. **Make the Share modal prompt the correct URL**. ~~The reason we currently use `window.location.href` is to avoid hardcoding a Scenario URL in the code, since it may change over time (see #365, #377). It would become one more place that needs to be updated if the URL structure changes, and thus add fragility and brittleness to the codebase.~~ There is a method `getReferenceUrl` on the `Project` model which can get us this value, and in fact is updated with URL changes https://github.com/WikiWatershed/model-my-watershed/pull/513/files#diff-9abeff034ba5a68b4814ee90721888ebR223. However, we would also have to do the same treatment for every item in that dropdown list, which right now only includes `Print`, but could have more options in the future.
 2. **Make clicking the dropdown icon activate the tab**. This is what the original card asked for, but unfortunately the dropdown button and the containing tab header are two different views, and stringing them together will create unnecessary coupling. Furthermore, we use Bootstrap's implementations for tabs and dropdown lists which aren't very easy to hook in to or override the behavior of.
 3. **Hide the dropdown button on inactive tabs**. This is straightforward to implement and is also in line with conventions of Google Sheets. However, it now makes sharing inactive scenarios slightly more difficult, since users have to click twice instead of once, first to make that scenario active and then click the dropdown. There is also a slight visual vacuum when the icons is hidden, although perhaps I feel that way because I'm so used to seeing it before. But it is still better than removing that space, which would disturb visual positioning of unrelated tabs.

I went with 3 in this case:

image

but if others feel strongly about going with 1 or 2 I can comply.

Connects #378

### DIFF
--- a/src/mmw/js/src/modeling/templates/scenarioTabPanel.html
+++ b/src/mmw/js/src/modeling/templates/scenarioTabPanel.html
@@ -1,21 +1,23 @@
 <a href="#{{ cid }}" data-scenario-cid="{{ cid }}" aria-controls="home" role="tab" data-toggle="tab" class="tab-name"><span>{{ name }}</span></a>
-<div class="scenario-btn-dropdown">
-    <div class="dropdown">
-        <button class="btn btn-sm btn-icon dark dropdown-toggle" type="button" data-toggle="dropdown" aria-expanded="true">
-        <i class="fa fa-caret-down"></i>
-        </button>
-        <ul class="dropdown-menu menu-right" role="menu">
-            {% if editable %}
-                <li role="presentation"><a role="menuitem" tabindex="-1" data-action="share">Share</a></li>
-            {% endif %}
-            <li role="presentation"><a role="menuitem" tabindex="-1" data-action="print">Print</a></li>
-            {% if not is_current_conditions %}
+{% if active %}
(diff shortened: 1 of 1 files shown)


kdeloach:
+1

src/mmw/js/src/modeling/views.js
```diff
@@ -274,6 +274,7 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
 
     templateHelpers: function() {
         return {
+            active: this.model.get('active'),
```
caseypt: Do we need this? I thought you could reference `active`, like how we access `is_current_conditions` because the model is passed to the template.

rajadain: Oh! Let me test it out and if so remove it. That didn't even occur to me!

rajadain:
Removed `active` templateHelper since it is already available to the template as part of the model. Thanks for catching that @caseypt

caseypt:
+1. Tested. Looks good.

caseypt:
@azavea-bot rebuild

PR #513, opened 2015-06-29
Change /model to /project

caseypt:
* Change the /model route to /project, and update all
references and tests. This route name provides a more
direct mapping the to resource that is being loaded.

**To Test:**
- Navigate to the project page from the root page using the UI controls.
- Save the project.
- Revisit the project using the URL.
- Run the tests.

Connects to #365

### DIFF
--- a/src/mmw/apps/home/tests.py
+++ b/src/mmw/apps/home/tests.py
@@ -37,20 +37,20 @@ class RouteAccessTestCase(TestCase):
         self.c.login(username='test', password='test')
 
     def test_any_user_can_get_model_route_without_project_id(self):
-        response = self.c.get('/model/')
+        response = self.c.get('/project/')
 
         self.assertEqual(response.status_code, 200)
 
         self.c.logout()
 
-        response = self.c.get('/model/')
+        response = self.c.get('/project/')
 
         self.assertEqual(response.status_code, 200)
 
     def test_project_owner_can_get_model_route_with_project_id(self):
         project_id = self.create_private_project()
 
-        response = self.c.get('/model/' + project_id + '/')
+        response = self.c.get('/project/' + project_id + '/')
 
         self.assertEqual(response.status_code, 200)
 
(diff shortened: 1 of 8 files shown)


rajadain:
+1 tested. All tests work correctly, as does the UI. Created a new project as anonymous user, logged in to save the project, reloaded page to fetch from server, loaded an older project which had been saved previously: everything works.

There are still some remaining references, more conceptual than programmatic, such as `ModelPackage`, the `modeling` folder, and the `/api/modeling` API route. The latter two aren't as egregious, and I suppose that we should use `ModelPackage` if it is consistent with TR-55 naming conventions, however if we could update that we might get around names such as `ModelPackageModel`.

Given that we are in pre-Beta, I suppose it is okay to release a breaking URL change. Do we need to have redirects from the old routes for any outside users at this point? /cc @mmcfarland

mmcfarland:
@rajadain I considered it, but instead I'm just going to send out work the previously shared project urls won't work (but can be edited).  Nobody is using this for anything real yet.

rajadain:
Great. In this case no more work _needs_ to be done, and my comments above are just _nice-to-have_.

caseypt:
Thanks @rajadain.

In regards to the remaining references, I left them as is because I felt that what the user is largely doing on that page is still modeling, and that the app name should be closer tied to the functionality rather than the route. 

Specifically regarding `ModelPackage`, I would more be in favor of changing our model and collection naming pattern to use singular nouns for models, and plural nouns for collections, rather than appending `Model` or `Collection`. For example, `Scenario` would be a single scenario models, but `Scenarios` would be a collection of scenarios. `ModelPackage` would just be `ModelPackage` then, which I think best describes what that Backbone model represents.

caseypt:
@azavea-bot rebuild

caseypt:
@azavea-bot rebuild

caseypt:
@azavea-bot rebuild

kdeloach:
RE: Removing "Model" suffix. Sounds good.

Issue #454, opened 2015-06-18
Slow tiles can cause boundary selection to fail


mmcfarland:
This can be replicated on the existing Congressional Districts layer, but is more pronounced on slower to render tile layers that are forthcoming.  It appears that the utfgrid layers aren't loaded until the entire tile set for the extent is loaded, so if you click on a visible tile, the `id` does not exist and there is an exception.  The click handler has already removed the overlay layer, so the process must be restarted.

To reproduce, load a boundary layer and click immediately on a feature before the remaining tiles load.

PR #517, opened 2015-06-29
Fix slow UTF8 grid bug

lewfish:
When zooming out (to see the whole country) and then selecting Congressional Districts, the tiles load from East to West. On the develop branch, if you click on a shape in the West too soon, the shape id is not available. This causes a request for a null shape, and the boundaries to be cleared. In this PR, if the shape id isn't available it is polled for, and even if it can't get the shape id within the limit, the boundaries aren't cleared so the user can try clicking again. The error messages (and some debug messages which I will remove before merging) currently go to the console. 

To test, you should try clicking a region that probably isn't loaded yet (as described above) and look for the polling debug messages. Also try clicking a region after everything is loaded. 

A known issue is that sometimes even after the click listener is added, after clicking, the click listener isn't called. This seems strange since the listener is added after the tile layer and UTF8 grid layers are added.

Connects #454

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -29,10 +29,6 @@ var ToolbarView = Marionette.LayoutView.extend({
         resetRegion: '#reset-draw-region'
     },
 
-    modelEvents: {
-        'change:toolsEnabled': 'manageOutlineLayer'
-    },
-
     initialize: function() {
         var map = App.getLeafletMap(),
             ofg = L.featureGroup();
@@ -60,16 +56,7 @@ var ToolbarView = Marionette.LayoutView.extend({
         this.resetRegion.show(new ResetDrawView({
             model: this.model
         }));
-    },
-
-    manageOutlineLayer: function() {
-        var enabled = this.model.get('toolsEnabled');
-
-        if (!enabled) {
-            clearBoundaryLayer(this.model);
-        }
     }
-
 });
 
 var SelectAreaView = Marionette.ItemView.extend({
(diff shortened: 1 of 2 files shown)


rajadain:
JSHint fails on `draw/tests.js` lines 140 and 145:

```
$ ./scripts/npm.sh run lint
+ ARGS='run lint'

js/src/draw/tests.js: line 140, col 9, '$el' is defined but never used.
js/src/draw/tests.js: line 145, col 9, 'deferred' is defined but never used.

2 errors
```

rajadain:
+1 works well. All tests pass and the app does not crash if I click a place without waiting for district tiles to load first. Nice work with the console messages, they helped me understand the sequence of things and which behavior was expected vs unexpected.

At first I assumed that the app would also crash if the basemap wasn't loaded, but that doesn't seem to happen. The failure only occurs if the congressional district tiles are unavailable, and the presence or absence of the underlying basemap seems to have no effect.

I will read through the code tomorrow.

src/mmw/js/src/draw/views.js
```diff
@@ -201,6 +188,61 @@ var ResetDrawView = Marionette.ItemView.extend({
...
+        deferred = $.Deferred(),
+        shapeId = e.data ? e.data.id : null;
+
+        if (shapeId) {
+            _getShapeAndAnalyze();
+        } else {
+            pollForShapeId();
+        }
+
+    function _getShapeAndAnalyze() {
```
mmcfarland: I don't think the underscore prefix is necessary for this internal function, or, I would keep it consistent with the other internal function.

src/mmw/js/src/draw/views.js
```diff
@@ -201,6 +188,61 @@ var ResetDrawView = Marionette.ItemView.extend({
...
+            deferred.reject();
+        }).always(function() {
+            model.enableTools();
+        });
+    }
+
+    function pollForShapeId() {
+        console.log('Shape ID not available yet.');
+        if (pollCount < maxPolls) {
+            var shapeData = grid._objectForEvent(e).data;
```
mmcfarland: Handy function for this use case.

src/mmw/js/src/draw/views.js
```diff
@@ -201,6 +188,61 @@ var ResetDrawView = Marionette.ItemView.extend({
...
+                shapeId = shapeData.id;
+                _getShapeAndAnalyze();
+            } else {
+                window.setTimeout(pollForShapeId, pollInterval);
+                pollCount++;
+            }
+        } else {
+            console.log('Failed to get shape ID within time limit.');
+            model.enableTools();
+            deferred.reject();
```
mmcfarland: Let's add a popup up here on the click point that apologizes for not being able to select the shape and to try again.

src/mmw/js/src/draw/views.js
```diff
@@ -208,31 +250,14 @@ function changeOutlineLayer(endpoint, tableId, model) {
     if (endpoint && tableId !== undefined) {
         var ol = new L.TileLayer(endpoint + '.png'),
             grid = new L.UtfGrid(endpoint + '.grid.json?callback={cb}', { resolution: 4 });
```
mmcfarland: Check out the option `maxRequests` https://github.com/danzel/Leaflet.utfgrid#other-options.

It indicates that it will usually prioritize the json requests over the image requests, potentially reducing the chances of this situation occurring.  Can you play with tweaking that option up to 8 and see how it affects the behavior?

rajadain:
Read through the code, I don't have any comments to add there. Good job!

lewfish:
Revised according to suggestions. Increasing maxRequests had no apparent effect.

rajadain:
Confirmed that I got an error popup apologizing for the delay:

image

rajadain:
JSHint now fails on `draw/views.js` line 208:

```
$ ./scripts/npm.sh run lint
+ ARGS='run lint'

js/src/draw/views.js: line 208, col 32, 'getShapeAndAnalyze' is already defined.

1 error
```

rajadain:
Looks like a bunch of `console.log` statements got merged

Issue #518, opened 2015-06-29
Navigating to /model does not capture full AoI extent


mmcfarland:
The fitBounds call may be happening at a point in the animation before the map reaches it's final size, or perhaps the size of the map includes an invisible portion. The full shape should be visible on the map but it often is clipped on the northerly side.

mmcfarland:
This is no longer an issue.

Issue #401, opened 2015-06-15
Consider not fitting bounds on output panel resize


mmcfarland:
If the user is zoomed to an extent of the AoI, minimizes the output panel, then restores it, we resize the zoom and center to fit the bounds in the new bbox.  This can be annoying if we are focused on a region and need to occasionally look at the results output.  I don't think we need to fit to bounds.

lewfish:
So, we want minimizing and restoring to have no effect on the position and zoom level?

mmcfarland:
That's my inclination - would be good to get other feedback from those of us who have been "using" the app.  Thoughts?

caseypt:
+1

rajadain:
I agree, the resize is disruptive.

This may have the effect of Leaflet attribution never being visible in the `/model` page. Created #497 for that.

PR #519, opened 2015-06-29
Don't fit bounds when toggling the model output

mmcfarland:
Restructure how views tell the map to go half size, so they can also
specify whether or not the bounds should be fit to the new size.  All
main region switching will still initially fit the AoI in the map, but
once the user begins toggling the model output pane, the bounds remain
consistent (and clipped if the map is shrinking).

The previously existing issue where the map fit going towards modelling doesn't fit the entire AoI is still present.  I suspect the fit is happening at a middle point in the animation or possible the map div reaches above and below a header.  The issue created for that is #518

Connects #401

### DIFF
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -109,7 +109,7 @@ var AnalyzeWindow = Marionette.LayoutView.extend({
 
         this.$el.animate({ height: '55%' }, 200, function() {
             self.trigger('animateIn');
-            App.map.set('halfSize', true);
+            App.map.setHalfSize(true);
         });
         if (this.lock !== undefined) {
             this.lock.resolve();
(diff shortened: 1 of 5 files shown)


caseypt:
@azavea-bot rebuild

lewfish:
+1, tested.

src/mmw/js/src/analyze/views.js
```diff
@@ -109,7 +109,7 @@ var AnalyzeWindow = Marionette.LayoutView.extend({
 
         this.$el.animate({ height: '55%' }, 200, function() {
             self.trigger('animateIn');
-            App.map.set('halfSize', true);
+            App.map.setHalfSize(true);
```
kdeloach: Consider calling `App.map.fitToAoi()` instead of using a boolean arg.

kdeloach:
+1

src/mmw/js/src/analyze/views.js
mmcfarland: I wrapped it up together in a single call & attribute because I didn't want the fit call to happen during an animation and wanted the map view to be able to handle the transition internally - but I see that the code never accounted for that anyway, and just pops a class on.  I still kind of like having the single call so that it could manage that, which I now suspect is the cause of #518

PR #521, opened 2015-06-30
Add modal to resend activation

lewfish:
To test:
 * Create a new account, but don't activate.
 * Click Send Activation Email (located between 'Forgot?' and 'Sign Up' on the Login modal)
 * Enter the email for the account just created. It should send an email (on the console in development mode).
 * Try triggering various error messages. Make sure the enter button works in addition to clicking the button.
   * Try entering a non-email address string.
   * Try entering a valid email address that is not associated with a user.
   * Try entering an email address for a user who has already activated their account.

This was pretty easy thanks to the refactoring done by @rajadain 

Connects #361

### DIFF
--- a/src/mmw/apps/user/urls.py
+++ b/src/mmw/apps/user/urls.py
@@ -8,6 +8,7 @@ from django.conf.urls import patterns, url
 from apps.user.views import (login,
                              sign_up,
                              forgot,
+                             resend,
                              logout,
                              itsi_login,
                              itsi_auth,
@@ -21,5 +22,6 @@ urlpatterns = patterns(
     url(r'^itsi/sign_up', itsi_sign_up, name='itsi_sign_up'),
     url(r'^login$', login, name='login'),
     url(r'^sign_up$', sign_up, name='sign_up'),
+    url(r'^resend$', resend, name='resend'),
     url(r'^forgot$', forgot, name='forgot')
 )
--- a/src/mmw/apps/user/views.py
+++ b/src/mmw/apps/user/views.py
@@ -5,7 +5,10 @@ from django.contrib.auth.models import User
 from django.core.urlresolvers import reverse
 from django.shortcuts import redirect, render_to_response
 from django.contrib.auth.forms import PasswordResetForm
+from django.core.exceptions import ObjectDoesNotExist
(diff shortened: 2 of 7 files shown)


lewfish:
@azaveaci rebuild

rajadain:
It seems like there's a missing file in the commit: `user/templates/resendModal.html` referenced in `user/views.js`. Running `./scripts/bundle.sh` gives the following error:

```
Error: Cannot find module './templates/resendModal.html' from '/opt/app/js/src/user'
    at /opt/app/node_modules/browserify/node_modules/resolve/lib/async.js:55:21
    at load (/opt/app/node_modules/browserify/node_modules/resolve/lib/async.js:69:43)
    at onex (/opt/app/node_modules/browserify/node_modules/resolve/lib/async.js:92:31)
    at /opt/app/node_modules/browserify/node_modules/resolve/lib/async.js:22:47
    at Object.oncomplete (fs.js:107:15)
```

lewfish:
You're right. I just added the missing file.

rajadain:
I'm unsure if the "Send Activation Email" button should always be visible. How about making it visible only when login fails due to the account not being activated?

image

rajadain:
Since the ultimate goal of this user interaction is to log in, we should always redirect the user back to the Login modal, instead of letting them continue as guest:

image

rajadain:
+1 tested, works as advertised, great job. Very clean code! Apart from the two comments above I'm very happy with this.

lewfish:
I kind of agree with about only showing the "Send Activation Email" if the user logs in unsuccessfully to avoid cluttering the modal, but what if the user realizes they forgot to activate, and don't realize they can get to the button by logging in unsuccessfully? @mmcfarland What do you think?

rajadain:
I guess what I'm thinking is: How will a user realize they forgot to activate? Either when they see the email, from where they can click the activation link directly, or if they try to login and it tells them that they haven't activated yet (and also shows them the button to resend activation email).

rajadain:
But my concern is not cluttering the modal: I think it looks pretty fine where it is. It just doesn't feel like it fits with every other user action at that time. I also wonder if it could be a security issue.

But I don't feel very strongly about it, so we can also just leave it there.

PR #522, opened 2015-06-30
Invert tiler health conditional

hectcastro:
If there is no database error, nor cache error, then return 200 as the HTTP status code.


mmcfarland:
+1 Code looks right, not tested.

kdeloach:
+1 Sorry, didn't confirm the status code when I tested.

Issue #520, opened 2015-06-30
Separate modifications from inputs in scenario data model


kdeloach:
Fix scenario data model so there is a clearer distinction between modifications and inputs.

PR #524, opened 2015-06-30
Separate modifications from inputs in scenario data model

kdeloach:
Fix scenario data model so there is a clearer distinction between
modifications and inputs.

To test:
1. Run migrations
2. Try to add/remove/duplicate/delete/etc. scenarios

Existing scenarios probably won't work so I suggest clearing out your dev DB.

Connects #520


mmcfarland:
I'm getting these same test failures locally.

kdeloach:
Sorry about that, I failed to run the python unit tests before creating this PR. Will fix now.

kdeloach:
Fixed tests and lint.

mmcfarland:
After the fixups, I'm not able to add/display modifications any more, with errors generated from `updateModifications` in core/views.js.

kdeloach:
I fixed a bug where modifications wouldn't appear on the map and squashed all fixups. Thanks for reviewing.

mmcfarland:
+1 everything is still working and I see the precip input saved distinctly from modifications in the database.  Code improvements look good.

src/mmw/apps/modeling/tests.py
```diff
@@ -15,16 +15,17 @@
...
         model_input = {
-            'modifications': [
+            'inputs': [
                 {
                     'name': 'precipitation',
                     'value': 1.2
                 }
             ],
-            'precip': 1.0
+            'modifications': []
```
caseypt: Just checking if we should be using 1.0 or 1.2 here.

caseypt: Ignore, didn't notice that this was a test.

caseypt:
I don't think this causing problems, but I think at some point we should create a separate model and collection type for inputs. At the moment, they are saved with the default values for modifications:

selection_063

caseypt:
+1. Tested. Nice job.

src/mmw/js/src/modeling/modificationConfigUtils.js
```diff
@@ -6,9 +6,9 @@ var modificationConfig = require('./modificationConfig.json');
 function getHumanReadableName(modKey) {
     if (modificationConfig[modKey]) {
         return modificationConfig[modKey].name;
-    } else {
-        throw 'Unknown Land Cover or Conservation Practice: ' + modKey;
     }
+    console.warn('Unknown Land Cover or Conservation Practice: ' + modKey);
```
lliss: This is sensible.

src/mmw/js/src/modeling/filters.js
```diff
@@ -0,0 +1,16 @@
+"use strict";
+
+var nunjucks = require('nunjucks'),
```
lliss: Much cleaner.

lliss:
+1. Tests pass. Found no errors or unexpected behaviour while testing.

PR #525, opened 2015-07-01
Remove debug messages and change maxPolls

lewfish:
Some fixups for PR https://github.com/WikiWatershed/model-my-watershed/pull/517

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -189,12 +189,11 @@ var ResetDrawView = Marionette.ItemView.extend({
 });
 
 function getShapeAndAnalyze(e, model, ofg, grid, tableId) {
-    console.log('clicked');
     // The shapeId might not be available at the time of the click
     // because the UTF Grid layer might not be loaded yet, so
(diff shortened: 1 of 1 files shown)


caseypt:
Cool with the fixes here, but I think that commit message style is for temporary commits that get squashed, correct @kdeloach?

mmcfarland:
Yeah, I think that commits that are going to get merged should follow the normal conventions.

kdeloach:
Check out the docs for [fixup commits](http://git-scm.com/docs/git-commit). They are intended to be squashed before merging.

lewfish:
Ok, thanks for letting me know.

rajadain:
There's still a `console.log` in line 217:

```javascript
        }).fail(function() {
            console.log('Shape endpoint failed');
            deferred.reject();
        }).always(function() {
```

lewfish:
I wanted to keep that console message in because it shouldn't come up very often, and there's no other way to see that the error is occurring.

rajadain:
Ah good thinking. Alright then, looks good! :+1:

Issue #370, opened 2015-06-11
Add a health check endpoint to the application servers


hectcastro:
Ideally, the route would mount at `/health-check`.

kdeloach:
For reference, here is the stub for the health check endpoint: https://github.com/kdeloach/model-my-watershed/commit/2a19a7fd8709a2b83c8478c34db2c7142b92987c.

PR #528, opened 2015-07-01
Add application health check

hectcastro:
Each probe initiates a connection attempt to the PostgreSQL database and Redis. Failure of either returns an HTTP 503, while success of both returns an HTTP 200. The health check is mounted at `/health-check`.

The inclusion of `BypassMiddleware` is needed because when the connection to PostgreSQL is severed, 
other middleware end up triggering exceptions before the `health_check` response can be returned. This hurts our ability to know if the health check failure is just PostgreSQL or PostgreSQL and Redis.

Connects #370

---

**Testing**

I'm worried that the `BypassMiddleware` will cause problems with other pages. Thus far, I haven't been able to find anything broken and all tests still pass. If there are other solutions to avoid processing middleware, feel free to suggest them.

- [x] Provision the `app` server
- [x] Login to the `services` virtual machine and stop the Redis or PostgreSQL service
- [x] Inspect the JSON payload provided by [http://localhost:8000/health-check](http://localhost:8000/health-check)
- [x] Ensure that nothing breaks because of the `BypassMiddleware` middleware

### DIFF
--- a/src/mmw/apps/monitoring/views.py
+++ b/src/mmw/apps/monitoring/views.py
@@ -1,0 +1,68 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import division
+
+from rest_framework import status
+
+from django.http import JsonResponse
+from django.core.cache import caches
+from django.db import connections
+
+from mmw.middleware import bypass_middleware
+
+import uuid
+
+
+@bypass_middleware
+def health_check(request):
+    response = {}
+
+    for check in [_check_cache, _check_database]:
+        response.update(check())
+
+    if all(map(lambda x: x[0]['default']['ok'], response.values())):
+        return JsonResponse(response, status=status.HTTP_200_OK)
+    else:
+        return JsonResponse(response,
+                            status=status.HTTP_503_SERVICE_UNAVAILABLE)
+
+
+def _check_cache(cache='default'):
+    key = 'health-check-{}'.format(uuid.uuid4())
+
+    try:
+        caches[cache].set(key, uuid.uuid4())
+
+        # Loss of connectivity to Redis does not always throw an
+        # exception because DJANGO_REDIS_IGNORE_EXCEPTIONS is True.
+        if caches[cache].get(key) is None:
+            response = {cache: {'ok': False}}
+        else:
+            response = {cache: {'ok': True}}
+            caches[cache].delete(key)
+    except Exception as e:
+        response = {
+            cache: {
+                'ok': False,
+                'msg': str(e)
+            },
+        }
+
+    return {'caches': [response]}
+
+
+def _check_database(database='default'):
+    try:
+        connections[database].introspection.table_names()
+
+        response = {database: {'ok': True}}
+    except Exception as e:
+        response = {
+            database: {
+                'ok': False,
+                'msg': str(e)
+            },
+        }
+
+    return {'databases': [response]}
--- a/src/mmw/mmw/middleware.py
+++ b/src/mmw/mmw/middleware.py
@@ -1,0 +1,40 @@
+from django.conf import settings
+from django.core.handlers.base import BaseHandler
+from django.core import urlresolvers
+
+
+def bypass_middleware(view):
+    view.bypass_middleware = True
+
+    return view
+
+
+class BypassMiddleware(object):
+    """
+    Customized version of a gist detailing this technique by @bryanhelmig
+
+    See also: https://gist.github.com/bryanhelmig/9d09d1bd9a63504371d2
+    """
+    def process_request(self, request):
+        """Replicates a lot of code from BaseHandler#get_response."""
+        # Setup URL resolver
+        urlconf = settings.ROOT_URLCONF
+        urlresolvers.set_urlconf(urlconf)
+        resolver = urlresolvers.RegexURLResolver(r'^/', urlconf)
+        callback, callback_args, \
+            callback_kwargs = resolver.resolve(request.path_info)
+
+        if getattr(callback, 'bypass_middleware', False):
+            # bypass_middleware decorator was used; zero out all
+            # middleware and return the response.
+            handler = BaseHandler()
+
+            handler._request_middleware = []
+            handler._view_middleware = []
+            handler._template_response_middleware = []
+            handler._response_middleware = []
+            handler._exception_middleware = []
+
+            response = handler.get_response(request)
+
+            return response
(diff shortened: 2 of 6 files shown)


hectcastro:
@mmcfarland, you may want to take a look at this. It is a pretty low level change. I'd love to find a less intrusive solution.

src/mmw/mmw/middleware.py
```diff
@@ -0,0 +1,40 @@
...
+
+            handler._request_middleware = []
+            handler._view_middleware = []
+            handler._template_response_middleware = []
+            handler._response_middleware = []
+            handler._exception_middleware = []
+
+            response = handler.get_response(request)
+
+            return response
```
kdeloach: There may be enough args on `process_request` to call the view directly (Source: https://docs.djangoproject.com/en/1.8/topics/http/middleware/#process_view). This could reduce the amount of code needed for this middleware (Reference implementation: http://stackoverflow.com/a/7260757).

kdeloach:
+1 Confirmed messages and status codes are correct. I also tested that the health check yields an error message if Redis is down when `DJANGO_REDIS_IGNORE_EXCEPTIONS` is `False`.

src/mmw/mmw/middleware.py
hectcastro: I may be missing something here, but what other args on `process_request`?

If I try to intercept at `process_view` and conditionally call `view_func`, the database connectivity exception occurs during or after `view_func`:

```python
def process_view(self, request, view_func, view_args, view_kwargs):
    health_check_func = '.'.join([view_func.__module__,
                                  view_func.func_name])

    if health_check_func == 'apps.monitoring.views.health_check':
        print("joker")
        return view_func(request, *view_args, **view_kwargs)
```

In the logs, I see:

```
joker
Internal Server Error: /health-check/
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 223, in get_response
    response = middleware_method(request, response)
  File "/usr/local/lib/python2.7/dist-packages/debug_toolbar/middleware.py", line 94, in process_response
    new_response = panel.process_response(request, response)
  File "/usr/local/lib/python2.7/dist-packages/debug_toolbar/panels/request.py", line 55, in process_response
    for k in sorted(request.session.keys(), key=force_text)]
  File "/usr/local/lib/python2.7/dist-packages/django/contrib/sessions/backends/base.py", line 119, in keys
    return self._session.keys()
  File "/usr/local/lib/python2.7/dist-packages/django/contrib/sessions/backends/base.py", line 181, in _get_session
    self._session_cache = self.load()
  File "/usr/local/lib/python2.7/dist-packages/django/contrib/sessions/backends/db.py", line 21, in load
    expire_date__gt=timezone.now()
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/manager.py", line 127, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 328, in get
    num = len(clone)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 144, in __len__
    self._fetch_all()
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 965, in _fetch_all
    self._result_cache = list(self.iterator())
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 238, in iterator
    results = compiler.execute_sql()
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/sql/compiler.py", line 838, in execute_sql
    cursor = self.connection.cursor()
  File "/usr/local/lib/python2.7/dist-packages/debug_toolbar/panels/sql/tracking.py", line 42, in cursor
    return state.Wrapper(connection._djdt_cursor(), connection, panel)
  File "/usr/local/lib/python2.7/dist-packages/django/db/backends/base/base.py", line 162, in cursor
    cursor = self.make_debug_cursor(self._cursor())
  File "/usr/local/lib/python2.7/dist-packages/django/db/backends/base/base.py", line 135, in _cursor
    self.ensure_connection()
  File "/usr/local/lib/python2.7/dist-packages/django/db/backends/base/base.py", line 130, in ensure_connection
    self.connect()
  File "/usr/local/lib/python2.7/dist-packages/django/db/utils.py", line 97, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "/usr/local/lib/python2.7/dist-packages/django/db/backends/base/base.py", line 130, in ensure_connection
    self.connect()
  File "/usr/local/lib/python2.7/dist-packages/django/db/backends/base/base.py", line 119, in connect
    self.connection = self.get_new_connection(conn_params)
  File "/usr/local/lib/python2.7/dist-packages/django/db/backends/postgresql_psycopg2/base.py", line 176, in get_new_connection
    connection = Database.connect(**conn_params)
  File "/usr/local/lib/python2.7/dist-packages/psycopg2/__init__.py", line 164, in connect
    conn = _connect(dsn, connection_factory=connection_factory, async=async)
OperationalError: could not connect to server: Connection refused
	Is the server running on host "33.33.34.30" and accepting
	TCP/IP connections on port 5432?
```

kdeloach: Sorry, I mixed up process_view and process_request. Disregard.

src/mmw/apps/monitoring/views.py
```diff
@@ -0,0 +1,68 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import division
+
+from rest_framework import status
+
+from django.http import JsonResponse
```
hectcastro: It is worth noting that I originally went with the DRF approach for this endpoint, but then switched to JsonResponse because I wanted to avoid the overhead. If there are reasons that approach might be better, feel free to suggest them.

mmcfarland:
Taking a look now

mmcfarland:
+1 I don't think it's too intrusive and messages/status codes work as expected.  No unusual behavior noticed in the app or tests.

src/mmw/mmw/middleware.py
```diff
@@ -0,0 +1,40 @@
...
+from django.core import urlresolvers
+
+
+def bypass_middleware(view):
+    view.bypass_middleware = True
+
+    return view
+
+
+class BypassMiddleware(object):
```
mmcfarland: One question is what is the type of value for `object` here?  Can it be removed or renamed to reflect its content, doesn't look like it's used at all?

mmcfarland: Nevermind -this is python and it's a class.

PR #531, opened 2015-07-01
Logging out redirects to home

rajadain:
**Testing Instructions**

 * Login as a user, go to [`/user/logout`](http://localhost:8000/user/logout), ensure you are logged out and redirected to home 

Connects #214


lliss:
+1 tested.

Issue #530, opened 2015-07-01
fix failing SuggestionModel unit test


lewfish:
I get the following unit test failure about half of the time when running testem on the  develop branch.

SuggestionModel
#setMapViewLocation
sets map model's position attributes to the geocode result position ‣
TypeError: Cannot read property 'x' of undefined
    at Object.L.DomUtil.getTranslateString (http://localhost:7357/js/test.vendor.js:44453:22)
    at Object.L.DomUtil.setPosition (http://localhost:7357/js/test.vendor.js:44470:47)
    at L.PosAnimation.L.Class.extend.stop (http://localhost:7357/js/test.vendor.js:52038:13)
    at L.Map.include.setView (http://localhost:7357/js/test.vendor.js:52112:18)
    at Marionette.ItemView.extend.updateView (http://localhost:7357/js/test.bundle.js:3577:30)
    at triggerEvents (http://localhost:7357/js/test.vendor.js:21368:57)
    at Backbone.Events.trigger (http://localhost:7357/js/test.vendor.js:21308:19)
    at _.extend.set (http://localhost:7357/js/test.vendor.js:21528:16)
    at Backbone.Model.extend.setMapViewToLocation (http://localhost:7357/js/test.bundle.js:4895:17)
    at Context.<anonymous> (http://localhost:7357/js/test.bundle.js:5270:23)

lewfish:
It seems like I'm the only one who gets this failure.

lewfish:
The failure doesn't occur when running it in isolation from the other tests.

Issue #516, opened 2015-06-29
Sharing should not be available on scenarios for unsaved projects.


caseypt:
Just like projects, the share option should only appear once the project is saved.

Add a test after making the change.

PR #532, opened 2015-07-01
Only show Share menu item when saved

lewfish:
To test, when logged out, check that there is no 'Share' item in the Project and Scenario menus. When logged in, check that the 'Share' item appears in both menus.

The testem tests fail some of the time (see #530) but this also occurs on the develop branch.

Connects #516

### DIFF
--- a/src/mmw/js/src/modeling/templates/projectMenu.html
+++ b/src/mmw/js/src/modeling/templates/projectMenu.html
@@ -8,7 +8,7 @@
         <i class="fa fa-caret-down"></i>
         </button>
         <ul class="dropdown-menu menu-right" role="menu">
-            {% if id and editable %}
+            {% if not is_new and editable %}
                 <li role="presentation"><a role="menuitem" tabindex="-1" data-toggle="modal" id="share-project">Share</a></li>
                 <li role="presentation"><a role="menuitem" tabindex="-1" data-toggle="modal" id="project-privacy">{% if is_private %}Make Public{% else %}Make Private{% endif %}</a></li>
                 <li role="presentation"><a role="menuitem" tabindex="-1" data-toggle="modal" id="delete-project">Delete</a></li>
@@ -17,7 +17,7 @@
             {% if editable %}
                 <li role="presentation"><a role="menuitem" tabindex="-1" data-toggle="modal" id="rename-project">Rename</a></li>
             {% endif %}
-            {% if not id %}
+            {% if is_new %}
                 <li role="presentation"><a role="menuitem" tabindex="-1" id="save-project">Save</a></li>
             {% endif %}
             <li role="presentation"><a role="menuitem" tabindex="-1" id="print-project">Print</a></li>
--- a/src/mmw/js/src/modeling/templates/scenarioTabPanel.html
+++ b/src/mmw/js/src/modeling/templates/scenarioTabPanel.html
@@ -6,7 +6,7 @@
             <i class="fa fa-caret-down"></i>
             </button>
             <ul class="dropdown-menu menu-right" role="menu">
-                {% if editable %}
+                {% if editable and not is_new %}
                     <li role="presentation"><a role="menuitem" tabindex="-1" data-action="share">Share</a></li>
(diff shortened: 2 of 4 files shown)


src/mmw/js/src/modeling/models.js
```diff
@@ -275,7 +277,9 @@ var ScenarioModel = Backbone.Model.extend({
         modifications: null, // ModificationsCollection
         active: false,
         job_id: null,
-        results: null // ResultCollection
+        results: null, // ResultCollection
+        id: null // Scenario ID from server
```
mmcfarland: Check out the `isNew` method to see if that can be used instead of an id check.

src/mmw/js/src/modeling/templates/projectMenu.html
```diff
@@ -8,7 +8,7 @@ <h3 class="light">{{ name }}</h3>
         <i class="fa fa-caret-down"></i>
         </button>
         <ul class="dropdown-menu menu-right" role="menu">
-            {% if id and editable %}
+            {% if id and not isNew %}
```
rajadain: Checking both is redundant since they will always have exactly the same value. Consider using just `if id` or `if not isNew`

rajadain: Or perhaps you meant to replace `id` instead of `editable` with `not isNew`?

src/mmw/js/src/modeling/views.js
```diff
@@ -275,7 +276,8 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
     templateHelpers: function() {
         return {
             cid: this.model.cid,
-            editable: isEditable(this.model)
+            editable: isEditable(this.model),
+            isNew: this.model.isNew()
```
rajadain: Since Nunjucks templates follow Python conventions, consider naming this variable `is_new`

lewfish:
Please run the testem tests a few times locally. Let me know if you get a failure in #setMapViewLocation. I get a failure locally, but it doesn't seem to happen on staging or for other people locally.

src/mmw/js/src/modeling/tests.js
```diff
@@ -90,6 +90,14 @@ describe('Modeling', function() {
                 assert.equal($('#sandbox ul.nav.nav-tabs > li').length, 3);
             });
 
+            function checkItemsMatch(expectedItems) {
```
rajadain: Since this function is not generic and only matches menu items, consider renaming to `checkMenuItemsMatch`

src/mmw/js/src/modeling/templates/scenarioTabPanel.html
```diff
@@ -6,7 +6,7 @@
             <i class="fa fa-caret-down"></i>
             </button>
             <ul class="dropdown-menu menu-right" role="menu">
-                {% if editable %}
+                {% if editable and not is_new %}
```
rajadain: Perhaps the conditional hierarchy in this page can be simplified? Unrelated to your changes, but a good opportunity to simplify. Currently we have:

```
if editable and not is_new
    Share
endif
Print
if not is_current_conditions
    if editable
        Rename
        Duplicate
        Delete
    endif
endif
```

Which might be simplified to:

```
if editable
    if not is_new
        Share
    endif
    if not is_current_conditions
        Rename
        Duplicate
        Delete
    endif
endif
Print
```

However, this would change the order of the items, which might not be desirable at this time.

rajadain: Actually, disregard that entire comment. It might be simplest to lift the `editable` condition into its parent `if`:

```
if editable and not is_new
    Share
endif
Print
if editable and not is_current_conditions
    Rename
    Duplicate
    Delete
endif
```

rajadain:
+1 tested, works well. Good, clean code. Nice job!

The tests pass for me when running from the command line, but after re-running testem a bunch of times I did get the `#setMapViewLocation` error which I couldn't get to go away once it came. Please make an issue for it.

Issue #337, opened 2015-06-09
Setup read replica for ElastiCache


hectcastro:
This is to help protect against a Redis master failure.

hectcastro:
This should probably happen along with https://github.com/WikiWatershed/model-my-watershed/issues/339.

hectcastro:
`t2.*` instances aren't supported for setting up an ElastiCache replica set. The next cheapest instance is an `m1.small`. We'd be going from $0.018 to $0.075 per hour.

mmcfarland:
Is this resolved by #533?

hectcastro:
Yes!

PR #533, opened 2015-07-01
Setup a Redis read replica

hectcastro:
Two Redis cache clusters are created in different availability zones with one node each. Then, a replication group is created to house both cache clusters. At that time, one is set as the primary, and the other its read replica:

image

**Fun learnings**:

- Redis replication is asynchronous. Therefore, when a primary cluster fails over to a replica, a small amount of data might be lost due to replication lag.

- When selecting the replica to promote to primary, ElastiCache selects the replica with the least replication lag (the one that is most current).

- When you enable Multi-AZ on a replication group, a replica cluster cannot be manually promoted to primary. Thus, if the primary in AZ-1 fails over to a replica in AZ-2, the primary cluster stays in AZ-2. To promote the new replica in AZ-1 to primary, you must first disable Multi-AZ on the replication group, then do the promotion, and finally re-enable Multi-AZ.

- In the case where a cluster's failure is caused by the rare event of an entire Availability Zone failing, the replica replacing the failed primary is created only when the Availability Zone is back up. For example, if the primary cluster is in AZ-1 with replicas in AZ-2 and AZ-3, if the primary cluster fails the replica with the least replication lag is promoted to primary. ElastiCache then creates a new replica in AZ-1 (where the failed primary was) only when AZ-1 is back up and available.

- Rebooting a primary cluster does not trigger auto failover. When the primary node is rebooted, it is cleared of data when it comes back online. When the read replicas see the cleared primary, they clear their copy of the data, incurring data loss.

- Once a read replica has been promoted, the other replicas sync with their new primary. After the initial sync, the replicas’ content is deleted and they sync the data from the new primary, causing a brief interruption during which the replicas are not accessible. This sync process also causes temporary load on the primary while syncing with the replicas. This behavior is native to Redis and isn’t unique to ElastiCache Multi-AZ.

- Multi-AZ is not supported on `t1` and `t2` cache nodes (which is why we're on `m1.small` now).

Connects #337

---

**Testing**

I already updated the data plane stack with these changes, so reapplying this stack won't do much. All of the changes in `CachePrivateDNSRecord` where undone manually to make way for the new CloudFormation DNS resources. I'd say just reviewing the changes and making sense of the learnings is enough.


hectcastro:
Reminder to reapply this template to the data plane stack after merging.

lliss:
I'm thinking we should capture some of the learnings in a readme so we have a more permanent place to access this information.

lliss:
+1

Issue #473, opened 2015-06-18
0 modifications should not drop down an empty list


mmcfarland:
A tiny dialog stub is shown when clicking on the modifications list when there are none present.  I think we should not display this list, or otherwise have a consistent header but in the case of no modifications, display a message saying as much.

![screenshot from 2015-06-18 17 51 24](https://cloud.githubusercontent.com/assets/1014341/8243180/ad19c0da-15e2-11e5-8d8a-c2d5121fa84f.png)

PR #534, opened 2015-07-01
Disable "modifications" dropdown.

jamesmcclain:
Disable the "modifications" dropdown when there are no modifications.  Also, the button has been removed entirely when the scenario is not editable.

**To test**
   1. Create or go to a scenario with no modifications.
   2. Notice the disabled dropdown button.
   3. Add some modifications.
   4. Now the button should behave exactly as it did before.
   5. Remove the modifications.
   6. The button is disabled again.

Connects #473

screenshot from 2015-07-01 17 25 10
 screenshot from 2015-07-01 17 26 08
screenshot from 2015-07-01 17 26 26

### DIFF
--- a/src/mmw/js/src/modeling/templates/scenarioToolbarTabContent.html
+++ b/src/mmw/js/src/modeling/templates/scenarioToolbarTabContent.html
@@ -2,9 +2,15 @@
 
 <div id="modification-btn-wrapper"> <!-- Modifications Dropdown -->
     <div class="dropdown"> <!-- Dropdown Button -->
-        <button class="btn btn-sm btn-default inverted" type="button" data-toggle="dropdown" aria-expanded="true">
-        <span id="modification-number"><strong>{{ shapes|length }}</strong></span> Modifications
-        </button>
+        {% if not is_current_conditions %}
+            {% if shapes|length == 0 %}
+                <button class="btn btn-sm btn-default inverted" type="button" disabled="true">
+            {% else %}
+                <button class="btn btn-sm btn-default inverted" type="button" data-toggle="dropdown" aria-expanded="true">
+            {% endif %}
+                    <span id="modification-number"><strong>{{ shapes|length }}</strong></span> Modifications
+                </button>
+        {% endif %}
         <div id="modifications" class="dropdown-menu menu-right" role="menu"> <!-- DropdownContent -->
         {% for controlName, models in groupedShapes %}
             <div class="pad-1 brd-bottom">


src/mmw/js/src/modeling/templates/scenarioToolbarTabContent.html
```diff
@@ -2,8 +2,12 @@
 
 <div id="modification-btn-wrapper"> <!-- Modifications Dropdown -->
     <div class="dropdown"> <!-- Dropdown Button -->
+        {% if shapes|length == 0 %}
```
rajadain: Since Nunjucks follows Python convention, consider using `if not shapes` instead of `if shapes|length == 0`

rajadain:
+1 tested, works as advertised, nice job!

Since Current Conditions can never have any modifications, consider always hiding the modifications button in Current Conditions.

src/mmw/js/src/modeling/templates/scenarioToolbarTabContent.html
rajadain: Well, try it, and if it doesn't work this is fine. Perhaps we can do `===` to be even more explicit.

jamesmcclain:
Great suggestions, I will follow up on both of them.  Thanks!

src/mmw/js/src/modeling/templates/scenarioToolbarTabContent.html
```diff
@@ -2,9 +2,15 @@
 
 <div id="modification-btn-wrapper"> <!-- Modifications Dropdown -->
     <div class="dropdown"> <!-- Dropdown Button -->
+        {% if not is_current_conditions %}
+        {% if shapes|length == 0 %}
```
rajadain: Consider improving readability by adding indentation:

```jinja2
{% if not is_current_conditions %}
    {% if shapes|length == 0 %}
        <button ...
    {% else %}
        <button ...
    {% endif %}
            <span ...
        </button>
{% endif %}
```

src/mmw/js/src/modeling/templates/scenarioToolbarTabContent.html
```diff
@@ -2,9 +2,15 @@
...
+        <button class="btn btn-sm btn-default inverted" type="button" disabled="true">
+        {% else %}
         <button class="btn btn-sm btn-default inverted" type="button" data-toggle="dropdown" aria-expanded="true">
-        <span id="modification-number"><strong>{{ shapes|length }}</strong></span> Modifications
+        {% endif %}
+          <span id="modification-number"><strong>{{ shapes|length }}</strong></span> Modifications
         </button>
+	{% endif %}
```
rajadain: Tab instead of spaces here

rajadain:
+1 again, works well! Modifications button is no longer visible in Current Conditions scenario, and works as expected in all regular scenarios.

jamesmcclain:
Okay, will make those improvements and merge.  Thanks again for reviewing!

Issue #543, opened 2015-07-02
Enable StatsD plugin for Django


hectcastro:
Within the Django application, enable `STATSD_CLIENT`. The `django-statsd-mozilla` Python module is already installed.

PR #547, opened 2015-07-02
Enable StatsD metrics from Django

hectcastro:
Use `django-statsd-mozilla` to sends timings via middleware. Also allows you to import a StatsD client within any module to emit custom metrics. See `boto_ses_mailer.py` for an example.

I could see this being useful for subcomponents of the background jobs. Celery will automatically emit job level metrics to Statsite.

Connects #543

---

**Testing**

- [ ] Provision the `app` server
- [ ] Interact with the application
- [ ] Check the Graphite UI to see if there are metrics under the `statsite` namespace [http://localhost:8080](http://localhost:8080)

graphite_browser


mmcfarland:
+1 Got django
screenshot from 2015-07-06 09 07 28
 data in Graphite

PR #550, opened 2015-07-06
Fix SuggestionModel Unit Test

lewfish:
The SuggestionModel#setMapViewLocation test was failing when run alongside the Draw test "fails to add AOI when shape id cannot be retrieved by getShapeAndAnalyze", but worked in isolation. The Draw test opened a Leaflet popup which set map._panAnim to a truthy value, which led the SuggestionModel test to execute some code that tried to get a style on the window that doesn't exist. I tried to just close the popup, but that didn't work. I then manually set the _panAnim to null, and that fixed it. Setting an internal property doesn't seem like a good idea, but I'm not sure how else to fix this.

It seems like unit tests interfering with each other is a common problem. Maybe we can just clear the entire map state (or recreate it) after each test (or set of tests) is done.

To test:
Make sure that testem tests pass. Try reloading them a few times to make sure.

Connects #530

### DIFF
--- a/src/mmw/js/src/draw/tests.js
+++ b/src/mmw/js/src/draw/tests.js
@@ -75,7 +75,8 @@ describe('Draw', function() {
         });
 
         it('fails to add AOI when shape id cannot be retrieved by getShapeAndAnalyze', function(done) {
-            var successCount = 10,
+            // Set successCount high enough so that the polling will fail.
+            var successCount = 6,
                 deferred = setupGetShapeAndAnalyze(successCount),
                 success;
 
@@ -87,6 +88,8 @@ describe('Draw', function() {
                     success = false;
                 }).
                 always(function() {
+                    App.getLeafletMap().closePopup();
+                    App.getLeafletMap()._panAnim = false;
                     assert.equal(success, false);
                     done();
                 });


kdeloach:
I noticed that Leaflet Map has a `stop` method (Source: https://github.com/Leaflet/Leaflet/blob/62c71f5c1f6f912722446cc2608993261ee6e325/src/map/Map.js#L229). Would that work instead?

rajadain:
I agree, the map state shouldn't really interact between unit tests. Perhaps we can look into making some hooks that reset the map state after each test: http://mochajs.org/#hooks

On an unrelated but interesting note, if my testem browsers are not in focus, i.e. I'm looking at other tabs or other windows, on Mac OS X Yosemite, I get consistent timeout errors, and all tests take longer:

screen shot 2015-07-06 at 5 12 39 pm

But if I make bring all those tabs and windows to the foreground, tests execute much faster and I don't get any timeouts anymore:

image

This is true for both `develop` and your branch, so is unaffected by changes herein.

lewfish:
The stop method can't be called on the map. I'm trying to remove and re-create the Leaflet map, but it's throwing an exception because there's nothing in the DOM aside from the map container.

lewfish:
I think we should go with the quick fix, and find a better way as part of https://github.com/WikiWatershed/model-my-watershed/issues/558

kdeloach:
+1

caseypt:
+1 I'm not sure what this is an indication of, but it's strange that this wasn't happening on Jenkins. Though I wasn't seeing it locally either.

Issue #512, opened 2015-06-29
Cache intermediate model run output (GP)


mmcfarland:
Currently the app caches the end result of a complete model run in the scenario so we don't have rerun the entire job when reloading or sharing a scenario.  Instead, we only run the job when a modification or input is changed.  Given that model inputs, at least in the case of the Site Storm Model, are non-geographic, we could also cache the intermediate geoprocessing results which is far more the expensive process in the chain.  If we cache the GP results, changes to the input (precipitation) will generate much less load and complete quicker.

The view/job initiator could determine if the current scenario modifications have changed and pass in existing persisted results, or potentially the GP job itself could be smart enough to check - not sure what will work best for the distributed workflow.  

I think the implications mean we add 2 attributes to scenario:

* geop_results (text for json blob)
* inputs (text for json blob)

and we stop using `modifications` for both mods and inputs.  This bit will require a good few changes in the front end code for storing a loading the settings.

PR #553, opened 2015-07-07
Cache geoprocessing service results

caseypt:
When the modifications for a scenario don't change, but the inputs do, we now use the cached results from the geoprocessing endpoint that generates the landcover census.

**To Test:**
- Run `./scripts/manage migrate`, as well as `./scripts/npm install`.
- Regenerate the vendor bundle.
- Create a new project. The initial model run should take about 2 seconds.
- Change the precipitation level. The model run should finish in less than a second.
- Add a modification. This model run should take about 2 seconds.
- Change the precipitation level. The model run should finish in less than a second.
- Run the tests. Tests were added on both the front-end and back-end.

Connects to #512

### DIFF
--- a/src/mmw/apps/modeling/tests.py
+++ b/src/mmw/apps/modeling/tests.py
@@ -15,25 +15,57 @@ from rest_framework.test import APIClient
 
 
 class TaskRunnerTestCase(TestCase):
-
-    @override_settings(CELERY_ALWAYS_EAGER=True)
-    def test_tr55_job_runs_in_chain(self):
-        model_input = {
+    def setUp(self):
+        self.model_input = {
             'inputs': [
                 {
                     'name': 'precipitation',
                     'value': 1.2
                 }
             ],
-            'modifications': []
+            'modifications': [{
+                'name': 'conservation_practice',
+                'area': 15.683767964377065,
+                'value': 'rain_garden',
+                'shape': {
+                    'geometry': {
+                        'type': 'Polygon',
+                        'coordinates': [[
+                            [-75.09326934814453, 40.10092245173233],
+                            [-75.13343811035156, 40.060731050581396],
+                            [-75.0637435913086, 40.065460682065535],
+                            [-75.0692367553711, 40.095670021782404],
+                            [-75.09326934814453, 40.10092245173233]
+                        ]]
+                    },
+                    'type': 'Feature',
+                    'properties': {}
+                },
+                'units': 'km<sup>2</sup>',
+                'type': ''
+            }],
+            'area_of_interest': {
+                'type': 'MultiPolygon',
+                'coordinates': [[
+                    [[-75.06271362304688, 40.15893480687665],
+                     [-75.2728271484375, 39.97185812402586],
+                     [-74.99130249023438, 40.10958807474143],
+                     [-75.06271362304688, 40.15893480687665]]
+                ]]
+            },
+            'modification_hash': '39f488abec2de49f17652631ae843946'
         }
 
         created = now()
-        job = Job.objects.create(created_at=created, result='', error='',
-                                 traceback='', user=None, status='started')
-        job.save()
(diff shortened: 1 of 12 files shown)


src/mmw/apps/modeling/tests.py
```diff
@@ -45,6 +77,75 @@ def test_tr55_job_runs_in_chain(self):
...
+        self.assertEqual(task_count, 2,
+                         'Task list contains the wrong number of tasks')
+
+    @override_settings(CELERY_ALWAYS_EAGER=True)
+    def test_tr55_chain_generates_census_if_census_does_not_exist(self):
+        task_list = views._initiate_tr55_job_chain(self.model_input,
+                                                   self.job.id)
+
+        task_count = len(list(task_list._parents()))
+        self.assertEqual(task_count, 2,
```
caseypt: These tests aren't the greatest (ideally, we would inspect which tasks get run), but I spent a few hours trying to figure out how to do that and came up empty.

src/mmw/apps/modeling/views.py
```diff
@@ -210,10 +213,24 @@ def start_tr55(request, format=None):
...
+        # TODO: Remove. For demo purposes only
+        time.sleep(2)
+        return chain(tasks.prepare_census.s(model_input),
+                     tasks.run_tr55.s(model_input),
+                     save_job_result.s(job_id, model_input)) \
+            .apply_async(link_error=save_job_error.s(job_id))
+    else:
+        return chain(tasks.run_tr55.s(census, model_input),
+                     save_job_result.s(job_id, model_input)) \
+            .apply_async(link_error=save_job_error.s(job_id))
```
caseypt: As far as I know, `chain` doesn't directly support incrementally or conditionally adding tasks, so this isn't very DRY. It is however more readable than the alternatives I experimented with (and didn't get quite working).

src/mmw/apps/modeling/views.py
```diff
@@ -210,10 +213,24 @@ def start_tr55(request, format=None):
...
+    current_hash = model_input['modification_hash']
+    census = model_input.get('census')
+    census_hash = None
+
+    if census is not None:
+        census_hash = model_input['census'].get('modification_hash')
+
+    if census is None or current_hash != census_hash:
+        # TODO: Remove. For demo purposes only
+        time.sleep(2)
```
caseypt: This makes it easy to see when it's reusing the census. Not sure if we want to keep this though.

rajadain:
Before testing, we should also remember to restart Celery:

```bash
$ vagrant ssh worker -c 'sudo service celeryd restart'
```

caseypt:
@rajadain Good call. Though, just FYI, I didn't need to run that. It may be because I had Flower open the whole time, but I'm not totally sure. Changes to the tasks were picked up each time I saved the file.

src/mmw/apps/modeling/views.py
mmcfarland: I understand the temporary purpose of this sleep, though maybe it makes more sense to be in the `prepare_census` function, since that's what would get/not get called and take more/less time, depending, right?   Since it's async here, it should just return at about the same rate either way.  I realize it doesn't make a difference for testing, but I just want to make sure I'm understanding this flow.

caseypt: That's true. I'll move it there if you think we should keep it.

mmcfarland: I guess I don't think it's important to keep in for the merge.

mmcfarland:
+1 Tested:
* Works for Current Conditions as expected (only run 1x)
* Works for logged in & logged out users since it's generated on and sent from the client, not the server model.

caseypt:
Rebased. @azavea-bot rebuild.

rajadain:
+1 tested. Runs slower when I add modification / geo features, but faster when I only change the precipitation slider.

caseypt:
Rebased away the `time.sleep` in the view.

Issue #536, opened 2015-07-02
Setup prototype template for compare design


mmcfarland:
@ctaylo37 needs a place to develop a prototype for the compare designs, but being able to use the existing css.  We just need a sandbox that has the sass that we need so he can construct the html.

rajadain:
There already exists a [`compare.html`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/home/templates/home/compare.html). Should I update that so it plays well with changes made to `base.html` and looks like the prototype? Seems like it would be a fine sandbox since none of us touch it, and by extending `base.html` it gets all our styling.

mmcfarland:
Yeah, I think that's probably a good idea.

PR #554, opened 2015-07-07
Compare Design Sandbox

rajadain:
## Overview

Adds a sandbox for use by design team to mockup / prototype the Compare page. This sandbox imports all of our main styling (CSS) and behavior (JavaScript) that are available to all pages.

Changes include updating [`compare.html`](https://github.com/rajadain/model-my-watershed/blob/30283b2c3b41649938538314caa9e1ba989a67fb/src/mmw/apps/home/templates/home/compare.html) to match new conventions of `base.html`, adding some temporary custom styling to [`_compare.scss`](https://github.com/rajadain/model-my-watershed/blob/30283b2c3b41649938538314caa9e1ba989a67fb/src/mmw/sass/pages/_compare.scss), and moving the compare controller to its own file in [`compare/controllers.js`](https://github.com/rajadain/model-my-watershed/blob/30283b2c3b41649938538314caa9e1ba989a67fb/src/mmw/js/src/compare/controllers.js).

I had to change `<header id="compare-header">` to `<div id="compare-header">` because without it we would have two app headers:

image

This does not happen in the `/project` view because we actually [replace the second header after it is rendered](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/controllers.js#L108). Setting that up for this sandbox seemed extraneous, so I simply switched it to a `div`, adding a note to change back when making the view.

## Testing Instructions

Checkout the branch, run the bundle `./scripts/bundle.sh`, and go to [`/compare`](http://localhost:8000/compare). Ensure that things look reasonably like the prototype. This is what it the current mockup looks like:

image

Here's what the prototype looks like:

image

## Design Use Instructions

Once this has been merged, use `./scripts/bundle.sh --watch` (to auto convert SASS to CSS), then edit `/src/mmw/sass/pages/_compare.scss` and `/src/mmw/apps/home/templates/home/compare.html`, and see the output at [`http://localhost:8000/compare`](http://localhost:8000/compare)

Connects #536

### DIFF
--- a/src/mmw/apps/home/templates/home/compare.html
+++ b/src/mmw/apps/home/templates/home/compare.html
@@ -1,60 +1,63 @@
 {% extends 'base.html' %}
 
 {% block header %}
-<header id="compare-header"> <!-- Model Header -->
-
-     <div class="top"> <!-- Top Bar -->
-
-        <div class="project"> <!-- Project -->
-            <div class="back"> <!-- Back -->
-                <a href="model" class="btn btn-md btn-icon light"><i class="fa fa-arrow-left"></i></a>
-            </div> <!-- End Back -->
-            <div class="project-title"> <!-- Project Name -->
-                <h3 class="light">Compare Mode</h3>
-            </div> <!-- End Project Name -->
-        </div> <!-- End Project -->
-
-        <div class="top-bar"> <!-- Top Bar -->
-            <div class="tab-controls"> <!-- Tab Control Buttons -->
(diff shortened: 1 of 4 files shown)


mmcfarland:
Looks good to me :+1: I think @ctaylo37 will be making substantial changes to the general layout of this page.

@ctaylo37 note the instructions in the PR body for getting set up to work.

rajadain:
@ctaylo37 This is now merged and ready to be worked with. Please let me know if I can help you get set up with this.

Issue #538, opened 2015-07-02
Add basemap switcher and alternate basemaps


mmcfarland:
Use some free TMS satellite layer as a preliminary option (ESRI?)

Issue #450, opened 2015-06-17
Replace lodash with underscore


lliss:
We have to use underscore for Backbone. Lodash is almost the same thing but a bit better and we're using that on the project as well. We should try to reduce our bundle file size and code footprint so we should either:
A) Figure out how to allow backbone to use lodash and just use that.
B) Swap out lodash for underscore and remove lodash as a dependency.

I think A will be hard and B will be pretty easy.

caseypt:
I think lodash should work fine as a replacement for underscore with Backbone. There used to be a way to build a custom version of lodash that would work with underscore, but it looks like they stopped doing it and it should just work: https://github.com/lodash/lodash-cli/issues/32

alundiak:
Can someone give me OFFICIAL opinion by Backbone.JS how stable is migration/usage of Lodash.

I know these 3:
- Lodash
- Lodash.Underscore
- Lodash.Compat

I wanna make sure ALL of them will work correctly with ANY version of backbone starting from 1.1.x

PR #555, opened 2015-07-07
Add layer selector to map

lewfish:
To test, try switching to the ESRI World Imagery base layer using the Leaflet layer selector which should be visible on the map.

Connects #538

### DIFF
--- a/src/mmw/apps/core/templates/base.html
+++ b/src/mmw/apps/core/templates/base.html
@@ -32,6 +32,9 @@
     {% endblock modals %}
 
     {% block javascript %}
+    <script type="text/javascript">
+        window.clientSettings = {{ client_settings | safe }};
+    </script>
     <script type="text/javascript" src="{% static 'js/vendor.js' %}"></script>
     <script type="text/javascript" src="{% static 'js/main.js' %}"></script>
     {% endblock javascript %}
--- a/src/mmw/js/src/core/tests.js
+++ b/src/mmw/js/src/core/tests.js
@@ -14,6 +14,7 @@ var $ = require('jquery'),
     models = require('./models'),
     AppRouter = require('../router').AppRouter,
     chart = require('./chart'),
+    settings = require('./settings'),
     sandboxTemplate = require('./templates/sandbox.html');
 
 var TEST_SHAPE = {
@@ -178,6 +179,34 @@ describe('Core', function() {
 
                 model.setFullSize();
                 assert.isFalse($('#map').hasClass('half'));
+
+                view._leafletMap.remove();
+            });
+
+
+            it('creates a layer selector with the correct items', function() {
+                var baseLayers = {
+                    'A': {
+                        'url': 'https://{s}.tiles.mapbox.com/v3/ctaylor.lg2deoc9/{z}/{x}/{y}.png',
+                        'default': true
+                    },
+                    'B': {
+                        'url': 'https://{s}.tiles.mapbox.com/v3/examples.map-i86nkdio/{z}/{x}/{y}.png'
+                    }
+                };
+                settings.setSettings('base_layers', baseLayers);
+
+                var model = new models.MapModel(),
+                    view = new views.MapView({
+                        model: model
+                    }),
+                    $layers = $('.leaflet-control-layers-base'),
+                    layerNames = $layers.find('span').map(function() {
+                        return $(this).text().trim();
+                    }).get();
+
+                assert.equal($layers.length, 1, 'Did not add layer selector');
+                assert.deepEqual(layerNames, _.keys(baseLayers));
 
                 view._leafletMap.remove();
             });
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -2,7 +2,7 @@
 
 var $ = require('jquery'),
     L = require('leaflet'),
-    _ = require('lodash'),
+    _ = require('underscore'),
     router = require('../router.js').router,
     Marionette = require('../../shim/backbone.marionette'),
     TransitionRegion = require('../../shim/marionette.transition-region'),
@@ -15,7 +15,7 @@ var $ = require('jquery'),
     modalShareTmpl = require('./templates/shareModal.html'),
     modificationPopupTmpl = require('./templates/modificationPopup.html'),
     areaOfInterestTmpl = require('../core/templates/areaOfInterestHeader.html'),
-
+    settings = require('./settings'),
     ENTER_KEYCODE = 13,
     BASIC_MODAL_CLASS = 'modal modal-basic fade';
 
@@ -140,11 +140,6 @@ var MapView = Marionette.ItemView.extend({
 
     initialize: function() {
         var map = new L.Map('map', { zoomControl: false }),
-            // TODO: Replace tile layer, eventually.
-            tileLayer = new L.TileLayer('https://{s}.tiles.mapbox.com/v3/ctaylor.lg2deoc9/{z}/{x}/{y}.png', {
-                attribution: 'Map data &copy; <a href="http://openstreetmap.org">OpenStreetMap</a> contributors, <a href="http://creativecommons.org/licenses/by-sa/2.0/">CC-BY-SA</a>, Imagery © <a href="http://mapbox.com">Mapbox</a>',
-                maxZoom: 18
-            }),
             areaOfInterestLayer = new L.FeatureGroup(),
             modificationsLayer = new L.FeatureGroup(),
             maxZoom = 10,
@@ -154,9 +149,27 @@ var MapView = Marionette.ItemView.extend({
 
         map.addControl(new L.Control.Zoom({position: 'topright'}));
         addLocateMeButton(map, maxZoom, maxAge);
-        map.addLayer(tileLayer);
+
+        var baseLayers = _.mapObject(settings.getSettings().base_layers, function(layerData) {
+            return new L.TileLayer(layerData.url, {
+                attribution: layerData.attribution || '',
+                maxZoom: layerData.maxZoom || 18
+            });
+        }),
+            defaultLayerName = _.findKey(settings.getSettings().base_layers, function(layerData) {
+                return layerData.default;
+            }),
+            defaultLayer = baseLayers[defaultLayerName];
+
+        L.control.layers(baseLayers, {}, {autoZIndex:false}).addTo(map);
+
+        if (defaultLayer) {
+            map.addLayer(defaultLayer);
+        }
+
         map.addLayer(areaOfInterestLayer);
         map.addLayer(modificationsLayer);
+
         map.setView([40.1, -75.7], maxZoom); // center the map
 
         // Keep the map model up-to-date with the position of the map
--- a/src/mmw/mmw/settings/base.py
+++ b/src/mmw/mmw/settings/base.py
@@ -359,3 +359,16 @@ ITSI = {
     'access_token_url': 'auth/concord_id/access_token',
     'user_json_url': 'auth/concord_id/user.json',
 }
+
+BASE_LAYERS = {
+    'Default': {
+        'url': 'https://{s}.tiles.mapbox.com/v3/ctaylor.lg2deoc9/{z}/{x}/{y}.png',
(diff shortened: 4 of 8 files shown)


src/mmw/sass/base/_map.scss
```diff
@@ -24,6 +24,15 @@
 .leaflet-draw-toolbar a {
 	background-image: url('../images/spritesheet.png');
 }
+
 .leaflet-retina .leaflet-draw-toolbar a {
 	background-image: url('../images/spritesheet-2x.png');
 }
+
+.leaflet-control-layers-toggle{
+  background-image: url('/static/images/layers.png');
```
caseypt: I don't think we should reference `static` in the path, because it can change. FYI, it is configured here: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/base.py#L177.

Instead, we should use a relative path.

caseypt: Consider making a whitespace commit to fix the different indentation patterns in this file.

src/mmw/mmw/settings/base.py
```diff
@@ -368,3 +368,16 @@ def get_env_setting(setting):
     'access_token_url': 'auth/concord_id/access_token',
     'user_json_url': 'auth/concord_id/user.json',
 }
+
+BASE_LAYERS = {
+    'Default': {
+        'url': 'https://{s}.tiles.mapbox.com/v3/ctaylor.lg2deoc9/{z}/{x}/{y}.png',
```
caseypt: @mmcfarland Is there an Azavea or Stroud Mapbox account we can use?

src/mmw/js/src/core/views.js
```diff
@@ -154,11 +149,36 @@ var MapView = Marionette.ItemView.extend({
...
+                    map.addLayer(defaultLayer);
+                }
+            })
+            .fail(function() {
+                var defaultLayer = new L.TileLayer('https://{s}.tiles.mapbox.com/v3/ctaylor.lg2deoc9/{z}/{x}/{y}.png', {
+                    attribution: 'Map data &copy; <a href="http://openstreetmap.org">OpenStreetMap</a> contributors, <a href="http://creativecommons.org/licenses/by-sa/2.0/">CC-BY-SA</a>, Imagery © <a href="http://mapbox.com">Mapbox</a>',
+                    maxZoom: 18
+                });
+                map.addLayer(defaultLayer);
+            });
```
caseypt: I like this workflow for setting up the layer selector. Nice job.

src/mmw/js/src/core/views.js
```diff
@@ -154,11 +149,36 @@ var MapView = Marionette.ItemView.extend({
...
+                    });
+                }),
+                    defaultLayerName = _.findKey(response, function(layerData) {
+                        return layerData.default;
+                    });
+
+                L.control.layers(baseLayers).addTo(map);
+                if (defaultLayerName) {
+                    var defaultLayer = baseLayers[defaultLayerName];
+                    map.addLayer(defaultLayer);
```
caseypt: Consider abstracting out the `done` and `fail` handlers into helper functions.

src/mmw/apps/home/urls.py
```diff
@@ -4,7 +4,7 @@
 from __future__ import division
 
 from django.conf.urls import patterns, url
-from apps.home.views import home_page, project, compare
+from apps.home.views import home_page, project, compare, base_layers
```
lliss: As per @mmcfarland I'd strongly favor having this come in with the initial page rather than as a separate request.

src/mmw/js/src/core/views.js
```diff
@@ -154,11 +149,36 @@ var MapView = Marionette.ItemView.extend({
...
         map.setView([40.1, -75.7], maxZoom); // center the map
 
+         $.get('/base-layers/')
+            .done(function(response) {
+                var baseLayers = _.mapObject(response, function(layerData) {
+                    return new L.TileLayer(layerData.url, {
+                        attribution: layerData.attribution || '',
+                        maxZoom: layerData.maxZoom || 18
+                    });
+                }),
```
caseypt: Indentation seems off here.

src/mmw/js/src/core/views.js
```diff
@@ -2,7 +2,7 @@
 
 var $ = require('jquery'),
     L = require('leaflet'),
-    _ = require('lodash'),
+    _ = require('underscore'),
```
caseypt: Curious why this was changed. Every function that's in underscore should have a corollary in lodash.

src/mmw/js/src/core/tests.js
```diff
@@ -181,6 +181,54 @@ describe('Core', function() {
...
+                var model = new models.MapModel(),
+                    view = new views.MapView({
+                        model: model
+                    }),
+                    $layers = $('.leaflet-control-layers-base');
+
+                assert.equal($layers.length, 0, 'Added layer selector');
+
+                view._leafletMap.remove();
+            });
```
caseypt: Nice tests.

src/mmw/js/src/core/views.js
```diff
@@ -154,11 +149,36 @@ var MapView = Marionette.ItemView.extend({
...
+                    return new L.TileLayer(layerData.url, {
+                        attribution: layerData.attribution || '',
+                        maxZoom: layerData.maxZoom || 18
+                    });
+                }),
+                    defaultLayerName = _.findKey(response, function(layerData) {
+                        return layerData.default;
+                    });
+
+                L.control.layers(baseLayers).addTo(map);
```
lliss: I didn't realize the default functionality was that the popup shows on mouse hover. I'm not fond of that effect.

src/mmw/js/src/core/views.js
lliss: Might be related to: https://github.com/WikiWatershed/model-my-watershed/issues/450

caseypt:
@mmcfarland I'm into current approach as implemented. However, if we want to add more than just the base layer information, it might make more sense to load it all at once rather than having separate endpoints for different config groups or one that returns everything.

This is sort of a tanget, but I'm into this method because our initial page load time is pretty slow. It's mostly because of the JS vendor bundle, but I favor pushing things to async that aren't essential to help with the page load.

mmcfarland:
@caseypt Fair point.  I think I'd prefer there be a hierarchical api structure for these thing then, like `/config/base-layers` so they end up too scattered.

lliss:
We do need to reduce the vendor bundle if we can. That's one reason to replace lodash with underscore. For some reason I recall though haven't verified recently that lodash was a larger lib that underscore. Either way, I don't think adding a bit of config here is going to slow down the initial load very much. We're talking a few hundred bytes vs a full round trip HTTP request with the SSL handshake. Can't see how the former could possibly be slower.

caseypt:
@lliss True that this wouldn't add much to the page load. 

It might be worth seeing how other projects accomplish this, or if there is a recommended way from Django to get settings to the client.

lliss:
I'm noticing that all the landcover SVGs are on the initial page. Might be worth moving those off to a separate file since they won't get loaded until the user accesses a project.

caseypt:
@lliss has convinced me that the approach suggested by @mmcfarland is probably the way to go. There seems to be a bunch of low hanging fruit to reduce the page load (minify vendor in staging, remove unused or duplicate libraries, off-load things that aren't needed on the initial page), rather than have this async, which won't help as much.

src/mmw/js/src/core/views.js
caseypt: Adding `{ collapsed: false }` should fix it.

lliss:
All other comments aside, I did test this and it does work. It's a totally workable solution though I think we're all leaning towards a reworked implementation to reduce lag a bit.

src/mmw/js/src/core/views.js
lewfish: Yes, I did that because I thought we were switching to underscore.

rajadain:
I'm getting the following errors on testing:

image

Steps to reproduce:

 1. Switch to ESRI World Imagery and draw a shape
 2. In the `/analyze` view with the shape visible, select Default tiles
 3. Select ESRI World Imagery tiles. Tiles do not show up and there are two `Cannot read property 'length' of undefined` errors
 3. Select Default tiles. Tiles render, but there are three `Cannot read property 'parentNode' of null` errors.
 4. Keep switching to produce more errors.

src/mmw/js/src/core/views.js
rajadain: I don't mind the hover / collapse effect too much, although I should probably test it on a touch device. What bothers me more is how different the layer control looks compared to other icons:

image

kdeloach:
@caseypt 

> It might be worth seeing how other projects accomplish this, or if there is a recommended way from Django to get settings to the client.

Here's how OTM2 does it: https://github.com/OpenTreeMap/otm-core/blob/master/opentreemap/treemap/templates/treemap/settings.js

src/mmw/js/src/core/views.js
lliss: Considering the implications on a touch screen is a good call.

rajadain: Just tried it on the iPad, works well. Tap once to expand, then tap to select option. The expanded form stays open until you tap elsewhere.

rajadain:
@kdeloach 

Regarding using a templated JavaScript file, will it play well with Browserify which will only load it once? How do we load it again after server variables have been changed?

lewfish:
I think I've addressed all the comments. I'm now passing in the settings via home.html, which has a script tag that attaches the settings to `window.clientSettings`. The settings are accessed by the app indirectly via core/settings.js which references `window.clientSettings` when they are available, and uses default settings when they are not (which should only occur when running testem tests).

rajadain:
Something seems to be missing. `get_home_page_settings` is undefined. There are also other flake8 errors. See the build's console log for details.

lewfish:
The build passed. I'll start using the test.sh script before pushing for now on.

rajadain:
FYI I'm still seeing the map switching errors in the `/analyze` view

lewfish:
After spending some time trying to fix the bug that @rajadain found, and talking about it with @mmcfarland, we decided to postpone any further effort to fix the bug since we might not be showing the layer selector in the Analyze view. https://github.com/WikiWatershed/model-my-watershed/issues/559

rajadain:
It might be best to only show the selector in the `/` view, rather than just hide it in the `/analyze` view, because I'm pretty sure that it also doesn't work in the `/project` view.

lliss:
Switching the base map after drawing and moving to analyze mode results in errors.

screencap

lewfish:
@lliss This is what I was referring to two comments up and in #559.

lliss:
Ok. Then +1.

src/mmw/js/src/core/views.js
caseypt: @mmcfarland Any thoughts on this? Seems like we should drop one of the `_`'s.

src/mmw/apps/core/templates/base.html
```diff
@@ -32,6 +32,9 @@
     {% endblock modals %}
 
     {% block javascript %}
+    <script type="text/javascript">
+     window.clientSettings = {{ client_settings | safe }};
```
caseypt: Indentation

src/mmw/mmw/settings/base.py
mmcfarland: Not to my knowledge on either, but they've expressed appreciation of the basemap we've been using.  We'll have to go through the rigmarole of explaining the options and the potential cost of using the MapBox one, unless they have a free tile set that's similar in style to this. I'll put it on the agenda for next week.

src/mmw/js/src/core/views.js
mmcfarland: Yeah, we should drop one.  I'm a little partial to the expanded `lodash` methods, and if there is a version compatible with backbone, I'd be happy to use that.  Aside from that, they're both solid libraries and I haven't got a strong opinion.  I suspect the size difference between the two is negligible, so I think it comes down to backbone support.  If lodash brings in a ton of dependencies, that would also be a strike against it for me.

Issue #548, opened 2015-07-02
Store the results of complete model runs


mmcfarland:
Determine if the scenario has inputs/mods changed since the most recent results, which should be saved after the job runs.  Loading a saved project or refreshing the page should not generate any job requests.

PR #567, opened 2015-07-10
Cache Model Run Results to Scenarios

rajadain:
## Overview

Enables reading saved results from scenario when requesting existing project or scenario from server, thus reducing requirement for repeated model runs.

Previously, loading a saved project incurred extraneous model run requests, one per scenario:

image

Now we do not re-run the model if there are no changes:

image

However, if for whatever reason a scenario does not have a saved result, then the model is run on project load, and the results subsequently saved:

image

The scenario is **always** saved after a model run, even if a model run failed. In this case, the results would be set to `[]`, but other changes, such as inputs and modifications, will be saved to the database. This allows the user to re-open the project at a later date, at which time the model will be re-run with the saved inputs and modifications from last time.

~~Also, by saving only after a model run, we minimize duplicate save requests for the same scenario, instead grouping all changes into one `PUT`.~~

Since model runs can potentially take a long time, we also save the scenario inputs and modifications before starting the model run. This takes care of the edge case when a user closes the browser window before the model run completes, and loses their work. Whenever there are changes to `inputs` or `modifications`, we update the corresponding `inputmod_hash` and `modification_hash` fields. When starting model runs, we pass in these hashes, and they are included in the model output, which is then saved to the `modeling_scenario.results` field. When loading an existing project, we check to see if the `results` field is empty, and if it isn't, if the `inputmod_hash` matches that on the scenario, indicating a _stale_ result. If the `results` are empty or stale, we run the model, else we do nothing since everything is up to date.

## Data Changes

We had so far not marked `modeling_scenario.results` as a `JsonField`. Thus the existing data was incompatible, since it had the DRF `u` prefixed strings that was fixed for the modifications field in #320. Thus, we have a [data migration](https://docs.djangoproject.com/en/1.8/topics/migrations/#data-migrations) that sets the `results` field to `[]` for each row that had the `u` prefix. Thus, when such a scenario is loaded the next time, its results will be recalculated and saved in the correct format for the future.

We also add the `modeling_scenario.inputmod_hash` field in a second migration.

## Testing Instructions

 1. Run bundle and migrations `./scripts/bundle.sh && ./scripts/manage.sh migrate`
 2. Log in as a user and create a new project.
 3. Reload this existing project. Ensure that model calculations are not run for freshly loaded / unchanged project.
 4. Make arbitrary changes to the scenario: add modifications, change the slider, jump around between tabs. Ensure that after you stop and refresh the page, the previous state is recreated faithfully.
 5. Load an old project (if you have any on your system) which would have had its `results` emptied as part of the migration. Ensure that model calculations are run and saved. Reload the same project and ensure that model calculations are not re-run.
 6. Log out, and create a new project. Ensure that while model calculations are run, the scenario is not saved. Make arbitrary changes and then log in to save the project. Reload the project and ensure that model calculations were not run.
 7. Log out, and open a public project. Make changes, adjust the slider, ensure that the scenario is not saved after model calculations are run by reloading the page and finding it restored to its original state.
 8. Ensure that the model output graphs are rendered correctly when loading existing projects, and after model calculations are run.

Connects #548

### DIFF
--- a/src/mmw/js/src/core/utils.js
+++ b/src/mmw/js/src/core/utils.js
@@ -1,6 +1,7 @@
 "use strict";
 
-var _ = require('underscore');
+var _ = require('underscore'),
+    md5 = require('blueimp-md5').md5;
 
 var utils = {
     // Parse query strings for Backbone
@@ -28,6 +29,40 @@ var utils = {
             );
         }
         return params;
+    },
+
+    // Convert a backbone collection into a canonical MD5 hash
+    getCollectionHash: function(collection) {
+        // JSON objects are not guaranteed consistent order in string
+        // representation, so we convert to sorted array which can be
+        // stringified reliably.
+
+        // Converts an object with JSON representation:
+        //     {name: 'abc', type: 'xyz', shape: [...]}
+        // to:
+        //     [{name: 'abc'}, {shape: [...]}, {type: 'xyz'}]
+        var objectToSortedArray = function(object) {
+                var sorted = [];
+                Object.keys(object).sort().forEach(function(key) {
+                    var kvPair = {},
+                        value = object[key];
+                    // Recursively sort nested objects
+                    if (_.isObject(value) && value !== null && !_.isArray(value)) {
+                        value = objectToSortedArray(value);
+                    }
+                    kvPair[key] = value;
+                    sorted.push(kvPair);
+                });
+
+                return sorted;
+            },
+            // Sort the collection by given sortKey and convert
+            // each model to a sorted array
+            sortedCollection = collection.sort().map(function(model) {
+                return objectToSortedArray(model.toJSON());
+            });
+
+        return md5(JSON.stringify(sortedCollection));
     }
 };
 
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -2,7 +2,7 @@
 
 var Backbone = require('../../shim/backbone'),
     _ = require('underscore'),
-    md5 = require('blueimp-md5').md5,
+    utils = require('../core/utils'),
     App = require('../app'),
     coreModels = require('../core/models');
 
@@ -21,7 +21,11 @@ var ModelPackageControlModel = Backbone.Model.extend({
 });
 
 var ModelPackageControlsCollection = Backbone.Collection.extend({
-    model: ModelPackageControlModel
+    model: ModelPackageControlModel,
+
+    comparator: function(model) {
+        return model.get('name');
+    }
 });
 
 var ModelPackageModel = Backbone.Model.extend();
@@ -48,6 +52,10 @@ var ResultModel = Backbone.Model.extend({
 
 var ResultCollection = Backbone.Collection.extend({
     model: ResultModel,
+
+    comparator: function(model) {
+        return model.get('name');
+    },
 
     setPolling: function(polling) {
         this.forEach(function(resultModel) {
@@ -240,7 +248,16 @@ var ModificationModel = coreModels.GeoModel.extend({
 });
 
 var ModificationsCollection = Backbone.Collection.extend({
-    model: ModificationModel
+    model: ModificationModel,
+
+    comparator: function(model) {
+        // Even though model.get('area') is a numeric value, passing it
+        // along with the others here causes a string comparison. But that's
+        // alright, because we only want to sort consistently, not actually
+        // by area, and it serves as a tie-breaker when the other values are
+        // identical.
+        return [model.get('name'), model.get('value'), model.get('area')];
+    }
 });
 
 var ScenarioModel = Backbone.Model.extend({
@@ -280,6 +297,9 @@ var ScenarioModel = Backbone.Model.extend({
 
         this.on('change:project change:name change:census', this.attemptSave, this);
         this.get('inputs').on('add', this.attemptSave, this);
+        this.updateModificationHash();
+        this.updateInputModHash();
+
         this.get('modifications').on('add remove change', this.updateModificationHash, this);
         this.get('modifications').on('add remove', this.attemptSave, this);
 
@@ -350,6 +370,8 @@ var ScenarioModel = Backbone.Model.extend({
     // Poll the taskModel for results and reset the results collection when done.
     // If not successful, the results collection is reset to be empty.
     getResults: function() {
+        this.updateInputModHash();
+
         var self = this,
             results = this.get('results'),
             taskModel = this.get('taskModel'),
@@ -415,8 +437,18 @@ var ScenarioModel = Backbone.Model.extend({
         taskModel.start(taskHelper);
     },
 
+    updateInputModHash: function() {
+        var hash = utils.getCollectionHash(this.get('inputs'));
+
+        if (this.get('modification_hash')) {
+            hash += this.get('modification_hash');
+        }
+
+        this.set('inputmod_hash', hash);
+    },
+
     updateModificationHash: function() {
-        var hash = md5(JSON.stringify(this.get('modifications')));
+        var hash = utils.getCollectionHash(this.get('modifications'));
 
         this.set('modification_hash', hash);
     }
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -40,6 +40,7 @@ var ResultModel = Backbone.Model.extend({
     defaults: {
         name: '',
         displayName: '',
+        inputmod_hash: null,
         result: null,
         polling: false
     }
@@ -250,6 +251,7 @@ var ScenarioModel = Backbone.Model.extend({
         is_current_conditions: false,
         user_id: 0, // User that created the project
         inputs: null, // ModificationsCollection
+        inputmod_hash: null, // MD5 string
         modifications: null, // ModificationsCollection
         modification_hash: null, // MD5 string
         active: false,
@@ -360,7 +362,10 @@ var ScenarioModel = Backbone.Model.extend({
                     results.forEach(function(resultModel) {
                         var resultName = resultModel.get('name');
                         if (serverResults[resultName]) {
-                            resultModel.set('result', serverResults[resultName]);
+                            resultModel.set({
+                                'result': serverResults[resultName],
+                                'inputmod_hash': serverResults.inputmod_hash
+                            });
                         } else {
                             console.log('Response is missing ' + resultName + '.');
                         }
@@ -378,6 +383,7 @@ var ScenarioModel = Backbone.Model.extend({
                         modifications: self.get('modifications').toJSON(),
                         area_of_interest: App.currProject.get('area_of_interest'),
                         census: self.get('census'),
+                        inputmod_hash: self.get('inputmod_hash'),
                         modification_hash: self.get('modification_hash')
                     })
                 },
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -199,13 +199,12 @@ var ProjectModel = Backbone.Model.extend({
             var user_id = response.user.id,
                 scenariosCollection = this.get('scenarios'),
                 scenarios = _.map(response.scenarios, function(scenario) {
-                    // TODO We don't want to set the results until a future
-                    // PR when we intentionally cache results.
-                    delete scenario.results;
-
                     var scenarioModel = new ScenarioModel(scenario);
                     scenarioModel.set('user_id', user_id);
                     scenarioModel.get('modifications').reset(scenario.modifications);
+                    if (!_.isEmpty(scenario.results)) {
+                        scenarioModel.get('results').reset(scenario.results);
+                    }
 
                     return scenarioModel;
                 });
@@ -295,13 +294,11 @@ var ScenarioModel = Backbone.Model.extend({
         this.set('inputs', new ModificationsCollection(attrs.inputs));
         this.set('modifications', new ModificationsCollection(attrs.modifications));
 
-        this.on('change:project change:name change:census', this.attemptSave, this);
-        this.get('inputs').on('add', this.attemptSave, this);
         this.updateModificationHash();
         this.updateInputModHash();
 
+        this.on('change:project change:name', this.attemptSave, this);
         this.get('modifications').on('add remove change', this.updateModificationHash, this);
-        this.get('modifications').on('add remove', this.attemptSave, this);
 
         var debouncedGetResults = _.debounce(_.bind(this.getResults, this), 500);
         this.get('inputs').on('add', debouncedGetResults);
@@ -351,17 +348,23 @@ var ScenarioModel = Backbone.Model.extend({
         this.get('inputs').reset(response.inputs);
         delete response.inputs;
 
-        // TODO We don't want to set the results until a future
-        // PR when we intentionally cache results.
+        if (!_.isEmpty(response.results)) {
+            this.get('results').reset(response.results);
+        }
         delete response.results;
 
         return response;
     },
 
     getResultsIfNeeded: function() {
-        var needsResults = this.get('results').some(function(resultModel) {
-            return !resultModel.get('result');
-        });
+        var inputmod_hash = this.get('inputmod_hash'),
+            needsResults = this.get('results').some(function(resultModel) {
+                var emptyResults = !resultModel.get('result'),
+                    staleResults = inputmod_hash !== resultModel.get('inputmod_hash');
+
+                return emptyResults || staleResults;
+            });
+
         if (needsResults) {
             this.getResults();
         }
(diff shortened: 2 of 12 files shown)


mmcfarland:
Since reported failures write a `[]` to results, we're mostly avoiding the case where the inputs + mods can get out of sync with the results.  There are a few edge cases such as closing the window before polling ends, such that the saved results are not reflective of the mods + inputs.  I think a simple safety check would be to either:

1. Clear out the results prior to starting a model run
2. Include a hash of inputs + mods on the result set, and before assuming that the results don't need to be run, insure that recalculating the hash produces the same value on the existing results.  

(2) has the benefit of being a stronger check and doesn't incure the additional http request for the save that (1) will do on the attribute change event and is similar to what we're doing with the geop results.

rajadain:
We save the `modificaiton_hash` in both the Scenario and the Results set, since we have to check on the server side whether redoing the geoprocessing is required:

screen shot 2015-07-13 at 11 36 22 am

We don't _have_ to do that with an `input_hash` per se. Using a hash this is the workflow I'm envisioning:

 1. When requesting a model run, we calculate and send an `input_hash` in addition to the `modification_hash`.
 2. When we receive results of this model run, we save the results to the Scenario's `results` field.
 3. When loading an existing project, we calculate the `input_hash` on the client side after project has been loaded and check that against the `input_hash` field in each result. If any result is stale, we re-run the model.

Should I still add an `input_hash` field to the Scenario, even though we don't need it right now, for symmetry? Or leave it until we need it, a-la YAGNI?

rajadain:
Just talked to @caseypt to reorient myself since he helped me reason through this the first time. The edge case of "closing the window before polling ends, such that the saved results are not reflective of mods + inputs" is not really possible, since we only save **after** polling ends. Thus, the inputs, modifications, and results are always saved in sync. When they change inputs and / or modifications, we only trigger polling and not saving. If polling is interrupted, then neither the new inputs + mods nor the new results will be saved. Thus, the state in the server is still valid. If this happens and they reload the page, they will receive the old inputs and modifications and results, not new inputs and modifications with old results.

lliss:
Think this is worth adding a unit test?

rajadain:
It is definitely complicated enough to warrant some testing. I wasn't sure if unit testing would be enough or if we'd need some for of UI / integration testing. Let me see if I can add some unit tests to at least ensure the isolated functionality.

src/mmw/js/src/modeling/models.js
```diff
@@ -276,10 +275,8 @@ var ScenarioModel = Backbone.Model.extend({
         this.set('inputs', new ModificationsCollection(attrs.inputs));
         this.set('modifications', new ModificationsCollection(attrs.modifications));
 
-        this.on('change:project change:name change:census', this.attemptSave, this);
-        this.get('inputs').on('add', this.attemptSave, this);
+        this.on('change:project change:name', this.attemptSave, this);
         this.get('modifications').on('add remove change', this.updateModificationHash, this);
-        this.get('modifications').on('add remove', this.attemptSave, this);
```
lliss: So we no longer save at the time the modification is made. That's find presuming short turn around for processing. But we might need to be hanging out on the order of minutes and if we don't save until the very end, the chance for data loss increases significantly.

lliss:
Overall +1. Code looks good and everything seems to work. Let me know what you end up doing about the test. I think a simple backend API endpoint test would be a good start. UI tests wouldn't hurt either but will probably be a lot more complicated and I'd check with @mmcfarland about whether that's worth the time investment. I am mildly concerned about the issue I noted in that we don't save the modifications after the UI changes any more and I think that might cause issues if there's a long turn around time on the processing.

src/mmw/js/src/modeling/models.js
rajadain: This is true. I wonder if we can get @hectcastro to setup an alert / visualization to see average processing time for this to complete, and update our code based on that behavior. Also note that we poll for the job a limited number of times. If after that the job is still not complete, we abandon the results and save everything else.

src/mmw/apps/modeling/migrations/0008_scenario_results_to_json.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def convert_to_json(apps, schema_editor):
```
rajadain: Initially I had attempted to convert existing results to proper JSON, but clearing them out was simpler. I should probably update this functions name to reflect that.

src/mmw/js/src/modeling/models.js
lliss: I think we'll end up tweaking the timeout on the polling. Initial results of the real geotrellis endpoint is making me think we'll be in the 20sec+ range.

hectcastro: If we want to keep an eye on the duration of background tasks, we can start by looking at the metrics already kept by Celery. If we want something more specific, we can use any of the following techniques to instrument the functions run by Celery:

- http://statsd.readthedocs.org/en/latest/timing.html#calling-timing-manually
- http://statsd.readthedocs.org/en/latest/timing.html#using-a-context-manager
- http://statsd.readthedocs.org/en/latest/timing.html#using-a-decorator
- http://statsd.readthedocs.org/en/latest/timing.html#using-a-timer-object-directly

The StatsD client and configuration is available across all of the machine types. Anything working through Django should already have it configured, just need to import the client:

```python
from django_statsd.clients import statsd

with statsd.timer('foo'):
    foo()
```

rajadain:
Adding tests to this functionality is pretty complicated since we want to test the results of polling, which is triggered by a method inside the ScenarioModel which passes locally defined callbacks to an instance of TaskModel. In 46ccaf5, I'm using delays instead of callbacks since the actual callbacks are buried so many layers deep and I thought it best to do that as part of #393.

Testing the entire project load to ensure that we don't run the scenarios again seems more integration / UI test territory.

The fact that a model run can take 20 seconds or more is a little worrying. The reason I deferred saving inputs and modifications until after results are available was because it was almost instantaneous and we were POSTing twice, almost the same information, in a matter of milliseconds. However, in this case, I would go with @mmcfarland 1. alternative above, where we first save the inputs and modifications with cleared results when starting the model run, and then save it again with everything once it is finished. Does that sound okay?

lliss:
I'll defer to @mmcfarland's judgement here but I think going with the plan he outlined is probably going to be the best way to handle it. I think we're going to have to assume long turn around time on model runs once we have a real geoprocessing service.

mmcfarland:
I'm hoping for < 20 seconds GP time on these TR55 runs for sure, I think our VMs may not reflect production times.  That said, I'm still nervous about the gap between saving; I think presenting the user with model results that don't reflect the current inputs is about as bad of a silent error as we could give.  I think incurring the additional save and blanking the current results is probably the lowest-cost way to prevent this edge case.

rajadain:
The only problem with that is if the model run takes a long time and we clear the result _before_ starting it, the model output results pane will be empty for quite a while. I do like being able to see the results hang around until new ones are available, instead of seeing white and then the new results. Let me see if I can find a way to _save_ cleared results without clearing them in the model / affecting the view. Otherwise we might clear them out, save them, and then restore them pending the runs.

rajadain:
After working on this extensively last night, I have been unable to find a way to save cleared results without affecting the UI, thus we do get a white screen until the results are available. Given this, I am now attempting the second alternative @mmcfarland suggests above with using a new field for the input hash.

rajadain:
Okay I've updated the description and commits of this PR. Main new thing is ada86a1 and the `input_hash` field. Essentially we initialize it with a default value which corresponds to a precipitation of 1" when a new project is made, and update it whenever there is a change to the `inputs` collection. When a model run is triggered, we save the scenario, including `inputs`, `modifications`, and `input_hash`. We also send this `input_hash` to the modeling endpoint, and it is included in the results we receive. When we receive the results, we update the scenario with them and then save the scenario again.

When loading  an existing scenario, we check if the `results` are empty, and if they are not we compare the scenario's `input_hash` with that of the result. If they do not match, we assume that the results are _stale_. In both cases of empty or stale results, we run the model again. Otherwise, the results are current and the model does not need to be re-run.

rajadain:
I will squash and rebase before merging: I'm leaving the commits open now for clarity and readability.

src/mmw/js/src/modeling/models.js
```diff
@@ -270,16 +271,18 @@ var ScenarioModel = Backbone.Model.extend({
                     name: 'precipitation',
                     value: 1.0
                 }
-            ]
+            ],
+            // This is the md5 hash of the above default value
+            inputmod_hash: 'ab0d9b8a1d441b8ec7ddf81b5a49dd2d'
```
mmcfarland: Is it important to set this value initially?  I suppose that if we're defaulting a value, it's hard to pick up an event to calculate it on first run.  If we need it to be set, let's make it less brittle and just calculate the md5 via [hashlib(]https://docs.python.org/2/library/hashlib.html) so we can muck around with the default input with ease.

mmcfarland:
Great updates, looks good.  I'd recommend the one hash suggestion prior to merging

src/mmw/js/src/modeling/models.js
rajadain: This is a really funny problem to tackle. I agree with what you're saying though: having a hard coded hash is pretty brittle. The reason we set it here is that in Backbone the JSON string representation of the inputs is this:

```
[{"name":"precipitation","value":1.8,"type":"","shape":null,"area":"0","units":"m<sup>2</sup>"}]
```

whereas the value that is initialized from the server is this:

```
[{"name":"precipitation","area":"0","value":1.8,"shape":null,"units":"m<sup>2</sup>","type":""}]
```

Identical objects, but different string representations, leading to different MD5 hashes. Thus, when loading an existing project, a client-side generated MD5 hash (of the server representation) would not match the saved MD5 hash (of the client representation) from the server, leading to a redundant model run when it is not required.

I'm looking in to what can be done about it.

mmcfarland: That's a good point, and furthermore, json attributes are not guaranteed to be in any order (perhaps different browser `stringify` implementations could order them differently).  Maybe we should be very explicit about the string we hash, convert the object to an array of key/values (the attributes), order by key name, then hash?  That way we could unify across implementations/platforms/languages with higher confidence.

rajadain:
I've added sorting and standardization of inputs before hashing in a72de4c. Since this is now a costly operation, I've removed listeners on input change and instead trigger it just before saving to server so we use it as sparingly as possible.

By the same token, should we also update the modifications hash calculation since it would be vulnerable to the same inconsistencies?

I used [`Object.keys`](http://kangax.github.io/compat-table/es5/#Object.keys) to help sort. It seems to be supported in all major browsers.

src/mmw/js/src/modeling/models.js
```diff
@@ -400,9 +413,29 @@ var ScenarioModel = Backbone.Model.extend({
...
 
+        self.updateInputHash();
+        self.attemptSave();
         taskModel.start(taskHelper);
     },
 
+    updateInputHash: function() {
+        var standardizedInputs = this.get('inputs').sortBy('name').map(function(inputModel) {
+            var input = inputModel.toJSON(),
+                sortedInput = {};
```
mmcfarland: I don't think this implementation is correct, but the idea looks right.  The result here is an array with one element which is essentially the same object that was started with `[{name: "", shape: "", etc...]`, I think we really want an array with key/value pairs `[{name: ""}, {shape: ""}, etc..]` since array index order is guaranteed.  I believe we were avoiding stringifying on a multi-attributed object since key order was not assured.

rajadain: Ah! Got it. Let me fix that up real quick.

rajadain:
@mmcfarland I'm very impressed that you spotted the mistake in my implementation. It was so subtle to me that I would've never thought of it myself, and only come upon it by debugging.

I've just added the corrected version in 577df72. Also added a small comment to explain what is going on. You can inspect the results by adding these two lines [here](https://github.com/rajadain/model-my-watershed/commit/577df72b71abab487af203a60720ad4ec0b07273#diff-9abeff034ba5a68b4814ee90721888ebR441):

```javascript
console.log(JSON.stringify(standardizedInputs));
console.log(hash);
```

The result looks like this, and works correctly:

image

src/mmw/js/src/modeling/models.js
```diff
@@ -342,6 +349,9 @@ var ScenarioModel = Backbone.Model.extend({
     // Poll the taskModel for results and reset the results collection when done.
     // If not successful, the results collection is reset to be empty.
     getResults: function() {
+        this.updateInputHash();
```
rajadain: Had to move these up so that `taskHelper` is initialized with the correct values. `updateInputHash` must execute before we assign `inputmod_hash` to `taskHelper`.

rajadain:
I will probably rename the `updateInputHash` function to `updateInputModHash` before merging it in. Didn't want to do it right now since it would make in-progress code review more confusing.

mmcfarland:
+1 Verified that reloading a saved scenario does not result in any new jobs being submitted.

src/mmw/js/src/core/utils.js
```diff
@@ -0,0 +1,43 @@
...
+
+    getCollectionHash: function(collection, sortKey) {
+        // JSON objects are not guaranteed consistent order in string
+        // representation, so we convert to sorted array which can be
+        // stringified reliably.
+
+        // Converts an object with JSON representation:
+        //     {name: 'abc', type: 'xyz', shape: [...]}
+        // to:
+        //     [{name: 'abc'}, {shape: [...]}, {type: 'xyz'}]
```
lliss: Do any of our objects have sub-objects?

src/mmw/js/src/modeling/models.js
```diff
@@ -3,6 +3,7 @@
 var Backbone = require('../../shim/backbone'),
     _ = require('underscore'),
     md5 = require('blueimp-md5').md5,
```
lliss: I don't see md5 being used anywhere in this file.

src/mmw/js/src/modeling/tests.js
```diff
@@ -376,43 +377,124 @@ describe('Modeling', function() {
...
 
                     model.addModification(new models.ModificationModel(modificationsSample1));
                     assert.isTrue(modSpy.calledBefore(saveSpy));
+                });
+            });
+
+            describe('#scenarioPolling', function() {
+                it('saves after polling finishes successfully', function(done) {
+                    var model = getTestScenarioModel(),
+                        tr55Response = '{"status":"started", "job":"8aef636e-2079-4f87-98dc-471d090141ad"}',
```
lliss: We may want to move some of these huge json responses over to their own file for convenience. It might also encourage reuse.

src/mmw/js/src/core/utils.js
rajadain: Yes. A modification, for example, has this structure:

```json
    {
        "area": 648.1614728140878,
        "name": "landcover",
        "shape": {
            "geometry": {
                "coordinates": [
                    [
                        [
                            -75.1647502183914,
                            39.95236473500341
                        ],
                        [
                            -75.16454637050629,
                            39.95232772476147
                        ],
                        [
                            -75.16461610794067,
                            39.952011080761736
                        ],
                        [
                            -75.1648199558258,
                            39.952031642104906
                        ],
                        [
                            -75.1647502183914,
                            39.95236473500341
                        ]
                    ]
                ],
                "type": "Polygon"
            },
            "properties": {},
            "type": "Feature"
        },
        "type": "",
        "units": "m<sup>2</sup>",
        "value": "grassland"
    }
```
That's why we recursively sort nested objects in lines [22-25](https://github.com/WikiWatershed/model-my-watershed/pull/567/files#diff-8b62826bd98f0674cd1f97e9bb645f8cR22) below.

src/mmw/js/src/modeling/models.js
rajadain: Good catch. I missed it in lint. Will remove.

src/mmw/js/src/modeling/tests.js
rajadain: Splendid idea.

rajadain:
@lliss thanks for looking through the code. I just pushed 22b0eef with your suggestions.

Another problem with the current implementation of `getCollectionHash` that I've been thinking about is that while it sorts a collection by one key, it is very possible for there to be duplicates / collisions, thus causing inconsistencies (think multiple modifications of the same type). I think we might need to add a custom [`comparator`](http://backbonejs.org/#Collection-comparator) to all collections that we expect to hash (`inputs` and `modifications` so far) so that we have a canonical string representation of each.

rajadain:
Rebased on top of `develop` and added a custom comparator to address the issue of multiple modifications with identical `name`s and `value`s. We use `area` as the tie-breaker. Backbone does a string-comparison by default, so sorting by `area` won't be a numeric sort. But that's okay because we're only looking for a consistent ordering, not an ascending ordering.

I will merge this in this evening before leaving, in case there are any other final comments.

PR #568, opened 2015-07-10
Simplify Celery chain assembly and testing

rajadain:
## Overview

By separating out the construction for the TR55 job chain, we make it more readable and more testable. The tests are now more robust because we are checking for presence of specific methods, rather than just the number of tasks, which may change in the future if other tasks are added.

## Testing Instructions

Run all Python tests. Ensure that TR55 models can still be run from the UI.

Connects #512 by slightly improving #553


caseypt:
+1. Tested. Nice improvements on #553.

PR #571, opened 2015-07-13
Redirect STDERR for Celery service restart

hectcastro:
On `worker` provisioning attempts that trigger a restart of the Celery service, the shell provisioner step that follows triggers an error message indicating that the Celery service PID file already exists. This change suppresses that error message since it doesn't appear to impact the service's availability.


caseypt:
+1. Provisioned worker and didn't get any errors. Verified that Celery is up and running.

PR #572, opened 2015-07-13
Set `client_settings` in `/compare` sandbox

rajadain:
Without this JavaScript would fail on this page

## Testing Instructions

Go to [`/compare`](http://localhost:8000/compare). You should not see this error:

image


lewfish:
+1, tested.

Issue #6, opened 2015-03-25
Investigate feasibility of using spark-jobserver


hectcastro:
[`spark-jobserver`](https://github.com/spark-jobserver/spark-jobserver) provides a REST interface for submitting Apache Spark jobs, JARs, and contexts. I could see this being the interface for interacting with  GeoTrellis Spark from Celery workers.

Lots of questions still around how to spread the `spark-jobserver` across AZs, and how to remove or mitigate the Spark Context single point of failure.

---

### Resources

- https://www.youtube.com/watch?v=oznpiqBf4ZY

jamesmcclain:
I have been able to do a little work toward getting the MMW Scala code working on `spark-jobserver`: https://gist.github.com/jamesmcclain/86a131db281c6c490dbc#file-mmw-diff .  Applying that diff to the current mmw tree will allow you to build a jar file which *might* be compatible with `spark-jobserver`.

```shell
patch -p1 < mmw.diff
cd src/geop
sbt 'project  sparkjob' compile assembly
```

It is just the word count example grafted into the existing tree and Build.scala.  (I just realized that this diff does not include the word count code, but it is largely identical to https://github.com/spark-jobserver/spark-jobserver/blob/master/job-server-tests/src/spark.jobserver/WordCountExample.scala with `package spark.jobserver` replaced by `package com.azavea.mmw` and the addition of `import spark.jobserver.{SparkJob, SparkJobValidation, SparkJobValid, SparkJobInvalid}`).

I say that it *might* be compatible with `spark-jobserver` because I was not able to run it.  A 59 megabyte jar file is produced, but the `spark-jobserver` accepts a maximum of 30 megabytes by default.  I created this patch https://gist.github.com/jamesmcclain/86a131db281c6c490dbc#file-docker-jobserver-diff against https://github.com/azavea/docker-spark-jobserver to try to get around that, but unfortunately it does not work: when I try to do a `make test` docker throws something called "Error 7" and stops.  Oddly, the same behavior occurs when I try to use a similar technique to go below 30 megabytes, so clearly there is something wrong with this method.  It is probably better to just create a patched branch and use the git command in the Docker file to pull form that.  It also might make sense to pursue  some of the idea for reducing jar file size given here: https://github.com/ooyala/spark-jobserver/issues/5 .

hectcastro:
First of all, thanks for trying to make an example work with Spark Job Server (from now on known as SJS). I got your example to work with SJS using the steps below.

[...]

PR #574, opened 2015-07-14
Feature/lol/add scala vm

lliss:
Create a geotrellis vm and ansible role.

Connects #7 

Builds off of https://github.com/WikiWatershed/model-my-watershed/pull/318

To test:

 * vagrant up geotrellis
 * Use firefox poster plugin or wget as per instructions in https://github.com/WikiWatershed/model-my-watershed/pull/318 to post a geojson shape to localhost:8081
 * You should see results.

Only the commits after ea43a2f are relevant.

Note that this will take a while to build.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -13,6 +13,7 @@ ANSIBLE_GROUPS = {
   "app-servers" => [ "app" ],
   "services" => [ "services" ],
   "workers" => [ "worker" ],
+  "geoprocessing" => [ "geoprocessing" ],
   "monitoring-servers" => [ "services" ],
   "tile-servers" => [ "tiler" ]
 }
@@ -20,14 +21,14 @@ ANSIBLE_GROUPS = {
 if !ENV["VAGRANT_ENV"].nil? && ENV["VAGRANT_ENV"] == "TEST"
   ANSIBLE_ENV_GROUPS = {
     "test:children" => [
-      "app-servers", "services", "workers", "tile-servers"
+      "app-servers", "services", "workers", "tile-servers", "geoprocessing"
     ]
   }
   VAGRANT_NETWORK_OPTIONS = { auto_correct: true }
 else
   ANSIBLE_ENV_GROUPS = {
     "development:children" => [
-      "app-servers", "services", "monitoring-servers", "workers", "tile-servers"
+      "app-servers", "services", "monitoring-servers", "workers", "tile-servers", "geoprocessing"
     ]
   }
   VAGRANT_NETWORK_OPTIONS = { auto_correct: false }
@@ -116,6 +117,35 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     end
 
     worker.vm.provision "shell", inline: "service celeryd restart >> /dev/null 2>&1", run: "always"
+  end
+
+  config.vm.define "geoprocessing" do |geop|
+    geop.vm.hostname = "geoprocessing"
+    geop.vm.network "private_network", ip: ENV.fetch("MMW_GEOPROCESSING_IP", "33.33.34.48")
+
+    geop.vm.synced_folder ".", "/vagrant", disabled: true
+
+    if Vagrant::Util::Platform.windows? || Vagrant::Util::Platform.cygwin?
+      geop.vm.synced_folder "src/geop", "/opt/geop/", type: "rsync", rsync__exclude: ["node_modules/"]
+    else
+      geop.vm.synced_folder "src/geop", "/opt/geop/"
+    end
+
+    geop.vm.provider "virtualbox" do |v|
+      v.memory = 3096
+    end
+
+    # Geoprocessing endpoint via Nginx
+    geop.vm.network "forwarded_port", {
+      guest: 80,
+      host: 8081
+    }.merge(VAGRANT_NETWORK_OPTIONS)
+
+    geop.vm.provision "ansible" do |ansible|
+      ansible.playbook = "deployment/ansible/geoprocessing.yml"
+      ansible.groups = ANSIBLE_GROUPS.merge(ANSIBLE_ENV_GROUPS)
+      ansible.raw_arguments = ["--timeout=60"]
+    end
   end
 
   config.vm.define "app" do |app|
--- a/deployment/ansible/geoprocessing.yml
+++ b/deployment/ansible/geoprocessing.yml
@@ -1,0 +1,10 @@
+---
+- hosts: geoprocessing
+  sudo: True
+
+  pre_tasks:
+    - name: Update APT cache
+      apt: update_cache=yes cache_valid_time=3600
+
+  roles:
+    - { role: "model-my-watershed.geoprocessing" }
--- a/deployment/ansible/group_vars/all
+++ b/deployment/ansible/group_vars/all
@@ -31,3 +31,6 @@ elasticsearch_cluster_name: "logstash"
 nodejs_npm_version: 2.1.14
 
 java_version: "7u79-*"
+
+spark_version: "1.4.0-bin-hadoop2.6"
+scala_version: "2.10.5"
--- a/deployment/ansible/roles.txt
+++ b/deployment/ansible/roles.txt
@@ -22,3 +22,8 @@ azavea.java,0.1.1
 azavea.build-essential,0.1.0
 azavea.mapnik,0.1.0
 azavea.beaver,0.1.3
+azavea.java,0.2.1
+azavea.git,0.1.0
+azavea.spark,1.0.0
+azavea.scala,0.1.0
+azavea.libgdal-java,0.1.0
--- a/deployment/ansible/roles/model-my-watershed.geoprocessing/defaults/main.yml
+++ b/deployment/ansible/roles/model-my-watershed.geoprocessing/defaults/main.yml
@@ -1,0 +1,11 @@
+---
+geop_home: /opt/geop
+data_dir: "{{ geop_home }}/processing/src/test/data/"
+deploy_branch: "master"
+geoprocessing_log: /var/log/mmw-geoprocessing.log
+geoprocessing_log_rotate_count: 5
+geoprocessing_log_rotate_interval: daily
+
+
+app_config:
+  MMW_DATA: "{{ data_dir }}"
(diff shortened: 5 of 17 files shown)


mmcfarland:
General comment, I think it'd be more useful/less confusing to use the noun "Geoprocessing" or "geop" instead of "Geotrellis" to refer to this machine (and parts), especially as GT isn't even directly exposed, and it'll grow to do more general purpose geoprocessing using GT as a component.

lliss:
Noted. I'll go through and rename some things after you're done testing.

mmcfarland:
Aside from the naming comment, this PR is successfully building a geop VM and serving #318 correctly.  I'd appreciate an evaluation from @hectcastro, with an eye towards how we start to think about providing similar services in AWS.

deployment/ansible/roles/model-my-watershed.geotrellis/meta/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+dependencies:
+  - { role: "model-my-watershed.base" }
+  - { role: "azavea.git" }
+  - { role: "azavea.spark", spark_version: "1.3.1-bin-hadoop2.6" }
```
hectcastro: `spark_version` here should probably be moved up into `group_vars/all`. Also, I think we should be up to `1.4.0-bin-hadoop2.6`, unless we were told to pin to 1.3.1.

deployment/ansible/roles/model-my-watershed.geotrellis/meta/main.yml
```diff
@@ -0,0 +1,9 @@
...
+  - { role: "azavea.scala", scala_version: "2.10.5" }
```
hectcastro: Same with moving `scala_version` to `group_vars/all`.

deployment/ansible/roles/model-my-watershed.geotrellis/tasks/compile.yml
```diff
@@ -0,0 +1,6 @@
+---
+- name: Compile the scala code
+  shell: "sbt 'project services' compile assembly"
```
hectcastro: I usually try to avoid using `shell` unless some type of shell expansion is required. Was there a reason not to use `command`?

deployment/ansible/roles/model-my-watershed.geotrellis/templates/nginx-mmw-geotrellis.conf.j2
```diff
@@ -0,0 +1,23 @@
...
+
+  {% if ['packer'] | is_in(group_names) -%}
+  location /version.txt {
+    alias /srv/version.txt;
+  }
+  {% endif %}
+
+  location / {
+    {% if ['packer'] | is_in(group_names) -%}
+    expires max;
```
hectcastro: I'm not sure we'll want any HTTP caching on these requests. Are duplicate jobs being cached further up in the stack somewhere?

deployment/ansible/roles/model-my-watershed.geotrellis/tasks/compile.yml
lliss: No, just forgot about command. Fixed in 89a5306013238db270075965a62c3a5f0253fdb8

deployment/ansible/roles/model-my-watershed.geotrellis/meta/main.yml
lliss: Fixed in b08514d0301091ce3ee63bf301a73e3b2f767221

src/geop/services/src/main/scala/com/azavea/mmw/GeopServiceActor.scala
```diff
@@ -0,0 +1,69 @@
...
+  def receive = runRoute(myRoute)
+}
+
+trait GeopService extends HttpService {
+  import scala.util._
+
+  import MediaTypes._
+  import UnattractiveButTemporary._
+
+  val myRoute =
```
hectcastro: Is this setup to be a shortcut or the ultimate mechanism to submit jobs to GeoTrellis? It seems like we might benefit from using [`spark-jobserver`](https://github.com/spark-jobserver/spark-jobserver) here instead.

deployment/ansible/roles/model-my-watershed.geotrellis/templates/nginx-mmw-geotrellis.conf.j2
lliss: There is caching but those results are just being captured and stored in Django. I'll clean this up.

src/geop/services/src/main/scala/com/azavea/mmw/GeopServiceActor.scala
lliss: That might be a better question for @jamesmcclain. I didn't set up the scala rest endpoints. I'm open to any alternative setups.

jamesmcclain: I may not fully understand what `spark-jobserver` does, but it appears to be a mechanism for submitting spark jobs via a REST interface.  This is distinct (I think) from our application which is a long-lived spark job that happens to implement a REST interface.

hectcastro: The `spark-jobserver` route seems like it could be more maintainable in the long run (worth mentioning that I haven't spent much time with it yet). Mainly because it already has a standard interface, but can also handle job submission a few different ways:

- https://github.com/spark-jobserver/spark-jobserver#package-jar---send-to-server
- https://github.com/spark-jobserver/spark-jobserver#ad-hoc-mode---single-unrelated-jobs-transient-context
- https://github.com/spark-jobserver/spark-jobserver#persistent-context-mode---faster--required-for-related-jobs

The persistent context mode seems very similar to what we're trying to do here, but more flexible because another context could be added for different job types, or a newer version of the same job type. I also like the idea of deploying something like this vs. a long-lived `spark-submit` via Upstart.

I don't think this is something that needs to be solved in this PR, but wanted to bring it up to get an idea of where we plan on going. Being closer to the code than I am, does the persistent context mode seem like what we're trying to recreate with the Spray web service endpoints?

lliss:
I think I've addressed the concerns and have tested it locally again. Everything is still working. Will rebase and merge after #318 is in.

src/geop/services/src/main/scala/com/azavea/mmw/GeopServiceActor.scala
jamesmcclain: As I understand it, this looks like a mechanism for uploading jar files and having them execute on spark?  While not incompatible, this seems to be unrelated to what we are doing.  In other words, I see  no problem with submitting a module to spark using this mechanism which then sits and waits for connects, but I do not see a direct connection or synergy.  This is just based on my current understanding though ~ I think the definitive opinion would come from the Geotrellis team.

echeipesh: `spark-jobserver` is very attractive, but there are a couple of things to keep in mind with it.

- Some requests do no to use spark-context at all and will instead use `tileReader` to work on one tile at a time.
- Currently requests are expected to to terminate in request-response cycle, so job monitoring is mute.
- We have not use it yet and its not  clear what the limitations for job parameters are.

I think one we have a situation where the request can not be complete in request-response cycle it is a really good idea to start playing the the `spark-jobserver`. Particular interesting is if we can use the named/cached RDDs as a caching layer for calculations that cover a large area. This would probably require having a way to fetch those tiles in some way other than TMS service and placing them on a Leaflet.

hectcastro:
Confirmed that the workflow works:

**Request:**

```
POST / HTTP/1.1
Host: localhost:8081
Connection: close
User-Agent: Paw/2.2.2 (Macintosh; OS X/10.10.4) GCDHTTPRequest
Content-Length: 466

{
  "type": "Feature",
  "geometry": {
    "type": "Polygon",
    "coordinates": [
      [
        [
          -75.72533952081164,
          39.76
        ],
        [
          -75.72533952081164,
          39.8024743318554
        ],
        [
          -75.58521226772196,
          39.77
        ],
        [
          -75.58521226772196,
          39.7342308109474
        ],
        [
          -75.72533952081164,
          39.76
        ]
      ]
    ]
  }
}
```

**Response**:

```
HTTP/1.1 200 OK
Server: nginx/1.8.0
Date: Wed, 15 Jul 2015 17:00:41 GMT
Content-Type: application/json; charset=UTF-8
Content-Length: 1113
Connection: close
Vary: Accept-Encoding

[[[21, 5], 82599], [[81, 6], 909], [[81, 5], 10024], [[23, 2], 621], [[23, 7], 491], [[11, 6], 25], [[22, 5], 44285], [[52, 2], 602], [[42, 2], 268], [[11, 5], 66], [[24, 3], 80], [[21, 2], 5769], [[90, 7], 93], [[82, 2], 113], [[23, 5], 10390], [[43, 6], 1170], [[43, 0], 194], [[90, 3], 98], [[81, 2], 173], [[21, 3], 404], [[52, 5], 7259], [[82, 5], 740], [[24, 0], 5026], [[52, 7], 421], [[42, 7], 219], [[24, 6], 371], [[41, 7], 3257], [[31, 2], 48], [[52, 0], 165], [[21, 7], 5037], [[23, 6], 1391], [[22, 7], 1258], [[82, 0], 4], [[31, 5], 128], [[81, 7], 451], [[22, 0], 14319], [[43, 5], 4744], [[42, 6], 182], [[41, 5], 45882], [[81, 0], 164], [[22, 2], 1509], [[43, 7], 284], [[41, 6], 11114], [[24, 7], 54], [[23, 3], 279], [[43, 2], 821], [[42, 5], 5244], [[71, 7], 46], [[90, 5], 2752], [[22, 6], 3052], [[71, 5], 194], [[21, 6], 7570], [[41, 3], 228], [[42, 0], 197], [[90, 6], 2916], [[90, 0], 172], [[22, 3], 325], [[82, 6], 24], [[24, 5], 1691], [[52, 3], 19], [[11, 0], 1173], [[41, 0], 1223], [[41, 2], 7334], [[21, 0], 10179], [[90, 2], 5618], [[23, 0], 5704], [[24, 2], 173], [[52, 6], 793]]
```

:+1:

jamesmcclain:
Cool!  Let me know what changes if any you would like me to make to work with `spark-jobserver`.

hectcastro:
I think we need to allocate some time in an upcoming sprint for https://github.com/WikiWatershed/model-my-watershed/issues/6 to address `spark-jobserver` stuff.

mmcfarland:
+1 Same sentiment as #318, I think this is working and we'll keep building on it.

PR #575, opened 2015-07-14
Add stream slider

lewfish:
### Loading data
 * Get the three zip files from `smb://fileshare/users/lfishgold/mmw_stream_data` These are different than the ones in the original email because I transformed the data to use a different SRS.
 * Unzip and then transform the `.shp` file in each directory to a `.sql` file using `shp2pgsql -g geom DelDem4net20r.shp > DelDem4net20r.sql`, for example.
 * Log into the database using `psql -h localhost -p 5432 -U mmw` on your host machine. 
 * Load the data for each `.sql` file using `\i DelDem4net20r.sql`, for example.
 * Each file takes a long time load, maybe 30 minutes.
 * Reload the tiler VM.
 
### Testing
Initially, the slider should be set to 'Off.' After setting the slider to low, medium, and high resolution stream tiles, you should see lots of blue lines indicating possible streams. The tiler runs very slowly on my machine, and after a while, seems to require a reboot. I increased the number of CPUs to 2 on the services VM, but it didn't seem to help much.

screen shot 2015-07-14 at 10 34 43 am 2

It would be nice if the line width was a function of the zoom level, but I didn't see any obvious way of doing so.

Connects #3

### DIFF
--- a/src/mmw/apps/home/views.py
+++ b/src/mmw/apps/home/views.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 from __future__ import division
 
 import json
+from urlparse import urljoin
 
 from django.http import Http404
 from django.shortcuts import render_to_response, get_object_or_404
@@ -13,10 +14,27 @@ from django.conf import settings
 from apps.modeling.models import Project
 
 
+def get_stream_layers():
+    tiler_prefix = '//'
+    tiler_host = settings.TILER_HOST
+    tiler_postfix = '/{z}/{x}/{y}'
+    tiler_base = '%s%s' % (tiler_prefix, tiler_host)
+
+    stream_layers = []
+    for layer in settings.STREAM_LAYERS:
+        stream_layers.append({
+            'display': layer['display'],
+            'endpoint': urljoin(tiler_base, layer['code'] + tiler_postfix)
+        })
+
+    return stream_layers
+
+
 def get_client_settings():
     client_settings = {
         'client_settings': json.dumps({
-            'base_layers': settings.BASE_LAYERS
+            'base_layers': settings.BASE_LAYERS,
+            'stream_layers': get_stream_layers()
         })
     }
     return client_settings
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -13,7 +13,9 @@ var $ = require('jquery'),
     selectTypeTmpl = require('./templates/selectType.html'),
     drawAreaTmpl = require('./templates/drawArea.html'),
     resetDrawTmpl = require('./templates/reset.html'),
-    placeMarkerTmpl = require('./templates/placeMarker.html');
+    streamSliderTmpl = require('./templates/streamSlider.html'),
+    placeMarkerTmpl = require('./templates/placeMarker.html'),
+    settings = require('../core/settings');
 
 // Responsible for loading and displaying tools for selecting and drawing
 // shapes on the map.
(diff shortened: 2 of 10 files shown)


jamesmcclain:
It looks as though jshint has found a few things:
```
js/src/draw/views.js: line 230, col 19, Expected '===' and instead saw '=='.
js/src/draw/views.js: line 254, col 2, Unnecessary semicolon.
js/src/draw/views.js: line 261, col 2, Unnecessary semicolon.
```

jamesmcclain:
> It would be nice if the line width was a function of the zoom level, but I didn't see any obvious way of 
> doing so.

I do not know a nice way to do this either.  It might be possible to do this crudely using multiple styles.

Other comments:
   * As you observed above, things do seem to get a little bit slow, especially when enabling the streams while zoomed out, but this is not your fault.
   * The code looks good (to me) and it works as advertised.  Other than the issues found by the linter, +1.

mmcfarland:
CartoCSS provides for a way to style based on zoom level, see http://gis.stackexchange.com/a/53133 for an example.

src/mmw/js/src/draw/views.js
```diff
@@ -188,6 +200,66 @@ var ResetDrawView = Marionette.ItemView.extend({
...
+        // Preview slider value while dragging.
+        this.ui.displayValue.text(this.getSliderDisplay());
+    },
+
+    onSliderChanged: function() {
+        var ind = this.getSliderIndex();
+        clearStreamLayer(this.model);
+        if (ind > 0) {
+            var streamLayer = this.streamLayers[ind-1];
+            changeStreamLayer(streamLayer.endpoint, streamLayer.table_name, this.model);
```
lliss: You shouldn't be using the table name anywhere in client side code. We designed the base_layers to prevent that data from ever being something an end user could see, use, alter etc.

src/mmw/apps/home/views.py
```diff
@@ -13,10 +15,24 @@
...
+    tiler_prefix = '//'
+    tiler_host = settings.TILER_HOST
+    tiler_postfix = '/{z}/{x}/{y}'
+    tiler_base = '%s%s' % (tiler_prefix, tiler_host)
+
+    stream_layers = deepcopy(settings.STREAM_LAYERS)
+    for layer in stream_layers:
+        layer['endpoint'] = urljoin(tiler_base, layer['code'] + tiler_postfix)
+
+    return stream_layers
```
lliss: Avoid exposing the table name. See `api/modeling/boundary-layers/` for the data used to populate the other tiler data.

lliss:
It would be nice if you made the loading of the data a script or part of the `reload-dev-data` script. Seems like something we'll need to do often when developing.

lewfish:
I made line width a function of zoom level and removed the table names from getting to the client. The tiler is running unusably slowly for me and needs to be rebooted every few minutes (even after doubling RAM and CPUs). Should I make an issue to fix the slowness? Re: making a script to load data, I think @mmcfarland is doing that as part of his HUC PR.

lewfish:
I was able to speed up the tiler by adding a spatial index as follows:
`CREATE INDEX spatial_index20r ON deldem4net20r USING GIST ( geom );`
`CREATE INDEX spatial_index50r ON deldem4net50r USING GIST ( geom );`
`CREATE INDEX spatial_index100r ON deldem4net100r USING GIST ( geom );`

However, the high-res tiles at a high zoom level still load very slowly. We might want to look into using pre-computed tiles.

Issue #451, opened 2015-06-17
Queue multiple POST requests


rajadain:
This is proposed as a generalized extension to #448, more for the sake of record than urgent need of implementation.

There are use cases, for example when the user is on a flaky internet connection, when the user interface is faster than the network. In such cases, it is possible for the user to make multiple modifications to a project or scenario, while the individual POST requests are still in transit. Instead of firing them all at once and hoping they sort themselves out, we should queue ones with explicit relationships.

For example, if a user adds a new scenario and quickly makes a change to it, two POST requests are fired instead of a POST and a PUT, resulting in an error response of "The fields name, project must make a unique set." and an unpredictable outcome of which version was saved: the original or the one with changes.

Consider making a generalized approach that queues multiple POSTs, and converts them into PUTs after the first is returned successfully, or tries the next one if it fails.

rajadain:
This may end up being a big card that needs more thought than is provided in the initial description above.

kdeloach:
I would be in favor of...
  * Less event-driven behavior (models updating other models is a recipe for disaster)
  * Locking UI/async actions to prevent creating tons of requests (which can lead to race conditions)
  * Possibly some kind of task that runs in the background and saves data at intervals (instead of as a response to an event)

I mocked up a proof on concept for how we could wire up UI event locking http://jsfiddle.net/kdeloach/qz1d1cw9/ but I would like to put more thought into a better solution.

lliss:
See https://github.com/WikiWatershed/model-my-watershed/pull/577 for another place where this was becoming a problem. With the fix described in this PR in place we should be able to revert #577 and clean up the code here: 
https://github.com/WikiWatershed/model-my-watershed/blob/4f4d321792d447d0241ac3d3a7890fbce233012d/src/mmw/js/src/modeling/models.js#L151
and 
https://github.com/WikiWatershed/model-my-watershed/blob/4f4d321792d447d0241ac3d3a7890fbce233012d/src/mmw/js/src/modeling/models.js#L303

PR #577, opened 2015-07-14
Prevent double saving scenarios.

lliss:
We were seeing errors due to events firing very quickly which caused our
scenario to try to save itself twice in rapid succession. This resulted in an
API error. We'd seen similar behavior before with projects. This adds a simple
flag and locking mechanism to prevent the double save.

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -298,6 +298,12 @@ var ScenarioModel = Backbone.Model.extend({
             App.currProject.saveProjectAndScenarios();
             return;
         }
+        if (this.isNew() && this.saveCalled) {
+            return;
+        } else if (this.isNew() && !this.saveCalled) {
+            // Makeshift locking mechanism to prevent double saves.
+            this.saveCalled = true;
+        }
         this.save().fail(function() {
             console.log('Failed to save scenario');
         });


mmcfarland:
Is this limited to just new scenarios?

lliss:
Could happen anywhere but it's really tied up with the autosave functionality. I saw it happening here while testing something else. I had previously fixed this on projects and this is a similar approach to the fix I made there.

src/mmw/js/src/modeling/models.js
```diff
@@ -298,6 +298,12 @@ var ScenarioModel = Backbone.Model.extend({
             App.currProject.saveProjectAndScenarios();
             return;
         }
+        if (this.isNew() && this.saveCalled) {
+            return;
+        } else if (this.isNew() && !this.saveCalled) {
+            // Makeshift locking mechanism to prevent double saves.
+            this.saveCalled = true;
```
mmcfarland: Since this flag is playing an important role, I think we should define it explicitly on the class like we do for `ProjectModel`, not just implicitly when we first set it.

mmcfarland:
Since there is an underlying problem here, I'm becoming more interested at following up on https://github.com/WikiWatershed/model-my-watershed/issues/451.

lliss:
Reasonable... this seemed like an easy temporary fix for the moment but I'd support a card in the next sprint to fix this in a more robust way.

mmcfarland:
+1

Issue #562, opened 2015-07-09
Auto Login to ITSI when coming from an activity


mmcfarland:
Utilizing the same flag as was created in #560, an attempt to auto-login to ITSI should be made when receiving the request.  Under normal circumstances the user is coming from the ITSI portal, meaning they should be logged in already, resulting in a "single sign on" experience.  We shouldn't need to prompt them again.  If the auto-login fails, it should result in an `401`

Issue #561, opened 2015-07-09
Save project immediately when starting from blank project


mmcfarland:
#### Starting from a blank project

A second workflow for a new ITSI assignment would be for the teacher to not specify an existing project (#560) to clone from, but instead have the student select their own area of interest and begin the process from the beginning.  The URL should contain the same modification as made in #560, but the result should be that a new project is saved immediately.  This is essentially similar to coming to the site by default, but since we'll have to report to the portal a "saved state" - a URL specifying the project in the student's account - we'll want the URL sooner than the normal course.  And because the state is saved, the student returning to the site will not generate multiple new, blank projects in their account which is the reason for waiting to save until modifications are made under normal circumstances.

Potential URL structure:
http://mmw.azavea.com/?initialize_itsi=true

Assume:
* An ITSI user is currently logged in.

Issue #560, opened 2015-07-09
Clone project via URL flag


mmcfarland:
#### Start from an existing project

ITSI portal Activities will be initiating student assignments via a GET request to a per-assignment URL.  Teachers can specify a base project, which may include an AoI, scenarios, model package and any number of modifications.  By creating a project and copying the URL, they can embed that base project in an Activity.  When a student visits the Activity page, and the request is sent to our servers and could take the form of:

http://mmw.azavea.com/project/123/scenario/123?initialize_itsi=true

This would cause the original project to be copied to the currently logged in user's account and then redirect the browser to the new project, which the correspondingly mapped scenario tab (if present) active.

I'm not sure what form the `initialize_itsi=true` flag should take, but I do think it's important that we require the teacher to modify the URL from the "share" link as little as possible before submitting it to an activity.

For this issue, assume the following, whose nuances will be handled by separate issues:
* Original project is public
* Current user is logged in via ITSI

rajadain:
We can consider adding an "Get ITSI Embed Link" option to the project dropdown, only visible to ITSI users, which will, just like the "Share" option, prompt the user to first make the project public (or even do it under the hood without asking permission?). This will enforce both assumptions.

We can set a flag in the `user` object to indicate if they are an ITSI user or not, and then use that flag to customize the Project menu.

Having the "make public" a part of the sharing will ensure that the project is public before we get the link.

Having a popup already populated with the correct URL is easier than having the teacher edit a regular Share URL.

The existing Share menu should still be available.

mmcfarland:
@rajadain Good idea, the teacher shouldn't have to modify anything, we should have the app provide it like you suggest.

Issue #564, opened 2015-07-09
Embed MMW Interactive into ITSI Activity


mmcfarland:
Tying all the pieces of #560, #561, #562 & #563, create a new Interactive using the MMW URL structures developed in the previous issues.  This issue is for identifying and fixing problems and assumptions with the full site.

https://learn.staging.concord.org/

rajadain:
Scott mentioned that certain browsers like Safari may have problems setting third party cookies inside iframes. He also mentioned that the oAuth redirection dance can be fragile inside iframes. These are a few issues to watch out for when doing the final embed.

rajadain:
When I try to create a new activity / interactive as a Teacher user, I don't see a way to embed a custom app. I can only browse and search in the existing options:

![image](https://cloud.githubusercontent.com/assets/1430060/8913041/8546af14-3464-11e5-8c45-6789f65bfef3.png)

I think we may have to talk to Scott about getting MMW into the library.

rajadain:
With recent additions to #613 and the lack of URL manipulation as noted in https://github.com/WikiWatershed/model-my-watershed/issues/563#issuecomment-126824774, while we have a (somewhat) working implementation of #561 within an activity, we don't yet of #560 whose interface is URL-only. We will need to come up with a UI mechanism to enable cloning.

PR #578, opened 2015-07-14
Auto Login to ITSI when coming from an activity

rajadain:
## Overview

When a user initially accesses a URL from within an activity in the ITSI portal, we can have the portal append a flag to the URL indicating that they are coming from the portal. Currently, we assume this flag to be a ~~`?initialize_itsi=true`~~ `?itsi_embed=true` query parameter. This is configurable in the ITSI ~~`url_flag`~~ `embed_flag` setting in `base.py`.

We intercept all requests to check if this query parameter exists. If it does not, we do nothing. ~~If it does, we see if the user is already logged in, and if so, do they have an ITSI account. If they do, we do nothing.~~ Otherwise, we attempt to log them in automatically via the portal. Since they are already logged in, the redirection will be transparent and they will be returned to the initial request path after the oAuth dance is done. If there is any error during this, they will instead be redirected to `/error/itsi`.

In case they don't yet have an MMW account corresponding to their ITSI account, they will be prompted to sign up, after which they will be returned to the initial request path.

## Testing Instructions

 1. Go to the home page of the app. Log out.
 2. Go to ~~`/?initialize_itsi=true`~~ [`/?itsi_embed=true`](http://localhost:8000/?itsi_embed=true). An attempt should be made to automatically log you in to ITSI. If it is successful, you should be returned to ~~`/?initialize_itsi=true`~~ [`/`](http://localhost:8000/).
 3. Go to a public project, for example `/project/1/scenario/2`. Log out of your ITSI account. Go to the same project, but this time add the parameter: ~~`/project/1/scenario/2?initialize_itsi=true`~~ `/project/1/scenario/2?itsi_embed=true`. You should be automatically logged in and returned to the original URL.
 4. Try doing that with a private project. You will first be logged in to ITSI, then you should get a `404`.
 5. Delete your ITSI account from the database:

    ```sql
    DELETE FROM user_itsiuser WHERE user_id IN (
        SELECT id FROM auth_user WHERE username = 'tstudent2'
    );
    DELETE FROM auth_user WHERE username = 'tstudent2';
    ```

  Then try accessing the public project. You should be prompted to sign up as a new user, after which you should be returned to the original project.
 6. When logged in as one ITSI user in the system, `tstudent2`, log in to MMW. Then go back to ITSI and log out, and log in as another ITSI user `mmwterence`. Then come back to MMW and add the [`/?itsi_embed=true`](http://localhost:8000/?itsi_embed=true) flag. You should be logged in as the current ITSI user, namely `mmwterence`.

## Questions

 * ~~Currently, after logging in with ITSI, we **always** set the `initialize_itsi` flag. Should we always do this, or only when they are coming in from the ITSI portal?~~
 * ~~Currently, we only check if they are logged in with an ITSI account. Should we try and check to see if they are logged in as the same user as they are in ITSI? For example, a teacher might have a dummy student account to check what things look like from that perspective, and may forget to log out. They may be logged in with the dummy student account in MMW even if they are logged in as themselves in ITSI.  
  However, that would require some refactoring, since currently we use the fact that the logged in user is an ITSI user to escape the loop of always checking against the portal, since the `intialize_itsi` flag is set even when returning after logging in, to facilitate behavior expected in #560 and #561.~~

Connects #562

### DIFF
--- a/src/mmw/apps/user/middleware.py
+++ b/src/mmw/apps/user/middleware.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+
+from django.conf import settings
+from django.core.urlresolvers import reverse
+from django.shortcuts import redirect
+
+from apps.user.views import itsi_login
+
+EMBED_FLAG = settings.ITSI['embed_flag']
+LOGIN_URL = reverse(itsi_login)
+
+
+class ItsiAuthenticationMiddleware(object):
+    """
+    Middleware for automatically logging in ITSI users
+    and setting relevant flags.
+    """
+
+    def process_request(self, request):
+        """
+        Check if ITSI EMBED FLAG is set, and if so attempt to log
+        the user in with their ITSI credentials
+        """
+
+        # If flag is not set then return None so Django can proceed
+        # with the request as-is
+        if request.GET.get(EMBED_FLAG, 'false') != 'true':
+            return None
+
+        # Flag is set. Assume user is not logged in. Set session flag and
+        # Redirect them to ITSI LOGIN URL
+        request.session[EMBED_FLAG] = True
+        return redirect(LOGIN_URL)
--- a/src/mmw/apps/user/middleware.py
+++ b/src/mmw/apps/user/middleware.py
@@ -19,7 +19,8 @@ class ItsiAuthenticationMiddleware(object):
     def process_request(self, request):
         """
         Check if ITSI EMBED FLAG is set, and if so attempt to log
-        the user in with their ITSI credentials
+        the user in with their ITSI credentials and redirect them
+        to current page.
         """
 
         # If flag is not set then return None so Django can proceed
@@ -28,6 +29,6 @@ class ItsiAuthenticationMiddleware(object):
             return None
 
         # Flag is set. Assume user is not logged in. Set session flag and
-        # Redirect them to ITSI LOGIN URL
+        # Redirect them to ITSI LOGIN URL, with the current URL to return to
         request.session[EMBED_FLAG] = True
-        return redirect(LOGIN_URL)
+        return redirect('{0}?next={1}'.format(LOGIN_URL, request.path))
--- a/src/mmw/apps/user/views.py
+++ b/src/mmw/apps/user/views.py
@@ -84,7 +84,10 @@ itsi = ItsiService()
 
 
 def itsi_login(request):
-    redirect_uri = request.build_absolute_uri(reverse('itsi_auth'))
+    redirect_uri = '{0}?next={1}'.format(
+        request.build_absolute_uri(reverse(itsi_auth)),
+        request.GET.get('next', '/')
+    )
     params = {'redirect_uri': redirect_uri}
     auth_url = itsi.get_authorize_url(**params)
 
@@ -108,12 +111,14 @@ def itsi_auth(request):
     user = authenticate(itsi_id=itsi_user['id'])
     if user is not None and user.is_active:
         auth_login(request, user)
-        return redirect('/')
+        return redirect(request.GET.get('next', '/'))
     else:
-        # User did not authenticate. Save their ITSI ID and send to /register
+        # User did not authenticate. Save their ITSI ID and send to /sign-up
         request.session['itsi_id'] = itsi_user['id']
         return redirect(
-            '/sign-up/itsi/{username}/{first_name}/{last_name}'.format(
+            '/sign-up/itsi/{username}/{first_name}/{last_name}?next={0}'
+            .format(
+                request.GET.get('next', '/'),
                 **itsi_user['extra']
             )
         )
--- a/src/mmw/js/src/user/controllers.js
+++ b/src/mmw/js/src/user/controllers.js
@@ -1,6 +1,7 @@
 "use strict";
 
 var App = require('../app'),
+    utils = require('../core/utils'),
     models = require('./models'),
     views = require('./views');
 
@@ -12,13 +13,14 @@ var SignUpController = {
         }).render();
     },
 
-    itsiSignUp: function(username, first_name, last_name) {
+    itsiSignUp: function(username, first_name, last_name, queryParams) {
         new views.ItsiSignUpModalView({
             app: App,
             model: new models.ItsiSignUpFormModel({
                 username: username,
                 first_name: first_name,
-                last_name: last_name
+                last_name: last_name,
+                next: queryParams ? utils.parseQueryString(queryParams).next : '/'
             })
         }).render();
     }
--- a/src/mmw/js/src/user/views.js
+++ b/src/mmw/js/src/user/views.js
@@ -134,13 +134,15 @@ var LoginModalView = ModalBaseView.extend({
         password: '#password',
         signUp: '.sign-up',
         resend: '.resend',
-        forgot: '.forgot'
+        forgot: '.forgot',
+        itsiLogin: '.itsi-login'
     }, ModalBaseView.prototype.ui),
 
     events: _.defaults({
         'click @ui.signUp': 'signUp',
         'click @ui.resend': 'resend',
-        'click @ui.forgot': 'forgot'
+        'click @ui.forgot': 'forgot',
+        'click @ui.itsiLogin': 'itsiLogin'
     }, ModalBaseView.prototype.events),
 
     onModalShown: function() {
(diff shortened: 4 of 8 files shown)


src/mmw/mmw/settings/base.py
```diff
@@ -244,6 +244,7 @@ def get_env_setting(setting):
     'django.contrib.sessions.middleware.SessionMiddleware',
     'django.middleware.csrf.CsrfViewMiddleware',
     'django.contrib.auth.middleware.AuthenticationMiddleware',
+    'apps.user.middleware.ItsiAuthenticationMiddleware',
```
rajadain: Is this the right place to put the new middleware? I also see that we have `mmw.middleware.BypassMiddleware`, and I could have added my `ItsiAuthenticationMiddleware` class there, but felt like this was more cohesive since all the other ITSI stuff is contained within `apps.user`.

src/mmw/mmw/settings/base.py
```diff
@@ -358,6 +359,9 @@ def get_env_setting(setting):
     'authorize_url': 'auth/concord_id/authorize',
     'access_token_url': 'auth/concord_id/access_token',
     'user_json_url': 'auth/concord_id/user.json',
+
```
rajadain: This blank line added to differentiate the ITSI Portal specific settings, and MMW specific settings (which follow).

src/mmw/apps/user/views.py
```diff
@@ -81,10 +82,14 @@ def logout(request):
         return redirect('/')
 
 itsi = ItsiService()
+ITSI_URL_FLAG = settings.ITSI['url_flag']
 
 
 def itsi_login(request):
-    redirect_uri = request.build_absolute_uri(reverse('itsi_auth'))
+    redirect_uri = '{0}?next={1}'.format(
```
rajadain: Do we prefer `.format` or `%` for string interpolation?

src/mmw/apps/user/views.py
```diff
@@ -108,12 +113,18 @@ def itsi_auth(request):
     user = authenticate(itsi_id=itsi_user['id'])
     if user is not None and user.is_active:
         auth_login(request, user)
-        return redirect('/')
+        redirect_uri = '{0}?{1}=true'.format(
```
rajadain: Is it okay to redirect from strings or should we use something more explicitly defined like `request.build_absolute_uri(reverse('itsi_login'))` or something?

src/mmw/apps/user/middleware.py
```diff
@@ -0,0 +1,39 @@
...
+        # Check if user is already logged in, and if they are an
+        # ITSI user. If so, return None so Django can proceed
+        user = request.user
+        if (user.is_authenticated() and
+                ItsiUser.objects.filter(user_id=user.id)):
+            return None
+
+        # Flag is set and the user is not already logged in.
+        # Redirect them to ITSI LOGIN URL, with the current URL to return to
+        return redirect('{0}?next={1}'.format(LOGIN_URL, request.path))
```
rajadain: Is it okay to redirect from strings or should we use something more explicitly defined like `request.build_absolute_uri(reverse(...))` or something?

src/mmw/js/src/user/views.js
```diff
@@ -355,7 +355,7 @@ var ItsiSignUpModalView = ModalBaseView.extend({
 
     // Override to provide custom success handler
     primaryAction: function() {
-        var formData = this.$el.find('form').serialize();
+        var formData = this.model.toJSON();
```
rajadain: We do this since we set the `next` variable on the model, not in the form.

mmcfarland:
Functionality is working really well, and I like the approach.  Going to spend some time looking at the code, in the meantime:

* No, I think we should only be setting the itsi flag on the QS when it was provided in the initial request.  Seems like it'll be good for the app to know whether it is in an Interactive or not, and that flag seems like the clearest indication.
* I think if the user is logged in, and we've received the itsi flag, we should essentially re-login to ensure that we reflect the current session in ITSI.  Unsure based on your comment if that would require much refactoring.

mmcfarland:
Also, should we think about adding a session variable (or similar) to indicate coming from an ITSI session, rather than preserve the querystring after parsing and doing the login dance?

rajadain:
Ah, session variables! I'd forgotten about those. You're right, we should set that and not have it in the URL all the time. I had thought about using cookies but remembered that they sometimes behave strangely when in iframes.

The issue with resetting the flag in the URL is for the use case when the user lands on a project page, then we do the oAuth dance, and then redirect them back to the same page, which we must now clone into their account. Since cloning so far expects to use the same flag, it must be present in the URL. However, when we see the flag we will try to log them in again (to make sure they're using the current ITSI account), and thus be stuck in a loop.

The way out of this loop seems to be to make a separate flag / endpoint for project cloning. I'll meditate on that and add a solution here and comments on #560.

hectcastro:
@azavea-bot rebuild

rajadain:
I have simplified the PR according to received feedback and updated the main description and testing instructions. This is ready to be looked at again.

src/mmw/js/src/user/views.js
```diff
@@ -335,11 +343,6 @@ var ItsiSignUpModalView = ModalBaseView.extend({
         this.ui.username.focus();
     },
 
-    onModalHidden: function() {
```
rajadain: Removed this since we're handling the re-routing in the end of `handleSuccess`

src/mmw/mmw/settings/base.py
lewfish: If a middleware returns a redirect, then subsequent middlewares will not run, according to the Django docs. So it's safer to put it at the end of the list.

src/mmw/apps/user/views.py
```diff
@@ -84,7 +84,10 @@ def logout(request):
 
 
 def itsi_login(request):
-    redirect_uri = request.build_absolute_uri(reverse('itsi_auth'))
+    redirect_uri = '{0}?next={1}'.format(
+        request.build_absolute_uri(reverse(itsi_auth)),
+        request.GET.get('next', '/')
```
lewfish: This expression is repeated several times so you could pull it into a variable.

src/mmw/js/src/user/controllers.js
```diff
@@ -12,13 +13,14 @@ var SignUpController = {
...
-    itsiSignUp: function(username, first_name, last_name) {
+    itsiSignUp: function(username, first_name, last_name, queryParams) {
         new views.ItsiSignUpModalView({
             app: App,
             model: new models.ItsiSignUpFormModel({
                 username: username,
                 first_name: first_name,
-                last_name: last_name
+                last_name: last_name,
+                next: queryParams ? utils.parseQueryString(queryParams).next : null
```
lewfish: I'm not sure, but maybe the null should be a '/'.

lewfish:
+1, tested. Nicely done.

src/mmw/apps/user/views.py
kdeloach: Either is fine, I prefer `.format` slightly more. Those Python guidelines are not authored by anyone on the Civic Apps team and could probably be updated.

rajadain: I need `.format` for auto parsing keys from a dictionary here: https://github.com/WikiWatershed/model-my-watershed/pull/578/files#diff-76f6d7d05a4b988f9b9af9034c53a081R119 so I decided to be consistent and just use it everywhere. Reading up on recent Python guidelines, it seems like there is a gradual movement away from `%`

src/mmw/apps/user/middleware.py
```diff
@@ -0,0 +1,34 @@
...
+        """
+
+        # If flag is not set then return None so Django can proceed
+        # with the request as-is
+        if request.GET.get(EMBED_FLAG, 'false') != 'true':
+            return None
+
+        # Flag is set. Assume user is not logged in. Set session flag and
+        # Redirect them to ITSI LOGIN URL, with the current URL to return to
+        request.session[EMBED_FLAG] = True
```
rajadain: It's not used as part of this PR, but we might need it for #561 or #563 or #564.

src/mmw/js/src/user/controllers.js
rajadain: Hmm at first I was thinking that the back-end always has to redirect somewhere, so we default to `/` if no `next` is specified. The front-end on the other hand has the third option of not redirecting at all, which is what the current implementation does here: https://github.com/WikiWatershed/model-my-watershed/pull/578/files#diff-bbc329caa6aad1d9e2bf745db301390fR381. However, now that I think about it, I think we do always want to redirect away from the sign-up URL, even if nothing is specified, so you're right this should default to `/`. Good catch.

src/mmw/mmw/settings/base.py
rajadain: Agreed.

rajadain:
Moved middleware to end of chain, set client to default `next` to `/` so it always redirects even when `next` is not specified.

Issue #316, opened 2015-06-08
Version Celery tasks


hectcastro:
In cases where we attempt a blue/green deployment, there will be two versions of the Celery code attempting to process jobs. We need a way to version job specifications so that new Celery worker code avoids attempting older jobs while multiple stacks are present.

rajadain:
Since the `worker` VM also has access to the main git repo, we could include an abbreviated SHA as the version number from the `app` in job requests. Celery could then check git inside `worker` to see if that SHA is available on its end. If so, then Celery is running an equal or later codebase than `app`.

But that isn't explicit enough, since different versions of MMW will work with the same version of Celery and vice versa. So do we start some form of semantic versioning for the `worker`?

hectcastro:
Would it be less complicated to use stack color as the separator?

For example, `Blue` stack machines can only send tasks to the `Blue` queue (or add an attribute to the task indicating its color). Only `Blue` workers can process `Blue` tasks.

When we go to deploy, we bring up a new `Green` stack that functions similarly. As we cut over, we wait for tasks for `Blue` to drain as new tasks begin flowing down the `Green` path.

Getting the color from within the code may require interaction with the EC2 API via Boto. You can get the instance ID via the metadata service, but I don't think it gives you any information on tags. From there, you can query the EC2 API with the instance ID for tag details.

This might be something that goes in the production settings file. Non-production can get some consistent default queue name or color attribute. Failures in the process are OK, because those instances will just never get associated with the load balancer (failed health checks).

See also:

- https://boto.readthedocs.org/en/2.0/ref/boto.html#boto.utils.get_instance_metadata
- http://boto.readthedocs.org/en/latest/ref/ec2.html#module-boto.ec2.instance

PR #580, opened 2015-07-16
Feature/lol/celery versions

lliss:
Connects #316

Provides a way to prevent stacks (green/blue) from running jobs provided by an alternately colored stack. Use black as a default value for development and as a reasonable default if something fails to properly provision.


mmcfarland:
Is `Black` used just for environments which are not associated with a stack color?

lliss:
Correct. As a default for dev or errors.

lliss:
@azaveabot rebuild

mmcfarland:
Looks good to me.

mmcfarland:
You'll want to add 
Connects #316

Issue #582, opened 2015-07-16
Stream network layers don't show over certain basemap layers


mmcfarland:
To reproduce: 

* Turn on stream layer
* Switch to ESRI basemap
* Notice the stream layer is not visible

PR #584, opened 2015-07-16
Make stream layer visible after switching base layer

lewfish:
To test:

 * Turn on a stream layer
 * Switch base map 
 * Ensure that stream layer is still showing
 * Also check that Congressional Districts layer works alongside stream layer

Connects #582


jamesmcclain:
Just for future reference, the loading instructions given in https://github.com/WikiWatershed/model-my-watershed/pull/575 are required to test/use.

jamesmcclain:
I was able to see the congressional boundaries and the streams on top of both basemaps.
:+1:

Issue #569, opened 2015-07-13
Add Analyze Endpoint for GeoP REST service


mmcfarland:
The Geoprocessing endpoint will support general analysis (histogram) on a defined list of layers.  Initially it should just be the layers required for the TR-55 model.

PR #589, opened 2015-07-17
 Provide a survey of the Area of Interest.

jamesmcclain:
**Back-end**

Use Geotrellis to compute a survey of the area of interest that is compatible with that expected by the `/analyze` page.

**Front-end**

The `run_analyze` function was receiving an object with nothing in it.  It needs access to the area of interest in order to provide a survey.

Connects #569

**Caveats**
   1. The introduces a new environment variable `MMW_GEOPROCESSING_HOST`.
   1. The test extent  (shown below) is very small.  You will receive null results for any portion of the area of interest that does not intersect with  the test extent.
extent
   1. Some of the tiles do not have a soil classification (their value is "(null)" instead of "A" or "C/D" or the like).  All such tiles have been given a soil value of "?".  We will eventually have to come up with a strategy for dealing with these, because they make up a significant percentage of the tiles.
   1. They very first interaction with Geotrellis after the VMs have been (re)started seems to always fail (timeout?).

**Non-scientific validation**
There could be some issues with the alignment of the data.  For example, the two areas of interest below give 0% water and 11% water, respectively.

screenshot from 2015-07-17 15 10 28

screenshot from 2015-07-17 15 11 26

**To test**
   1. Draw an area of interest that intersects with the test extent.
aoi
   2. You should get back a survey of the area of interest.  Click on the "land" and "soil" tabs to examine it.
land
soil

### DIFF
--- a/deployment/ansible/group_vars/development
+++ b/deployment/ansible/group_vars/development
@@ -18,6 +18,8 @@ graphite_host: "{{ services_ip }}"
 statsite_host: "{{ services_ip }}"
 tiler_host: "{{ lookup('env', 'MMW_TILER_IP') | default('33.33.34.35', true) }}"
 
+geop_host: "{{ lookup('env', 'MMW_GEOPROCESSING_IP') | default('33.33.34.48', true) }}"
+
 celery_log_level: "DEBUG"
 celery_number_of_workers: 2
 celery_processes_per_worker: 1
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -6,115 +6,41 @@ from __future__ import absolute_import
 from celery import shared_task
 import logging
 
-# TODO Remove this when stub task is deleted.
-import time
+from apps.modeling.geoprocessing import geojson_to_survey
 
 from tr55.model import simulate_modifications, simulate_cell_day
 from tr55.tablelookup import lookup_ki
 
 logger = logging.getLogger(__name__)
 
+FE_BE_MAP = {
+    'lir':                 'li_residential',
+    'hir':                 'hi_residential',
+    'commercial':          'commercial',
+    'forest':              'mixed_forest',
+    'turf_grass':          'urban_grass',
+    'pasture':             'pasture',
+    'grassland':           'grassland',
+    'row_crop':            'row_crop',
+    'chaparral':           'chaparral',
+    'tg_prairie':          'tall_grass_prairie',
+    'sg_prairie':          'short_grass_prairie',
+    'desert':              'desert',
+    'rain_garden':         'rain_garden',
+    'veg_infil_basin':     'infiltration_trench',
+    'porous_paving':       'porous_paving',
+    'green_roof':          'green_roof',
+    'no_till_agriculture': 'no_till',
+    'cluster_housing':     'cluster_housing'
+}
+
 
 @shared_task
 def run_analyze(area_of_interest):
-    time.sleep(3)
-
-    results = [
-        {
-            "name": "land",
-            "displayName": "Land",
-            "categories": [
-                {
-                    "type": "Water",
-                    "area": 21,
-                    "coverage": 0.01
-                },
-                {
-                    "type": "Developed: Open",
-                    "area": 5041,
-                    "coverage": .263
-                },
-                {
-                    "type": "Developed: Low",
-                    "area": 5181,
-                    "coverage": .271
-                },
-                {
-                    "type": "Developed: Medium",
-                    "area": 3344,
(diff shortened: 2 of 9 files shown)


lliss:
At least in earlier versions of GT, there was an initialization period after starting. The first request warmed all the pipelines with data that could be used subsequently. As a result the first request was always considerable slower. I suspect you could reproduce this by bounding the GT service too. Since we have a fairly low timeout, I'd expect these to initially fail.

jamesmcclain:
That is consistent with what I am seeing.

src/mmw/apps/modeling/geotrellis.py
```diff
@@ -0,0 +1,153 @@
...
+
+def geotrellis(polygon):
+    """
+    Send a string containing a GeoJSON Polygon, MultiPolygon, or
+    GeometryCollection (containing only Polygons and Multipolygons) to
+    Geotrellis and get back a histogram of NLCD x SOIL pairs present
+    in that area.
+    """
+    host = settings.GEOP['host']
+    conn = httplib.HTTPConnection(host)
```
lliss: Should this be wrapped in a `try` in case something goes wrong?

src/mmw/apps/modeling/geotrellis.py
```diff
@@ -0,0 +1,153 @@
...
+    conn.request("POST", "", polygon, {"Content-type": "application/json"})
+    response = conn.getresponse()
+    data = response.read()
+    conn.close()
+    return data
+
+
+def transmute(data, nucleus, update_rule, after_rule):
```
lliss: Consider renaming this function to something more descriptive.

src/mmw/apps/modeling/geotrellis.py
```diff
@@ -0,0 +1,153 @@
...
+    for [[nlcd, soil], count] in data:
+        total += count
+        update_rule(nlcd, soil, count, retval)
+
+    after_rule(total, retval)
+
+    return retval
+
+
+def census(polygon):
```
lliss: Also could be slightly more descriptive.

src/mmw/apps/modeling/geotrellis.py
```diff
@@ -0,0 +1,153 @@
...
+            update_category(NLCD_MAPPING[nlcd][1], count, landCategories)
+        else:
+            update_category("?", count, landCategories)
+        if soil in SOIL_MAPPING:
+            update_category(SOIL_MAPPING[soil][1], count, soilCategories)
+        else:
+            update_category("?", count, soilCategories)
+
+    def after_rule(count, survey):
+        land = survey[0]["categories"].iteritems()
```
lliss: I think we generally prefer single to double quotes in python.

src/mmw/apps/modeling/tasks.py
```diff
@@ -6,8 +6,7 @@
 from celery import shared_task
 import logging
 
-# TODO Remove this when stub task is deleted.
-import time
+from apps.modeling.geotrellis import survey
```
lliss: When I was working on the VM setup for this, @mmcfarland suggested that I name it more generally "geoprocessing" since it was specifically doing only geotrellis work but rather was acting as a general server for handling all geoprocessing requests. We might consider a similar naming convention here.

lliss:
It would be nice if there was a test for the endpoint. We can assume a response from GT and just use some known returned data. Aside from that and the other comments, this looks really good and works in my test environment.

jamesmcclain:
I added some tests which pass locally but fail on Jenkins.  Since this code will need to be changed to use the S3 catalogues before inclusion, I am leaving this issue (for the time-being).

deployment/ansible/group_vars/development
```diff
@@ -18,6 +18,8 @@ graphite_host: "{{ services_ip }}"
 statsite_host: "{{ services_ip }}"
```
hectcastro: Unrelated to this PR, but I believe there is are two entries for `azavea.java` in `roles.txt`. Only the more recent version should be listed.

jamesmcclain: Is there anything that I can/should do?

hectcastro: Can you apply this change to your PR? https://github.com/hectcastro/model-my-watershed/commit/e00c02d32c81be418408563e2844de4a91e6412e

jamesmcclain: Sure, no problem.

jamesmcclain: I added this to https://github.com/WikiWatershed/model-my-watershed/pull/621 instead, since it is more likely to be included in the near future than this P/R.

PR #573, opened 2015-07-13
Added HTML and some styling for compare screen.

ctaylo37:
Basic HTML framework in place for compare feature.  Needs custom header and functionality wired up.

Issue #537, opened 2015-07-02
Implement views for Compare from prototype


mmcfarland:
Restructure the design elements from the compare prototype into a proper view.

caseypt:
Blocked until prototype is finished

mmcfarland:
See https://github.com/WikiWatershed/model-my-watershed/pull/573

PR #591, opened 2015-07-20
Make initial compare views

lewfish:
This PR includes HTML and CSS for the compare views created by @ctaylo37 which was separately reviewed in https://github.com/WikiWatershed/model-my-watershed/pull/573. It also breaks the HTML up into a bunch of views, and makes the forward and back navigation button operational. 

When going to /compare, you should see 7 scenario columns. Check that the forward and back buttons works and that you cannot scroll beyond the edge of the 0th or 6th column. Try resizing the window. The leftmost column should still be flush against the lefthand side of the window even though the column may have gotten smaller.

Most of the functionality will be implemented in future issues:
 * [Style list of modifications](https://github.com/WikiWatershed/model-my-watershed/issues/592)
 * [Pass in projects and show AOI](https://github.com/WikiWatershed/model-my-watershed/issues/594)
 * [Precipitation slider, charts, and run model](https://github.com/WikiWatershed/model-my-watershed/issues/593)
 * [Unit tests](https://github.com/WikiWatershed/model-my-watershed/issues/595)

Connects #537


lliss:
Can we make a big note in the templates to alert us to the fact that we'll need to deal with those duplicated ids at some point in the future before we add functionality to this. They're the kind of thing we can easily forget and that tend to cause real trouble when we end up writing JS that expects those to be unique.

rajadain:
Is this what it is supposed to look like currently? The views are different in Firefox vs Chrome:

image

image

lewfish:
Yes, that's what it's supposed to look like on Chrome. I didn't test on Firefox.

rajadain:
The "Compare" button in the `/project` page points to a different URL: `/project/XX/scenario/compare` instead of `/compare`. We should make a note of that for when we're handling data between the views:

image

rajadain:
I wish that @ctaylo37's mockup was still accessible from somewhere, since we're still adding to the `/compare` view, so that we could review it while developing. Consider adding a new route `/compare-mockup` or something that points to the mockup.

src/mmw/js/src/compare/controllers.js
```diff
@@ -1,26 +1,65 @@
...
-        });
+        App.rootView.footerRegion.show(compareWindow);
+    },
 
-        $('.selectpicker').selectpicker({
-            tickIcon: ''
-        });
-
-        $('[data-toggle="tooltip"]').tooltip();
+    analyzeCleanUp: function() {
```
rajadain: This should be called `compareCleanUp` by our conventions

ctaylo37:
The differing height issue was created when backbone added in two divs with no height defined.  The two divs added by backbone need to have a height of 100% and the map-container needs to be 20%.

ctaylo37:
@rajadain I will put the compare comp on fileshare.

src/mmw/js/src/compare/templates/compareScenarios.html
```diff
@@ -0,0 +1,7 @@
+<div id="edge-fade"></div><!-- Blur offcanvas scenario -->
+
+<table> <!-- Table allows for horizontal scrolling. -->
```
rajadain: Horizontal scrolling wasn't working for me, and https://github.com/lewfish/model-my-watershed/commit/684224a9879ee696edd387f8d5588fc1045ad6af makes me think that it shouldn't work. Personally I'd like it to be there, I think it would make it a lot more intuitive on touch screens.

ctaylo37:
@rajadain  smb://FILESHARE.internal.azavea.com/projects/Stroud_ModelMyWatershed/documents/design/visual design/v02/2015_07_20_Compare

src/mmw/js/src/compare/views.js
```diff
@@ -0,0 +1,168 @@
...
+    Marionette = require('../../shim/backbone.marionette'),
+    compareWindowTmpl = require('./templates/compareWindow.html'),
+    compareScenariosTmpl = require('./templates/compareScenarios.html'),
+    compareScenarioTmpl = require('./templates/compareScenario.html'),
+    compareMapTmpl = require('./templates/compareMap.html'),
+    compareChartTmpl = require('./templates/compareChart.html'),
+    compareModificationsTmpl = require('./templates/compareModifications.html');
+
+var CompareWindow = Marionette.LayoutView.extend({
+    //model: modelingModels.ProjectModel,
```
rajadain: I think I like this approach better than having it uncommented, like it is here: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/views.js#L265

src/mmw/js/src/compare/templates/compareScenarios.html
ctaylo37: We don't want horizontal scrolling because each scenario will have vertical scrolling in the modification list.  It gets to be too many scrollbars and hard to use.  I could see a use case for being able to swipe to scroll horizontally on touch.  Is this something we can enable just for touch?

rajadain: Ah good point. I forget that not all platforms have invisible scrollbars like OS X and iOS. But at the same time I think that allowing native scroll in one platform but not in another is kind of limiting. Perhaps we can look in to tapping / clicking the edge scenario bringing it into view. I definitely think we should have more than one way of scrolling the list.

rajadain:
Also note that the lack of overlay in Firefox actually allows the user to interact with the map which is probably undesirable:

image

ctaylo37:
Yea the overlay will get fixed with the height issue.

rajadain:
+1 overall, great architecture for the views. I agree that the scope of this PR is pretty limited, and once we have this set of views we can delve in to the individual tasks as noted above. I would recommend looking in to the Firefox styling issue as part of this PR, and maybe making a new issue for investigating scrolling in the compare view, which looks in to:

 * Is it possible to have horizontal scrolling only on touch screens?
 * Consider making it so that clicking / tapping the off-screen scenario bring it in to view
 * Consider adding sliding animation so that the change is more apparent

ctaylo37:
@lewfish can we also have a slide left and slide right animation when the user clicks to move the scenarios?

lewfish:
I was also thinking it would be good to have a sliding animation. Do you have any CSS pointers on how to achieve this?

ctaylo37:
I can write the CSS but im not sure how to coordinate it with the js.

lewfish:
How about you add a new commit to your PR and I'll handle the rest?

On Mon, Jul 20, 2015 at 12:30 PM, Christopher <notifications@github.com>
wrote:

> I can write the CSS but im not sure how to coordinate it with the js.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/WikiWatershed/model-my-watershed/pull/591#issuecomment-122941208>
> .
>



-- 
Lewis Fishgold, Software Developer
Azavea |  340 N 12th St, Ste 402, Philadelphia, PA
Web azavea.com <http://www.azavea.com/>  |  Blog azavea.com/blogs
<http://www.azavea.com/Blogs>  | Twitter @azavea <http://twitter.com/azavea>

ctaylo37:
add it to compare-prototype or here?

lewfish:
To compare-prototype

On Mon, Jul 20, 2015 at 12:37 PM, Christopher <notifications@github.com>
wrote:

> add it to compare-prototype or here?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/WikiWatershed/model-my-watershed/pull/591#issuecomment-122943164>
> .
>



-- 
Lewis Fishgold, Software Developer
Azavea |  340 N 12th St, Ste 402, Philadelphia, PA
Web azavea.com <http://www.azavea.com/>  |  Blog azavea.com/blogs
<http://www.azavea.com/Blogs>  | Twitter @azavea <http://twitter.com/azavea>

ctaylo37:
@lewfish transitions here:
https://github.com/WikiWatershed/model-my-watershed/commit/2b6df25ec230033d4fa36dc42e608c7fb16e5f6f

PR #596, opened 2015-07-20
Improve initial compare views

lewfish:
This PR is to respond to feedback on https://github.com/WikiWatershed/model-my-watershed/pull/591, which was prematurely merged. 

I also added a new issue https://github.com/WikiWatershed/model-my-watershed/issues/597

Connects #591

### DIFF
--- a/src/mmw/js/src/compare/controllers.js
+++ b/src/mmw/js/src/compare/controllers.js
@@ -56,7 +56,7 @@ var CompareController = {
         App.rootView.footerRegion.show(compareWindow);
     },
 
-    analyzeCleanUp: function() {
+    compareCleanUp: function() {
         App.rootView.footerRegion.empty();
     }
 
--- a/src/mmw/js/src/compare/templates/compareScenario.html
+++ b/src/mmw/js/src/compare/templates/compareScenario.html
@@ -1,6 +1,6 @@
 <div class="compare-column">
-    <div class="map-container"></div>
-    <div class="chart-container"></div>
-    <div class="precipitation-container"></div>
-    <div class="modifications-container"></div>
+    <div class="map-region"></div>
+    <div class="chart-region"></div>
+    <div class="precipitation-region"></div>
+    <div class="modifications-region"></div>
 </div>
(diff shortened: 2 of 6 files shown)


rajadain:
Sliding animation is great, makes it a lot more obvious. I'm not sure what is going on with the map height issue in Firefox, but I wouldn't call it fixed:

mmw-compare-map-height

Chrome behaves correctly.

lewfish:
@ctaylo37 Do you have any idea what might be causing this in Firefox? Everytime you click a drop down the map-regions grow vertically.

caseypt:
We should keep an eye on this page in terms of performance. Going to this page as it is now jumps from ~90mb to ~150mb of memory, I think largely from all of the Leaflet instances. I'm not sure how many scenarios people will actually have, but it may be something to look out for as more people use app.

ctaylo37:
@rajadain I told you not to feed the gremlins(leaflet map) after midnight!

rajadain:
Keeping in mind Lewis Carrol's ancient prophecy about browser performance:

> Beware the JavaScript, my son,
> With clause that byte and pause that catch
> Beware the `new Object` word,
> And shun the duplicating Leaflet map

I think we should consider using a screenshot of the map and AoI instead of a real live map, especially since that part of the UI is non-interactive. We might consider using something like this: https://github.com/niklasvh/html2canvas

mmcfarland:
Yeah, that was my concern about the multiple maps.  We are likely to have to implement a full map export service (similar to nyc and HL) so once that infrastructure is set up, we could use it to generate these thumbnails on the backend, too.

lewfish:
Through trial and error I found an imperfect solution to the growing height problem in Firefox https://github.com/lewfish/model-my-watershed/commit/3fe7f3f1b156b87dc69b6f860790378b22490211

I'll spend a little bit more time looking for a general solution, and if I can't find one, I'll create a new issue.

rajadain:
The effect is still there, but it is subtle enough to not impede the user in a significant way:

mmw-compare-map-height-2

I think we can create an issue for it and prioritize it at a later date.

+1, nice job.

lewfish:
That's strange. The effect isn't there at all for me on Firefox.

lewfish:
I'm on Firefox 38.0.5. Which version are you running?

rajadain:
Firefox 39.0 apparently:

image

Issue #583, opened 2015-07-16
Add interactivity to boundary layers


mmcfarland:
Boundary layers should have hover-interactivity (like name).

Issue #576, opened 2015-07-14
Port Congressional boundary layer to data load for stage/prod


mmcfarland:
Currently, congressional districts are loaded with a management command and have a model implemented in `modeling/models.py`.  This shouldn't necessary due to work done in #453.

* Remove the model for Districts
* Rename the table and corresponding code from `modeling_district` to `boundary_district` so it's not confused with Django app tables.
* `pg_dump` the table and load to the external data storage bucket.
* Update the `setupdb.sh` script appropriately

mmcfarland:
When porting, change the column name from `state_short` to name so that the interactivity can just default to that convention, and not have to keep a map of layer:name.

PR #590, opened 2015-07-17
Load stream data

mmcfarland:
Updates to the load script to load the stream network data.

To test, from an app server run:

`./setupdb.sh -s` to load the stream network

verify that
`./setupdb.sh -b` will load the boundary data

Ultimately, the stream network should be visible in the Delaware River Basin.

Connects #581 

cc/ @hectcastro

PR #598, opened 2015-07-20
Add utfgrid interactivity to boundary layers

mmcfarland:
Add a custom label for each feature of each boundary layer.

To test, turn on boundary layers and notice the name displayed on hover.  Reset, turn on other drawing tools, etc, to see the label turn on/off appropriately.

Known issues:
* Congressional Districts name do not appear (#576)
* Extreme right-hand side features will eventually clip the label.  I don't see this as too egregious of a problem at the moment, since the user needs only to pan the map.  If we get feedback regarding it, we can just buffer the `x` position to the width of the label.

Connects #583

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -1,6 +1,7 @@
 "use strict";
 
 var $ = require('jquery'),
+    _ = require('lodash'),
     L = require('leaflet'),
     Marionette = require('../../shim/backbone.marionette'),
     turfRandom = require('turf-random'),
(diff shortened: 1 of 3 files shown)


src/mmw/js/src/draw/views.js
```diff
@@ -72,9 +73,11 @@ var ToolbarView = Marionette.LayoutView.extend({
 });
 
 var SelectAreaView = Marionette.ItemView.extend({
+    $label: $('#boundary-label'),
```
mmcfarland: This element *must* be on the body directly to achieve proper offsets from the viewpoint origin, rather than a parent element origin, so it is not included in this view's template (unfortunately).

jamesmcclain:
Just for future reference, I had to do the following to test:
   1. Increase services VM ram to 2048
   2. Transfer `setupdb.sh` script to the app VM: `scp ./scripts/aws/setupdb.sh vagrant@33.33.34.10:` (there is probably some better way to do this)
   3. ssh to app VM and load the data as described in https://github.com/WikiWatershed/model-my-watershed/pull/590

Once all of that was done, everything worked well: The names of the HUC layers to appear when you enable the layer and hover the mouse over a HUC.  The popups come and go with the visible outline layers.  +1

PR #599, opened 2015-07-20
Port district boundary loading to setupdb

mmcfarland:
Remove the need for `Districts` to be a django model and include
the data in the db setup script. 

The referenced sql.gz file has been uploaded publicly to S3.

Connects #576


jamesmcclain:
I can confirm that the congressional district boundaries do appear after the old table has been removed and the new setupdb.sh script has been run.  This also allows mouse hovering to work with  https://github.com/WikiWatershed/model-my-watershed/pull/598, although just the name of the state, not the district, appears (I don't know whether this was intentional or not).

+1

mmcfarland:
Oh, yeah - great point.  Don't know why I selected state name as the name.  Will update the data to do a [state]-[number] display instead.

jamesmcclain:
Okay, sounds cool.

Issue #558, opened 2015-07-09
better isolation of state between testem tests


lewfish:
The testem tests share a lot of state which makes them harder to work with. I've been dealing with a problem where a test works in isolation but then breaks another test that is run later. I tried to do some clean up when the first test is finished (in particular, removing and re-creating the Leaflet map), but I can't because of some side-effects from yet another test (re: MapView). When trying to fix that test, it interferes with another test. Instead of playing whack-a-mole, it seems like we need to come up with a more general solution.

PR #605, opened 2015-07-22
Improve isolation between unit tests

lewfish:
The main improvements of this PR are to reset the App object after each test, and to pass the MapView a map container id, so it won't conflict with the one that is part of App. Some of the tests had to be updated to accommodate these changes. To test, run the unit tests and trying running the app to make sure it still works. Note that a kludge needed to reset part of the map state (https://github.com/lewfish/model-my-watershed/commit/127f69b17407988abd4e5a9c672164bcf2056001) is no longer necessary. 

Connects #558

### DIFF
--- a/src/mmw/js/src/app.js
+++ b/src/mmw/js/src/app.js
@@ -15,17 +15,22 @@ var App = new Marionette.Application({
 
         // This view is intentionally not attached to any region.
         this._mapView = new views.MapView({
-            model: this.map
+            model: this.map,
+            el: '#map'
         });
 
         this.rootView = new views.RootView();
         this.user = new userModels.UserModel({});
 
-        var header = new views.HeaderView({
+        this.header = new views.HeaderView({
             el: 'header',
             model: this.user
         });
-        header.render();
+        this.header.render();
+
+        // Not set until modeling/controllers.js creates a
+        // new project.
+        this.currProject = null;
     },
 
     load: function(data) {
--- a/src/mmw/js/src/core/tests.js
+++ b/src/mmw/js/src/core/tests.js
@@ -15,6 +15,7 @@ var $ = require('jquery'),
     AppRouter = require('../router').AppRouter,
     chart = require('./chart'),
     settings = require('./settings'),
+    testUtils = require('./testUtils'),
     sandboxTemplate = require('./templates/sandbox.html');
 
 var TEST_SHAPE = {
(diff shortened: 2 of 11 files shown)


src/mmw/js/src/app.js
```diff
@@ -15,17 +15,44 @@ var App = new Marionette.Application({
...
+    },
+
+    // Should be called after each unit test.
+    reset: function() {
+        this.map.off();
+        this.map = null;
+
+        this._mapView.destroy();
+
+        //this.rootView.destroy();
```
rajadain: Is there any benefit to commenting this line out rather than just removing it?

src/mmw/js/src/app.js
```diff
@@ -15,17 +15,44 @@ var App = new Marionette.Application({
...
-            model: this.map
+            model: this.map,
+            mapId: 'map'
         });
 
         this.rootView = new views.RootView();
         this.user = new userModels.UserModel({});
 
-        var header = new views.HeaderView({
+        this.header = new views.HeaderView({
```
caseypt: Outside the scope of this PR, but this should probably get moved to [here](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/draw/controllers.js#L16) and should not be part of the application.

src/mmw/js/src/core/tests.js
```diff
@@ -30,51 +30,55 @@ var chartData = [{x: 'a', y: 1},
...
     xValue = 'x',
     yValue = 'y',
-    sandboxHeight = '500',
-    sandboxWidth = '700',
-    sandboxSelector = '#display-sandbox';
+    displaySandboxHeight = '500',
+    displaySandboxWidth = '700',
+    displaySandboxSelector = '#display-sandbox',
+    sandboxId = 'sandbox',
+    sandboxSelector = '#' + sandboxId;
```
caseypt: Consider doing this for the other test files.

src/mmw/js/src/modeling/tests.js
```diff
@@ -490,19 +507,19 @@ describe('Modeling', function() {
 
                     model.addModification(new models.ModificationModel(mocks.modifications.sample1));
 
-                    // TODO: Refactor to use callbacks instead of delays
-                    setTimeout(delayedAssert, 1100);
+                    this.clock.tick(1000);
```
caseypt: Very cool.

src/mmw/js/src/modeling/models.js
```diff
@@ -288,7 +288,8 @@ var ScenarioModel = Backbone.Model.extend({
                     name: 'precipitation',
                     value: 1.0
                 }
-            ]
+            ], // Setting getResults to false can be useful when testing.
```
caseypt: Consider making this comment more clear that this is only here for unit testing purposes.

src/mmw/js/src/core/views.js
```diff
@@ -198,17 +193,22 @@ var MapView = Marionette.ItemView.extend({
         // supported, nothing more needs to be done since the map has
         // already been centered.
         if (navigator.geolocation) {
-            var options = {
+            var geolocationOptions = {
```
caseypt: Was this changed for clarity?

src/mmw/js/src/modeling/tests.js
```diff
@@ -275,13 +292,17 @@ describe('Modeling', function() {
...
+                });
 
+                /*
+                 TODO: fix this test as part of https://github.com/WikiWatershed/model-my-watershed/issues/393
+
+                it('calls #save on the project and sets the id on every scenario that\'s part of the project', function() {
                     var projectResponse = '{"id":57,"user":{"id":1,"username":"test","email":"test@azavea.com"},"name":"My Project","area_of_interest":{},"is_private":true,"model_package":"tr-55","created_at":"2015-06-03T20:09:11.988948Z","modified_at":"2015-06-03T20:09:11.988988Z"}';
                     this.server.respondWith('POST', '/api/modeling/projects/',
-                            [ 200, { 'Content-Type': 'application/json' }, projectResponse ]);
+                                            [ 200, { 'Content-Type': 'application/json' }, projectResponse ]);
```
caseypt: Extra indentation here.

caseypt:
All tests pass. +1. Nice work.

src/mmw/js/src/core/views.js
lewfish: Yes. At one point I was passing options into the initialize function, so I wanted to make it clearer, but it's not necessary anymore.

src/mmw/js/src/app.js
```diff
@@ -15,17 +15,51 @@ var App = new Marionette.Application({
...
-        header.render();
+        this.header.render();
+
+        // Not set until modeling/controllers.js creates a
+        // new project.
+        this.currProject = null;
+    },
+
+    // Should be called after each unit test.
+    reset: function() {
```
lliss: I feel like this should be part of the test code and not part of the App. We have no need to reset the app in non-test code do we?

src/mmw/js/src/core/tests.js
lliss: I think this is a good idea too. Also, this makes me think we might want to create a test bootstrap file which could also set up the app reset method.

caseypt: I think that idea.

src/mmw/js/src/core/tests.js
```diff
@@ -30,51 +30,55 @@ var chartData = [{x: 'a', y: 1},
...
+            $('<div>', {id: sandboxId}).appendTo('body');
         }
     });
 
     beforeEach(function() {
-        $('#display-sandbox').remove();
+        $(displaySandboxSelector).remove();
         // Use a special sandbox so that we can test responsiveness of chart.
-        $('body').append(sandboxTemplate.render({height: sandboxHeight, width: sandboxWidth}));
+        $('body').append(sandboxTemplate.render({height: displaySandboxHeight, width: displaySandboxWidth}));
```
lliss: This like got long. Consider replacing
`{height: displaySandboxHeight, width: displaySandboxWidth}`
with
```
{
    height: displaySandboxHeight, 
    width: displaySandboxWidth
}
```

src/mmw/js/src/modeling/tests.js
```diff
@@ -14,29 +14,39 @@ var _ = require('lodash'),
...
         App.currProject = new models.ProjectModel();
-    });
 
-    beforeEach(function() {
         this.server = sinon.fakeServer.create();
         this.server.respondImmediately = true;
+
+        // Use fake debounce so that things will execute more synchronously.
+        this.origDebounce = _.debounce;
+        _.debounce = _.identity;
```
lliss: This seems a bit heavy. Can you elaborate on why this was needed?

src/mmw/js/src/modeling/tests.js
```diff
@@ -294,26 +315,23 @@ describe('Modeling', function() {
                     assert.isTrue(scenario1Spy.calledOnce, 'Scenario1 was not saved once.');
                     assert.isTrue(scenario2Spy.calledOnce, 'Scenario2 was not saved once.');
                 });
+                 */
 
                 it('properly associates scenarios with the project when initially saving the project', function() {
                     var projectResponse = '{"id":21,"user":{"id":1,"username":"test","email":"test@azavea.com"},"name":"Test Project","area_of_interest":{},"is_private":true,"model_package":"tr-55","created_at":"2015-06-03T20:09:11.988948Z","modified_at":"2015-06-03T20:09:11.988988Z"}',
```
lliss: I think @rajadain moved some of these long responses to a separate file to keep things tidy. Consider a similar approach.

src/mmw/js/src/app.js
lewfish: Right, we have no need to reset in non-test code. I'll move it to a test utils file I'll put in core/.

src/mmw/js/src/modeling/tests.js
lewfish: I'll put this in a comment:

Using debounce causes adding modifications to become asynchronous due to using setTimeout, which makes it more difficult to test. Using the fake timer in sinon does not help because lodash has its own cached copy of the setTimeout function.

lewfish:
@azaveaci rebuild

hectcastro:
@azavea-bot rebuild

PR #606, opened 2015-07-22
Clone Projects / Embed in ITSI

rajadain:
## Overview

A teacher can create a project for an activity, and set it up with the appropriate Area of Interest, Scenarios, Modifications, etc before sharing with students. When the teacher is ready, they can select "Embed in ITSI" from the project dropdown menu, which will open a `shareModal` window with a URL that they can copy and paste in to an ITSI activity.

When a student logs in to the system, and opens that URL, MMW will first log them in (using the `?itsi_embed=true` flag as implemented for #562 in #578), then clone the project in their workspace and load that project.

## Testing Instructions

 1. Log in using the `mmwterence` user (see Wiki for passwords)
 2. Create a new project. Make arbitrary changes to scenarios.
 3. In the project dropdown, click "Embed in ITSI" and copy the link.
 4. Go to ITSI and log out as `mmwterence`, and log in as `tstudent2`. You should still be logged in as `mmwterence` in MMW.
 5. Open a new tab and paste in the link. You should first be logged in as `tstudent2` (and you might be prompted to create an account if it doesn't already exist), then the project should be cloned and you should be redirected to a new project, identical to the first one, but now owned by `tstudent2`.

 * Run all tests.

## Implementation Notes

There are two parts to this PR: one that introduces project cloning functionality, which is in fact available to all users via the `/project/XX/clone` URL, and the second which allows the front-end to distinguish between ITSI and non-ITSI users, so that it may show the link only to the former and not the latter.

~~Since a project _must_ be public before it can be cloned by another user, when the "Copy" button in the Embed in ITSI modal is clicked, we automatically make the project public. However, if the user copies the text directly, then the resulting link will fail. Given this, perhaps we should make the project public as soon as the link is clicked in the menu. Or perhaps prompt the user to explicitly make the project public before they are allowed to get the embed link.~~

Since a project _must_ be public before it can be cloned by another user, when the "Embed in ITSI" link is clicked, we check to see if the project is already public. If it isn't, then we prompt the user to explicitly make it public. Once they do, we show the share modal. This functionality has also been added to the "Share" link, since the same limitation applies there.

Please see commit messages for details.

Connects #560

### DIFF
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -4,6 +4,7 @@ var $ = require('jquery'),
     L = require('leaflet'),
     _ = require('underscore'),
     router = require('../router.js').router,
+    Backbone = require('../../shim/backbone'),
     Marionette = require('../../shim/backbone.marionette'),
     TransitionRegion = require('../../shim/marionette.transition-region'),
     ZeroClipboard = require('zeroclipboard'),
@@ -529,7 +530,29 @@ var ShareModal = BaseModal.extend({
             self.zc.clip(self.$el.find(self.ui.copy.selector));
         });
 
-        this.$el.modal('show');
+        if (this.model.get('is_private')) {
+            var question = 'This project is currently private. ' +
+                    'You must make it public before it can be shared with others. ' +
+                    'Once public, anyone with the URL can access it.',
+                confirm = new ConfirmModal({
+                    model: new Backbone.Model({
+                        question: question,
+                        confirmLabel: 'Make Public',
+                        cancelLabel: 'Cancel'
+                    })
+                });
+
+            confirm.render();
+            confirm.on('confirmation', function() {
+                var project = self.options.app.currProject;
+                project.set('is_private', false);
+                project.saveProjectAndScenarios();
+
+                self.$el.modal('show');
+            });
+        } else {
+            this.$el.modal('show');
+        }
     },
 
     onModalShown: function() {
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -527,7 +527,7 @@ var ShareModal = BaseModal.extend({
     onRender: function() {
         var self = this;
         this.$el.on('shown.bs.modal', function() {
-            self.zc.clip(self.$el.find(self.ui.copy.selector));
+            self.zc.clip(self.ui.copy);
         });
 
         if (this.model.get('is_private')) {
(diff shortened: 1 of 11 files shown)


rajadain:
Hey @mmcfarland, I'd love to hear your ideas on when / how to make the project public when users click "Embed in ITSI" link. See the **Implementation Notes** section above.

mmcfarland:
Not tested, but the code and tests look good.  As for the question with the timing of making it public, as we discussed, I think we should present the "public" modal first (if it's private) and use that as the mechanism to make public before showing the share link.  These modals will probably get refined and combined in the future.

rajadain:
Added the `.length` to tests in my cleanup commit.

Updated modal behavior according to what we discussed, so that the user is now always prompted if the project is private, in both the Share and Embed in ITSI modals. The transition isn't as jarring as I'd imagined it would be:

mmw-itsi-embed-demo

mmcfarland:
Agreed, transition is perfectly serviceable and I think the right level of explicitness.

src/mmw/js/src/core/views.js
```diff
@@ -526,10 +527,32 @@ var ShareModal = BaseModal.extend({
...
+            self.zc.clip(self.ui.copy);
         });
 
-        this.$el.modal('show');
+        if (this.model.get('is_private')) {
+            var question = 'This project is currently private. ' +
+                    'You must make it public before it can be shared with others. ' +
+                    'Once public, anyone with the URL can access it.',
+                confirm = new ConfirmModal({
+                    model: new Backbone.Model({
```
lewfish: Consider moving this model (and others you've added to views.js) to models.js

lewfish:
+1, tested. Your tests were very thorough. Consider putting some of that explanation about the roundtrip in a comment.

rajadain:
@lewfish I just pushed a modal refactor in c240c1c. Since we were now adding models to the structure, I grouped the views, templates, and models in a subdirectory. I'm not entirely convinced that this is useful, since `modeling/views.js` is no less verbose. Seems like we just moved stuff around without much benefit, except perhaps for some useful defaults added in `core/modals/models.js`.

rajadain:
@lewfish It didn't feel right doing the modal refactor as part of this PR, since it is quite tangentially related, so I moved it into its own #609. The update to the comment you suggested has been made here: https://github.com/rajadain/model-my-watershed/commit/92d0e0c4ac8b50c3b94229b27b051574fba80a0f#diff-18cc37e5fb8b51645e19fc27adad5cd0R89

PR #607, opened 2015-07-22
Add Google Map API key for production

mmcfarland:
This key is generated from a Stroud account and currently has
https://mmw.azavea.com as the approved referrer.  We'll want to have
them update the referrers when we launch this on multiple (or final)
domains.


lewfish:
+1, tested by putting key in developer.py settings.

PR #609, opened 2015-07-23
Group Core Modal Components

rajadain:
As part of the review of #606, it was suggested that we add Backbone models for the Modal Views in core. Since adding them to `/core/models.js` seemed rather messy, I created `core/modals` subdirectory with the relevant `models`, `views`, and `templates`.


lewfish:
+1, tested.

PR #610, opened 2015-07-24
Feature/lol/itsi saved blank project

lliss:
In some cases we want a project to be saved as soon as the user lands on the
site. This is determined by the addition of a URL query parameter. If this
parameter is set, a project will be instantly created and saved. If the user
designates an area of interest and then models it everything will be saved. If
they then move back out and redraw, the same project will be used and the data
will be overwritten.

Connects #561 

To test:

 * Alter `src/mmw/apps/home/views.py` and set `EMBED_FLAG: True` in `def get_client_settings(request):`
 * Visit localhost:8000
 * Check DB, you should see the new project has been save.
 * Draw AOI, and move to modeling page.
 * Create scenarios as normal and ensure they are saved.
 * Back up to analyze mode and return to the project. Everything should remain unchanged.
 * Back up the initial screen and then navigate back to the project. Everything should remain unchanged. 
 * Back up to the main drawing page and click a drawing tool.
 * You should be prompted about the possible hazards.
 * Cancel out. Move back to project page. The project should be unchanged.
 * Back up to the main drawing page and click a drawing tool.
 * Agree to make the changes.
 * Move back to modeling page. The new data should be saved.
 * Create scenarios as normal and ensure they are saved.
 * Reload page at the project URL and ensure that it properly reloads.
 * After reloading, you should remain in "activity mode" which has the same properties as described above.
 * Disable the hard coded value in `src/mmw/apps/home/views.py` 
 * Reload the URL for the project you created. This should have an is_activity flag set on it. Loading this should load the page in "activity mode."

### DIFF
--- a/src/mmw/js/src/app.js
+++ b/src/mmw/js/src/app.js
@@ -23,6 +23,10 @@ var App = new Marionette.Application({
             el: '#map'
         });
 
+        this._mapView.on('change:needs_reset', function(needs) {
+            App.currProject.set('needs_reset', needs);
+        });
+
         this.rootView = new views.RootView();
         this.user = new userModels.UserModel({});
 
--- a/src/mmw/js/src/app.js
+++ b/src/mmw/js/src/app.js
@@ -5,6 +5,7 @@ var $ = require('jquery'),
     Marionette = require('../shim/backbone.marionette'),
     views = require('./core/views'),
     models = require('./core/models'),
+    settings = require('./core/settings'),
     userModels = require('./user/models'),
     userViews = require('./user/views');
 
@@ -12,6 +13,9 @@ var App = new Marionette.Application({
     initialize: function() {
         this.restApi = new RestAPI();
         this.map = new models.MapModel();
+
+        // If in embed mode we are by default in activity mode.
+        this.activityMode = settings.getSettings().itsi_embed;
 
         // This view is intentionally not attached to any region.
         this._mapView = new views.MapView({
--- a/src/mmw/js/src/app.js
+++ b/src/mmw/js/src/app.js
@@ -15,7 +15,8 @@ var App = new Marionette.Application({
         this.map = new models.MapModel();
 
         // If in embed mode we are by default in activity mode.
-        this.activityMode = settings.getSettings().itsi_embed;
+        var activityMode = settings.get('itsi_embed');
+        settings.set('activityMode', activityMode);
 
         // This view is intentionally not attached to any region.
         this._mapView = new views.MapView({
--- a/src/mmw/js/src/core/modals/views.js
+++ b/src/mmw/js/src/core/modals/views.js
@@ -14,8 +14,18 @@ var _ = require('underscore'),
 var ModalBaseView = Marionette.ItemView.extend({
     className: BASIC_MODAL_CLASS,
 
-    attributes: {
-        'tabindex': '-1'
+    attributes: function() {
+        var defaults = { 'tabindex': '-1' },
+            feedbackRequired = {
+                'data-backdrop': 'static',
+                'data-keyboard': 'false'
+            };
+
+        // If feedback is required, we don't allow the user to get away from
+        // the modal without clicking confirm or deny. This way we ensure we
+        // get an event one way or another.
+        // http://getbootstrap.com/javascript/#modals-options
+        return this.model.get('feedbackRequired') ? _.extend(defaults, feedbackRequired) : defaults;
     },
 
     events: {
--- a/src/mmw/js/src/core/modals/views.js
+++ b/src/mmw/js/src/core/modals/views.js
@@ -55,16 +55,22 @@ var ConfirmView = ModalBaseView.extend({
     template: modalConfirmTmpl,
 
     ui: {
-        confirmation: '.confirm'
+        confirmation: '.confirm',
+        deny: '.btn-default'
     },
 
     events: _.defaults({
-        'click @ui.confirmation': 'primaryAction'
+        'click @ui.confirmation': 'primaryAction',
+        'click @ui.deny': 'dismissAction'
     }, ModalBaseView.prototype.events),
 
     primaryAction: function() {
         this.triggerMethod('confirmation');
         this.hide();
+    },
+
+    dismissAction: function() {
+        this.triggerMethod('deny');
     }
 });
 
--- a/src/mmw/js/src/core/models.js
+++ b/src/mmw/js/src/core/models.js
@@ -12,7 +12,8 @@ var MapModel = Backbone.Model.extend({
         zoom: 0,
         areaOfInterest: null,           // GeoJSON
         halfSize: false,
-        geolocationEnabled: true
+        geolocationEnabled: true,
+        previousAreaOfInterest: null
     },
 
     revertMaskLayer: function() {
@@ -44,6 +45,18 @@ var MapModel = Backbone.Model.extend({
 
             this.set('areaOfInterest', aoi);
         }
+    },
+
+    stashAOI: function() {
+        // Since we oscillate between an area of interest and a blank map, stash
+        // non-null AOI.
+        if (!_.isNull(this.get('areaOfInterest'))) {
+            this.set('previousAreaOfInterest', _.clone(this.get('areaOfInterest')));
+        }
+    },
+
+    revertAOI: function() {
+        this.set('areaOfInterest', _.clone(this.get('previousAreaOfInterest')));
     },
 
     setHalfSize: function(fit) {
--- a/src/mmw/js/src/core/settings.js
+++ b/src/mmw/js/src/core/settings.js
@@ -5,15 +5,24 @@ var defaultSettings = {
     stream_layers: {}
 };
 
-function setSettings(key, value) {
-    defaultSettings[key] = value;
+var settings = (function() {
+    return window.clientSettings ? window.clientSettings : defaultSettings;
+})();
+
+function set(key, value) {
+    settings[key] = value;
+    return value;
 }
 
-function getSettings() {
-    return window.clientSettings ? window.clientSettings : defaultSettings;
+function get(key) {
+    try {
+        return settings[key];
+    } catch (exc) {
+        return undefined;
+    }
 }
 
 module.exports = {
-    getSettings: getSettings,
-    setSettings: setSettings
+    get: get,
+    set: set
 };
--- a/src/mmw/js/src/core/utils.js
+++ b/src/mmw/js/src/core/utils.js
@@ -31,6 +31,17 @@ var utils = {
         return params;
     },
 
+    /**
+     * Returns the value of a query param from the URL if it exists or null:
+     * https://stackoverflow.com/questions/901115
+     */
+    getParameterByName: function(name) {
+        name = name.replace(/[\[]/, '\\[').replace(/[\]]/, '\\]');
+        var regex = new RegExp('[\\?&]' + name + '=([^&#]*)'),
+            results = regex.exec(location.search);
+        return results === null ? '' : decodeURIComponent(results[1].replace(/\+/g, ' '));
+    },
+
     // Convert a backbone collection into a canonical MD5 hash
     getCollectionHash: function(collection) {
         // JSON objects are not guaranteed consistent order in string
--- a/src/mmw/js/src/core/utils.js
+++ b/src/mmw/js/src/core/utils.js
@@ -8,22 +8,22 @@ var utils = {
     // Takes queryString of format "key1=value1&key2=value2"
     // Returns object of format {key1: value1, key2: value2}
     // From http://stackoverflow.com/a/11671457/2053314
-    parseQueryString: function (queryString){
+    parseQueryString: function(queryString) {
         var params = {};
-        if(queryString){
+        if (queryString) {
             _.each(
-                _.map(decodeURI(queryString).split(/&/g),function(el){
+                _.map(decodeURI(queryString).split(/&/g), function(el) {
                     var aux = el.split('='), o = {};
-                    if(aux.length >= 1){
+                    if (aux.length >= 1) {
                         var val;
-                        if(aux.length === 2) {
+                        if (aux.length === 2) {
                             val = aux[1];
                         }
                         o[aux[0]] = val;
                     }
                     return o;
                 }),
-                function(o){
+                function(o) {
                     _.extend(params,o);
                 }
             );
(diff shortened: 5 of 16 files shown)


src/mmw/js/src/core/utils.js
```diff
@@ -31,6 +31,17 @@ var utils = {
         return params;
     },
 
+    /**
+     * Returns the value of a query param from the URL if it exists or null:
+     * https://stackoverflow.com/questions/901115
+     */
+    getParameterByName: function(name) {
```
rajadain: I like this function a lot better than the `parseQueryString` above. We might consider removing `parseQueryString` entirely and replacing its one usage here: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/user/controllers.js#L23 with something like this:

```javascript
var nextParam = utils.getParameterByName('next');
...
model = new models.ItsiSignUpFormModel({
    ...
    next: nextParam ? nextParam : '/'
});
```

src/mmw/js/src/draw/controllers.js
```diff
@@ -27,6 +29,8 @@ var DrawController = {
         App.rootView.geocodeSearchRegion.show(geocodeSearch);
         App.rootView.drawToolsRegion.show(toolbarView);
 
+        enableSingleProjectMode();
```
rajadain: This function's name seems to imply that Single Project Mode is _always_ enabled. Consider renaming to `enableSingleProjectModeIfItsi`, or to place the condition here instead of within the function:

```javascript
if (coreUtils.getParameterByName('initialize_itsi') === 'true') {
    enableSingleProjectMode();
}
```

src/mmw/js/src/draw/controllers.js
```diff
@@ -51,6 +55,28 @@ var DrawController = {
...
 };
 
+/**
+ * If itsi flag is set, prepare a project immedialty upon visiting the page.
+ * This will be the only project the user can save during this session.
+ */
+function enableSingleProjectMode() {
+    if (coreUtils.getParameterByName('initialize_itsi') === 'true') {
+
+        App.singleProjectMode = true;
```
rajadain: Consider explicitly setting this to `false` by default.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -43,40 +44,65 @@ var ModelingController = {
...
+
+                // Clear current scenarios and start over.
+                // Must convert to an array first to avoid conflicts with
+                // collection events that are disassociating the model during
+                // the loop.
+                var locks = [];
+                _.each(project.get('scenarios').toArray(), function(model) {
+                    var $lock = $.Deferred();
+                    locks.push($lock);
+                    model.destroy({
```
rajadain: Since having once gone back from the `/project` view to `/analyze`, even if I make no changes, all the Scenarios are killed, I'm afraid that people will easily make mistakes and end up deleting a lot of work. I think we should not delete scenarios unless the AoI has changed, so if one goes from `/project` to `/analyze` to `/project` without making any changes, their old work is preserved.

Also, when resetting the AoI, the user should be shown a Confirm Modal that informs them of scenario deletion, and gives them a chance to cancel that.

rajadain:
Great work. I think we can follow up with the changes talked about with @mmcfarland. Have a look at [`middleware.py`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/user/middleware.py#L33) and `client_settings` in [`views.py`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/home/views.py) for ideas about stringing through a flag from Django to Backbone without query parameter parsing.

src/mmw/js/src/core/views.js
```diff
@@ -414,6 +459,135 @@ var ModificationPopupView = Marionette.ItemView.extend({
...
+        this.ui.input.focus().select();
+    },
+
+    signIn: function() {
+        this.options.app.getUserOrShowLogin();
+    }
+
+});
+
+>>>>>>> Trigger event when canceling a confirmation modal.
```
rajadain: Can't really test this until merge conflicts are resolved.

rajadain:
Having a look now.

rajadain:
+1, functionally works very well. Great work! By the way, we don't have to hard-set the value of `EMBED_FLAG` in `home/views.py`, we can just visit the URL with `/?itsi_embed=true` which will enable it (and be a more accurate test).

Tested that project is saved as soon as one lands on the draw page when `activityMode` is enabled, that one can go back and forth at ease without harming the Scenarios, that one is prompted a warning when resetting the Area of Interest, that "Cancel" and "Make Changes" in that warning work as expected, and that loading a project that was started as an Activity turns `activityMode` on.

A loophole I found is when a project that is marked as an activity is made public, and accessed by a different / anonymous user. Since `activityMode` is still set to `true`, they can go back and change the AoI, which gives them the same warning, but when they try to proceed again the `DELETE` scenario requests are rejected with `403`s. I think that `activityMode` should only be set if the current user is the project owner, but perhaps we need to think more about this.

Going to look through the code now.

src/mmw/js/src/core/views.js
```diff
@@ -1,6 +1,7 @@
 "use strict";
 
-var $ = require('jquery'),
+var Backbone = require('../../shim/backbone'),
```
rajadain: `Backbone` is not used in this file, is causing lint errors.

lliss:
@azavea-bot rebuild

src/mmw/js/src/core/views.js
lliss: Fixed and rebased.

lliss:
@azavea-bot rebuild

src/mmw/js/src/core/views.js
```diff
@@ -203,6 +210,45 @@ var MapView = Marionette.ItemView.extend({
         }
     },
 
+    aoiChangeWarning: _.debounce(function() {
```
rajadain: Cool use of `_.debounce` here, I didn't know we could define view methods this way.

src/mmw/js/src/app.js
```diff
@@ -13,11 +14,19 @@ var App = new Marionette.Application({
         this.restApi = new RestAPI();
         this.map = new models.MapModel();
 
+        // If in embed mode we are be default in activity mode.
+        var activityMode = settings.get('itsi_embed');
+        settings.set('activityMode', activityMode);
```
rajadain: Consider renaming to `activity_mode` to better match other setting keys.

src/mmw/js/src/draw/controllers.js
```diff
@@ -51,6 +55,29 @@ var DrawController = {
...
+ * If we are in embed mode then the project is an activity and we want to keep
+ * the same project reguardless of changes to the AOI. This prepares a project
+ * immedialty upon visiting the page and will be the only project the user can
+ * save during this session.
+ */
+function enableSingleProjectModeIfActivity() {
+    if (settings.get('activityMode')) {
+        if (!App.currProject) {
+            var project = new modelingModels.ProjectModel({
+                name: 'New Project',
```
rajadain: New projects are named "Untitled Project" by default in the regular case.

src/mmw/js/src/core/settings.js
```diff
@@ -5,15 +5,24 @@ var defaultSettings = {
...
+        return settings[key];
+    } catch (exc) {
+        return undefined;
+    }
 }
 
 module.exports = {
-    getSettings: getSettings,
-    setSettings: setSettings
+    get: get,
```
rajadain: Thanks, I like this refactor a lot better than the original implementation.

src/mmw/js/src/core/modals/views.js
```diff
@@ -55,16 +55,22 @@ var ConfirmView = ModalBaseView.extend({
...
+        'click @ui.confirmation': 'primaryAction',
+        'click @ui.deny': 'cancelAction'
     }, ModalBaseView.prototype.events),
 
     primaryAction: function() {
         this.triggerMethod('confirmation');
         this.hide();
+    },
+
+    cancelAction: function() {
```
rajadain: A similar method exists in `user/views.js` where it is called `dismissAction`.

This function only works when the user clicks cancel. If they click outside the modal, or use <kbd>Escape</kbd>, then the app is in an inconsistent state where neither event was triggered:

mmw-clear-aoi-issue

Consider refactoring so that instead of listening to the `deny` event, the consumer listens for the `hidden.bs.modal` and checks for a flag that indicates if `primaryAction` was executed or not.

src/mmw/js/src/app.js
lliss: Trying to stick with camelCase throughout the application where possible we use snake_case where we are getting attributes directly from the API so we don't have to do nasty conversion of key formats. I think that was decided to be the lesser evil though I think that is the exception rather than the rule.

src/mmw/js/src/draw/controllers.js
lliss: Originally I had this named in a way that would differentiate it from the normal projects. This is generic enough that I should either make it more specific such as "New Activity" or just make it the same "Untitled Project" we are using elsewhere.

src/mmw/js/src/core/modals/views.js
lliss: Oh that's an annoying issue. I might make a separate card for that.

src/mmw/js/src/app.js
rajadain: Cool, makes sense.

src/mmw/js/src/draw/controllers.js
rajadain: Since we have the `is_activity` flag, it allows us (the backend people) to easily differentiate activities and projects, and it isn't different enough to enable users to clearly differentiate between the two. I like the "New Activity" idea, which is clear and explicit.

Any thoughts on this, @mmcfarland ?

rajadain: Thanks for clarifying the thought process though, it makes sense and I appreciate it. I made my original comment to ensure this wasn't an oversight.

src/mmw/js/src/core/modals/views.js
lliss: Fixed in efb8daf

lliss:
Fixed the issue with the failing DELETE requests in 4cf078e.

rajadain:
Taking another look

src/mmw/js/src/core/modals/views.js
```diff
@@ -14,8 +14,18 @@ var _ = require('underscore'),
...
+            feedbackRequired = {
+                'data-backdrop': 'static',
+                'data-keyboard': 'false'
+            };
+
+        // If feedback is required, we don't allow the user to get away from
+        // the modal without clicking confirm or deny. This way we ensure we
+        // get an event one way or another.
+        // http://getbootstrap.com/javascript/#modals-options
+        return this.model.get('feedbackRequired') ? _.extend(defaults, feedbackRequired) : defaults;
```
rajadain: Very cool, I love this solution!

src/mmw/js/src/modeling/controllers.js
```diff
@@ -41,42 +43,82 @@ var ModelingController = {
                         project.get('scenarios').makeFirstScenarioActive();
                     }
                     project.getResultsIfNeeded();
+
+                    // If this project is an activity then the application's
+                    // behaior changes but only if the project is owned by the
```
rajadain: Typo, should be "behavior" instead of "behaior"

rajadain:
If going back to existing AoI and selecting "Boundary Layers", they are still clickable if they are within the AoI, and lead to an inconsistent state:

mmw-clear-aoi-issue-3

But the good thing is that the inconsistency isn't persisted, as seen in the above GIF. The only thing it might affect are the model results for which we send the AoI each time, and they aren't recalculated when loading the project next time because the inputs and modifications are the same, but the AoI is different.

If this is not easily resolved, we might make a separate card for this, since it doesn't seem to be too damaging / not too easily reproduced.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -41,42 +43,82 @@ var ModelingController = {
...
+                    // current user.
+                    if (App.user.loggedInUserMatch(project.get('user_id')) && project.get('is_activity')) {
```
rajadain: We cannot use `App.user.loggedInUserMatch` here because the `App.user` isn't initialized until later. In the following example, I revisit my own project made using `itsi_embed=true` in non-embedded mode, and even though it is an activity and I own it, `activityMode` is `false`:

image

lliss:
Okay. I removed the test to ensure the user owns the activity. That's an edge case and we'll address it in a new issue. Same for the boundary layer selector not being properly discarded.

rajadain:
Alright, going in for one final pass

rajadain:
Very minor style recommendation: consider constraining commit body to 79 characters to prevent odd wrapping in GitHub:

image

src/mmw/js/src/app.js
```diff
@@ -13,11 +14,19 @@ var App = new Marionette.Application({
         this.restApi = new RestAPI();
         this.map = new models.MapModel();
 
+        // If in embed mode we are be default in activity mode.
```
rajadain: Typo, "be default" should be "by default"

src/mmw/js/src/core/models.js
```diff
@@ -46,6 +47,18 @@ var MapModel = Backbone.Model.extend({
         }
     },
 
+    stashAOI: function() {
+        // Since we oscilate between an area of interest and a blank map, stash
```
rajadain: Typo, "oscilate" should be "oscillate"

src/mmw/js/src/core/views.js
```diff
@@ -10,7 +10,9 @@ var $ = require('jquery'),
     modificationConfigUtils = require('../modeling/modificationConfigUtils'),
     headerTmpl = require('./templates/header.html'),
     modificationPopupTmpl = require('./templates/modificationPopup.html'),
-    areaOfInterestTmpl = require('../core/templates/areaOfInterestHeader.html'),
```
rajadain: Hah this was a funny self-reference. Great catch fixing that.

lliss:
I always thought it wrapped at 80.

rajadain:
+1 great work, great code. Works very well. I hope we have to do minimal modifications to this as we go along, although I suspect that as we start embedding it inside ITSI we may come across some edge cases that we didn't think about here. But this is good to merge (once the conflicts have been resolved).

I did find one another issue regarding changing the AoI, making a separate card for that. Will reference it when ready.

rajadain:
Added #638 for the above mentioned issue.

Issue #563, opened 2015-07-09
Report and Accept ITSI Saved State


mmcfarland:
ITSI Portal will communicate with MMW via [post messages](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) that communicate a saved state in an Activity.  My understanding of the workflow is as follows:

1. Upon visiting an Activity page, a URL is requested as descripted in #560 via an iframe (or new tab, probably for the former).
2. When the iframe loads, a post message is received with any existing saved state. 
3. The iframe is then polled periodically to provide any updated saved state.
4. Repeat for every page load

Currently, the only anticipated state to save is the student's project URL which may/will be different from the one originally requested in point 1 above due to the redirections mentioned in #560 and #561.  If state is received with a URL, the page should redirect to it.  The two methods ITSI uses for loading and save state are `loadInteractive` and `saveInteractive` and can be demoed at:

http://concord-consortium.github.io/lara-interactive-api/
(click "Mock Interactive" on the Interactive Source container to manually manipulate the values).

mmcfarland:
API Documentation

http://concord-consortium.github.io/lara-interactive-api/docs/

mmcfarland:
It appears Concord has developed a library for handling postMessages, may be worth an evaluation:

https://github.com/concord-consortium/iframe-phone

rajadain:
Since all the state of our app is contained within the URL, perhaps we don't need to use the `loadInteractive` and `saveInteractive` methods in LARA, and instead use only the [`setLearnerUrl`](http://concord-consortium.github.io/lara-interactive-api/docs/#setlearnerurl) method to update the URL to be loaded for the student once we have the final state.

rajadain:
I think we may need to revise some our reliance on the mutability of embedded URLs. After creating an activity with the Staging URL appended with `/?itsi_embed=true`, assigning it to a student, logging in as the student and then starting the activity, the following things happened:

 1. I was automatically logged in as my student account
 2. Because of #561, the blank project was automatically saved to my account
 3. I was able to proceed and manipulate the application with minimal deterrence (the only difficulty was because of the narrow viewport of the iframe)

However, once I got to my project and made some changes:

![image](https://cloud.githubusercontent.com/assets/1430060/9018173/4d9af18a-37aa-11e5-9d58-54b95367da13.png)

Sent a `setLearnerUrl` message:

```javascript
var message = {
    type: 'setLearnerUrl',
[...]

PR #613, opened 2015-07-24
ITSI Activity Interaction

rajadain:
## Overview

ITSI uses [LARA](https://github.com/concord-consortium/LARA) for all their teacher / learner activities. MMW will be embedded as in iframe within such an activity. In order to save state and update the activity with MMW's state as a user progresses through the workflow, we must tell ITSI of the most relevant URL for the user which will be used to resume their session should they return at a later date. Using the [iFrame Phone](https://github.com/concord-consortium/iframe-phone) library, we can send the most recent URL to the parent frame.

## Testing Instructions

 1. Login as an ITSI user to MMW and create a project
 2. Go to http://concord-consortium.github.io/lara-interactive-api/ and in the "Interactive Source" text box, paste in the URL of the project you created in Step 1, with the additional suffix of `?itsi_embed=true` which enables the interface and press <kbd>Enter</kbd>:  
  image
 3. You should see the parent frame log the incoming message:  
  image
 4. Click "getLearnerUrl" in the parent frame. You should see a similar message appear in the log:  
  image

## Notes

This needs to be tested inside an actual activity with the proper workflow to see if it is enough to send the learner URL, or if we must send something for `saveInteractive` and `loadInteractive` calls as well. If we are unable to test this on our own, we may need to ask for Scott's help.

Connects #563

### DIFF
--- a/src/mmw/js/src/app.js
+++ b/src/mmw/js/src/app.js
@@ -6,6 +6,7 @@ var $ = require('jquery'),
     views = require('./core/views'),
     models = require('./core/models'),
     settings = require('./core/settings'),
+    itsi = require('./core/itsiEmbed'),
     userModels = require('./user/models'),
     userViews = require('./user/views');
 
@@ -17,6 +18,11 @@ var App = new Marionette.Application({
         // If in embed mode we are by default in activity mode.
         var activityMode = settings.get('itsi_embed');
         settings.set('activityMode', activityMode);
+
+        // Initialize embed interface if in activity mode
+        if (activityMode) {
+            this.itsi = new itsi.ItsiEmbed();
+        }
 
         // This view is intentionally not attached to any region.
         this._mapView = new views.MapView({
--- a/src/mmw/js/src/core/settings.js
+++ b/src/mmw/js/src/core/settings.js
@@ -1,6 +1,7 @@
 "use strict";
 
 var defaultSettings = {
+    itsi_embed: false,
     base_layers: {},
     stream_layers: {}
 };
--- a/src/mmw/js/src/modeling/controllers.js
+++ b/src/mmw/js/src/modeling/controllers.js
@@ -44,9 +44,14 @@ var ModelingController = {
                     }
                     project.fetchResultsIfNeeded();
 
-                    // If this project is an activity then the application's behaior changes.
+                    // If this project is an activity then the application's behavior changes.
                     if (project.get('is_activity')) {
                         settings.set('activityMode', true);
+                    }
+
+                    // Send URL to parent if in embed mode
+                    if (settings.get('itsi_embed')) {
+                        App.itsi.setLearnerUrl(window.location.href);
                     }
                 });
         } else {
@@ -95,6 +100,9 @@ var ModelingController = {
                     initViews(project);
                     project.fetchResultsIfNeeded();
                     router.navigate(project.getReferenceUrl());
+                    if (settings.get('itsi_embed')) {
+                        App.itsi.setLearnerUrl(window.location.href);
+                    }
                 }
             } else {
                 project = new models.ProjectModel({
@@ -110,6 +118,9 @@ var ModelingController = {
                 setupNewProjectScenarios(project);
                 project.on('change:id', function() {
                     router.navigate(project.getReferenceUrl());
+                    if (settings.get('itsi_embed')) {
+                        App.itsi.setLearnerUrl(window.location.href);
+                    }
                 });
 
                 initScenarioEvents(project);
@@ -141,6 +152,9 @@ function initScenarioEvents(project) {
     scenariosColl.on('change:activeScenario change:id', function(scenario) {
         mapView.updateModifications(scenario.get('modifications'));
         router.navigate(project.getReferenceUrl());
+        if (settings.get('itsi_embed')) {
+            App.itsi.setLearnerUrl(window.location.href);
+        }
     });
 }
 
--- a/src/mmw/bundle.sh
+++ b/src/mmw/bundle.sh
@@ -105,7 +105,7 @@ CONCAT_VENDOR_CSS_COMMAND="cat \
 JS_DEPS=(jquery backbone backbone.marionette bootstrap bootstrap-select \
          leaflet leaflet-draw leaflet.locatecontrol leaflet-plugins lodash \
          underscore d3 nunjucks turf-area turf-bbox-polygon turf-buffer \
-         turf-destination turf-random zeroclipboard blueimp-md5)
+         turf-destination turf-random zeroclipboard blueimp-md5 iframe-phone)
 BROWSERIFY_EXT=""
 BROWSERIFY_REQ=""
 for DEP in "${JS_DEPS[@]}"
--- a/src/mmw/js/src/core/itsiEmbed.js
+++ b/src/mmw/js/src/core/itsiEmbed.js
@@ -2,6 +2,7 @@
 
 var _ = require('lodash'),
     iframePhone = require('iframe-phone'),
+    App = require('../app.js'),
     EMBED_FLAG = 'itsi_embed',
     QUERY_SUFFIX = EMBED_FLAG + '=true';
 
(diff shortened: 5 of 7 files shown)


hectcastro:
@azavea-bot rebuild

src/mmw/apps/home/views.py
```diff
@@ -30,9 +30,11 @@ def get_stream_layers():
     return stream_layers
 
 
-def get_client_settings():
+def get_client_settings(request):
+    EMBED_FLAG = settings.ITSI['embed_flag']
     client_settings = {
         'client_settings': json.dumps({
+            EMBED_FLAG: request.session.get(EMBED_FLAG, False),
```
rajadain: I'm just reusing the existing `embed_flag`, so it is relatively easy to change (although not really, since we explicitly define it on the Backbone side). Effectively, this will look like:

```javascript
{
    itsi_embed: true,
    ...
}
```

src/mmw/apps/home/views.py
```diff
@@ -44,7 +46,7 @@ def get_client_settings():
 def get_context(request):
     context = {}
     context.update(csrf(request))
-    context.update(get_client_settings())
+    context.update(get_client_settings(request))
```
rajadain: Passing in the `request` since this function reads a session variable.

src/mmw/js/src/core/settings.js
```diff
@@ -1,6 +1,7 @@
 "use strict";
 
 var defaultSettings = {
+    itsi_embed: false,
```
rajadain: Off by default

src/mmw/js/src/modeling/controllers.js
```diff
@@ -41,6 +41,10 @@ var ModelingController = {
                         project.get('scenarios').makeFirstScenarioActive();
                     }
                     project.getResultsIfNeeded();
+
+                    if (App.itsi_embed) {
+                        App.itsi.setLearnerUrl(window.location.href);
```
rajadain: Call `setLearnerUrl` at the end of rendering an existing project.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -71,6 +75,9 @@ var ModelingController = {
 
             project.on('change:id', function() {
                 router.navigate(project.getReferenceUrl());
+                if (App.itsi_embed) {
+                    App.itsi.setLearnerUrl(window.location.href);
```
rajadain: Call `setLearnerUrl` once a new project has been assigned an ID.

src/mmw/js/src/core/itsiEmbedInterface.js
```diff
@@ -0,0 +1,27 @@
...
+    QUERY_SUFFIX = EMBED_FLAG + '=true';
+
+var ItsiEmbedInterface = function() {
+    this.phone = new iframePhone.getIFrameEndpoint();
+
+    this.url = window.location.href + '?' + QUERY_SUFFIX;
+
+    this.setLearnerUrl = function(url) {
+        if (url) {
+            this.url = url + (url.indexOf('?') > 0 ? '&' : '?') + QUERY_SUFFIX;
```
rajadain: Will attach the query string after a `&` if query strings already exist, else will use `?`

src/mmw/js/src/core/itsiEmbedInterface.js
```diff
@@ -0,0 +1,27 @@
...
+    iframePhone = require('iframe-phone'),
+    EMBED_FLAG = 'itsi_embed',
+    QUERY_SUFFIX = EMBED_FLAG + '=true';
+
+var ItsiEmbedInterface = function() {
+    this.phone = new iframePhone.getIFrameEndpoint();
+
+    this.url = window.location.href + '?' + QUERY_SUFFIX;
+
+    this.setLearnerUrl = function(url) {
```
rajadain: We expose `setLearnerUrl` so it can be called from outside (currently only `modeling/controllers`)

src/mmw/js/src/core/itsiEmbedInterface.js
```diff
@@ -0,0 +1,27 @@
...
+        if (url) {
+            this.url = url + (url.indexOf('?') > 0 ? '&' : '?') + QUERY_SUFFIX;
+        }
+
+        this.phone.post('setLearnerUrl', this.url);
+    };
+
+    this.phone.addListener('getLearnerUrl', _.bind(this.setLearnerUrl, this));
```
rajadain: If parent frame (an activity in ITSI) requests URL with a `getLearnerUrl` message, execute `setLearnerUrl` in response.

src/mmw/js/src/app.js
```diff
@@ -3,6 +3,8 @@
 var $ = require('jquery'),
     _ = require('lodash'),
     Marionette = require('../shim/backbone.marionette'),
+    settings = require('./core/settings').getSettings(),
+    itsi = require('./core/itsiEmbedInterface'),
```
rajadain: I'm open to suggestions for names better than `itsiEmbedInterface`. The PascalCasing looks weird and unwieldy but I couldn't come up with anything shorter that wasn't confusing.

lliss: We don't seem to have a convention about file naming in general. The closest thing I can find is "NTFS file/folder names are not [case sensitive], so be careful..." from out C# guidelines. I'm generally in favor of lowercase names with dashes or underscores. Traditionally I've preferred underscores.

src/mmw/js/src/modeling/controllers.js
lliss: Do we not care about changing the learnerUrl when changing scenarios?

rajadain: Ah good point! I'll add that.

rajadain: Added here: https://github.com/rajadain/model-my-watershed/commit/d7baa08e042ee8e35d2333df5f189cb3ae72fe0e#diff-18cc37e5fb8b51645e19fc27adad5cd0R104

lliss:
+1. I have verified that this works properly. Code looks good.

rajadain:
There will be potential merge conflicts / alterations after #610, and I need to test this actually inside an ITSI embed, so will hold off on merging until these prerequisites are merged / available.

rajadain:
Rebased. Now that we've heard from Scott, I should be able to test this within in actual Activity before merging.

rajadain:
After adding behavior for `loadInteractive` and `interactiveState` (which is what they call `saveInteractive` under the hood), the app is working inside an activity:
 
ezgif com-optimize

Unfortunately, since we can't act on a `loadInteractive` call until after the front-end has been rendered, we automatically create a new project for the user (from #561), which must then be deleted before redirecting to the existing project.

## Testing Instructions

To test the API calls, just follow the instructions in the main description of this PR, but also try the `loadInteractive` and `saveInteractive` buttons. Copy the values received in "Inbound Data" by `saveInteractive` to the "Outbound Data" field before clicking `loadInteractive`, like so:

image

To test inside an activity:

 1. Checkout the branch and run the bundle `./scripts/bundle.sh --debug`
 2. Run [ngrok](https://ngrok.com/) `ngrok http 8000`
 3. Log on to [ITSI](https://learn.staging.concord.org) as `mmwterence` and edit the [**MMW Example 6abdbbf1**](https://learn.staging.concord.org/eresources/506/matedit?iFrame=true) activity
 4. Update the Activity Name to have the URL of your ngrok instance for identification purposes and click "Save":  
  image
 5. Edit the first page of the activity:  
  image
 6. Edit the embedded app:  
  image
 7. Update the name and URL to match your ngrok instance:  
  image
 8. Click "Save"
 9. Log on to [ITSI](https://learn.staging.concord.org) as `tstudent2` and "Run" the activity:  
  image
 10. Play around in the environment. Ensure that your work is saved if you close this tab and "Run" the activity again. Click "Undo All My Work" to reset the state of the app:  
  image  
  Note that this will reset the saved URL in the ITSI activity, thus the next time the user will start from an empty / new project. This will not delete the project in the user's account, which is still alive and available at the original URL. Of course, since the URL is never displayed to the user in this embed mode, without some UI to browse their own projects, they can never return to it.

lewfish:
Link to activity: https://6abdbbf1.ngrok.io/?itsi_embed=true

lewfish:
+1, tested.

rajadain:
Rebased on top of `develop`.

rajadain:
Seeing some flaky behavior in the embedded app where it sometimes doesn't properly save / load the project. Will test some more and see if I can figure that out before merging.

rajadain:
I created a sample activity within ITSI with [Concord's test iframe](https://concord-consortium.github.io/lara-interactive-api/iframe.html), which allowed me to observe the behavior of the parent activity. This is what was logged:

```
(log messages show up here) Starting the dummy iframe interactive
Phone ready
Phone call: getExtendedSupport: undefined
Phone responded: extendedSupport - [object Object]
Phone call: getLearnerUrl: undefined
Phone responded: setLearnerUrl - https://concord-consortium.github.io/lara-interactive-api/iframe.html
Phone call: loadInteractive: [object Object]
Logging RPC call: lara-logging-present
Phone call: getInteractiveState
Phone responded: interactiveState
Phone call: getInteractiveState
Phone responded: interactiveState
...
```

So `getLearnerUrl` is called when initializing, and then `getInteractiveState` is routinely called roughly every second. Previously, we would respond to every `getLearnerUrl` and `getInteractiveState` call with the current state of the app. This was problematic because `getLearnerUrl` is called before `loadInteractive`, thus we would overwrite the saved "correct" state with the current "incorrect" state, before the project could be loaded. Since we do receive the old URL as part of `loadInteractive`, we would attempt to load that URL, but we would also be getting `getInteractiveState` calls in the meantime, leading to confusing race conditions and unpredictable results.

The latest commit splits the logic of `setLearnerUrl`, which _always_ sets the current URL and sends it to the parent, giving us full control, and `sendLearnerUrlOnlyFromProjectView`, which is the event handler listening to `getLearnerUrl` and `getInteractiveState` postMessages from the parent. Thus, we now respond to those state inquiries only when we are in a (reliable) project view, and not otherwise. This is ensured by checking for a `project/123` type presence in the URL being shared.

PR #614, opened 2015-07-27
Update comment in migration

rajadain:
Otherwise Django wants to create a new one since the content differs from that in `models.py`


hectcastro:
:+1:

Issue #510, opened 2015-06-29
When renaming a scenario, if element loses focus, name should be uneditable


caseypt:
Currently, the only way to cancel editing is to click on the div holding the name.

PR #619, opened 2015-07-27
Stop editing scenario when clicking outside

rajadain:
## Overview

Clicking outside the edit area when renaming scenarios was not canceling the edit operation. Now it does.

## Testing Instructions

 1. Open a scenario
 2. Start renaming a scenario, but don't make any changes to the name
 3. Click anywhere outside. The scenario tab should return to its non-editing state.
 4. Start renaming a scenario and make changes.
 5. Click anywhere outside. Your changes should be saved and the tab should return to its non-editing state.

## Demo

mmw-rename-scenario

Connects #510

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -499,7 +499,7 @@ var ScenariosCollection = Backbone.Collection.extend({
 
         // Bail early if the name actually didn't change.
         if (model.get('name') === newName) {
-            return;
+            return false;
         }
 
         var match = this.find(function(model) {
@@ -507,10 +507,13 @@ var ScenariosCollection = Backbone.Collection.extend({
         });
 
         if (match) {
+            window.alert("There is another scenario with the same name. " +
+                    "Please choose a unique name for this scenario."
+            );
             console.log('This name is already in use.');
-            return;
+            return false;
         } else if (model.get('name') !== newName) {
-            model.set('name', newName);
+            return model.set('name', newName);
         }
     },
 
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -326,7 +326,12 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
     },
 
     renameScenario: function() {
-        var self = this;
+        var self = this,
+            setScenarioName = function(name) {
+                if (!self.model.collection.updateScenarioName(self.model, name)) {
+                    self.render();
+                }
+            };
 
         this.ui.nameField.attr('contenteditable', true).focus();
 
@@ -343,16 +348,12 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
                 // Don't add line returns to the text.
                 e.preventDefault();
 
-                if (self.model.get('name') !== $(this).text() && $(this).text() !== '') {
-                    self.model.collection.updateScenarioName(self.model, $(this).text());
-                } else {
-                    self.render();
-                }
+                setScenarioName($(this).text());
             }
         });
 
         this.ui.nameField.on('blur', function() {
-            self.model.collection.updateScenarioName(self.model, $(this).text());
+            setScenarioName($(this).text());
         });
     },
 
--- a/src/mmw/js/src/modeling/tests.js
+++ b/src/mmw/js/src/modeling/tests.js
@@ -619,11 +619,29 @@ describe('Modeling', function() {
             });
 
             describe('#updateScenarioName', function() {
+                var realAlert, spyAlert;
+
+                // Swap out window.alert so that testing can complete
+                // automatically without the user being prompted that the
+                // scenario name must be unique.
+                before(function() {
+                    realAlert = window.alert;
+                });
+
+                after(function() {
+                    window.alert = realAlert;
+                });
+
+                beforeEach(function() {
+                    window.alert = spyAlert = sinon.spy();
+                });
+
                 it('trims whitespace from the provided new name', function() {
                     var collection = getTestScenarioCollection();
 
                     collection.updateScenarioName(collection.at(0), 'New Name       ');
 
+                    assert.isFalse(spyAlert.called);
                     assert.equal(collection.at(0).get('name'), 'New Name');
                 });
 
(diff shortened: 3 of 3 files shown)


src/mmw/js/src/modeling/views.js
```diff
@@ -342,9 +342,11 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
             if (keycode === ENTER_KEYCODE) {
                 // Don't add line returns to the text.
                 e.preventDefault();
+                
+                var new_name = $(this).text();
```
lliss: prefer camelCase variable names.

src/mmw/js/src/modeling/views.js
```diff
@@ -352,7 +354,13 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
         });
 
         this.ui.nameField.on('blur', function() {
-            self.model.collection.updateScenarioName(self.model, $(this).text());
+            var new_name = $(this).text();
```
lliss: Lines are nearly identical as above. Might be able to DRY it up.

rajadain:
Published new version a8ac7fb incorporating @lliss's suggestions

src/mmw/js/src/modeling/views.js
```diff
@@ -326,7 +326,14 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
     },
 
     renameScenario: function() {
-        var self = this;
+        var self = this,
+            setScenarioName = function(name) {
+                if (name && name !== '' && name !== self.model.get('name')) {
```
rajadain: Empty string is always _falsy_ so the second check is redundant. Fixing ...

rajadain: Also, we [check if the name is the same inside `updateScenarioName`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/models.js#L511) so checking it here in the `if` condition is redundant. Removing that too ...

hectcastro:
@azavea-bot rebuild

lewfish:
+1, tested.

rajadain:
@caseypt since you wrote the original issue, could you verify that this fix is what you were looking for?

caseypt:
@rajadain Sure! Will take a look soon.

src/mmw/js/src/modeling/views.js
```diff
@@ -326,7 +326,14 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
     },
 
     renameScenario: function() {
-        var self = this;
+        var self = this,
+            setScenarioName = function(name) {
```
caseypt: Does this need to be defined within `renameScenario`? If not, consider pulling it out.

rajadain: It doesn't _need_ to be, but since we're packaging local functionality for use twice locally (in response to https://github.com/WikiWatershed/model-my-watershed/pull/619#discussion_r35756091), I kept it local. I suppose it could be useful to have this outside, but I didn't have an immediate use case for that and wanted to contain my change to have a small footprint (keep it within `renameScenario`). If this is not satisfactory, then I can move it out.

caseypt: Okay cool. That's fine with me

caseypt:
If you enter a scenario name that already exists, like "Current Conditions", and then click away. It leaves edit mode and shows the name as "Current Conditions". But then if you click on one of the other scenario tabs, it will rerender the tabs and show that the scenario name wasn't changed (which is correct). If the validation fails, we should reset the value in the tab when clicking away.

I thought a prompt showed up if you used an invalid name of the scenario, but it looks like that isn't happening on staging either.

rajadain:
Hmm good catch. This is likely a problem with the name checking logic inside `updateScenarioName` (as evidenced by this also being on Staging) but I will investigate and fix as part of this PR.

rajadain:
Could use one final look through. I'm now reusing the logic available in `updateScenarioName` by making it always return a value. If the return value is falsy we manually render. I also added an alert box asking the user for unique names. Previously we only logged to the console.

We should probably make a card for adding a custom alert box.

src/mmw/js/src/modeling/models.js
```diff
@@ -517,10 +517,13 @@ var ScenariosCollection = Backbone.Collection.extend({
         });
 
         if (match) {
+            window.alert("There is another scenario with the same name. " +
```
lliss: Can we use one of the modals for alerts?

lliss:
@azavea-bot rebuild

src/mmw/js/src/modeling/models.js
lliss: Strike that just saw the note in your comment.

src/mmw/js/src/modeling/models.js
```diff
@@ -517,10 +517,13 @@ var ScenariosCollection = Backbone.Collection.extend({
...
+                    "Please choose a unique name for this scenario."
+            );
             console.log('This name is already in use.');
-            return;
+            return false;
         } else if (model.get('name') !== newName) {
-            model.set('name', newName);
+            return model.set('name', newName);
```
lliss: If I don't need the returned value on the other side, I personally try to return boolean rather than truthy and falsey values.

lliss:
+1

caseypt:
+1

src/mmw/js/src/modeling/models.js
rajadain: This should hopefully take care of the unlikely edge case when a `model.set` fails (for example, if there are [validation errors](http://backbonejs.org/docs/backbone.html#section-68))

rajadain:
@azavea-bot rebuild

rajadain:
@azavea-bot rebuild

rajadain:
The tests were stalling because of the `window.alert` that was added recently. Rebased on top of `develop`, and updated tests in 5b1c198 to handle that.

@lewfish could you please have a look at the tests and see if they are stylistically consistent?

rajadain:
Fixed lint and repushed d0db809

caseypt:
Cool fix.

rajadain:
Added #645 for adding an alert modal.

lewfish:
Taking a look.

lewfish:
+1, tested.

Issue #570, opened 2015-07-13
Load raster data from the S3 catalog


mmcfarland:
From Eugene

> bucket: com.azavea.datahub
prefix: catalog
layer name = “nlcd-tms”
layer zoom = 13

You should be able to: `S3RasterCatalog(“com.azavea.datahub”, “catalog”)`

PR #621, opened 2015-07-27
Use data from S3.

jamesmcclain:
This patch modifies the Scala code to use the NLCD data stored in S3 instead of the small, local test extent.  Since we do not yet have soil data, it passes the NLCD data through a function to produce deterministic but pseudo-random soil values.

Connect #570

**To build and run the service:**
First, put S3 credentials in environment or in ~/.aws/credentials, then type:
```
cd src/geop
./sbt 'project services' compile assembly
$SPARK_HOME/bin/spark-submit services/target/scala-2.10/mmw-services-assembly-0.0.1.jar
```

**To test the service:**
Post a valid geojson polygon (such as this one https://gist.github.com/jamesmcclain/42ec932910b1014d7137), multipolygon, or geometry collection of polygons and/or multipolygons to the endpoint:
```
% wget localhost:8107 --post-file=polygon.json -O -
--2015-07-27 18:45:00--  http://localhost:8107/
Resolving localhost (localhost)... 127.0.0.1
Connecting to localhost (localhost)|127.0.0.1|:8107... connected.
HTTP request sent, awaiting response... 200 OK
Length: 241 [application/json]
Saving to: ‘STDOUT’

[[[82, 4], 1408], [[52, 2], 7493], [[21, 1], 89141], [[42, 4], 4979], [[22, 1], 45787], [[23, 3], 13818], [[31, 4], 147], [[71, 3], 343], [[41, 1], 49996], [[43, 4], 5434], [[11, 0], 1290], [[90, 2], 7764], [[24, 2], 2399], [[81, 1], 11965]]
2015-07-27 18:45:07 (39.4 MB/s) - written to stdout [241/241]
```

### DIFF
--- a/src/geop/services/src/main/scala/com/azavea/mmw/GeopServiceActor.scala
+++ b/src/geop/services/src/main/scala/com/azavea/mmw/GeopServiceActor.scala
@@ -4,13 +4,7 @@ import akka.actor.Actor
 import spray.routing._
 import spray.http._
 
-/**
- * This loader will shortly be replaced by one that loads from S3
- * catalogues instead of small, local test files.  This entire object
- * will probably go away, with the possible exception of
- * jsonHistogram.
- */
-object Loader {
+object S3Loader {
   import geotrellis.raster.io.geotiff.reader._
   import geotrellis.raster._
   import geotrellis.raster.reproject._
@@ -22,6 +16,8 @@ object Loader {
   import geotrellis.spark.tiling._
   import geotrellis.raster.mosaic._
   import geotrellis.proj4._
+  import geotrellis.spark.io.s3._
+  import geotrellis.spark.op.local._
 
   import org.apache.spark._
   import org.apache.spark.{SparkConf, SparkContext}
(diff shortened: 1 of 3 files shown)


mmcfarland:
running the `sbt compile` command is resulting in the following:

```
Detected sbt version 0.13.8
Downloading sbt launcher 0.13.8:
  From  http://typesafe.artifactoryonline.com/typesafe/ivy-releases/org.scala-sbt/sbt-launch/0.13.8/sbt-launch.jar
    To  ./.lib/0.13.8/sbt-launch.jar
Download failed. Obtain the jar manually and place it at ./.lib/0.13.8/sbt-launch.jar
```

The URL provided indeed returns a 404, any ideas?

src/geop/services/src/main/scala/com/azavea/mmw/GeopServiceActor.scala
```diff
@@ -32,23 +28,57 @@ object Loader {
 
   import scala.util._
 
-  import DataFiles._
   import Histogram._
 
+  val bucket = "com.azavea.datahub"
+  val prefix = "catalog"
+  val layerName = "nlcd-tms"
+  val layerZoom = 13
```
mmcfarland: What's the implication of a fixed layerZoom level here and how did you come to the current value, 13?

mmcfarland:
I copied the jar manually to get past that issue, but now compilation actually fails.

```
GeopServiceActor.scala:60: value query is not a member of geotrellis.spark.io.s3.S3RasterCatalog
[error]     val nlcd = catalogue.query[SpatialKey]((layerName, layerZoom))
```

Is there a step I need to run to update the versions?  I did a `vagrant provision geoprocessing` prior to executing the code.

src/geop/services/src/main/scala/com/azavea/mmw/GeopServiceActor.scala
jamesmcclain: I got the bucket, prefix, layer name, and layer zoom from the information that you provided in the card.  My understanding is that 13 is the highest-resolution that we have?  We seem to be encountering issues with the speed of drawing these from S3 -- if that persists we may want to go to a lower resolution?

jamesmcclain:
Yeah, regarding the sbt jar: I normally copy that manually.  Sometimes it will download automatically and other times it will not (I think the sbt shell script that we have is older and pointing the wrong place).

mmcfarland:
In order to execute this code on the geop VM, I had to make the following changes:
* Manually install sbt-launcher.jar
* Had to manually edit Build.scala to change the version of GeoTrellis for 0.10.1-SNAPSHOT to M1
* `$SPARK_HOME` is not defined on the geop VM, I used the path `/usr/bin/spark-submit`
* Define and set the AWS key/secret variables

Once I did, the referenced polygon and similarly sized polygons all produced reasonable looking results.  Trying on a fairly large polygon (see attached) consistently killed the service.  Considering the caveats of setup - let's talk about how this fits into the plan for releasing an initial production version of this service.

jamesmcclain:
Yeah, I think that since the entire local spark cluster is running inside of the VM, the 3 gigabytes will be consumed very quickly if you attempt a larger area.

jamesmcclain:
If we are interested in merging this code, I can go ahead and change to the `-M1` release.  We will eventually want to move back to `-SNAPSHOT` in order to be to use future improvements, though.  (The name `-SNAPSHOT` is unfortunately overloaded, being the name of a release on sonatype and the default name of a freshly-built-from-master version of Geotrellis).

mmcfarland:
+1

Issue #612, opened 2015-07-24
Add boundary name to footer label


mmcfarland:
For any boundary selection (HUC/District) we now have a display name in the utfgrid that we can use instead of "Selected Area" for those types.

![screenshot from 2015-07-24 16 14 02](https://cloud.githubusercontent.com/assets/1014341/8883741/2b78886e-321f-11e5-957e-a8f6df78a899.png)

PR #627, opened 2015-07-28
Add boundary name to Analyze window

rajadain:
## Overview

Since boundary layers have names, we want to show those names instead of "Selected Area" in the Analyze window.

### Bonus

I also set "Delineate Watershed" dropdown to behave similarly, which sets the name to delineation type used. This is in a separate commit can be easily removed if it is not desired, since it was not asked for in the original issue.

## Testing Instructions

 1. Draw a shape on the map. Analyze window should show "Selected Area".
 2. Select a boundary on the map. Analyze window should show the selected boundary name.
 3. Delineate a watershed on the map. Analyze window should show the method used for delineation.

## Demo

mmw-analyze-window-label

Connects #612

### DIFF
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -67,6 +67,7 @@ var AnalyzeWindow = Marionette.LayoutView.extend({
             App: App,
             model: new coreModels.AreaOfInterestModel({
                 shape: this.model.get('area_of_interest'),
+                place: App.map.get('areaOfInterestName'),
                 can_go_back: true,
                 next_label: 'Model',
                 url: 'project'
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -213,9 +213,10 @@ var PlaceMarkerView = Marionette.ItemView.extend({
         'change:toolsEnabled': 'render'
     },
 
-    onItemClicked: function() {
+    onItemClicked: function(e) {
         var self = this,
             map = App.getLeafletMap(),
+            itemName = $(e.target).text(),
             revertLayer = clearAoiLayer();
 
         this.model.disableTools();
@@ -240,7 +241,7 @@ var PlaceMarkerView = Marionette.ItemView.extend({
                     max_radial_length: 0.25
                 });
 
-            addLayer(shape);
+            addLayer(shape, itemName);
             navigateToAnalyze();
         }).fail(function() {
             revertLayer();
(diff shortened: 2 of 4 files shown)


src/mmw/js/src/analyze/views.js
```diff
@@ -67,6 +67,7 @@ var AnalyzeWindow = Marionette.LayoutView.extend({
             App: App,
             model: new coreModels.AreaOfInterestModel({
                 shape: this.model.get('area_of_interest'),
+                place: App.map.get('areaOfInterestName'),
```
rajadain: This was easier than passing the `areaOfInterestName` to the [memoized `AnalyzeTaskModel`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/analyze/controllers.js#L42), since we already have a reference to `App` here.

src/mmw/js/src/draw/views.js
```diff
@@ -213,9 +213,10 @@ var PlaceMarkerView = Marionette.ItemView.extend({
         'change:toolsEnabled': 'render'
     },
 
-    onItemClicked: function() {
+    onItemClicked: function(e) {
         var self = this,
             map = App.getLeafletMap(),
+            itemName = e.target.innerHTML,
```
jamesmcclain: This is a minor thing, but is it true that the stuff that comes out of `e.target.innerHTML` is html whereas the location names (in this case `itemName`) are assumed to be plain text?  If so, it possible that at some point something changes so that '<b>Big City</b>' appears as the location?  (Once again, very minor.)

jamesmcclain:
Tested, works as advertised.  Very clean and efficient.  Other than the minor question about `e.target.innerHTML`, looks great.

:+1:

src/mmw/js/src/draw/views.js
rajadain: Good catch. We expect `itemName` to be a text value, not an HTML value, but since in this case the `innerHTML` is just a plain text value we are getting away with it. `innerText` is not a browser standard. The alternative would be to use jQuery using `$(e.target).text()` which may be a little less efficient because it causes a DOM lookup but the clarity may well be worth it.

rajadain:
Changed `e.target.innerHTML` to `$(e.target).text()` in 6489e9f

jamesmcclain:
Looks good to me.

kdeloach:
+1 Nice.

lliss:
+1

PR #637, opened 2015-07-29
Remove duplicate entry for `azavea.java`

hectcastro:
Having two inside `roles.txt` makes the dependency resolver go off all of the time.

[Review on Reviewable](https://reviewable.io/reviews/wikiwatershed/model-my-watershed/637)

### DIFF
--- a/deployment/ansible/roles.txt
+++ b/deployment/ansible/roles.txt
@@ -18,7 +18,6 @@ azavea.elasticsearch,0.2.0
 azavea.logstash,0.1.0
 azavea.kibana,0.3.1
 azavea.collectd,0.2.0
-azavea.java,0.1.1
 azavea.build-essential,0.1.0
 azavea.mapnik,0.1.0
 azavea.beaver,0.1.3


hectcastro:
I originally thought this was introduced in your pull request, but later realized it was in `develop`.

jamesmcclain:
Oh, okay.  Included a commit in https://github.com/WikiWatershed/model-my-watershed/pull/621 to remove the second entry, but I will remove it.

jamesmcclain:
Is the correct way to test this to re-provision the `geoprocessing` VM and confirm that it is still working?

hectcastro:
Before removing this, any attempt to `vagrant up` or `vagrant provision` against any virtual machine would trigger Ansible role dependency resolution. Afterwards, it should only happen if version numbers are tweaked in `roles.txt`.

jamesmcclain:
I see.  Well, I made this change and re-provisioned, and everything seems to be good.

Will remove this change from my pull request.

+1

lliss:
+1

PR #639, opened 2015-07-29
Test running models

lewfish:
Added a bunch of tests around running models.

Connects #393

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -360,6 +360,28 @@ var ScenarioModel = Backbone.Model.extend({
         }
     },
 
+    setResults: function() {
+        var rawServerResults = this.get('taskModel').get('result');
+        if (rawServerResults === "" || rawServerResults === null) {
+            this.get('results').setNullResults();
+        } else {
+            var serverResults = JSON.parse(rawServerResults);
+            this.get('results').forEach(function(resultModel) {
+                var resultName = resultModel.get('name');
+                if (serverResults[resultName]) {
+                    resultModel.set({
+                        'result': serverResults[resultName],
+                        'inputmod_hash': serverResults.inputmod_hash
+                    });
+                } else {
+                    console.log('Response is missing ' + resultName + '.');
+                }
+            });
+
+            this.set('census', serverResults.census);
+        }
+    },
+
     // Poll the taskModel for results and reset the results collection when done.
     // If not successful, the results collection is reset to be empty.
     getResults: function() {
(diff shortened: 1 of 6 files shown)


src/mmw/js/src/modeling/modificationConfigUtils.js
```diff
@@ -2,6 +2,15 @@
 
 var modificationConfig = require('./modificationConfig.json');
 
+// Useful for testing.
+function setConfig(config) {
```
rajadain: Consider renaming this and `resetConfig` to `setModificationConfig` and `resetModificationConfig` to be more representative. Currently we're using "settings" for runtime configuration in `clientSettings` but I can imagine `setConfig` being confusing down the line.

lewfish: I thought `setConfig` was unambiguous because it's in the `modificationConfigUtils` module.

rajadain: Ah, I missed that :expressionless: No change necessary.

src/mmw/js/src/modeling/tests.js
```diff
@@ -556,21 +693,92 @@ describe('Modeling', function() {
...
+                it('sets results to null on start failure', function(done) {
+                    var self = this;
+                    self.server.respondWith('POST', '/api/modeling/start/tr55/',
+                                            [ 400, // Make request fail.
+                                              { 'Content-Type': 'application/json' },
+                                              JSON.stringify(self.startResponse) ]);
+
+                    self.scenarioModel.getResults().startPromise.always(function() {
+                        assert(self.setNullResultsSpy.calledOnce, 'setNullResults should have been called once');
+                        assert(self.setResultsSpy.callCount === 0, 'setResults should not have been called');
```
rajadain: Consider changing assertion to `self.setResultsSpy.called === false` which is more in line with the test "setResults should not have been called"

rajadain: Same for https://github.com/WikiWatershed/model-my-watershed/pull/639/files#diff-2e0f78ab63ec412d23f471bbbbb3cf7eR766 and https://github.com/WikiWatershed/model-my-watershed/pull/639/files#diff-2e0f78ab63ec412d23f471bbbbb3cf7eR777

lewfish: I knew there had to be a better way. Taking it a step further, I can do `assert.isFalse(self.setResultsSpy.called)`.

rajadain: Even better!

src/mmw/js/src/modeling/tests.js
```diff
@@ -556,21 +693,92 @@ describe('Modeling', function() {
...
+                                              JSON.stringify(this.startResponse) ]);
+                    this.pollingResponse = mocks.polling.getJobSuccess;
+                    this.server.respondWith('GET', '/api/modeling/jobs/1/',
+                                            [ 200,
+                                              { 'Content-Type': 'application/json' },
+                                              this.pollingResponse ]);
+                });
+
+                function getResultsAssertions(self) {
+                    assert(inputModSpy.calledTwice, 'updateInputModHash should have been called twice');
```
rajadain: This went from being called once to twice because we are now initializing the model before each test and it is called once in `initialize` and once in `getResults`, correct?

lewfish: Yes.

src/mmw/js/src/modeling/models.js
```diff
@@ -370,6 +370,28 @@ var ScenarioModel = Backbone.Model.extend({
...
+                }
+            });
+
+            this.set('census', serverResults.census);
+        }
+    },
+
     // Poll the taskModel for results and reset the results collection when done.
     // If not successful, the results collection is reset to be empty.
     getResults: function() {
```
rajadain: This might be a big ask, since this function is so pervasive and may require touching many files, but —

Since the value returned by `getResults` is not a ResultsCollection but a promise, consider renaming to `fetchResults` instead.

rajadain:
+1 great work, very thorough tests, very useful refactors. I would recommend another set of eyes, but everything looks spectacular to me.

If possible, consider renaming `getResults` to `fetchResults` since that would be more accurate, and replacing the `.callCount === 0` with `!.called` as discussed above.

kdeloach:
:+1:  :trophy:

lliss:
This is a ton of work! Nice job.

mmcfarland:
Yeah, agree with the common sentiment here, good effort and result.

Issue #608, opened 2015-07-23
Ensure that modifications, BMPs are non-overlapping


jamesmcclain:
The Geotrellis/TR-55 work in https://github.com/WikiWatershed/model-my-watershed/pull/604 assumes that land use modifications and BMPs are non-overlapping (actually it is okay for a modification and a BMP to overlap, but not a modification and a modification or a BMP and a BMP to overlap).  

The client code should send a non-overlapping list of polygons.

caseypt:
Related to #255

jamesmcclain:
Ah, I was unaware that we already had an issue for this.  Should I close this issue?

caseypt:
I'll close #255, this is more detailed.

jamesmcclain:
Okay.

PR #641, opened 2015-07-29
 Generate and send non-overlapping modifications.

jamesmcclain:
The TR-55 code needs a set of modifications with the following property: "No point on the map is covered by more than one 'BMP' polygon, nor by more than one 'reclassification' polygon."  This patch introduces a function to compute such a set, applies it to the list of modifications, and sends the output to the back-end TR-55 code.

Connects #608

**To build**
   * This patch involves two new javascript dependencies, so you may need to do a `./script/bundle.sh --vendor` or similar.

**Expected behavior**
   * There are four modifications drawn in the area of interest shown below:
      1. A rectangle of *low-intensity residential*, a triangle of *high-intensity residential* that overlaps the rectangle
      2. A square of *row crop* to the right of those two
      3. A pentagon of *chaparral* (hard to see) on top of that
      4. A square of *porous paving* that intersects the *row crop*.

   * The non-overlapping version of this input is:
      1. A non-convex, six-sided square-minus-square of *row crop* with a pentagonal hole where the *chaparral* was drawn (both are land-use types, so the last-to-be-drawn overwrites the overlapping portion of the first-to-be-drawn).  The missing square portion comes from *point v* below.
      2. A rectangle of *low-intensity residential* with a triangular notch missing.
      3. A triangle of *high-intensity residential*
      4. A pentagon of *chaparral*
      5. A square of both *porous paving* (a BMP) and *row crop* (a land-use).
      6. A non-convex six-sided area of *porous paving* (the square mentioned in *point v* has been subtracted)

The idea is to enforce the rule that *"No point on the map should be covered by more than one BMP, nor more than one reclassification"*.

**Caveats**
   * Not so good in terms of theoretical efficiency (n^2 per TR-55 call).  However, as long as the person does not draw more than a couple hundred modifications, this should not be noticeable.

**To test**
   * Draw a collection of modifications on an area of interest (see first picture below)
   * With the chrome developer tools open (or whatever their equivalent is in the browser that you are using) examine the network traffic.
   * Click on the most recent `tr55/` listing (see the second picture below)
   * Check out the form data associated with that entry (see the third picture below)
   * Paste that into pro.jslint.com or similar and look at the value associated with the `modification_pieces` key (the block of text below)
   * See whether the output is consistent with the rule stated above.

screenshot from 2015-07-29 13 56 09

tr55

traffic

```
    "modification_pieces": [
        {
            "name": "landcover",
            "shape": {
                "type": "Feature",
                "properties": {},
                "geometry": {
                    "type": "Polygon",
                    "coordinates": [
                        [
                            [
                                -75.12451171875,
                                40.08542661727012
                            ],
                            [
                                -74.75784301757812,
                                40.11588965267845
                            ],
                            [
                                -74.74421825428698,
                                39.96808723793004
                            ],
                            [
                                -74.82513427734374,
                                39.95922773254976
                            ],
                            [
                                -74.81385482298451,
                                39.925749640339724
                            ],
                            [
                                -75.11077880859375,
                                39.87918107556866
                            ],
                            [
                                -75.12451171875,
                                40.08542661727012
                            ]
                        ],
                        [
                            [
                                -74.97894287109375,
                                40.07071544306934
                            ],
                            [
                                -75.04623413085938,
                                40.01394257474289
                            ],
                            [
                                -74.9871826171875,
                                39.95185892663005
                            ],
                            [
                                -74.90203857421875,
                                39.95185892663005
                            ],
                            [
                                -74.87182617187499,
                                40.02235635293741
                            ],
                            [
                                -74.97894287109375,
                                40.07071544306934
                            ]
                        ]
                    ]
                }
            },
            "value": "row_crop"
        },
        {
            "name": "landcover",
            "shape": {
                "type": "Feature",
                "properties": {},
                "geometry": {
                    "type": "Polygon",
                    "coordinates": [
                        [
                            [
                                -75.6132880326546,
                                39.9468414749074
                            ],
                            [
                                -75.77682495117188,
                                39.940277770390324
                            ],
                            [
                                -75.82626342773438,
                                40.069664523297774
                            ],
                            [
                                -75.28244018554688,
                                40.08542661727012
                            ],
                            [
                                -75.25222778320312,
                                39.961332959837826
                            ],
                            [
                                -75.49264941290215,
                                39.9516834167795
                            ],
                            [
                                -75.56533813476562,
                                40.01499435375046
                            ],
                            [
                                -75.6132880326546,
                                39.9468414749074
                            ]
                        ]
                    ]
                }
            },
            "value": "lir"
        },
        {
            "name": "landcover",
            "shape": {
                "type": "Feature",
                "properties": {},
                "geometry": {
                    "type": "Polygon",
                    "coordinates": [
                        [
                            [
                                -75.56533813476562,
                                40.01499435375046
                            ],
                            [
                                -75.3936767578125,
                                39.86547951378614
                            ],
                            [
                                -75.67794799804688,
                                39.854937988531276
                            ],
                            [
                                -75.56533813476562,
                                40.01499435375046
                            ]
                        ]
                    ]
                }
            },
            "value": "hir"
        },
        {
            "name": "landcover",
            "shape": {
                "geometry": {
                    "type": "Polygon",
                    "coordinates": [
                        [
                            [
                                -74.97894287109375,
                                40.07071544306934
                            ],
                            [
                                -75.04623413085938,
                                40.01394257474289
                            ],
                            [
                                -74.9871826171875,
                                39.95185892663005
                            ],
                            [
                                -74.90203857421875,
                                39.95185892663005
                            ],
                            [
                                -74.87182617187499,
                                40.02235635293741
                            ],
                            [
                                -74.97894287109375,
                                40.07071544306934
                            ]
                        ]
                    ]
                },
                "type": "Feature",
                "properties": {}
            },
            "value": "chaparral"
        },
        {
            "name": "both",
            "shape": {
                "type": "Feature",
                "properties": {},
                "geometry": {
                    "type": "Polygon",
                    "coordinates": [
                        [
                            [
                                -74.74421825428698,
                                39.96808723793004
                            ],
                            [
                                -74.74136352539062,
                                39.93711893299021
                            ],
                            [
                                -74.81385482298451,
                                39.925749640339724
                            ],
                            [
                                -74.82513427734374,
                                39.95922773254976
                            ],
                            [
                                -74.74421825428698,
                                39.96808723793004
                            ]
                        ]
                    ]
                }
            },
            "value": {
                "bmp": "porous_paving",
                "reclass": "row_crop"
            }
        },
        {
            "name": "conservation_practice",
            "shape": {
                "type": "Feature",
                "properties": {},
                "geometry": {
                    "type": "Polygon",
                    "coordinates": [
                        [
                            [
                                -74.74421825428698,
                                39.96808723793004
                            ],
                            [
                                -74.70977783203125,
                                39.97185812402586
                            ],
                            [
                                -74.70016479492186,
                                39.87707332113241
                            ],
                            [
                                -74.79354858398438,
                                39.86547951378614
                            ],
                            [
                                -74.81385482298451,
                                39.925749640339724
                            ],
                            [
                                -74.74136352539062,
                                39.93711893299021
                            ],
                            [
                                -74.74421825428698,
                                39.96808723793004
                            ]
                        ]
                    ]
                }
            },
            "value": "porous_paving"
        }
    ],
```

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -5,7 +5,10 @@ var Backbone = require('../../shim/backbone'),
     utils = require('../core/utils'),
     settings = require('../core/settings'),
     App = require('../app'),
-    coreModels = require('../core/models');
+    coreModels = require('../core/models'),
+    turfArea = require('turf-area'),
+    turfErase = require('turf-erase'),
+    turfIntersect = require('turf-intersect');
 
 var ModelPackageControlModel = Backbone.Model.extend({
     defaults: {
@@ -237,6 +240,123 @@ var ProjectModel = Backbone.Model.extend({
     }
 });
 
+/**
+ * A predicate for a filter function used in the _modifyModifications.
+ * Returns true if piece has a valid shape with non-zero area, and
+ * false otherwise.
+ */
+function validShape(piece) {
+    return (piece.shape !== undefined) && (turfArea(piece.shape) > 0.33);
+}
+
+/**
+ * This function takes a collection of modifications as drawn on
+ * screen and returns an array that has the following property: no
+ * point on the map is covered by more than one 'BMP' or more than one
+ * 'reclassification'.
+ */
+function _modifyModifications(rawModifications) {
+    var pieces = [],
+        reclass = 'landcover',
+        bmp = 'conservation_practice',
+        both = 'both',
+        n = rawModifications.size(),
+        n2 = n * n;
+
+    for (var i = 0; i < n; ++i) {
+        var rawModification = rawModifications.at(i),
+            newPiece = {
+                name: rawModification.get('name'),
+                shape: rawModification.get('shape'),
+                value: rawModification.get('value')
+            };
+
+        for (var j = 0; (j < pieces.length) && (newPiece.shape !== undefined) && (j < n2); ++j) {
+            var oldPiece = pieces[j];
+
+            /* If the new piece and the old piece are both BMPs or
+             * both are reclassifications, then simply subtract the
+             * shape of the new piece from that of the old piece to
+             * enforce the invariant. */
+            if (oldPiece.name === newPiece.name) {
+                try {
+                    oldPiece.shape = turfErase(oldPiece.shape, newPiece.shape);
+                } catch(e) {
+                    /* This "can only happen" if oldPiece.shape is
+                     * undefined (that is, the empty set), but that is
+                     * explicitly codified here. */
+                    oldPiece.shape = undefined;
+                }
+            }
+            /* If the new piece and the old piece are not both BMPs or
+             * not both reclassifications, then there are a number of
+             * possible scenarios: the old piece might be a BMP, might
+             * be a reclassification, or might represent an area where
+             * BMPs and reclassification overlap. */
+            else {
+                var newOldIntersection;
+
+                try {
+                    newOldIntersection = turfIntersect(oldPiece.shape, newPiece.shape);
+                } catch(e) {
+                    /* Once again, this "can only happen" if
+                     * oldPiece.shape is undefined, but make it
+                     * explicit. */
+                    oldPiece.shape = undefined;
+                    newOldIntersection = undefined;
+                }
+
+                /* New overlap pieces are born here. */
+                if ((newOldIntersection !== undefined) && (turfArea(newOldIntersection) > 0.33)) {
+                    var oldPieceShape = oldPiece.shape, // save a copy, need this for later
+                        overlapPiece = {
+                            name: both,
+                            shape: newOldIntersection,
+                            value: {}
+                        };
+
+                    /* compute overlap piece and add to array */
+                    if ((oldPiece.name === both) && (newPiece.name === reclass)) {
+                        overlapPiece.value.bmp = oldPiece.value.bmp;
+                        overlapPiece.value.reclass = newPiece.value;
+                    } else if ((oldPiece.name === both) && (newPiece.name === bmp)) {
+                        overlapPiece.value.bmp = newPiece.value;
+                        overlapPiece.value.reclass = oldPiece.value.reclass;
+                    } else if (oldPiece.name === reclass) {
+                        overlapPiece.value.bmp = newPiece.value;
+                        overlapPiece.value.reclass = oldPiece.value;
+                    } else {
+                        overlapPiece.value.bmp = oldPiece.value;
+                        overlapPiece.value.reclass = newPiece.value;
+                    }
+                    pieces.push(overlapPiece);
+
+                    /* remove the overlapping portion from both parents */
+                    try {
+                        oldPiece.shape = turfErase(oldPiece.shape, newPiece.shape);
+                    } catch(e) {
+                        /* Once again, should only happen if
+                         * oldPiece.shape is already undefined. */
+                        oldPiece.shape = undefined;
+                    }
+                    try {
+                        newPiece.shape = turfErase(newPiece.shape, oldPieceShape);
+                    } catch(e) {
+                        /* This can (should) never happen. */
+                    }
+                }
+            }
+        }
+
+        pieces.push(newPiece);
+        pieces = _.filter(pieces, validShape);
+    }
+
+    return pieces;
+}
+
+var modifyModifications = _.memoize(_modifyModifications, function(_, hash) {return hash;});
+
 var ModificationModel = coreModels.GeoModel.extend({
     defaults: _.extend({
             name: '',
@@ -396,6 +516,7 @@ var ScenarioModel = Backbone.Model.extend({
                     model_input: JSON.stringify({
                         inputs: self.get('inputs').toJSON(),
                         modifications: self.get('modifications').toJSON(),
+                        modification_pieces: modifyModifications(self.get('modifications'), self.get('modification_hash')),
                         area_of_interest: App.currProject.get('area_of_interest'),
                         census: self.get('census'),
                         inputmod_hash: self.get('inputmod_hash'),
--- a/src/mmw/package.json
+++ b/src/mmw/package.json
@@ -49,6 +49,8 @@
     "turf-bbox-polygon": "1.0.1",
     "turf-buffer": "1.0.4",
     "turf-destination": "1.2.1",
+    "turf-erase": "1.3.2",
+    "turf-intersect": "1.4.2",
     "turf-random": "1.0.2",
     "underscore": "1.8.3",
     "watchify": "3.1.0",
--- a/src/mmw/bundle.sh
+++ b/src/mmw/bundle.sh
@@ -105,7 +105,8 @@ CONCAT_VENDOR_CSS_COMMAND="cat \
 JS_DEPS=(jquery backbone backbone.marionette bootstrap bootstrap-select \
          leaflet leaflet-draw leaflet.locatecontrol leaflet-plugins lodash \
          underscore d3 nunjucks turf-area turf-bbox-polygon turf-buffer \
-         turf-destination turf-random zeroclipboard blueimp-md5 iframe-phone)
+         turf-destination turf-erase turf-intersect turf-random zeroclipboard \
+	 blueimp-md5 iframe-phone)
 BROWSERIFY_EXT=""
 BROWSERIFY_REQ=""
 for DEP in "${JS_DEPS[@]}"
--- a/src/mmw/npm-shrinkwrap.json
+++ b/src/mmw/npm-shrinkwrap.json
@@ -4041,6 +4041,46 @@
         }
       }
     },
+    "turf-erase": {
+        "version": "1.3.2",
+        "from": "turf-erase@",
+        "dependencies": {
+            "jsts": {
+                "version": "0.15.0",
+                "from": "https://registry.npmjs.org/jsts/-/jsts-0.15.0.tgz",
+                "resolved": "https://registry.npmjs.org/jsts/-/jsts-0.15.0.tgz",
+                "dependencies": {
+                    "javascript.util": {
+                        "version": "0.12.5",
+                        "from": "https://registry.npmjs.org/javascript.util/-/javascript.util-0.12.5.tgz",
+                        "resolved": "https://registry.npmjs.org/javascript.util/-/javascript.util-0.12.5.tgz"
+                    }
+                }
+            },
+            "turf-featurecollection": {
+                "version": "1.0.1",
+                "from": "turf-featurecollection@1.0.1"
+            }
+        }
+    },
+    "turf-intersect": {
+        "version": "1.4.2",
+        "from": "turf-intersect@",
+        "dependencies": {
+            "jsts": {
+                "version": "0.15.0",
+                "from": "https://registry.npmjs.org/jsts/-/jsts-0.15.0.tgz",
+                "resolved": "https://registry.npmjs.org/jsts/-/jsts-0.15.0.tgz",
+                "dependencies": {
+                    "javascript.util": {
+                        "version": "0.12.5",
+                        "from": "https://registry.npmjs.org/javascript.util/-/javascript.util-0.12.5.tgz",
+                        "resolved": "https://registry.npmjs.org/javascript.util/-/javascript.util-0.12.5.tgz"
+                    }
+                }
+            }
+        }
+    },
     "turf-random": {
       "version": "1.0.2",
       "from": "https://registry.npmjs.org/turf-random/-/turf-random-1.0.2.tgz",


src/mmw/package.json
```diff
@@ -49,6 +49,8 @@
     "turf-random": "1.0.2",
     "underscore": "1.8.3",
     "watchify": "3.1.0",
-    "zeroclipboard": "2.2.0"
+    "zeroclipboard": "2.2.0",
+    "turf-intersect": "1.4.2",
```
rajadain: Consider keeping the packages in `package.json` sorted alphabetically, since that is the behavior of `npm install <packagename> --save`. Thus, this would be: 

```json
    "turf-area": "1.1.1",
    "turf-buffer": "1.0.4",
    "turf-erase": "1.3.2",
    "turf-intersect": "1.4.2",
    "turf-random": "1.0.2",
```

jamesmcclain: This was actually generated by `npm install ... --save` :smile:

rajadain: Did you run it locally or via `npm.sh` in the VM? When I run it in the VM I get the sorted version (although with carets in the version numbers which should be removed):

```bash
$ ./scripts/npm.sh install turf-erase turf-intersect --save
+ ARGS='install turf-erase turf-intersect --save'
+ vagrant ssh app -c 'cd /opt/app && envdir /etc/mmw.d/env npm install turf-erase turf-intersect --save'
npm WARN package.json model-my-watershed@0.0.1 No description
npm WARN package.json model-my-watershed@0.0.1 No README data
npm WARN package.json iframe-phone@1.1.3 No repository field.
npm WARN package.json retina.js@1.3.0 No repository field.
turf-intersect@1.4.2 node_modules/turf-intersect
└── jsts@0.15.0 (javascript.util@0.12.5)

turf-erase@1.3.2 node_modules/turf-erase
├── turf-featurecollection@1.0.1
└── jsts@0.15.0 (javascript.util@0.12.5)
```

```shell
$ git diff
```

```diff
diff --git a/src/mmw/package.json b/src/mmw/package.json
index f7ace2d..3100bca 100644
--- a/src/mmw/package.json
+++ b/src/mmw/package.json
@@ -32,8 +32,8 @@
     "jstify": "0.9.0",
     "leaflet": "0.7.3",
     "leaflet-draw": "git://github.com/michaelguild13/Leaflet.draw#e2fe3a4",
-    "leaflet.locatecontrol": "0.43.0",
     "leaflet-plugins": "git://github.com/azavea/leaflet-plugins#feature/browserify",
+    "leaflet.locatecontrol": "0.43.0",
     "livereloadify": "2.0.0",
     "lodash": "3.6.0",
     "minifyify": "7.0.0",
@@ -46,11 +46,11 @@
     "testem": "0.5.15",
     "turf-area": "1.1.1",
     "turf-buffer": "1.0.4",
+    "turf-erase": "^1.3.2",
+    "turf-intersect": "^1.4.2",
     "turf-random": "1.0.2",
     "underscore": "1.8.3",
     "watchify": "3.1.0",
-    "zeroclipboard": "2.2.0",
-    "turf-intersect": "1.4.2",
-    "turf-erase": "1.3.2"
+    "zeroclipboard": "2.2.0"
   }
 }
```

jamesmcclain: Yeah, I just ran it locally.  It is no biggie, I can put those in sorted order.

rajadain:
Please add `turf-intersect` and `turf-erase` to the `vendor` bundle in `bundle.sh`

src/mmw/package.json
rajadain: Thanks. What version of `npm` do you have locally? The VM has 2.1.17

jamesmcclain: 1.4.3, it look like.

rajadain: Upgrading it would certainly help, but for the sake of consistency I would recommend using `npm.sh` if possible.

kdeloach:
Clever algorithm, but consider clipping features as they are drawn on the map to get better visual feedback?

jamesmcclain:
@kdeloach This was my initial approach but there is a bit of a problem.  I want to preserve the list of modifications that the user supplies.  One can produce a valid set of polygons very easily if only insertions are allowed, but allowing deletions complicates matters so much that you might as well just recompute the whole thing every time unless you are dealing with a thousand vertices (in C/C++/Fortran, that is -- I don't know what the threshold is in Javascript).

For that reason, I just left the original modification list alone.  If we want to later, I suppose we can display the non-overlapping set to the user, but we will still need to keep the original list around.

kdeloach:
One solution may be to write a function that takes a set of existing modifications and a new modification and returns a set of non-intersecting modifications. That should allow for insertions/deletions, but then again, I have not been hands-on with this codebase for some time.

rajadain:
So there's a problem in the "original list of modifications". Inside an AoI, draw two inverted triangles that form a hexagon in their intersection, one with Landcover of type Commercial and the other with Landcover of type Chaparral, like so:

image

Regardless of drawing order, the `modification_pieces` are calculated as:

```json
    "modification_pieces": [
        {
            "name": "landcover",
            "shape": {
                "geometry": {
                    "coordinates": [
                        [
                            [
                                [
                                    -75.20014504621149,
                                    39.95041267683133
                                ],
                                [
                                    -75.1845932006836,
                                    39.96712200067156
                                ],
                                [
                                    -75.17612383312957,
                                    39.95071161518519
                                ],
                                [
                                    -75.20014504621149,
                                    39.95041267683133
                                ]
                            ]
                        ],
                        [
                            [
                                [
                                    -75.20353883087786,
                                    39.90429174371303
                                ],
                                [
                                    -75.23677825927734,
                                    39.9110529610132
                                ],
                                [
                                    -75.21937222326301,
                                    39.92975447730698
                                ],
                                [
                                    -75.20353883087786,
                                    39.90429174371303
                                ]
                            ]
                        ],
                        [
                            [
                                [
                                    -75.15491571927986,
                                    39.90961842593674
                                ],
                                [
                                    -75.1461410522461,
                                    39.892616487170955
                                ],
                                [
                                    -75.16409645207723,
                                    39.89626878703937
                                ],
                                [
                                    -75.15491571927986,
                                    39.90961842593674
                                ]
                            ]
                        ]
                    ],
                    "type": "MultiPolygon"
                },
                "properties": {},
                "type": "Feature"
            },
            "value": "commercial"
        },
        {
            "name": "landcover",
            "shape": {
                "geometry": {
                    "coordinates": [
                        [
                            [
                                -75.23197174072266,
                                39.9500166010947
                            ],
                            [
                                -75.12622833251953,
                                39.951332552969085
                            ],
                            [
                                -75.18218994140625,
                                39.86995917164515
                            ],
                            [
                                -75.23197174072266,
                                39.9500166010947
                            ]
                        ]
                    ],
                    "type": "Polygon"
                },
                "properties": {},
                "type": "Feature"
            },
            "value": "chaparral"
        }
    ],
```

i.e. three corner cutouts for Commercial, and the entire triangle for Chaparral. This happens correctly when Chaparral is drawn after Commercial, but incorrectly when Chaparral is drawn _before_ Commercial.

I'm pretty sure this is happening because the order of modifications is no longer the order in which the user drew them, as of #567. See https://github.com/WikiWatershed/model-my-watershed/commit/220a6e82f4cfbe5d67d6b541786a3d59d2286974, specifically this `comparator` that was introduced: [modeling/models.js#L259-265](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/models.js#L259-L265). At the time, the actual ordering of the modifications was meaningless, thus it was alright. Now, however, we want to add meaning to the ordering, and thus that code must be changed.

Since I am familiar with that code, I will make the changes necessary and open a PR that should restore user ordering of modifications. I imagine this card is blocked until that is merged. I will reference it here when ready.

jamesmcclain:
So the motivation for the comparator was to facilitate caching (I assume to get a consistent hash of the collection)?

rajadain:
@jamesmcclain Exactly. I'm going to move the sort into the hashing function, rather than maintaining a sorted Collection throughout. Furthermore, I'm only going to sort inputs, not modifications, which is what I meant to do all along, it's just that they use the same Collection type under the hood.

jamesmcclain:
@kdeloach If I understand what you are suggesting, that is a viable approach.  Handling insertions is easy.  However, since handling deletions requires a rebuilt (barring lots of effort), it is probably easier to just use the same code for both insertion and deletion.

kdeloach:
Would deletions require a rebuild if drawing a shape always deletes the modification area underneath it? I'm interested in knowing what makes deletions difficult, so I can get a better grasp on the data model.

Something else to think about... It may be confusing for people to get 2 different results by drawing the same modifications area in a different order. I haven't actually confirmed this behavior but it's something to consider.

jamesmcclain:
> Would deletions require a rebuild if drawing a shape always deletes the modification area underneath 
> it? I'm interested in knowing what makes deletions difficult, so I can get a better grasp on the data 
> model.

No, if drawing an overlap deletes everything underneath it, then everything is much nicer.

> Something else to think about... It may be confusing for people to get 2 different results by drawing 
> the same modifications area in a different order. I haven't actually confirmed this behavior but it's 
> something to consider.

Agreed, some kind of visual indication of what is happening is definitely needed.  However, it does make  sense for the order to matter (just like splashing blue paint on the wall then red will produce a different result from splashing red, then blue).

caseypt:
In addition to this (and in a different issue), I think we should also prevent people from drawing overlapping polygons. We could look to PSI for an example of how to implement this.

jamesmcclain:
@caseypt As I understand it, we need to be able to support BMPs and reclassifications overlapping since changing both of those at a given spot does make sense.

mmcfarland:
I think there's merit in the approach of leaving the unaltered/unclipped polygons on the map for display.  It does make it easier to handle deletions, which is a nice benefit, but it also imposes less restriction on  non-precise drawing.  Reclassifying Forest to Agriculture, and then applying an agricultural BMP on that reclassified area also makes sense, so I think we should keep that overlapping allowed.  

Probably the visual feedback that is required is more textual/conceptual than graphical - when we notice overlapping features that ultimately will be clipped for computation, we should give them a non-disruptive message like "Looks like you've added X over Y, in those sections Y will be applied".

I also think we may need to update the modification list/popup to display the drawn area and the "effective area" to further indicate that a portion of the feature may not be being applied to the calculation.

I'm happy to defer those two requirements to further issues, but I think starting visually simple and then iterating on more complex behavior is a good way to go for now.

mmcfarland:
Other additional nice features that we could add would be to edit existing shapes (which would be another reason to keep the original displayed) and allowing to change the z-order to certain overlaps get prioritized.

rajadain:
@jamesmcclain thanks for reviewing #643, I've just merged it in. I rebased your branch on top of that before merging it in and found that the order of pieces generated seems to be reversed. For example, when Chaparral is drawn on top of Commercial, Chaparral should have one large triangle and Commercial three small ones, but if you look at the values the reverse is true:

image

```json
    "modification_pieces": [
        {
            "name": "landcover",
            "shape": {
                "geometry": {
                    "coordinates": [
                        [
                            [
                                [
                                    -75.18733076087285,
                                    39.960200719452025
                                ],
                                [
                                    -75.17635345458983,
                                    39.97790935965889
                                ],
                                [
                                    -75.1643393622671,
                                    39.96055017690412
                                ],
                                [
                                    -75.18733076087285,
                                    39.960200719452025
                                ]
                            ]
                        ],
                        [
                            [
                                [
                                    -75.1822700320134,
                                    39.8923961424834
                                ],
                                [
                                    -75.2292251586914,
                                    39.892616487170955
                                ],
                                [
                                    -75.2083811863427,
                                    39.926242077756505
                                ],
                                [
                                    -75.1822700320134,
                                    39.8923961424834
                                ]
                            ]
                        ],
                        [
                            [
                                [
                                    -75.1365271727653,
                                    39.92036429623975
                                ],
                                [
                                    -75.11695861816406,
                                    39.892089657886366
                                ],
                                [
                                    -75.15277861591436,
                                    39.89225774914223
                                ],
                                [
                                    -75.1365271727653,
                                    39.92036429623975
                                ]
                            ]
                        ]
                    ],
                    "type": "MultiPolygon"
                },
                "properties": {},
                "type": "Feature"
            },
            "value": "chaparral"
        },
        {
            "name": "landcover",
            "shape": {
                "geometry": {
                    "coordinates": [
                        [
                            [
                                -75.2340316772461,
                                39.959490889505346
                            ],
                            [
                                -75.11283874511719,
                                39.961332959837826
                            ],
                            [
                                -75.16536712646484,
                                39.87048617098581
                            ],
                            [
                                -75.2340316772461,
                                39.959490889505346
                            ]
                        ]
                    ],
                    "type": "Polygon"
                },
                "properties": {},
                "type": "Feature"
            },
            "value": "commercial"
        }
    ],
```

When we try the other way, i.e. Commercial is drawn atop Chaparral, the pieces are three of Commercial and one of Chaparral, we still get reversed results:

image

```json
    "modification_pieces": [
        {
            "name": "landcover",
            "shape": {
                "geometry": {
                    "coordinates": [
                        [
                            [
                                [
                                    -75.24124145507812,
                                    39.946068593571304
                                ],
                                [
                                    -75.19786115112917,
                                    39.94906545456428
                                ],
                                [
                                    -75.22153756957772,
                                    39.91865062886813
                                ],
                                [
                                    -75.24124145507812,
                                    39.946068593571304
                                ]
                            ]
                        ],
                        [
                            [
                                [
                                    -75.15781109657789,
                                    39.951832250444546
                                ],
                                [
                                    -75.12313842773438,
                                    39.95422755797632
                                ],
                                [
                                    -75.13868657416948,
                                    39.9304440667295
                                ],
                                [
                                    -75.15781109657789,
                                    39.951832250444546
                                ]
                            ]
                        ],
                        [
                            [
                                [
                                    -75.16316259420245,
                                    39.89300389914165
                                ],
                                [
                                    -75.18218994140625,
                                    39.863898388233295
                                ],
                                [
                                    -75.2017211787621,
                                    39.89107611297296
                                ],
                                [
                                    -75.16316259420245,
                                    39.89300389914165
                                ]
                            ]
                        ]
                    ],
                    "type": "MultiPolygon"
                },
                "properties": {},
                "type": "Feature"
            },
            "value": "commercial"
        },
        {
            "name": "landcover",
            "shape": {
                "geometry": {
                    "coordinates": [
                        [
                            [
                                -75.17807006835938,
                                39.974489161874125
                            ],
                            [
                                -75.24467468261719,
                                39.88892859714545
                            ],
                            [
                                -75.1076889038086,
                                39.89577737784395
                            ],
                            [
                                -75.17807006835938,
                                39.974489161874125
                            ]
                        ]
                    ],
                    "type": "Polygon"
                },
                "properties": {},
                "type": "Feature"
            },
            "value": "chaparral"
        }
    ],
```

Please rebase your work on top of `develop` and see if you can reproduce and fix this. Thanks.

jamesmcclain:
@rajadain Yeah, you are probably right.  I reversed the order of the inner loop trying to get things in the right order yesterday, but I think it was being caused by the sorting of the collection.  I'll change that and see if it fixes it.

jamesmcclain:
@mmcfarland You asked a question in a line comment which has been obscured by later changes.  Regarding the try/catch block around the computation of the intersection of the new and old pieces: this was out of an abundance of caution.

`newPiece` is always valid due to the stopping condition on the loop.  `oldPiece` may have been reduced to nothing by this point, so I could have done a check to see if it was undefined, but I thought it was more general to handle an exception.  If the exception is raised, then `newOldIntersection` is left undefined, which is equivalent to an empty intersection (which is desired).

rajadain:
+1, my tests show predictable results, no infinite looping, and correct ordering of modification pieces.

jamesmcclain:
@rajadain Many thanks for the thorough review.

src/mmw/js/src/modeling/models.js
```diff
@@ -237,6 +240,120 @@ var ProjectModel = Backbone.Model.extend({
...
+                    }
+                }
+            }
+        }
+
+        /* add remenants of new piece */
+        pieces.push(newPiece);
+
+        /* remove pieces that did not survive */
+        pieces = _.filter(pieces, function(p) {
```
rajadain: This anonymous function is causing lint errors and the build to fail:

```

js/src/modeling/models.js: line 348, col 10, Don't make functions within a loop.

1 error
```

Consider moving this out of the loop and defining at the beginning with all other local variables.

jamesmcclain:
@azaveaci rebuild

src/mmw/js/src/modeling/models.js
jamesmcclain: Thanks, I think Jenkins should rebuild it soon.

rajadain: I think you'll have to move this function and then rebuild it in Jenkins for the build to be greenlit

Issue #624, opened 2015-07-28
Model output tabs are reversed


mmcfarland:
Runoff should remain the initial tab on the output window, currently Water Quality is.

![screenshot from 2015-07-28 12 37 17](https://cloud.githubusercontent.com/assets/1014341/8937298/7ce54058-3525-11e5-8c89-d0ad40e0d815.png)

rajadain:
Modifications also seem to be reversed:

![image](https://cloud.githubusercontent.com/assets/1430060/8961605/35bd6fb0-35e4-11e5-8ff4-abf4ed63cbde.png)

I wonder if it has anything to do with the re-ordering of internal representation done for #548

PR #643, opened 2015-07-30
Preserve original ordering of Collections

rajadain:
For the purpose of producing consistent MD5 hashes for caching, #567
added code for keeping many Collections in an internally sorted order
in 220a6e8. The actual order (usually the order in which they were added
to the collection) was not meaningful at the time.

Now, however, we are solving problems which do require us to respect the
order in which models are added to the collection, e.g. #641. Thus, we
undo automatic sorting and also remove tests for it. We adjust some more
tests to respect the order of models within a collection, but still test
that models with differently sorted keys are still evaluated the same.

Since Collections are stringified to Arrays and not Objects in JSON, the
sorting there should be preserved automatically, without us having to
sort each Collection. Thus, this change should not affect the stability
of our caching approach.

Connects #624 
Connects #641

### DIFF
--- a/src/mmw/js/src/core/utils.js
+++ b/src/mmw/js/src/core/utils.js
@@ -67,9 +67,8 @@ var utils = {
 
                 return sorted;
             },
-            // Sort the collection by given sortKey and convert
-            // each model to a sorted array
-            sortedCollection = collection.sort().map(function(model) {
+            // Convert each model to a sorted array
+            sortedCollection = collection.map(function(model) {
                 return objectToSortedArray(model.toJSON());
             });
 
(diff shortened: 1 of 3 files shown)


mmcfarland:
+1 Correct order is restored

mmcfarland:
You'll want to add a Connects #624 (if this comment doesn't do it)

mmcfarland:
Sorry - got confused about the issue you trying to resolve (just read the thread on #641) but I think this PR actually resolves #624 as well, for a similar reason.

jamesmcclain:
I just want to confirm that the `modification_hash` is still behaving correctly: is it supposed to give the same hash when the collection contains two identical modifications regardless of order?

jamesmcclain:
I understand now, +1

Issue #635, opened 2015-07-29
Fix rounding issues in microsite 


mmcfarland:
In some combinations, the total amount of R + ET + Inf does not equal the amount of precipitation that fell.  We need to generate the model results with more precision so that these rounding errors aren't displayed.

Issue #541, opened 2015-07-02
Investigate ET/R/Infl values for microsite


mmcfarland:
There were numerous comments from teachers indicating that the numbers provided by the microsite were incorrect or inconsistent.  This is partly do to misconceptions (ET is fixed a specific rate) and rounding - but there was some concern that the values were incorrect.  

* Fix rounding (try increasing precision from model output)
* Check values against previous implementation

I'll update the issue with specific comments for specific combinations if I can dig them up.

jamesmcclain:
Did they say what in particular they were having trouble with?

   * If scenarios with less than two inches of precipitation are the issue, it may be because the Pitt Small Storm Hydrology model is used in that regime instead of the regular TR-55.
   * Are there some use/soil pairs that are suspicious and not others?  If so, it may be incorrect curve numbers in the lookup table.  If the majority of scenarios are correct when there is > 2 inches of rain, then it the problem is unlikely to be the actual model implementation.

rajadain:
It should also be noted that the values for LIR, HIR, Commercial, Forest, and Turf Grass will be different between the old flash version and the new Micro Site, because of reasons specified in #185:

>**Incongruous Land Code Interpretation**
>
>The Flash app interpreted the `land` values as follows:
>
>```actionscript
>	var landCodes = {
>		21: LandCovers.LOW_INTENSITY_RESIDENTIAL,
>		22: LandCovers.HIGH_INTENSITY_RESIDENTIAL,
>		23: LandCovers.COMMERCIAL_INDUSTRIAL,
>		43: LandCovers.FOREST,
>		85: LandCovers.TURF_GRASS,
>		81: LandCovers.XYZ,
>		71: LandCovers.GRASSLAND,
>		82: LandCovers.ROW_CROPS,
>		91: LandCovers.WETLANDS
>	}
>```
>
>But the official definition of these codes is different: http://www.mrlc.gov/nlcd11_leg.php
>
>This official description is the one used by TR-55: https://github.com/WikiWatershed/tr-55/blob/develop/tr55/tables.py#L387-L414, and also by the new mini-app:
>
>```javascript
>    landMap: {
>        '21': 'turfGrass',
>        '22': 'lir',
>        '23': 'hir',
>        '24': 'commercial',
>        '41': 'forest',
>        '71': 'grassland',
>        '81': 'pasture',
>        '82': 'rowCrops',
>        // TODO The following are missing NLCD identifiers.
[...]

PR #653, opened 2015-07-31
Update mini-app data.

jamesmcclain:
The `model.csv` table now has many more significant digits.  In order to accommodate this, numbers must be rounded before they can be displayed (because the display only accepts one digit after the decimal point).

**To Test**
   1. `./scripts/bundle.sh --debug --vendor --tests`
   2. `./scripts/manage.py collectstatic`
   3. Play with the mini-app and see if the values look a little more reasonable.

Connects #635


rajadain:
Cases when a value (such as ET) is 0", the bar chart should reflect that. Currently, even when ET is shown to have 0", the bar chart has some small value:

image

mmcfarland:
Generally, where reasonable, we should be matching the behavior (and results) of the existing app:

http://www.stroudcenter.org/mmw/mini/

rajadain:
Here's the equivalent screenshot from the existing app:

image

While the numbers are different (as discussed in #541), and they don't add up to Precipitation value (as fixed in this PR), there is correspondence between numbers in the illustration and values in the bar chart.

jamesmcclain:
Okay, I have made that change.

Just to point out: there are now more significant digits in the table, but they are not used anywhere in the mini-app!

Also: this does not solve the problem of the totals not summing to the precipitation, because that is impossible (in general).

rajadain:
+1 great work! Tested that the numbers make sense and the bar chart correctly reflects labels.

jamesmcclain:
Thanks for the review.  I think I'll wait until we've had a discussion about how to index the mini-app csv before merging this.

mmcfarland:
Can you elaborate on why we are unable to get the component totals to sum up to the total precipitation?  I was under the impression that the was the original behavior.

jamesmcclain:
I guess formally the answer is something like this:

None of the precipitation values have any significant digits beyond the 1/10 th position.  If `round` rounds a value to the 1/10th position then wanting the sum of the rounded `runoff`, `et`, and `inf` to always sum to the `precipitation` is equivalent to wanting: precip = `round`(precip) = `round`(runoff + et +inf) = `round`(runoff) + `round`(et) + `round`(inf),  which is equivalent to wanting `round` to be a linear function which it isn't (if you graph it, it would be a step function).

So in general, the sum of the rounded values will not equal the sum of the unrounded values.  They might just have to be tweaked by hand (if that is allowed).

mmcfarland:
If that's the case, then we should evaluate the original implementation and see how they handle it.  I believe that app always has matching totals, and so should this one.

rajadain:
The original app doesn't always have matching totals, as can be seen in this screenshot:

image

2.0" of precipitation is broken down into 0" of evapotranspiration, 0.4" of runoff, and 1.5" of infiltration, which totals to 1.9" instead of 2.0".

jamesmcclain:
Another issue that we have is the infiltration decreasing as the precipitation goes from 2" to the next increment above that on built types such as *high intensity residential* with soil type C or D.  That is caused by the runoff jumping upward as we cross from using the Pitt SSH model at 2" and below to the regular TR-55 calculation at greater than 2".

rajadain:
+1, numbers work well and app behaves as expected.

I think this hash will cause problems in the future:

```javascript
    landMap: {
        'urban_grass': ['turfGrass'],
        'tall_grass_prairie': ['tallGrass'],
        'li_residential': ['lir'],
        'hi_residential': ['hir'],
        'commercial': ['commercial'],
        'desert': ['desert'],
        'deciduous_forest': ['forest'],
        'chaparral': ['chaparral'],
        'grassland': ['grassland'],
        'pasture': ['pasture'],
        'short_grass_prairie': ['shortGrass'],
        'row_crop': ['rowCrops']
    },
```

Before, we were translating from numeric NLCD encoding to alphabetic labels, used in CSS and JavaScript, so it was okay. Now we're converting one alphabetic labeling to another, which adds more noise than signal. This will get worse when we start to actually process the TR-55 results in the back end, since we send the JavaScript style labeling in our API calls which will have to be translated back to Python style labeling there before processing.

Instead of this dual reversal, we should either switch the styling in the front-end to use Pythonic labels, or the labeling in the back-end to use JavaScripty labels. I think that changing the back-end would be easier, since the front-end labels are spread across the app in multiple places. However, we have made a policy decision in #237 to prefer Pythonic labeling in the front-end as well, so I think that would be a a more stylistically fitting choice.

I don't think this work should be done as part of this PR. I'm going to create a new issue for this.

Issue #625, opened 2015-07-28
Add 1km² "stamp" drawing tool


mmcfarland:
Allow the user to place a fixed-size (1km²) square as an Area of Interest selector.  Ideally, we could show the square size as it's being placed, but I suspect it will be much easier to place a point and generate the polygon similarly to how we're mocking the delineate watershed feature.

caseypt:
See this project for a potential implementation: http://apps.npr.org/moore-oklahoma-tornado-damage/

mmcfarland:
We are going to change the "Draw Area" button to a "Draw" dropdown that includes both Draw Area and a new option for this "Draw Square Km" tool.  This is similar to the Delineate dropdown.

PR #656, opened 2015-08-03
Feature/lol/one km stamp

lliss:
Connects #625 

When the user select the one square km drawing tool, they are given a point to
drop on the map. This acts as the center of the square. We use turf
destination, which uses the Haversine formula to calculate four points around
the center at a distance of one half Km with the bearings -180, -90, 0, 90.
This is then used to create a polygon and start the analysis.

To test:

* Run the npm install shell script.
* Bundle with the vendor flag.
* Locate the "1 Km square" button and click it.
* Click anywhere on the map.
* Check to make sure the corners of the box actually are about 1 km apart.


mmcfarland:
Should the new vendor deps also be added to the vendor bundle?

src/mmw/js/src/draw/views.js
```diff
@@ -194,6 +198,51 @@ var DrawAreaView = Marionette.ItemView.extend({
...
+                // points. Then map each array of two to a single point by
+                // taking the lat from one and lng from the other.
+                swNe = _.map(_.toArray(_.groupBy(halfKmbufferPoints, function(p, i) {
+                    // split the array of four in half.
+                    return i < 2;
+                })), function(pointGroup) {
+                    return L.latLng(pointGroup[0].lat, pointGroup[1].lng);
+                }),
+                bounds = L.latLngBounds(swNe),
+                bbox = [
```
mmcfarland: Previous comment may have been dropped in an update, but I think you can use `bounds.toBboxString()` here.

lliss:
@azavea-bot rebuild

lliss:
@azavea-bot rebuild

lliss:
@mmcfarland Dug up your earlier comment too and simplified the geoJSON creation from the lan/lng.

lliss:
Will rebase after review.

src/mmw/js/src/draw/views.js
```diff
@@ -194,6 +198,40 @@ var DrawAreaView = Marionette.ItemView.extend({
...
+                    return i < 2;
+                })), function(pointGroup) {
+                    return L.latLng(pointGroup[0].lat, pointGroup[1].lng);
+                }),
+                bounds = L.latLngBounds(swNe),
+                box = turfBboxPolygon(bounds.toBBoxString().split(','));
+
+            addLayer(box, '1 Square Km');
+            navigateToAnalyze();
+        }).fail(function() {
```
rajadain: Should this `fail` execute when `placeMarker` fails, or when `function(latlng)` fails? Since we're using `.then` instead of `.done`, currently it will execute in the latter case, but I'm not sure if it will in the former.

mmcfarland:
Nice updates, results look good and accurate. +1

src/mmw/js/src/draw/views.js
lliss: I think we only care if we couldn't get a lat/lng from the marker for some reason.

rajadain: Great! I just learnt the differences between `.done` and `.then`, good to know we're using them right!

rajadain:
+1 tested, works really well.

PR #658, opened 2015-08-03
Don't let shrinkwrap pull from cache.

lliss:
We had accidentally updated the npm shrinkwrap file so that some dependencies
were being loaded from a local cache. This reverts those changes.

### DIFF
--- a/src/mmw/npm-shrinkwrap.json
+++ b/src/mmw/npm-shrinkwrap.json
@@ -41,7 +41,7 @@
     },
     "bootstrap-select": {
       "version": "1.6.4",
-      "from": "../../tmp/npm-31030-7f32f0c4/1436971651298-0.12548802117817104/f1755efa102a41e43fc367ba36ce905b8f2b90e1",
+      "from": "bootstrap-select@git://github.com/azavea/bootstrap-select",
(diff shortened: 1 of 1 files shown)


lliss:
@rajadain This is a straight reversion to what was there before those lines were changed so I think it's probably safe.

rajadain:
@lliss I've been trying to read up on what's the difference between `from` and `resolved` fields in `npm-shrinkwrap.json`, but it is pretty poorly documented. If this makes npm run something like

```
npm install leaflet-draw@0.2.3
```

instead of

```
npm install git://github.com/michaelguild13/Leaflet.draw.git#e2fe3a4
```

then that could be problematic.

Can you tell me which revision you are using as reference? I can check and confirm the install there.

lliss:
Updated.

caseypt:
+1, though these changes may get blown away the next time `shrinkwrap` is run. I'm not sure why, but it seems like the `from` and `resolved` keys switch back and forth for some dependencies each time `shrinkwrap` is run.

rajadain:
+1.

There are open issues regarding this behavior in the NPM repo, e.g. https://github.com/npm/npm/issues/3581.

We just have to be mindful about staging and committing only the relevant lines in `npm-shrinkwrap.json` when adding / updating dependencies and ensure that unrelated packages are not affected.

Issue #636, opened 2015-07-29
Improve contrast for boundary styles so they are more visible on all basemaps


mmcfarland:
The slim black border is difficult to distinguish, especially on imagery maps.  These style changes should be done before #620 is complete so we don't have to mess with the cache.

<img width="1239" alt="screen shot 2015-07-29 at 9 17 40 am" src="https://cloud.githubusercontent.com/assets/1014341/8960011/d783ffde-35db-11e5-8f9a-bc6533f31786.png">

PR #660, opened 2015-08-03
Update boundary layer style

caseypt:
* Update boundary color to provide more contrast across multiple
basemaps.
* Remove unused fill style.

**To Test:**

- Reload the tiler VM or run `./scripts/debugtiler.sh`.
- Load the app and select one of the tile layers.
- See how the tiles look across different basemaps.

Connects to #636

### DIFF
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -147,15 +147,15 @@ var MapView = Marionette.ItemView.extend({
         addLocateMeButton(map, maxZoom, maxAge);
 
         var baseLayers = _.mapObject(settings.get('base_layers'), function(layerData) {
-            if (layerData.googleType) {
-                return new L.Google(layerData.googleType);
-            } else {
-                return new L.TileLayer(layerData.url, {
-                    attribution: layerData.attribution || '',
-                    maxZoom: layerData.maxZoom || 18
-                });
-            }
-        }),
+                if (layerData.googleType) {
+                    return new L.Google(layerData.googleType);
+                } else {
+                    return new L.TileLayer(layerData.url, {
+                        attribution: layerData.attribution || '',
+                        maxZoom: layerData.maxZoom || 18
+                    });
+                }
+            }),
             defaultLayerName = _.findKey(settings.get('base_layers'), function(layerData) {
                 return layerData.default;
             }),
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -306,7 +306,10 @@ var MapView = Marionette.ItemView.extend({
             this.updateAreaOfInterest();
         } else {
             this._areaOfInterestLayer.clearLayers();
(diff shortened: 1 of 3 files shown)


caseypt:
It was tough to find a color that worked across all three basemaps (blacks, grays, and whites were out). In short I used the selected color for a couple reasons:

- It is not present in any of the current basemaps.
- It works well across dark and light basemaps.
- It somewhat fits with the other colors used in the app.

screen shot 2015-08-03 at 3 14 49 pm

screen shot 2015-08-03 at 3 15 31 pm

screen shot 2015-08-03 at 3 15 11 pm

rajadain:
+1 great color choice, I think it works very well across all three basemaps. While some highways in the Google layer are highlighted with a near color:

image

The hovering label helps differentiate, so it's not a big deal.

Is there a way to add a slight white border to these lines? That could potentially help, but if not then it's not a problem, and this looks pretty good by itself.

caseypt:
@rajadain Good idea. Here they are with a white border and slightly slimmer line. 

I also added a rule to round out the edges where the lines meet, which makes things look smoother.

screen shot 2015-08-03 at 3 33 27 pm
screen shot 2015-08-03 at 3 33 33 pm

rajadain:
Great! +1

rajadain:
I'd love to look at the code that allows for the white border

caseypt:
Updated!

mmcfarland:
Looks good

rajadain:
To be honest, I'm a little conflicted. The white border does make it easier to differentiate, but the pure orange looked cleaner and more minimalistic. I suppose we must be pragmatic and choose function over form.

caseypt:
I forgot that we use a similar color for drawing shapes. Since the drawing tool and boundary layers won't be on the map at the same time, I thought it might be good to reuse that existing color for the boundary layer borders. What do you think @mmcfarland @rajadain?

screen shot 2015-08-03 at 5 25 20 pm
screen shot 2015-08-03 at 5 25 25 pm
screen shot 2015-08-03 at 5 25 29 pm

rajadain:
+1 This works just as well, and I like that we're reusing an existing color rather than introducing a new one.

mmcfarland:
Yeah, reusing the same color is a good idea.  I kinda like the newer orangy color more than the pinkish one, though.  I don't think the original was "chosen" so much as "used".

caseypt:
Cool. I'll switch it to the new color.

caseypt:
Turns out we were just using the Leaflet Draw default color. Updated the drawn area polygon to match the boundary color, and set the AOI to be drawn in this color when you move back to draw from analyze. It was previously set to the default Leaflet blue.

screen shot 2015-08-04 at 9 59 28 am
screen shot 2015-08-04 at 9 59 34 am

mmcfarland:
Good call, +1

rajadain:
+1. That pink / blue inconsistency always bothered me, glad we've removed it.

PR #661, opened 2015-08-03
Allow ITSI users to clone projects via UI

rajadain:
## Overview

Create a new "Clone Project" link in the user drop-down for ITSI users that allows them to specify a project ID and clone it.

This is required because users won't be able to paste in a URL when the app is embedded inside an activity.

## Testing Instructions

 1. Log in as an ITSI user.
 2. Open the user drop-down in the header
 3. Click "Clone Project"
 4. Enter the project ID of a project you have access to. In reality this will be supplied by the teacher in the assignment.
 5. The project should now be cloned into your account.

Connects #560 
Connects #564

### DIFF
--- a/src/mmw/js/src/core/templates/header.html
+++ b/src/mmw/js/src/core/templates/header.html
@@ -12,6 +12,9 @@
                         <span class="caret"></span>
                     </a>
                     <ul class="dropdown-menu">
+                        {% if itsi %}
+                            <li><a href="#" class="clone-project">Clone Project</a></li>
+                        {% endif %}
                         <li><a href="#">Profile</a></li>
                         <li><a href="/user/logout" class="user-logout">Logout</a></li>
                     </ul>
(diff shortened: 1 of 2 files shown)


jamesmcclain:
Tested, works as advertised.  The only potential problem that I have been able to spot is that it gives a big, scary error screen when you type in a project id that does not exist:

screenshot from 2015-08-04 12 52 41

If that isn't a problem, then +1

rajadain:
I was thinking about that use-case last night as well. It is especially egregious because users inside the embedded view (the target audience for this PR) can't use their browser's back button to go back, even when we have a nicer 404 page.

I solved this by doing a test fetch of the entered project ID. If the user can successfully fetch the project, then they have read access and can clone as well. Otherwise they are shown an alert message describing why they can't clone the project:

image

Care to take another look?

jamesmcclain:
Yeah, that looks much friendlier.  :+1:

Issue #632, opened 2015-07-29
Survey and standardize all use of display units in the applications


mmcfarland:
We're going to standardize on displaying metric units _everywhere_ in the app.  Survey all the display, and update accordingly.  Be sure to include a label specifying the unit for table headers or similar.  Ultimately, it would be nice to provide a toggle, but in the meantime if we include a conversion utility and convert on display, I don't think we need to change the units we're using for calculations.  This will make it easier to toggle the display in the future.

Known measurement displays:

* [ ] Analysis Output
* [ ] Model Output
* [ ] Micro Site output
* [ ] AoI areas
* [ ] Modification and CP areas

PR #664, opened 2015-08-05
Feature/lol/standarize units

lliss:
Connects #632 

* Converts all display units to metric.
* Inputs to TR55 from front end remain in Imperial but display in metric.
* Job results that show pixel counts are converted to area based on approximate latitude.
* Micro site updates.
* Precipitation slider starts with 2.5 cm rather than 1 inch.

To test:

* Use site, as per normal. Note any numeric values that show on the screen and check that they are in metric.

### DIFF
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -4,7 +4,9 @@ from __future__ import unicode_literals
 from __future__ import absolute_import
 
 from celery import shared_task
+import json
 import logging
+import math
 
 # TODO Remove this when stub task is deleted.
 import time
@@ -14,9 +16,23 @@ from tr55.tablelookup import lookup_ki
 
 logger = logging.getLogger(__name__)
 
+KG_PER_POUND = 0.453592
+CM_PER_INCH = 2.54
+
 
 @shared_task
 def run_analyze(area_of_interest):
+
+    # Find the width in meters of a pixel at the average lat of the shape.
+    pairs = json.loads(area_of_interest)['coordinates'][0][0]
+    average_lat = reduce(lambda total, p: total+p[1], pairs, 0) / len(pairs)
+
+    # Zoom level is a hard coded value in the geoprocessing library.
+    zoom = 13
+    # See https://msdn.microsoft.com/en-us/library/bb259689.aspx
+    pixel_width = (math.cos(average_lat * math.pi / 180) * 2 * math.pi *
+                   6378137) / (256 * math.pow(2, zoom))
+
     time.sleep(3)
 
     results = [
@@ -114,6 +130,8 @@ def run_analyze(area_of_interest):
         }
     ]
 
+    convert_result_areas(pixel_width, results)
+
     return results
 
 
@@ -163,10 +181,28 @@ def format_quality(model_output):
         measure, code = input
         return {
             'measure': measure,
-            'load': model_output['modified'][code]
+            'load': model_output['modified'][code] * KG_PER_POUND
         }
 
     return map(fn, zip(measures, codes))
+
+
+def format_runoff(model_output):
+    """
+    Convert model output values that are in inches into centimeters.
+    """
+    inch_keys = ['runoff', 'inf', 'et']
+    for key in model_output:
+        for item in inch_keys:
+            model_output[key][item] = model_output[key][item] * CM_PER_INCH
+
+        distribution = model_output[key]['distribution']
+        for k in distribution:
+            for item in inch_keys:
+                model_output[key]['distribution'][k][item] = \
+                    model_output[key]['distribution'][k][item] * CM_PER_INCH
+
+    return model_output
 
 
 @shared_task
@@ -196,7 +232,7 @@ def run_tr55(census, model_input):
     return {
         'inputmod_hash': model_input['inputmod_hash'],
         'census': census,
-        'runoff': model_output,
+        'runoff': format_runoff(model_output),
         'quality': format_quality(model_output)
     }
 
@@ -208,6 +244,18 @@ def _get_precip_value(model_input):
         return precips[0]['value']
     except Exception:
         return None
+
+
+def convert_result_areas(pixel_width, results):
+    """
+    Updates area values on the survey results. The survey gives area as a
+    number of pixels in a geotiff. This converts the area to square meters.
+    """
+    for i in range(len(results)):
+        categories = results[i]['categories']
+        for j in range(len(categories)):
+            categories[j]['area'] = (categories[j]['area'] * pixel_width *
+                                     pixel_width)
 
 
 # TODO: For demonstration purposes.
--- a/src/mmw/js/src/core/utils.js
+++ b/src/mmw/js/src/core/utils.js
@@ -73,6 +73,54 @@ var utils = {
             });
 
         return md5(JSON.stringify(sortedCollection));
+    },
+
+    convertToMetric: function(value, fromUnit) {
+        fromUnit = fromUnit.toLowerCase();
+        switch (fromUnit) {
+            case 'in':
+                // return cm.
+                return value * 2.54;
+
+            case 'ft':
+                // return meters.
+                return value * 0.3048;
+
+            case 'mi':
+                // return Km.
+                return value * 1.60934;
+
+            case 'lb':
+                // return Kg.
+                return value * 0.453592;
+
+            default:
+                throw 'Conversion not implemented.';
+        }
+    },
+
+    convertToImperial: function(value, fromUnit) {
+        fromUnit = fromUnit.toLowerCase();
+        switch (fromUnit) {
+            case 'cm':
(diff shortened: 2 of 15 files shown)


lliss:
This is a draft. Looking for a little feedback.

lliss:
@azavea-bot rebuild.

src/mmw/apps/modeling/tasks.py
```diff
@@ -163,12 +176,30 @@ def fn(input):
         measure, code = input
         return {
             'measure': measure,
-            'load': model_output['modified'][code]
+            'load': model_output['modified'][code] * 0.453592  # Lbs to Kg.
```
rajadain: Consider replacing this with a constant declared at the top. That would make it easier to adjust the resolution if we ever need to, and also better communicate the meaning of this (otherwise) magic number.

src/mmw/apps/modeling/tasks.py
```diff
@@ -163,12 +176,30 @@ def fn(input):
...
 
 
+def format_runoff(model_output):
+    """
+    Convert model output values that are in inches into centimeters.
+    """
+    inch_keys = ['runoff', 'inf', 'et']
+    for key in model_output:
+        for item in inch_keys:
+            model_output[key][item] = model_output[key][item] * 2.54
```
rajadain: Same here for `2.54`.

src/mmw/apps/modeling/tasks.py
```diff
@@ -210,6 +241,19 @@ def _get_precip_value(model_input):
...
 
+def convert_result_areas(pixel_width, results):
+    """
+    Updates area values on the result set from tr55. TR55 gives area as a
+    number of pixels in a geotiff. This converts the area to square meters.
+    """
+    for i in range(len(results)):
+        categories = results[i]['categories']
+        for j in range(len(categories)):
+            print(categories[j])
```
rajadain: I assume this is here for debugging. Probably should be removed prior to merging.

lliss: Correct.

rajadain:
The coverage label implies percentage, but the values in the table are decimal fractions:

image
image

They should be multiplied by 100 to be formatted as percentages, and match the charts.

lliss:
Good catch.

rajadain:
The y-axis labels in the runoff chart don't _say_ inches, but that is what they are:

image

rajadain:
Given that we don't default precipitation to the nice round figure of 1.0" anymore, and since 2.54 can't be selected by the users even if they try (since we only move in increments of 0.25 from 0.00), consider changing that default to something better like 2.50 or 3.00:

image

src/mmw/js/src/core/utils.js
```diff
@@ -73,6 +73,54 @@ var utils = {
             });
 
         return md5(JSON.stringify(sortedCollection));
+    },
+
+    convertToMetric: function(value, fromUnit) {
```
caseypt: Consider modifying `convertToMetric` and `convertToImperial` to take in the desired output unit as a parameter. I think this could make it more clear when using these functions.

src/mmw/js/src/main_water_balance.js
```diff
@@ -39,10 +39,10 @@ var initialize = function(model) {
 
         var result = model[soil][land][precip];
 
-        $precipText.text(precip + '"');
-        $evapoText.text(result.et + '"');
-        $runoffText.text(result.r + '"');
-        $infilText.text(result.i + '"');
+        $precipText.text(convertToMetric(precip) + 'cm');
```
rajadain: There should always be a space between the numeric value and the unit. See http://physics.nist.gov/cuu/Units/checklist.html#15 "Unit Spacing":

> There is a space between the numerical value and unit symbol, even when the value is used in an adjectival sense, except in the case of superscript units for plane angle.
> 
> proper: 	a 25 kg sphere
> an angle of 2° 3'  4"
> If the spelled-out name of a unit is used, the normal rules of English apply: "a roll of 35-millimeter film."

src/mmw/js/src/modeling/tr55/quality/templates/table.html
```diff
@@ -2,7 +2,7 @@
     <thead>
         <tr>
             <th>Quality Measure</th>
-            <th class="text-left">Load (lbs)</th>
+            <th class="text-left">Load (Kg)</th>
```
rajadain: SI representation uses all lowercase letters, so "kg" instead of "Kg". See https://en.wikipedia.org/wiki/Kilogram

rajadain:
+1, great work! I think you got all of the occurrences, except for the silently imperial y-axis labels in the runoff chart. If we can add a space before "cm" and replace "Kg" with "kg", I think we're good to go!

The 2.54 cm default value of precipitation looks funny, but I appreciate that it's wired in to the default 1.0" original value. I'm unsure how difficult it would be to switch it.

I'm surprised that you only had to change one test, and that too only checked for the table headers. I assumed there were more tests that looked at the numbers themselves. Are they still running because things are still imperial under the hood, or do we simply not have tests for checking the numbers?

rajadain:
Uhhh the micro site seems to be completely broken:

image

I get the following error:

```
TypeError: model[soil] is undefined
```

here: https://github.com/WikiWatershed/model-my-watershed/pull/664/files#diff-410ada084f8c0311b241bc1e4801ca80R40

lliss:
@rajadain RE: "The y-axis labels in the runoff chart don't say inches, but that is what they are:"

Did you reload the celery workers? I think if you do that, you'll start getting metric results back. The output of that graph is actually just raw numbers from the output of the job run and since we now do a metric conversion prior to returning the results, you should see the top of the graph match the precipitation slider.

lliss:
@rajadain RE: "micro site seems to be completely broken:"

I'll need to see this on your machine.

src/mmw/js/src/modeling/tr55/quality/templates/table.html
lliss: Got it. Will update.

src/mmw/js/src/main_water_balance.js
lliss: Makes sense. +1 for standards.

src/mmw/js/src/core/utils.js
lliss: I'm not thrilled with that idea because it allows for the possibility of converting to things that don't make any sense. Ex: lbs -> km 

If this seems a lot more helpful, I'd create more function
convertDistanceToMetric
convertDistanceToImperial
convertMassToMetric
convertMassToImperial

just to avoid a ton of logic to ensure the input and output units are compatible.

caseypt: Right. Good point. I'm surprised there aren't that many great libraries out that there that do this (I looked around a bit). I think it's fine as is.

rajadain:
@lliss Thanks, restarting Celery fixed the chart. Also, the Micro app is working correctly now. Probably a bad checkout on my part. Sorry about that.

lliss:
@azavea-bot rebuild

lliss:
@jamesmcclain I'd like you to take a look at what's going on with the area calculations based on the results of the pixel counts.

jamesmcclain:
Certainly, will do.

lliss:
@azavea-bot rebuild

rajadain:
Taking one final look

src/mmw/apps/modeling/tasks.py
```diff
@@ -210,6 +244,18 @@ def _get_precip_value(model_input):
         return None
 
 
+def convert_result_areas(pixel_width, results):
+    """
```
jamesmcclain: I might have accidentally misinformed you.  Talking about the areas of geotiff pixels: the TR-55 module expects that number as input, not output.  For example:

> The cell_res parameter gives the resolution (size) of each cell. It is used for converting runoff, 
> evapotranspiration, and infiltration amounts from inches to volumes.

The part above were you calculate the pixel area can be used for this, I think.

jamesmcclain: Actually now that I think about it, since a pixel count in the area of interest is always available, it is probably easy to infer the area of each pixel given the area of the polygon (computer by `turf-area`) or similar.

rajadain:
+1. Tested all the places I did before, and everything looks consistently metric. Very detailed work!

cm values in the Micro Site now go up to 3 digits, which expands the containing white box, but it doesn't look odd:

image

src/mmw/apps/modeling/tasks.py
```diff
@@ -210,6 +244,18 @@ def _get_precip_value(model_input):
...
+    Updates area values on the result set from tr55. TR55 gives area as a
```
lliss: After talking it seem that the comment is misleading but the results are accurate. I'll update these and rebase.

lliss:
@azavea-bot rebuild

Issue #594, opened 2015-07-20
Compare views: pass in Project and show area of interest on maps


lewfish:
Currently, a test Project model is passed to the Compare window. Instead, pass in the real project (being edited in the modeling view) and set the extent, area of interest, and modifications on the maps.

See the mockup in #573 and #591

PR #665, opened 2015-08-05
Load projects into /compare

lewfish:
Situations to test:
 * As a guest, switch the base layer. In the modeling view, add a few modifications to a scenario, and click compare.You should see the modifications, AOI, correct extent on the map and the selected base layer for that scenario. The url should be `/project/compare`. Clicking back should lead to `/project` and the same project should be loaded (ie. it shouldn't create a new one). Clicking back again (using the browser's back button or the modeling view's back button) should lead to `/analyze`.
 * As a guest, in the modeling view, log in, and then click compare. The url should be `/project/project_id/compare`. Clicking back should lead to `/project/project_id/scenario/scenario_id`. Clicking back again should lead to `/analyze`.
 * Do the same thing except directly visit the URL `/project/project_id/compare`. It should work the same as above.
 * Directly visit the URL `/project/compare`. It should give a 404 error.
 * As a guest, in the modeling view, click compare. Then log in. This should save the project, and update the url to `/project/project_id/compare`. Clicking back should lead to `/project/project_id/scenario/scenario_id`. Clicking back again should lead to `/analyze`.

Known bug: If you are in `/project/project_id/scenario/scenario_id` and then hit the browser's back button, and then the browser's forward button, there is a problem loading the modeling view. This behavior also doesn't work on the develop branch. I'll see if I can figure it out, but if not, I'll make an issue for it.

Connects #594

### DIFF
--- a/src/mmw/js/src/analyze/controllers.js
+++ b/src/mmw/js/src/analyze/controllers.js
@@ -4,7 +4,8 @@ var _ = require('underscore'),
     App = require('../app'),
     router = require('../router').router,
     views = require('./views'),
-    models = require('./models');
+    models = require('./models'),
+    settings = require('../core/settings');
 
 var AnalyzeController = {
     analyzePrepare: function() {
@@ -17,6 +18,10 @@ var AnalyzeController = {
         // analyze mode
         if (!App.map.get('maskLayerApplied')) {
             App.map.set('maskLayerApplied', true);
+        }
+
+        if (!settings.get('activityMode')) {
+            App.currProject = null;
         }
     },
 
--- a/src/mmw/js/src/compare/views.js
+++ b/src/mmw/js/src/compare/views.js
@@ -2,13 +2,14 @@
 
 var _ = require('lodash'),
     $ = require('jquery'),
-    L = require('leaflet'),
     Marionette = require('../../shim/backbone.marionette'),
+    App = require('../app'),
+    coreModels = require('../core/models'),
+    coreViews = require('../core/views'),
     modConfigUtils = require('../modeling/modificationConfigUtils'),
     compareWindowTmpl = require('./templates/compareWindow.html'),
     compareScenariosTmpl = require('./templates/compareScenarios.html'),
     compareScenarioTmpl = require('./templates/compareScenario.html'),
-    compareMapTmpl = require('./templates/compareMap.html'),
     compareChartTmpl = require('./templates/compareChart.html'),
     compareModificationsTmpl = require('./templates/compareModifications.html');
 
@@ -41,6 +42,18 @@ var CompareWindow = Marionette.LayoutView.extend({
         // so the offset of the container needs to be
         // recomputed.
         $(window).bind('resize.app', _.debounce(_.bind(this.updateContainerPos, this)));
+        this.listenTo(App.user, 'change:guest', this.saveAfterLogin);
+    },
+
+    saveAfterLogin: function(user, guest) {
+        if (!guest && this.model.isNew()) {
+            var user_id = user.get('id');
+            this.model.set('user_id', user_id);
+            this.model.get('scenarios').each(function(scenario) {
+                scenario.set('user_id', user_id);
+            });
+            this.model.saveProjectAndScenarios();
+        }
     },
 
     destroy: function() {
@@ -101,10 +114,28 @@ var CompareScenarioView = Marionette.LayoutView.extend({
         modificationsRegion: '.modifications-region'
     },
 
+    initialize: function(options) {
+        this.projectModel = options.projectModel;
+    },
+
     onShow: function() {
-        this.mapRegion.show(new CompareMapView({
-            model: this.model
-        }));
+        this.mapModel = new coreModels.MapModel({});
+        this.mapModel.set('areaOfInterest', this.projectModel.get('area_of_interest'));
+        this.mapView = new coreViews.MapView({
+            model: this.mapModel,
+            el: $(this.el).find('.map-container').get(),
+            addZoomControl: false,
+            addLocateMeButton: false,
+            addLayerSelector: false,
+            showLayerAttribution: false,
+            initialLayerName: App.getMapView().getActiveBaseLayerName()
+        });
+
+
+        this.mapView.fitToAoi();
+        this.mapView.updateAreaOfInterest();
+        this.mapView.updateModifications(this.model.get('modifications'));
+        this.mapRegion.show(this.mapView);
         this.chartRegion.show(new CompareChartView({
             model: this.model
         }));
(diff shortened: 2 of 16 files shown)


src/mmw/js/src/analyze/controllers.js
```diff
@@ -18,6 +18,11 @@ var AnalyzeController = {
         if (!App.map.get('maskLayerApplied')) {
             App.map.set('maskLayerApplied', true);
         }
+
+        // Clear currProject so that
+        // the ModelingController will know to make a
+        // new project when coming from the Analyze views.
+        App.currProject = null;
```
rajadain: I wonder how this would play with the Single Project Mode made in #610 which initializes `App.currProject` in the `draw` view when given the flag `?itsi_embed=true`. Will test and report.

lewfish: Thanks, it would be great if you could evaluate this from the standpoint of ITSI compatibility.

lewfish: In particular, changes to `src/mmw/js/src/modeling/controllers.js'

rajadain:
The modeling controller is getting pretty complicated. I had to write this pseudocode down to understand what was going on:

```python
if projectId:
    # Project ID was specified, load given project and
    # Initialize project
else:
    # Project ID was not specified. We should create a new project
    if App.currProject and settings.get('activityMode'):
        # We are in activity mode and a project has already been
        # created
        if App.currProject.get('needs_reset'):
            # The project has been reset, we should get rid of all
            # scenarios and setup new ones, then
            # Initialize project
        else:
            # Initialize project
    else:
        # Either not in activity mode, or project not set
        if not App.currProject:
            # Create new project
        else:
            # Navigate to project

        # Initialize project
```

rajadain:
Logging out from the `/compare` screen does not redirect to home like it does in all other cases.

src/mmw/js/src/analyze/controllers.js
rajadain: Yes this does indeed break the single project mode. I'm not sure why this line is required at all (since the previous behavior was the same when loading an existing project `/project/XX/scenario/YY/`, going to back to Analyze, then coming forward to Modeling which would create a new project), but if we must keep it then it should only run when not in single project mode:

```javascript
if (!settings.get('activityMode')) {
    App.currProject = null;
}
```

src/mmw/js/src/compare/views.js
```diff
@@ -41,6 +42,18 @@ var CompareWindow = Marionette.LayoutView.extend({
...
         $(window).bind('resize.app', _.debounce(_.bind(this.updateContainerPos, this)));
+        this.listenTo(App.user, 'change:guest', this.saveAfterLogin);
+    },
+
+    saveAfterLogin: function(user, guest) {
+        if (!guest && this.model.isNew()) {
+            var user_id = user.get('id');
+            this.model.set('user_id', user_id);
+            this.model.get('scenarios').each(function(scenario) {
+                scenario.set('user_id', user_id);
```
lliss: Should we move this into `saveProjectAndScenarios` ? Seems like we'd always want this to happen based off of the project.

src/mmw/js/src/compare/views.js
```diff
@@ -101,10 +114,28 @@ var CompareScenarioView = Marionette.LayoutView.extend({
...
+
     onShow: function() {
-        this.mapRegion.show(new CompareMapView({
-            model: this.model
-        }));
+        this.mapModel = new coreModels.MapModel({});
+        this.mapModel.set('areaOfInterest', this.projectModel.get('area_of_interest'));
+        this.mapView = new coreViews.MapView({
+            model: this.mapModel,
+            el: $(this.el).find('.map-container').get(),
```
lliss: Why do we need `.get()` here?

src/mmw/js/src/compare/views.js
rajadain: This seems to be a replica of [`saveAfterLogin`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/views.js#L69-L78) in `modeling/views.js`. That line, originally added in #385, was placed outside of `saveProjectAndScenarios` since we only need to set it for saving existing projects and scenarios that don't have a user assigned to them. Every project and scenario created when a user is logged in already has this set, so doing so every time in `saveProjectAndScenarios` would be redundant.

src/mmw/js/src/modeling/templates/scenariosBar.html
```diff
@@ -10,5 +10,5 @@
 </div>
 <div class="fade-tabs"></div>
 <div class="compare-top-btn">
-    <a href="compare" class="btn btn-compare"><i class="fa fa-th"></i> Compare</a>
```
lliss: Isn't it better to leave this on? I'd be inclined to leave this on and have it update when the project id changes:

`<a href="/project/{{ pid }}/compare" class="btn btn-compare"><i class="fa fa-th"></i> Compare</a>`

lewfish:
I added a commit to fix the bug re: square AOIs.

src/mmw/js/src/compare/views.js
lewfish: To return a DOM element as opposed to a jQuery selector, since that's what el is supposed to be. Maybe I could just do `$el: $(this.el).find('.map-container')`?

src/mmw/js/src/modeling/controllers.js
```diff
@@ -94,24 +94,30 @@ var ModelingController = {
...
-                });
-
-                // TODO evalutate if we can remove this global by reworking this
-                // code.
-                App.currProject = project;
-                setupNewProjectScenarios(project);
-                project.on('change:id', function() {
-                    router.navigate(project.getReferenceUrl());
-                });
+                if (!App.currProject) {
```
lliss: Looks like removing this and the null setting src/mmw/js/src/analyze/controllers.js puts activity mode back into effect and doesn't cause any major problems. It does cause new projects to be created when jumping back and forth to compare and modeling views. But I'd suggest that maybe we make the guest behaviour a lot more like activity mode in that you get a clean project and just keep adjusting the same one rather that starting over each time  you hit the modeling page.

src/mmw/js/src/compare/views.js
lliss: The get seemed odd to me. I also thought that el needed to be a string. Didn't realize you could pass in a DOM reference.

lliss:
Looks pretty good. Only issue I see is making sure we don't break activity mode and assessing the routing that I noted in `scenariosBar.html`

src/mmw/js/src/modeling/templates/scenariosBar.html
lewfish: I changed it to use `data-url` to avoid doing a full page reload and instead just go to the CompareController. What's the advantage of leaving it the way it is?

src/mmw/js/src/routes.js
```diff
@@ -12,8 +12,8 @@ router.addRoute(/^/, DrawController, 'draw');
 router.addRoute(/^analyze/, AnalyzeController, 'analyze');
 router.addRoute('project(/:projectId)(/scenario/:scenarioId)(/)', ModelingController, 'project');
 router.addRoute('project/:projectId/clone(/)', ModelingController, 'projectClone');
-//TODO Add route for /project/XX/scenario/compare
-router.addRoute(/^compare/, CompareController, 'compare');
+router.addRoute('project(/:projectId)(/scenario/:scenarioId)(/)', ModelingController, 'project');
```
rajadain: Isn't this the same as line 13?

lewfish: Yes, looks like a mistake. I'll fix it in #673

rajadain: I was just checking, I'm also fixing in my PR, wanted to make sure before removing it.

Issue #662, opened 2015-08-04
Refactor TR-55 Labels to be consistent


rajadain:
As of #653 we have switched to using labels instead of NLCD codes:

```javascript
    landMap: {
        'urban_grass': ['turfGrass'],
        'tall_grass_prairie': ['tallGrass'],
        'li_residential': ['lir'],
        'hi_residential': ['hir'],
        'commercial': ['commercial'],
        'desert': ['desert'],
        'deciduous_forest': ['forest'],
        'chaparral': ['chaparral'],
        'grassland': ['grassland'],
        'pasture': ['pasture'],
        'short_grass_prairie': ['shortGrass'],
        'row_crop': ['rowCrops']
    },
```

This mapping between labels of one type to another is redundant and confusing, and should be refactored away with one standard labeling.

Before, we were translating from numeric NLCD encoding to alphabetic labels, used in CSS and JavaScript, so it was okay. Now we're converting one alphabetic labeling to another, which adds more noise than signal. This will get worse when we start to actually process the TR-55 results in the back end, since we send the JavaScript style labeling in our API calls which will have to be translated back to Python style labeling there before processing.

Instead of this dual reversal, we should either switch the styling in the front-end to use Pythonic labels, or the labeling in the back-end to use JavaScripty labels. I think that changing the back-end would be easier, since the front-end labels are spread across the app in multiple places. However, we have made a policy decision in #237 to prefer Pythonic labeling in the front-end as well, so I think that would be a a more stylistically fitting choice.

### Places to Update in the Front-End

 * [ ] `src/mmw/apps/core/templates/patterns.html`
 * [ ] `src/mmw/apps/water_balance/templates/home_page/index.html`
 * [ ] `src/mmw/js/src/modeling/modificationConfig.json`
 * [ ] `src/mmw/js/src/modeling/templates/controls/landCover.html`
 * [ ] `src/mmw/js/src/modeling/templates/controls/conservationPractice.html`
 * [ ] `src/mmw/js/src/water_balance/models.js`

PR #671, opened 2015-08-10
Make land use and BMP names in the frontend consistent with those in the backend

jamesmcclain:
The internal names for various land uses and BMPs have been changed to be consistent with those found in the TR-55 module.

   * [lh]ir -> (li|hi)_residential
   * forest -> deciduous_forest
   * turf(Grass|_grass) -> urban_grass
   * wetland -> woody_wetland
   * ((tg|sg)_prairie|(tall|Short)Grass) -> (tall|short)_grass_prairie
   * no_till_agriculture -> no_till
   * veg_infil_basin -> infiltration_trench
   * rowCrops -> row_crop

Connects #662

**To test**
   * Bring up some previously-created projects and scenarios and make sure that they still work as expected
   * Create some new projects and scenarios and make sure that they work as expected
   * Add some land uses and/or BMPs to an existing scenario and confirm that this mixed scenario works as intended
   * Play with the micro-app and make sure that it still behaves as expected

### DIFF
--- a/src/mmw/js/src/modeling/modificationConfig.json
+++ b/src/mmw/js/src/modeling/modificationConfig.json
@@ -14,8 +14,9 @@
         "summary": "This category includes areas of bedrock, desert pavement, scarps, talus, slides, volcanic material, glacial debris, sand dunes, strip mines, gravel pits and other accumulations of earthen material. Generally, vegetation accounts for less than 15% of total cover.",
         "color": "#f1edd9"
     },
-    "forest": {
-        "name": "Forest",
+    "deciduous_forest": {
+        "name": "Deciduous Forest",
+        "shortName": "Forest",
         "summary": "This category includes areas that are covered by vegetation with more than 20% of the land surface covered by trees that are taller than 5 meters.",
         "color": "#53e9b4"
     },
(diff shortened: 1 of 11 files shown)


src/mmw/js/src/modeling/modificationConfig.json
```diff
@@ -14,8 +14,9 @@
         "summary": "This category includes areas of bedrock, desert pavement, scarps, talus, slides, volcanic material, glacial debris, sand dunes, strip mines, gravel pits and other accumulations of earthen material. Generally, vegetation accounts for less than 15% of total cover.",
         "color": "#f1edd9"
     },
-    "forest": {
-        "name": "Forest",
+    "deciduous_forest": {
+        "name": "Deciduous Forest",
+	"shortName": "Forest",
```
rajadain: Tab instead of spaces

src/mmw/js/src/water_balance/models.js
```diff
@@ -31,26 +16,24 @@ var WaterBalanceModel = {
...
+                'i' : round(line[4]),
+                'r' : round(line[5])
+            };
         }
 
         for (var j = 1; j < lines.length; j++) {
             var line = lines[j].split(',');
             var soil = this.soilMap[line[2]];
-            var lands = this.landMap[line[1]];
+            var lands = line[1];
```
rajadain: Perhaps we can now rename `lands` to `land` since it always has exactly one value.

rajadain:
+1 to the functionality, works as specified, great job!

Regarding the existing modifications, I see the value in translating it on the UI end without touching the underlying data. However, I don't think that we will ever be able to remove that translation function, [as hoped for](https://github.com/jamesmcclain/model-my-watershed/commit/656204b3b932e2821a96145c0e3d71035813f010#diff-e037f7a38fdeab17f9c52f9b74808ebcR15), unless those existing modifications are deleted.

Ideally we could write a migration that would update all the values in the database, ready for the next session which expects them, however editing JSON values is difficult through the Django Migrations API. Perhaps we can have a function which, after loading an existing scenario, checks to see if it has any modifications whose `value` is an `oldKey`, and translate it to the `new_key`, so that whenever the scenario is saved next it's value is updated in the database. This way, slowly over time, most (if not all) scenarios will be updated to use the new keys.

jamesmcclain:
> Ideally we could write a migration that would update all the values in the database, ready for the 
> next session which expects them, however editing JSON values is difficult through the Django 
> Migrations API. Perhaps we can have a function which, after loading an existing scenario, checks 
> to see if it has any modifications whose value is an oldKey, and translate it to the new_key, so that 
> whenever the scenario is saved next it's value is updated in the database. This way, slowly over 
> time, most (if not all) scenarios will be updated to use the new keys.

Good idea, I have implemented it.  Be careful which scenarios you load-up if you choose to test this patch, because it will automatically and silently upgrade any scenarios that you load.

lliss:
+1.

rajadain:
Taking another look.

rajadain:
+1, upgrading is very smooth and works perfectly. Great job!

jamesmcclain:
Sounds good, thanks!

Issue #593, opened 2015-07-20
Compare views: Precipitation slider, charts and run model


lewfish:
Add a precipitation slider and bar charts and make it so that changes to the slider trigger model runs. The results of these model runs and (input values) should not be saved as in the modeling views.

See the mockup in #573 and #591

PR #673, opened 2015-08-10
Run models from compare views

lewfish:
The compare views are designed to be read-only in the sense that any changes to the inputs or results are not saved to the project. 

I realize the layout is a bit off -- I'll fix it in a forthcoming commit.

To test:
 * In the modeling views, change the precipitation level on a scenario.
 * Go to the compare views
 * Check that the precipitation value and results were copied over.
 * Change the precip value, and check that new results load. Make sure the spinner appears, and that you can switch to Water Quality results. While Water Quality results are visible, change the precip value and check that it hasn't switched back to the Runoff results.
 * Click back to go to the modeling views, and check that the new results and precip value were *not* copied.
 * The previous steps should work whether logged in or not.
 * Try going to the compare views as a guest and then login. The project should be saved and the URL updated, although any changes introduced by the compare views should not be saved.

Connects #593

### DIFF
--- a/src/mmw/js/src/compare/controllers.js
+++ b/src/mmw/js/src/compare/controllers.js
@@ -1,6 +1,7 @@
 "use strict";
 
-var App = require('../app'),
+var _ = require('lodash'),
+    App = require('../app'),
     router = require('../router').router,
     views = require('./views'),
     modelingModels = require('../modeling/models.js');
@@ -8,6 +9,7 @@ var App = require('../app'),
 var CompareController = {
     compare: function(projectId) {
         if (App.currProject) {
+            setupProjectCopy();
             showCompareWindow();
         } else if (projectId) {
             App.currProject = new modelingModels.ProjectModel({
@@ -16,28 +18,96 @@ var CompareController = {
             App.currProject
                 .fetch()
                 .done(function() {
+                    setupProjectCopy();
                     showCompareWindow();
                 });
         }
-        App.currProject.on('change:id', updateUrl);
         // else -- this case is caught by the backend and raises a 404
     },
 
     compareCleanUp: function() {
+        App.user.off('change:guest', saveAfterLogin);
+        App.origProject.off('change:id', updateUrl);
+
+        // Switch back to the origProject so any changes are discarded.
+        App.currProject.off();
+        App.currProject = App.origProject;
+
         App.rootView.footerRegion.empty();
-        App.currProject.off('change:id', updateUrl);
     }
 };
 
+function setupProjectCopy() {
+    /*
+    Create a copy of the project so that:
+      -Changes to the results and inputs are not saved to the server
+      -Changes to the results and inputs are not present after hitting the back button (and project is unsaved).
+       When the user is logged in as a guest, the only copy of the original inputs and results
+       are in App.currProject, and modifying these values in the compare views will result in those new
+       values being back in the modeling views, which is not what we want.
+      -The original project (without changes) is saved when logging in. If the user is a guest,
+       goes into the compare views, and then logs in, the project should be saved with the inputs
+       and results that were present before going to the compare views. This is to enforce the
+       constraint that inputs and results entered in the compare views should never be saved.
+       Without holding onto the original copies of these values, it's not possible to do this.
+     */
+    App.origProject = App.currProject;
+    App.currProject = copyProject(App.origProject);
+
+    App.user.on('change:guest', saveAfterLogin);
+    App.origProject.on('change:id', updateUrl);
+}
+
+// Creates a special-purpose copy of the project
+// for the compare views, since creating a true deep
+// clone is more difficult and unnecessary.
+function copyProject(project) {
+    var scenariosCopy = new modelingModels.ScenariosCollection();
+    project.get('scenarios').forEach(function(scenario) {
+        var scenarioCopy = new modelingModels.ScenarioModel({});
+        scenarioCopy.set({
+            is_current_conditions: scenario.get('is_current_conditions'),
+            modifications: scenario.get('modifications'),
+            modification_hash: scenario.get('modification_hash'),
+            results: new modelingModels.ResultCollection(scenario.get('results').toJSON()),
+            inputs: new modelingModels.ModificationsCollection(scenario.get('inputs').toJSON()),
+            inputmod_hash: scenario.get('inputmod_hash'),
+            allow_save: false
+        });
+        scenarioCopy.get('inputs').on('add', _.debounce(_.bind(scenarioCopy.fetchResults, scenarioCopy), 500));
+        scenariosCopy.add(scenarioCopy);
+    });
+    return new modelingModels.ProjectModel({
+        name: App.origProject.get('name'),
+        area_of_interest: App.origProject.get('area_of_interest'),
+        model_package: App.origProject.get('model_package'),
+        scenarios: scenariosCopy,
+        allow_save: false
+    });
+}
+
+function saveAfterLogin(user, guest) {
+    if (!guest && App.origProject.isNew()) {
+        var user_id = user.get('id');
+        App.origProject.set('user_id', user_id);
+        App.origProject.get('scenarios').each(function(scenario) {
+            scenario.set('user_id', user_id);
+        });
+        // Save the origProject (as opposed to the currProject)
+        // to not include changes to inputs and results.
+        App.origProject.saveProjectAndScenarios();
+    }
+}
+
 function updateUrl() {
     // Use replace: true, so that the back button will work as expected.
-    router.navigate(App.currProject.getCompareUrl(), { replace: true });
+    router.navigate(App.origProject.getCompareUrl(), { replace: true });
 }
 
 function showCompareWindow() {
     var compareWindow = new views.CompareWindow({
-        model: App.currProject
-    });
+            model: App.currProject
+        });
     App.rootView.footerRegion.show(compareWindow);
 }
 
--- a/src/mmw/js/src/compare/templates/compareModeling.html
+++ b/src/mmw/js/src/compare/templates/compareModeling.html
@@ -1,0 +1,19 @@
+<h3>Model Output</h3>
+{% if polling %}
+    <i class="fa fa-circle-o-notch fa-spin"></i>
+{% endif %}
+<div class="pull-right">
+    <select class="form-control btn btn-small btn-primary">
+        {% for result in results %}
+            {% if result.active %}
+                <option value="{{ result.name }}" selected>
+            {% else %}
+                <option value="{{ result.name }}" >
+            {% endif %}
+            {{ result.displayName }}
+                </option>
+        {% endfor %}
+    </select>
+</div>
+<div class="result-region"></div>
+<div class="controls-region"></div>
--- a/src/mmw/js/src/compare/views.js
+++ b/src/mmw/js/src/compare/views.js
@@ -6,11 +6,13 @@ var _ = require('lodash'),
     App = require('../app'),
     coreModels = require('../core/models'),
     coreViews = require('../core/views'),
+    modelingModels = require('../modeling/models'),
+    modelingViews = require('../modeling/views'),
     modConfigUtils = require('../modeling/modificationConfigUtils'),
     compareWindowTmpl = require('./templates/compareWindow.html'),
     compareScenariosTmpl = require('./templates/compareScenarios.html'),
     compareScenarioTmpl = require('./templates/compareScenario.html'),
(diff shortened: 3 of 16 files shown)


src/mmw/js/src/compare/controllers.js
```diff
@@ -16,28 +18,87 @@ var CompareController = {
             App.currProject
                 .fetch()
                 .done(function() {
+                    setupProjectCopy();
```
caseypt: Instead of copying the project, is it possible for us to use and/or expand some of the logic we are using for public/read-only projects? Copying the project seems a bit heavy-handed to me. I think it would be great if we had some sort of read-only switch that we could put the project in, where changes to the project wouldn't trigger a save. 

I'm curious what others think.

caseypt: Also, copying the project can add a lot to the page's footprint, since there's now twice as many models and event listeners as before, along with all of the Leaflet maps.

caseypt:
Going to `/project/compare` goes to a 404. Not sure if that is correct or not.

caseypt:
Things are working pretty well.

I'm having a little trouble testing this out

> Try going to the compare views as a guest and then login. The project should be saved and the URL updated, although any changes introduced by the compare views should not be saved."

Should I be able to visit `/project/compare`? Or if I go to an existing project from another user, should it save the project in my account when I log in?

caseypt:
Your comment about the layout being off might include these things, but just in case:

screen shot 2015-08-11 at 9 35 11 am

Should the panels be expanding to take up the full width of the page?

The legend for the water quality chart is unreadable.

src/mmw/js/src/modeling/models.js
```diff
@@ -46,7 +46,8 @@ var ResultModel = Backbone.Model.extend({
         displayName: '',
         inputmod_hash: null,
         result: null,
-        polling: false
+        polling: false,
+        active: false
```
caseypt: Consider adding some comments, like in `ProjectModel`, for these attributes.

src/mmw/js/src/compare/controllers.js
mmcfarland: @caseypt makes a great point.  It looks like for "read only" in share mode we are checking for a user match (https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/models.js#L153).  Could that call be extracted into a function `isReadOnly` that checks either that the user matches or that a new `readOnly` attribute has been set to serve both purposes?  Would be a much lighter solution that the clone, and give us the ability to decide later if we should, in fact, be persisting any of these values.

lewfish: This idea would address point 1 of the comment in setupProjectCopy 
 -changes to the results and inputs are not saved to the server

But would not address points 2 and 3
-changes to the results and inputs are not present after hitting the back button (and project is unsaved)
-the original project (without changes) is saved when logging in

Originally, I was setting the user_id of the copied project to 0, so that the logic for shared projects would be used, but I decided to add an attempt_save flag to the ProjectModel and ScenarioModel to be more explicit about not attempting to save.

caseypt: Can you explain this point

> the original project (without changes) is saved when logging in

I'm not sure I understand.

lewfish:
Going to /project/compare directly should indeed result in a 404. You should only be able to navigate to /project/compare when you are logged in as guest, and the project is stored in App.currProject.

The panels should not expand to fill the screen, but you are right about the axis on water quality being squished. There's also a strange issue with the charts growing everytime you resize the screen, which I'm looking into.

Logging in when in the compare views with a shared project should not result in a save.

src/mmw/js/src/compare/controllers.js
lewfish: If the user is a guest, goes into the compare views, and then logs in, the project should be saved with the inputs and results that were present before going to the compare views. This is to enforce the constraint that inputs and results entered in the compare views should never be saved. Without holding onto the original copies of these values, it's not possible to do this.

caseypt: Got it. Thanks.

> the original project (without changes) is saved when logging in

Consider asking the user to log in before going to compare view. I may have missed some discussion around this, but I don't see much value in not saving the project by this point.

> changes to the results and inputs are not present after hitting the back button (and project is unsaved)

Consider fetching the project from the server when returning from the compare view to repopulate the page with the saved version of the project.

Alternatively, consider doing the project "copying" a bit more dynamically and on an as-needed basis. For example, once a precipitation slider is changed for a scenario, have that scenario store it's current attributes in a temporary attribute. When the page is moved back to model, then apply the saved attributes to the changed scenario. I haven't used it much but `model.previousAttributes` and/or `model.previous()` may be of some use here.

mmcfarland: > Consider asking the user to log in before going to compare view. I may have missed some discussion around this, but I don't see much value in not saving the project by this point.

Unfortunately, we're on the hook to allow a full worflow for unauthenticated users.

caseypt:
^^ Thanks for that. Following that workflow works as described.

I'm going to open an issue for the panel sizing.

src/mmw/js/src/compare/controllers.js
lewfish: > Consider fetching the project from the server when returning from the compare view to repopulate the page with the saved version of the project.

We can do that for saved projects, but not unsaved.

> Alternatively, consider doing the project "copying" a bit more dynamically and on an as-needed basis. For example, once a precipitation slider is changed for a scenario, have that scenario store it's current attributes in a temporary attribute. When the page is moved back to model, then apply the saved attributes to the changed scenario. I haven't used it much but model.previousAttributes and/or model.previous() may be of some use here.

I was thinking of doing that, but I think it makes the code more complicated. I'll think it about some more though.

lewfish:
Any ideas about what to do about the Water Quality chart getting unusably squished when the width of the screen gets too small?

screen shot 2015-08-11 at 4 39 31 pm 2

mmcfarland:
Can we reorient the chart to vertical and angle the labels up at ~45d?

src/mmw/js/src/compare/views.js
```diff
@@ -164,14 +152,77 @@ var CompareScenariosView = Marionette.CompositeView.extend({
     }
 });
 
-var CompareChartView = Marionette.ItemView.extend({
-    template: compareChartTmpl,
+var CompareModelingView = Marionette.LayoutView.extend({
+    //model: modelingModels.ScenarioModel
```
lliss: Left over?

lewfish: No, I've been putting the model property as a comment in views as a reminder of the type of model associated with a view. If you leave it uncommented, it sometimes creates a problem, although I can't remember when/why. Also, having it as a comment makes sense because that's what it essentially is, since the property is overridden by what is passed to the constructor.

lewfish:
Re: the bar chart getting squished, what about using a stack bar chart? It would take less work and has the advantage of being both vertical and horizontal space efficient?

src/mmw/js/src/modeling/templates/scenarioToolbarTabContent.html
```diff
@@ -1,41 +1,43 @@
...
-            {% else %}
-                <button class="btn btn-sm btn-default inverted" type="button" data-toggle="dropdown" aria-expanded="true">
+{% if not compareMode %}
+    <div id="modification-btn-wrapper"> <!-- Modifications Dropdown -->
+        <div class="dropdown"> <!-- Dropdown Button -->
+            {% if not is_current_conditions %}
+                {% if shapes|length == 0 %}
+                    <button class="btn btn-sm btn-default inverted" type="button" disabled="true">
+                {% else %}
+                        <button class="btn btn-sm btn-default inverted" type="button" data-toggle="dropdown" aria-expanded="true">
```
lliss: Extra indentation here.

src/mmw/js/src/compare/templates/compareModeling.html
```diff
@@ -0,0 +1,19 @@
...
+    <i class="fa fa-circle-o-notch fa-spin"></i>
+{% endif %}
+<div class="pull-right">
+    <select class="form-control btn btn-small btn-primary">
+        {% for result in results %}
+            {% if result.active %}
+                <option value="{{ result.name }}" selected>
+            {% else %}
+                <option value="{{ result.name }}" >
+            {% endif %}
```
rajadain: This entire `if` block can probably be simplified using an [`if` expression](https://mozilla.github.io/nunjucks/templating.html#if-expression):

```jinja
<option value="{{ result.name }}" {{ "selected" if result.active }}>
```

src/mmw/js/src/modeling/tr55/runoff/views.js
```diff
@@ -45,9 +46,9 @@ var ResultView = Marionette.ItemView.extend({
                     depDisplayNames: ['Infiltration', 'Runoff', 'Evaporation']
                 };
 
-            if (this.scenario.get('is_current_conditions')) {
+            if (this.scenario.get('is_current_conditions') || this.compareMode) {
                 data = [
-                    getBarData('', 'unmodified'),
+                    getBarData('', 'modified'),
```
rajadain: Is the reasoning here that both `unmodified` and `modified` are identical for Current Conditions so it doesn't matter for them, and by choosing `modified` as the single bar-chart we ensure that when in `compareMode` the chart reflects the modifications?

lewfish: Yes, I'll put in a comment to that effect.

src/mmw/js/src/compare/templates/compareModeling.html
lewfish: Thanks, I didn't know about if expressions.

caseypt:
On the modeling page, the water quality table is missing the data and the formatting looks a little off:

`develop`
screen shot 2015-08-12 at 1 32 17 pm

`lewfish-feature/compare-run-model`
screen shot 2015-08-12 at 1 32 58 pm

caseypt:
The water quality stacked bar chart on the compare page looks good.

caseypt:
On the compare page, after adjusting the precipitation for a scenario, the chart get's pushed down by the modeling spinner and moved partially behind the perception slider.

In this screenshot, the precipitation was not adjusted for the left scenario, but was adjusted for the right scenario:

screen shot 2015-08-12 at 1 35 48 pm

caseypt:
While the stacked bar chart fits better on the compare scenario panels, the proportions never change, only the amounts of each due, which just changes the legend. It's a little hard to catch this.

If this isn't because of dummy data, we may want to figure out a better solution.

caseypt:
Verified that the water quality table is back.

lewfish:
^^ I agree that it's hard to tell when the proportions never change. It's a general problem we've been having, and can be seen in the modeling view, so I think we should make a new issue for it.

caseypt:
Clicking on a modification on a map in the compare view and deleting the modification removes the modification from the scenario and leaves the compare view map in a weird state:

screen shot 2015-08-12 at 1 52 13 pm

We may want to disable interactivity on these maps or remove the trashcan button from the modification pop-up.

caseypt:
If you are on on the compare page, and are logged in, when you log out, the url is updated to `http://localhost:8000/` but the page does not navigate there.

lewfish:
^^ For me, it navigates to http://localhost:8000/ and shows the home page with the AOI still showing and the analyze view minimized. That's also what happens when you log out while in the modeling view. I think this is the desired behavior, but maybe not.

lewfish:
Good point about disabling interactivity on the maps.

caseypt:
Re: the routing, false alarm. I'm now seeing the behavior you described, and can't reproduce the behavior I was initially seeing.

src/mmw/js/src/modeling/models.js
```diff
@@ -42,11 +42,12 @@ var Tr55TaskModel = coreModels.TaskModel.extend({
 
 var ResultModel = Backbone.Model.extend({
     defaults: {
-        name: '',
-        displayName: '',
-        inputmod_hash: null,
-        result: null,
-        polling: false
+        name: '', // Code name for type of result, eg. runoff
+        displayName: '', // Human-readable name f or type of result, eg. Runoff
```
caseypt: Typo: `f or` -> `for`

caseypt:
+1. Followed the testing instructions and did a bunch of other tests, and things are generally things are working well. Nice job.

lewfish:
Some new issues prompted by this PR:
https://github.com/WikiWatershed/model-my-watershed/issues/685
https://github.com/WikiWatershed/model-my-watershed/issues/686

Issue #648, opened 2015-07-30
Saved projects don't persist area name


mmcfarland:
Footer label for Area of Interest name is set correctly when starting from the Draw screen, but when moving back from /project the name is lost.

PR #674, opened 2015-08-10
Save the AOI name with the project.

lliss:
We had not been saving the AOI name when saving the project. As a result, if
you reloaded a project and then went back to analyze mode, the name would be
blank. This change allows the AOI name to be persisted as part of the modeling
project and then restores it to the map when the project is reloaded.

Connects #648 

To test:
 * Run migrations
 * Log in and start a project. Note the AOI name on the analyze view.
 * After the project is saved, refresh the page.
 * Hit the back button in the UI to get you to the analyze view.
 * Ensure the name above the area calculation for the AOI matches what you'd previously seen.

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -71,13 +71,14 @@ var ProjectModel = Backbone.Model.extend({
 
     defaults: {
         name: '',
-        created_at: null,          // Date
-        area_of_interest: null,    // GeoJSON
-        model_package: '',         // Package name
-        scenarios: null,           // ScenariosCollection
-        user_id: 0,                // User that created the project
-        is_activity: false,        // Project that persists across routes
-        needs_reset: false         // Should we overwrite project data on next save?
+        created_at: null,            // Date
+        area_of_interest: null,      // GeoJSON
+        area_of_interest_name: null, // Human readable string for AOI.
+        model_package: '',           // Package name
+        scenarios: null,             // ScenariosCollection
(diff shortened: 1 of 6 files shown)


src/mmw/js/src/modeling/models.js
```diff
@@ -71,13 +71,14 @@ var ProjectModel = Backbone.Model.extend({
...
-        is_activity: false,        // Project that persists across routes
-        needs_reset: false         // Should we overwrite project data on next save?
+        created_at: null,           // Date
+        area_of_interest: null,     // GeoJSON
+        model_package: '',          // Package name
+        scenarios: null,            // ScenariosCollection
+        user_id: 0,                 // User that created the project
+        is_activity: false,         // Project that persists across routes
+        needs_reset: false,         // Should we overwrite project data on next save?,
+        area_of_interest_name: null // Human readable string for AOI.
```
rajadain: Since the ordering of these fields doesn't matter, can we move `area_of_interest_name` right below `area_of_interest` so they are close together? I also liked that `needs_reset`, which doesn't have a corresponding database field, was at the very end, after `is_activity` a related field.

rajadain:
+1 tested, works perfectly. Tried with both UI and browser back and forward buttons. Great work!

I like that the introduced changes were minimal, but I agree with @lewfish that we may take this opportunity to add some DRYness to the code.

rajadain:
+1, edits look great!

caseypt:
I'll pull this down and rebase it.

caseypt:
Superseded by #688

Issue #668, opened 2015-08-05
Better loading of existing projects in Embed Mode


rajadain:
After #613, we have the following use-cases:

 1. User opens the app from inside an activity, and begins a new project. The user can start interacting immediately.
 2. User opens the app from inside an activity, expecting to resume work on an existing project. While the app is waiting for a `loadInteractive` call from the parent, we automatically create a new project (since the app is loaded in Embed Mode), and if the user starts interacting before the project is loaded, then they will lose their work in a jarring way when the project finally does load.

We cannot change the behavior via URL because that cannot be changed once set, and must support both use cases.

Implement a loading screen that only shows when running in Embed Mode, that stays up for around 1 - 2 seconds which gives the app time to load and listen for the `loadInteractive` call. The user should not be able to interact with the app at all while this screen is up. If the app receives a `loadInteractive` call the load screen should go away, or it should go away by itself after the set timeout. The screen should only display when on the `/` route, not on `/analyze` or `/project` or anything else.

rajadain:
I'm finding that if we change the re-route we do for loading existing projects from a full page refresh:

```javascript
window.location.href = interactiveState.url;
```

to a Backbone route navigation:

```javascript
router.navigate(interactiveState.route);
```

The loading is much snappier and removes the need for a load screen. Unfortunately, we redirect before the "New Activity" project is saved, thus we don't delete it, but it does indeed get saved after the redirect. One way to address this is to add another flag, something like `loadingExistingActivity`, and have the [`enableSingleProjectModeIfActivity`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/draw/controllers.js#L66) method check for that flag before creating / saving.

Further thought and learning about how educators are going to design activities makes it seem like there may be sessions where the learners focus completely on the Draw view, and don't move on to Analyze view at all until a later session.

We should refactor the `interactiveState` object to include a project ID in addition to a route, load the appropriate project, and then navigate to the appropriate route. We may consider adding more save-points to Activity Mode so that changes to the area of interest cause saves.

PR #675, opened 2015-08-10
ITSI Improved Session Resumption

rajadain:
## Overview

Previously there were two issues with loading existing projects in ITSI Embed mode:

 1. Loading took too long, a user could begin interacting with the app before loading completed, and could interfere with the saved state.
 2. It was not possible to save and load a project that hadn't reached the modeling stage. We were not saving a project with only an AoI at the draw stage, and there was no way to resume from this stage.

This PR addresses both concerns by doing the following:

 1. Switches to Backbone routing instead of full page refresh which is orders of magnitude faster and does not let the user interact with the app before project is loaded.
 2. Adds saving and loading mechanism to the draw stage.

### Prepare an Activity for Testing

 1. Checkout the branch and run the bundle `./scripts/bundle.sh --debug`
 2. Run [ngrok](https://ngrok.com/) `ngrok http 8000`
 3. Log on to [ITSI](https://learn.staging.concord.org) as `mmwterence` and edit the [**MMW Example 6abdbbf1**](https://learn.staging.concord.org/eresources/506/matedit?iFrame=true) activity
 4. Update the Activity Name to have the URL of your ngrok instance for identification purposes and click "Save":  
  image
 5. Edit the first page of the activity:  
  image
 6. Edit the embedded app:  
  image
 7. Update the name and URL to match your ngrok instance:  
  image
 8. Click "Save"

### Test the PR

 1. Log in as `tstudent2`, and open the activity.
 2. Draw an AoI.
 3. Close the tab. Open the activity again. It should be initialized to the draw view with the correct AoI.
 4. Proceed to Analyze. Proceed to Model.
 5. Close the tab. Open the activity again. It should be initialized to the project view.

Connects #668

### DIFF
--- a/src/mmw/js/src/app.js
+++ b/src/mmw/js/src/app.js
@@ -21,7 +21,7 @@ var App = new Marionette.Application({
 
         // Initialize embed interface if in activity mode
         if (activityMode) {
-            this.itsi = new itsi.ItsiEmbed();
+            this.itsi = new itsi.ItsiEmbed(this);
         }
 
         // This view is intentionally not attached to any region.
--- a/src/mmw/js/src/core/itsiEmbed.js
+++ b/src/mmw/js/src/core/itsiEmbed.js
@@ -1,22 +1,23 @@
 "use strict";
 
 var _ = require('lodash'),
+    Backbone = require('../../shim/backbone'),
     iframePhone = require('iframe-phone'),
-    App = require('../app.js'),
+    router = require('../router.js').router,
     EMBED_FLAG = 'itsi_embed',
     QUERY_SUFFIX = EMBED_FLAG + '=true';
 
-var ItsiEmbed = function() {
+var ItsiEmbed = function(App) {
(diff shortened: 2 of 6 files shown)


rajadain:
Please hold off on reviewing this until after today's demo.

rajadain:
Okay this is ready for review now. Embed URL https://e0ebfe64.ngrok.io/?itsi_embed=true

lewfish:
+1, tested.

mmcfarland:
Also tested in demo yesterday, worked well and everyone was pleased with the progress/result.

rajadain:
Encapsulated ITSI notification in `updateItsiFromEmbedMode`, and now nullify `App.itsi` if it exists from `testUtils.resetApp`, according to comments on https://github.com/rajadain/model-my-watershed/commit/e3b81d2469b8dfc3118bf30a4e675aaa1b8fcaba. Will merge shortly.

PR #677, opened 2015-08-11
Allow AMI lookup to filter by users with execute bit

hectcastro:
By default, AMI lookups were occurring based on account ownership. This doesn't work for AMI sharing because the account making use of shared AMIs doesn't own them, instead it only has execute privileges. This updates the filter so that if the account looking up AMIs does not own any AMIs, it'll fall back to looking for AMIs it can execute.


lliss:
+1. Looks good to me and I have seen this successfully provision a stack.

PR #683, opened 2015-08-12
Update ALLOWED_HOSTS setting

hectcastro:
The new staging environment is under a different domain, so we need to add that domain to Django's `ALLOWED_HOSTS` to avoid HTTP 400 errors.


caseypt:
+1

Issue #481, opened 2015-06-19
Add "Alert" to Micro Site for Excessive Runoff


chitchens:
It would be a nice enhancement to replicate the feature of the previous Simple Model where there was a visual indication of "alarm" when the settings result in excessive runoff.

PR #670, opened 2015-08-06
Animate Arrows in Micro Site

rajadain:
## Overview

This builds on top of work by @ctaylo37 in #663 to add animations to the arrows depending on input values. The formula used for calculating arrow sizes has been taken from the original Flash app, but can be tweaked herein as deemed aesthetically necessary.

## Testing Instructions

 1. Checkout the branch
 2. Run:  

        $ ./scripts/manage.sh collectstatic
        $ ./scripts/bundle.sh

 3. Open the [`/micro`](http://localhost:8000/micro) site
 4. Test on various browsers and devices

Connects #480 
Connects #663 
Connects #666

PR #684, opened 2015-08-12
Added alert image and transition to water column.

ctaylo37:
This will need wired up to the model.

Connects #481


caseypt:
Connects to #481

caseypt:
**Testing Instructions:**

- Select different land uses and soil types, and adjust the precipitation slider. 
- Ensure that any time runoff is < 2 inches, the runoff alert icon is not shown, and any time runoff is > 2 inches, the alert icon is shown.

Connects to #481.

rajadain:
Works really well in Chrome, I like the breathing animation a lot!

A "nice-to-have" would have been if the alert icon animates in and out, instead of just popping in to existence. Another minor design issue is that the alert icon is stuck a the top of the bar-chart, even when ET has a non-zero value and runoff starts after it, thus obscuring the green column:

No Evapotranspiration:
image

Non-Zero Evapotranspiration:
image

In Firefox, the icon isn't center aligned like it is in Chrome, and creates a horizontal scroll bar:
image

In Safari, the alignment of the icon is correct, but the breathing animation is absent. ~~This is likely due to an issue discovered in #670, where we had to animate the height of the SVG element itself, instead of its containing `div`.~~ Just tried that out, didn't work. Will need more fiddling. Since one of the target use cases of the micro site is the iPad, which uses Mobile Safari, we should consider fixing this.

caseypt:
@ctaylo37 Do you have time to address @rajadain's comments today?

ctaylo37:
Unfortunately not, but @rajadain said he was going to be able to take care of it.

rajadain:
Fixed Safari animation and Firefox width issues. Tried adding animations to appear / disappear of the alert, and to dynamically adjust `top` of the icon so as not to obscure the ET column, but both are non-trivial tasks. This is primarily because CSS animations take 0.3s to finish, but our code executes too fast and thus acts before CSS height has reached its final value. Will either have to build in the delay or try something else.

I'm willing to let that slide for now and come back to it at a later time. The more problematic issues have been fixed.

caseypt:
+1. Icon looks good in Safari and Firefox.

Noticed this Firefox though: 
screen shot 2015-08-13 at 1 26 34 pm

Are you seeing that too @rajadain? If so, I'll create an issue.

rajadain:
@caseypt Yes that's been there for a while. Please do.

Issue #681, opened 2015-08-11
Don't create new project when loading existing in Embed Mode


rajadain:
When loading an existing project in Embed Mode, we have to wait for a `loadInteractive` postMessage from the parent container #563. Unfortunately, by that time a new project has already been created and saved for the user #561. Currently, we delete this temporary project before loading the existing one #613. This leads to a large number of "dead" project IDs which are never associated with a real project, because they are deleted as soon as they are created.

Refactor the project creation / saving so that the project is not saved until a later state, such as when an AoI is available or when the user has arrived at the Analyze view.

The work for this should build on top of #675.

PR #687, opened 2015-08-12
Defer saving projects in Embed Mode until AoI is available

rajadain:
## Overview

Previously we would save a new project as soon as the app initialized in embed mode. This had the effect of creating a number of empty projects when the app was loading an existing project. We would delete this empty project before loading the specified one.

Now, we still create the project as soon as the app starts, but defer saving it to the database until an Area of Interest has been drawn and the user lands in the Analyze view. This ensures that the project is always saved as soon as an AoI is available, which is fine since we're not tracking any other changes that we could restore (such as the maps zoom or location, or values in the search box).

This further simplifies the loading logic, since now we don't have to check for and delete an existing project. We simply reset `currProject` and navigate to given route, and let the controllers handle the rest.

## Testing Instructions

If an activity is not setup in ITSI, do so according to instructions in #675. Then:

 1. Log in to [ITSI](https://learn.staging.concord.org/) as `tstudent2` and open the activity
 2. Open Chrome dev tools, go to the Network tab, and filter by XHR and "ngrok":  
  image
 3. Star the interactive. Observe network traffic. No new project should be saved, and you should not see a `POST` to `/projects`
 4. Draw an AoI. Now the project should save, and you should see a `POST` to `/projects`. The interactive state should also be saved.
 5. Refresh the page and start the interactive. Observer network traffic. Your old project should load, but no new project should save.
 6. Proceed to the modeling stage. Your existing project should be saved with a `PUT`, and new scenarios should be created.
 7. Refresh the page and start the interactive. Observer network traffic. Your old project should load, but no new project should save.

## Notes

Once #674 gets merged, this code should be updated to [save](https://github.com/rajadain/model-my-watershed/blob/refactor/itsi-activity-saving/src/mmw/js/src/analyze/controllers.js#L30) and [load](https://github.com/rajadain/model-my-watershed/blob/refactor/itsi-activity-saving/src/mmw/js/src/modeling/controllers.js#L156) the Area of Interest Name.

Since now we only always save after an AoI is available, should we revert the database constraint change made in #610 that made AoI a nullable field?

Connects #681

### DIFF
--- a/src/mmw/js/src/analyze/controllers.js
+++ b/src/mmw/js/src/analyze/controllers.js
@@ -20,7 +20,29 @@ var AnalyzeController = {
             App.map.set('maskLayerApplied', true);
         }
 
-        if (!settings.get('activityMode')) {
+        if (settings.get('activityMode')) {
+            // Only one project allowed in Activity Mode. Save current project
+            // and if in embedded mode, update interactive state for container.
+            var project = App.currProject,
+                map = App.map;
+
+            if (project && project.get('scenarios').isEmpty()) {
+                project.set({
+                    'area_of_interest': map.get('areaOfInterest'),
+                    'area_of_interest_name': map.get('areaOfInterestName')
+                });
+                project
+                    .save()
+                    .done(function() {
+                        if (settings.get('itsi_embed')) {
+                            App.itsi.setLearnerUrl('project/' + project.id + '/draw');
+                        }
+                    });
+            }
+        } else {
+            // Multiple projects allowed in Regular Mode. Nullify current
+            // project since a new one will be created and saved by the
+            // Modelling Controller.
             App.currProject = null;
         }
     },
--- a/src/mmw/js/src/core/itsiEmbed.js
+++ b/src/mmw/js/src/core/itsiEmbed.js
@@ -37,25 +37,13 @@ var ItsiEmbed = function(App) {
     };
 
     this.loadInteractive = function(interactiveState) {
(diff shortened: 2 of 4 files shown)


rajadain:
The ITSI activity is currently setup and running from my machine. If you test this tomorrow, then I will re-setup the activity and comment here when ready.

mmcfarland:
Behavior looks good, network traffic corresponds to the save/do-not-save workflow described.

+1

rajadain:
Rebased on top of `develop` and now save AoI name since #688 got merged.

src/mmw/js/src/analyze/controllers.js
```diff
@@ -20,7 +20,29 @@ var AnalyzeController = {
             App.map.set('maskLayerApplied', true);
         }
 
-        if (!settings.get('activityMode')) {
+        if (settings.get('activityMode')) {
+            // Only one project allowed in Activity Mode. Save current project
+            // and if in embedded mode, update interactive state for container.
+            var project = App.currProject,
+                map = App.map;
```
caseypt: Are these here to make referencing the variables later shorter?

rajadain: Yes. `project` is referred to a number of times, but `map` only twice. The second is a bit of a relic since this code was originally came from a place where `map` was a function parameter, now we source it from `App`.

caseypt:
Untested, but the code looks good. +1.

PR #703, opened 2015-08-14
Put CompareModelingView header in div

lewfish:
So that the "Model Output" text will stay on the same level as the results dropdown after polling for results.


rajadain:
+1, tested in Chrome, Firefox and Safari. The dropdown stays where it starts out and the container div does not expand.

caseypt:
+1. Tested.

Issue #705, opened 2015-08-17
Failing Test in Chrome / Safari


rajadain:
The test [modeling/tests.js:Modeling/Models/ProjectModel/#parse](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/tests.js#L505-L523) fails in Chrome and Safari when running `./scripts/testem.sh` but not in Firefox. Since `./scripts/test.sh` uses Firefox for testing, this error is not caught.

The error is:

```
Modeling Models ProjectModel #parse correctly constructs the nested project model from the server response 
    ✘ Uncaught TypeError: Fake server request processing threw exception: Cannot assign to read only property 'value' of [ (http://localhost:7357/js/test.vendor.js:11558)
        Error: Uncaught TypeError: Fake server request processing threw exception: Cannot assign to read only property 'value' of [ (http://localhost:7357/js/test.vendor.js:11558)
            at global.onerror (http://localhost:7357/testem/mocha.js:5462:10)
```

The line causing errors is [modeling/models.js#464](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/models.js#L464):

```javascript
// Change old names to new ones.  This could eventually be
// removed when enough of old scenarios have been upgraded.
_.forEach(attrs.modifications, function(m) {
    m.value = oldToNewName(m.value);
});
```

Fix the test.

PR #706, opened 2015-08-17
Unquote arrays to fix test

rajadain:
## Overview

The quotes were causing some tests to fail in Chrome and Safari, since the value `"[]"` would be parsed as a string not an array, but not in Firefox which automatically casted the string to an empty array.

## Testing Instructions

Run `./scripts/testem.sh` in Chrome and Firefox, ensure that all tests pass.

Connects #705


caseypt:
Were you seeing consistent failures for these tests? I've run them a couple times, and haven't seem them fail at all.

caseypt:
In any case, all the tests pass on this branch for me in Firefox, Chrome, Safari, and from the command line. +1

rajadain:
Yes on `develop` I'm still seeing it:

image

image

On Chrome for Mac 44.0.2403.155 (64-bit)

caseypt:
Same version here. Weird.

Issue #695, opened 2015-08-13
Align the color scheme between micro R/ET/Inf and model output


mmcfarland:
Favor the color scheme in micro and apply it to the main web app R/ET/Inf charts.

Issue #549, opened 2015-07-06
Fix graph label


rajadain:
What is currently labeled as "Evaporation" in the TR-55 Runoff Model Output should probably be called "Evapotranspiration" as it is in the Micro app:

![image](https://cloud.githubusercontent.com/assets/1430060/8525948/aefcb20c-23d1-11e5-902d-f201e58b3c58.png)

![image](https://cloud.githubusercontent.com/assets/1430060/8525978/cef510f4-23d1-11e5-97ac-15f7ccde6500.png)

PR #707, opened 2015-08-17
Fix Runoff colors, labels

rajadain:
Change colors and labels to match the values in the micro site:

image

Connects #695 
Connects #549


lewfish:
The water quality stacked bar chart in the compare view needs to be changed as well. You'll have to find a fourth color that looks good with those three.

rajadain:
I've left that chart as it is right now, since they represent different values, and adds to the visible differentiation:

image

But I can make them the same colors if that is deemed appropriate.

jamesmcclain:
+ 1 Tested.

Just out of curiosity, where did this color scheme come from?

rajadain:
@jamesmcclain I took the colors from the Micro Site, which I think was designed by @ctaylo37.

jamesmcclain:
Autumn colors

Issue #691, opened 2015-08-13
Safari makes multiple "Get Your Location" requests


mmcfarland:
Users are reporting that Safari will request to gain access to your location multiple times, sometimes being satiated after 3 times, but often occurring infinitely, requiring a Force Stop.

Reported version:
Safari  Version 8.0 (10600.1.25.1)

caseypt:
Verified this is happening in 8.0.7. Stuck in an infinite loop.

PR #708, opened 2015-08-18
Fix Safari infinte location request loop.

lliss:
In some cases, it seems requesting geolocation at page load will cause Safari
to continue to ask for location permission in an infinite loop.
https://stackoverflow.com/questions/27150465/
The solution to this seems to be triggering the location request based on some
user event or supplying a short delay to allow whatever is causing Safari the
issue time to resolve itself.

Connects #691 

To test:

 * Open the application in Safari
 * Click allow location
 * Ensure that this happens only once.

### DIFF
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -265,11 +265,17 @@ var MapView = Marionette.ItemView.extend({
                 maximumAge : maxAge,
                 timeout : timeout
             };
-            navigator.geolocation.getCurrentPosition(
-                geolocation_success,
-                _.noop,
-                geolocationOptions
-            );
+
+            // Wait a bit and then get the location. Suppresses issues on Safari
+            // in which immediately requesting the location results in an
+            // infinite request loop for geolocation permissions.
+            setTimeout(function() {
+                navigator.geolocation.getCurrentPosition(
+                    geolocation_success,
+                    _.noop,
+                    geolocationOptions
+                );
+            }, 500);
         }
     },
 


rajadain:
+1 tested. It was difficult to get Safari to do a hard refresh because of the constant prompts, but once it loaded the new JavaScript I was only prompted once. I am still prompted at each full page reload, unless I check the "Remember for one day" checkbox.

src/mmw/js/src/core/views.js
```diff
@@ -265,11 +265,17 @@ var MapView = Marionette.ItemView.extend({
...
+            // Wait a bit and then get the location. Suppresses issues on Safari
+            // in which immediately requesting the location results in an
+            // infinite request loop for geolocation permissions.
+            setTimeout(function() {
+                navigator.geolocation.getCurrentPosition(
+                    geolocation_success,
+                    _.noop,
+                    geolocationOptions
+                );
+            }, 500);
```
rajadain: Since this is an arbitrary number, consider calling it out as `GEOLOCATION_DELAY`or something where we define other defaults: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/views.js#L189-L191

lliss:
I ended up force quitting Safari and then opening it and clearing this history and local files before revisiting the page.

src/mmw/js/src/core/views.js
lliss: I'm not against this but it's not really and important part of the application that I think we need to tune. We just need some delay and I feel like adding a variable for it just adds a little more indirection without any real gain.

rajadain: True, and I guess the comment above does add context. Alright let's leave it as is.

caseypt:
+1. Tested.

Issue #696, opened 2015-08-13
Rename basemap layers to more conventional language


mmcfarland:
I'm thinking "Streets", "Satellite with Roads" and "Satellite" or similar.  Also add the Google Terrain option in.  In general, let's let the attribution of the layer on the map cite the provider and not use name brands in the list.

Issue #692, opened 2015-08-13
Maximize zoom levels by layer


mmcfarland:
Each baselayer should maximize the potential maximum zoom level so that users can digitize features on the map easily.  This may involve base map settings, layer specific options or the `maxNativeZoom` property if it looks respectable. Currently, the GMaps App layer and the actual Google Maps differ in available zoom 

![screenshot from 2015-08-13 11 54 56](https://cloud.githubusercontent.com/assets/1014341/9254564/2d8866f4-41b2-11e5-8c06-5f6e921918ed.png)

![screenshot from 2015-08-13 11 55 03](https://cloud.githubusercontent.com/assets/1014341/9254565/323f90c8-41b2-11e5-9076-c0c3454bea38.png)

PR #709, opened 2015-08-18
Use layer max zoom levels + layer name change

caseypt:
Uses the maximum zoom level of each layer instead of a fixed value for all. Also changes the layer names to be generic and adds the Google Terrain layer.

**Testing different maximum zoom level**
- Keep the default basemap.
- Center the map on City Hall.
- Switch to the "Satellite" basemap.
- You should be able to zoom one level closer.
- Switch back to the "Streets" basemap.
- You should be zoomed out one level further than before, and zooming in should not be available.
- Switch to the "Satelitte with Streets" basemap.
- You should be able to zoom in two levels.
- Switch back to the "Streets" basemap.
- You should be zoomed out two levels further than before, and zooming in should be not be available.

**Testing base layer name changes**
- You should see non-branded names in the layer selector, like so:

screen shot 2015-08-18 at 1 59 48 pm

Connects to #692 
Connects to #696

### DIFF
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -78,6 +78,7 @@ var HeaderView = Marionette.ItemView.extend({
 
     userLogout: function(event) {
         event.preventDefault();
+
         this.model.logout().done(function() {
             router.navigate('', {trigger: true});
         });
@@ -85,6 +86,7 @@ var HeaderView = Marionette.ItemView.extend({
 
     cloneProject: function() {
         event.preventDefault();
+
         var view = new modalViews.InputView({
             model: new modalModels.InputModel({
                 title: 'Clone Project',
@@ -111,7 +113,6 @@ var HeaderView = Marionette.ItemView.extend({
         });
         view.render();
     }
-
 });
 
 // Init the locate plugin button and add it to the map.
@@ -192,6 +193,7 @@ var MapView = Marionette.ItemView.extend({
             self = this;
 
         this.interactiveMode = options.interactiveMode;
+
         if (!options.interactiveMode) {
             // Disable panning and zooming.
             // http://gis.stackexchange.com/questions/54454/disable-leaflet-interaction-temporary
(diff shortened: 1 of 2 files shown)


caseypt:
Rebased on `develop`.

caseypt:
Tests fixed.

rajadain:
New sessions start out with the "Streets" basemap at max zoom to my current location, as opposed to the current behavior:

image
image

Is this desirable?

caseypt:
That behavior is because we set the `maxZoom` to 10 when we first added the map: https://github.com/WikiWatershed/model-my-watershed/commit/330b5542a1301791cfcc715f70e0b7d926d4b308.

It's now set to the actual max zoom of the layer.

I don't have a strong preference on where it should be set, so happy to defer to others.

I do think we should get rid of this line though or modify it to center on the US: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/views.js#L240

src/mmw/js/src/core/views.js
```diff
@@ -461,12 +483,53 @@ var MapView = Marionette.ItemView.extend({
...
+
+        this._googleMaxZoomService.getMaxZoomAtLatLng(latLng, function(response) {
+            if (response.status !== self._googleMaps.MaxZoomStatus.OK) {
+                // Leave the max layer zoom as is
+                return;
+            } else {
+                // Set layer zoom level to the max for the current area
+                _.each(self.baseLayers, function(layer) {
+                    // Needs to be updated if other Google Base maps are added
+                    if (layer._type === 'HYBRID') {
```
mmcfarland: Should the Terrain option be listed here as well?  Maybe all `layer` types that are of Google should get an additional attribute set so this conditional won't need to be updated?  Like `layer.isGoogle`?

caseypt: As far as I can tell from the [Google docs](https://developers.google.com/maps/documentation/javascript/3.exp/reference#MaxZoomService), it's just satellite based layers. Unfortunately, we can't add any custom properties to the Leaflet layer object, but I'll add the other map type that uses imagery.

mmcfarland: Got it - I was just going off of the comment.

mmcfarland:
Re: default zoom and location, I agree, I think a default of regional scale - similar to where it was originally is a pretty good level.  I don't think eastern PA is a good starting point either.  Somewhere central seems good, and then most people will geolocate to their location.

src/mmw/js/src/core/views.js
caseypt: Not a great comment, considering adding support for the other layer was less characters.

caseypt:
Update with support for both satellite-based Google layers, as well as new setting for the default extent and new zoom level for geolocating.

Default extent:

screen shot 2015-08-19 at 3 11 57 pm

Geolocation zoom level:

screen shot 2015-08-19 at 3 11 25 pm

mmcfarland:
Looks good.  Let's hope it doesn't encourage them to turn on HUC12 at that scale...

rajadain:
+1

Issue #697, opened 2015-08-13
Re-evaluate color choices for CP SVG items


mmcfarland:
Try to align the conceptual Conservation Practice with a reasonable real-world color.  For instance, Green Roof is currently blue, but could instead be Green.

PR #714, opened 2015-08-19
Modify conservation practice colors

caseypt:
* Use colors for the conservation practice tools that may better match
what those practices look like in the real world, without too closely
colliding with colors from the NLCD standard legend.

Connects to #697 

screen shot 2015-08-19 at 4 44 46 pm
screen shot 2015-08-19 at 4 45 01 pm

### DIFF
--- a/src/mmw/apps/core/templates/patterns.html
+++ b/src/mmw/apps/core/templates/patterns.html
@@ -4807,7 +4807,7 @@
         <!-- No Till farming -->
         <pattern height="74.8" id="fill-no_till" overflow="visible" patternUnits="userSpaceOnUse" viewBox="72.15 -74.8 72 74.8" width="72">
             <g>
-                <polygon fill="#53e9b4" points="72.15,-74.8 144.15,-74.8 144.15,0 72.15,0"/>
+                <polygon fill="#CDCD00" points="72.15,-74.8 144.15,-74.8 144.15,0 72.15,0"/>
                 <g>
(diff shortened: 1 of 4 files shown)


caseypt:
Open to suggestions.

caseypt:
Tagging @rajadain as well.

rajadain:
Unsure if there is ever really going to be a satisfactory solution to this, but Cluster Housing is hard to see over Map Box

image

And Porous Paving is hard to see over Google Hybrid

image

One way of solving such a problem is to swap their border colors. This way, both of them have dark and bright colors, one for fill and one for stroke, and will be visible on both dark and light basemaps.

For the record, they're both pretty visible on ESRI World Imagery

image

+1 on the color choices. I really like them and they make a lot of intuitive sense.

lliss:
+1

caseypt:
@rajadain Updated those two practices based on your suggestion.

screen shot 2015-08-21 at 9 50 49 am
screen shot 2015-08-21 at 9 50 55 am
screen shot 2015-08-21 at 9 51 01 am

caseypt:
Also added a commit to make modifying the colors a bit clearer in the future.

lliss:
Failed because of this test:
```
not ok 82 Firefox 39.0 - Modeling modificationConfigUtils #getDrawOpts returns the provided colors
    ---
        message: >
            expected '#888' to equal '#000'
```

caseypt:
@lliss Thanks. Didn't even think to check the tests.

rajadain:
+1 the new colors look good! Tried on all four basemaps, and they are much more accessible.

Issue #682, opened 2015-08-11
Add Boundary Type to Label


chitchens:
After a boundary polygon is selected, its name appears on the bar below the map. In addition to the HUC or congressional district name, please also add the boundary type. For example, 

We currently have names like:
  * PA-06
  * East Branch White Clay Creek

It would be helpful if they could be more like:
  * Congressional District: PA-06
  * HUC 12: East Branch White Clay Creek

PR #718, opened 2015-08-24
Display layer type with named boundary feature

mmcfarland:
When using a named boundary feature, always display a short contextual
name for the feature type.

Connects #682


caseypt:
+1. Tested. Saved and restored a project and the prefix was still there.

screen shot 2015-08-25 at 9 31 43 am
screen shot 2015-08-25 at 9 32 02 am

lliss:
+1. Tested.

Issue #719, opened 2015-08-25
Target area for selecting a boundary is incorrect


hectcastro:
On staging, clicks have to occur on the leftmost or rightmost edges of the dropdown in order for its events to trigger:

<img width="313" alt="model_my_watershed" src="https://cloud.githubusercontent.com/assets/43639/9475418/97cf2486-4b34-11e5-80d8-f3fc45d1be38.png">

caseypt:
This may be due to new CSS rules or event binding changes introduced by https://github.com/WikiWatershed/model-my-watershed/pull/715

caseypt:
You can also click in the space between the name and question mark.

PR #721, opened 2015-08-25
Allow boundary dropdown to be clicked.

lliss:
Boundary layers were being triggered based on the data attribute of the
clicked element in the dropdown. But since spans were added to the text for
styling this broke the existing functionality. By changing the click handler
to work with the currentTarget (the specific element the event is bound to) we
will always capture the proper element with the data attribute in question.
Also added bounding to the popovers to ensure they stay within the bounds of
the main area.

Connects #719


rajadain:
+1 tested. Works perfectly.

Issue #722, opened 2015-08-26
Sync output type display in compare view


mmcfarland:
While in compare view, the display type (Water Quality vs Runoff) should be synced against all scenarios.

PR #734, opened 2015-08-28
Synchronize output type in compare view.

jamesmcclain:
When the user changes the output display type (runoff vs. quality) of one scenario in the compare view, all of them change.

**To Test**
   * Go to the compare view and change the display type a few times

Connects #722


rajadain:
+1 tested, works very well.

I wish Marionette offered a better way of accessing parent members from child views, which would simplify a lot of the passing around we have to do now, but it is fairly minimal and achieves the desired effect. Good job.

jamesmcclain:
Many thanks.

Issue #730, opened 2015-08-27
Add school district layer


caseypt:
- Find a school district shapefile for the whole U.S.
- Load it as part of the `setupdb` script.
- Include it in the new layer control that is being developed in #629 

Depends on #629

Issue #629, opened 2015-07-29
Add overlay layers and control


mmcfarland:
Initially, let's reuse the 4 existing boundary layers + a new layer for school districts (we need to source this data).

We should merge the basemap control this control with input from @ctaylo37

mmcfarland:
These layers are merely for display and don't need initially need utfgrid support initially.  Overlays should be available in all draw/analyze/model maps windows.

caseypt:
Re: school districts, are we looking for a dataset that covers the entire U.S.?

mmcfarland:
> Re: school districts, are we looking for a dataset that covers the entire U.S.?

Yes, we were told it exists nationally.

caseypt:
School district work should be done in #730

PR #735, opened 2015-08-31
Load school districts

mmcfarland:
Update setup script to load a national school district boundary layer.

A new script option `-f` allows you to load up a single, name layer and not have to reload all boundary data sets to add a new one.  For example, from an app server you could run

`./setupdb.sh -f boundary_school_district.sql.gz`

to add just this new dataset.

The data set was sourced as a shapefile from http://nces.ed.gov/programs/edge/geographicDistrictBoundary.aspx and originally processed into PostGIS with the following settings:

```bash
shp2pgsql -s 4269:4326 -d -g geom -I -W "LATIN1" schooldistrict_sy1314_tl14 public.boundary_school_district
```

##### Testing
From you local app server run either 
* `./setupdb.sh -f boundary_school_district.sql.gz` or `./setupdb.sh -d` to load the new dataset
* Test that the tiles can be generated by requesting http://localhost:4000/district/10/300/387.png

Connects #730 and will be used in #629

### DIFF
--- a/scripts/aws/setupdb.sh
+++ b/scripts/aws/setupdb.sh
@@ -9,32 +9,33 @@ usage="$(basename "$0") [-h] [-b] [-s] \n
 where: \n
     -h  show this help text\n
     -b  load/reload boundary data\n
+    -f  load a named boundary sql.gz\n
     -s  load/reload stream data\n
 "
 
 # HTTP accessible storage for initial app data
 FILE_HOST="https://s3.amazonaws.com/data.mmw.azavea.com"
 load_boundary=false
+file_to_load=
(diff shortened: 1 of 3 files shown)


caseypt:
:+1: 

screen shot 2015-08-31 at 5 07 48 pm

caseypt:
Also, thanks for not messing with the app layer settings. I'll add this layer in the PR I'm working on for #629.

lliss:
+1

mmcfarland:
Yeah, I knew you were rearranging the guts of it and didn't want to get involved.

Issue #733, opened 2015-08-27
Table output should use same threshold as AoI area for units


mmcfarland:

mmcfarland:
When the AoI size reaches > 1km  (or similar), the display units switch from sq m to sq km.  The table output units should use the same units at the same threshold

PR #736, opened 2015-08-31
Unit thresholding in /analyze view

jamesmcclain:
The table is now displayed in units of square kilometers when one or more <use,soil> pairs occupies at least one square kilometer. Otherwise, the table is displayed in units of square meters.

Connects #733

**To Test**
   * Load the `/analyze` page
   * Observe that when the <use,soil> pair with greatest area has area of one square kilometer or greater, the table is displayed in units of square kilometers, otherwise it is displayed in units of square meters.
   * In order to make this observation, it may be necessary to edit the `run_analyze` function in the file `src/mmw/apps/modeling/tasks.py`.

**Square meters:**
screenshot from 2015-08-31 11 28 23

**Square kilometers:**
screenshot from 2015-08-31 11 32 39

### DIFF
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -1,12 +1,14 @@
 "use strict";
 
 var $ = require('jquery'),
+    _ = require('lodash'),
     Marionette = require('../../shim/backbone.marionette'),
     App = require('../app'),
     models = require('./models'),
     coreModels = require('../core/models'),
     coreViews = require('../core/views'),
     chart = require('../core/chart'),
+    utils = require('../core/utils'),
     windowTmpl = require('./templates/window.html'),
     messageTmpl = require('./templates/message.html'),
     detailsTmpl = require('./templates/details.html'),
@@ -196,11 +198,14 @@ var TabContentView = Marionette.LayoutView.extend({
         chartRegion: '.analyze-chart-region'
     },
     onShow: function() {
-        var dataCollection = new coreModels.DataCollection(
-            this.model.get('categories')
-        );
+        var categories = this.model.get('categories'),
+            largestArea = _.max(_.pluck(categories, 'area')),
+            units = utils.magnitudeOfArea(largestArea),
+            dataCollection = new coreModels.DataCollection(categories);
 
         this.tableRegion.show(new TableView({
+            units: units,
+            model: new coreModels.GeoModel({units: (units === 'km2') ? 'km<sup>2</sup>' : 'm<sup>2</sup>'}),
             collection: dataCollection
         }));
 
@@ -222,16 +227,26 @@ var TabContentsView = Marionette.CollectionView.extend({
 var TableRowView = Marionette.ItemView.extend({
     tagName: 'tr',
     template: tableRowTmpl,
-        templateHelpers: function() {
+    templateHelpers: function() {
+        var area = this.model.get('area'),
+            units = this.options.units;
+
         return {
             // Convert coverage to percentage for display.
-            coveragePct: (this.model.get('coverage') * 100).toFixed(1)
+            coveragePct: (this.model.get('coverage') * 100),
+            // Scale the area to display units.
+            scaledArea: utils.changeOfAreaUnits(area, 'm2', units)
         };
     }
 });
 
 var TableView = Marionette.CompositeView.extend({
     childView: TableRowView,
+    childViewOptions: function() {
+        return {
+            units: this.options.units
+        };
+    },
     childViewContainer: 'tbody',
     template: tableTmpl
 });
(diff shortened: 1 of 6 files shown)


mmcfarland:
There's some duplication here with GeoModel - perhaps this could be combined so we only have this kind of code in one place.

https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/models.js#L186

jamesmcclain:
> There's some duplication here with GeoModel - perhaps this could be combined so we only have this kind of code in one place.

Okay, will-do.

src/mmw/js/src/analyze/views.js
```diff
@@ -196,11 +199,14 @@ var TabContentView = Marionette.LayoutView.extend({
         chartRegion: '.analyze-chart-region'
     },
     onShow: function() {
-        var dataCollection = new coreModels.DataCollection(
-            this.model.get('categories')
-        );
+        var categories = this.model.get('categories'),
+            largest = _.max(_.map(categories, function(cat) {return cat.area;})),
```
rajadain: This might be more easily written as:

```javascript
largest = _.max(categories.pluck('area'));
```

See [pluck](http://backbonejs.org/#Collection-pluck)

jamesmcclain: I *knew* there had to be a better way to do that, thanks.

src/mmw/js/src/analyze/views.js
```diff
@@ -16,6 +17,8 @@ var $ = require('jquery'),
     tabContentTmpl = require('./templates/tabContent.html'),
     barChartTmpl = require('../core/templates/barChart.html');
 
+var M_IN_KM = 1000000;
```
lewfish: Maybe rename to M_IN_KM2

src/mmw/js/src/analyze/views.js
lewfish: Consider renaming to maxArea.

src/mmw/js/src/analyze/views.js
lewfish: Actually, M2_IN_KM2 makes more sense. However, there really isn't a good way to represent squaring in a variable name. Maybe just use a comment.

src/mmw/js/src/analyze/views.js
```diff
@@ -222,16 +228,25 @@ var TabContentsView = Marionette.CollectionView.extend({
 var TableRowView = Marionette.ItemView.extend({
     tagName: 'tr',
     template: tableRowTmpl,
-        templateHelpers: function() {
+    templateHelpers: function() {
+        var quotient = this.options.kiloMeters ? M_IN_KM : 1;
```
lewfish: I think by quotient you meant denominator?

jamesmcclain: Yes, you are correct.

src/mmw/js/src/analyze/views.js
jamesmcclain: Okay, will rename.

lewfish:
+1, tested.

src/mmw/js/src/analyze/views.js
lliss: M2_IN_KM2 seems reasonable to me.

src/mmw/js/src/analyze/views.js
```diff
@@ -222,16 +228,25 @@ var TabContentsView = Marionette.CollectionView.extend({
...
+
         return {
             // Convert coverage to percentage for display.
-            coveragePct: (this.model.get('coverage') * 100).toFixed(1)
+            coveragePct: (this.model.get('coverage') * 100).toFixed(1),
```
lliss: We handle display formatting for scaledArea in the template using nunjucks filters. It may make sense to do the same thing for coveragePct so that the logic is in the same place.

lliss:
Regarding @mmcfarland's comment about duplication, there are some core utility helpers for unit conversion. It might be worth centralizing this code GeoModel code under coreUtils:
https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/utils.js#L78-L125

src/mmw/js/src/analyze/views.js
jamesmcclain: This function seems to make sure that there is only one digit after the decimal point.  There does not seem to be a similar function among the default supplied by nunjucks, is there an economical way to create a custom filter to do this?

src/mmw/js/src/analyze/views.js
rajadain: There's also an M_IN_KM here: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/models.js#L187

jamesmcclain:
@lliss : Okay, I have moved the logic to the utility file.

lliss:
Neato. +1. I'll add a card to replace the code in https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/models.js#L186 with this new stuff.

jamesmcclain:
Okay, sounds good.

Issue #712, opened 2015-08-19
Consider using NLCD colors for land use modifications


caseypt:
Since the land use modifications come from the NLCD, we should consider using the official color scheme: http://www.mrlc.gov/nlcd11_leg.php.

Current colors:

<img width="322" alt="screen shot 2015-08-19 at 4 07 00 pm" src="https://cloud.githubusercontent.com/assets/1042475/9367589/4f2bcc7c-468c-11e5-9255-cf948bac47df.png">

Issue #601, opened 2015-07-21
Add NLCD overlay


mmcfarland:
@echeipesh is loading tiled display rasters for the NLCD data set into S3.  Add this layer as a map overlay option when we have a URL scheme for it.

mmcfarland:
https://s3.amazonaws.com/com.azavea.datahub.tms/nlcd/{z}/{x}/{y}.png although we may end up putting this URL behind a CloudFront CDN or similar.

caseypt:
Is it possible to follow whatever process generated these tiles, and do the same for the watershed, congressional district, and stream layers? Just thinking of consolidating where we serve our layers from, and that this (without knowing how they were generated) seems simpler then our current set-up.

mmcfarland:
These were generated from the raw original raster set by the GT team and use the default color classes and have no interactivity.  The other layers are sourced from vector files and also generate utfgrids of attributes we want to display and are custom styled.  I'm more hesitant to maintain a permanent cache of these things for the whole country at every zoom level since they are tweakable.  I think the changes coming with #447 provide a good mix of performance and flexibility.

PR #737, opened 2015-08-31
Use NLCD colors for land uses.

jamesmcclain:
The fill color for each has been changed to that which corresponds with its NLCD code.  The stroke colors have been changed to those on the opposite side of the color wheel.

Connects #712

screenshot from 2015-08-31 13 01 05

screenshot from 2015-08-31 13 00 36


rajadain:
+1, tested on various basemaps, the polygons are easily seen on all.

I see that since they use the same NLCD codes, Turf Grass and Tall Prairie, and Pasture and Short Prairie, share colors:

image

Turf Grass and Tall Prairie still remain easily distinguishable because of their very different patterning. Perhaps we may need some design assistance to develop similar distinguishable patterning for Pasture and Short Prairie.

caseypt:
Consider updating the outline colors, which can be configured here: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/modificationConfig.json

rajadain:
I wonder how it will look once #601 is available. If the modifications are the same colors as the basemap, it might be confusing.

mmcfarland:
@rajadain We're actually going for an effect like that, since applying these tools is essentially a reclassification of use.  If you extend an existing area of agriculture land with more agriculture, they should look reasonable contiguous or at least imply that they are the same value.  We may need to dial down the opacity so the existing use type doesn't bleed through too much.

lliss:
Appearance-wise, I think this looks good. You have my +1

jamesmcclain:
> Consider updating the outline colors, which can be configured here: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/modificationConfig.json

Okay, will-do.

mmcfarland:
:+1:  Looks good.  Can you make an issue to develop a new distinguishing pattern for Pasture/Prairie?  We can tweak the small stuff like opacity once we have the overlays loading.

kdeloach:
+1 Tested

Issue #700, opened 2015-08-13
Enable screenshot ability in ITSI portal


mmcfarland:
The ITSI portal snapshot feature is unable to snapshot within an iframe.  Work with Scott from Concord to work out a solution that enables this feature.

mmcfarland:
Concord runs a PhantomJS DOM rendering service and has released a JS utility for interacting with it.  We can use these tools to develop snapshotting ability while within the ITSI portal.  Based on the docs, the service should render our map markup:

>  Currently shutterbug supports HTML, SVG and Canvas elements.

There are instructions on how to interact with the service 
https://github.com/concord-consortium/shutterbug
and 
https://github.com/concord-consortium/shutterbug.js


From Scott:
> You don't need to use shutterbug.js, but in order to integrate into ITSI (without work on the ITSI side) you'll need to use the same postMessage API that shutterbug.js uses.

> The html snapshot is sent off to our server side app whose repo is here:
https://github.com/concord-consortium/shutterbug
That turns the html snapshot into an image. We host that, so you shouldn't need to worry about it, but you'll want to test the html snapshot being generated to see if the image returned is what you expect. You can test this result using the same page you used before: http://concord-consortium.github.io/lara-interactive-api/

> It is also possible that instead of a html snapshot you can generate the image yourselves. We don't have an way to handle that with our current API, so you'd need to return html with an image tag and data-uri.  If you do want to go this route, I'll see if can update shutterbug to handle this case better.

PR #738, opened 2015-08-31
Allow screenshot capture from ITSI parent frame

rajadain:
## Overview

Concord uses a library called Shutterbug to convert DOM elements to an image, with built in tools to communicate over their postMessage API. The original library is built using [Brunch](http://brunch.io/) and relies on dynamic require registration, which allows one to define modules at run time. Since that is [incompatible with Browserify](https://github.com/substack/node-browserify/issues/377) which needs modules to be defined at compile time, we adapt their work into a shim which fits in our stack.

Original library is: https://github.com/concord-consortium/shutterbug.js and this shim has been adapted from the latest release [v0.5.4](https://github.com/concord-consortium/shutterbug.js/releases/tag/v0.5.4).

## Testing Instructions

Check out the branch, bundle the scripts, and run `ngrok`:

    $ ngrok http 8000

We need to serve over a publicly accessible endpoint rather than `localhost` because Shutterbug.js compiles all asset URLs and sends them to a server on Concord's side. If we send `localhost` URLs they don't resolve and the screenshot looks odd.

Open their iframe test environment http://concord-consortium.github.io/lara-interactive-api/ and replace the iframe URL with your `ngrok` one, for example: https://23c23e62.ngrok.io/?itsi_embed=true

Then, click the Screenshot button, and verify that the correct screenshot appears. It should look like this:

image

## Known Issues

 * Currently the maps in the Compare view don't render:  
  image
 * Sliders reflect the correct numeric values in labels, but are always fixed to the mid-point in the screenshot.
 * Taking a screenshot causes the following error message in the iframe-phone library:  

        cant handle type: htmlFragResponse

    Unsure how it can be fixed.

Connects #700

### DIFF
--- a/src/mmw/js/shim/shutterbug.js
+++ b/src/mmw/js/shim/shutterbug.js
@@ -1,0 +1,460 @@
+"use strict";
+
+/**
+ * This shim has been developed for Shutterbug.js v0.5.4:
+ * https://github.com/concord-consortium/shutterbug.js/releases/tag/v0.5.4,
+ * which allows the ITSI parent frame to capture a screenshot of the running
+ * application.
+ *
+ * We make this shim since the original code was developed using Brunch which
+ * supports dynamically defined require modules, which are not supported by
+ * Browserify. Thus, we rewrite the library to not need dynamic require.
+ */
+
+var $ = require('jquery'),
+    /**
+     * Taken from shutterbug.js/app/scripts/default-server.js
+     */
+    DEFAULT_SERVER = "https://snapshot.concord.org/shutterbug",
+
+    /**
+     * Taken from shutterbug.js/app/scripts/html-tools.js
+     */
+    htmlTools = {
+        cloneDomItem: function($elem, elemTag) {
+            var $returnElm = $(elemTag);
+            $returnElm.addClass($elem.attr('class'));
+            $returnElm.attr('style', $elem.attr('style'));
+            $returnElm.css('background', $elem.css('background'));
+            $returnElm.attr('width', $elem.width());
+            $returnElm.attr('height', $elem.height());
+            return $returnElm;
+        },
+
+        generateFullHtmlFromFragment: function(fragment) {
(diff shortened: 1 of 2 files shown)


mmcfarland:
This seems to be working great with the few issues you reported.  Another I found is that after selecting any Google base layer (Hybrid or Terrain), the CSS fails to render correctly henceforth in any screenshot until the app is reloaded.

screenshot from 2015-09-01 11 00 04

mmcfarland:
There was some concern about allowing screenshots of a map with a Google basemap on it anyway, as it (apparently) violates their terms of service for storing map data without going through their API.  We had previously discussed with Concord the possibility of forcing an alternate basemap when the screenshotting utility requests the page.  Does shutterbug allow overriding of the URL at all so we might pass a flag for the subsequent request?

rajadain:
The Shutterbug server URL (which generates the image given DOM elements) is defined in the library here: https://github.com/rajadain/model-my-watershed/blob/feature/itsi-screenshot/src/mmw/js/shim/shutterbug.js#L18, and we can easily make it dynamic. Concord would then implement something on their end to handle the flags?

mmcfarland:
I was assuming that if we could add the query string argument to the URL that shutterbug captures, which presumably is requested by the PhantomJS service, that way we could switch out the basemap.  This is presuming that the client library is not actually chunking out the whole `body` of our doc and passing that to the service.  If that's the case, we may need to override the postMessage event, do some housecleaning, and then execute the shutterbug callback.  

Ultimately, I think we should create a comprehensive issue of the known rendering problems and the possible political ones (using GMaps at all, even if they did render) and consult Concord.  I think it'll be easier to work through them if this base work is merged into the public site.

rajadain:
Unfortunately it is the latter, where the Shutterbug.js extracts the entire DOM and POSTs that to their server, which parses it into an image, using public URLs only to download assets:

image

I will make a card cataloguing the issues discovered herein.

mmcfarland:
+1

Issue #732, opened 2015-08-27
Delay in chart refresh after spinner ends


rajadain:
In the Model view, for large, complex AoIs, the "updating" spinner in the Model Results pane goes away first, and then the charts are updated after a delay. This is seen on `develop` and staging:

![mmw-slider-chart-delay](https://cloud.githubusercontent.com/assets/1430060/9532653/ed4891a2-4cda-11e5-8269-febdd46020d0.gif)

Also see related #672

Issue #672, opened 2015-08-10
Cannot change precipitation slider while job is running


lliss:
If you attempt to change the precipitation slider while the job is running, it will be reset while you are dragging it because the view re-renders.

rajadain:
This also happens if you try to select a new modification to draw: the dropdown will close before you have a chance to select a modification because the view re-renders.

Ideally, only the model output view would re-render when the results are returned, but I believe because we update the Scenario model with the returned results, it causes a full page re-render. We might investigate if addition of results to scenario can be done silently.

mmcfarland:
This might become more of an obvious issue of the jobs take a while to complete, we should maybe up our sleep time to a more realistic (seeming) time of ~ 5 seconds.

PR #739, opened 2015-08-31
Save scenarios silently

rajadain:
## Overview

We trigger scenario saves when there are changes to the scenario name or modifications or inputs, so as to facilitate auto-saving and relieve the user of the responsibility to save manually. Unfortunately, this call to `save` would trigger a call to `parse` which would update the collection attributes with new values received from server, thus triggering many views to refresh.

We only want to update views when loading new values, thus when `parse` is called after a `fetch`. To ensure that `parse` does not affect values after a `save`, we pass in a `silent` option which is inspected within `parse`, and if found we simply return existing values without changing any, which does not trigger a view refresh.

Individual views track specific fields on the scenario, thus they update correctly, and in isolation.

## Testing Instructions

Try changing the precipitation slider quickly. Try changing modifications quickly. Changing inputs and modifications should trigger saves (observable in Network tab of dev tools), but not update the entire UI disruptively.

Observe the spinner on the results tab. The spinner on the results tab should appear and disappear in sync with changes to the charts.

Connects #732
Connects #672

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -497,9 +497,11 @@ var ScenarioModel = Backbone.Model.extend({
             // Makeshift locking mechanism to prevent double saves.
             this.saveCalled = true;
         }
-        this.save().fail(function() {
-            console.log('Failed to save scenario');
-        });
+        // Save silently so server values don't trigger reload
+        this.save(null, { silent: true })
+            .fail(function() {
+                console.log('Failed to save scenario');
+            });
     },
 
     addModification: function(modification) {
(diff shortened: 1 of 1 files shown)


lliss:
+1. Tested.

src/mmw/js/src/modeling/models.js
```diff
@@ -497,7 +497,8 @@ var ScenarioModel = Backbone.Model.extend({
             // Makeshift locking mechanism to prevent double saves.
             this.saveCalled = true;
         }
-        this.save().fail(function() {
+        // Save silently so server values don't trigger reload
+        this.save(null, { silent: true }).fail(function() {
```
mmcfarland: Formatting-wise, I prefer the second chain beginning its own line so it stands out a bit more.

```javascript
this.save(null, {...})
    .fail(function() {});
```

Issue #743, opened 2015-09-01
Boundary label appears above map controls


caseypt:
Some quick, obvious CSS rule changes (z-index, etc), did not fix this.

![screen shot 2015-09-01 at 9 50 25 am](https://cloud.githubusercontent.com/assets/1042475/9605999/e48d6df0-508e-11e5-8360-48d17d6f02a5.png)

caseypt:
@jfrankl Suggested we take a look at updating the CartoCSS to make the label part of the tile. This change will solve this issue, as well as allow for the user to see what areas are on the map without having to rollover each feature.

mmcfarland:
Not a bad idea - I was thinking of doing this specifically for the "boundary overlay" versions, but maybe just dropping the hover and going label is good for both.  Not sure if the labeling gets good enough at every level to entirely drop the hover (will a user see all labels for an area of several intersecting congressional district at a high zoom level?)

caseypt:
![image](https://cloud.githubusercontent.com/assets/1042475/19931483/143b78f8-a0e3-11e6-9af2-a5a18eeeff35.png)

This is still reproducible.

Issue #742, opened 2015-09-01
Bottom of map is covered up by analyze window header


caseypt:
![screen shot 2015-08-31 at 5 40 52 pm](https://cloud.githubusercontent.com/assets/1042475/9591144/74445662-5007-11e5-8c86-0a86d25e6cd3.png)

(The layer control is obscured by the header in this screenshot).

To reproduce:

- Select an AoI
- Move back to the draw page from the analyze page.
- The layer control will now be covered up by the analyze header.

This is because we aren't actually resizing the map when returning to the draw page.

Currently, the map only has a [full size and half size](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/views.js#L480-L494).

To fix the above issue, should we implement more sizes? Or maybe adjust our sass so that the map can't overlap with the analyze header (make the analyze header take up vertical space)?

Also, if you are in this this state, and you start drawing, the analyze header disappears. Then, if you don't finish drawing but press escape, the analyze header is brought back. Any solution should consider this workflow as well.

caseypt:
This also happens on the modeling page when the modeling results window is minimized:

![screen shot 2015-09-02 at 9 26 20 am](https://cloud.githubusercontent.com/assets/1042475/9632364/c3672082-5154-11e5-89f1-9d3f87e4cd18.png)

mmcfarland:
Similar to #497.

There is however some design work being done to investigate a different approach to the horizontal panel navigation (and resulting slim map area).  Let's hold off on working these issues until we see where that leads.

mmcfarland:
@jfrankl has provided some css based fixes that can accommodate proper map sizing in #831 that need to be implemented.

Issue #741, opened 2015-09-01
How should boundary layers and overlay layers be differentiated on the map?


caseypt:
Currently, layers that are used as both boundary layers (layers used for AoI selection) and overlay layers (layers that aren't interactive and are just for reference) use the same style configuration when used in both contexts.

If you are on the initial page, turning on an overlay and a boundary layer makes it hard to see what's what.

![screen shot 2015-09-01 at 9 31 36 am](https://cloud.githubusercontent.com/assets/1042475/9605537/45fc9884-508c-11e5-8dd7-9fd7b8fb72d3.png)

Also, just turning on multiple overlay layers makes it hard to tell which overlay layer is which, because all overlay layers use the same style.

![screen shot 2015-08-31 at 5 43 55 pm](https://cloud.githubusercontent.com/assets/1042475/9591214/e76521e4-5007-11e5-8eec-043a63f6a4c8.png)

How should we deal with this?

rajadain:
We can either use a different color, or a dashed line, or both. We may also add a subtle highlight on hover behavior to boundary layers since they are clickable.

caseypt:
Re: the highlight, the way I've done this in the past was passing along the feature shape in GeoJSON as an attribute in the UTFGrid data. It works well, but can substantially increase the payload of the UTFGrid files.

caseypt:
Another thing to consider is that you can turn on the same layer from the boundary layer selector and the layer control, so that there are duplicate layers.

rajadain:
Regarding duplicate layers with overlapping lines, in addition to different color and different stroke style (dashed or dotted), we can also differ stroke widths, so that the underlying layer has fatter strokes and the overlaying one has a thinner one.

I see what you mean about the increased payload.

mmcfarland:
Since we're now (or rather, soon) caching the tiles heavily, I think we can use the flexibility of having a dynamic tiler to start distinguishing between these surfaces.  As for the grid data, if we can work out a solution without having to include the shape data, I'm be greatly in favor of that.

mmcfarland:
We also discussed using CartoCSS styles to apply static labels on overlay features.  We could potentially  include a global or per overlay-layer option for including grids and therefore hover tips.  I'm hoping to not have too much interactivity because these are meant to largely be contextual pieces of information.

caseypt:
The current boundary and overlay color doesn't work well when the NLCD overlay is on. This screenshot has the congressional district boundaries turned on:

[...]

PR #744, opened 2015-09-01
Add custom layer control + overlay layer support

caseypt:
Adds a custom layer control to the map, and adds support for overlay layers.

This took a bit longer than I expected because I needed to do some refactoring of how we implemented layer support so that this could be implemented easily. 

This doesn't implement the layer control exactly as it's represented in the visual designs, but I thought that getting that right could be done by UI/UX.

As designed:

screen shot 2015-08-31 at 5 24 19 pm

As implemented:

screen shot 2015-08-31 at 5 24 41 pm

**Testing instructions**:

- Using the layer control, switch the active base map.
- Activate the "Overlays" tab of the layer control, and turn on and off the overlays for which you've loaded the data.
- Select an AoI.
- Any active overlays should still be on, and the layer control should still be present.

Known issues:

- The control is hidden by the analyze bar that appears when moving back from the analyze page to the draw page after having previously selected an AoI. I started to fix this in this PR, but it's not as simple as it first appears. We may need design support. https://github.com/WikiWatershed/model-my-watershed/issues/742

- The overlay layers that are also used as boundary layers have the same style.  https://github.com/WikiWatershed/model-my-watershed/issues/741

- The boundary label is drawn above map controls. https://github.com/WikiWatershed/model-my-watershed/issues/743

Connects to #629

### DIFF
--- a/src/mmw/apps/modeling/views.py
+++ b/src/mmw/apps/modeling/views.py
@@ -18,8 +18,6 @@ from django.contrib.gis.geos import GEOSGeometry
 
 from celery import chain
 
-from urlparse import urljoin
-
 from apps.core.models import Job
 from apps.core.tasks import save_job_error, save_job_result
 from apps.modeling import tasks
@@ -238,38 +236,16 @@ def _construct_tr55_job_chain(model_input, job_id):
 
 @decorators.api_view(['GET'])
 @decorators.permission_classes((AllowAny, ))
-def boundary_layers(request, table_id=None, obj_id=None):
-    layer_list = settings.BOUNDARY_LAYERS
-    layer_ids = [layer['code'] for layer in layer_list]
+def boundary_layer_detail(request, table_code, obj_id):
+    layers = [layer for layer in settings.LAYERS
+              if layer.get('code') == table_code]
+    table_name = layers[0]['table_name']
+    json_field = layers[0].get('json_field', 'geom')
 
-    if not table_id and not obj_id:
-        tiler_prefix = '//'
-        tiler_host = settings.TILER_HOST
-        tiler_postfix = '/{z}/{x}/{y}'
-        tiler_base = '%s%s' % (tiler_prefix, tiler_host)
+    query = 'SELECT {field} FROM {table} WHERE id = %s'.format(
+            field=json_field, table=table_name)
 
-        def augment(dictionary):
-            code = dictionary['code']
-            return {
-                'display': dictionary['display'],
-                'tableId': code,
-                'endpoint': urljoin(tiler_base, code + tiler_postfix),
-                'short_display': layer['short_display'],
-                'helptext': dictionary['helptext']
-            }
-
-        layers = [augment(layer) for layer in layer_list]
-        return Response(layers)
-
-    elif table_id in layer_ids and obj_id:
-        layers = filter(lambda l: l['code'] == table_id, layer_list)
-        table_name = layers[0]['table_name']
-        json_field = layers[0].get('json_field', 'geom')
-
-        query = 'SELECT {field} FROM {table} WHERE id = %s'.format(
-                field=json_field, table=table_name)
-
-        cursor = connection.cursor()
+    with connection.cursor() as cursor:
         cursor.execute(query, [int(obj_id)])
         row = cursor.fetchone()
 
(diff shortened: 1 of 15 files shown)


caseypt:
Connects to #629

caseypt:
Updated with @mmcfarland suggestions.

src/mmw/js/src/core/views.js
```diff
@@ -169,59 +171,43 @@ var MapView = Marionette.ItemView.extend({
...
+                position: 'bottomright',
+                collapsed: false
+            }).addTo(map);
         }
 
         if (options.addZoomControl) {
             map.addControl(new L.Control.Zoom({position: 'topright'}));
         }
 
+        var maxGeolocationAge = 60000;
```
lliss: Good call moving this here.

src/mmw/js/src/core/templates/layerControlContainer.html
```diff
@@ -0,0 +1,23 @@
+<div>
+    <div class="layer-control-button-container leaflet-control leaflet-bar">
+        <a class="leaflet-bar-part leaflet-bar-part-single" href="#" title="Show layer control">
+            <span class="fa fa-globe"></span>
```
lliss: Why did you choose globe here rather than the more traditional layer icon?

lliss:
Looks good. I tested it out and everything looks sharp. +1

src/mmw/apps/modeling/views.py
```diff
@@ -238,45 +236,21 @@ def _construct_tr55_job_chain(model_input, job_id):
...
+def boundary_layer_detail(request, table_code, obj_id):
+    layers = [layer for layer in settings.LAYERS
+              if layer.get('code') == table_code]
+    table_name = layers[0]['table_name']
+    json_field = layers[0].get('json_field', 'geom')
+
+    query = 'SELECT {field} FROM {table} WHERE id = %s'.format(
+            field=json_field, table=table_name)
+
+    cursor = connection.cursor()
```
rajadain: Should we be using a `with` here to ensure that the cursor is closed afterwords?

src/mmw/js/src/core/settings.js
```diff
@@ -3,7 +3,8 @@
 var defaultSettings = {
     itsi_embed: false,
     base_layers: {},
```
rajadain: Should `base_layers` be default to an array instead of an object now?

rajadain: Oh I see we're using it as an associative array, so an empty object is the correct initialization. Never mind.

src/mmw/mmw/settings/base.py
```diff
@@ -12,6 +12,8 @@
 from os.path import abspath, basename, dirname, join, normpath
 from sys import path
 
+from layer_settings import LAYERS  # NOQA
```
rajadain: Why do we need the `# NOQA` here?

src/mmw/js/src/core/templates/layerControlContainer.html
caseypt: I couldn't find it in the icon set we are using (Font Awesome). I'll file an issue.

caseypt: https://github.com/WikiWatershed/model-my-watershed/issues/749

src/mmw/mmw/settings/base.py
caseypt: Thanks. Will remove. Leftover from a previous version (`from layer_settings import *`)

src/mmw/js/src/core/templates/layerControlContainer.html
lliss: Annoying. I generally think we should be using fontello (http://fontello.com/) to pick icons we need so we don't have to load 500 others that never get used.

caseypt: Agreed.

src/mmw/mmw/settings/base.py
caseypt: Never mind. Removing it reminded me why it was there. We want to be able to access `LAYERS` from where ever settings is imported (including the client-side), but we don't actually need to use it directly in the settings file. This triggers the following PEP8 warning:

`'LAYERS' imported but unused F401`

src/mmw/apps/modeling/views.py
caseypt: Probably a good idea. Updated.

src/mmw/mmw/settings/base.py
rajadain: Ah, makes sense. Thanks for the investigation.

PR #745, opened 2015-09-01
Rename options under "Draw Area" dropdown

kdeloach:
Having a "Draw Area" option in the "Draw Area" dropdown looked funny.

image


jamesmcclain:
Tested, +1

jamesmcclain:
Arguably, the label "draw area" is kind of strange as well, given that you do not always draw.  This is an improvement, though.

kdeloach:
Yeah, I was thinking about changing that to "Select Area" but didn't want to change too much.

jamesmcclain:
That makes sense.

caseypt:
+1.

PR #750, opened 2015-09-02
Add NLCD layer as an overlay layer.

caseypt:
* Add the National Land Cover Database layer as an overlay layer
option.

screen shot 2015-09-02 at 11 56 10 am

**Testing Instructions**

- Verify that the layer control overlay tab contains the NLCD as a option and that turning it on adds tiles at zoom levels <= 13.

**Known issues**

Updated #741 with NLCD visibility issues.

Connects to #601

### DIFF
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -315,11 +315,13 @@ var MapView = Marionette.ItemView.extend({
                     });
                 } else {
                     var tileUrl = (layer.url.match(/png/) === null ?
-                                    layer.url + '.png' : layer.url);
+                                    layer.url + '.png' : layer.url),
+                        zIndex = layer.overlay ? 1 : 0;
 
                     leafletLayer = new L.TileLayer(tileUrl, {
                         attribution: layer.attribution || '',
-                        maxZoom: layer.maxZoom
+                        maxZoom: layer.maxZoom,
+                        zIndex: zIndex
                     });
                 }
 
--- a/src/mmw/mmw/settings/layer_settings.py
+++ b/src/mmw/mmw/settings/layer_settings.py
@@ -82,6 +82,15 @@ LAYERS = [
         'overlay': True,
     },
     {
+        'display': 'National Land Cover Database',
+        'short_display': 'NLCD',
+        'helptext': 'National Land Cover Database defines'
+                    'land use across the U.S.',
+        'url': 'https://s3.amazonaws.com/com.azavea.datahub.tms/'
+               'nlcd/{z}/{x}/{y}.png',
+        'overlay': True,
+    },
+    {
         'code': 'stream-low',
         'display': 'Low-Res',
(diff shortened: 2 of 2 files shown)


caseypt:
If possible, we may want to make the null values in the raster transparent. Right now they are black.

screen shot 2015-09-02 at 1 16 23 pm

rajadain:
I realize this is outside the current scope, but do we intend on adding a contextual legend at some point? To indicate what the colors mean?

rajadain:
The black border doesn't stretch forever, but stops at a certain level:

image

+1 for the idea of making those areas transparent, or at least gray with some transparency. Since the tiles are delivered as `.png`s adding transparency should be relatively straightforward.

rajadain:
Beyond a certain zoom level the NLCD tiles just disappear:

mmw-nlcd-zoom-in

If we can specify the max-zoom level for each overlay, we can give the user some feedback (such as the checkbox and label being grayed out) for zoom levels where certain layers are not available, like this:

image

Or even a ["not visible"](http://fortawesome.github.io/Font-Awesome/icon/eye-slash/) icon with a tooltip explaining that the overlay is not available at this zoom level.

This might be better handled as its own issue, though. If you like this idea I'll make one for it.

rajadain:
+1 tested.

rajadain:
Spoke too soon. There's odd behavior when the NLCD overlay is active and we change the basemaps:

mmw-nlcd-odd-overlay

Google basemaps correctly render underneath the overlay, but Mapbox and ESRI basemaps render over the overlay.

caseypt:
@rajadain Thanks. I'll look into it.

caseypt:
@rajadain Added a commit to fix the issue you reported.

caseypt:
@rajadain I'll add an issue about the legend.

caseypt:
@rajadain re: graying out the layer names, I do like that approach. It's how ArcGIS handles it, and I think it works pretty well. I think we should create an issue and consult with @jfrankl.

rajadain:
Created #754 for graying out inapplicable overlays. Taking a look at the z-index fix now.

rajadain:
Confirmed that the overlays are now always on top of basemaps.

Regarding the ordering of the overalays themselves, it seems like they operate in a stack-fashion, arranged in the order in which they were added:

mmw-nlcd-overlay-ordering

This allows the user to arrange multiple overlays in a custom fashion by enabling them in the right order. On the other hand, if the user expects a fixed ordering according to the list (e.g. Congressional Districts will always be on top because they are listed first) then they might be surprised.

Ideally, the list ordering would reflect the map ordering, and the user would be able to reorder elements in the list to reorder them on the map. But that is a feature in and of itself.

I don't mind the current behavior, but we should be consciously aware of it.

Everything else works perfectly well. A big **+1**.

caseypt:
@rajadain Yeah, that's a good solution to the ordering problem.

Issue #751, opened 2015-09-02
Clicking a Modification does not show Popup


rajadain:
Clicking a Modification in the `/project` used to show a popup box with detailed information and a delete button. This is no longer visible.

As seen on staging.

PR #752, opened 2015-09-02
Fix scenario modification popup.

caseypt:
* The interactiveMode property of the MapView is no longer
a direct property of the view, but is available in the view
options.

**Testing instructions**
- Create a new project with a scenario.
- Add a modification to an editable scenario.
- Click the modification. A pop-up should appear.

Connects to #751


caseypt:
Thanks @jamesmcclain.

jamesmcclain:
:+1: Tested.

Issue #690, opened 2015-08-13
Saved project should persist in forward navigation


mmcfarland:
Currently, if you are on a /project page for a saved model, navigate back to Analyze via the UI back button, and then move forward again, you are started at a blank project.  The saved project should persist during this kind of navigation - perhaps even back to the Draw page, similarly to the "Single Activity Mode" created for ITSI integration.

caseypt:
I imagine this will require a bit of reworking how we are storing the current project.

When doing this, consider renaming `currProject` to `currentProject`.

PR #756, opened 2015-09-03
Remember current project while navigating.

jamesmcclain:
It is now possible to navigate away from the /project page using the UI arrow buttons then return (using the UI arrow buttons) back to the project on which you have been working.

Connects #690

**To Test**
   * Do some activities which take you to the `/project` page.
   * Navigate away from the project page using the UI arrow buttons (back to the `/analyze` page or back to the root page), then back to the `/project` page.
   * You should find yourself back on the project that you were on when you left.

**Caveats**
   * I originally wanted to be able to take you back to the scenario that you had been working on, but unfortunately,this seems difficult.  One cannot set the scenario unless the DOM is in particular state: I am not able to give a precise description of this state and moreover I do not know how to wait for such a state without polling, which is a no-no (the usual expedient of releasing a lock within the `onShow`, `onRender` and `onAttach` functions of the views which create the DOM elements which I believe are required did not work).

### DIFF
--- a/src/mmw/js/src/analyze/controllers.js
+++ b/src/mmw/js/src/analyze/controllers.js
@@ -23,7 +23,7 @@ var AnalyzeController = {
         if (settings.get('activityMode')) {
             // Only one project allowed in Activity Mode. Save current project
             // and if in embedded mode, update interactive state for container.
-            var project = App.currProject,
+            var project = App.currentProject,
                 map = App.map;
 
             if (project && project.get('scenarios').isEmpty()) {
@@ -43,7 +43,7 @@ var AnalyzeController = {
             // Multiple projects allowed in Regular Mode. Nullify current
             // project since a new one will be created and saved by the
             // Modelling Controller.
-            App.currProject = null;
+            App.currentProject = null;
         }
     },
 
--- a/src/mmw/js/src/app.js
+++ b/src/mmw/js/src/app.js
@@ -30,7 +30,7 @@ var App = new Marionette.Application({
         });
 
         this._mapView.on('change:needs_reset', function(needs) {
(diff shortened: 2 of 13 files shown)


kdeloach:
One problem I found is that if you save a project, then analyze a new area, the previous project is loaded  instead of a blank project.

jamesmcclain:
> One problem I found is that if you save a project, then analyze a new area, the previous project is loaded instead of a blank project.

Thanks for detecting this.  I have addressed this and the other things you mentioned.  Please have another look when you get a chance.

lliss:
It seems that this feature negates the need for single project mode.
That was designed to ensure that a student who returns to the site outside of the ITSI portal would be able to continue work on the same project. It feels a lot like this handles the same cases. If this is true the singleProjectMode logic could be removed and some of the logic for loading the project page might become simpler.

kdeloach:
+1 tested

caseypt:
@lliss Should we create an issue to consolidate/remove "Single Project Mode" with this code?

lliss:
Yes. This PR covers most but not all of the features of "Single Project Mode" notably the idea that you can move around without destroying the current project. There are some differences that we should resolve. For instance "Single Project Mode" will create a project for the user as soon as they visit the home page. This is then passed back to ITSI portal as the "submission" for the assignment.

The biggest difference is that new projects are not created in Single Project Mode. So when you redraw an area of interest, the current project data is wiped out and the new AOI data is added to it. Therefore you are always on the same project. The idea being that a student can rework the submission a few times but the submission id that the teacher will be aware of remains the same.

So in short, there is overlap and we can consolidate features but this doesn't totally superseded the need for single project mode.

Issue #758, opened 2015-09-04
Trouble Downloading School District Boundaries


jamesmcclain:
When I attempt to load the boundaries using the `setupdb.sh` script, they fail to load.  The following can also be observed:

```
vagrant@app:~$ curl -s https://s3.amazonaws.com/data.mmw.azavea.com/boundary_school_districts.sql.gz
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>...</RequestId><HostId>...</HostId></Error>
vagrant@app:~$
```

mmcfarland:
There's a typo in the script, `districts` should not be plural.  PR forthcoming, as a temporary fix, you can drop the superfluous `s`.

PR #759, opened 2015-09-04
Correct bad boundary file name

mmcfarland:
* Update name to point to actual data file in S3
* Remove problematic `,` delimeter

Connects #758


jamesmcclain:
+1 I can now see the school districts.

Issue #757, opened 2015-09-03
Stretch overlays when zooming beyond available levels


rajadain:
Currently, beyond a certain zoom level the NLCD tiles just disappear:

![mmw-nlcd-zoom-in](https://cloud.githubusercontent.com/assets/1430060/9639066/1c886680-5178-11e5-88a0-a08ddfedffb7.gif)

They should instead still be visible at all visible levels. When tiles for a particular zoom level are not available, we should display the closest available level tiles stretched out.

See #750 and #754 for further discussion.

caseypt:
A broader implementation could use any Leaflet settings added to layers in the [layer settings](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/layer_settings.py) when the layers are [being created](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/views.js#L312-L324).

A simple implementation:

Layer settings:
```python
    {
        'display': 'National Land Cover Database',
        'short_display': 'NLCD',
        'helptext': 'National Land Cover Database defines'
                    'land use across the U.S.',
        'url': 'https://s3.amazonaws.com/com.azavea.datahub.tms/'
               'nlcd/{z}/{x}/{y}.png',
        'overlay': True,
        # Leaflet specific settings
        'maxNativeZoom': 13,
        'opacity': 0.5
    },
```

Layer creation:

```
    buildLayers: function(layerConfig) {
        var self = this,
            layers = {};
     

            _.each(layerConfig, function(layer) {
                var leafletLayer;

             // Make sure there are defaults
             var zIndex = layer.overlay ? 1 : 0;
             var defaults =  {
                  opacity: 1.0,
                  maxNativeZoom: null,
                  zIndex: zIndex
             }

                // Check to see if the google api service has been loaded
                // before creating a google layer
                if (self._googleMaps && layer.type === 'google') {
                    leafletLayer = new L.Google(layer.googleType, settings);
                } else {
                       var tileUrl = (layer.url.match(/png/) === null ?
                                    layer.url + '.png' : layer.url),
                       settings = _defaults(defaults, layer);

                    leafletLayer = new L.TileLayer(tileUrl, settings);
                }

                layers[layer['display']] = leafletLayer;
            });

        return layers;
    },
```

mmcfarland:
[...]

Issue #755, opened 2015-09-03
Alter overlay opacity


mmcfarland:
We should provide the ability to alter overlay layers opacity, and probably not start them out as 100% opaque, at least for ones with a fill (ie, not a boundary layer).

PR #761, opened 2015-09-08
Overlay Alpha

jamesmcclain:
This patch adds the capability to give overlay layers independent opacity sliders.  Only the *NLCD* layer has been so-equipped.

Connects #755 

The NLCD overlay layer has also been stretched so that it does not disappear when the user zooms in beyond level 13.

Connects #757 

**To Test**
   * Select the NLCD overlay
   * A slider should appear in the upper-right corner of the map when the layer is toggled on.  That slider will allow you to adjust the opacity of the layer.  The slider will disappear when the layer is toggled off (#755).
   * Zoom in past level 13.  Notice that instead of disappearing, the NLCD tiles are stretched (#757)

**Caveats**
   * Probably needs attention from designer.

### DIFF
--- a/src/mmw/js/src/core/opacityControl.js
+++ b/src/mmw/js/src/core/opacityControl.js
@@ -1,0 +1,83 @@
+/*
+Copyright (c) 2013, Jared Dominguez
+Copyright (c) 2013, LizardTech
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification,
+are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice, this
+  list of conditions and the following disclaimer in the documentation and/or
+  other materials provided with the distribution.
+
+* Neither the name of LizardTech nor the names of its
+  contributors may be used to endorse or promote products derived from
+  this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR
+ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+        Leaflet.OpacityControls, a plugin for adjusting the opacity of a Leaflet map.
+        (c) 2013, Jared Dominguez
+        (c) 2013, LizardTech
+
+        https://github.com/lizardtechblog/Leaflet.OpacityControls
+
+
+Modified by Azavea, 2015.
+*/
+
+'use strict';
+
+var L = require('leaflet'),
+    $ = require('jquery'),
+    Marionette = require('../../shim/backbone.marionette'),
+    sliderTmpl = require('./templates/opacitySlider.html');
+
+module.exports = L.Control.extend({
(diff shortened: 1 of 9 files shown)


src/mmw/mmw/settings/layer_settings.py
```diff
@@ -89,6 +89,9 @@
         'url': 'https://s3.amazonaws.com/com.azavea.datahub.tms/'
                'nlcd/{z}/{x}/{y}.png',
         'overlay': True,
+        'maxNativeZoom': 13,
+        'maxZoom': 18,
+        'opacity': 0.618,
```
caseypt: Consider documenting why this number is so specific or selecting a less precise number (0.6, 0.75, etc)

src/mmw/js/src/core/views.js
```diff
@@ -321,6 +321,8 @@ var MapView = Marionette.ItemView.extend({
                     leafletLayer = new L.TileLayer(tileUrl, {
                         attribution: layer.attribution || '',
                         maxZoom: layer.maxZoom,
+                        maxNativeZoom: layer.maxNativeZoom || layer.maxZoom,
+                        opacity: layer.opacity ? layer.opacity : 1,
                         zIndex: zIndex
```
caseypt: Consider making this more dynamic so that other Leaflet layer settings can be specified in the layer config and will work without further modifications (as described here: https://github.com/WikiWatershed/model-my-watershed/issues/757#issuecomment-137548508).

src/mmw/mmw/settings/layer_settings.py
jamesmcclain: It is 1 / $\phi$ and also 1 - $\phi$, the golden ratio.  It does not have any special significance beyond that.

caseypt:
@mmcfarland Just wanted to check that #755 was not describing a client-side feature ("We should provide the ability to alter overlay layers opacity").

mmcfarland:
Yeah, I did have in mind exposing a slider per layer for manual adjustment of opacity.  It implies a bit of UI design - but I think we can start out a little crude at first.

jamesmcclain:
I can work on the slider idea.

src/mmw/js/src/core/views.js
jamesmcclain: Okay, I did not see that comment before.  Will add that.

lewfish:
+1, tested. The slider looks good aside from needing to be centered horizontally relative to the two controls above it.

src/mmw/js/src/core/opacityControl.js
```diff
@@ -0,0 +1,80 @@
+/*
```
caseypt: This should probably go in https://github.com/WikiWatershed/model-my-watershed/tree/develop/src/mmw/js/shim

caseypt:
It looks like the slider and the layer opacity are being initialized at different values. When you turn on the NLCD layer, and just barely adjust the slider, you'll see that the opacity level makes a big jump.

src/mmw/js/src/core/opacityControl.js
```diff
@@ -0,0 +1,80 @@
...
+
+    setOpacityLayer: function (layer) {
+            this.opacity_layer = layer;
+    },
+
+    onAdd: function (map) {
+        var opacity_slider_div = L.DomUtil.create('div', 'opacity_slider_control'),
+            self = this;
+
+        $(opacity_slider_div).slider({
```
mmcfarland: Since we're shimming this small library anyway, what do you think the level of effort is to swap out the `jquery-ui` slider and replace it with a native slider?  Would be nice to in incur that dependency and also keep the sliders stylistically and behaviorally consistent.

jamesmcclain: I don't see any reason in principle why it would be difficult.  I can give this a try.

src/mmw/js/src/core/opacityControl.js
```diff
@@ -0,0 +1,83 @@
...
+*/
+
+'use strict';
+
+var L = require('leaflet'),
+    $ = require('jquery'),
+    Marionette = require('../../shim/backbone.marionette'),
+    sliderTmpl = require('./templates/opacitySlider.html');
+
+require('jquery-ui');
```
rajadain: `jquery-ui` must be added to `package.json`, `npm-shrinkwarp.json`, or `bundle.sh` before it can be used here.

jamesmcclain: I thought I got rid of all of the references to `jquery-ui`, thanks for catching.

jamesmcclain:
> +1, tested. The slider looks good aside from needing to be centered horizontally relative to the two controls above it.

Many thanks for the review.  I think a designer (or somebody with more design experience) will need to take a look at this to get it looking better.

> It looks like the slider and the layer opacity are being initialized at different values. When you turn on the NLCD layer, and just barely adjust the slider, you'll see that the opacity level makes a big jump.

Thanks, I have corrected this.

src/mmw/js/src/core/opacityControl.js
caseypt: Consider adding a comment so that we remember that we modified this from the original.

jamesmcclain: > This should probably go in https://github.com/WikiWatershed/model-my-watershed/tree/develop/src/mmw/js/shim

At one time I had it there, but I decided to move it to its current place because I noticed that it does something very analogous to `layerControl.js`.

caseypt: Re: the `shim` directory, external libraries that can't be brought in via NPM or need to be browserified go in there. But since we've changed most of the code, I think it's fine here.

Issue #763, opened 2015-09-08
Move stream layer display to new grouping


mmcfarland:
The stream slider should be moved from a "draw area" tool to a new map layers/navigation grouping as displayed in the newest visual designs.  

Relevant extracts from that document:

![screenshot from 2015-09-08 10 45 31](https://cloud.githubusercontent.com/assets/1014341/9737924/07e11e22-5617-11e5-9cd6-4e59e4eabfa0.png)
![screenshot from 2015-09-08 10 45 45](https://cloud.githubusercontent.com/assets/1014341/9737925/07e34b5c-5617-11e5-84d1-f7e03dd72e60.png)

PR #770, opened 2015-09-08
Feature/lol/move stream layer tool

lliss:
Connects #763

Provides a custom leaflet control to show and hide the stream selector.

To test:
 * Find the new map control and click it.
 * Move slider back and forth.
 * Ensure streams appear.

Notes:
 * Need a custom icon to match the mocks. Used some "streamy" looking arrows for the time being.
 * Controls jump upon opening. Could use some designer love to prevent this.

### DIFF
--- a/src/mmw/js/src/core/streamSliderControl.js
+++ b/src/mmw/js/src/core/streamSliderControl.js
@@ -1,0 +1,112 @@
+'use strict';
+
+var L = require('leaflet'),
+    Marionette = require('../../shim/backbone.marionette'),
+    streamSliderTmpl = require('./templates/streamSlider.html'),
+    settings = require('./settings');
+
+module.exports = L.Control.extend({
+    options: {
+        position: 'bottomright'
+    },
+
+    initialize: function (foo, options) {
+        L.Util.setOptions(this, options);
+        this.streamFeatures = L.featureGroup();
+    },
+
+    onAdd: function(map) {
+        var container = this.createContainer();
+        // Copied directly from the parent class.
+        if (!L.Browser.touch) {
+            L.DomEvent
+                .disableClickPropagation(container)
+                .disableScrollPropagation(container);
+        } else {
+            L.DomEvent.on(container, 'click', L.DomEvent.stopPropagation);
+        }
+        this.streamFeatures.addTo(map);
+        return container;
+    },
+
+    createContainer: function() {
+        var container = new StreamSliderView({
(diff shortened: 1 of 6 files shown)


lliss:
@azavea-bot rebuild

caseypt:
Should opening the layer control or the stream slider close the other if that one is open? It seems like that needs to happen or we should have the control window stack horizontally.

caseypt:
To have the control buttons stack (like the zoom in/out buttons), it looks like we have to wrap them all in a `div` with the class `leaflet-bar`. And then each control gets the class `leaflet-bar-part`. Consider doing taht here.

src/mmw/js/src/core/streamSliderControl.js
```diff
@@ -0,0 +1,111 @@
...
+        if (!L.Browser.touch) {
+            L.DomEvent
+                .disableClickPropagation(container)
+                .disableScrollPropagation(container);
+        } else {
+            L.DomEvent.on(container, 'click', L.DomEvent.stopPropagation);
+        }
+        this.streamFeatures.addTo(map);
+        return container;
+    },
```
caseypt: Missing space.

caseypt:
+1. Test. Nice job.

@mmcfarland Can we get some designer time next sprint to help with this and the layer control?

mmcfarland:
@caseypt Yep, however in the meantime we're going to just be hiding the streams (#778)

caseypt:
To help with styling efforts, consider moving the locate control to the bottom right of the map. https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/views.js#L104

lliss:
> To have the control buttons stack (like the zoom in/out buttons), it looks like we have to wrap them all in a div with the class leaflet-bar. And then each control gets the class leaflet-bar-part. Consider doing that here.

@casey Moving these around is actually pretty difficult considering how we're building the HTML for them. Were you able to make these look the right way on your end.

caseypt:
@lliss I didn't realize there was a special classes for keeping the buttons together when I was building the layer control. I thought they would just stack if another control was added.

lliss:
> @lliss I didn't realize there was a special classes for keeping the buttons together when I was building the layer control. I thought they would just stack if another control was added.

My preference is to punt on this and try to resolve all the style issues at once later on.

caseypt:
Sounds good to me.

Issue #767, opened 2015-09-08
Text changes


mmcfarland:
* Change name of “pre-Columbian” condition to “100% Forest”
* In the scenarios, the first bar graph should be labeled “Current Conditions”, not “Original”.
* The title should be “Model My Watershed: Geospatial Analysis”  in the upper black bar on the https://mmw.azavea.com/analyze page.
* The title should be “Model My Watershed: Model”  in the upper black bar on the modeling pages.

PR #775, opened 2015-09-08
Text Changes

caseypt:
Update modeling bar chart titles and show the page title in the header. See commit messages for details.

**Testing instructions:**

Bar chart titles
- Create a new project or open an existing one.
- Verify that the second chart for the current conditions scenario is titled "100% Forest"
- Verify that the first chart for any other scenario is titled "Current Conditions".

Page header
- On each page of the app, verify that the title is shown in the header.
- Visit the project page directly and verify that the page title is in the header.

While it is helpful to add a title to each page, I think the current location is problematic. It feels like the title should be on a line below the app title or otherwise differentiated. In it's current placement, it looks like it's the subtitle to the app, rather than the bread crumb or current location. cc @jfrankl.

screen shot 2015-09-08 at 5 05 51 pm

Connects to #767

### DIFF
--- a/src/mmw/js/src/analyze/controllers.js
+++ b/src/mmw/js/src/analyze/controllers.js
@@ -54,6 +54,8 @@ var AnalyzeController = {
                 model: createTaskModel(aoi)
             });
 
+        App.state.set('current_page_title', 'Geospatial Analysis');
+
         App.rootView.footerRegion.show(analyzeWindow);
     },
 
--- a/src/mmw/js/src/app.js
+++ b/src/mmw/js/src/app.js
@@ -13,6 +13,7 @@ var App = new Marionette.Application({
     initialize: function() {
         this.restApi = new RestAPI();
         this.map = new models.MapModel();
+        this.state = new models.AppStateModel();
 
         // If in embed mode we are by default in activity mode.
         var activityMode = settings.get('itsi_embed');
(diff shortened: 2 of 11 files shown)


jamesmcclain:
+1 Tested.  I was able to observe all of the changes.

rajadain:
It seems like the page title is more related to app state than user state. Was the user chosen for tracking this because it was a convenient app-level Backbone Model?

caseypt:
@rajadain Yes, and it's what is currently powering the header. I could create a header model or some sort of general app state model if you think that would be better.

rajadain:
Since `current_page_title` is not persisted to database (so the `user` is independent of this value), and retains its value when the user state changes (for example, when an anonymous user logs in at a late stage) (so the value is independent of the `user`), it feels like adding it to the user model is conflation. While creating a whole new model for this may seem like overengineering now, I do think it would be better, and it would be a place for us to store future app state variables.

caseypt:
@rajadain Added a commit to use an app state model. The header view is a little weird since it needs access to the app, but the app creates it, but let me know what you think.

rajadain:
This looks good! We're injecting the state into the Header View, so there's no circular dependencies, and we have other views that listen to multiple models, so its not like we're doing something very abhorrent. Another restructuring we may do is to make the App State the main model for the Header view, and make a sub User View for which the User Model is the main model. That way each view will have a single model to listen to. But I'm not sure if that is worth it.

I would +1 this in its current state.

caseypt:
Cool. Rebased and added a test.

caseypt:
@azavea-bot rebuild.

Issue #778, opened 2015-09-09
Hide Stream and Delineate Watershed tools


mmcfarland:
While these features are being rolled out nationally, they should not be visible to the user.

It would be good to hide them on production and show them on staging/dev.  We could add a simple settings feature flag and pass it to the client.

PR #780, opened 2015-09-09
Make draw tools configurable

rajadain:
## Overview

Add settings to enable or disable drawing tools in the UI.

## Testing Instructions

 * Checkout the branch and open the home page, ensure that you can see all draw tools.
 * Edit the `development.py` settings file and comment out arbitrary tools, reload the home page and see those tools disappear.
 * Run all tests to ensure that disappearing tools don't affect any other behavior.

## Notes

We should probably wait until #770 is in, and then update this branch to adapt to that new placement of the stream slider. But the general idea will be the same and thus can be tested.

Connects #778

### DIFF
--- a/src/mmw/js/src/core/tests.js
+++ b/src/mmw/js/src/core/tests.js
@@ -118,6 +118,18 @@ describe('Core', function() {
 
     describe('Views', function() {
         describe('MapView', function() {
+            before(function() {
+                // Ensure that map controls are enabled before testing
+
+                settings.set('map_controls', [
+                    'LayerAttribution',
+                    'LayerSelector',
+                    'LocateMeButton',
+                    'StreamControl',
+                    'ZoomControl',
+                ]);
+            });
+
             it('adds layers to the map when the map model attribute areaOfInterest is set', function() {
                 var model = new models.MapModel(),
                     view = new views.MapView({
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -16,7 +16,8 @@ var $ = require('jquery'),
     selectTypeTmpl = require('./templates/selectType.html'),
     drawTmpl = require('./templates/draw.html'),
     resetDrawTmpl = require('./templates/reset.html'),
-    placeMarkerTmpl = require('./templates/placeMarker.html');
+    placeMarkerTmpl = require('./templates/placeMarker.html'),
+    settings = require('../core/settings');
 
 // Responsible for loading and displaying tools for selecting and drawing
 // shapes on the map.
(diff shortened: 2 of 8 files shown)


src/mmw/js/src/draw/views.js
```diff
@@ -56,21 +56,32 @@ var ToolbarView = Marionette.LayoutView.extend({
...
-            model: this.model
-        }));
-        this.resetRegion.show(new ResetDrawView({
-            model: this.model
-        }));
-        this.streamRegion.show(new StreamSliderView({
-            model: this.model
-        }));
+        var draw_tools = settings.get('draw_tools');
+        if (_.indexOf(draw_tools, 'SelectArea') >= 0) {
```
rajadain: I tried to think of a more elegant way of doing this, but instantiating a dynamic class using reflection from a string seemed too indirect and obtuse to be helpful, even though it would reduce some repetition.

If there's a better way of doing this, I'd love to hear about it.

src/mmw/mmw/settings/development.py
```diff
@@ -72,5 +72,17 @@
 
 # END TOOLBAR CONFIGURATION
 
+# UI CONFIGURATION
+
+DRAW_TOOLS = [
+    'SelectArea',   # Boundary Selector
```
rajadain: The keys were chosen to be identical to the Backbone View classes they control, but since in conversation we refer to these controls differently, those conversational names have been added as comments.

src/mmw/js/src/draw/views.js
caseypt: Not too much shorter, but `_.contains` may work better here.

http://underscorejs.org/#contains

rajadain: Thanks, that is what I was looking for. I just kept seeing `_.has` which was only for objects, and then looked only at the Arrays list, completely forgot about Collections functions. Updated.

caseypt:
The build is failing because I believe you also need to add the `DRAW_TOOL` settings to `settings/test.py`. I think they should match the production settings.

jamesmcclain:
I am seeing different (I think) than those occurring on Jenkins.

rajadain:
Fixing tests and rebasing ...

rajadain:
@azavea-bot rebuild

rajadain:
Alright, updated tests to ensure that draw tools are properly configured before testing. One more look?

jamesmcclain:
+1 Tested: I did not test all 32 possibilities, but the handful that I did chose worked correctly (the correct ones were shown or not shown on screen and the tests passed).

One thing: the stream slider is present in the list in `development.py` but has already been removed from its former place.  The new stream slider button is present regardless of the whether it is present or not in the `DRAW_TOOLS` list.

rajadain:
Ah thanks for the reminder, yes I have to update the code to hide the new stream slider.

rajadain:
Alright, this update adds the ability to configure Map Controls as well, since Stream Selector is now a map control. We also consolidate settings in `base.py`, only overriding when needed e.g. in `production.py`.

jamesmcclain:
+1

rajadain:
Squashed commits, rebased on `develop`. Will merge in shortly.

caseypt:
Nice inclusion of the map UI controls.

caseypt:
Also, FYI if you haven't seen this related PR: https://github.com/WikiWatershed/model-my-watershed/pull/784

rajadain:
Thanks for pointing that out. It should not affect this work, since these are the defaults that get overridden by that.

src/mmw/js/src/core/tests.js
```diff
@@ -118,6 +118,18 @@ describe('Core', function() {
 
     describe('Views', function() {
         describe('MapView', function() {
+            before(function() {
+                // Ensure that draw tools are enabled before testing
```
rajadain: Ack these are `map_controls` not `draw_tools`. Fixing...

PR #781, opened 2015-09-09
Add scenario name to compare view.

lliss:
The wireframes show the scenario name as an overlay on the map. Here we add
the scenario name to each scenario model passed into the compare view and
then create an overlay over the map view that includes then text.

Connects #781 

To test:
 * Create a project.
 * Name the scenarios.
 * Move to compare view. 
 * Ensure names are visible:

screencap


jamesmcclain:
+1 Tested.

lliss:
@lewfish, may want to take a look before I merge it.

lewfish:
Looks good to me.

PR #784, opened 2015-09-10
Do not show stream control on compare view maps.

lliss:
Found this on dev while testing #781 
screencap

Should now look like this:

screencap


lewfish:
+1, tested

Issue #786, opened 2015-09-11
Analyze Service to include NLCD type


mmcfarland:
Add an `nlcd` key to each entry in the returned census so that the resulting bar chart can be colored to the official NLCD legend.  Presently, the hard-coded Analyze results define this key.

Issue #774, opened 2015-09-08
LC Analysis results should match NLCD color scheme


mmcfarland:
The bar chart land color value should be that of the NLCD class color value.  This will also make it serve double duty as a legend.

http://www.mrlc.gov/nlcd11_leg.php

![screenshot from 2015-09-08 16 56 12](https://cloud.githubusercontent.com/assets/1014341/9747258/aec92936-564a-11e5-882b-f1ee11ab72dd.png)

Issue #765, opened 2015-09-08
Align MMW land covers to NLCD


mmcfarland:
A review of what is currently being used by MMW and MMW-micro resulted in several changes required. 

We all believe that it is not helpful to offer land cover names that do not match the underlying data or provide additional modeling output options.
Some of the current mapping of names to NLCD Level II classes is incorrect and gives bad model output as a result (i.e. turf_grass/urban_grass in clay loam is producing less run-off than forest)
There are 3 places that the Land Cover names get shown
* https://mmw.azavea.com/analyze, under the Land tab. These should show all 16 NLCD Level II Classes (20 classes minus 52, 72, 73, 74, which are in Alaska only), exactly as they are listed here: http://www.mrlc.gov/nlcd11_leg.php.
* https://mmw.azavea.com/micro/ presently shows 12 Land Cover types, but these do not have 1:1 mappings with NLCD classes
* Land cover button in Model New Scenarios shows the same Land Cover classes as the Micro model.

These are the 12 land cover names we should use, in both New Scenario and Micro apps. In the help text, we should include the NLCD code and definition, but also list examples and alternative names to help users choose.  Sara and Melinda will work on the help text.  

1. Open Water (NLCD 11, including ponds and lakes). Specifically want to include this because someone might want to modify a scenario by adding a pond or waterway
1. Developed, Open Space (NLCD 21 - turf grass, urban grass, park)
1. Developed, Low Intensity (NLCD 22 - low density single family homes) title could be shorted to “Developed-Low”
1. Developed, Medium Intensity (NLCD 23 - high density single family homes)
title could be shorted to “Developed-Med.”
1. Developed, High Intensity (NLCD 24 - commercial and industrial)
title could be shorted to “Developed-High”
1. Barren Land (NLCD 31, including desert)
1. Forest (NLCD 41, 42, 43, including deciduous, evergreen and mixed forest)
1. Shrub/Scrub (NLCD 52, including chaparral, )
1. Grassland/Herbaceous (NLCD 71 including tall and shortgrass prairie, savanna)
title could be shorted to “Grassland”
1. Pasture/Hay (NLCD 81)
1. Cultivated Crops (NLCD 82, includes row crops, orchards, vineyards)
title could be shorted to “Crops”
1. Wetlands (NLCD 90, 95, including wooded wetland, herbaceous wetland, marsh, bog)

[...]

PR #787, opened 2015-09-11
Analysis results coded to NLCD colors

mmcfarland:
Allows NLCD colors to be applied to the bar chart for analysis results.  The Analyze endpoint will need to include NLCD type (may be a representive type, see #765).  Only the 12 NLCD classes to be included in MMW are provided, and the hover highlight effect was left in place, can be tweaked as needed. 

New Issue to add this feature to the real service: #786

##### Testing
* Restart your worker `celeryd` service
* Define an area of interest and notice several bars in the chart are the corresponding NLCD color
* Check the soil chart, notice the the default color is used 

Connects #774

### DIFF
--- a/src/mmw/js/src/core/chart.js
+++ b/src/mmw/js/src/core/chart.js
@@ -294,7 +294,10 @@ function makeBarChart(el, data, indVar, depVars, options) {
         barGroups = chartGroup.selectAll('.bar')
                 .data(data)
                 .enter().append('g')
-                .attr('class', 'g')
+                .attr('class', function(d) { 
+                    return d.className ? d.className : ''; 
+                })
+                .classed('g', true)
                 .attr('transform', function(d) {
                     return 'translate(0,' + y(d[indVar]) + ')';
                 });
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -269,7 +269,13 @@ var ChartView = Marionette.ItemView.extend({
     addChart: function() {
         var chartEl = this.$el.find('.bar-chart').get(0),
             chartData = this.collection.map(function(model) {
-                return model.attributes;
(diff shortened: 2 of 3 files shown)


mmcfarland:
Representative style:

screenshot from 2015-09-10 23 56 49

rajadain:
+1 works very well.

If you follow the discussion from https://github.com/WikiWatershed/model-my-watershed/pull/653#issuecomment-127647366 to #662 to #671, you will see that we tried to move away from NLCD codes and use string identifiers instead. I'm not sure how that interacts with the revelations in #765, but we should consider using a unified scheme throughout.

src/mmw/js/src/core/chart.js
```diff
@@ -294,7 +294,10 @@ function makeBarChart(el, data, indVar, depVars, options) {
         barGroups = chartGroup.selectAll('.bar')
                 .data(data)
                 .enter().append('g')
-                .attr('class', 'g')
+                .attr('class', function(d) { 
+                    return d.className ? d.className : ''; 
+                })
+                .classed('g', true)
```
rajadain: Clever combination of `attr` and `classed`.

mmcfarland:
I think the new understanding of adhering to NLCD more strictly whenever we display or calculate on land cover means we should stick to the numerical class ids which are well defined.  They don't really present strict strings, and in fact some of the strings we've used are being removed as a concept per #765.  I'm not sure what the best way to consolidate the LC tool vector styles with this style, but there will be a 1:1 relationship eventually between it and this graph.  It might be as simple as extracting the hex values out to sass variables and using them in different places.

rajadain:
Understood. I guess we'll converge on the codes one way or another as we progress. I like the idea of extracting NLCD colors into variables, but since we use them in Django templates, JavaScript variables, and SASS, it might not be as centralized as we wish.

Issue #776, opened 2015-09-08
Add grouping to Layer Control


mmcfarland:
We'll allow two groups of overlays, for convenience, "vector" and "raster", though they are all burned image tiles.  The boundary layers and the NLCD/Soil will each be in their own group and be mutually exclusive within the group.  You can select one "vector" and one "raster" layer at a time. 

Should help partially resolve #741

caseypt:
In other words, there should now be the following categories with the following layers?

- Basemaps
- Vector
    - County
    - Schools
    - HUCs
    - Congressional Districts
- Raster
    - NLCD
    - Soil

mmcfarland:
Correct-ish, I think that the vector/raster groups remain on the same "overlays" tab, though.  And we need better names.  Boundary and coverage?  Not sure.

caseypt:
Ohh. I see. Same tab, but groups within the tab.

So something like this:

- Basemap
- Overlays
    - "Boundary"
        - County
        - Schools
        - HUCs
        - Congressional Districts
    - "Coverage"
    - NLCD
    - SURRGO

mmcfarland:
Yep, updated the indentation under "Coverage" (probably a typo).  Looks right to me.  Each group under overlays can have one layer selected at a time

PR #791, opened 2015-09-15
Feature/lol/group layer selection

lliss:
Connects #776 

 * Change layer settings from change for overlay layers in subgroups (raster and vector)
 * Extend leaflet's Layer control to use these special groups to create radios rather than checkboxes.
 * Enable a "no selection" option to disable all overlay layers.

To test:
 * Open the layer control.
 * Switch to the overlays tab.
 * Choose a vector layer.
 * Choose a raster layer.
 * Ensure you can disable both layers.
 * Ensure you cannot enable more than one of each type of layer.

### DIFF
--- a/src/mmw/apps/home/views.py
+++ b/src/mmw/apps/home/views.py
@@ -115,7 +115,8 @@ def get_client_settings(request):
             'base_layers': get_layer_config('basemap'),
             'stream_layers': get_layer_config('stream'),
             'boundary_layers': get_layer_config('boundary'),
-            'overlay_layers': get_layer_config('overlay'),
+            'vector_layers': get_layer_config('vector'),
+            'raster_layers': get_layer_config('raster'),
             'draw_tools': settings.DRAW_TOOLS,
             'map_controls': settings.MAP_CONTROLS,
             'google_maps_api_key': settings.GOOGLE_MAPS_API_KEY
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -188,7 +188,8 @@ var MapView = Marionette.ItemView.extend({
         var map = new L.Map(this.el, {
                 zoomControl: false,
                 attributionControl: options.showLayerAttribution
-            });
+            }),
+            overlayLayers = settings.get('vector_layers').concat(settings.get('raster_layers'))
 
         // Center the map on the U.S.
         map.fitBounds([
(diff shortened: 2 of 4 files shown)


src/mmw/js/src/core/views.js
```diff
@@ -338,7 +351,7 @@ var MapView = Marionette.ItemView.extend({
                     leafletLayer = new L.Google(layer.googleType, {
                         maxZoom: layer.maxZoom
                     });
-                } else {
+                } else if (!layer.empty) {
```
rajadain: Apparently `layer` is undefined in some cases, causing all tests to fail.

rajadain:
+1. Functionally works very well, and looks pretty nice:

image

I wish that there was more room above the "Coverage" heading, but bootstrap by default only gives those bottom paddings, not top. It will probably be addressed when Design takes a look at that entire box.

Code is great too, except for that one failing use case which is preventing tests.

Great work!

src/mmw/apps/home/views.py
```diff
@@ -115,7 +115,8 @@ def get_client_settings(request):
             'base_layers': get_layer_config('basemap'),
             'stream_layers': get_layer_config('stream'),
             'boundary_layers': get_layer_config('boundary'),
-            'overlay_layers': get_layer_config('overlay'),
+            'vector_layers': get_layer_config('vector'),
+            'raster_layers': get_layer_config('raster'),
```
caseypt: Consider adding the word "overlay" in the attribute names just so it's clear that's what they are.

src/mmw/js/src/core/views.js
```diff
@@ -188,7 +188,20 @@ var MapView = Marionette.ItemView.extend({
...
+                empty: true
+            },
+            nullRasterLayer = {
+                display: 'nullRaster',
+                raster: true,
+                empty: true
+            },
+            vectorLayers = [nullVectorLayer].concat(settings.get('vector_layers')),
+            rasterLayers = [nullRasterLayer].concat(settings.get('raster_layers')),
+            overlayLayers = vectorLayers.concat(rasterLayers);
```
caseypt: Consider splitting this logic out into a separate function that's called in initialize and returns `overayLayers`, just for the sake of keeping the `initialize` function shorter (it was previously ~300 lines).

caseypt:
+1. Works well. Nice job.

src/mmw/js/src/core/views.js
lliss: Looking into that now.

PR #792, opened 2015-09-15
Remove Flower UI for Celery

hectcastro:
Flower keeps its knowledge of past Celery tasks in memory by default. There is an optional configuration setting for enabling persistence, but it is limited to a file on the local file system. In our setup, we planned to support multiple worker machines, each with their own copy of Flower, behind a load balancer. Viewing the Flower UI in this configuration shows inconsistent results because there is no centralized storage.


mmcfarland:
+1 Destroyed and re-provisioned worker, jobs still running.

Issue #771, opened 2015-09-08
Compare Mode: Lock/Sync precipitation slider values


mmcfarland:
The initial implementation of Compare Mode should lock all the precipitation values for all scenarios.  Adjusting any slider should adjust all scenarios and re-run the model output.  This shouldn't bee too expensive, because the geoprocessing results are cached.

rajadain:
What should happen if there are three scenarios, each with different precipitation values, in the Model view, and then the user enters the Compare view? Should the Compare view start with different values, but as soon as the user changes one it updates all of them to the same value? Should there be a toggle button that allows the user to update one value at a time?

mmcfarland:
I think we should default all the models precipitation to the same value we start new scenarios to.  I think it should be clear that we're starting from a new universal value.  We may also luck out if they haven't changed the sliders for some, because the result should be 100% cached in that case and no job is submitted.  In all cases, we should be verifying that we're not requesting a geop job.

FYI, in the future this feature will be expanded to let users unlock the sliders, and be able to compare based on "climate scenarios", that is, the same/similar modifications against different predictions of future climate.

PR #793, opened 2015-09-15
Synchronize precipitation sliders in Compare View

jamesmcclain:
This patch exports a new object from the file `src/modeling/controls.js` which contains functions which can be used to keep precipitation sliders in sync.

Connects #771

**To Test**
   * Create a project with multiple scenarios
   * Go to the compare view
   * Move the precipitation slider around -- all should stay in sync

**Caveats**
   * If you have too many scenarios, it might overwhelm the worker and cause you to not get results

### DIFF
--- a/src/mmw/js/src/compare/controllers.js
+++ b/src/mmw/js/src/compare/controllers.js
@@ -4,9 +4,15 @@ var _ = require('lodash'),
     App = require('../app'),
     router = require('../router').router,
     views = require('./views'),
-    modelingModels = require('../modeling/models.js');
+    modelingModels = require('../modeling/models.js'),
+    modelingControls = require('../modeling/controls'),
+    synchronizer = modelingControls.PrecipitationSynchronizer;
 
 var CompareController = {
+    comparePrepare: function() {
+        synchronizer.on();
+    },
+
     compare: function(projectId) {
         if (App.currentProject) {
             setupProjectCopy();
@@ -28,6 +34,7 @@ var CompareController = {
     },
 
     compareCleanUp: function() {
+        synchronizer.off();
         App.user.off('change:guest', saveAfterLogin);
         App.origProject.off('change:id', updateUrl);
 
--- a/src/mmw/js/src/modeling/controls.js
+++ b/src/mmw/js/src/modeling/controls.js
@@ -106,6 +106,69 @@ var ConservationPracticeView = ModificationsView.extend({
     }
 });
 
+var PrecipitationSynchronizer = (function() {
+    var isEnabled = false,
+        precipViews = [];
+
+    // Add a view to the list of views to be kept syncrhonized
+    function add(precipView) {
+        if (isEnabled) {
+            precipViews.push(precipView);
+        }
+    }
+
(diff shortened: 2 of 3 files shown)


src/mmw/js/src/modeling/controls.js
```diff
@@ -106,6 +106,72 @@ var ConservationPracticeView = ModificationsView.extend({
...
+    function add(precipView) {
+        if (isEnabled) {
+            precipViews.push(precipView);
+        }
+    }
+
+    // Remove a view from the list
+    function remove(precipView) {
+        if (isEnabled) {
+            precipViews = _.filter(precipViews, function(otherPrecipView) {
```
kdeloach: Consider using `_.without`?

kdeloach:
+1 Nice. Very easy to understand.

jamesmcclain:
> +1 Nice. Very easy to understand.

Thanks! I am gradually learning how to Javascript.

src/mmw/js/src/modeling/controls.js
```diff
@@ -106,6 +106,72 @@ var ConservationPracticeView = ModificationsView.extend({
...
+    }
+
+    // Synchronize the group to the first slider
+    function sync() {
+        if (precipViews.length > 0) {
+            syncTo(precipViews[0]);
+        }
+    }
+
+    return {
```
rajadain: I really like this way of returning functions defined elsewhere. Makes the API very easy to grok.

src/mmw/js/src/compare/controllers.js
```diff
@@ -4,9 +4,16 @@ var _ = require('lodash'),
     App = require('../app'),
     router = require('../router').router,
     views = require('./views'),
-    modelingModels = require('../modeling/models.js');
+    modelingModels = require('../modeling/models.js'),
+    modelingControls = require('../modeling/controls');
+
+var synchronizer = modelingControls.PrecipitationSynchronizer;
```
rajadain: Consider combining the above into:

```javascript
modelingModels = require('../modeling/models'),
synchronizer = require('../modeling/controls').PrecipitationSynchronizer;
```

rajadain:
+1 tested, works very well. Code is also very readable. Great job!

src/mmw/js/src/compare/controllers.js
jamesmcclain: I actually tried to do this, but it did not work.  I do not understand why.

rajadain: That's odd, because we do use this in other places, like this: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/controllers.js#L8

I wonder if it's because we're not using the `new` keyword, or because of the [IIFE](https://en.wikipedia.org/wiki/Immediately-invoked_function_expression).

But it's not a big deal. I don't mind if we leave it.

src/mmw/js/src/modeling/controls.js
jamesmcclain: Thanks, implemented.

PR #794, opened 2015-09-15
Move alert icon

jfrankl:
The alert icon would often cover the evapotranspiration portion of the stacked bar. I downsized it and moved it up above the bar.

image


lewfish:
What is the meaning of the alert icon? Maybe there should be a tooltip for it. (You don't need to make any changes to address this in this PR, but maybe make a new issue.) +1, tested.

Issue #747, opened 2015-09-01
Issues with ITSI Screenshot


rajadain:
We added the ability for the ITSI portal to request screenshots in #700 and #738. The following are known issues with the current implementation:

 1. Currently the maps in the Compare view don't render:  
  ![image](https://cloud.githubusercontent.com/assets/1430060/9587223/4ce57942-4fef-11e5-8fea-2d02ae73ac95.png)
 2. Sliders reflect the correct numeric values in labels, but are always fixed to the mid-point in the screenshot.
 3. Taking a screenshot causes the following error message in the `iframe-phone` library, seen in the developer console:  

        cant handle type: htmlFragResponse

 4. After selecting a Google basemap (Terrain or Hybrid), the CSS fails to render correctly henceforth in any screenshot until the app is reloaded:  
  ![screenshot from 2015-09-01 11 00 04](https://cloud.githubusercontent.com/assets/1014341/9607701/9d68cbf4-5098-11e5-97d4-afdbf0ad091a.png)
 5. Google Maps apparently does not allow screenshots without attribution, so in cases when they are set as basemaps we must switch to a different layer under the hood, take the screenshot, and then switch back.
 6. It is unclear what the UI is within an activity in ITSI for taking screenshots. While the [Lara Interactive sandbox](http://concord-consortium.github.io/lara-interactive-api/) has a "Screenshot" button, I couldn't find it within ITSI itself.

See comments on #738 for more discussion.

scytacki:
Hi guys, if you want to break these into separate issues it would be easier to discuss them.

For #5 take a look here: https://github.com/concord-consortium/shutterbug.js#shutterbug-jquery-custom-events
the `saycheese` and `asyouwere` events are intended for this.  For example we've used them change from webgl rendering to non-webgl rendering.  If you can switch the base layer in the background then you could do that on `saychesse`, and then switch it back on `asyouwere`.

#3 should be harmless

#2 will probably require some more discussion

rajadain:
Thanks @scytacki. Tagging @mmcfarland

rajadain:
A new issue seen on Staging: taking a screenshot when a user is logged in shows the login screen:

![image](https://cloud.githubusercontent.com/assets/1430060/9619549/24a90908-50df-11e5-93b9-0b77f78fc704.png)

rajadain:
Another issue is sometimes the screenshot is of the login screen because it is accessing the URL from an endpoint which hasn't logged in.

[...]

PR #795, opened 2015-09-15
Correct slider values in ITSI screenshots

rajadain:
## Overview

The PhantomJS parser assigns a default mid-point value to sliders when no "value" attribute is present on the input tag. We add that attribute and ensure that it is updated every time the value changes for all three sliders: precipitation, stream, and opacity.

## Testing Instructions

Checkout the branch and ensure that ITSI is correctly configured in your environment as per the README. Then, run `ngrok`:

```
$ ngrok http 8000
```

since we can't use the [Lara Interactive Playground](http://concord-consortium.github.io/lara-interactive-api/) from `localhost:8000`. Then, open the URL of your `ngrok` instance, akin to `https://????????.ngrok.io/?itsi_embed=true` in the iframe area. Navigate to the project details / modeling screen, and take a screenshot. The slider values in the screenshot should be correctly rendered:

image

image

Also test the compare view, the stream slider, and the opacity slider.

Connects #747, partially.

### DIFF
--- a/src/mmw/js/src/core/opacityControl.js
+++ b/src/mmw/js/src/core/opacityControl.js
@@ -57,6 +57,7 @@ module.exports = L.Control.extend({
     onAdd: function (map) {
         var opacity_slider_div = L.DomUtil.create('div', 'opacity_slider_control'),
             opacityLayer = this.opacityLayer,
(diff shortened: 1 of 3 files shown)


mmcfarland:
+1 Looks good.

PR #800, opened 2015-09-16
Upgrade node-sass to 3.3.2

kdeloach:
This may fix some problems we have experienced with "watch" sometimes not
rebuilding the sass bundle.


kdeloach:
Test to see if this fixes your file watching issues... I didn't really notice any difference (except it runs a little slower) but it seems like the latest version has a lot of fixed around file watching.

rajadain:
+1 the sass files are being watched correctly now. Also tested with random changes to js files and they were picked up as well.

PR #801, opened 2015-09-16
After SJS HTTP socket binds, wait an additional second

hectcastro:
We were attempting to `POST` the SJS job JAR too quickly after the SJS service starts. This adds a one second sleep before we attempt the first `POST` (in addition to the `nc` in a loop at the end of the SJS service definition waiting for a successful connection to port 8090).


mmcfarland:
:+1: I wasn't experiencing this issue prior, but provisioning the worker on this branch still starts geop for me on startup.

hectcastro:
I couldn't reproduce it on the Vagrant setup either, but it reliably fails in the AMI cutting process. The SJS service start blocks until the socket is active, but somehow the `POST` is happening too quickly between the activation and when SJS is actually up.

If this starts happening again, we may need to address that SJS startup in a different way.

PR #802, opened 2015-09-16
Add more stuff to the vendor bundle

kdeloach:
This shrinks `main.js` from 488K to 365K and should make bundling
slightly faster.

The vendor bundle is now 2.5MB unminified and 1.2MB minified.


rajadain:
+1, bundle is now _much_ smaller and all tests still pass. Great job!

Issue #808, opened 2015-09-17
Remove extra width in Compare view in ITSI screenshot


rajadain:
Compare view screenshots look like this:

![image](https://cloud.githubusercontent.com/assets/1430060/9945008/ccfef768-5d58-11e5-8b8e-887aa5610769.png)

Because PhantomJS doesn't handle `vw`, which was used to fix #642. Find a solution that works for both cases.

rajadain:
Also see related, from https://github.com/WikiWatershed/model-my-watershed/pull/1025#issuecomment-157739010:

> The additional columns are now showing up:

> ![screenshot from 2015-11-18 09 52 21](https://cloud.githubusercontent.com/assets/1014341/11244157/179e9ba0-8dda-11e5-893f-ec9550a9a6ab.png)

> but the screenshot tries to fit all of the columns in on the shot - not sure if this was the behavior from before or is related to this changeset.  Also not sure what the expected behavior should be, since we can't fit them all in at once on the actual screen.

> https://ccshutterbugtest.s3.amazonaws.com/293cdeef199.png

ajrobbins:
Compare View Revamp makes this issue moot.

PR #809, opened 2015-09-17
Show Compare Map in Screenshot

### LINKED ISSUE #642: Height keeps growing in /compare on Firefox
Before adding a kludge to fix it, in the compare views in Firefox, every time a drop down was clicked, the map-regions grew vertically. The kludge was to make the sum of the heights of the regions in each column sum to 95%. This fixed the problem for me on Firefox 38.0.5, but not for @rajadain on Firefox 39.0. We should find a better solution if possible. See #596.

rajadain:
## Overview

PhantomJS cannot handle `vh` and `vw` units, thus did not render the map in the Compare page. By replacing those values with percent units, the map renders.

## Testing Instructions

Checkout the branch and ensure that ITSI is correctly configured in your environment as per the README. Then, run `ngrok`:

```
$ ngrok http 8000
```

since we can't use the [Lara Interactive Playground](http://concord-consortium.github.io/lara-interactive-api/) from `localhost:8000`. Then, open the URL of your `ngrok` instance, akin to `https://????????.ngrok.io/?itsi_embed=true` in the iframe area. Navigate to the compare view and take a screenshot. The maps in the screenshot should be correctly rendered:

image

Before:
image

After:
image

## Notes

The extraneous width in the screenshot comes from the usage of `vw` to fix #642. We will need a solution that handles both cases. Created #808 for that.

Connects #747

### DIFF
--- a/src/mmw/sass/pages/_compare.scss
+++ b/src/mmw/sass/pages/_compare.scss
@@ -24,7 +24,7 @@
     .scenario-title{
         text-shadow: 0px 0px 5px black;
         position: absolute;
-        top: 140px;
+        top: 5px;
         z-index: 101;
         left: 15px;
         opacity: 1;
@@ -64,7 +64,7 @@
           }
 
           .fa-minus-square{
-            opacity: 54%!important;
+            opacity: 0.54 !important;
           }
 
           .map-region{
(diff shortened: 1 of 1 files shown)


mmcfarland:
+1, works well.

Issue #804, opened 2015-09-17
Project URL Not Updated in ITSI


rajadain:
When a user moves from "Analyze" to "Model" view, the ITSI saved state should be updated with the project and scenario IDs and point to the "Model" view, but instead they continue to point to the "Draw" view until a Scenario is clicked:

![image](https://cloud.githubusercontent.com/assets/1430060/9940015/3266489a-5d3b-11e5-8f26-9e683488e1d5.png)

This can cause returning users to create duplicate scenarios if they leave before clicking another scenario on screen.

Ensure that an updated URL is sent to ITSI as soon as the user enters the "Model" view.

PR #821, opened 2015-09-21
Update ITSI with URL of new project upon creation

rajadain:
## Overview

Previously, the project ID would be available to the app but not present in the URL, resulting in a use case when a student who has progressed to the project stage closes the window and returns to the draw stage, since the project URL was never updated / communicated to the containing activity. By ensuring that the project URL is updated as soon as the project stage is entered, we avoid that use case.

## Testing Instructions

 1. Checkout the branch and run [`ngrok http 8000`](https://ngrok.com)
 2. Go to the [LARA Sandbox](http://concord-consortium.github.io/lara-interactive-api/)
 3. Put in your `ngrok` URL `https://????????.ngrok.io/?itsi_embed=true` and load MMW in the iframe
 4. Login as an ITSI user `mmwterence` or `tstudent2`
 5. Create an AoI and move to the Analyze stage
 6. Move to the Project stage
 7. Ensure that you see a similar message in the parent log window:  

        App: interactiveState called with: {"route":"project/277"}

## Demo

image

Connects #804

### DIFF
--- a/src/mmw/js/src/modeling/controllers.js
+++ b/src/mmw/js/src/modeling/controllers.js
@@ -100,6 +100,8 @@ var ModelingController = {
                         // Now render.
                         initViews(project);
                         project.fetchResultsIfNeeded();
+
+                        updateUrl();
                     });
                 } else {
                     initViews(project);


mmcfarland:
+1 - Successfully tested sequence listed above.

rajadain:
Thanks for reviewing!

Issue #813, opened 2015-09-18
Analysis `Area` Output is not sorted correctly


mmcfarland:
When sorting the table values on the `Area` column, the numbers are not appropriately sorted.  Note that the `Coverage` column _is_ sorted correctly.

![screenshot from 2015-09-18 15 23 27](https://cloud.githubusercontent.com/assets/1014341/9968970/adc228fe-5e19-11e5-9f27-65ab135372c1.png)

mmcfarland:
As @jamesmcclain [pointed out](https://github.com/WikiWatershed/model-my-watershed/issues/812#issuecomment-141696557), the numbers are being sorted correctly lexicographical and not numerically.

mmcfarland:
`bootstrap-table`'s default sort example also sorts numerical data incorrectly:  http://issues.wenzhixin.net.cn/bootstrap-table/#options/basic-sort.html

jamesmcclain:
It looks as though there is a way to define a custom sorting function.

PR #822, opened 2015-09-21
Sort numeric columns numerically

jamesmcclain:
This patch causes numeric columns of the tables on the `/analyze` and `/project` pages to be sorted numerically instead of lexicographically.

Connects #813

### DIFF
--- a/src/mmw/js/src/main.js
+++ b/src/mmw/js/src/main.js
@@ -1,5 +1,7 @@
 "use strict";
 
+var utils = require('./core/utils');
+
 require('./core/setup');
 require('./routes');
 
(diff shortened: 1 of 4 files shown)


src/mmw/js/src/main.js
```diff
@@ -29,6 +29,19 @@ App.on('start', function() {
 
 App.start();
 
+window.numericSort = function(_x, _y) {
```
kdeloach: I think this is the right place to attach this to window but consider putting the sort function in its own module. Also, consider adding a comment why this needs to be attached to window.

kdeloach:
+1

jamesmcclain:
Thanks for reviewing.

caseypt:
Consider dropping that function into this existing utility module: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/utils.js

jamesmcclain:
Okay, will-do.

PR #827, opened 2015-09-22
Clean scss

jfrankl:


mmcfarland:
+1 - Cleaner code and found no side effects in the site.

Issue #812, opened 2015-09-18
Format Analysis Output with 2 digits of precision


mmcfarland:
Sorting on `Area` can produce visually confusing results since some values don't report a second digit of precision if it's `0`, making it look like a smaller number than following values.

![screenshot from 2015-09-18 15 23 27](https://cloud.githubusercontent.com/assets/1014341/9968903/5db27ef4-5e19-11e5-8dae-0686529660ae.png)

jamesmcclain:
It looks like it is sorting lexicographically instead of numerically.

jamesmcclain:
It seems like some of this could also be due to the fact that we are using a variable-width font for these numbers.

PR #828, opened 2015-09-22
Pad results in /analyze table

jamesmcclain:
Values in the "Area" column of the `/analyze` table are now padded to at least two digits.

**To Test**
   * Click around until you find an AoI where some of the values in the "Area" column end in zeros  (you may have to try a few times before you can find one).  Previously, those would have had zero or one digit after the decimal point.

screenshot from 2015-09-22 15 19 33

Connects #812

### DIFF
--- a/src/mmw/js/src/analyze/templates/tableRow.html
+++ b/src/mmw/js/src/analyze/templates/tableRow.html
@@ -1,3 +1,3 @@
 <td>{{ type }}</td>
-<td class="strong text-right">{{ scaledArea|round(2)|toLocaleString }}</td>
+<td class="strong text-right">{{ scaledArea|round(2)|toLocaleString(2) }}</td>
 <td class="strong text-right">{{ coveragePct|toFixed(1) }}</td>
--- a/src/mmw/js/src/analyze/tests.js
+++ b/src/mmw/js/src/analyze/tests.js
@@ -106,7 +106,7 @@ function checkTableBody(subData) {
         // displayed.
         var expectedRowVals = [
             subData.categories[trInd].type,
-            subData.categories[trInd].area.toLocaleString('en'),
+            subData.categories[trInd].area.toLocaleString('en', {minimumFractionDigits: 2}),
             (subData.categories[trInd].coverage * 100).toFixed(1)
         ];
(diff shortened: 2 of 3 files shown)


rajadain:
+1 tested and works well:

image

0 value percentages are reported as `0.0`, with one digit after decimal. However, 0 value areas are reported as just `0`, with no places, even though we ensure that all other values have two digits after decimal. Consider ensuring that 0 value areas are reported as `0.00` for visual consistency.

src/mmw/js/src/analyze/templates/tableRow.html
```diff
@@ -1,3 +1,3 @@
 <td>{{ type }}</td>
-<td class="strong text-right">{{ scaledArea|round(2)|toLocaleString }}</td>
+<td class="strong text-right">{{ scaledArea|round(2)|toLocaleString|pad(2) }}</td>
```
rajadain: [`toLocaleString`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) also takes optional parameters which tell it to display a certain number of `minimumFractionDigits`. It may be simpler to modify the [`toLocaleString` filter](https://github.com/WikiWatershed/model-my-watershed/blob/feb1affef89f412d038ff47385851bb7c5ad38d1/src/mmw/js/src/core/filters.js#L7-L9) to take an optional parameter which does the parsing for us.

jamesmcclain: Good idea, I will loo into it.

jamesmcclain:
> Consider ensuring that 0 value areas are reported as 0.00 for visual consistency.

Okay, will-do.

rajadain:
You may have to update some failing tests.

jamesmcclain:
Yeah, I just noticed that.  Must be the 0 versus 0.00 issue (because visually that is the only thing that has changed since the last time this PR passed).

rajadain:
+1 looks and works great now

image

jamesmcclain:
Thanks, there was actually very little to do after your comment about `toLocaleString` already having this ability.

PR #826, opened 2015-09-22
Make asynchronous calls to SJS

jamesmcclain:
SJS's asynchronous mode is now used.  That works in the following way: after the initial request is made, the requester is given a job ID which it can periodically check until the job has either succeeded or failed.

The polling is done by using a celery task which checks the SJS endpoint to see if the result is ready.  If so, it is returned; if not the task asks to be put back onto the queue and run again in two seconds.

Connects #701
Connects #823

Would like to create separate issues for:
   * Gathering metrics (https://github.com/WikiWatershed/model-my-watershed/issues/871)
   * Caching of censuses (https://github.com/WikiWatershed/model-my-watershed/issues/870)

Issue #815, opened 2015-09-21
Communicate geoprocessing failure to front-end and stop polling.


caseypt:
When an error/exception like https://github.com/WikiWatershed/model-my-watershed/issues/811 happens in the geoprocessing pipeline, we should respond to the client-side so that it can stop polling and report the failure, rather than waiting for the client-side timeout to occur.

Issue #811, opened 2015-09-18
Geoprocessing division error


mmcfarland:
Running a sq km AoI as seen below resulted in the following error reported from my `debugcelery` window.  Note some values are still reported.  Drawing very similar AoIs don't always reproduce this error and I failed to save the exact geometry of the AoI :disappointed: 

We should try to protect against doing calculations when `count` is 0.

![screenshot from 2015-09-18 15 17 55](https://cloud.githubusercontent.com/assets/1014341/9968788/784c9958-5e18-11e5-89a3-2cecdc82312d.png)

Error:
```bash
[2015-09-18 15:14:34,596: ERROR/MainProcess] Task apps.modeling.tasks.run_analyze[288065ed-c629-4192-8b00-91a1f7348199] raised unexpected: ZeroDivisionError('float division by zero',)
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/opt/app/apps/modeling/tasks.py", line 42, in run_analyze
    results = geojson_to_survey(area_of_interest)
  File "/opt/app/apps/modeling/geoprocessing.py", line 224, in geojson_to_survey
    return data_to_survey(data)
  File "/opt/app/apps/modeling/geoprocessing.py", line 214, in data_to_survey
    return histogram_to_x(data, nucleus, update_rule, after_rule)
  File "/opt/app/apps/modeling/geoprocessing.py", line 97, in histogram_to_x
    after_rule(total, retval)
  File "/opt/app/apps/modeling/geoprocessing.py", line 198, in after_rule
    survey[0]['categories'] = [update_pcts(v, count) for k, v in land]
  File "/opt/app/apps/modeling/geoprocessing.py", line 168, in update_pcts
    entry['coverage'] = float(area) / count
ZeroDivisionError: float division by zero
```

caseypt:
I can reproduce this consistently with the following AoI (working on getting my host to be shareable):

![screen shot 2015-09-21 at 11 24 34 am](https://cloud.githubusercontent.com/assets/1042475/9997663/6d260100-605b-11e5-8164-8ec04f2e144b.png)

This is caused by bad results from the geoprocessing service. If you make a quick fix for this error ...

```
[...]

PR #829, opened 2015-09-22
Stop polling for Celery job if job fails.

caseypt:
* Adjust the client-side polling logic to stop polling Celery if the
the job being polled has failed. The response from Celery is a
200 even if the job completes, which is why the errors/failed jobs
were not being caught by the fail function of the polling promise.

The change in 9c5659d is needed for testing, otherwise the client-side timeout will be reached before the job has failed. It doesn't necessarily need to be included, though it will probably be increasing due to #826.

**Testing:**

- Open a new terminal window and run `scripty debugcelery`.
- Visit the MMW homepage.
- Open the JS debugger.
- Draw a large AoI thats at least covers half of PA.
- Watch the Celery log and the console log. After about 10 seconds, Celery should report #811.
- At this time, the app should stop polling and show the error message, even though the timeout hasn't been hit. If it's hard to see. Increase the client-side timeout to something much higher, like one minute. 

Connects to #815

### DIFF
--- a/src/mmw/js/src/core/models.js
+++ b/src/mmw/js/src/core/models.js
@@ -165,11 +165,13 @@ var TaskModel = Backbone.Model.extend({
             self.fetch()
                 .done(function(response) {
                     console.log('Polling ' + self.url());
-                    if (response.status !== 'complete') {
+                    if (response.status === 'started') {
                         elapsed = elapsed + self.get('pollInterval');
                         window.setTimeout(getResults, self.get('pollInterval'));
-                    } else {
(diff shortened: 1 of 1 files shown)


lliss:
This looks good. Minimal code changes... very nice.

One note is that since the spark job keeps going in the background, starting a huge download for half of PA caused the new jobs to fail for quite a while. This should be fixed by the PR noted in your comment but for testing I had to jump onto the worker and restart the geoprocessing service as well as spark job server.

+1 tested.

Issue #816, opened 2015-09-21
Clip modifications that intersect with an AoI 


mmcfarland:
When modifications are drawn to an Area of Interest, the area used for calculation must be clipped to be internal to the bounding polygon.  

I think there is a good argument for not preventing the drawn modification from being added to the map or to be visibly trimmed on the outside of the border.  If a user is studying a specific area of interest and sketching out an existing plan that interests that area, we should allow them to draw the full plan or as-built without unnecessary friction - perhaps with a message/indication that the entire sketch is not being used for calculations in the model.

PR #830, opened 2015-09-23
Clip modifications to AoI for modeling purposes.

caseypt:
* Clip modifications to the boundary of the area of interest and save
the resulting shape and area as effectiveShape and effectiveArea,
respectively.
* Add tests to exercise effective field functionality.
* Update modification templates to display effective area.
* Add message to bottom of modification dropdown that describes
effective area.

**Testing instructions:**
- Start a new project.
- Progress to the modeling page.
- On an editable scenario, draw a modification that overlaps the AoI border. The effective area of the modification should be displayed in the modification pop-up, as well as in the modification drop down.
- Refresh the page. The modification area should still represent the area within the AoI.
- Run the tests.

Connects to #816

### DIFF
--- a/src/mmw/js/src/core/templates/modificationPopup.html
+++ b/src/mmw/js/src/core/templates/modificationPopup.html
@@ -2,7 +2,11 @@
 <table class='table'>
     <tr>
         <td>Total Area</td>
-        <td class='strong text-right'>{{ area|round(1)|toLocaleString }} {{ units }}</td>
+        <td class='strong text-right'>{{ area|round(2)|toLocaleString(2) }} {{ units }}</td>
+    </tr>
+    <tr>
+        <td>Effective Area</td>
+        <td class='strong text-right'>{{ effectiveArea|round(2)|toLocaleString(2) }} {{ effectiveUnits }}</td>
     </tr>
 </table>
 <span>
--- a/src/mmw/js/src/modeling/templates/scenarioToolbarTabContent.html
+++ b/src/mmw/js/src/modeling/templates/scenarioToolbarTabContent.html
@@ -15,18 +15,24 @@
             <div id="modifications" class="dropdown-menu menu-right" role="menu"> <!-- DropdownContent -->
                 {% for controlName, models in groupedShapes %}
                     <div class="pad-1 brd-bottom">
-                        <label class="medium">
-                            {% if controlName == 'landcover' %}
-                                Land Cover
-                            {% elif controlName == 'conservation_practice' %}
-                                Conservation Practice
-                            {% endif %}
-                        </label>
                         <table id="mod-{{ controlName|lower|replace(' ', '') }}" class="table modifications custom-hover">
+                            <thead>
+                                <tr>
+                                    <td class="text-left modification-type">
+                                        {% if controlName == 'landcover' %}
+                                            Land Cover
+                                        {% elif controlName == 'conservation_practice' %}
+                                            Conservation Practice
+                                        {% endif %}
+                                    </td>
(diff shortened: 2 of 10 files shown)


rajadain:
The effective area clipping works well when tested against the AoI boundary:

screen shot 2015-09-24 at 10 32 22 am

It does not work (contrary to the message) when there are overlapping modifications:

screen shot 2015-09-24 at 10 33 24 am

Also, the formatting of values, which skip decimal point when `.0`, makes it visually look like the effective area is larger than the total area in some cases:

screen shot 2015-09-24 at 10 34 30 am

caseypt:
@rajadain regarding the clipping of features by features, I'm adding a card for that because it's a bit complicated. It's currently being done, but there isn't a way to link back the clipped shape with the original modification.

I'll fix the digits thing.

caseypt:
Updated with padded numbers:

screen shot 2015-09-24 at 12 03 28 pm

src/mmw/js/src/core/templates/modificationPopup.html
```diff
@@ -2,7 +2,11 @@
 <table class='table'>
     <tr>
         <td>Total Area</td>
-        <td class='strong text-right'>{{ area|round(1)|toLocaleString }} {{ units }}</td>
+        <td class='strong text-right'>{{ area|round(1)|toLocaleString(2) }} {{ units }}</td>
+    </tr>
+    <tr>
+        <td>Effective Area</td>
+        <td class='strong text-right'>{{ effectiveArea|round(1)|toLocaleString(2) }} {{ effectiveUnits }}</td>
```
rajadain: Cool reuse of `toLocaleString` for padding. If we're rounding to 1 digit, should we also display only 1 fractional digit? Or round to 2 for visual consistency with the tables? Rounding to 1 and but showing 2 to gives a false impression of accuracy.

caseypt: Good catch. Will switch to 2.

caseypt:
Updated:

screen shot 2015-09-24 at 12 08 43 pm

src/mmw/js/src/modeling/templates/scenarioToolbarTabContent.html
```diff
@@ -15,18 +15,24 @@
...
+                                    </td>
+                                    <td class="text-right">Effective Area</td>
+                                    <td></td>
+                                </tr>
+                            </thead>
                             {% for model in models %}
                                 <tr>
                                     <td>{{ model.get('value')|tr55Name }}</td>
-                                    <td class="strong text-right">{{ model.get('area')|round(1)|toLocaleString }} {{ model.get('units') }}</td>
+                                    <td class="strong text-right">{{ model.get('effectiveArea')|round(1)|toLocaleString }} {{ model.get('effectiveUnits') }}</td>
```
rajadain: Might want to use `round(2)|toLocaleString(2)` here too.

I wonder if we can forego `round` and just use `toLocaleString`? But that may clip instead of round.

caseypt: Yeah, I'll switch it over.

caseypt:
Updated: 

screen shot 2015-09-24 at 12 25 35 pm

rajadain:
+1 tested

image

You may have to adjust one last test:

```
not ok 65 Firefox 41.0 - Modeling Views ToolbarTabContentView lists all of the modifications and their effective area
    ---
        message: >
            expected '44.42 km2' to equal '44.4 km2'
    ...
```

caseypt:
@rajadain. Thanks. Sorry for all the quick fixes.

PR #832, opened 2015-09-24
Replace livereloadify with livereload

kdeloach:
This project has much better documentation and supports injecting changes
live on the page without having to do a full page refresh.

Reprovision the app VM then run these commands:

`./scripts/bundle.sh --debug --watch`

`./scripts/npm.sh run livereload`

If you have the Chrome livereload extension, any changes to the SASS files should be injected on to the page in real time.


caseypt:
Very cool. I'll give it a try.

jfrankl:
+1

This works great Kevin—thank you! I think your build system is ready for a TSM.

Issue #844, opened 2015-09-24
Remove "Add Tags" and "Print" as project options


arobbi7:
Since these don't have functionality yet, they should be removed since the site is already starting to be used by teachers, etc.

Issue #838, opened 2015-09-24
Remove "I Agree to Something"


mmcfarland:
The user sign up process includes a place holder for a Terms of Service agreement.  Hide this element and make it non-required.  Users signing up should not have this field marked as accepted, ToS will most likely be introduced in the future and users who have not agreed will be prompted to review and agree to the terms at that time

kdeloach:
Shares a point with #844

PR #855, opened 2015-09-24
Remove Items

rajadain:
## Overview

Removes "Add Tags" and "Print" menu options from the Project Menu. Also removes the "I Agree to Something" checkbox from Sign Up and ITSI Sign Up modals.

## Testing Instructions

Test the Project Menu while logged in, when logged out, and when viewing public projects as third-party.
Test the sign-up procedure.
Run all tests.

## Notes

Should I also remove the "Print" option from the Scenario Menu?

Connects #844 
Connects #838

### DIFF
--- a/src/mmw/apps/user/views.py
+++ b/src/mmw/apps/user/views.py
@@ -155,8 +155,6 @@ def itsi_sign_up(request):
         errors.append("First name must be specified")
     if 'last_name' not in request.POST or not request.POST.get('last_name'):
         errors.append("Last name must be specified")
-    if 'agreed' not in request.POST or not request.POST.get('agreed'):
-        errors.append("You must agree to the terms")
 
     if len(errors) > 0:
         response_data = {"errors": errors}
--- a/src/mmw/js/src/modeling/templates/projectMenu.html
+++ b/src/mmw/js/src/modeling/templates/projectMenu.html
@@ -3,27 +3,27 @@
 </div>
 <div class="project-title">
     <h3 class="light">{{ name }}</h3>
-    <div id="project-settings" class="dropdown pull-right">
-        <button class="btn btn-sm btn-icon light dropdown-toggle" type="button" data-toggle="dropdown" aria-expanded="true">
-        <i class="fa fa-caret-down"></i>
-        </button>
-        <ul class="dropdown-menu menu-right" role="menu">
-            {% if not is_new and editable %}
-                <li role="presentation"><a role="menuitem" tabindex="-1" data-toggle="modal" id="share-project">Share</a></li>
-                <li role="presentation"><a role="menuitem" tabindex="-1" data-toggle="modal" id="project-privacy">{% if is_private %}Make Public{% else %}Make Private{% endif %}</a></li>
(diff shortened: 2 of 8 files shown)


mmcfarland:
Looks good.  I poked around and even though we're sending an "agreed" value to the server, it looks like we're not persisting it anywhere.  I just want to make sure that we can distinguish in the future users who have signed up and not actually "agreed" after we do introduce a real ToS.

mmcfarland:
> Should I also remove the "Print" option from the Scenario Menu

Yep, good catch.

rajadain:
Re: distinguishing users who haven't explicitly agreed to a yet unwritten ToS, since we don't persist the agreement it will just have to be date of sign up vs the date we deploy the ToS.

Will remove "Print" from Scenario Menu as well.

rajadain:
"Print" removed from Scenario Menu in https://github.com/rajadain/model-my-watershed/commit/e2dd613f69627443068b41a9ebb981832cf9f0a0

src/mmw/js/src/user/templates/baseModal.html
```diff
@@ -16,6 +16,15 @@
     </div>
 {% endmacro %}
 
+{% macro hidden_checkbox(value, name, label='') %}
+    <div class="checkbox" style="display: none;">
+        <label>
+            <input name="{{ name }}" type="checkbox" id="{{ name }}" checked>
```
lliss: According to the card (emphasis added)
> Users signing up should **not** have this field marked as accepted

I think we probably will also want to clear out that dummy text as well.

rajadain:
@lliss the ToS Agreement is **completely** removed in https://github.com/rajadain/model-my-watershed/commit/36c499eb5fe62fe75af504508407390f0e10b728. Is this closer to your interpretation of the card?

lliss:
> @lliss the ToS Agreement is completely removed in rajadain@36c499e. Is this closer to your interpretation of the card?

It's fine with me since it's fairly easy to add back. I thought we'd leave the checkbox there but hidden and remove the silly "I agree to something" text and the validation checks for it. So with the exception of the deletion:

```
-{% macro hidden_checkbox(value, name, label='') %}
-    <div class="checkbox" style="display: none;">
-        <label>
-            <input name="{{ name }}" type="checkbox" id="{{ name }}" checked>
-            {{ label }}
-        </label>
-    </div>
-{% endmacro %}
``` 

this is pretty much what I had in mind.

rajadain:
The [original checkbox macro](https://github.com/WikiWatershed/model-my-watershed/blob/4f16ea139a899eb85b9db81df2befd74e39220fc/src/mmw/js/src/user/templates/baseModal.html#L10-L17) is still there, so it should be easy to add it back.

This does look cleaner, and when we want to add a checkbox it may be easier to add it back to our then current style, rather than going back and un-commenting code, which is why I just outright deleted it. Git will preserve history.

lliss:
I totally agree. +1

lewfish:
+1, tested

rajadain:
Just pushed https://github.com/rajadain/model-my-watershed/commit/470ea8a044355db08f000bf478d3a59056e22166 which addresses #862. Will squash and merge shortly.

Issue #836, opened 2015-09-24
NLCD Chart Order by Class ID


mmcfarland:
The order of the NLCD chart on the Geospatial Analysis page should be that of the NLCD class ID, ascending.

caseypt:
http://www.mrlc.gov/nlcd01_leg.php

mmcfarland:
NLCD Class Id should be a part of the response as of 59881a94a6c0c016b78e2a0ca14a7bc0374ef244 under the `nlcd` key.

PR #856, opened 2015-09-25
Order analyze chart by NLCD class number

caseypt:
* In the bar chart on the analyze page, order the land uses categories
by NLCD class number. This puts the categories in an order that matches the
official order and legend (http://www.mrlc.gov/nlcd01_leg.php).

**Testing instructions:**
- Select an area.
- Verify that the order of the NLCD categories in the bar chart matches the order here: http://www.mrlc.gov/nlcd01_leg.php

screen shot 2015-09-25 at 1 53 18 pm

Connects to #836

### DIFF
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -268,13 +268,12 @@ var ChartView = Marionette.ItemView.extend({
 
     addChart: function() {
         var chartEl = this.$el.find('.bar-chart').get(0),
-            chartData = this.collection.map(function(model) {
-                var attrs = model.attributes;
+            chartData = _.map(this.collection.toJSON(), function(model) {
                 return {
-                    area: attrs.area,
-                    coverage: attrs.coverage,
-                    type: attrs.type,
-                    className: attrs.nlcd ? 'nlcd-' + attrs.nlcd : null
+                    area: model.area,
+                    coverage: model.coverage,
+                    type: model.type,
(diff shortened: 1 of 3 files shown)


src/mmw/js/src/analyze/views.js
```diff
@@ -268,15 +268,16 @@ var ChartView = Marionette.ItemView.extend({
...
-            chartData = this.collection.map(function(model) {
-                var attrs = model.attributes;
-                return {
-                    area: attrs.area,
-                    coverage: attrs.coverage,
-                    type: attrs.type,
-                    className: attrs.nlcd ? 'nlcd-' + attrs.nlcd : null
-                };
-            }),
+            chartData = _.chain(this.collection.toJSON())
```
lliss: Cool.

lliss:
+1. Looks good.

lliss:
+1. One question though... Should the the default sort order for the table also match this order?

rajadain:
I agree with the suggestion above: if we can't have the orders _always_ match, we should at least have them _initially_ match.

caseypt:
@lliss Good suggestion. I'll make an update.

caseypt:
Updated:

screen shot 2015-09-28 at 9 01 47 am

rajadain:
+1, great job!

caseypt:
Added one more commit to make things more clear.

mmcfarland:
+1

Issue #835, opened 2015-09-24
Align Select by Boundary and Overlay Boundaries


mmcfarland:
The two lists should share all the same features and should appear in the following order in both:

* HUC 8
* HUC 10
* HUC 12
* US counties 
* Congressional districts
* School districts

PR #858, opened 2015-09-25
Reorder and turn on additional layers.

caseypt:
* Reorder boundary and overlay layers according to layer order
preferred by Stroud.
* Turn on Schools and Counties as boundary layers.

**Testing instructions:**
- Verify that the layers boundary selector and the layers in the boundary section of the overlay control are in the same order.

screen shot 2015-09-25 at 2 34 49 pm

Connects to #835

### DIFF
--- a/src/mmw/mmw/settings/layer_settings.py
+++ b/src/mmw/mmw/settings/layer_settings.py
@@ -59,7 +59,7 @@ LAYERS = [
     },
     {
         'code': 'county',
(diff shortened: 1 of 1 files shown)


lliss:
+1. Tested.

caseypt:
@mmcfarland Updated with text change.

mmcfarland:
+1

Issue #834, opened 2015-09-24
Add 100% Forest column to Compare View


mmcfarland:
Similarly to the added "Current Conditions" column, add the model output for 100% Forest as a default column in the Compare View.

PR #859, opened 2015-09-25
Show 100% Forest Cover scenario in Compare View

rajadain:
## Overview

Since the Compare View needs scenarios, and 100% Forest Cover data is contained inside the Current Conditions scenario, we make a copy of it with `is_current_conditions` set to `false` and `is_pre_columbian` set to `true` before passing it to the Compare View.

When rendering TR55 Runoff views, we check if the current scenario is Current Conditions, Pre Columbian, or a regular one in Compare View. We use `unmodified`, `pc_unmodified`, or `modified` results respectively.

## Testing Instructions

Take any project to the Compare View. Update sliders, charts. Ensure that the view works as it should. Come back to Model View. Ensure that there is no new 100% Forest Cover scenario here.

## Demo

image

Connects #834

### DIFF
--- a/src/mmw/js/src/compare/controllers.js
+++ b/src/mmw/js/src/compare/controllers.js
@@ -16,6 +16,7 @@ var CompareController = {
     compare: function(projectId) {
         if (App.currentProject) {
             setupProjectCopy();
+            addForestCoverScenario();
             showCompareWindow();
         } else if (projectId) {
             App.currentProject = new modelingModels.ProjectModel({
@@ -25,6 +26,7 @@ var CompareController = {
                 .fetch()
                 .done(function() {
                     setupProjectCopy();
+                    addForestCoverScenario();
                     showCompareWindow();
                 });
         }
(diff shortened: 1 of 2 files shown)


lliss:
+1. Looks good. Probably not related but I did notice that the precipitation slider does not persist after changing it in compare view.

rajadain:
> the precipitation slider does not persist after changing it in compare view.

I believe that is by design, since changing the precipitation of one changes all, and you may have them set-up a certain way in the Modeling View which we don't want to overwrite.

mmcfarland:
Looks good visually, but perhaps makes more sense as the first column?  It'll almost always be the outlier in the group, and as a stand-in for "pre-columbian" has a certain chronological order to go Forest -> Current -> Modified

We may want to be consistent and have it be the first bar on the CC tab to keep it consistent.

rajadain:
100% Forest now precedes Current Conditions in both Compare and Model Views, as of https://github.com/WikiWatershed/model-my-watershed/commit/a0a32f29e1a395027b81226450da519d48c906e6.

image

image

jamesmcclain:
It looks like other folks had a chance to chime in before I did, but looks good to me.

+1 Tested.

Issue #782, opened 2015-09-09
Add Feature Switching based on Environment


rajadain:
Arising out of issues such as #778, there is a need to enable different features across different environments. Investigate feature switching options, such as https://github.com/jsocol/django-waffle, and implement a mechanism that will allow different features in different environments.

kdeloach:
Can we close this?

rajadain:
From @hectcastro [here](https://github.com/WikiWatershed/model-my-watershed/pull/861#issuecomment-143776906):

> Technically, I think it can, but I also think that this approach doesn't scale very well. I'm fine with using it for the number of dynamic settings we have across environments now, but I would prefer to work toward a more flexible solution if we plan on adding a dozen more settings.
> 
> Right now, the AMI will get baked using whatever settings are put in place by Ansible. That same AMI will be used in all EC2 environments, but stacks will have the option to override specific settings via CloudFormation parameters.

So I'd get input from @mmcfarland if he envisions us having a lot more environment variables going forward or not. If not, we can close this.

mmcfarland:
I do not anticipate many more environment based features, I think we should close and revisit if that assumption proves false.

PR #861, opened 2015-09-28
Make ITSI credentials and base URL customizable per stack

### LINKED ISSUE #833: Update ITSI Portal URL for oAuth
Our `mmw` deployment will be the production deployment and can get pointed to use the production ITSI portal environment.  Existing credentials will be ported over shortly from `staging.learn`.

##### New URL
`itsi.portal.concord.org`

The existing `mmw-dev` site should remain pointed at the existing staging portal site.

hectcastro:
When new stacks are launched, parameters that change across stacks are now injected at launch time. This allows us to have a different set of ITSI credentials and base URL for staging vs. production.

Attempts to resolve #833.

---

**Testing**

Using the staging stack settings file on Jenkins as a base (`/var/lib/jenkins/.aws/staging.yml`), try launching a stack of the opposite color that's currently active:

```bash
$ cd deployment
$ ./mmw_stack.py launch-stacks --aws-profile mmw-stg --mmw-config-path default.yml --mmw-profile staging --stack-color green --activate-dns
```

Once launched, login to the following instances and confirm that the settings meant to change across stacks are present:

- `AppServer`
- `TilerServer`
- `Worker`

In addition, you can query the EC2 metadata API to inspect the user data for each instance type:

```bash
$ curl "http://169.254.169.254/latest/user-data/"
```


rajadain:
Would this also address #782? Since now we can have different variables per environment? Or do we promote the same AMI to production which has been deployed to staging?

hectcastro:
Technically, I think it can, but I also think that this approach doesn't scale very well. I'm fine with using it for the number of dynamic settings we have across environments now, but I would prefer to work toward a more flexible solution if we plan on adding a dozen more settings.

Right now, the AMI will get baked using whatever settings are put in place by Ansible. That same AMI will be used in all EC2 environments, but stacks will have the option to override specific settings via CloudFormation parameters.

mmcfarland:
+1.  Verified that development and staging both have the appropriate values set on the individual vagrant/ec2 instances.  oAuth to ITSI on both environment continues to work.

Issue #497, opened 2015-06-24
Leaflet attribution not visible in /model view when results panel minimized


rajadain:
Because of the way the results panel is positioned over the map, the Leaflet attribution is hidden behind it:

Visible Full Size
![image](https://cloud.githubusercontent.com/assets/1430060/8333404/b256a928-1a60-11e5-91fb-fc2e07a5aee7.png)

Visible Half Size
![image](https://cloud.githubusercontent.com/assets/1430060/8333415/bca03f7a-1a60-11e5-8804-6a99e5cdc65c.png)

Not Visible Full Size with Results Panel Header
![image](https://cloud.githubusercontent.com/assets/1430060/8333432/ccdd46c6-1a60-11e5-9810-6918374bb90a.png)

If we get rid of the Half Size view in #401 then the `/model` view will never have the attribution visible. Adjust the map size so that the attribution is visible when the panel is minimized.

caseypt:
Since the analyze and modeling pages have been redesigned, this is no longer an issue.

PR #863, opened 2015-09-28
Keep map sized in container

mmcfarland:
* New classes for specifying `top` and `bottom` positions for a new map container
* Add more size setting methods than just half and full

This is more or less temporary until we move to a sidebar design, which has been approved.  The new version will remove the map size transitions altogether.

Connects #742 
Connects #497

### DIFF
--- a/src/mmw/js/src/core/models.js
+++ b/src/mmw/js/src/core/models.js
@@ -60,13 +60,25 @@ var MapModel = Backbone.Model.extend({
         this.set('areaOfInterest', _.clone(this.get('previousAreaOfInterest')));
     },
 
-    setHalfSize: function(fit) {
-        this.set('size', { half: true, fit: !!fit });
+    setDoubleHeaderSmallFooterSize: function(fit) {
+        this._setSizeOptions({ double: true  }, { small: true }, fit);
     },
 
-    setFullSize: function(fit) {
-        this.set('size', { half: false, fit: !!fit });
+    setDoubleHeaderHalfFooterSize: function(fit) {
+        this._setSizeOptions({ double: true  }, { large: true }, fit);
     },
+
+    setDrawWithBarSize: function(fit) {
+        this._setSizeOptions({ single: true  }, { med: true }, fit);
+    },
+
+    setAnalyzeSize: function(fit) {
+        this._setSizeOptions({ single: true  }, { large: true }, fit);
+    },
+
+    _setSizeOptions: function(top, bottom, fit) {
+        this.set('size', { top: top, bottom: bottom, fit: !!fit });
+    }
 
 });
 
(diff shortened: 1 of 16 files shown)


src/mmw/js/src/core/views.js
```diff
@@ -563,19 +563,26 @@ var MapView = Marionette.ItemView.extend({
...
+        $container.toggleClass('map-container-bottom-1', !!size.bottom.min);
+        $container.toggleClass('map-container-bottom-2', !!size.bottom.small);
+        $container.toggleClass('map-container-bottom-3', !!size.bottom.med);
+        $container.toggleClass('map-container-bottom-4', !!size.bottom.large);
 
-        if (size.fit) {
-            this.fitToAoi();
-        }
+
+        _.delay(function() {
```
mmcfarland: The alternative to adding a delay is to trap a `transitionend` event to fit and invalidate the map size.  This was easier and will be removed, hopefully, in the next month in the new design implementation.

src/mmw/js/src/core/models.js
```diff
@@ -60,14 +60,26 @@ var MapModel = Backbone.Model.extend({
...
 
-    setHalfSize: function(fit) {
-        this.set('size', { half: true, fit: !!fit });
+    setDoubleSmallSize: function(fit) {
+        this._setSizeOptions({ double: true  }, { small: true }, fit);
     },
 
-    setFullSize: function(fit) {
-        this.set('size', { half: false, fit: !!fit });
+    setDoubleHalfSize: function(fit) {
```
caseypt: Consider renaming these to something like `setDoubleHeaderSmallFooterSize' and `setDoubleHeaderHalfFooterSize'.

caseypt:
+1. There are some tests to update, but otherwise looks good. Went through the workflow a couple times, and the map was always sized properly.

screen shot 2015-09-28 at 11 39 49 am

mmcfarland:
Renamed and added representational tests for class existence testing.

rajadain:
+1 tested, map and attribution are always visible

rajadain:
It seems that the top-bar (with login button) is no longer visible in the Model View:

image

Is this intentional?

mmcfarland:
> It seems that the top-bar (with login button) is no longer visible in the Model View:

I bet the new map-container element hides it, I'll have to look into this.

mmcfarland:
@jfrankl - got an idea of why #app-header isn't showing up on the /project pages from your commit?

lliss:
Looks like the culprit is a header rule that is too generic and applies absolute position to two different header elements so the second just sits on top of the first.

lliss:
I find this a little jarring.
screencap

I'm hoping the new design will address this.

lliss:
+1. Tested.

Code looks good. I'm glad, however that this is a temporary measure. The animation logic is somewhat difficult to follow and I think it could be made cleaner with an overhaul.

mmcfarland:
Closing in favor of #873 which has the header fix + these commits.

Issue #840, opened 2015-09-24
Add accent colors to land cover items


mmcfarland:
Each land cover should have its corresponding NLCD class color associated with it as:
* The selected/hover highlight border color around the swatch
* The header of the description for that popup

The soils will have a similar color scheme which isn't defined yet, but we can appropriate temporary colors to get the mechanics down.

PR #864, opened 2015-09-28
Use NLCD color for popover and highlight for micro site

caseypt:
Use NLCD color for popover and highlight
* Set popover to use a custom template that has a dynamic
class for each land use.
* Move popover init code to app JS file.
* Add NLCD color SASS variables.

**Testing Instructions:**
- Visit the micro app.
- Verify that for each land use, the popover header and highlight color match, and that they are the correct NLCD color.
- Verify that the soil choices still use the default color.

screen shot 2015-09-28 at 11 03 59 am

I also updated the text color of the "Grassland" and "Pasture/Hay" popovers because it was hard to read with the new background colors:

screen shot 2015-09-28 at 11 08 43 am

screen shot 2015-09-28 at 11 10 31 am

Also, I created SASS variables for the NLCD colors and updated the analysis chart to use these variables. Verify that the analysis chart still uses NLCD colors:

screen shot 2015-09-28 at 11 03 34 am

Is the extra gray space between the image and the highlight a known issue? I assumed it was caused by the missing images.

screen shot 2015-09-28 at 11 04 31 am

Connects to #840

### DIFF
--- a/src/mmw/apps/water_balance/templates/home_page/index.html
+++ b/src/mmw/apps/water_balance/templates/home_page/index.html
@@ -209,85 +209,85 @@
                 <ul id="thumbs-land" class="nav" role="tablist"> <!-- Landcover Thumbs -->
 
                     <li id="thumb-open_water">
-                        <a href="#land-open_water" aria-controls="filter-tab" role="tab" data-toggle="tab">
-                            <img src="" style="background-color: gray;" alt="Open Water" data-container="body" data-toggle="popover" data-placement="left" data-title="Open Water" data-content="Areas of open water, generally with less than 25% cover of vegetation or soil.">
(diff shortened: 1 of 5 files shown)


rajadain:
Created #865 for the extra whitespace in thumbnails.

rajadain:
Not sure how it looks on your screen, but I think we could use a similar dark text treatment for Wetlands:

image

caseypt:
Re: wetlands, will do.

caseypt:
Updated:

screen shot 2015-09-28 at 11 47 15 am

rajadain:
+1 tested. Code look great too!

mmcfarland:
+1

Issue #842, opened 2015-09-24
Overlay layers disappear when switching basemaps


arobbi7:
To replicate:
1. Turn on any overlay layer (boundary or coverage)
2. Switch to satellite basemap (overlay layers disappear)
3. Switch back to streets basemap (overlay layers still gone)

The overlay layers reappear if you switch to either satellite with roads basemap or terrain basemap. They also reappear if you turn on a different overlay layer.

PR #868, opened 2015-09-28
Fix overlay layer zIndex setting.

caseypt:
* Bring back overlay key in layer config for layers that are used
as overlays.
* Support multiple layer keys when retrieving layers from layer config.

**Testing instructions:**
- Turn on any overlay layer.
- Switch to satellite basemap. 
- Switch back to streets basemap.
- The overlay layer should still be visible.
- Try with other layers and ensure that the overlay layers are always on top of the basemap.

Connects to #842


mmcfarland:
+1 Tested.

PR #873, opened 2015-09-29
Feature/show all headers

jfrankl:
The initial, top-level header was being hidden behind the other headers. I also changed the map positioning to factor in this extra header.

image


rajadain:
+1 tested

mmcfarland:
+1, headers are where they are meant to be.

Issue #725, opened 2015-08-26
Using browser back and forward buttons can lead to broken compare view


caseypt:
To reproduce:

- Draw or select an area.
- Navigate to the compare view using the application navigation elements.
- Navigate back to the analyze view using the browser back button.
- Navigate forward to the analyze view using the browser forward button.
- You URL should indicate that you are on the compare page, but it should look something like this:

![screen shot 2015-08-26 at 11 37 23 am](https://cloud.githubusercontent.com/assets/1042475/9498265/d7ad6b00-4be6-11e5-81cf-0f553f61c41c.png)

caseypt:
In fact all of the pages are broken once this happens.

Relevant errors:

![screen shot 2015-08-26 at 11 40 06 am](https://cloud.githubusercontent.com/assets/1042475/9498333/42fadf46-4be7-11e5-8e49-588bcb54c5ed.png)

Issue #724, opened 2015-08-26
Add back arrow to allow navigation from compare to the project modeling page


caseypt:
The only way I see is by modifying the URL

rajadain:
Or clicking the "back" button in the browser. But I agree, there should be UI for this.

caseypt:
Yeah, and using the browser buttons can sometimes get the compare view into a broken state. I think I'm finally able to reproduce it consistently, so I'll open another issue.

mmcfarland:
I'm not sure if there were wireframe/mockups for this, but the assumption is that the back arrow element on the other screens would be present here.  I'm going to change the issue title to reflect adding it in.

caseypt:
I think I thought to give this two points instead of one because I thought this issue and this one ( https://github.com/WikiWatershed/model-my-watershed/issues/725) were one in the same. But feel free to ignore if that's too much.

jamesmcclain:
Okay, I might as well investigate #725 and see if I can do it, as well.

PR #874, opened 2015-09-29
Compare view back button

jamesmcclain:
This patch does two things:

   * **Allow use of back and forward buttons.**  This attempts to fix a bug that prevents navigation form the modeling page, to the analyze page, then back to the modeling page via the browser back and forward buttons.
   * **Add back button to compare view.**  (Of course, this will  not be the final place or appearance for this button.)

screenshot from 2015-09-29 14 03 28
 
Connects #724
Connects #725

**To Test**
   * Navigate to the compare view.  Notice that there is now a left-arrow at the bottom-left of the page.  Pressing it should send you back to the modeling page.  It should also work if you go directly to the compare view without going through the modeling page.
   * Navigate around the site using the browser arrow buttons.  In particular, try moving between the analyze view and the modeling page view the browser arrow buttons.

### DIFF
--- a/src/mmw/js/src/compare/controllers.js
+++ b/src/mmw/js/src/compare/controllers.js
@@ -43,6 +43,9 @@ var CompareController = {
         // Switch back to the origProject so any changes are discarded.
         App.currentProject.off();
         App.currentProject = App.origProject;
+        if (!App.map.get('areaOfInterest')) {
+            App.map.set('areaOfInterest', App.currentProject.get('area_of_interest'));
+        }
 
         App.rootView.footerRegion.empty();
     }
--- a/src/mmw/js/src/compare/templates/compareWindow.html
+++ b/src/mmw/js/src/compare/templates/compareWindow.html
@@ -1,5 +1,8 @@
 <div id="compare-scenarios-region"></div>
 <div id="compare-footer">
+    <div class="back"> <!-- Back -->
+      <a data-url="/project" class="btn btn-md btn-icon"><i class="fa fa-arrow-left"></i></a>
+    </div> <!-- End Back -->
     <h3>Navigate Scenarios</h3>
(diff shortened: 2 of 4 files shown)


jamesmcclain:
@caseypt I had originally assigned you since you created the two cards, but @rajadain requested to review.  Hope that is okay.

caseypt:
Cool with me.

jamesmcclain:
Thanks, I probably should have asked first.

caseypt:
One quick comment though: It would be helpful to provide, in the commit message a synopsis of the problem that was causing the app to break when using the browser buttons.

rajadain:
Oh just saw this thread. Sorry, thought this issue was still unassigned. I'd already started taking a look, will continue.

jamesmcclain:
I just realized that the UI back button does not currently work when you go directly to the compare view.  I will fix that and update the commit message at the same time.

rajadain:
+1 very smooth. Tried:

 * Going back and forth between Analyze and Model, Model and Compare, and all the way from Analyze to Compare
 * Tried navigating via browser buttons and UI buttons
 * Tried landing on the Compare page directly and then hitting the back UI button
 * When returning to the Model view, the last activated scenario is selected, which is great

I feel the button placement, although a little odd, is good enough for now, and uses meaningful color conventions. Sounds like we're going to get a design update pretty soon, so it's good to have the functionality in there ready to go.

rajadain:
Perhaps @caseypt could have a look anyway and provide input on whether this satisfies his original vision.

jamesmcclain:
Thanks for looking, much appreciated!

caseypt:
+1. Nice fix. Tested.

jamesmcclain:
Thanks, I'll merge into develop.

PR #879, opened 2015-09-29
Increase options for EC2 and RDS instance types

hectcastro:
Adding small through large for both.

### DIFF
--- a/deployment/cfn/utils/constants.py
+++ b/deployment/cfn/utils/constants.py
@@ -1,11 +1,15 @@
 EC2_INSTANCE_TYPES = [
     't2.micro',
     't2.small',
-    't2.medium'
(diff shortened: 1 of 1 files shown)


deployment/cfn/utils/constants.py
```diff
@@ -1,11 +1,15 @@
 EC2_INSTANCE_TYPES = [
     't2.micro',
     't2.small',
-    't2.medium'
+    't2.medium',
+    't2.large'
```
caseypt: Is the purpose of this list to expand the available options?

hectcastro: Yes.

caseypt:
+1. Looks good to me.

Issue #843, opened 2015-09-24
Layers pane partially hidden in geospatial analysis view


arobbi7:
The bottom half of the layers pane is hidden by the geospatial analysis pane in Analysis view.

PR #881, opened 2015-09-29
Adjust layer control so that it is never obscured.

caseypt:
Previously, the element that contained the layer list had a parent with
a very small size and a very deep placement within the map container
hierarchy. To better size and position the list, we make it a direct
sibling of the map container, which is the controlling factor for the
size of the list.

* Refactor layer control code to create two separate elements, one for
the control button and one with the layer list.
* Adjust style of layer list so that it grows and shrinks along with the
map height, is positioned near the top of the map so it has more
vertical space, and so that any hidden layers are accessible by a
scroll bar.
* Adjust style of draw tools container so that it does not overlap the
layer list on small screens.

**Testing Instructions:**
- Start the app from the home page.
- Open the layer control.
- Navigate through the app, and verify that the layer control list is never hidden and changes size based on the map container size.

Thanks to @jfrankl for help figuring this out.

screen shot 2015-09-29 at 2 58 14 pm

screen shot 2015-09-29 at 2 58 31 pm

screen shot 2015-09-29 at 2 58 37 pm

Connects to #843


mmcfarland:
I'm getting controls breaking a line closer than they were previously, on large screens:

screenshot from 2015-09-29 18 11 12

Does this seem related?

mmcfarland:
Layer control behaves very nicely now.

caseypt:
Yeah, definitely related. The container for those elements overlaps the layer control without some modification. Let me see if I can improve it.

jamesmcclain:
Tests passed for me and the layer control is (almost) always on top.  If you make the screen absurdly tiny, the control appears under the controls at the top left and also under part of the circle that surrounds the right arrow button on the analyze page, but those issues seem minor and unavoidable.

jamesmcclain:
+1 from me

caseypt:
@mmcfarland Updated. Made the rule responsive.

caseypt:
Sorry, that was poorly worded. I placed the rule that was causing the two line problem in a media query so it only takes effect when the map elements need more room.

caseypt:
Just added a quick fix to make the draw tools look better when they collapse on small screens (or embeds).

screen shot 2015-09-29 at 6 24 56 pm

mmcfarland:
+1

PR #887, opened 2015-09-30
Re-enable NLCD overlay layer

rajadain:
We need to add this flag as of #868 

because NLCD Overlays are not available without them:

image

but are with them:

image


mmcfarland:
+1

caseypt:
Sorry!

Issue #888, opened 2015-09-30
Micro Site is Broken


rajadain:
![image](https://cloud.githubusercontent.com/assets/1430060/10196081/ffa44974-675e-11e5-96b3-dec20fd69e75.png)

Likely due to template changes in #873.

PR #894, opened 2015-09-30
Adjust Micro site to new Base template

rajadain:
Fixed:

image

Connects #888


mmcfarland:
+1, nice resolution

rajadain:
Merged, ready for redeploy.

Talking to @lliss, we think we should consider refactoring the template inheritance model to be less coupled, which will help prevent this in the future.

Issue #820, opened 2015-09-21
Implement My Projects page


mmcfarland:
A simple listing of the current user's saved projects.  From the design below, do not include Search or Tags.  Allow navigation to this page from the Login/User dropdown (replace "Profile") and from the Project dropdown .

![screenshot from 2015-09-21 12 45 52](https://cloud.githubusercontent.com/assets/1014341/9998390/d0e97854-605e-11e5-9577-010c9f45d1a3.png)

mmcfarland:
Consider incorporating an indication of the public/private status of the project.   Maybe a [locked](http://fortawesome.github.io/Font-Awesome/icon/lock/)/[unlock](http://fortawesome.github.io/Font-Awesome/icon/unlock/) icon?

rajadain:
Should the "My Projects" page navigation be hidden when in ITSI embed mode, so that users can't switch to a different project?

mmcfarland:
Good question.  I think so, by switching to another project, they'd probably end up associating it with the assignment.

PR #896, opened 2015-10-01
My Projects pt 1: Basic Structure and Routing

rajadain:
## Introduction

To help evaluate and manage this work, I'm splitting it into smaller PRs. The full work can be followed here: https://github.com/rajadain/model-my-watershed/tree/feature/my-projects

## Overview

In this PR a basic outline for the Projects view is added, with some basic routing. The idea is talk about the general approach, independent of design details, which will be present in a forthcoming PR.

## Testing Instructions

Checkout the code and go to [`/projects/`](http://localhost:8000/projects/) and make sure you see a "Projects" heading, like this:

image

## Notes

Input on design decisions is welcome. Questions we might explore include:

 * What is the best way of having a full-page layout when our fundamental layout is split in three parts: a `subHeaderRegion`, a `mapRegion`, and a `footerRegion`? Currently I put everything in `footerRegion`, just like in Compare view.
 * Although the map is obscured by a full height `footerRegion` as here and in the Compare view, it still exists in the background. This is not a major problem, since the background is not viewable, and this saves us from having to reinitialize the map (which is a heavy operation) more than once. Is there any value in hiding the map (which adds a `display: none` to the entire map div): `App.getMapView().$el.hide();`? Currently I'm not hiding the map, just like in Compare view.
 * Do the routes make sense? They are now more forgiving (`/project/` now redirects to `/projects/` instead of 404, `/projects/` redirects to `/` if user is not logged in instead of 403), and also follow the current scheme.

Connects #820

### DIFF
--- a/src/mmw/apps/home/tests.py
+++ b/src/mmw/apps/home/tests.py
@@ -3,11 +3,22 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import division
 
+import re
+
 from django.contrib.auth.models import User
 
 from django.test import TestCase
 
 from rest_framework.test import APIClient
+
+
+def getRoute(url):
+    """
+    Gets relative route from full URL
+    :param url: Full URL, usually like `http://testserver/projects/`
+    :return: Relative path, like `/projects/`
+    """
+    return re.search(r'http://.*?(/.*)', url).group(1).strip()
 
 
 class RouteAccessTestCase(TestCase):
@@ -36,16 +47,31 @@ class RouteAccessTestCase(TestCase):
 
         self.c.login(username='test', password='test')
 
-    def test_any_user_cannot_get_model_route_without_project_id(self):
+    def test_logged_in_user_can_get_project_listing(self):
+        response = self.c.get('/projects/')
+
+        self.assertEqual(response.status_code, 200)
+
+    def test_anon_user_getting_listing_redirects_to_home(self):
+        self.c.logout()
+
+        response = self.c.get('/projects/')
+
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(getRoute(response.url), '/')
+
+    def test_getting_model_route_without_project_id_redirects_to_listing(self):
         response = self.c.get('/project/')
 
-        self.assertEqual(response.status_code, 404)
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(getRoute(response.url), '/projects/')
 
         self.c.logout()
 
         response = self.c.get('/project/')
 
-        self.assertEqual(response.status_code, 404)
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(getRoute(response.url), '/projects/')
 
     def test_project_owner_can_get_model_route_with_project_id(self):
         project_id = self.create_private_project()
--- a/src/mmw/apps/home/urls.py
+++ b/src/mmw/apps/home/urls.py
@@ -4,12 +4,13 @@ from __future__ import unicode_literals
 from __future__ import division
 
 from django.conf.urls import patterns, url
-from apps.home.views import home_page, project, project_clone
+from apps.home.views import home_page, projects, project, project_clone
 
 
 urlpatterns = patterns(
     '',
     url(r'^$', home_page, name='home_page'),
+    url(r'^projects/$', projects, name='projects'),
     url(r'^project/$', project, name='project'),
     url(r'^project/(?P<proj_id>[0-9]+)/$', project, name='project'),
     url(r'^project/(?P<proj_id>[0-9]+)/clone/?$',
--- a/src/mmw/apps/home/views.py
+++ b/src/mmw/apps/home/views.py
@@ -19,6 +19,13 @@ def home_page(request):
     return render_to_response('home/home.html', get_context(request))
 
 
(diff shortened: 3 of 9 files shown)


caseypt:
Might want to get @jfrankl's input on the questions you outlined.

caseypt:
I'm not sure of the best way to do the layout, but sticking everything in the footer seems odd. I didn't realize we were doing that on the compare page. 

Ideally, we wouldn't initiate the map at all if the only page that was accessed was the "My Projects" page, though that may not be possible with our current layout.

We may want to consider loading the my projects page on a different template all together, and handling all the routing on the server-side. I'm not sure there's an advantage to having it as part of the existing "single page" app.

Curious for others thoughts on this.

caseypt:
I did test the current code and it works well. This is a good foundation if we stick with keeping this as part of the current template.

rajadain:
Well, the advantage of keeping everything in the "single page" app is that navigation is much quicker, once the app is all loaded. So if a person decides to go into the project list and open a bunch of projects to find the right one (because they didn't rename them and went with the same name, and there are no thumbnails in the My Projects screen), that would be **very** expensive, since we'd load the entire app again and again. It would also allow the user to open projects in new windows, which would lead to multiple client sessions of the same user, which we haven't really thought about in regards to autosaving and keeping a consistent state. So to avoid all that, I do think that keeping it all in a "single page" app is a good idea.

The map in the background isn't _really_ hurting anything, except in some edge case of taking ITSI screenshots with Firefox where it would be visible, but I haven't been able to consistently reproduce that. Now, if one were to land on the My Projects view directly, that may cause a certain amount of extraneous wait while the map (never to be seen) renders in the background possibly preceding the visible contents on the screen. But I think that the majority of visits to this page would be from UI navigation within the app, when the map would have already been rendered, and thus should be pretty snappy.

Our general layout has four real content areas: the top nav (which is fixed in height and always has the same thing, namely the title and the user menu), the sub header (which is used as a toolbar in the Model view), the map region, and the footer area (which is bit of a misnomer, since it is used more as a content area than anything, housing the results in Analyze and the model runs in Model view). Since the "footer" is already hosting content in other views, it seemed like a natural progression to use it to host content in full screen views, such as Compare and My Projects. It sounds odd at first, but once you think about it within the confines of our visual layout it is the best place to put it.

image

jfrankl:
I don't have a strong opinion here. Loading it in the footer seems less
than ideal in terms of semantics. We might want to rename the section that
is currently map-related to something more general.

Is there a way to use a different template for these kind of fullscreen,
no-map pages and also have them load quickly? Using a different template
seems like it could be cleaner and could prevent some potential issues.
These pages work in a really different way—the map one has lots of absolute
positioning and the projects page flows like a normal website.

Jeff

On Mon, Oct 5, 2015 at 12:10 PM, Terence Tuhinanshu <
notifications@github.com> wrote:

> Well, the advantage of keeping everything in the "single page" app is that
> navigation is much quicker, once the app is all loaded. So if a person
> decides to go into the project list and open a bunch of projects to find
> the right one (because they didn't rename them and went with the same name,
> and there are no thumbnails in the My Projects screen), that would be
> *very* expensive, since we'd load the entire app again and again. It
> would also allow the user to open projects in new windows, which would lead
> to multiple client sessions of the same user, which we haven't really
> thought about in regards to autosaving and keeping a consistent state. So
> to avoid all that, I do think that keeping it all in a "single page" app is
> a good idea.
>
> The map in the background isn't *really* hurting anything, except in some
> edge case of taking ITSI screenshots with Firefox where it would be
> visible, but I haven't been able to consistently reproduce that. Now, if
> one were to land on the My Projects view directly, that may cause a certain
> amount of extraneous wait while the map (never to be seen) renders in the
> background possibly preceding the visible contents on the screen. But I
> think that the majority of visits to this page would be from UI navigation
> within the app, when the map would have already been rendered, and thus
> should be pretty snappy.
>
> Our general layout has four real content areas: the top nav (which is
> fixed in height and always has the same thing, namely the title and the
> user menu), the sub header (which is used as a toolbar in the Model view),
> the map region, and the footer area (which is bit of a misnomer, since it
> is used more as a content area than anything, housing the results in
> Analyze and the model runs in Model view). Since the "footer" is already
> hosting content in other views, it seemed like a natural progression to use
> it to host content in full screen views, such as Compare and My Projects.
> It sounds odd at first, but once you think about it within the confines of
> our visual layout it is the best place to put it.
>
> [image: image]
> <https://cloud.githubusercontent.com/assets/1430060/10286107/16c73f04-6b5a-11e5-9451-efefff1ade98.png>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/WikiWatershed/model-my-watershed/pull/896#issuecomment-145584799>
> .
>



-- 
Jeff Frankl, UI/UX Designer

Azavea  |  340 N 12th St, Suite 402, Philadelphia, PA
Web azavea.com  |  Blog azavea.com/blogs  |  Twitter @azavea
<http://twitter.com>

caseypt:
Hopefully if we get to implement the sidebar layout, we can fix up the layout to be more semantically correct, as well as not load the map until we need to.

+1. Seems like everyone is okay with load this page dynamically.

rajadain:
Thanks. Closing in favor of #905

PR #901, opened 2015-10-02
Move to mmw-geoprocessing version 0.2.0

jamesmcclain:
This change enables the new geoprocessing code.

**To Test**
   * Rebuild or reprovision your `worker` VM (I rebuilt, but reprovisioning is probably sufficient).
   * Use the site and see if it still works.

Tagging @mmcfarland , as well.


mmcfarland:
+1.  Correct jar is loaded and verified that the feature in `0.2.0` are working in the app.

Issue #885, opened 2015-09-30
Remove backend geoprocessing source from this project


hectcastro:
I believe that everything in `src/geop` is an old version of `WikiWatershed/mmw-geoprocessing`.

jamesmcclain:
I am for removing this, as well (FWIW).

PR #902, opened 2015-10-02
Remove old geop code

jamesmcclain:
Connects #885 

This simply removes the old `geop` code.


mmcfarland:
+1, App functions as expected without this code.

PR #903, opened 2015-10-02
My Projects pt 2: Markup Mockup

rajadain:
## Overview

This builds on top of #896, and adds an HTML mockup in 53d558f. This mockup does not have tagging, community projects, or a "my profile" link, or pagination support, as in the original visual mockup here: https://cloud.githubusercontent.com/assets/1014341/9998390/d0e97854-605e-11e5-9577-010c9f45d1a3.png

It does have a search bar, a name and date (created? last modified?) for each project, an icon indicating public or private, buttons which are the same as in the project dropdown menu, a launch button, and a create new project button.

image

I went with text instead of icons for buttons because we have the room and that is clearer. I also switched the color palette, so we now have white rows on a gray background instead of vice versa, which provides greater contrast for the text.

Existing SASS variables have been used wherever possible.

Connects #820

### DIFF
--- a/src/mmw/apps/home/tests.py
+++ b/src/mmw/apps/home/tests.py
@@ -3,11 +3,22 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import division
 
+import re
+
 from django.contrib.auth.models import User
 
 from django.test import TestCase
 
 from rest_framework.test import APIClient
+
+
+def getRoute(url):
+    """
+    Gets relative route from full URL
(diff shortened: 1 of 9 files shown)


rajadain:
Added AoI name to project listing, which can potentially help identification, but only for projects that use a boundary layer:

image

I wonder if us including this here will inadvertently push users to prefer Boundary Layers over custom drawing.

rajadain:
Closing this in favor of #905. We can use that PR for visual / design ideas as well as implementation details.

PR #905, opened 2015-10-05
My Projects pt 3: Basic Structure and Routing

rajadain:
## Overview

Builds on top of #903 and translates the mock-up into a basic set of views, to look like this:

image

Summary of changes:

 * 44fd081 Switch to a light-weight project serializer which gives us only the information we need for the "My Projects" screen
 * c1fd1fc Has the actual view breakdown, and does the following:
   - Adds a `ProjectCollection`
   - Decomposes mockup into separate templates and adjusts styling
   - Adds a new `toFriendlyDate` Nunjucks filter for date format assist
   - The projects are populated and show the `name`, `area_of_interest_name`, `created_at` fields, the lock / globe icon is shown based on `is_private`, and the "Launch Project" button points to the correct project URL based on the project's `id`
 * d5d92c1 Adds navigation to the "My Projects" page from the user dropdown in the header and the project menu for logged in users.

Still remaining:

 * Search functionality
 * "Create New Project" button
 * Rename, Share, Make Public / Private, Delete functionality

Known bugs:

 * User dropdown has stopped working in Model view

Connects #820

Issue #883, opened 2015-09-29
Resetting AoI after Analyze doesn't resize map


mmcfarland:
When the AoI bar exists on the Draw page and you reset the tools, the bar is removed but the map is not resized to fill the gap.

![screenshot from 2015-09-29 19 49 18](https://cloud.githubusercontent.com/assets/1014341/10181216/4e57f092-66e3-11e5-92af-8e8b6ded3dcf.png)

PR #908, opened 2015-10-06
My Projects pt 4: Release Candidate 1

rajadain:
## Overview

Building on top of #905, this branch adds functionality to all user facing operations on the page, except for Search which is disabled as per instructions in #820.

image

### Summary of Changes

 * eb0c5f5 and df12c4d are different from #905 because of whitespace changes.
```diff
$ git diff c1fd1fc61ec9e2c08c526f6356130b1d0800ec82..eb0c5f5e7e749458ae0305b6508102787968ebc9
diff --git a/src/mmw/js/src/projects/views.js b/src/mmw/js/src/projects/views.js
index b10f595..3a67d77 100644
--- a/src/mmw/js/src/projects/views.js
+++ b/src/mmw/js/src/projects/views.js
@@ -18,9 +18,9 @@ var ProjectsView = Marionette.LayoutView.extend({
         projects
             .fetch()
             .done(function() {
-               self.rowContainer.show(new ProjectRowsView({
+                self.rowContainer.show(new ProjectRowsView({
                     collection: projects
-               }));
+                }));
             });
     }
 });
```
 * fb81e55 addresses #883, by ensuring that when we come back to the Draw view the map is positioned and cleared properly.
 * b3ceaa0 Clicking "Create New Project" clears any current project and jumps to Draw view. Also disables Search, which is now non-clickable. If we want it completely gone, I can edit to make it so. If real users are going to be using this page, then we should not have the search bar there unless they can use it.
 * bace65f Adds `PATCH` saving for projects, both on client and server sides. Since in the My Projects view we use the new `ProjectListingSerializer` which only sends a subset of information, we must evolve the backend to also save a project with only a subset of the information, instead of clearing out the missing information.

The final four commits add rename, share, privacy changing, and deleting capabilities to the page. The first two required subtle changes to generic renaming and sharing to allow working with the partial data set available in the Project listing. See commit messages for detailed explanations.

### Known Issues:

 * User dropdown is not visible in Modeling view #912 
 * User dropdown is clipped in Draw view #910 

Connects #820 
Connects #883

### DIFF
--- a/src/mmw/js/src/core/modals/views.js
+++ b/src/mmw/js/src/core/modals/views.js
@@ -152,10 +152,10 @@ var ShareView = ModalBaseView.extend({
 
             confirm.render();
             confirm.on('confirmation', function() {
-                var project = self.options.app.currentProject;
+                var project = self.options.project || self.options.app.currentProject;
 
                 project.set('is_private', false);
-                project.saveProjectAndScenarios();
+                project.saveProjectListing();
 
                 self.$el.modal('show');
             });
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -278,6 +278,12 @@ var ProjectModel = Backbone.Model.extend({
     }
 });
 
+var ProjectCollection = Backbone.Collection.extend({
+    url: '/api/modeling/projects/',
+
+    model: ProjectModel
+});
+
 /**
  * A predicate for a filter function used in the _modifyModifications.
  * Returns true if piece has a valid shape with non-zero area, and
@@ -809,6 +815,7 @@ module.exports = {
     ModelPackageControlModel: ModelPackageControlModel,
     Tr55TaskModel: Tr55TaskModel,
     ProjectModel: ProjectModel,
+    ProjectCollection: ProjectCollection,
     ModificationModel: ModificationModel,
     ModificationsCollection: ModificationsCollection,
     ScenarioModel: ScenarioModel,
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -177,6 +177,11 @@ var ProjectModel = Backbone.Model.extend({
             ],
             attrs = _.pick(this.toJSON(), listingAttrs);
 
+        // Server expects user to be id, not object
+        if (attrs.user.id) {
+            attrs.user = attrs.user.id;
+        }
+
         this.save(attrs, { patch: true });
     },
 
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -171,6 +171,16 @@ var ProjectModel = Backbone.Model.extend({
     // Flag to prevent double POSTing of a project.
     saveCalled: false,
 
+    saveProjectListing: function() {
+        var listingAttrs = [
+                'id', 'name', 'area_of_interest_name', 'is_private',
+                'model_package', 'created_at', 'modified_at', 'user'
+            ],
+            attrs = _.pick(this.toJSON(), listingAttrs);
+
+        this.save(attrs, { patch: true });
+    },
+
     saveProjectAndScenarios: function() {
         if (!this.get('allow_save') || !App.user.loggedInUserMatch(this.get('user_id'))) {
             // Fail fast if the user can't save the project.
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -120,7 +120,6 @@ var ProjectModel = Backbone.Model.extend({
         this.set('is_activity', settings.get('activityMode'));
 
         this.listenTo(this.get('scenarios'), 'add', this.addIdsToScenarios, this);
-        this.on('change:name', this.saveProjectAndScenarios, this);
     },
 
     createTaskModel: function() {
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -133,6 +133,7 @@ var ProjectMenuView = Marionette.ItemView.extend({
 
         rename.on('update', function(val) {
             self.model.updateName(val);
+            self.model.saveProjectAndScenarios();
         });
     },
 
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -4,6 +4,7 @@
     $ = require('jquery'),
     Marionette = require('../../shim/backbone.marionette'),
     App = require('../app'),
+    settings = require('../core/settings'),
     models = require('./models'),
     controls = require('./controls'),
     modalModels = require('../core/modals/models'),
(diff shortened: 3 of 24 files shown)


caseypt:
I think we should remove the search box until we have time to implement it.

src/mmw/js/src/modeling/models.js
```diff
@@ -171,6 +170,21 @@ var ProjectModel = Backbone.Model.extend({
     // Flag to prevent double POSTing of a project.
     saveCalled: false,
 
+    saveProjectListing: function() {
+        var listingAttrs = [
+            'id', 'name', 'area_of_interest_name', 'is_private',
+            'model_package', 'created_at', 'modified_at', 'user'
```
caseypt: Consider indenting the two lines of items.

src/mmw/js/src/projects/views.js
```diff
@@ -0,0 +1,160 @@
...
+        'click @ui.newProject': 'newProject'
+    },
+
+    regions: {
+        rowContainer: '.project-rows-container'
+    },
+
+    onShow: function() {
+        var self = this,
+            projects = new models.ProjectCollection();
```
caseypt: Consider adding a line break here.

caseypt:
This is an existing issue, but more apparent with these changes: https://github.com/WikiWatershed/model-my-watershed/issues/910.

caseypt:
This is out of scope for this card, but it would be neat to have an image of the AoI for each project.

rajadain:
> it would be neat to have an image of the AoI for each project.

I agree! It would make identifying them very easy, because I imagine not many people would take the time to name the projects meaningfully. I remember at some point we talked about a server-side screenshot solution, but now with the ITSI screenshot mechanism in place, we could potentially use that to generate images when creating new projects and saving on the server side.

src/mmw/js/src/modeling/models.js
```diff
@@ -278,6 +292,12 @@ var ProjectModel = Backbone.Model.extend({
     }
 });
 
+var ProjectCollection = Backbone.Collection.extend({
+    url: '/api/modeling/projects/',
+    model: ProjectModel,
+    comparator: 'created_at'
```
caseypt: Not sure what's most desirable, but consider ordering this in descending order.

caseypt:
+1. This is really well done. I feel like I've exercised just about every way to enter and use the projects page, and I haven't encountered any new issues.

One thing this highlights is that we may need a way to skip animations when changing pages. Things look a little chunky when going from the analyze page to my projects, then selecting a project. That's more of a polishing step, however and is out of scope for this card.

caseypt:
I just noticed that Known Issues section of your PR. I created issues for those things.

lliss:
I'm unable to use the dropdown to get to the projects page from the modeling page.

rajadain:
@lliss The User dropdown not working from the Modeling view is a known issue: #912. Can you try the Project dropdown instead?

lliss:
I just noticed that the modals have no protection against double clicking or hitting enter twice.

src/mmw/js/src/modeling/models.js
lliss: I agree. It feels natural to see the things I just created at the top.

lliss:
I tried all the features and everything worked as expected. I'll give the code a closer look in the morning when I am fresh.

src/mmw/js/src/projects/views.js
```diff
@@ -0,0 +1,160 @@
...
+            .fetch()
+            .done(function() {
+                self.rowContainer.show(new ProjectRowsView({
+                    collection: projects
+                }));
+            });
+    },
+
+    newProject: function() {
+        App.currentProject = null;
```
lliss: I could be wrong, but I seem to recall this happening in a few places now. We might want to make this a method of the App.

src/mmw/js/src/modeling/views.js
```diff
@@ -131,6 +133,7 @@ var ProjectMenuView = Marionette.ItemView.extend({
 
         rename.on('update', function(val) {
             self.model.updateName(val);
+            self.model.saveProjectAndScenarios();
```
lliss: Removing the listener on the model and moving the call here might introduce problematic edge cases. I haven't seen any but its just something to be cautious about.

src/mmw/js/src/core/modals/views.js
```diff
@@ -152,10 +152,10 @@ var ShareView = ModalBaseView.extend({
 
             confirm.render();
             confirm.on('confirmation', function() {
-                var project = self.options.app.currentProject;
+                var project = self.options.project || self.options.app.currentProject;
 
                 project.set('is_private', false);
-                project.saveProjectAndScenarios();
```
lliss: If you share your project immediately after creating it now will it save the scenarios?

lliss:
+1. Nicely done. Tested.

rajadain:
Added commits that remove the search box, add whitespace, and sort newest projects to the top:

image

src/mmw/js/src/projects/views.js
rajadain: These are all the places where we set `currentProject` to `null`:

 * [`AnalyzeController.analyzePrepare`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/analyze/controllers.js#L46)
 * [`core/testUtils:resetApp`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/testUtils.js#L28)
 * [`ItsiEmbed.loadInteractive`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/itsiEmbed.js#L48)
 * [`ModelingController.projectDraw`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/controllers.js#L202)

While we are clearing the `App.currentProject` (and notably not clearing `App.projectNumber` which I believe was a later addition), we're not always clearing the AoI, and I don't think we want to in all these cases. I agree that clearing the project and AoI should be a function on `App`, but I'm not sure if it will be able to replace all these occurrences.

src/mmw/js/src/modeling/views.js
rajadain: Here are all the places where we are setting the name:

 * [`compare/controllers:copyProject`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/compare/controllers.js#L96): but it doesn't count since we also set `allow_save` to `false`, so we don't intend to save.
 * [`draw/controllers:enableSingleProjectModeIfActivity`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/draw/controllers.js#L69): but we're not so much setting the name as creating a new project, which would get saved when we enter the Analyze view as of #687 
 * [`modeling/controllers:makeProject`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/controllers.js#L264): same as above, where we will save if any changes are made to the project or scenarios after it has been made.

Thanks for pointing this out, glad I checked, but I think we're okay.

rajadain:
Thanks for reviewing, all! Merging this in.

Issue #912, opened 2015-10-06
User menu is not clickable on modeling page.


caseypt:
Clicking on the username in the header has no effect on the modeling page. The menu opens properly on all other pages.

Issue #910, opened 2015-10-06
User menu is cut-off


caseypt:
<img width="219" alt="screen shot 2015-10-06 at 5 01 39 pm" src="https://cloud.githubusercontent.com/assets/1042475/10322490/ecc0fafe-6c4b-11e5-997f-be5523472618.png">

PR #919, opened 2015-10-07
Adjust CSS related to header dropdown.

lliss:
Connects #910 
Connects #912 

Because of the absolute positioning, our dropdown for the user menu in the top
left was being pushed off the screen. Likewise, because the secondary menu bar
had an unnecessary z-index it was sitting on top of the open dropdown.

### To Test
 * Log in
 * Use user dropdown in top right
 * Ensure text is readable
 * Go to modeling page
 * Use user dropdown in top right
 * Ensure text is readable

### DIFF
--- a/src/mmw/sass/base/_header.scss
+++ b/src/mmw/sass/base/_header.scss
@@ -50,6 +50,10 @@ header {
           right: 0px;
           a {
             color: $ui-secondary;
+
+            &:hover, &:active {
+              color: $ui-secondary;
+            }
           }
         }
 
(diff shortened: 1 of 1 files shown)


rajadain:
+1 much better! Also rebased on top of #908 to ensure that it worked, and it does:

image

The only thing I'd ask for is that we use dark text on the highlighted elements, as the white on light grey is nearly impossible to read.

lliss:
Adjusted text color in 67ca47c

rajadain:
+1 much better!

image

Issue #921, opened 2015-10-07
Removing BMP/modification causes graphs to vanish


jamesmcclain:
On the modeling page, removing a BMP or reclassification causes the graphs to disappear:

No reclassification initially:
![screenshot from 2015-10-07 17 10 54](https://cloud.githubusercontent.com/assets/11281373/10351414/6f5c939e-6d17-11e5-82c4-b6a0538a8abb.png)

Reclassification added:
![screenshot from 2015-10-07 17 11 21](https://cloud.githubusercontent.com/assets/11281373/10351429/843d65f4-6d17-11e5-8646-14819c401412.png)

Reclassification removed:
![screenshot from 2015-10-07 17 11 29](https://cloud.githubusercontent.com/assets/11281373/10351438/8b8a312a-6d17-11e5-8a9e-62beef50393c.png)

jamesmcclain:
Update: this is happening on my local setup but not on mmw.azavea.com, which should be identical at this moment, so the problem might be mine.

caseypt:
I got this locally and on mmw.azavea.com. I think it's a problem with the caching.

Here's the stack trace from Celery:

```
[2015-10-08 11:09:03,619: ERROR/MainProcess] Task apps.modeling.tasks.run_tr55[bb18452e-1804-495e-9978-640f470a178f] raised unexpected: Exception(u'Missing pieces or censuses for modifications',)
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/opt/app/apps/modeling/tasks.py", line 185, in run_tr55
    raise Exception('Missing pieces or censuses for modifications')
Exception: Missing pieces or censuses for modifications
[2015-10-08 11:09:03,627: ERROR/Worker-1] Task bb18452e-1804-495e-9978-640f470a178f run from job 1339 raised exception: Missing pieces or censuses for modifications
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
[...]

PR #923, opened 2015-10-08
Fix model runs with no modifications.

caseypt:
* Model runs with no modifications that previously had modifications
were throwing exceptions, but this case should be supported. If there
are pieces but no censuses, we should throw an exception because the
model results will be incorrect. However, if there are no pieces
but a census, we can just delete the census and run the model.

**Testing instructions:**
- Create a new project.
- Add a modification to the "New Scenario". The model should run correctly.
- Remove the modification. The bar chart results should still appear, and the modified column should match the current conditions column.

Connects to #921

### DIFF
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -180,9 +180,10 @@ def run_tr55(censuses, model_input, cached_aoi_census=None):
     aoi_census = cached_aoi_census if cached_aoi_census else censuses[0]
 
     modifications = []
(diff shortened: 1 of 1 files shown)


mmcfarland:
+1 Logical change, not tested.

rajadain:
Testing now

rajadain:
+1 tested. Had to restart Celery for the new code to kick in, but it worked as advertised afterwards. Nice job.

caseypt:
Thanks @rajadain!

Issue #917, opened 2015-10-07
Address selections show below draw tools


mmcfarland:
On less-wide screens when the draw button row breaks into two rows, the address suggestions show below the other tools.

![screenshot 2015-10-07 at 11 19 53 am](https://cloud.githubusercontent.com/assets/1014341/10342226/1e423e22-6ce6-11e5-917d-e5bce725b541.png)

PR #925, opened 2015-10-08
Ensure that search results are on top

rajadain:
## Overview

Previously they would be obscured by Draw Tools on small screens. This ensures that when they are visible they are on top of all other items.

## Demo

See here: https://1cfc461b.ngrok.io

image

Connects #917


mmcfarland:
+1 Looks good, suggestions show above all others.

rajadain:
Thank you!

Issue #886, opened 2015-09-30
Celery retires go to on another machine; no SJS affinity


hectcastro:
Celery workers and Spark Job Server instances share the same virtual machine. In a scenario where there are two worker machines, a retry of a Celery task kick-off on machine `A` may get executed on machine `B`. The SJS instance on machine `B` doesn't know anything about the job running inside SJS on machine `A`.

Simplest potential solution I can come up with right now is to bind workers to queues using their hostnames, then ensure that retries re-enqueue into queues named after their hostname.

PR #927, opened 2015-10-08
Celery Worker Affinity

jamesmcclain:
The parts of the the job chain which must remain on the same worker, namely the top- and bottom-halves of the SJS submission/polling functionality, are forced to remain on the same worker.  This is done so that the polling task is not scheduled on another worker which does not have access to the spark-jobserver on which the geoprocessing query has been submitted.

Connects #886

It might be helpful to install the `vagrant-cachier` plugin, if you have not done so already.

**Test with two workers**
   * Halt the `worker` VM.
   * Pull down the following branch https://github.com/jamesmcclain/model-my-watershed/tree/affinity-testing, and switch to it.  This branch is identical to the pull request branch except for the last commit which adds workers.  This last commit should not be merged, and I also don't want to put it into the P/R branch for fear of messing up Jenkins.
   * Type `vagrant up worker1 worker2` to create the two VMs.
   * Reload the `app` VM.
   * Use the site as normal.  Hopefully everything works, and you will be able to verify that jobs are being executed on both VMs by looking at the Celery logs (`/var/log/celery/Black-worker[01].log` on both `worker1` and `worker2` [that is the purpose of SSHing into the VMs]).
   * Run tests: `./scripts/test.sh`.
   * After testing, halt or destroy the two worker VMs.

**Test with one worker**
   * Now switch to the P/R branch https://github.com/jamesmcclain/model-my-watershed/tree/feature/celery-worker-affinity
   * Reprovision the worker VM: `vagrant provision worker`
   * Reload the `app` VM.
   * Conform that the site works as normal with one worker.
   * Run tests: `./scripts/test.sh`.

### DIFF
--- a/src/mmw/apps/modeling/views.py
+++ b/src/mmw/apps/modeling/views.py
@@ -198,19 +198,24 @@ def get_job(request, job_uuid, format=None):
 
 
 def choose_worker():
-    workers = celery.current_app.control.inspect().ping().keys()
-    index = random.randint(0, len(workers) - 1)
-    return workers[index]
+    workers = filter(lambda worker: settings.STACK_COLOR in worker,
+                     celery.current_app.control.inspect().ping().keys())
+    return random.choice(workers)
 
 
 def _initiate_analyze_job_chain(area_of_interest, job_id, testing=False):
-    return chain(tasks.start_histogram_job.s(area_of_interest),
-                 tasks.get_histogram_job_results.s(),
-                 tasks.histogram_to_survey.s(),
-                 save_job_result.s(job_id, area_of_interest)) \
-        .apply_async(link_error=save_job_error.s(job_id),
-                     exchange='C.dq',
-                     routing_key=choose_worker())
+    exchange = 'C.dq'
+    routing_key = choose_worker()
+
+    return chain(tasks.start_histogram_job.s(area_of_interest).set(
+                    exchange=exchange, routing_key=routing_key),
+                 tasks.get_histogram_job_results.s().set(
+                     exchange=exchange, routing_key=routing_key),
+                 tasks.histogram_to_survey.s().set(
+                     exchange=exchange, routing_key=routing_key),
+                 save_job_result.s(job_id, area_of_interest).set(
+                     exchange=exchange, routing_key=routing_key)) \
+        .apply_async(link_error=save_job_error.s(job_id))
 
 
 @decorators.api_view(['POST'])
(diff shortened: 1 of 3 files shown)


src/mmw/apps/modeling/views.py
```diff
@@ -229,6 +242,9 @@ def _initiate_tr55_job_chain(model_input, job_id):
 
 
 def _construct_tr55_job_chain(model_input, job_id):
+    exchange = 'C.dq'
```
lliss: Does this have special meaning?

src/mmw/apps/modeling/views.py
```diff
@@ -195,9 +197,20 @@ def get_job(request, job_uuid, format=None):
...
+def choose_worker():
+    workers = filter(lambda worker: settings.STACK_COLOR in worker,
+                     celery.current_app.control.inspect().ping().keys())
+    return random.choice(workers)
+
+
 def _initiate_analyze_job_chain(area_of_interest, job_id, testing=False):
-    return chain(tasks.start_histogram_job.s(area_of_interest),
-                 tasks.get_histogram_job_results.s(),
+    exchange = 'C.dq'
```
lliss: This identical string is defined elsewhere. Should it be a constant?

src/mmw/apps/modeling/views.py
jamesmcclain: Yes, this is a special magic exchange name that is created by Celery when you ask it to create per-worker queues (by setting `CELERY_WORKER_DIRECT = True`).

src/mmw/apps/modeling/views.py
jamesmcclain: How does one define constants in Python?

src/mmw/apps/modeling/views.py
lliss: A comment to that effect would be good. I also think if it is something non-arbitrary and somewhat magic, it probably should be a constant.

jamesmcclain: Okay, will-do.

src/mmw/apps/modeling/views.py
lliss: https://github.com/jamesmcclain/model-my-watershed/blob/cd24b832bf46c9934eabd28931038b307387877e/src/mmw/apps/modeling/tasks.py#L18

lliss:
I followed your instructions but was not able to successfully submit jobs to the workers.

jamesmcclain:
I see.  If you have time on Tuesday maybe we can investigate a little bit.

lliss:
Ok. Tested everything out. Looks good. +1.

Issue #893, opened 2015-09-30
Drawing twice creates empty space in footer


kdeloach:
1 Draw an area

2 Click back button

![image](https://cloud.githubusercontent.com/assets/43062/10196622/cf49b8f6-6761-11e5-9f92-ed9708a70430.png)

3 Draw area

![image](https://cloud.githubusercontent.com/assets/43062/10196658/ecf53434-6761-11e5-85b1-c58ff115721c.png)

There's a big empty space where the footer used to be.

Source: mmw.azavea.com version ccdf9d75de0fa44435e37a92e64b51546864b0ae.

mmcfarland:
Duplicate of #883

kdeloach:
Not a duplicate and still observable on staging.

PR #929, opened 2015-10-08
Fill empty space when selecting Draw tools

rajadain:
## Overview

Previously, if there was a location bar left over from the Analyze view, and the user reset the AoI or selected a new draw tool, the location bar would disappear but leave an empty white space.

2e57f36 tried to fix this but made the mistake of resetting the map only after the user clicked "Reset", and not when the user simply selected a different draw tool. By moving the call to `setDrawSize` within `clearAoILayer`, we ensure that the map is always resized whenever the AoI is cleared.

We switch from `setDrawSize(true)` to `setDrawSize(false)` because there is no AoI to fit to, thus we don't attempt fitting, making the resize transition slightly smoother.

## Testing Instructions

 1. Draw a shape. Click the back arrow in the Analyze view to return to Draw view.
 2. Click "Reset". Ensure that when the black bar disappears the map is resized to occupy all space and there is no left over whitespace.
 3. Draw another shape. Click the back arrow in the Analyze view to return to Draw view.
 4. Click any draw tool. Ensure that when the black bar disappears the map is resized to occupy all space and there is no left over whitespace.

Connects #893

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -332,7 +332,6 @@ var ResetDrawView = Marionette.ItemView.extend({
         utils.cancelDrawing(App.getLeafletMap());
         clearAoiLayer();
         clearBoundaryLayer(this.model);
-        App.map.setDrawSize(true);
     }
 });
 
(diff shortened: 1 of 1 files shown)


mmcfarland:
+1, tested.

kdeloach:
+1

Issue #930, opened 2015-10-09
Draw tools popover is hidden


mmcfarland:
When hovering over the help icons on the various draw tools, the anchor of the popover is visible, but the content is clipped outside of the dropdown region, although as seen in the screenshot below, it is applied to the DOM.

![screenshot from 2015-10-09 09 17 27](https://cloud.githubusercontent.com/assets/1014341/10394683/c1c85d34-6e66-11e5-9450-9c3222cacef2.png)

PR #931, opened 2015-10-09
Fix draw popovers

rajadain:
## Overview

Since we now use the [`map`](https://github.com/WikiWatershed/model-my-watershed/blob/c392ed6c7548ef4b6cf34984606aecfda0e22a40/src/mmw/apps/core/templates/base.html#L31-L34) block rather than the [`content`](https://github.com/WikiWatershed/model-my-watershed/blob/c392ed6c7548ef4b6cf34984606aecfda0e22a40/src/mmw/apps/core/templates/base.html#L36-L39) block in [`base.html`](https://github.com/WikiWatershed/model-my-watershed/blob/c392ed6c7548ef4b6cf34984606aecfda0e22a40/src/mmw/apps/core/templates/base.html), which is wrapped in `.map-container` rather than `.container-fluid`, we have to adjust the target viewport in the given popovers.

## Testing Instructions

Hover over the help icons in entries in the Select by Boundary, Draw Area, and Delineate Watershed draw tools, and ensure that the popovers are shown correctly. Test in different browsers.

## Demo

image

Connects #930

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -105,7 +105,7 @@ var SelectAreaView = Marionette.ItemView.extend({
     onRender: function() {
         this.ui.helptextIcon.popover({
             trigger: 'hover',
-            viewport: '.container-fluid.top-nav'
+            viewport: '.map-container'
         });
     },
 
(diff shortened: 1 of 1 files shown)


caseypt:
+1. Looks good.

screen shot 2015-10-12 at 4 28 07 pm

Issue #926, opened 2015-10-08
Limit size of AoI drawn


mmcfarland:
Determine a crude approximation of likelihood of success (probably area) and don't allow the user to perform analysis, with some sort of explanatory message like "Sorry, the application doesn't support analysis of areas of this size at the moment, please try a smaller area".  

We are intending to support this but need to work out the performance implications of the volume of RDD requested for large AoIs.  

The goal here is to prevent users from seeing errors we know they'll get and to prevent jobs which will max out the queue polling time and then fail from clogging the pipeline.

Issue #645, opened 2015-07-30
Add Alert Modal


rajadain:
Currently we have a Share Modal, an Input Modal, and a Confirmation Modal. We need to add an Alert Modal which shows a message, perhaps an Icon (to differentiate Info vs Warning vs Error or something along those lines), and a dismiss button (OK or something). The user should also be able to dismiss it by clicking anywhere outside or using <kbd>Esc</kbd>.

Replace current usage of `window.alert` with this Modal.

jamesmcclain:
The code introduced in https://github.com/WikiWatershed/model-my-watershed/pull/932 should be updated to use this new modal (the `window.alert` should be replaced).

mmcfarland:
We have a number of `window.alert` usage now and I think it's time to create this and replace the usage across the app.

PR #932, opened 2015-10-09
Limit AoI Size

jamesmcclain:
The maximum size of an Area of Interest has been limited to 10,000 square km.  In my testing, this seems to be the maximum size that can be reliably operated-on.

Connects #926

**To Test**
   * Choose a small AoI.  Everything should proceed as normal.
   * Choose a very large AoI (drawn or selected).  If it is >= 10000 square kilometers, you will receive an alert informing you that it is too large, and the AoI will not be set.

I chose the value of 10000 square kilometers because that seems to be the cutoff point between reliable and unreliable in my testing.  I am open to suggestions on this value.

screenshot from 2015-10-09 14 50 32

screenshot from 2015-10-09 14 51 00

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -4,13 +4,15 @@ var $ = require('jquery'),
     _ = require('lodash'),
     L = require('leaflet'),
     Marionette = require('../../shim/backbone.marionette'),
-    turfRandom = require('turf-random'),
-    turfBuffer = require('turf-buffer'),
+    turfArea = require('turf-area'),
     turfBboxPolygon = require('turf-bbox-polygon'),
+    turfBuffer = require('turf-buffer'),
     turfDestination = require('turf-destination'),
+    turfRandom = require('turf-random'),
     router = require('../router').router,
     App = require('../app'),
     utils = require('./utils'),
+    coreUtils = require('../core/utils'),
     toolbarTmpl = require('./templates/toolbar.html'),
     loadingTmpl = require('./templates/loading.html'),
     selectTypeTmpl = require('./templates/selectType.html'),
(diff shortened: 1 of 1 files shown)


src/mmw/js/src/draw/views.js
```diff
@@ -19,6 +21,17 @@ var $ = require('jquery'),
...
 
+var maxArea = 10000; // 10,000 km^2
+
+function areaWarning(area) {
+    var message = '';
+    message += 'Sorry, your Area of Interest is too large.\n\n';
+    message += Math.floor(area) + ' square km were selected, ';
+    message += 'but the maximum supported size is currently ';
+    message += maxArea + ' square km.';
+    window.alert(message);
```
mmcfarland: I think we should use a friendlier Modal to display this message, we should have some generic approaches in the codebase.

rajadain: I was almost going to suggest bringing #645 into this sprint since we were looking for things to do

caseypt:
@jamesmcclain Does the TR-55 manual specify a maximum area that the results are valid for? If so, we may want to limit it to that.

jamesmcclain:
@caseypt : No, it more-or-less just takes a dictionary as input, operates on it, and produces a dictionary as output.  The size is unlimited, because the amount of time that it takes to run TR-55 is more proportional to the diversity of the area than its actual size (and even an area with every possible pair would not take long).

src/mmw/js/src/draw/views.js
jamesmcclain: Ah, thanks for the comment, I was just grepping for modals.  Would the "Alert Modal" be a good choice?

mmcfarland:
@caseypt As I suspect you suspect, TR55 is intended for small parcel-sized areas, but I think Stroud intends for it to be run on any size area (for a reasonable definition of "any").

src/mmw/js/src/draw/views.js
rajadain: No, because the Confirmation Modal has two (configurable) buttons: OK and Cancel. We need an Alert Modal, with only one response.

jamesmcclain:
Oh, I misunderstood what you meant.  Yeah, the model is only "valid" for small areas.

src/mmw/js/src/draw/views.js
jamesmcclain: It seems that we are going to use this `window.alert` as a stop-gap until https://github.com/WikiWatershed/model-my-watershed/issues/645 is done.

rajadain:
+1 tested. Tried manually drawing a large area, and my AoI was cleared after I dismissed the alert. Tried selecting a very large HUC-8, and no AoI was selected after I dismissed the alert but the HUC-8 Boundary Layer was not cleared, which is good. Could not test with Delineation, because it draws such small AoIs, but since it is disabled in Production I suppose we don't have to pay attention to it now. The code does look like it would work the same for that case.

I think that the alert is _okay_ for now, and @jamesmcclain has added a comment on #645 to convert this modal to a styled one, so that work can be deferred.

kdeloach:
+1 Tested.

PR #933, opened 2015-10-09
Update Microsite CSV

jamesmcclain:
The microsite CSV has been regenerated with precipitation levels which are whole numbers of centimeters.  The five levels of precipitation (in centimeters) are: 1, 3, 5, 8, 21.

**To Test**
   * Update the CSV by typing `./scripts/manage.sh collectstatic`
   * Use the microsite.  All of the precipitation levels should be whole-numbers of centimeters, and everything should appear more-or-less as before.

Connects WikiWatershed/tr-55#44


rajadain:
Tested the micro site, and it looks good:

image

But the CSV file looks very odd, as do the [`precipKey` values](https://github.com/WikiWatershed/model-my-watershed/pull/933/files#diff-410ada084f8c0311b241bc1e4801ca80R17).

Instead, could we generate the CSV file in cm instead of inches, and just remove the [`convertToMetric`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/main_water_balance.js#L108-L111) [references](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/main_water_balance.js#L76-L79)?

jamesmcclain:
Okay, making those changes.

jamesmcclain:
Okay, updated.

rajadain:
Much  better! The developer / code-reader UX is now much nicer.

However, in the spirit of consistency, we should ensure that all numeric values always have one decimal place:

image

Instead of adding it in the string, like [we do here](https://github.com/WikiWatershed/model-my-watershed/pull/933/files#diff-410ada084f8c0311b241bc1e4801ca80R76), I would recommend using [`.toFixed(1)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed) for all four values.

jamesmcclain:
Good call, will correct.

rajadain:
+1 tested, works perfectly and the code reads very well! Great work!

jamesmcclain:
Many thanks!

Issue #846, opened 2015-09-24
Model Output results disappear when modification is added outside of area being analyzed 


arobbi7:
To replicate:
1. Draw/select any area for analysis
2. Proceed to Model view
3. Add land cover or conservation area modification outside of analysis area/polygon

Connects #818

arobbi7:
This is what it looks like when I try:
![screen shot 2015-09-24 at 9 46 45 am](https://cloud.githubusercontent.com/assets/12868342/10075066/c9cc85b2-62a1-11e5-9fbb-d79c6615e78c.png)

rajadain:
Currently if a modification is drawn completely outside the AoI, it has an effective area of 0, which causes the model run to fail, even if there are other non-zero modifications:

![image](https://cloud.githubusercontent.com/assets/1430060/10372588/7b4ee162-6db6-11e5-96bb-75b93e46624c.png)

Consider warning the user or simply ignoring those modifications.

PR #934, opened 2015-10-09
Fix model runs with modifications outside of AoI

caseypt:
See commit message for details.

**Testing Instructions**
- Start a new project.
- Draw a bunch of modifications outside of the AoI. Make sure you get results each time a modification is added.

This is being caused by modifications whose census includes land cover types that are not in the AoI census. I'm working on a change to TR-55 so that it throws checks for this and throws an exception, rather than getting a `KeyError` later down the line.

Connects to #846

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -318,11 +318,11 @@ function _alterModifications(rawModifications) {
         reclass = 'landcover',
         bmp = 'conservation_practice',
         both = 'both',
-        n = rawModifications.size(),
+        n = rawModifications.length,
         n2 = n * n;
 
     for (var i = 0; i < n; ++i) {
-        var rawModification = rawModifications.at(i),
+        var rawModification = rawModifications[i],
             newPiece = {
                 name: rawModification.get('name'),
                 shape: rawModification.get('effectiveShape') || rawModification.get('shape'),
@@ -631,11 +631,14 @@ var ScenarioModel = Backbone.Model.extend({
         var self = this,
             results = this.get('results'),
             taskModel = this.get('taskModel'),
+            nonZeroModifications = this.get('modifications').filter(function(mod) {
+                return mod.get('effectiveArea') > 0;
+            }),
             taskHelper = {
                 postData: {
                     model_input: JSON.stringify({
                         inputs: self.get('inputs').toJSON(),
-                        modification_pieces: alterModifications(self.get('modifications'), self.get('modification_hash')),
+                        modification_pieces: alterModifications(nonZeroModifications, self.get('modification_hash')),
                         area_of_interest: App.currentProject.get('area_of_interest'),
                         aoi_census: self.get('aoi_census'),
                         modification_censuses: self.get('modification_censuses'),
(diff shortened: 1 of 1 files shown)


caseypt:
screen shot 2015-10-09 at 4 45 24 pm

src/mmw/js/src/modeling/models.js
```diff
@@ -625,11 +631,14 @@ var ScenarioModel = Backbone.Model.extend({
         var self = this,
             results = this.get('results'),
             taskModel = this.get('taskModel'),
+            filteredModifications = this.get('modifications').filter(function(mod) {
+                return mod.get('effectiveArea') !== 0;
```
rajadain: Although extremely unlikely, I would recommend protecting against negative values by replacing with

```javascript
return mod.get('effectiveArea') > 0;
```

src/mmw/js/src/modeling/models.js
```diff
@@ -625,11 +631,14 @@ var ScenarioModel = Backbone.Model.extend({
         var self = this,
             results = this.get('results'),
             taskModel = this.get('taskModel'),
+            filteredModifications = this.get('modifications').filter(function(mod) {
```
rajadain: Consider renaming to `nonZeroModifications` or `modificationsInsideAoI` or something more descriptive of intent.

rajadain:
Tested and it works beautifully. Tried various combinations of completely outside, partially inside, completely inside, and various overlapping combinations, and they all work as they should:

image

rajadain:
+1, with the few code comments above.

src/mmw/js/src/modeling/models.js
caseypt: Good call.

caseypt:
Re: how effectiveArea was incorrect before, `effectiveShape` was undefined, and so when it hit [this line](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/models.js#L225), it returned without setting `effectiveArea`. `Math.round(null) == 0`, so that's why it showed up as 0 in the UI. When the page was refreshed, since `effectiveArea` was still `null`, we tried to [set it again](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/models.js#L430-L436). This time, `App.map.AoI` isn't set yet, so the intersect returns the whole modification shape, and `effectiveArea` becomes equal to `area`. I'll add some more detail in the commit message.

src/mmw/js/src/modeling/models.js
caseypt: Will do.

caseypt:
Updated with @rajadain's suggestions.

Another way to see that this is working is that when adding modifications to a scenario, the model run should be quick because no new censuses have to be generated.

rajadain:
Ah I see. Also in my earlier screenshot, when a shape was completely outside the AoI I would just see "0", not "0.00 m<sup>2</sup>" which I do now:

image

When adding or deleting a modification that is completely outside the AoI, the calcs take the same amount of time as for when adding or deleting a modification which does have a non-zero effective area though. I cannot observe the model run speedup you mention above.

caseypt:
@rajadain and I discussed the model run performance issues noted above, and I was interpreting things incorrectly. As long as all of the modifications are outside of the AoI, the model doesn't need a census. However, if one modification is added within the AoI, any subsequent modifications added outside the AoI will trigger census generation because, while we are still only sending one modification to the geoprocessing workflow, the modification hash still reflects the modifications outside the AoI. For simplicities sake, we thought it still made sense to generate the hash off of the modifications before any processing.

rajadain:
+1

lliss:
+1

Issue #897, opened 2015-10-02
When creating a new scenario, pass along AoI census if it exists from previous scenario.


caseypt:
Since a new scenario doesn't have any modifications and contains the same AoI as other scenarios in a project, we can share the AoI census. This will help speed things up and reduce the load on the geoprocessing worker each time a new scenario is created.

PR #937, opened 2015-10-12
Share AoI Census

jamesmcclain:
When a new scenario is created, the census of the Area of Interest (AoI) is automatically attached to the new scenario.  Since the AoI is the same for all scenarios in a project, this can be done.  Doing this reduces the load the Geoprocessing service.

Connects #897

**To Test**
   * Open up the developer tools so that you can see the network traffic generated by the application, and also watch the spark-jobserver logs.
   * Go to a project and create a new scenario.
   * There should be no polling for the AoI census with this patch applied, but without it there should be.
   * Caching works on the compare page, as well.

**Limitations**
When a new project is created, there will still be two separate queries: one for "current conditions" and one for the "new scenario".  Since the census of the AoI is not available at that point, this cannot be helped without delaying the creation of the "new scenario" scenario.

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -737,9 +737,10 @@ var ScenariosCollection = Backbone.Collection.extend({
         return this.setActiveScenario(this.get({ cid: cid }));
     },
 
-    createNewScenario: function() {
+    createNewScenario: function(aoi_census) {
         var scenario = new ScenarioModel({
-            name: this.makeNewScenarioName('New Scenario')
+            name: this.makeNewScenarioName('New Scenario'),
+            aoi_census: aoi_census
         });
 
         this.add(scenario);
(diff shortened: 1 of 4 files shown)


caseypt:
Code looks good, but I'm going to do some more testing tomorrow. Nice work.

src/mmw/js/src/modeling/models.js
```diff
@@ -494,6 +494,12 @@ var ScenarioModel = Backbone.Model.extend({
             ]
         });
 
+        // If the census of the Area of Interest is available, then
+        // attach it so that it does not need to be recomputed.
+        if (attrs.aoi_census) {
+            this.set('aoi_census', attrs.aoi_census);
+        }
```
caseypt: Does the above code block (488 - 495) not do the job of setting `aoi_census` as well as any other attributes passed in on initialization?

jamesmcclain: Hmm ... I think it probably does ...

caseypt:
+1. Tested. Nice job.

jamesmcclain:
Thanks, I will make the cleanups you suggested before merging.

Issue #913, opened 2015-10-07
Precipitation and opacity sliders do not render in Internet Explorer


arobbi7:
Supercedes #504 

Running Internet Explorer 11 on a Windows VM: precipitation and opacity sliders do not render on landing page, model view, or compare view.
![screen shot 2015-10-06 at 5 16 24 pm](https://cloud.githubusercontent.com/assets/12868342/10339794/17961d24-6cdb-11e5-8eb8-06a003b393be.png)
![screen shot 2015-10-06 at 5 19 42 pm](https://cloud.githubusercontent.com/assets/12868342/10339798/1f9336ba-6cdb-11e5-8cc0-e918dc5d1f7b.png)
![screen shot 2015-10-07 at 10 09 56 am](https://cloud.githubusercontent.com/assets/12868342/10339901/993d00d6-6cdb-11e5-85ea-265b7d6659f7.png)

PR #938, opened 2015-10-12
Fix sliders in Internet Explorer

rajadain:
## Overview

The invisibility of the sliders in Internet Explorer was caused by its inability to automatically determine their width. By giving them, or their containers, explicit widths, we ensure that the sliders are always visible.

In addition, their behavior was different because IE doesn't trigger `input` event when the slider is being dragged, but instead triggers `change` continuously. This is different from Firefox, Chrome and Safari, which trigger `input` when dragging the slider, and `change` when the user stops dragging.

Previously we would only update the UI on the `input` event to preview the value selected by the user. Unfortunately, since IE never fires it, the UI would not update even when the user stopped dragging. This edit ensures that the UI is updated with the final value.

The actual fetching of the results is debounced, so even though we are firing a lot more `change` events than were originally expected, there are no more network requests and performance is quite consistent.

## Testing Instructions

Create a project, and proceed to the Model view. Test the sliders. Go to the Compare view. Test the sliders.

Test in IE, but also in other browsers to ensure that the expected behavior hasn't changed.

## Demo

Chrome 45 on Mac:

image

Safari 9 on Mac:

image

Firefox 41 on Mac:

image

IE 11 on Windows:

image

Firefox 41 on Windows:

image

## Notes

I can't get the opacity slider to actually change the opacity of the coverage layer in either IE or Firefox. These observations should probably be made into a new issue.

Connects #913

### DIFF
--- a/src/mmw/js/src/modeling/controls.js
+++ b/src/mmw/js/src/modeling/controls.js
@@ -212,6 +212,12 @@ var PrecipitationView = ControlView.extend({
                 value: imperialValue
             });
 
+        // Update values for IE which doesn't trigger onSliderDragged. Will
+        // effectively noop on other browsers, since the same values were
+        // already set by onSliderDragged.
+        this.ui.slider.attr('value', value);
+        this.ui.displayValue.text(this.getDisplayValue(value));
+
         PrecipitationSynchronizer.syncTo(this);
 
         this.addOrReplaceInput(modification);
--- a/src/mmw/sass/base/_header.scss
+++ b/src/mmw/sass/base/_header.scss
@@ -214,7 +214,9 @@ header {
         // Undo bootstrap input[type="range"] syling
         input {
           display: inline-block;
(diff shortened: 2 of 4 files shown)


lliss:
+1

jamesmcclain:
+1 The precipitation sliders work in Firefox, Chrome, and IE11.

During testing, I noticed that the stream slider and NLCD opacity slider do not work in IE 11 and the NLCD opacity slider does not work in Firefox.  I understand that those were not part of this card, though.

If they can be very easily fixed, perhaps that can be done as well.  If not, we can make a separate card.

rajadain:
Thanks for confirming that behavior, I saw it and documented it in my Notes section above. I've created #943 for it.

jamesmcclain:
Cool.  Nice work on getting the stream sliders working -- I know that debugging these types of edge cases is probably not easy.

Issue #915, opened 2015-10-07
Search bar does not render in Internet Explorer


arobbi7:
Running IE 11 on a Windows VM:
![screen shot 2015-10-06 at 5 12 18 pm](https://cloud.githubusercontent.com/assets/12868342/10340129/a24fb316-6cdc-11e5-89b3-099f9e51fe29.png)

PR #942, opened 2015-10-12
Fix IE search bar.

lliss:
IE will not respect the line height for search boxes to make the box larger.
Creating an explicit height fixes this.

Connects #915 

### To test
 * bundle and then view in IE.
 * search bar should look the same as other browsers.

screenshot from 2015-10-12 18 47 19


rajadain:
+1 tried in different browser widths, works consistently.

Issue #899, opened 2015-10-02
Omitting layer table name produces exception


hectcastro:
```
Internal Server Error: /api/modeling/boundary-layers/
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/views.py", line 452, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/views.py", line 449, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/decorators.py", line 50, in handler
    return func(*args, **kwargs)
  File "/opt/app/apps/modeling/views.py", line 241, in boundary_layer_detail
    table_name = layers[0]['table_name']
KeyError: u'table_name'
```

mmcfarland:
I'm not sure how these invalid requests are getting generated, I wondered about a crawlerbot but it looks like it's from an Android tablet:

`Mozilla/5.0 (Linux; Android 5.1.1; Nexus 7 Build/LMY48I) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/45.0.2454.94 Safari/537.36`

We should try our Nexus 7 to see if it's building the layers dropdown incorrectly, and either way, guard against this key error producing a 500 and return a 400 Bad Request or similar.

PR #947, opened 2015-10-13
Return 404s for invalid boundary layer requests.

caseypt:
* Previously a 500 error would be returned if an table name wasn't
provided or was invalid. A 404 error is more appropriate for these
cases.

I tested the site out on our Nexus table and my iPhone, and the boundary selection worked fine in both cases. I wasn't able to reproduce this error using the app. I had to visit the bad URL directly.

**Testing instructions**
- Visit http://localhost:8000/api/modeling/boundary-layers/
- A DRF 404 error page should be returned.
- Visiting the same URL or `develop`, production or staging should throw a 500.

Connects to  #899

### DIFF
--- a/src/mmw/apps/modeling/views.py
+++ b/src/mmw/apps/modeling/views.py
@@ -276,13 +276,17 @@ def _construct_tr55_job_chain(model_input, job_id):
 @decorators.api_view(['GET'])
 @decorators.permission_classes((AllowAny, ))
 def boundary_layer_detail(request, table_code, obj_id):
-    layers = [layer for layer in settings.LAYERS
-              if layer.get('code') == table_code]
-    table_name = layers[0]['table_name']
-    json_field = layers[0].get('json_field', 'geom')
+    try:
+        layers = [layer for layer in settings.LAYERS
+                  if layer.get('code') == table_code and
+                  layer.get('boundary')]
+        table_name = layers[0]['table_name']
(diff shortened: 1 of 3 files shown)


src/mmw/apps/modeling/views.py
```diff
@@ -276,13 +276,16 @@ def _construct_tr55_job_chain(model_input, job_id):
...
+    try:
+        layers = [layer for layer in settings.LAYERS
+                  if layer.get('code') == table_code]
+        table_name = layers[0]['table_name']
+        json_field = layers[0].get('json_field', 'geom')
+
+        query = 'SELECT {field} FROM {table} WHERE id = %s'.format(
+                field=json_field, table=table_name)
+    except KeyError:
+        return Response(status=status.HTTP_404_NOT_FOUND)
```
mmcfarland: I think it might be useful to distinguish between the two failure scenarios: a request to a non-existent resource and not actually specifying a resource.  Consider checking if `table_code` is not provided and returning early with a `400` in that case.

caseypt: Will do.

caseypt:
Updated.

I modified the URL capture logic for this endpoint so that you can't not provide parameters. This takes care of the case of a missing table code or feature id.

I also added a test case for the bad table code scenario.

mmcfarland:
+1, tested.

Issue #918, opened 2015-10-07
Limit zoom levels for overlay tiles


mmcfarland:
Generating tiles at low zoom levels [without caching](https://github.com/WikiWatershed/model-my-watershed/issues/727) is too expensive and causes production tilers to fail.  We should limit the zoom levels at which boundaries can be displayed, per layer, and indicate that in the UI.

This applies to both the select AoI by boundary and the overlays.

* Add max zoom level to layer settings.  This should be appropriate for each set, for instance, we should stop drawing HUC12 at 9 because there's not a lot of utility in having them displayed below that, plus there's a lot of them.  Congressional districts would need to stop around 6 or 5 because they're larger and simpler.

* Communicate this in the UI.  Presumably by disabling the option and perhaps adding some explanation text about need to zoom in to enable that layer.

This should help resolve the immediate tiler issues in "production", currently.

PR #949, opened 2015-10-13
Limit Overlay Zoom Levels

jamesmcclain:
This pull request includes the following changes:
   * When the zoom level is sufficiently low (zoomed-out) it
      * prevents unavailable overlay layers from being displayed,
      * makes sure that no unwanted after-image is displayed, and
      * makes sure that unavailable overlay layers cannot be activated with the radio buttons
   * Does something similar to the predefined shape overlays that are available from the "Select By Boundary" dropdown.

The minimum zoom levels (`layer_settings.py`) -- except for the two given in the issue -- are best-guesses based on my testing.  If they do not meet everyone's needs, they can be changed.

**Issues**
   * Clicking a disabled entry in the "Select By Boundary" dropdown still causes the dropdown to close.
   * Zooming in very far with google layers enabled causes the overlay tiles to disappear.  This has been observed on staging, so was not caused by this PR.

**To Test**
   * Active the overlay layers menu (as in the first picture below).
   * Zoom in and out.  If you zoom far enough out, some of the selections will be disabled.
   * If you zoom too far out with an overlay active, it will disappear until you zoom in far enough to see it again.
   * Try the same thing with the predefined shape overlays from the "Select By Boundary" menu.

screenshot from 2015-10-13 15 38 09

screenshot from 2015-10-13 15 38 13

screenshot from 2015-10-13 15 40 21

screenshot from 2015-10-13 15 40 30

Connects #918

### DIFF
--- a/src/mmw/js/src/draw/templates/selectType.html
+++ b/src/mmw/js/src/draw/templates/selectType.html
@@ -10,7 +10,8 @@
         <a role="menuitem" tabindex="-1" href="#"
            data-tile-url="{{ item.url }}"
            data-layer-code="{{ item.code }}"
-           data-short-display="{{ item.short_display }}">
+           data-short-display="{{ item.short_display }}"
+           data-min-zoom="{{ item.minZoom }}">
           <span>{{ item.display }}</span> <i class="split fa fa-question-circle"
                                              data-content="{{ item.helptext }}"></i>
         </a>
--- a/src/mmw/js/src/core/layerControl.js
+++ b/src/mmw/js/src/core/layerControl.js
@@ -78,6 +78,9 @@ module.exports = L.Control.Layers.extend({
         }
 
         input.layerId = L.stamp(obj.layer);
+        if (obj.overlay && obj.layer && obj.layer.options && obj.layer.options.code) {
+            $(input).attr('id', obj.layer.options.code);
+        }
 
         L.DomEvent.on(input, 'click', this._onInputClick, this);
 
--- a/src/mmw/js/src/core/utils.js
+++ b/src/mmw/js/src/core/utils.js
@@ -6,6 +6,22 @@ var _ = require('underscore'),
 var M2_IN_KM2 = 1000000;
 
 var utils = {
+    // A function to enable/disable UI entries in response to zoom
+    // level changes.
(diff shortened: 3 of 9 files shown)


src/mmw/js/src/draw/templates/selectType.html
```diff
@@ -10,7 +10,8 @@
         <a role="menuitem" tabindex="-1" href="#"
            data-tile-url="{{ item.url }}"
            data-layer-code="{{ item.code }}"
-           data-short-display="{{ item.short_display }}">
+           data-short-display="{{ item.short_display }}"
+	   data-min-zoom="{{ item.minZoom }}">
```
rajadain: Spacing / Indentation discrepancy

jamesmcclain: Will correct, thanks.

rajadain:
I really like how the boundaries are grayed out in the "Select by Boundary" dropdown. I wish that the same effect was applied to the Overlays text, because it is hard to tell from just the radio button that it is disabled:

image

rajadain:
The Google basemaps (Satellite with Roads and Terrain) have higher available zoom levels than the Streets and Satellite ones. If I zoom in all the way while some overlays are enabled (such as HUC-8), the overlays disappear, but the selector doesn't disable.

Zoom level limitations seem to apply in both extremes: too far in and too far out. But the disabled selector seems to only look at one of them: zooming out. It would be consistent if the selector was disabled every time it went outside its supported range.

jamesmcclain:
The overlays are not supposed to disappear when you zoom in past the maxZoom level, I'll fix that (though I would be curious to see whether that was the case before these changes).

jamesmcclain:
I am able to zoom all the way in with (e.g.) HUC-8 enabled, and it does not disappear (with any of the basemaps).  Are sure that the problem is not your tiler (or perhaps I am misunderstanding what you are saying).

rajadain:
+1 new styling edits look good, works well overall. The disappearing boundary layers at superzoom levels in Google basemaps are hard to reproduce and not deal breaking. I think we're okay with that behavior.

jamesmcclain:
Cool beans, thanks for the thorough review, as always.

Issue #950, opened 2015-10-13
ITSI Login and Screenshot Issues


rajadain:
If a user wants to register with ITSI, but the autopopulated name is already present in MMW, the user gets a "Username already exists" prompt:

![image](https://cloud.githubusercontent.com/assets/1430060/10468136/68547ae0-71cb-11e5-9f4d-a043e200ae11.png)

If, discouraged by this and unaware that they can change their username, the user clicks "Continue as Guest", the draw tools don't render.

If the user can somehow get to a stage where the draw tools do render, the screenshot functionality does not work, and returns a login screen instead.

SRGDamia1:
This is happening to me every time I try to log-in with ITSI.  I am redirected the ITSI page where I successfully log-in, but then I am taken back to this page where it seems that the application is trying to create a new user based on the information I just gave ITSI instead of logging in with it.

mmcfarland:
@SRGDamia1 that actually is what's happening.  We create a user account in MMW (with no password) and link it to the ITSI user that was just authenticated.  If the ITSI user has no account in MMW yet, the dialog you see is the one to create that account based on the user info provided by ITSI.  We need to prompt because there could actually be an ITSI username (say mmcfarland) account already in MMW.  We're not going to default the MMW name to be `{{ITSI_USER}}.itsi` to make name collisions less likely.

We're doing some work now (https://github.com/WikiWatershed/model-my-watershed/pull/951) to make that whole process more explicit.

SRGDamia1:
Question:  will there be a way to link an already existing ITSI account to an already existing MMW account?  This probably will not be a common workflow; I'm just curious.

mmcfarland:
@SRGDamia1 We aren't planning a UI mechanism for that and there might some strange behavior when having both a setting to try to authenticate against oAuth ITSI and having a password in the system.  If it's something that comes up we'd have to investigate.

mmcfarland:
@SRGDamia1 In the next release, there will be no "sign up" stage for users authenticating against ITSI for the fist time.  When they first visit MMW, if we find no MMW account for the ITSI user, we will simply create one automatically and use a unique name for their MMW account which they don't need to select.  This should make the workflow much more straightforward.  See the screenshots in #995 for a preview.

SRGDamia1:
Looks awesome, thank you!

PR #951, opened 2015-10-13
ITSI Login and Screenshot Issue Fixes

rajadain:
## Overview

Fixes login and screenshot issues as described in #950. See commit messages for details.

## Testing Instructions

 1. Login to https://learn.staging.concord.org/ as `tstudent2` and open the MMW Example da76db03 activity.
 2. Start the activity. It should automatically try to log you in, but will not find a corresponding account, and will try to create one for you.
 3. The account name should autopopulate as `tstudent2.itsi`. Change it to `tstudent2`. Try signing up. You should see the "Username already exists" prompt.
 4. Dismiss the modal by clicking outside, or by selecting "Continue as Guest". You should be redirected to the home page and the draw tools should render.
 5. Try and take a screenshot. It should render correctly and not show you an ITSI or MMW login screen.
 6. Try and login. Select "ITSI" in the login box. Sign up as `tstudent2.itsi`. Ensure you can still take screenshots.

Connects #950

### DIFF
--- a/src/mmw/apps/user/views.py
+++ b/src/mmw/apps/user/views.py
@@ -130,7 +130,7 @@ def itsi_auth(request):
         # User did not authenticate. Save their ITSI ID and send to /sign-up
         request.session['itsi_id'] = itsi_user['id']
         return redirect(
-            '/sign-up/itsi/{username}/{first_name}/{last_name}?next={0}'
+            '/sign-up/itsi/{username}.itsi/{first_name}/{last_name}?next={0}'
             .format(
                 request.GET.get('next', '/'),
                 **itsi_user['extra']
--- a/src/mmw/js/src/app.js
+++ b/src/mmw/js/src/app.js
@@ -2,6 +2,7 @@
 
 var $ = require('jquery'),
     Marionette = require('../shim/backbone.marionette'),
+    shutterbug = require('../shim/shutterbug'),
     views = require('./core/views'),
     models = require('./core/models'),
     settings = require('./core/settings'),
@@ -48,6 +49,9 @@ var App = new Marionette.Application({
         // Not set until modeling/controllers.js creates a
         // new project.
         this.currentProject = null;
+
+        // Enable screenshot functionality
+        shutterbug.enable('body');
     },
 
     load: function(data) {
--- a/src/mmw/js/src/core/itsiEmbed.js
+++ b/src/mmw/js/src/core/itsiEmbed.js
@@ -3,7 +3,6 @@
 var _ = require('lodash'),
     Backbone = require('../../shim/backbone'),
     iframePhone = require('iframe-phone'),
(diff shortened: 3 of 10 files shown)


rajadain:
Alright I've re-setup the application to use a new `ngrok` URL running off my local. Ready for testing.

mmcfarland:
I think we should add a label to this screen, which is hard to distinguish from a login screen, that we are creating a MMW account linked to their ITSI Account.  Something along the lines of:

> We need to create a new account for you in Model My Watershed associated with your ITSI username: {{itsi_user}}.  Please pick a display name and sign up below.  You should always log in with the "Log in as ITSI" button.

screenshot from 2015-10-14 09 56 22

mmcfarland:
I'm not seeing the Draw Tools rendered as indicated in step 4 above:
screenshot from 2015-10-14 09 57 12

rajadain:
Okay taking another look. Will delete the `tstudent2.itsi` user to recreate some of this.

mmcfarland:
Guest screenshot works well, as does signup/login screenshot.

rajadain:
The reason the draw tools weren't rendering was because the activity had saved a state for a project that no longer existed (because I'd deleted the user in preparation for this):

image

And that error would break the Backbone controller. I've cleared the ITSI saved state and the draw tools render correctly now.

mmcfarland:
Given that user can manage their projects, is this in error condition that we should make an error for to handle better?

mmcfarland:
Tools are rendering in Guest mode now.

rajadain:
The error condition has now been minimally handled in dfbf4fd5ec1105caa796b4a9b942c7be8a1dafbe, and I've created #953 to make that more robust.

I've also added help text to the ITSI sign up modal. We can't have too much of it, because even one more line will make the user have to scroll to be able to press buttons on the modal:

image

Any feedback on further improving the help text would be welcome.

mmcfarland:
Looks good - I wonder if the "Sign Up with ITSI" button text could be improved.  Something like "Complete signup with ITSI" or "Finish linking account with ITSI" - both are pretty verbose, but "Sign Up With ITSI sounds like a _new_ option (like "Continue as Guest") instead of completing this current action, aka "Done".

rajadain:
Since we're already saying "This will link to your {username} ITSI account", we don't need to repeat "with ITSI" in the button, and can just say something like "Complete Sign Up" or "Finish Linking Accounts". On the other hand, we do have a lot of space in the wide button, and could potentially fit in more.

I also wanted to improve the "Always pick ITSI when logging in" line, which sounds incomplete and not very friendly. I want to say something more like "Just use ITSI when logging in to log in" but that sounds very redundant, verbose and confusing.

Perhaps we could get some design assistance?

rajadain:
Closing in favor of #955.

PR #955, opened 2015-10-14
ITSI Speedy Sign Up

rajadain:
## Overview

This aggressively streamlines the ITSI sign-up procedure, by removing all input from the user. We now automatically create an account for the user using information received from ITSI's OAuth end point. If there are any username collisions, we simply add a number to the username, and keep adding it until we get a unique username.

The sign up modal has been considerably simplified, and now serves a purely informational role. The exact wording may be improved. Currently it looks like this:

image

## Demo

mmw-itsi-speedy-sign-up

Connects #950 
Supersedes #951

### DIFF
--- a/src/mmw/apps/user/urls.py
+++ b/src/mmw/apps/user/urls.py
@@ -11,15 +11,13 @@ from apps.user.views import (login,
                              resend,
                              logout,
                              itsi_login,
-                             itsi_auth,
-                             itsi_sign_up)
+                             itsi_auth)
 
 urlpatterns = patterns(
     '',
     url(r'^logout$', logout, name='logout'),
     url(r'^itsi/login$', itsi_login, name='itsi_login'),
     url(r'^itsi/authenticate$', itsi_auth, name='itsi_auth'),
-    url(r'^itsi/sign_up', itsi_sign_up, name='itsi_sign_up'),
     url(r'^login$', login, name='login'),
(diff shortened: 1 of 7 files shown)


rajadain:
Only the last two commits are meaningfully different from #951 and worthy of inspection. I will setup an activity running off my local in ITSI tomorrow where it can be tested.

rajadain:
Ready for testing, just use MMW EXAMPLE 7461A830. Talked to @arobbi7 about the language. Might consider different messaging for embed auto sign-up and non-embed manual sign-up.

arobbi7:
This new workflow makes a lot of sense to me - nice job! I do think it's confusing to be told, within the ITSI embed portal, "When logging in, simply select ITSI." since the only login used to reach that point was the ITSI main login, not anything on MMW. 

For more clarity I think that modal should just read "We've created an account for you using your <name> ITSI account. When logging in to Model My Watershed, simply select ITSI." This message would work in both the embed auto sign-up and the non-embed manual sign up.

rajadain:
I've updated the language to be more generic and removed the username entirely:

image

The reason this was done was because we found users wanting to enter their ITSI credentials in the username / password fields for Model My Watershed. By removing the username from the conversation entirely, we try and make it clearer that users should just press **ITSI** when logging in to Model My Watershed.

mmcfarland:
+1, tested.  Very seamless, I think this is a very nice improvement to the workflow.

Issue #940, opened 2015-10-12
Delay in chart refresh after spinner ends for new scenarios


caseypt:
Similar to https://github.com/WikiWatershed/model-my-watershed/issues/732, but only happens when the scenario is initially added.

As of now, this is happening on `develop` and `mmw-dev.azavea.com`, but not `mmw.azavea.com`, so the cause may be into between 13ff68067f9c44c250ac85592bef5fe0d56d7525 and c645cbd1732801330c0febf92fdc48d5102c4ae3.

![spinner](https://cloud.githubusercontent.com/assets/1042475/10438928/94b80d1c-7104-11e5-8e6f-b10231d6018b.gif)

jamesmcclain:
You may need to repoint this.

PR #957, opened 2015-10-15
Remove Chart Display Delay

jamesmcclain:
When the user clicked the little "+" button to create a new scenario, there was a slight delay between the time the spinners stop spinning and the chart was displayed.

That was caused by the fact that the `reset` method on the `ResultCollection` was firing a `reset` event which caused the spinner to stop spinning.  Now, that `reset` call is made with the `silent` option set to true so that the event is not fired, then it is later manually fired after the model results are obtained.

Connects #940

**To Test**
   * Go to a project and create a new scenario.
   * It should be possible to observe that there is no delay between the time the spinner stops spinning and the chart shows up.

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -571,7 +571,11 @@ var ScenarioModel = Backbone.Model.extend({
         delete response.inputs;
 
         if (!_.isEmpty(response.results)) {
-            this.get('results').reset(response.results);
+            if (this.get('waitToRefresh') === true) {
+                this.get('results').reset(response.results, {silent: true});
+            } else {
+                this.get('results').reset(response.results);
+            }
         }
 
         delete response.results;
@@ -652,6 +656,9 @@ var ScenarioModel = Backbone.Model.extend({
 
                 pollSuccess: function() {
                     self.setResults();
+                    if (self.get('waitToRefresh')) {
+                        self.get('results').trigger('reset');
+                    }
                 },
 
                 pollFailure: function() {
@@ -751,6 +758,7 @@ var ScenariosCollection = Backbone.Collection.extend({
             aoi_census: aoi_census
         });
 
+        scenario.set('waitToRefresh', true);
         this.add(scenario);
         this.setActiveScenarioByCid(scenario.cid);
     },


caseypt:
@rajadain I remember you doing something similar to this to solve another problem and then having to undo part of it later. Do you recall what that was?

src/mmw/js/src/modeling/models.js
```diff
@@ -652,6 +656,9 @@ var ScenarioModel = Backbone.Model.extend({
 
                 pollSuccess: function() {
                     self.setResults();
+                    if (self.waitToRefresh) {
+                        self.get('results').trigger('reset');
```
caseypt: Would `self.get('results').reset()` work here? Also, if there a case in `pollSuccess` where we don't want to trigger a reset? In other words, do we need the conditional or can we just call `reset` always?

src/mmw/js/src/modeling/models.js
```diff
@@ -751,6 +758,7 @@ var ScenariosCollection = Backbone.Collection.extend({
             aoi_census: aoi_census
         });
 
+        scenario.waitToRefresh = true;
```
caseypt: Attributes on models should be set using `scenario.set('foo', 'bar').` I know this is sort of like a meta attribute, but I think we should still probably use the Backbone methods.

jamesmcclain: This is not actually meant to be an official attribute, it is just an extra piece of state.  If the consensus is that it should be an attribute, then I can make it so.

rajadain:
The request here is similar to #732, which was solved by silently saving scenarios in #739. Unfortunately, that was too broad and resulted in scenarios not saving at all in #876, which was solved in #877 which made the silent saving more selective.

caseypt:
Thanks @rajadain. @jamesmcclain it might be worth looking through those series of issues and PRs and making sure we don't end up in a similar scenario. Initially it doesn't appear so.

src/mmw/js/src/modeling/models.js
jamesmcclain: Doing a reset there without the silent option set to true will definitely do a reset, but it also has other side effects (including emptying the existing collection, I believe).

jamesmcclain:
I am not too familiar with what went on earlier, but this problem actually seems to be fairly simple.  The problem was that the `reset` method was firing a `reset` event which was causing the spinner to stop spinning prematurely.  All I am doing here is delaying the firing of that event until we have everything that is needed to paint create the chart.

caseypt:
+1. Tested.

rajadain:
I'm also taking a look at the code. Please hold off on merging this until I've had a chance to finish the review. Thanks.

jamesmcclain:
Sure, no problem.

src/mmw/js/src/modeling/models.js
```diff
@@ -751,6 +758,7 @@ var ScenariosCollection = Backbone.Collection.extend({
             aoi_census: aoi_census
         });
 
+        scenario.set('waitToRefresh', true);
```
rajadain: Since this is set to `true` when a new scenario is created, and is never set to `false`, couldn't we simply remove it and all the corresponding `if` conditions?

jamesmcclain: Maybe, but the problem is that that `parse` function is being called by the framework, so it is not clear (to me anyway) exactly when and where that is called.  I wanted to be as conservative as possible.

rajadain: Once `waitToRefresh` is set to `true`, `if (this.get('waitToRefresh') === true)` will _always_ resolve to `true`, even if the `parse` is called by the framework, and all resets will be silent as per https://github.com/WikiWatershed/model-my-watershed/pull/957/files#diff-9abeff034ba5a68b4814ee90721888ebR575. If we want to be conservative, then at some point (when we know we are done) we should set it to `false`.

Or, instead of not firing the `reset` event, we might be able to change the spinners to listen to a different signal, which can be manually triggered, just like `reset` is in https://github.com/WikiWatershed/model-my-watershed/pull/957/files#diff-9abeff034ba5a68b4814ee90721888ebR660.

jamesmcclain: Okay, if I made that path unconditional, would that work for you?

rajadain:
After looking in to this a bit, this is what I've found:

 1. When we create a new scenario with the `+` button, we create a new scenario and save it immediately [`models.js:754`](https://github.com/WikiWatershed/model-my-watershed/blob/a8b7b27314decd86e60cbf424f35327342b51c18/src/mmw/js/src/modeling/models.js#L754).
 2. This triggers a POST to `/api/modeling/scenarios`, and also triggers a `fetchResultsIfNeeded` [`models.js:160`](https://github.com/WikiWatershed/model-my-watershed/blob/a8b7b27314decd86e60cbf424f35327342b51c18/src/mmw/js/src/modeling/models.js#L160)
 3. Since the new scenario has `null` `results` [`models.js:48`](https://github.com/WikiWatershed/model-my-watershed/blob/a8b7b27314decd86e60cbf424f35327342b51c18/src/mmw/js/src/modeling/models.js#L48), we trigger a `fetchResults` [`models.js:592`](https://github.com/WikiWatershed/model-my-watershed/blob/a8b7b27314decd86e60cbf424f35327342b51c18/src/mmw/js/src/modeling/models.js#L592)
 4. When `fetchResults` begins polling for the results, it sets `polling` to `true` on each of the scenario's `result`models [`models.js:650`](https://github.com/WikiWatershed/model-my-watershed/blob/a8b7b27314decd86e60cbf424f35327342b51c18/src/mmw/js/src/modeling/models.js#L650). This starts the spinners spinning, since that is what they listen to [`views.js:745`](https://github.com/WikiWatershed/model-my-watershed/blob/64b3c900423ffe33e9ad5eef041859ccf07c3a07/src/mmw/js/src/modeling/views.js#L745)
 5. While we're waiting for the results, the POST from step 2 returns, triggering the `parse` [`models.js:561`](https://github.com/WikiWatershed/model-my-watershed/blob/a8b7b27314decd86e60cbf424f35327342b51c18/src/mmw/js/src/modeling/models.js#L561). The `parse` method takes an optional `silent` parameter, which would prevent it from from firing any events, and this is indeed what we used in #739. We had to stop doing that in #877 for _new_ scenarios because we need to trigger a refresh after receiving the new `id`. So now the post-save `parse` for a new scenario is not silent. Unfortunately, the `results` we receive from the server have `polling` set to `false` by default, and when this is set it triggers the spinners (that are listening for `change:polling` [`views.js:745`](https://github.com/WikiWatershed/model-my-watershed/blob/64b3c900423ffe33e9ad5eef041859ccf07c3a07/src/mmw/js/src/modeling/views.js#L745)) to stop spinning.
 6. The results from step 4 return, triggering a silent `parse` and causing the graphs to render. This is the step that is _supposed_ to stop the spinners, but they have already stopped in step 5 above.

So, I think that preventing all `reset` event notifications is too broad a change. What we should do is set `polling` to true for new scenarios. But there is no clean place that jumps out to add this: these views are not separated enough and tightly coupled. I'm going to think a little bit more about how best to express this.

rajadain:
This is one way of solving it: https://github.com/rajadain/model-my-watershed/commit/777b6b87988ce6ed559f24b44a24b270676f5d53, where we set the value of polling to whatever the client's copy says it should be, but I'm not too happy with it. Will continue to think some more.

jamesmcclain:
Okay, I'll just close my PR and turn it over to you.

jamesmcclain:
My original solution for this problem was a small change that was guaranteed not to have side-effects because different behavior was triggered only if there was some special state (which later became an attribute on a model).

> Once waitToRefresh is set to true, if (this.get('waitToRefresh') === true) will always resolve to true, even if the parse is called by the framework, and all resets will be silent as per https://github.com/WikiWatershed/model-my-watershed/pull/957/files#diff-9abeff034ba5a68b4814ee90721888ebR575. 

My original solution just used some extra data that would not have been persisted, so that the behavior would have applied to that session only.  I could have gone further and set the `waitToRefresh` field to false after the first refresh if this was a concern.

> So, I think that preventing all reset event notifications is too broad a change. 

I thought that that was what you were suggesting above when you suggested getting rid of the conditionals.  I any case, no big deal.

rajadain:
> My original solution just used some extra data that would not have been persisted, so that the behavior would have applied to that session only.

Yes, the more I think about it the more I see that something along those lines is the right path. I can't seem to decide whether `polling` should or should not be a part of the persisted `ResultModel`. On one hand, it is not information that is useful long-term, and so should not be stored permanently on the server. On the other, it _is_ a part of a result's model state, and so should be kept together with `results` whose emptiness might be explained by `polling = true`.

> I thought that that was what you were suggesting above when you suggested getting rid of the conditionals

I apologize for that, it was poorly worded, I should've just talked to you in person. What I was trying to imply was that the conditions effectively didn't matter, since they would always apply, resulting in too broad a change.

---

I guess my true concern is that this code is becoming really complex (not just as a part of this PR but all the compounded work we've done to make the spinners be in sync with the graphs), and it feels like it is becoming more complex than it needs to be for such simple functionality. Instead of adding more variables to track more state, I wonder if it is possible to refactor in a way that makes this simpler and easier to manage.

rajadain:
Yet another solution might be this: https://github.com/rajadain/model-my-watershed/commit/f69d3409e8ae39b6b9424caf899cee7f0e642783. Unfortunately I didn't bring my work machine home this weekend so can't test it, but I think it should work. Essentially it transforms adding new scenarios to an existing project to mimic the paradigm used in adding the initial scenarios to a new project (where this out of sync spinners / results doesn't happen): add scenarios to collection silently [`controllers.js:255`](https://github.com/WikiWatershed/model-my-watershed/blob/683b1583ad00f32981d756132fddde77334e2c6a/src/mmw/js/src/modeling/controllers.js#L255) and manually call `fetchResultsIfNeeded` [`controllers.js:148`](https://github.com/WikiWatershed/model-my-watershed/blob/683b1583ad00f32981d756132fddde77334e2c6a/src/mmw/js/src/modeling/controllers.js#L148). This prevents the first POST request in step 2 above from happening, and allows the rest of our flow to proceed as usual. Of all the solutions here, this is probably the cleanest so far.

I apologize for the insistence, and I'm very grateful for your patience, but I feel responsible for the exceeding complexity we find ourselves dealing with for the syncing, having worked on earlier bits of autosaving. As long as we can find ways to make it work within those design decisions, we'll know that they still hold. But when there is no way to fix the problem except for very specific, surgical, hopefully self-contained fixes, we'll know that the design decisions no longer apply and we need to refactor things.

jamesmcclain:
Okay, it looks like there is a pretty good menu of options here.

rajadain:
Unfortunately upon testing the above I found that I'd have to modify it to handle duplication of scenarios, which was broken as a result, so one ends up with https://github.com/rajadain/model-my-watershed/commit/e1be51edde6dc377a000e9b7e0f5a10bd1754726. But that's not very elegant since it isn't DRY, and just shifts the mental burden of figuring out how the events are wired up to the mental burden of having to remember to manually fetch results. Furthermore, having so many `{ silent: true }` didn't feel right either.

Upon further investigation I found adding to a Backbone collection doesn't automatically save, but we add IDs to underlying scenarios every time [`models.js:122`](https://github.com/WikiWatershed/model-my-watershed/blob/a8b7b27314decd86e60cbf424f35327342b51c18/src/mmw/js/src/modeling/models.js#L122), which triggers a save [`models.js:508`](https://github.com/WikiWatershed/model-my-watershed/blob/a8b7b27314decd86e60cbf424f35327342b51c18/src/mmw/js/src/modeling/models.js#L508). Changing this would be wrong, since it is tied in to so many other things. Furthermore: this is functioning as it should, and shouldn't cause other problems.

So I went back to [`parse`](https://github.com/WikiWatershed/model-my-watershed/blob/a8b7b27314decd86e60cbf424f35327342b51c18/src/mmw/js/src/modeling/models.js#L561), in which we refresh all fields with server values for new scenarios (as described in step 5 a few comments above). This is excessive, since the only real new value we care about from the server is the scenario's `id`.

Thus, we have https://github.com/rajadain/model-my-watershed/commit/4d5acc0c566d0b6395787a33fa5fb7f8c0b9ba03, which is a more conservative implementation of #877, which only updates the client model with _new_ fields returned from the server, but keeps the values of _existing_ fields. This allows the scenario model to have `id`, `created_at`, and `modified_at` as populated by the server, but keep up-to-date values of everything else instead of defaults from the server.

I believe this bug (spinners ending early) was caused by my faulty implementation in #877, and by bettering that implementation we fix this, but also any other side-effects that may have caused.

jamesmcclain:
It seems that you have a pretty good idea of what you want to do.  Perhaps I should close this pull request and turn the card over to you?

rajadain:
Yes I would greatly appreciate that. Thanks @jamesmcclain for your patience and sportsmanship, and thanks @caseypt for catching the initial similarities and drawing my attention to this.

jamesmcclain:
No problem ... closing now ...

Issue #943, opened 2015-10-13
Coverage layer opacity does not work in IE or Firefox


rajadain:
To reproduce:

Open MMW in Firefox or Internet Explorer, open the Layers panel, select Overlays and then National Land Cover Database under Coverage. Once the Opacity Slider is visible, drag it all the way down to 0. Observe no change in the coverage layer opacity.

![image](https://cloud.githubusercontent.com/assets/1430060/10457185/2fd3b6c8-7193-11e5-99b2-6bcbf026d14a.png)

PR #962, opened 2015-10-16
Fix Opacity Slider in IE and Firefox

rajadain:
## Overview

`toElement` is not standard anymore (Chrome has it, IE used to have it, Firefox never had it). Switch to `target` which works across browsers.
See https://developer.mozilla.org/en-US/docs/Web/API/Event/target and http://stackoverflow.com/questions/8600174/event-toelement-in-ie8-and-firefox

## Testing Instructions

Test in Firefox and IE. Also test in Chrome to ensure it is still working there. Test in any other browsers you might have access to.

Connects #943

### DIFF
--- a/src/mmw/js/src/core/opacityControl.js
+++ b/src/mmw/js/src/core/opacityControl.js
@@ -68,7 +68,7 @@ module.exports = L.Control.extend({
         });
 
         $(view).on('mouseup', function(e) {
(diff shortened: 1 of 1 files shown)


jamesmcclain:
:+1: Works in Chromium and Firefox on Linux and IE 11 on Windows.

Issue #922, opened 2015-10-07
Modals don't protect against double clicks


rajadain:
See original comment https://github.com/WikiWatershed/model-my-watershed/pull/908#issuecomment-146012938

rajadain:
@lliss could you please elaborate?

lliss:
If you click on the button in the modal a few times very fast, you should be able to trigger a few events. The first one usually works and the send one usually throws an error or at least does an HTTP request that fails.

lliss:
Attempt to reproduce by going to the projects page, selecting delete and then confirming the deletion by clicking the button a few times very quickly. You should see a successful network request and a failed one as well.

lliss:
Same things happens with hitting "enter" a few times.

PR #964, opened 2015-10-16
Prevent Modal Double-Clicks

jamesmcclain:
The main buttons of various modals are now disabled after selection to prevent accidental double-clicking.

Connects #922

**To Test**
   * It should be possible to use the various modals as normal, in particular the "sign up" modal, the "sign in" modal, and the various modals that appear when you make choices on the "/projects" page.
   * It should no longer be possible to double-tap the main buttons on these modals such that they are activated more than once before the modal disappears.  (This is true using the mouse as well as the enter key).
   * By setting break points, for example at model-my-watershed/src/mmw/js/src/projects/views.js line 152, it should be possible to see that the buttons are now disabled after they are activated, preventing them from being double-clicked.

**Limitations**
   * Not all buttons have received this treatment.  I tried to limit it to buttons which I thought would cause a problem if double-activated.
   * I took on-board the suggestion that the modal should stay open in the event of failures, for example when deleting a project, but simply removing the calls to the `hide` method was not sufficient.  It may be the case that this behavior is coming from deeper within the framework.

### DIFF
--- a/src/mmw/js/src/core/modals/views.js
+++ b/src/mmw/js/src/core/modals/views.js
@@ -1,6 +1,7 @@
 "use strict";
 
 var _ = require('underscore'),
+    coreUtils = require('../utils.js'),
     Marionette = require('../../../shim/backbone.marionette'),
     ZeroClipboard = require('zeroclipboard'),
     models = require('./models'),
@@ -48,6 +49,17 @@ var ModalBaseView = Marionette.ItemView.extend({
         }
     },
 
+    getDisabledState: function($el) {
+        var model = this.model;
+
+        if (coreUtils.modalButtonDisabled(model, $el)) {
+            return true;
+        } else {
+            coreUtils.modalButtonToggle(model, $el, false);
+            return false;
+        }
+    },
+
     primaryAction: function() {
         // Not implemented.
     },
@@ -75,6 +87,10 @@ var ConfirmView = ModalBaseView.extend({
     }, ModalBaseView.prototype.events),
 
     primaryAction: function() {
+        if (this.getDisabledState(this.ui.confirmation)) {
+            return;
+        }
+
         this.triggerMethod('confirmation');
         this.hide();
     },
@@ -102,6 +118,10 @@ var InputView = ModalBaseView.extend({
     },
 
     primaryAction: function() {
+        if (this.getDisabledState(this.ui.save) === true) {
+            return;
+        }
+
         var val = this.ui.input.val().trim();
 
         if (val) {
--- a/src/mmw/js/src/core/utils.js
+++ b/src/mmw/js/src/core/utils.js
@@ -6,6 +6,23 @@ var _ = require('underscore'),
 var M2_IN_KM2 = 1000000;
 
 var utils = {
+    // A function for enabling/disabling modal buttons.  In additiion
+    // to adding the disabled class, it is also important to somehow
+    // note the fact that the button is disabled becaue there is an
+    // "enter key" listener which ignores the CSS class.
+    modalButtonToggle: function(model, $el, enabled) {
+        $el.attr('disabled', !enabled);
+        if (enabled === true) {
+            $el.removeClass('disabled');
+        } else if (enabled === false) {
+            $el.addClass('disabled');
+        }
+    },
+
+    modalButtonDisabled: function(model, $el) {
+        return $el.attr('disabled') ? true : false;
+    },
+
     // A function to enable/disable UI entries in response to zoom
     // level changes.
     zoomToggle: function(map, layerData, actOnUI, actOnLayer) {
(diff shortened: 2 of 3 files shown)


src/mmw/js/src/user/views.js
```diff
@@ -219,6 +239,14 @@ var LoginModalView = ModalBaseView.extend({
 
     // Login with ITSI
     itsiLogin: function() {
+        var selector = '.itsi-login',
+            model = this.model;
+
+        if (coreUtils.modalButtonDisabled(model, selector)) {
+            return;
+        }
+        coreUtils.modalButtonToggle(model, selector, false);
```
caseypt: This logic is slightly different than all of the other additions (it lacks an else statement). Is that intentional?

caseypt:
Consider implementing this code chunk ...

```js
var selector = '.class',
    model = this.model;

if (coreUtils.modalButtonDisabled(model, selector)) {
    return;
} else {
    coreUtils.modalButtonToggle(model, selector, false);
}
```

... as a method on `ModalBaseView` to make this more DRY.

caseypt:
I think we should consider merging `modals/views.js#ModalBaseView` and `users/views.js#ModalBaseView`, but that's probably a separate issue since it would involve a bit of moving around and testing.

caseypt:
Tested that you can no longer double click the modal when renaming or deleting projects.

src/mmw/js/src/user/views.js
jamesmcclain: No, it was just an oversight.  Since there is a return in the if-clause, there is no difference between putting this in an else and not.

jamesmcclain:
> Consider implementing this code chunk ...

I would like to do something like that, but that code's purpose prune the execution of the function.  In order to put it into another function, wouldn't I need to be able to do a non-local return?

src/mmw/js/src/core/utils.js
```diff
@@ -1,11 +1,29 @@
...
+    // A function for enabling/disabling modal buttons.  In additiion
+    // to adding the disabled class, it is also important to somehow
+    // note the fact that the button is disabled becaue there is an
+    // "enter key" listener which ignores the CSS class.
+    modalButtonToggle: function(model, selector, enabled) {
+        model.set(selector + '-disabled', !enabled);
+        if (enabled === true) {
+            $(':button' + selector).removeClass('disabled');
+        } else if (enabled === false) {
+            $(':button' + selector).addClass('disabled');
```
lliss: Is there a reason we don't use the disabled attribute as opposed to the class. I'm nearly certain that by adding `disabled` to the button, the enter key is also suppressed.

jamesmcclain: No, no reason.  Will try.

lliss: I think that should make things simpler.

src/mmw/js/src/user/views.js
caseypt: Got it. Consider making them uniform.

caseypt:
@jamesmcclain You could return a binary from that function and go from there.

caseypt:
Something like this:

```js
getDisabledState: function(selector) {
    var model = this.model;

    if (coreUtils.modalButtonDisabled(model, selector)) {
        return true;
     } else {
         coreUtils.modalButtonToggle(model, selector, false);
    }

    return false;
},

primaryAction: function() {
    if ((this.getDisabledState(this.ui.selector) === true) {
       return;
    }

    this.triggerMethod('confirmation');
    this.hide();
},

```

lliss:
+1. Tested.

src/mmw/js/src/core/utils.js
jamesmcclain: After some testing, it appears that setting the `disabled` attribute is not enough.

jamesmcclain:
Okay, I think that I have made changes for all of the comment received up to this point.  I will leave this open for a little while before merging.

src/mmw/js/src/core/modals/views.js
```diff
@@ -75,6 +87,10 @@ var ConfirmView = ModalBaseView.extend({
     }, ModalBaseView.prototype.events),
 
     primaryAction: function() {
+        if (this.getDisabledState(this.ui.confirmation) === true) {
```
lliss: Since you created getDisabledState, you know that it returns a boolean so you could just write
`if (this.getDisabledState(this.ui.confirmation)) { ...`

jamesmcclain: Truth.

lliss:
Failing because
```
js/src/core/utils.js: line 4, col 5, '$' is defined but never used.

1 error
...ignoring
```

lliss:
+1 Tested.

jamesmcclain:
Thanks for testing.

Issue #935, opened 2015-10-12
Check that name is in object before accessing values


caseypt:
On some legacy projects (I haven't figured out what is different about them yet, just that they are from a few months ago), the following error occurs:

![screen shot 2015-10-12 at 10 17 11 am](https://cloud.githubusercontent.com/assets/1042475/10429970/6846b9de-70ca-11e5-9bbe-000f71b54a42.png)

![screen shot 2015-10-12 at 10 17 36 am](https://cloud.githubusercontent.com/assets/1042475/10429977/75c186ca-70ca-11e5-8554-440d219e8729.png)

We should check that `name` exists in the object before trying to get values for it.

PR #716, opened 2015-08-24
 Pre-Columbian scenario

jamesmcclain:
This adds a "Pre-Columbian" chart to the "Current Conditions" scenario.

Connects #626

**To Test**
   * Apply the patch
   * Either reprovision the app and the worker, or do a `sudo pip uninstall tr55` `sudo pip install tr55` on both the app and the worker
   * Reload the worker `vagrant reload worker`
   * Go to a project
   * Go to the current conditions scenario.  There should now be a second chart showing the results for the AoI in Pre-Columbian times.

PR #966, opened 2015-10-19
Remove Incompatible Results from Scenarios

jamesmcclain:
This patch provides a data migratin which removes incompatible results from scenarios created prior to https://github.com/WikiWatershed/model-my-watershed/pull/716.

Connects #935

**To Test**
   * For testing purposes, you may want to take a snapshot of your services VM (halt the services VM, open up the Virtual Box GUI, click on the services VM [in the list on the left], click on snapshots [in the upper-right of the window], click on the "take snapshot" icon).
   * Run the new migration `./scripts/manage migrate`
   * Try to open an old scenario in the MMW application.  It should be possible now, whereas it was not before.

### DIFF
--- a/src/mmw/apps/modeling/migrations/0016_old_scenarios.py
+++ b/src/mmw/apps/modeling/migrations/0016_old_scenarios.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+def clear_old_scenario_results(apps, schema_editor):
+    Scenario = apps.get_model('modeling', 'Scenario')
+    for scenario in Scenario.objects.all():
+        if 'pc_modified' not in scenario.results or 'pc_unmodified' not in scenario.results:
+            scenario.results = "[]"
+            scenario.modification_hash = ""
+            scenario.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('modeling', '0015_remove_scenario_census'),
+    ]
+
+    operations = [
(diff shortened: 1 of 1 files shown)


jamesmcclain:
@azaveaci rebuild

jamesmcclain:
@azaveaci rebuild

jamesmcclain:
The CI check failed to a problem with the Geocoder, this is ready to be reviewed.

hectcastro:
@azavea-bot rebuild

rajadain:
+1 tested, was able to open old projects that would fail before. Works very well. Great job!

src/mmw/apps/modeling/migrations/0016_old_scenarios.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+def clear_old_scenario_results(apps, schema_editor):
+    Scenario = apps.get_model('modeling', 'Scenario')
+    for scenario in Scenario.objects.all():
+        if 'pc_modified' not in scenario.results or 'pc_unmodified' not in scenario.results:
```
rajadain: Great targeting of results missing the necessary values

caseypt:
+1. I can't test this because I don't have any legacy scenarios anymore, but I verified that it's only setting fields which essentially function as a cache and/or will be generated when the model is populated on the client.

jamesmcclain:
Okay, thanks for testing.  Merging now...

Issue #969, opened 2015-10-19
Model view screenshot missing disclaimer and map controls


rajadain:
In addition to #854 and #961, we have:
#### Expected:
![image](https://cloud.githubusercontent.com/assets/1430060/10593921/dd671602-7695-11e5-949a-a2925569eba8.png)
#### Observed:
![image](https://cloud.githubusercontent.com/assets/1430060/10593914/cc9e86ac-7695-11e5-8dab-8789d3a7693e.png)

Issue #961, opened 2015-10-16
Water quality results do not render in ITSI screenshot


arobbi7:
To replicate:
- Enter ITSI MWW staging activity as student
- Select an AoI, proceed through to Model view
- Click on Water Quality tab and take a snapshot

I saw this issue in Chrome, Firefox, and Safari on Mac, and Chrome & Firefox on Windows.
![screen shot 2015-10-16 at 10 50 07 am](https://cloud.githubusercontent.com/assets/12868342/10545371/6458eee4-73f6-11e5-83dc-b7060b5f3123.png)

PR #970, opened 2015-10-20
Fix Modeling view screenshots

rajadain:
## Overview

Because of the [5000px width set on `.scenario-tabs-wrapper`](https://github.com/WikiWatershed/model-my-watershed/blob/74485b4f67667d6e48288bea851400d9229c725f/src/mmw/sass/base/_header.scss#L149), and the way PhantomJS sees things, some elements which are positioned absolutely and have percent widths or `right: 0` would render way off screen.

We use the `shutterbug-saycheese` and `shutterbug-asyouwere` [events fired by Shutterbug](https://github.com/concord-consortium/shutterbug.js#shutterbug-jquery-custom-events) immediately before and after taking a screenshot to add a fixed width, equivalent to the viewport's width, to `#model-output-wrapper`, ensuring that all its contents are displayed within visible bounds.

## Testing Instructions

Checkout the branch and run [`ngrok http 8000`](https://ngrok.com/). Go to http://concord-consortium.github.io/lara-interactive-api/ and resize the iframe (using developer tools) to be `936px` wide and `702px` tall (these are the dimensions of an ITSI Activity Embed iframe). Then load the `ngrok` URL into the iframe, draw a shape and navigate to the Modeling view. Switch to the "Water Quality" tab in model results and take a snapshot using the "Snapshot" button:

image

Ensure that the screenshot renders both the table and the graph in Water Quality, and that the Map Controls are rendered correctly.

You can right click the tiny screenshot and open it in a new tab to see it in full size.

## Demo

**Before**

image

**After**

image

Connects #961 
Connects #969

### DIFF
--- a/src/mmw/js/src/app.js
+++ b/src/mmw/js/src/app.js
@@ -51,7 +51,7 @@ var App = new Marionette.Application({
         this.currentProject = null;
 
         // Enable screenshot functionality
-        shutterbug.enable('body');
+        initializeShutterbug();
     },
 
     load: function(data) {
@@ -99,6 +99,24 @@ function RestAPI() {
             });
         }
     };
+}
+
+function initializeShutterbug() {
+    $(window)
+        .on('shutterbug-saycheese', function() {
+            // Set fixed width before screenshot to constrain width to viewport
+            $('#model-output-wrapper, .map-container').css({
+                'width': window.innerWidth
+            });
+        })
+        .on('shutterbug-asyouwere', function() {
+            // Reset after screenshot has been taken
+            $('#model-output-wrapper, .map-container').css({
+                'width': ''
+            });
+        });
+
+    shutterbug.enable('body');
 }
 
 module.exports = App;


caseypt:
What was the effect of setting the above-mentioned element to a width < 5000px? There was probably a reason it was set to that, but it just seems odd at first glance.

rajadain:
If we set that width to be less than the width of all the scenario tabs, then the scenario tabs wrap, like so:

image

I tried looking in to ways to stop that from happening, and tried `white-space: nowrap` and `display: flex, flex-direction: row` and `overflow-x: hidden` but nothing worked. I suspect it is an effect of a combination of `absolute` and `relative` positioning we're using for all the components in there.

caseypt:
Thanks for the explanation.

src/mmw/js/src/app.js
```diff
@@ -50,6 +50,19 @@ var App = new Marionette.Application({
...
+                $('#model-output-wrapper, .map-container').css({
+                    'width': window.innerWidth
+                });
+            })
+            .on('shutterbug-asyouwere', function() {
+                // Reset after screenshot has been taken
+                $('#model-output-wrapper, .map-container').css({
+                    'width': ''
+                });
+            });
```
caseypt: Consider moving this code (and the following related `shutterbug` line`) to a helper function that can be called from `initialize` just for the sake of making the length of the function shorter.

caseypt:
+1. Tested. Nice work.

screen shot 2015-10-21 at 10 56 33 am

src/mmw/js/src/app.js
rajadain: Originally I tried to put all the Shutterbug code in [`setup.js`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/setup.js), but I was also looking at #971 and realized that we might need these handlers to be able to access `App.getLeafletMap()`, so it can't be initialized before `App`. Thus I just added it here.

Extracting it into an initialization function within this file sounds like a good compromise. I'll do that.

rajadain:
Encapsulated code into [`initializeShutterbug`](https://github.com/WikiWatershed/model-my-watershed/pull/970/files#diff-3a59f6b08229d723d3407c2ce03acea7R54) function.

src/mmw/js/src/app.js
```diff
@@ -101,4 +101,22 @@ function RestAPI() {
...
+        .on('shutterbug-asyouwere', function() {
+            // Reset after screenshot has been taken
+            $('#model-output-wrapper, .map-container').css({
+                'width': ''
+            });
+        });
+
+    shutterbug.enable('body');
+}
+
```
caseypt: :+1:

Issue #965, opened 2015-10-18
Deleting Project leads to inconsistent state


rajadain:
1. Log in and create a new project (so it is saved to the server).
 2. Delete the project. You should be taken back to the Draw view with the same AoI
 3. Proceed to Analyze. Proceed to Model view.
 4. The app should crash into an inconsistent state, with a 404 for the old project ID in the console window

![screen shot 2015-10-17 at 9 22 56 pm](https://cloud.githubusercontent.com/assets/1430060/10562040/3ce46e2c-7515-11e5-849a-144a1574a343.png)

mmcfarland:
I think this should take you back to your My Projects page upon encountering that 404.

Issue #762, opened 2015-09-08
Consolidate "Single Project Mode" navigation with new nav functionality.


lliss:
As of https://github.com/WikiWatershed/model-my-watershed/pull/756 some aspects of single project mode exist as the default condition of the application. Some duplicate code can almost certainly be removed.

rajadain:
Also consolidate resetting the app state / clearing the currently loaded project and Area of Interest to one place. An attempt to do so was made in https://github.com/WikiWatershed/model-my-watershed/commit/8608671cc3f428ee9957b18b50931c0a2b77e714, but it broke the functionality of #756.

PR #972, opened 2015-10-21
Project Deletion Flow

rajadain:
## Overview

Previously we would navigate to the Draw view after deleting a project, but we weren't clearing the `currentProject` or the Area of Interest correctly. Now, we clear those elements and navigate to the My Projects view instead. This view is always available for logged in users, and only logged in users can have saved projects, thus only logged in users can delete projects.

## Testing Instructions

Log in and draw an area. Proceed to the Modeling view. Delete the project. You should be redirected to My Projects list. Ensure that you can create a new project, and proceed to the Draw view, and create a new project all the way to the Modeling view. Also try launching an existing project from the My Projects view and ensure it works correctly.

## Notes

I tried to consolidate the project clearing functionality in one place (see https://github.com/WikiWatershed/model-my-watershed/compare/13f1904...rajadain:e5f2e93), specifically in https://github.com/rajadain/model-my-watershed/commit/8608671cc3f428ee9957b18b50931c0a2b77e714. But that ended up breaking #756. I've added a comment to #762 which I think would be a good fit for doing together.

Connects #965

### DIFF
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -159,7 +159,17 @@ var ProjectMenuView = Marionette.ItemView.extend({
                     confirmLabel: 'Delete',
                     cancelLabel: 'Cancel'
                 })
-            });
+            }),
+            navigateAfterDelete = function() {
+                App.currentProject = null;
+                App.projectNumber = undefined;
+                App.map.set({
+                    'areaOfInterest': null,
+                    'areaOfInterestName': null
+                });
+
+                router.navigate('projects/', { trigger: true });
+            };
 
         del.render();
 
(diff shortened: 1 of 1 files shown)


arobbi7:
Followed the testing instructions and it worked perfectly. I'm not totally sold on the actual workflow, though. If someone deletes their project from the Model page, I think they'd probably like to "start over" on the draw page, as opposed to adding the step of going to "Create a new project" from the My Projects page.

rajadain:
Furthermore, since the Draw view is also the "Start" view, it would make sense to return the user to this starting stage. I made the user go to My Projects on @mmcfarland's suggestion here https://github.com/WikiWatershed/model-my-watershed/issues/965#issuecomment-149716256, but re-reading that comment perhaps I interpreted it incorrectly.

mmcfarland:
I still kind of like the My Projects redirect, where you can still start a new project.  In most cases if I delete something, I expect to end up at some sort of home page which an empty map doesn't really fulfill for me - we also don't know that a user started at the map page, it could have been from My Projects as well, if they were returning to a saved project - which may be more likely over time.  Not wedded to the idea and happy to be overruled if there is consensus otherwise.

rajadain:
If we assume both use cases (to go to the Draw view, and to the My Projects view) to be equal in value, then we can differentiate on the cost of having to use the non-default one. Thus, if we make My Projects the default redirect, it only takes one click to go to the Draw view. But if we make the Draw view the default one, it would take two clicks (once to open the user dropdown, another to click the My Projects link) to go to the My Projects view. Thus, defaulting to My Projects is more cost effective.

mmcfarland:
The argument works for me.

arobbi7:
Makes sense to me.

caseypt:
+1. Tested. I agree with the workflow as implemented.

rajadain:
Thank you!

Issue #963, opened 2015-10-16
The choose_worker function fails to work after debugcelery has been run


jamesmcclain:
The function, https://github.com/WikiWatershed/model-my-watershed/blob/8e99574cad399db37f4cf0629d73c2ae7a2a3d1c/src/mmw/apps/modeling/views.py#L204-L207, filters out all workers that do not have the correct stack color in their name -- including the debug workers created by the `debugcelery` script.

PR #898, opened 2015-10-02
Cache censuses

caseypt:
Cache AoI and modification censuses.
* If a AoI census or modification censuses have already been generated
and are still valid, skip the geoprocessing tasks and run TR55 directly.
* If only the AoI census is available, generate the modification
censuses and provide the cached AoI census to TR-55.
* Modify the structure of the Scenario.census to contain the
modification hash from when the censuses were generated, as well
as both the AoI census and the modification censuses in the structure
expected in the geoprocessing tasks.
* Update and add tests to test caching logic.

Connects to #870

**Testing instructions:**
- Start a new project with a fairly large AoI so that it's easy to see the difference in computation time.

screen shot 2015-10-02 at 12 04 56 pm

- Switch to the "New Scenario"
- The initial model run should take a few seconds.
- Change just the precipitation slider.
- The model run should be faster than the previous run.
- Add a modification. The model run should be slower than the last run, and maybe a bit slower than the first run (more modifications means a longer model run time).
- Change just the precipitation slider again.
- The model run should be fast again.
- Repeat until you are confident there is some caching going on.
- Run the tests.

@jamesmcclain had a good idea that I didn't take on in this card because I spent a lot of time upfront reading through and rewriting things, and it's a discrete task. When a new scenario is added, we should pass along the AoI census from the previous or first scenario if the AoI census exists. Issue here: https://github.com/WikiWatershed/model-my-watershed/issues/897

Looking for input from @jamesmcclain re: new function and variable names, and function comments.

Issue #346, opened 2015-06-10
Early creation of a Project, persist in navigation


mmcfarland:
Right now, Projects are only created once the user navigates through to the `/model` page.  We should consider creating the Project earlier in the sequence in order to reduce the number of places we persist the area of interest shape and also to be able to navigate back and forth more consistently.  Currently, if you got to a `/model/id` page, navigate back to `/analyze` then forward again, you are looking a new project for the same area of interest.  This may or may not be what the user wanted, but it's pretty confusing.

ajrobbins:
We now save the project so that if you navigate back to analyze and then forward again, the user is back at the initial project.

PR #973, opened 2015-10-21
Restore Celery Debugging

jamesmcclain:
Contains a fix to the celery debugging functionality.

Connects #963

### DIFF
--- a/scripts/debugcelery.sh
+++ b/scripts/debugcelery.sh
@@ -6,6 +6,6 @@ set -e
 
 STOP_SERVICE="(sudo service celeryd stop || /bin/true)"
 CHANGE_DIR="cd /opt/app/"
-RUN_CELERY="envdir /etc/mmw.d/env celery -A 'mmw.celery:app' worker --autoreload -l debug"
+RUN_CELERY="envdir /etc/mmw.d/env celery -A 'mmw.celery:app' worker --autoreload -l debug -n debug@%n"
 
 vagrant ssh worker -c "$STOP_SERVICE && $CHANGE_DIR && $RUN_CELERY"
--- a/src/mmw/apps/modeling/views.py
+++ b/src/mmw/apps/modeling/views.py
@@ -202,7 +202,10 @@ def get_job(request, job_uuid, format=None):
 
 
 def choose_worker():
-    workers = filter(lambda worker: settings.STACK_COLOR in worker,
+    def predicate(worker_name):
+        return settings.STACK_COLOR in worker_name or 'debug' in worker_name
+
+    workers = filter(predicate,
                      celery.current_app.control.inspect().ping().keys())
     return random.choice(workers)
 


src/mmw/js/src/projects/controllers.js
```diff
@@ -4,6 +4,10 @@ var App = require('../app'),
     views = require('./views');
 
 var ProjectsController = {
+    projectsCleanUp: function() {
```
rajadain: [`projectsCleanup` already exists below](https://github.com/WikiWatershed/model-my-watershed/blob/54d7d6244e537fb7c051dbb2ae2d2189ee20e5db/src/mmw/js/src/projects/controllers.js#L15-L17), consider adding the line

```javascript
App.aoiCensus = null;
```

to it instead.

src/mmw/js/src/analyze/controllers.js
```diff
@@ -55,8 +57,11 @@ var AnalyzeController = {
             });
 
         App.state.set('current_page_title', 'Geospatial Analysis');
-
         App.rootView.footerRegion.show(analyzeWindow);
+
+        analyzeWindow.model.get('aoiCensus').done(function(aoiCensus) {
+            App.aoiCensus = aoiCensus;
```
caseypt: I think we should try to avoid attaching more things to `App` and try to attach to `App.project` or something similar if we need to persist this across pages.

jamesmcclain: Okay, will-do

jamesmcclain: I cannot find the string App.project in the codebase.

caseypt: Sorry, it's `currentProject`, which may just be an integer ID at the moment.

jamesmcclain: `App.currentProject` is possibly undefined at the point that I need to store the area of interest in `analyze/controllers.js`.  I am a little bit nervous about reusing this name to store an incompatible object.

caseypt:
After reviewing this, I don't think we should implement it without first restructuring our code a bit. I understand why this was implemented the way it was, and I can't think of a better way myself, but we should really hesitate before attaching more things to `App` and having to manage more global state. I think we need to refactor how we are managing the project and our models.

If the `aoi_census` is something that we want to persist across the analyze page to the model page, we should consider creating the `Project` model sooner and storing the `aoi_census` there. We currently store the `aoi_census` on each scenario, but it may make more sense to store it on the project level. This could make it simpler to share the census between scenarios.

Curious for other's thoughts on this.

jamesmcclain:
I agree with the second paragraph.

jamesmcclain:
I am grateful to everyone who looked at this.  I would be open to either merging this (with the small changes suggested) and refactoring later, or refactoring then adding these changes back in again.  I do not know what the relative pros and cons are.

Regardless of what happens to the geop-reduction commit, I would still like to merge the commit that fixes the debugging situation.

caseypt:
@jamesmcclain I did test the first commit and verified that I can now run `debugcelery` and get results in the app.

mmcfarland:
I agree, it's not super desirable to have the current AoI census on `App`, when it logically belongs to the project.  Perhaps it's time to get around to https://github.com/WikiWatershed/model-my-watershed/issues/346?

rajadain:
I also think we should move `aoi_census` to the Project model instead of the Scenario model, since it is the same for the entire project. The Project model already stores the actual Area of Interest, and also the Area of Interest Name, so it makes sense for it to store the Census as well.

I guess the reason the AoI Census was not added to the Project model in the first place in #898 was because it wouldn't be available until a Scenario is created and triggers a model run. Still, it wouldn't be difficult to populate it on the Project after we have run the first Scenario.

I don't think we should merge in the `aoiCensus` work in this PR, but I think that we should keep the branch alive until after the refactor has been done, for it can act as a reference for that future implementation.

jamesmcclain:
Can I merge the debugging fix?

caseypt:
+1 to merging the debug fix. I tested it out.

jamesmcclain:
kk

Issue #971, opened 2015-10-20
Map tiles / shapes positioned incorrectly in screenshot


rajadain:
Related to #854 and #817.

Because of Leaflet's use of `translate3d` transforms, which are not supported in PhantomJS, odd things happen. To reproduce the following, open the [LARA Interactive Playground](http://concord-consortium.github.io/lara-interactive-api/) in Chrome and load MMW Staging in the iframe, using the developer tools to resize it to 936px wide and 702px tall, the same height as in ITSI activities.

Draw a shape and move to the Modeling view, then move the map around. Take a snapshot. Notice the differences in shape positioning and map tiles between the snapshot and the real thing:

**Actual Page**

![image](https://cloud.githubusercontent.com/assets/1430060/10619660/e121be0a-7744-11e5-81d1-b7bc6099e1df.png)

**PhantomJS Snapshot**

![image](https://cloud.githubusercontent.com/assets/1430060/10619650/d31fd2c4-7744-11e5-9190-4c8e20a972f1.png)

This is because Leaflet is using CSS 3D transforms to position tiles and overlay objects:

![image](https://cloud.githubusercontent.com/assets/1430060/10619737/4e7c0384-7745-11e5-8361-d0484f11eab5.png)

But these 3D transforms are not supported by PhantomJS. These are also the underlying cause behind tiles differing in a screenshot without moving the map in Firefox, where they are [natively rendered using 3D transforms](https://github.com/WikiWatershed/model-my-watershed/issues/817#issuecomment-142601839).

While Leaflet does have [a way to disable 3D transforms](http://leafletjs.com/reference.html#global-l_disable_3d), it must be declared before including `leaflet.js` on the page:

> You need to set the switch as a global variable to `true` _before_ including Leaflet on the page, like this:
>```html
><script>L_PREFER_CANVAS = true;</script>
><script src="leaflet.js"></script>
>```

But that is not possible in our browserified approach. Furthermore, we don't want to disable 3D transforms for all cases, since they are 3D accelerated and improve performance on all modern platforms.

Find a way to switch out the 3D transforms for older equivalents (perhaps using [2D transforms](http://stackoverflow.com/a/8144361/2053314) instead) using the `shutterbug-saycheese` and `shutterbug-asyouwere` [events fired by Shutterbug](https://github.com/concord-consortium/shutterbug.js#shutterbug-jquery-custom-events) as used in #970 to make PhantomJS render the correct map.

Issue #854, opened 2015-09-24
Follow up on #817


kdeloach:
Follow up on #817 by fixing screenshot in Firefox.

SRGDamia1:
I sent an email to @chitchens and @mmcfarland about it; but in short I'm having problem 1 from #817 happen every single time in every browser I've tried.  A teacher trying the system with students pointed it out as well.

rajadain:
There seems to be an issue in ITSI where, instead of sending the entire contents of the iframe only sends the container element:

![image](https://cloud.githubusercontent.com/assets/1430060/10440995/119221c6-7112-11e5-8db0-5c71328fd436.png)

Curiously, if we use the same URL in the [LARA Interactive Playground](http://concord-consortium.github.io/lara-interactive-api/) it works correctly:

![image](https://cloud.githubusercontent.com/assets/1430060/10441014/42c7c23c-7112-11e5-89c4-28605b919c85.png)

It is most probably an ITSI configuration error. Perhaps @scytacki can offer some guidance.

It should be noted that we have had successful screenshots from within ITSI using Chrome in the past.

scytacki:
It worked for me in Chrome and Firefox in ITSI on a OS X.  If MMW in the iframe doesn't respond to the request for the html of the iframe, then you will see the behavior above. This could also happen if MMW takes too long to respond.  When it works you should see in the console: `cant handle type: htmlFragResponse`. If you don't see that message then it means the iframe isn't sending back its html.

This might still be a bug in ITSI, but it is hard to tell when I can reproduce it.

SRGDamia1:
The snapshot is showing nothing but the log-in page for me on Chrome, Firefox, and Internet Explorer on two different Windows computers.  MMW doesn't feel slow to me, but the snapshot takes forever.
![snapshot](https://cloud.githubusercontent.com/assets/5374350/10457594/2697afcc-7195-11e5-9aa7-dcb4fd42f0c7.jpg)

rajadain:
Screenshots seem to be working now in Firefox on Windows (with the notable exception of offset maps):

![image](https://cloud.githubusercontent.com/assets/1430060/10458987/7afe750e-719b-11e5-93b3-0a467b61ce5f.png)

And accurately in Chrome on Windows:

![image](https://cloud.githubusercontent.com/assets/1430060/10459041/b6f851e2-719b-11e5-9f54-d21281cf4d97.png)

[...]

Issue #817, opened 2015-09-21
ITSI Issues in Firefox


rajadain:
The following issues are seen on Firefox when running Staging at 78ee147510:

 1. Taking a screenshot results in a login window (in the screenshot) (hard to reproduce) Original issue #747 
 2. Taking a screenshot results in a blank screenshot (e.g. https://ccshutterbugtest.s3.amazonaws.com/1817cae5052.png, captured source https://ccshutterbugtest.s3.amazonaws.com/1817cae5052.html) (hard to reproduce) Original issue #747 
 3. ~~Taking a screenshot of the micro site results in a blank screenshot (e.g. https://s3.amazonaws.com/labbook.concord.org/snapshots/f1f89c44d3aab09921da426490fb49b95f9364ac.png), but with more interesting source HTML (https://s3.amazonaws.com/labbook.concord.org/snapshots/f1f89c44d3aab09921da426490fb49b95f9364ac.html). Original issue #788~~ 
 4. When screenshot is successful, the map is usually offset. It appears that the map is magnified and moved up and left in the viewport. Original issue #747 
 5. ~~"Undo all my work" button is still visible in Firefox, even though `{ "reset": false }` is being sent correctly from the app. Original issue #790~~

rajadain:
Regarding 4:

So far I've been unable to pinpoint the difference in the HTML being sent by Firefox and Chrome which cause these discrepancies. Interesting samples are https://ccshutterbugtest.s3.amazonaws.com/6e8aebce4d8.png (made by Chrome) and https://ccshutterbugtest.s3.amazonaws.com/a8bd8d08ceb.png (made by Firefox). The corresponding HTML pages render identically in Chrome (https://ccshutterbugtest.s3.amazonaws.com/6e8aebce4d8.html and https://ccshutterbugtest.s3.amazonaws.com/a8bd8d08ceb.html), and not at all in Firefox because the data URI in the iframe's `src` is much larger than Firefox will accept.

rajadain:
Regarding 5:

It seems to have been a deployment issue which has fixed itself.

rajadain:
[...]

PR #978, opened 2015-10-22
Disable 3D transforms in embed mode to fix screenshots

rajadain:
## Overview

PhantomJS used for taking screenshots doesn't support CSS 3D transforms, causing the map and drawn shapes to appear askew and displaced. The flag `L_DISABLE_3D` instructs Leaflet not to use 3D transforms, but we cannot abandon them entirely because of performance improvements they offer due to hardware acceleration.

By disabling 3D transforms only for cases when we know we are running in an embedded environment, we achieve a solution that helps in most cases.

## Testing Instructions

Checkout the branch and run [`ngrok http 8000`](https://ngrok.com/). Go to http://concord-consortium.github.io/lara-interactive-api/ and resize the iframe (using developer tools) to be `936px` wide and `702px` tall (these are the dimensions of an ITSI Activity Embed iframe). Then load the `ngrok` URL into the iframe (for example `https://12cdaec6.ngrok.io/?itsi_embed=true`), draw a shape and navigate to the Modeling view. Move the map around and take a snapshot using the "Snapshot" button:

image

Ensure that the screenshot renders both the map tiles and any shapes on it correctly.

You can right click the tiny screenshot and open it in a new tab to see it in full size.

Test in Chrome and Firefox.

## Demo

**Actual**

image

**Screenshot: Before**

image

**Screenshot: After**

image

Connects #971 
Connects #854 
Connects #817

### DIFF
--- a/src/mmw/apps/core/templates/base.html
+++ b/src/mmw/apps/core/templates/base.html
@@ -52,6 +52,13 @@
         </script>
         <script type="text/javascript">
             window.clientSettings = {{ client_settings | safe }};
+
+            if (window.clientSettings.itsi_embed) {
+                // Instruct Leaflet to not use CSS 3D transforms in embed mode
+                // since they are not supported by PhantomJS which is used for
+                // taking screenshots.
+                window.L_DISABLE_3D = true;
+            }
         </script>
         <script type="text/javascript" src="{% static 'js/vendor.js' %}"></script>
         <script type="text/javascript" src="{% static 'js/main.js' %}"></script>


src/mmw/apps/core/templates/base.html
```diff
@@ -52,6 +52,11 @@
         </script>
         <script type="text/javascript">
             window.clientSettings = {{ client_settings | safe }};
+
+            if (window.clientSettings.itsi_embed) {
+                // Disable CSS 3D transforms in embed mode to fix screenshots
```
caseypt: Consider updating this comment and noting that this is a Leaflet-specific setting.

caseypt:
Still seeing the issue in Chrome:

screen shot 2015-10-22 at 12 52 19 pm

caseypt:
Cancel that. Missed the detail about the flag in the URL.

rajadain:
Updated comment. You may have to refresh the page (right click the iframe and select "Reload Frame") after setting the flag for the JavaScript to get uncached.

Also, for an overview of how this is working is in #971

caseypt:
+1. Tested via walk-through from @rajadain.

rajadain:
Thanks!

Issue #974, opened 2015-10-22
ITSI Signup field length errors


mmcfarland:
The production site is reporting unhandled errors relating to ITSI user sign up requests with a too long field name.

```
Internal Server Error: /user/itsi/authenticate
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/app/apps/user/views.py", line 132, in itsi_auth
    return itsi_create_user(request, itsi_user)
  File "/opt/app/apps/user/views.py", line 155, in itsi_create_user
    last_name=last_name,
  File "/usr/local/lib/python2.7/dist-packages/django/contrib/auth/models.py", line 187, in create_user
    **extra_fields)
  File "/usr/local/lib/python2.7/dist-packages/django/contrib/auth/models.py", line 182, in _create_user
    user.save(using=self._db)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py", line 734, in save
    force_update=force_update, update_fields=update_fields)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py", line 762, in save_base
    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py", line 846, in _save_table
    result = self._do_insert(cls._base_manager, using, fields, update_pk, raw)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py", line 885, in _do_insert
    using=using, raw=raw)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/manager.py", line 127, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 920, in _insert
    return query.get_compiler(using=using).execute_sql(return_id)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/sql/compiler.py", line 974, in execute_sql
    cursor.execute(sql, params)
  File "/usr/local/lib/python2.7/dist-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python2.7/dist-packages/django/db/utils.py", line 97, in __exit__
[...]

PR #986, opened 2015-10-27
ITSI user field length restriction

mmcfarland:
Due to the automatic nature of ITSI signups, the app would have
attempted to create a username beyond the db column limit.  This commit
prevents excessively long user names from being entered.  Long First/Last names are just arbitrarily truncated.

For a rational as to why I didn't create a custom user model, see the issue discussion in #974

##### Test:
Use the long ITSI username on the staging portal `mmcfarlandjunioresquirethesecondsenior` to create a new account in MMW, it should be successful.

Connects #974

### DIFF
--- a/src/mmw/apps/user/tests.py
+++ b/src/mmw/apps/user/tests.py
@@ -6,7 +6,9 @@ from __future__ import division
 import requests
 
 from django.test import LiveServerTestCase
+from django.test import TestCase
 from django.contrib.auth.models import User
+from apps.user.views import trim_to_valid_length
 
 
 class TaskRunnerTestCase(LiveServerTestCase):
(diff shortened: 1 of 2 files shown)


rajadain:
@azavea-bot rebuild

rajadain:
+1 tested, the username was truncated automatically and an account with a unique suffix was created:

image

All tests passed.

src/mmw/apps/user/tests.py
```diff
@@ -97,3 +99,15 @@ def test_no_token_bad_credentials_returns_400(self):
         self.assertEqual(response.status_code, 400,
                          'Incorrect server response. Expected 400 found %s'
                          % response.status_code)
+
+
+class ItsiSignupTestCase(TestCase):
+
+    def test_small_name(self):
+        username = trim_to_valid_length('shortname', '.itsi')
+        self.assertLessEqual(len(username), 30)
```
rajadain: Consider adding a second assert to ensure that `username` is "shortname.itsi". The concatenation isn't very obvious given that the function is named `trim_to_valid_length`, so it would be valuable to have a test to communicate and verify that behavior.

mmcfarland:
@azavea-bot rebuild

Test updated.

src/mmw/apps/user/tests.py
```diff
@@ -97,3 +99,17 @@ def test_no_token_bad_credentials_returns_400(self):
...
+        self.assertEqual(username, 'shortname.itsi', 'The short name should ' +
+                         'concatenated but otherwise unmodified')
```
rajadain: Missing "be" before "concatenated"

PR #988, opened 2015-11-03
Avoid AZ auto-assignment; use us-east-1a for AMI creation

hectcastro:
Without `availability_zone` set, Packer allows AWS to automatically select an availability zone (AZ) for an instance launch. Unfortunately, AWS appears to be selecting `us-east-1b`, which does not have a default VPC subnet. Here we tell Packer to always use `us-east-1a`, because there is a default subnet setup for it.

vpc_management_console


jamesmcclain:
I think the only way to really test this is to merge it.  It looks okay to me, though.

jamesmcclain:
+1

Issue #990, opened 2015-11-04
Remove "Results are for demonstration purposes..." text


mmcfarland:
The app is no longer demonstrating values, but calculating them - we don't need the fine print anymore.  It may be just a revert on https://github.com/caseypt/model-my-watershed/commit/f479dfd5c71f5fcf1185efa486ee665d4c5cebf8, but it can at least be used as a guide as to what to remove.

PR #991, opened 2015-11-04
Remove disclaimer

lewfish:
Remove disclaimer text since it is no longer accurate along with associated CSS. To test, check that it is removed from the Analyze, Model, and Compare views.

Connects #990


caseypt:
+1. Tested.

screen shot 2015-11-04 at 6 08 18 pm
screen shot 2015-11-04 at 6 08 23 pm
screen shot 2015-11-04 at 6 08 30 pm

PR #993, opened 2015-11-05
Update Java version.

lliss:
OpenJDK had a security update and version 7u79 is no longer available. This was
causing the stack to not properly provision.

### To test:
Destroy one of the VMs that requires Java and then reprovision it.


jamesmcclain:
+1 I was able to rebuild `worker` and reprovision `app` on this branch.

hectcastro:
:+1:

Issue #871, opened 2015-09-29
Add code for taking metrics


jamesmcclain:
@hectcastro suggested adding code to take metrics on the geoprocessing code.

The following suggestion was made:

> Some example usage of the StatsD client is in the [Boto mailer backend](https://github.com/WikiWatershed/model-my-watershed/blob/53c325afa9b0ba209ce9ca7c29ccf2b3eec8ee3a/src/mmw/apps/core/mail/backends/boto_ses_mailer.py). More detail into the available data types can be found [here](http://statsd.readthedocs.org/en/latest/types.html). 

And suggested areas for doing that are in the `histogram_start` and `histogram_finish` functions in `geoprocessing.py` which will be introduced in https://github.com/WikiWatershed/model-my-watershed/pull/826

mmcfarland:
I'm specifically interested in getting some timing metrics for how much of the geoprocessing time is due to networking latency.  I suspect this might be hard to isolate since it's embedded int he trellis library code.  At the very least, we need to know how long the job takes to complete.

PR #997, opened 2015-11-05
Add statsd metrics to geoprocessing jobs

kdeloach:
Add timing metrics around geoprocessing jobs and SJS requests.

Connects #871

Test by verifying that statsd metrics appear for geoprocessing tasks in the graphite dashboard:

image

### DIFF
--- a/src/mmw/apps/modeling/geoprocessing.py
+++ b/src/mmw/apps/modeling/geoprocessing.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 from __future__ import absolute_import
 
 from django.conf import settings
+from django_statsd.clients import statsd
 from celery.exceptions import MaxRetriesExceededError
 
 import requests
@@ -51,6 +52,7 @@ SOIL_MAPPING = {
 }
 
 
+@statsd.timer(__name__ + '.histogram_start')
 def histogram_start(polygons):
     """
     Together, histogram_start and histogram_finish implement a
@@ -60,6 +62,10 @@ def histogram_start(polygons):
 
     This is the top-half of the function.
     """
+    @statsd.timer(__name__ + '.histogram_start.sjs_post')
+    def post(url, data):
+        return requests.post(url, data)
+
     host = settings.GEOP['host']
     port = settings.GEOP['port']
     path = settings.GEOP['path']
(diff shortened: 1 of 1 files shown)


jamesmcclain:
It looks good to me, but it might be worth checking to see if everyone's needs have been met.

For example: someone had expressed interest in knowing how much time was being spent (on average) in the Geotrellis code.  I think that this can be computed by dividing one less than the count of `histogram_finish.sjs_get` (the number of times SJS was polled) by `histogram_start.sjs_post` and multiplying by the polling interval.  If people are happy with indirect numbers like that, then everything should be good (I don't know if it is possible to get more direct numbers).

+1

jamesmcclain:
Based on what I know of this package (which is limited), the fact that the communication with the backend code is split into two pieces would make it difficult to get the number directly.

kdeloach:
The `sum` field for this metric captures duration so we should be able to derive average duration spent on geoprocessing jobs (and how much time we spent waiting for SJS). On the other hand, this doesn't do a good job at measuring network latency because I have not added any additional metrics inside the actual geotrellis code.

jamesmcclain:
This is truth.

jamesmcclain:
I should point out that virtually 100% of the time spent in Geotrellis code is network latency.

kdeloach:
:open_mouth:

mmcfarland:
Is `histogram_finish.sum` the cumulative time for a job to complete?  It'll be a good metric to establish the length of time for a few known boundaries, and then try those same boundaries under different proposed worker scenarios.

jamesmcclain:
If that quantity is a time, I am not sure whether it includes the polling interval or not.

kdeloach:
This should measure the duration of individual tasks, not the entire celery chain. Polling is not included.

mmcfarland:
+1

hectcastro:
Not sure that instrumenting the `mmw-geoprocessing` project would help much with tracking S3 latency either (because that happens inside of GeoTrellis Spark). I think the most effective approach would be to instrument GT with some StatsD client, and then have it expose configuration so that we can tell it where to spit metrics.

mmcfarland:
Yeah, I think this just gets us overall timings to provide a baseline to see if any of the different approaches make an overall improvement to any of the jobs.  Unless we get something inside of the GT code, I don't know if we can a finer resolution than that.

Issue #992, opened 2015-11-04
Proxy Observation API through nginx


mmcfarland:
The Observation API does not have CORS and the metadata request needs to be cached daily, so set up a caching proxy to http://www.wikiwatershed-vs.org/

PR #999, opened 2015-11-05
Setup cached proxy for Vizer endpoint

mmcfarland:
The Vizer webservices at http://www.wikiwatershed-vs.org/ are not CORS
enabled and the meta (asset) request was asked to be cached.

* Proxy through the app server so client can make ajax requests
* Cache the response for a period of time, the details of which may need
  to be worked out further.

To Test:
* Reprovision `app`
* Request http://localhost:8000/observation/services/get_asset_info.php?opt=meta&asset_type=siso and see that X-Proxy-Cache is `MISS`
* Refresh that request, see that X-Proxy-Cache is `HIT`

The endpoint is being proxied and cached.

Connects #992 

Big h/t to @hectcastro for help.

### DIFF
--- a/deployment/ansible/roles/model-my-watershed.app/tasks/reverse-proxy.yml
+++ b/deployment/ansible/roles/model-my-watershed.app/tasks/reverse-proxy.yml
@@ -1,4 +1,13 @@
 ---
+- name: Ensure Nginx cache directory exists
+  file: path="{{ nginx_cache_dir }}"
+        state=directory
+        owner="{{ ansible_ssh_user }}"
+        group=mmw
+        mode=0775
+  notify:
+    - Restart Nginx
+
 - name: Configure Nginx site
   template: src=nginx-app.conf.j2
             dest=/etc/nginx/sites-available/mmw-app.conf
--- a/deployment/ansible/roles/model-my-watershed.app/templates/nginx-app.conf.j2
+++ b/deployment/ansible/roles/model-my-watershed.app/templates/nginx-app.conf.j2
@@ -1,3 +1,10 @@
+# Protects against scenarios where DNS records for the
+# site being proxied change.
+resolver 8.8.8.8 valid=300s;
+resolver_timeout 10s;
+
+proxy_cache_path {{ nginx_cache_dir }} levels=1:2 keys_zone=OBSERVATION:10m max_size=100m;
+
 server {
   listen *:80;
   server_name _;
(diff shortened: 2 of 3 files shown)


deployment/ansible/roles/model-my-watershed.app/templates/nginx-app.conf.j2
```diff
@@ -1,3 +1,10 @@
+# Protects against scenarios where DNS records for the
+# site being proxied change.
+resolver 8.8.8.8 valid=300s;
```
kdeloach: Does the Google public DNS update quicker than others?

kdeloach:
+1 Tested

deployment/ansible/roles/model-my-watershed.app/templates/nginx-app.conf.j2
mmcfarland: Not sure, but it's used in quite a lot of examples on setting up a `resolver` when proxying to a remote public server.

lliss: I had to hit the page twice to get cached results. That might have been a fluke though. It looks good.

deployment/ansible/roles/model-my-watershed.app/tasks/reverse-proxy.yml
```diff
@@ -1,4 +1,13 @@
 ---
+- name: Ensure Nginx cache directory exists
+  file: path="{{ nginx_cache_dir }}"
+        state=directory
+        owner="{{ ansible_ssh_user }}"
```
hectcastro: Ideally, this directory would be owned by `www-data`. I think `mode=0775` makes it work in this case though.

Also just a little worried about how this will impact other things going on in the root volume, but the lowest entry point for ephemeral storage is an `m3.medium`. We're currently running `t2` instances everywhere else.

deployment/ansible/roles/model-my-watershed.app/templates/nginx-app.conf.j2
hectcastro: I think going with `8.8.8.8` is just the quickest way of getting one value to work across all environments. We could use a resolver provided by EC2, or something else local to the office, but then we'd have to make it a variable.

deployment/ansible/roles/model-my-watershed.app/tasks/reverse-proxy.yml
mmcfarland: Good call on `www-data`, I'll make that change.

Issue #976, opened 2015-10-22
`run_tr55` doesn't handle key errors


mmcfarland:
I'm not sure if this was an old project that has an invalid key, but I believe `developed_high` is (supposed to be) a valid land type.

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "apps/modeling/tasks.py", line 194, in run_tr55
    model_output = simulate_day(aoi_census, precip, cell_res=resolution)
  File "/usr/local/lib/python2.7/dist-packages/tr55/model.py", line 395, in simulate_day
    return simulate_modifications(census, fn, cell_res, precolumbian)
  File "/usr/local/lib/python2.7/dist-packages/tr55/model.py", line 343, in simulate_modifications
    mod = create_modified_census(census)
  File "/usr/local/lib/python2.7/dist-packages/tr55/model.py", line 228, in create_modified_census
    parent = mod['distribution'][cell]
KeyError: u'a:developed_high'
```

jamesmcclain:
The code where the exception originated (`tr55/model.py:228`) does not exist in newer version of the code.

PR #1004, opened 2015-11-05
Catch TR55 key errors.

lliss:
Connects #976

The new version of TR55 raises KeyError when there it receives simulation data
with unsupported types. This updates our worker to catch such errors and
gracefully fallback to avoid crashes.

### To test:
 - Reprovision worker to get new TR55.
 - Start celery debug script and watch logs.
 - Supply junk data to TR55 (Try the attached patch to let the UI do it for you.)
 - Ensure that we log the exception and do not crash.

### DIFF
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -195,12 +195,23 @@ def run_tr55(censuses, model_input, cached_aoi_census=None):
 
     # Run the model under both current conditions and Pre-Columbian
     # conditions.
-    model_output = simulate_day(aoi_census, precip, cell_res=resolution)
-    precolumbian_output = simulate_day(aoi_census, precip,
-                                       cell_res=resolution, precolumbian=True)
+    try:
+        model_output = simulate_day(aoi_census, precip, cell_res=resolution)
+        precolumbian_output = simulate_day(aoi_census,
+                                           precip,
+                                           cell_res=resolution,
+                                           precolumbian=True)
+        model_output['pc_unmodified'] = precolumbian_output['unmodified']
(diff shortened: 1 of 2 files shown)


lliss:
Let the UI give you bad data by applying this patch.
[bad-data.patch.txt](https://github.com/WikiWatershed/model-my-watershed/files/28325/bad-data.patch.txt)

src/mmw/apps/modeling/tasks.py
```diff
@@ -195,12 +195,23 @@ def run_tr55(censuses, model_input, cached_aoi_census=None):
...
+        model_output['pc_modified'] = precolumbian_output['modified']
+        runoff = format_runoff(model_output)
+        quality = format_quality(model_output)
+
+    except KeyError as e:
+        model_output = None
+        precolumbian_output = None
+        runoff = {}
+        quality = []
+        logger.info('Bad input data to TR55: %s' % e)
```
mmcfarland: Should this be `logging.error`?

lliss: Yeah. That would probably be better. I'll change it.

jamesmcclain:
I did not see this before sprint review.  I will try to get this back to you quickly.

lliss:
Rebased with logging.error

lliss:
> I did not see this before sprint review. I will try to get this back to you quickly.

I only assigned it to you moments before review. No worries.

jamesmcclain:
+1 I was able to provision on the new branch and everything worked as expected without the bad data patch.  With the bad data patch, I was able to observe the errors being caught and displayed in red in `debugcelery.sh` output.

PR #1010, opened 2015-11-06
Change owner of nginx cache directory

mmcfarland:
`www-data` is the more appropriate owner of this dir.

After provisioning `app`:

http://localhost:8000/observation/services/get_asset_info.php?opt=meta&asset_type=siso

should still work and produce a cache hit on the second request.


mmcfarland:
@azavea-bot rebuild

hectcastro:
:+1:

Issue #1007, opened 2015-11-05
Add Observation assets to the map based on layer selection


mmcfarland:
When selecting a layer for #1006, place all the assets for that provider as markers on the map.  Each provider should allow for it's own icon style.

Issue #1006, opened 2015-11-05
Add provider based Observation layers


mmcfarland:
http://www.wikiwatershed-vs.org/services/get_asset_info.php?opt=meta&asset_type=siso
This returns all of the stationary in-situ objects (siso), which are the only asset types available.

For all assets in result, we'll extract the unique list of provider values. This list of values will represent the "layers" in the new Observation tab of the layer selector.

This can use the proxy url from #992

PR #1012, opened 2015-11-10
Add Observation layers to the layer control

mmcfarland:
The existing Vizer implementation at http://wikiwatershed-vs.org is
providing an API to observation sensors in the DRB.  Show these systems
as layers than can be added to the map.  Layers are defined by parsing
the entire network by data provider and symbolizing the sensor by
platform type.

To test, open the layer selector and see a new Observations tab with 6 layers that add markers around the DRB.

Connects #1006 
Connects #1007 

**Note: I'm merging this into a long-running feature branch on upstream so that we won't deploy it until the full feature is complete.**


mmcfarland:
screenshot from 2015-11-10 11 58 14

lewfish:
+1, tested

mmcfarland:
Updated comments, selector names and whitespace.

Issue #1000, opened 2015-11-05
Increase timeout for geospatial analysis to support larger AoIs


arobbi7:
**We need to confirm this with Stroud**, but now that we have reduced the maximum allowed AoI to 1,000 km^2 we have precluded selection by most of our boundaries and many AoI's as a result. It makes it really hard not to hit the size limit while exploring AoIs in the current version of the app.  We should consider increasing timeout/allowable area, or potentially removing the larger boundary layers (HUC-8. for example) as a temporary measure while they are inactive.

jamesmcclain:
Not to flog a deceased equine, but we still have these issues: https://github.com/WikiWatershed/mmw-geoprocessing/issues/5 and https://github.com/geotrellis/geotrellis/issues/1183.  (These are substantively the same, but the second one is more clearly written.)

It may be worthwhile to consider terminating the source of the problem.

hectcastro:
> equine, adjective
>
> of, relating to, or affecting horses or other members of the horse family: equine infectious anemia.

jamesmcclain:
I will start by doubling it.  Will change in response to feedback if need-be.

PR #1013, opened 2015-11-11
Lengthen Geoprocessing Timeout

jamesmcclain:
The geoprocessing timeout has been lengthened from ~20 seconds to ~40 seconds.  The maximum supported AoI size has been increased from 1,000 square kilometers to 2,700.

Connects #1000

The time limit exists in two places: the `TaskModel` model in the front-end and the combination of the polling interval and maximum number of retries in the back-end.

**To Test**
   * Confirm that the maximum AoI size is now 2,700 square kilometers.
   * Confirm analysis can now take more than 20 seconds -- possibly as many as 40 (do that by providing an AoI that is close to the size limit).

**Note**.  You may need to wait for the internet connection problem to be resolved before testing this.

### DIFF
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -56,7 +56,7 @@ def start_histograms_job(polygons):
     }
 
 
-@shared_task(bind=True, default_retry_delay=1, max_retries=20)
+@shared_task(bind=True, default_retry_delay=1, max_retries=42)
 def get_histogram_job_results(self, incoming):
     """ Calls a function that polls SJS for the results
     of the given Job. Self here is Celery.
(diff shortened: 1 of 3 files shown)


src/mmw/apps/modeling/tasks.py
```diff
@@ -56,7 +56,7 @@ def start_histograms_job(polygons):
     }
 
 
-@shared_task(bind=True, default_retry_delay=1, max_retries=20)
+@shared_task(bind=True, default_retry_delay=1, max_retries=42)
```
lewfish: Since the polling interval on the frontend is 500 and timeout is 45,000, it seems that there could be up to 90 polls. So, why is max_retries set to 42?

jamesmcclain: There is a slight delay between when the clock starts in the front-end and when it does in the back-end.  The attempt here is to have them both finish at roughly the same time, with the back-end finishing slightly before.

lewfish: Oh, ok. You might want to put that in a comment. +1, tested.

jamesmcclain: Okay, will-do, thanks for reviewing.

Issue #994, opened 2015-11-05
Soil bar graph labels overlapping in analyze view


arobbi7:
Observed in Chrome, Firefox, and Safari on 13" Mac and in IE 11 on Windows.

![screen shot 2015-11-05 at 9 45 25 am](https://cloud.githubusercontent.com/assets/12868342/10971403/1c6c7c3e-83a2-11e5-98ac-760618c52963.png)

Issue #857, opened 2015-09-25
Chart labels are cut off or obscured on small screens


caseypt:
The left side of the longer labels are cut-off.

<img width="717" alt="screen shot 2015-09-25 at 2 04 24 pm" src="https://cloud.githubusercontent.com/assets/1042475/10108340/5a9a8732-638e-11e5-900a-15d8bd1d73e5.png">

This is in Chrome on a 13" MacBook.

rajadain:
They're all really **crammed** in there too. Perhaps we can add some vertical spacing as well?

caseypt:
In addition, on mobile the axis labels are unreadable

![img_1963](https://cloud.githubusercontent.com/assets/1042475/10459013/8f526ad8-719b-11e5-8bf8-2866ae95f1d1.PNG)
![img_1962](https://cloud.githubusercontent.com/assets/1042475/10459011/8f4dabec-719b-11e5-807a-daaf4cb5a461.PNG)

lewfish:
Below a certain screen size, it's impossible to fit all the labels in without cramming. So I'm thinking we need make the axis font smaller, or leave out some labels when possible (like with tick marks).

rajadain:
I agree. In addition we can also consider staggering the labels on two rows, so they alternate and leave more room. But not showing all labels / showing only a subset seems like the best way forward.

mmcfarland:
I was also thinking about a two row alternating approach.  Additionally, for squished values that are numeric we could abbreviate them 1,000,000 -> 1m, etc.

lewfish:
We could also display 1,000,000 as 1.0 and append ' (million)' to the axis label (eg. Level (millions)) to save more space on the tick labels.

lewfish:
Could we just turn the vertical bar chart above (and also the one in https://github.com/WikiWatershed/model-my-watershed/issues/994) into horizontal bar charts? That would avoid the issue with the tick labels overlapping. I don't think the staggered approach is going to look very nice, and will be more work to implement. The only vertical bar charts left in the app would be in the compare view, but those aren't problematic since there's only one bar in each chart.

rajadain:
I would be okay with us transitioning the charts to horizontal below a certain width. In the runoff results above there seems like a lot of unused horizontal space, so perhaps we can ensure we use all the available horizontal space before deciding to switch.

Another possibility is to use vertical or slanted labels, like this:

[...]

Issue #853, opened 2015-09-24
First digit of load numbers is cut off in Compare view - Firefox & Safari


arobbi7:
![screen shot 2015-09-24 at 11 50 43 am](https://cloud.githubusercontent.com/assets/12868342/10080336/28d75468-62bd-11e5-837e-aa2b846b83a2.png)

lliss:
It turns out that this happens in all browsers with an appropriately large AIO. After 5 digits, there is not enough space for the numbers in the axis.

arobbi7:
@lewfish, will the work you're doing on charting in MMW address this?

lewfish:
Yes!

Issue #1011, opened 2015-11-06
Slim down fat bars


lewfish:
Some of the vertical bar charts look ridiculous because they have a single bar in them which fills the entire width of the container they are in. We should cap the width of individual bars to avoid this.

![screen shot 2015-11-06 at 4 17 37 pm 2](https://cloud.githubusercontent.com/assets/1896461/11008989/fbcb08f0-84a1-11e5-8d2b-88ce17038e86.png)

PR #1016, opened 2015-11-12
Switch to NVD3 graph library and increase responsiveness

lewfish:
This PR switches from a hand-rolled graphing library to the open source (Apache 2.0 license) NVD3 library and in doing so, adds more responsiveness to the graphs to better support a variety of screen sizes. In particular, changes in the container size trigger changes in the number of ticks, the staggering of the ticks, and the layout of the legend. In addition, the numbers in the quality graphs in the compare view are now formatted to take up less space.

To test, run `npm install` and `bundle.sh --vendor`. Make sure all charts look similar to the originals  and that things don't  start overlapping or get cut off when the height and/or width of the screen gets small.

##### Known Issues
* ~~There is an error message that shows up in Chrome (but not Firefox) when showing vertical charts despite things rendering properly. I'm having a hard time figuring this out.~~ Fixed by https://github.com/lewfish/model-my-watershed/commit/f0344f744d7f38d01dfc6a0aa3a68621a403c629
screen shot 2015-11-12 at 9 55 50 am
* The NVD3 library doesn't currently support ordering the legend using the same order as the bar segments. It also doesn't let you put the legend to the right, so this looks different than before. I think we can patch NVD3 in a separate PR.
screen shot 2015-11-12 at 9 56 09 am 2
screen shot 2015-11-12 at 9 55 37 am 2

Connects #857 
Connects #994 
Connects #853

### DIFF
--- a/src/mmw/bundle.sh
+++ b/src/mmw/bundle.sh
@@ -100,6 +100,7 @@ CONCAT_VENDOR_CSS_COMMAND="cat \
     ./node_modules/leaflet-draw/dist/leaflet.draw.css \
     ./node_modules/font-awesome/css/font-awesome.min.css \
     ./node_modules/bootstrap-table/dist/bootstrap-table.min.css \
+    ./node_modules/nvd3/build/nv.d3.min.css \
     > $VENDOR_CSS_FILE"
 
 JS_DEPS=(backbone
@@ -121,6 +122,7 @@ JS_DEPS=(backbone
          leaflet-plugins/layer/tile/Google
          lodash
          nunjucks
+         nvd3
          turf-area
          turf-bbox-polygon
          turf-buffer
--- a/src/mmw/js/src/analyze/tests.js
+++ b/src/mmw/js/src/analyze/tests.js
@@ -10,28 +10,29 @@ var _ = require('lodash'),
     models = require('./models'),
     views = require('./views'),
     App = require('../app'),
-    testUtils = require('../core/testUtils'),
-    sandboxTemplate = require('../core/templates/sandbox.html');
+    testUtils = require('../core/testUtils');
 
-var sandboxHeight = '500',
-    sandboxWidth = '700',
-    displaySandboxId = 'display-sandbox',
-    displaySandboxSelector = '#' + displaySandboxId;
-
-var SandboxRegion = Marionette.Region.extend({
-    el: displaySandboxSelector
-});
+var sandboxId = 'sandbox',
+    sandboxSelector = '#' + sandboxId,
+    SandboxRegion = Marionette.Region.extend({
+        el: sandboxSelector
+    });
 
 describe('Analyze', function() {
+    before(function() {
+        if ($(sandboxSelector).length === 0) {
+            $('<div>', {id: sandboxId}).appendTo('body');
+        }
+    });
+
     beforeEach(function() {
-        $(displaySandboxSelector).remove();
-        // Use a special sandbox so that we can test responsiveness of chart.
-        $('body').append(sandboxTemplate.render({height: sandboxHeight, width: sandboxWidth}));
+
     });
 
     afterEach(function() {
+        $(sandboxSelector).remove();
+        $('<div>', {id: sandboxId}).appendTo('body');
         testUtils.resetApp(App);
-        $(displaySandboxSelector).remove();
     });
 
     describe('DetailsView', function() {
@@ -50,10 +51,6 @@ describe('Analyze', function() {
             var dataSet = analyzeDataSets[dataSetInd];
             it('renders tables that match the data when there are ' + dataSetInd + ' categories in the dataset', function(done) {
                 setupViewAndTest(dataSet, checkTable, done);
-            });
-
-            it('renders charts that match the data when there are ' + dataSetInd + ' categories in the dataset', function(done) {
-                setupViewAndTest(dataSet, checkChart, done);
             });
         }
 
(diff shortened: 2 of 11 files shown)


lewfish:
I just realized I need to test this on Safari and IE, and check that this PR covers #853, but you can starting reviewing if you want.

lliss:
If for some reason the modeling results fail to return (happens often depending on AOI size) we get a type error and the application fails to work properly from this point on:

```TypeError: chart.update is not a function```

**Comment by lliss** [comment]
Re: #853 
 * Not a firefox and safari issue.
 * Also still an issue with this branch...
screenshot-model my watershed - mozilla firefox

**Comment by lewfish** [comment]
I just need to bump up the left margin on that chart or change the axis format. Should be easy. I never tested it on such a big region...

**Comment by lewfish** [comment]
New commit to fix Compare/quality issue: https://github.com/lewfish/model-my-watershed/commit/a81251bbd8327e35f8a10caf759a4c005a6b0f72 

screen shot 2015-11-13 at 8 56 41 am 2

**Comment by lewfish** [comment]
> If for some reason the modeling results fail to return (happens often depending on AOI size) we get a type error and the application fails to work properly from this point on:
> 
> TypeError: chart.update is not a function

I'm unable to replicate this bug. I've set the polling timeout to 1000 in TaskModel which causes polling to fail with message 'Failed to get modeling results', but I don't see any other error messages.

**Comment by rajadain** [comment]
@azavea-bot rebuild

**Comment by rajadain** [comment]
I like that hovering on the coverage graph darkens the color subtly (instead of the horrid teal from before) and also shows the numerical value of the max:

image

The labels have more vertical spacing and are much easier to read.

**Comment by rajadain** [comment]
Observation noted about the inability to order the legend in the chart. I do like that it is an interactive legend and allows me to focus on certain aspects:

image

The label for 100% Forest is a little off, though, using Chrome on Mac.

**Comment by rajadain** [comment]
The right x-axis label on the Water Quality chart is a little clipped

image

**Comment by rajadain** [comment]
Maybe as part of a separate PR, unless it is pretty straightforward, but filtering the data on one chart in the Compare view should filter on all charts, otherwise the comparison becomes meaningless:

image

**Comment by lewfish** [comment]
> The label for 100% Forest is a little off, though, using Chrome on Mac.
This is because the labels are staggered labels to prevent overlap. It would be nice to turn it off though when there is a lot of room.

> The right x-axis label on the Water Quality chart is a little clipped
I'll add abbreviatedTicks to that axis.

**Comment by lewfish** [comment]
> Maybe as part of a separate PR, unless it is pretty straightforward, but filtering the data on one chart in the Compare view should filter on all charts, otherwise the comparison becomes meaningless:

I don't think it will be that easy to do that. I think it would be better to just disable filtering using the legend.

**Comment by rajadain** [comment]
>  it would be better to just disable filtering using the legend.

I agree.

**Comment by rajadain** [comment]
Here's the current graphs from staging:

mmw-graph-flash-1

And the ones from this PR:

mmw-graph-flash-2

The new ones are substantially fatter, and less saturated (which is okay given the hover behavior). There's also a flash of rendering (which unfortunately isn't always captured in the GIF, but is present on screen) as we switch scenarios in the new implementation that is absent in the old one. On one hand I don't really mind it because we're switching from once scenario to another and so redrawing the graph is expected and shouldn't surprise the user. On the other hand it feels like we have reduced performance, and almost gives the impression of the app feeling slower.

I don't really have a recommendation out of this – this is just an observation.

**Comment by rajadain** [comment]
There are some linting issues in the JavaScript:

```
$ ./scripts/npm.sh run lint
+ ARGS='run lint'

js/src/analyze/tests.js: line 14, col 5, 'sandboxTemplate' is defined but never used.

js/src/core/chart.js: line 19, col 50, Missing semicolon.
js/src/core/chart.js: line 96, col 48, Missing semicolon.
js/src/core/chart.js: line 163, col 91, Expected an assignment or function call and instead saw an expression.
js/src/core/chart.js: line 172, col 48, Missing semicolon.
js/src/core/chart.js: line 184, col 13, Missing semicolon.

js/src/core/tests.js: line 18, col 5, 'sandboxTemplate' is defined but never used.

js/src/modeling/tr55/quality/views.js: line 158, col 13, Expected an assignment or function call and instead saw an expression.

8 errors
```

rajadain:
The above flash of the chart when switching scenarios actually doesn't happen in Safari, where it is completely smooth!

rajadain:
The following use case breaks:

 1. Go to home page of MMW http://localhost:8000/ and log in
 2. Use the My Projects button in the user dropdown to load an existing project
 3. Click the Compare view button
 4. Things break:

image

Also happens if you try to go back to Analyze view instead.

rajadain:
I'll take another look once these issues have been fixed.

lewfish:
> Consider using prop instead of attr http://stackoverflow.com/questions/5874652/prop-vs-attr

In this case, setting prop className does not work. I'm guessing it's because it's an SVG.

lewfish:
Added some commits to address the feedback. I was unable to disable the legend filter and disable the staggering when it is unnecessary. These will require patching the library, and I would like to postpone this to another PR. Also, the bar width will be addressed in a pre-existing issue https://github.com/WikiWatershed/model-my-watershed/issues/1011

lewfish:
I'm also not sure what to do about the chart flashing when switching tabs in Chrome.

rajadain:
Taking another look

rajadain:
+1 tested, works great. The error scenario described above no longer happens.

I agree with fixing the couple stragglers later on. The work here is a significant improvement and I love the new UX of all charts.

Splendid job!

lewfish:
New issues spawned by this PR:
* #1021 
* #1022 
* #1023

Issue #1008, opened 2015-11-05
Add popup for observation assets


mmcfarland:
Based on #1007, clicking on an asset will show a list of variables available for the asset and be able to select a variable to plot its time series.  This will need maybe a little design from @jfrankl prior to implementation.

PR #1018, opened 2015-11-12
Add observation popups

mmcfarland:
This is very unstyled because @jfrankl is working on an ultimate design/layout style.  This merely get's the mechanics in place, the ultimate flow will be added later.

* Observation Popups have basic details and a listing of all variables measured at that station, with most recent values
* Clicking a variable will open a detail view modal with a timeseries plot of 2 months of data

Connects #1008 

**This is getting merged into the `feature/observation` branch on `origin` to avoid being deployed until it is ready.**

To Test:
1.  Add an observation layer to the map via the Layer Selector (USGS adds some points in Philly)
1. Click on an asset marker to see the popup.
1. Click on any `li` measurement item
1. See a modal appear plotting a few months worth of observation data for that asset and measurement


mmcfarland:
screenshot from 2015-11-12 14 25 06

screenshot from 2015-11-12 14 25 17

jfrankl:
This is somewhat superficial but will we be able to remove that Highcharts.com label?

mmcfarland:
Yeah, we might just have to add a css directive to hide it.

mmcfarland:
@azavea-bot rebuild

lewfish:
I think I found a bug.

screen shot 2015-11-13 at 2 31 47 pm 2
screen shot 2015-11-13 at 2 31 47 pm

mmcfarland:
@lewfish Looks like the api returns 200 but `success: false` for cases where it fails to load the sensor data.  I now check the flag and give a warning on the dialog (again, this is more or less a placeholder for real content) (306db2c)
screenshot from 2015-11-13 16 23 32

lewfish:
Sometimes the background turns black and the modal looks different. Here's an example:

screen shot 2015-11-13 at 4 29 14 pm 2
screen shot 2015-11-13 at 4 29 32 pm 2

mmcfarland:
@lewfish Any other details to get that case to happen?  I'm loading that modal/plot fine:

screenshot from 2015-11-13 16 41 10

lewfish:
Now it's looking like it only happens when double-clicking on an item, or clicking an item once, and then clicking another item before the modal loads. I think I also got it to happen intermittently when single clicking a lone item, but I'm not able to replicate that now.

lewfish:
The code looks good to me.

mmcfarland:
Good catch, I'll try to guard against double clicks to prevent that behavior.

mmcfarland:
Updated with 7c9c97a to not get the double overlay if double clicking.  There can be a flash of the first modal, might try to get rid of that in a styling phase.  Not sure if clicking the `li` will actually remain the way the modal is launched.

lewfish:
I'm thinking that instead of solving the double-click problem with https://github.com/WikiWatershed/model-my-watershed/commit/7c9c97a5c4fd71877e1ca56cbb49fe02f41c38b9, you could open the popup immediately and have a message or spinner to indicate that the data is loading. I'm suggesting this because it would also handle another issue which is that on a slow network connection, when the user clicks a list item, it can take a long time to get the data and show the popup which could be confusing.

lewfish:
I came across the situation shown in the screenshot below, but am unable to replicate it. I thought maybe it was because the data was still loading, and I tried to simulated this using a throttled network connection, but that just causes a delay in the popup appearing, as described in the previous comment.

screen shot 2015-11-16 at 11 01 50 am 2

mmcfarland:
I'm actually going to take care of this in #1020 which essentially follows what you described.  The modal will be responsible for loading the plot data, so it should open immediately and indicate it's loading.  This is to allow for a dropdown on the modal to switch between different variables, but should fix that issue as a side effect.

lewfish:
+1, tested

mmcfarland:
@azavea-bot rebuild

mmcfarland:
@azavea-bot rebuild

Issue #1024, opened 2015-11-17
Compare view scenarios not showing in ITSI snapshot


arobbi7:
The snapshot works perfectly in compare view in Safari and Firefox, but on Chrome (Mac), only the first scenario renders:
![screen shot 2015-11-17 at 3 29 05 pm](https://cloud.githubusercontent.com/assets/12868342/11224242/f99f2b12-8d41-11e5-915b-57ca0b5b70cd.png)

rajadain:
It's actually broken on all browsers. Was probably a fluke that it worked on Firefox and Safari earlier.

PR #1025, opened 2015-11-17
Fix screenshots of compare view

rajadain:
## Overview

Because the CSS selector wasn't specific enough, it would set the width of each `.map-container` in the compare view to 100%. By specifically targeting the one immediately below the body, we ensure that the views render correctly.

## Testing Instructions

If you would like to test, follow the instructions in #809.

## Demo

image

Connects #1024

### DIFF
--- a/src/mmw/js/src/app.js
+++ b/src/mmw/js/src/app.js
@@ -105,13 +105,13 @@ function initializeShutterbug() {
     $(window)
         .on('shutterbug-saycheese', function() {
             // Set fixed width before screenshot to constrain width to viewport
-            $('#model-output-wrapper, .map-container').css({
+            $('#model-output-wrapper, body > .map-container').css({
                 'width': window.innerWidth
             });
         })
         .on('shutterbug-asyouwere', function() {
             // Reset after screenshot has been taken
-            $('#model-output-wrapper, .map-container').css({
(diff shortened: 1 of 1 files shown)


mmcfarland:
The additional columns are now showing up:

screenshot from 2015-11-18 09 52 21

but the screenshot tries to fit all of the columns in on the shot - not sure if this was the behavior from before or is related to this changeset.  Also not sure what the expected behavior should be, since we can't fit them all in at once on the actual screen.

https://ccshutterbugtest.s3.amazonaws.com/293cdeef199.png

rajadain:
I'm not sure if we've tested more than three scenarios before, I can try and do that on an older commit (although I'll have to recreate all the VMs because of recent migrations which would be incompatible with old code).

The fundamental issue with the Compare View is that it is all designed with `vh` and `vw` units, which are unfortunately not supported in PhantomJS, used to take the screenshots. We may end up having to redesign that page so that it can function without those viewport relative units, as requested in #808.

I think the expected behavior should be to capture exactly what is shown on the screen. In case of multiple scenarios, the user can scroll until they have the desired set on screen, and then take a snapshot.

rajadain:
I've added the multiple column issue to #808 to be addressed at a later time. Merging this in for now.

Issue #1020, opened 2015-11-16
Updates to observation popups for styling


mmcfarland:
To align with the new designs for the observation popups:

* Sparkline of some recent values per measurement (not necessary)
* Historical Values button to launch the plot view
* Selection mechanism on plot view to cylce through measurements

mmcfarland:
@jfrankl I might not be finishing the task myself, can you attach a screenshot of the popup and modal so whoever picks it up has a frame of reference?

jfrankl:
Default view:

![image](https://cloud.githubusercontent.com/assets/1809908/11280333/8d9893f8-8ec3-11e5-88f9-a57244ff3768.png)

Show elevation on hover:

![image](https://cloud.githubusercontent.com/assets/1809908/11280340/95a36f28-8ec3-11e5-8a3d-e53ee8d0b9f4.png)

Show historical data in modal when clicking on "View historical data":

![image](https://cloud.githubusercontent.com/assets/1809908/11280353/a9ca3a54-8ec3-11e5-8053-cb5c8cfe511f.png)

mmcfarland:
A WIP commit that moves toward this design:
https://github.com/WikiWatershed/model-my-watershed/commit/6bbc261a41b9e9cb1dd0287cd3472898a20e3771

Specifically, opening the modal immediately and then having the timeseries requests happen from there to support the measurement dropdown pictured above.

PR #1038, opened 2015-11-25
Style observation popup and modal

lewfish:
This PR changes the style of the observations popups and plot modals according to wireframes in #1020, with the following exceptions. Note that only the final three commits were added by this PR, and the others are from the feature/observations branch.
* I omitted the sparklines in the popups because the data for each measurement takes a few seconds to load, and the scale and unit for each measurement is different which means the lines would be useless for comparison.
* I also omitted the last updated time on the popup because each measurement has a different time stamp (but implemented it for the plot modal since it only shows one measurement at a time). 
* I attempted to implement the hovering behavior for showing the depth of measurements, but there was a bug (only present in Chrome) which I could not fix after much effort, so I just put that info in a third column. I noticed that the depth is always 0.9m in the data sources we're using which is probably erroneous. Maybe we should not display this info?
* The CSS will be fine-tuned by @jfrankl in a separate commit later in the sprint to get the layout, fonts, and button icon to look good.

screen shot 2015-11-25 at 10 43 50 am 2

screen shot 2015-11-25 at 10 44 01 am 2

screen shot 2015-11-25 at 11 00 22 am 2

##### Testing
* Click on an observation marker.
* Click on a row which should open the plot modal with that measurement selected.
* A spinner should appear while loading.
* Selecting from the dropdown should load another plot. Selecting a new measurement while loading is in progress should work.
* Selecting a previously displayed measurement from the dropdown should not require loading since the data is cached.
* Closing the modal and reopening should work as expected.
* Clicking the 'Historical Data' button should open the plot modal with the first element in the dropdown selected.

Connects #1020


rajadain:
Took me a while to restore my setup. Taking a look now.

rajadain:
For the sparklines, the time it takes to load the data is a legitimate concern. There would still be value in seeing trendlines, see what trends up and down, and how they correspond with each other. But yes, in the face of slow, non-normalized values, this is something we should hold off on.

rajadain:
We may need more information from the data source before making this decision, but if the value of `depth` is the same for all measurements perhaps we can show it once instead of for each item:

image

rajadain:
In the CSBCZO Observation dataset, I'm only getting values in the blue circle, not the brown or yellow ones:

image

Do we know why that is? Also, if there is no immediate data available, as in the above screenshot, perhaps we can hide the "Historical Data" button, clicking which currently does nothing.

rajadain:
Consider moving the field selection drop-down into the modal header to better use horizontal space and save vertical space:

image

rajadain:
Is this value of Wind Direction correct? **288.9 deg (from)**. Seems like it is missing the direction.

image

lewfish:
> Is this value of Wind Direction correct? 288.9 deg (from). Seems like it is missing the direction.

Yes, I think it's saying that the wind is coming from 288.9 degrees (as in degrees around a circle).

rajadain:
Ah got it.

Since the "Loading Data" and spinner icon will always be shown together, consider moving them to one line instead of two:

image

rajadain:
We should probably have a timeout for loading the data. Also, when we receive a `success: false` or error response, we should show an error icon and message. Currently the spinner spins forever:

image

rajadain:
Although you _can_ click outside to close it, all our other modals have buttons on them that let you close the modal. Consider adding a Close button in the top-right corner to close this modal.

image

rajadain:
Will take another look once some of these issues have been addressed.

lewfish:
Added a commit to address comments.

rajadain:
+1 tested all use cases mentioned above and they have been fixed.

I still think that the dropdown can be slightly higher, but perhaps that's something @jfrankl can look in to when he's doing design fixes.

image

Will have a look at the code Monday morning, but functionally it looks good to me!

rajadain:
I cannot open the select box if I click in the lower half, only in the upper half:

mmw-observation-popup-dropdown-noclick

lewfish:
The dropdown was only clickable on the top half because it was inside of a div with class=pull-right. For now I just pulled the dropdown out of that div. Hopefully @jfrankl can find a way to position the dropdown without causing that problem to occur.

Issue #1043, opened 2015-12-01
Observation popup external link should target new tab


mmcfarland:
Clicking the provider URL should open a new tab and not load up in the current tab.  Currently it destroys the state you have.

PR #1047, opened 2015-12-02
Open new tab for observation links

lewfish:
Clicking one of the links in the observation popups should open a new tab.

Connects #1043


rajadain:
+1 tested.

Issue #995, opened 2015-11-05
Water quality load units are crowded/overlapping in Model view graph


arobbi7:
When the load numbers are large enough, the x-axis interval numbers start to get really crowded and overlap on the water quality bar graph in Model view.

Observed in Chrome, Firefox, and Safari on 13" Mac, and in IE 11 on Windows.

This project can be used to replicate: https://mmw-dev.azavea.com/project/443/scenario/882

![screen shot 2015-11-05 at 10 24 46 am](https://cloud.githubusercontent.com/assets/12868342/10972631/1a24e17c-83a8-11e5-80c7-652bfebaeaf9.png)

jamesmcclain:
Who do I need to login as in order to be able to see this project?

jamesmcclain:
This issue has already been solved.

lewfish:
There are still some situations that result in crowded labels. I think to fix this you would have to change the threshold at which NVD3 changes the number of tick marks. 

![screen shot 2015-12-02 at 11 31 49 am 2](https://cloud.githubusercontent.com/assets/1896461/11536898/7f155f94-98e8-11e5-80a5-78d7514f5a4f.png)

jamesmcclain:
Should I reopen this, then?

PR #1050, opened 2015-12-02
Fix Squashed Labels on Horizontal Charts

jamesmcclain:
This is a fix for squashed labels which sometimes occur in horizontal charts.  The attempt is to make sure that each tick on the y-axis is given at least 60 pixels.

Connects #995

## Note ##

A much more obvious solution would have been something like:

```diff
diff --git a/src/mmw/js/src/core/chart.js b/src/mmw/js/src/core/chart.js
index 510b892..7f74108 100644
--- a/src/mmw/js/src/core/chart.js
+++ b/src/mmw/js/src/core/chart.js
@@ -117,7 +117,9 @@ function renderHorizontalBarChart(chartEl, data, options) {
 
         setChartHeight();
         chart.tooltip.enabled(false);
-        chart.yAxis.ticks(5);
+        chart.yAxis
+            .axisLabelDistance(90)
+            .ticks(5);
         handleCommonOptions(chart, options);
 
         d3.select(svg)
```

but unfortunately, that does not work.

## To Test ##

Load the application and find a horizontal bar chart.  Try stretching and compression the browser horizontally.  Previously, squashing of the labels on the horizontal axis could be observed.

screenshot from 2015-12-02 15 46 23

screenshot from 2015-12-02 15 46 32

screenshot from 2015-12-02 15 46 41

screenshot from 2015-12-02 15 46 52

screenshot from 2015-12-02 15 47 01

screenshot from 2015-12-02 15 47 13

### DIFF
--- a/src/mmw/js/src/core/chart.js
+++ b/src/mmw/js/src/core/chart.js
@@ -51,7 +51,8 @@ function getNumBars(data) {
 */
 function renderHorizontalBarChart(chartEl, data, options) {
     var chart = nv.models.multiBarHorizontalChart(),
-        svg = makeSvg(chartEl);
+        svg = makeSvg(chartEl),
+        $svg = $(svg);
 
     // Augment data so that it is part of a single series to match
     // the format expected by NVD3.
@@ -84,7 +85,7 @@ function renderHorizontalBarChart(chartEl, data, options) {
         var numBars = getNumBars(data),
             maxHeight = options.margin.top + options.margin.bottom +
                         numBars * options.maxBarHeight,
-            actualHeight = $(svg).height();
+            actualHeight = $svg.height();
 
         if (actualHeight > maxHeight) {
             chart.height(maxHeight);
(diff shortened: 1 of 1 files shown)


src/mmw/js/src/core/chart.js
```diff
@@ -94,8 +95,13 @@ function renderHorizontalBarChart(chartEl, data, options) {
     }
 
     function updateChart() {
-        if($(svg).is(':visible')) {
+        var width = $svg.width(),
+            availableWidth = width - 240,
```
lewfish: It seems like the constant 240 should actually be a variable that changes as a function of the margins.

jamesmcclain: Good call.

src/mmw/js/src/core/chart.js
```diff
@@ -94,8 +95,13 @@ function renderHorizontalBarChart(chartEl, data, options) {
...
+            yticks = Math.floor(availableWidth / 60);
```
lewfish: Consider extracting 60 to a variable called minLabelWidth or similar.

lewfish:
+1, tested.

src/mmw/js/src/core/chart.js
jamesmcclain: Okay, will-do

jamesmcclain:
Thanks for reviewing, merging now ...

PR #1053, opened 2015-12-04
Style observations popups

lewfish:
Made some changes to improve the style of the observation popups/modals. Things should now look like:

screen shot 2015-12-07 at 2 31 41 pm 2

screen shot 2015-12-07 at 2 31 55 pm 2

### DIFF
--- a/src/mmw/js/src/core/modals/templates/plotModal.html
+++ b/src/mmw/js/src/core/modals/templates/plotModal.html
@@ -2,10 +2,9 @@
     <div class="modal-content plot-modal">
         <div class="modal-header">
             <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
-            <div class="measurements-platform"> {{ platform_type }} </div>
-            <div class="measurements-name"> {{ measurement.name }} </div>
+            <div class="measurements-platform">{{ name }}, {{ platform_type }}</div>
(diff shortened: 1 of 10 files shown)


lewfish:
@jfrankl: You can create a branch off of this, make a new commit with your changes, and then push it to your fork. Then I'll cherry-pick that commit back into my branch.

lewfish:
Pushed a commit to show the most recently updated measurement.

rajadain:
Looks very nice! Great job both of you!

It would be useful to add the units in the detail modal though:

image

I don't know what 1030 means.

rajadain:
Oh I see the units if I hover over the graph:

image

It would still be useful to have them somewhere without hovering. Also, what is with the numeric spacing in the tooltip? It seems to be using a European standard of " " instead of "," as digit separator. Can we configure that to be comma instead of a space?

rajadain:
How do the "last updated" timestamps work? I'm getting "16 hours ago" in the map marker, but "25 minutes ago" in all the detail modals, in this above example:

image

Also, can we consider moving the linked source of the data in "Data from NOS/CO-OPS" into the next line? The height of the "Historical Data" button should allow that, and we'd be less likely to encounter a line-break. Like this:

image

I just added a `<br />` before the `a` tag.

lewfish:
Good catches. I fixed things and will merge shortly.

Issue #1058, opened 2015-12-07
Stub out worker task and UI for RWD


lewfish:
This is mostly done.

PR #1061, opened 2015-12-08
Stub out RWD

lewfish:
This PR stubs out RWD. To test, click on 'Hill Slope' in the 'Delineate Watershed' dropdown and then click a point on the map. This will make a spinner visible next to 'Hill Slope', and will poll the backend, which should return a diamond shaped polygon. The other menu items should work as before (creating random polygons on the frontend). If you disable your network connection before clicking 'Hill Slope', you should see an error icon instead of a spinner, which will be removed if you make another selection from the dropdown. 

I'm not sure how to display the actual error message. We could associate a tooltip with the error icon, but that might not be obvious enough to the user.

Connects #1058

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -6,7 +6,6 @@ var $ = require('jquery'),
     Marionette = require('../../shim/backbone.marionette'),
     turfArea = require('turf-area'),
     turfBboxPolygon = require('turf-bbox-polygon'),
-    turfBuffer = require('turf-buffer'),
     turfDestination = require('turf-destination'),
     router = require('../router').router,
     App = require('../app'),
@@ -351,7 +350,6 @@ var PlaceMarkerView = Marionette.ItemView.extend({
         var self = this,
             map = App.getLeafletMap(),
             itemName = $(e.currentTarget).text(),
-            itemType = $(e.currentTarget).data('shape-type'),
             revertLayer = clearAoiLayer();
 
         this.model.set('pollError', false);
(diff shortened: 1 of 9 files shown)


src/mmw/js/src/draw/views.js
```diff
@@ -338,24 +340,77 @@ var PlaceMarkerView = Marionette.ItemView.extend({
...
+                            point.geometry.coordinates[1],
+                            point.geometry.coordinates[0]
+                        ])}
+                    };
+
+                    self.rwdTaskModel.start(taskHelper);
+                    return deferred;
+                } else {
                     shape = turfRandom('polygons', 1, {
                         max_vertcies: 50,
```
kdeloach: Unrelated, but I just noticed there's a typo on this line `max_vertcies`. Probably do not need turf-random anymore though.

src/mmw/js/src/modeling/models.js
```diff
@@ -40,6 +40,16 @@ var Tr55TaskModel = coreModels.TaskModel.extend({
...
 
+var RwdTaskModel = coreModels.TaskModel.extend({
+    defaults: _.extend(
+        {
+            taskName: 'rwd',
+            taskType: 'modeling'
+        },
+        coreModels.TaskModel.prototype.defaults
+    )
+});
```
kdeloach: Looks like this model is not exported. It's also defined in `src/mmw/js/src/draw/models.js`.

src/mmw/js/src/draw/views.js
kdeloach: Oh right, there's other delineation models that still use a random polygon. Ignore that remark.

kdeloach:
Very cool. One question: Should RWD be activated by clicking the "Stream Network" option instead of the "Hill Slope" option?

kdeloach:
+1 Tested

mmcfarland:
Here's a brief description I got for what each of the 3 "Delineate Watershed" options will/should do:

* Trace Flow Down:: Interactive visualization of blue line to the ocean
* Hill Slope:: Just local operation of Stream Network not snapped to blue line
* Stream Network:: Rapid Watershed Deliniation snap to blue line and compute contributing area

lewfish:
@azaveaci rebuild

Issue #1032, opened 2015-11-19
Ingest 30m gSSURGO and 30m NLCD to datahub


mmcfarland:
30m gSSURGO is on @chitchens external drive.  This needs to be tiled, reprojected and ingested (see @notthatbreezy for script)

30m NLCD _may_ already be ingested, but I think at the wrong projection

mmcfarland:
This will include a pairing session to go over the process.

mmcfarland:
Also get painted display tiles (no data == transparent)

PR #1062, opened 2015-12-16
Move to 30m Soil and NLCD tiles

mmcfarland:
* Point application to the new 30m EPSG:5070 soil and nlcd tiles.
* Update pixel / sqkm calculation
* Update size of bounding area allowed

Test:
Restart your worker or `debugcelery` to get updated tasks and check that modeling analyze/modeling results are found
* Test that larger areas (HUC8) return results

The max area will probably have to be tweaked.  Our current office network issues are making it hard to tell what is an acceptable limit.  We may also want to dial back down the polling time, but I think we should play with it in dev before making that decision.

Connects #1032 
Connects https://github.com/WikiWatershed/mmw-geoprocessing/issues/17

### DIFF
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -101,13 +101,14 @@ def aoi_resolution(area_of_interest):
     average_lat = reduce(lambda total, p: total+p[1], pairs, 0) / len(pairs)
 
     max_lat = 48.7
-    max_lat_count = 10025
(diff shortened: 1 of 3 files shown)


mmcfarland:
The area calculation is more variable than in the 10m version, I believe this is due to the larger cell size, but I generated the pixel count ramp in the same manner as it was done previously.

lewfish:
+1

Issue #998, opened 2015-11-05
Bring Microsite numbers in front of arrows on small screens


arobbi7:
On really small screens and windows, the runoff and infiltration numbers can become obscured by their respective arrows or by the water column graphic. We don't need to support a full rendering of the app on screens/windows this small, but we do want to prioritize the numbers over the graphics. 

![screen shot 2015-11-05 at 11 40 45 am](https://cloud.githubusercontent.com/assets/12868342/10974941/5230f47a-83b2-11e5-84bb-5991d9b3d406.png)
![screen shot 2015-11-05 at 11 40 54 am](https://cloud.githubusercontent.com/assets/12868342/10974943/53ad3eee-83b2-11e5-8874-0d57b86d2c4b.png)

PR #1063, opened 2015-12-16
Ensure that micro site numbers are always on top

rajadain:
## Overview

Previously the micro site numbers would be obscured on small screens. This change ensures that they are on top of the effect arrows as well as the water column.

## Testing Instructions

Open the micro site and make the browser window really small.

## Demo

mmw-small-micro-site

Connects #998


lewfish:
+1, tested

rajadain:
Thanks!

Issue #1068, opened 2015-12-17
Make RWD geodata available to script


mmcfarland:
This may include:

* S3QL
* EBS Volume Snapshot restored on Worker
* Configuring the RWD code to point to the correct mounted path

kdeloach:
Uploaded all shapefiles to S3 (located at s3://rwd-data) and, although RWD executed successfully using s3ql, I wasn't able to figure out an easy way to bypass the one-machine-at-a-time mounting limitation. Now I'm creating a volume snapshot containing all the necessary shapefiles.

PR #1073, opened 2016-01-04
Add RWD data volume snapshot to worker instances

kdeloach:
Connects #1068

### DIFF
--- a/deployment/cfn/worker.py
+++ b/deployment/cfn/worker.py
@@ -339,6 +339,9 @@ class Worker(StackNode):
     def get_cloud_config(self):
         return ['#cloud-config\n',
                 '\n',
+                'mounts:\n',
+                '  - [xvdf, /opt/rwd-data, ext4, "defaults,nofail,discard", 0, 2]\n'
+                '\n',
                 'write_files:\n',
                 '  - path: /etc/mmw.d/env/MMW_STACK_COLOR\n',
                 '    permissions: 0750\n',
--- a/deployment/packer/template.js
+++ b/deployment/packer/template.js
@@ -62,6 +62,14 @@
             "instance_type": "t2.large",
             "ssh_username": "ubuntu",
             "ami_name": "mmw-worker-{{timestamp}}-{{user `version`}}",
+            "ami_block_device_mappings": [
+                {
+                    "device_name": "/dev/sdf",
+                    "snapshot_id": "snap-3d8f22ab",
+                    "volume_type": "gp2",
+                    "delete_on_termination": true
+                }
+            ],
             "run_tags": {
                 "PackerBuilder": "amazon-ebs"
             },


kdeloach:
Updated cloud config to mount block device.

kdeloach:
Fixed the cloud config.

deployment/cfn/worker.py
```diff
@@ -339,6 +339,9 @@ def create_auto_scaling_resources(self, worker_security_group, worker_lb):
     def get_cloud_config(self):
         return ['#cloud-config\n',
                 '\n',
+                'mounts:\n',
+                '  - [xvdf, /mnt/rwd-data, "auto", "defaults"]\n'
```
hectcastro: I think that explicitly setting the following arguments is important:

```yaml
- [xvdf, /mnt/rwd-data, ext4, defaults,nofail,discard, 0, 2
```

Here `nofail` turns into `_netdev` (when mounted), which signals to the OS that this file system depends on the network and `discard` enables TRIM support for `gp2` volumes.

The `0` is to signal that the file system shouldn't be `dump(8)`'d and the `2` is for all non-root file systems (root file systems get `1`).

See also: `man fstab`

kdeloach: Thanks, will give this a try now.

deployment/packer/template.js
```diff
@@ -62,6 +62,14 @@
             "instance_type": "t2.large",
             "ssh_username": "ubuntu",
             "ami_name": "mmw-worker-{{timestamp}}-{{user `version`}}",
+            "ami_block_device_mappings": [
+                {
+                    "device_name": "/dev/sdf",
+                    "snapshot_id": "snap-7bf80133",
```
hectcastro: I think that we could copy this snapshot in order to rename it, and then ensure that it is tagged for the `Production` environment, so that we never delete it.

hectcastro: I created a newly labeled snapshot based on the previous. Its ID is `snap-3d8f22ab`.

kdeloach: Updated.

hectcastro:
Launched a new `Worker` stack using `ami-49bae523` and everything worked:

```bash
ubuntu@ip-10-0-3-165:~$ df -h
Filesystem      Size  Used Avail Use% Mounted on
udev            988M   12K  988M   1% /dev
tmpfs           201M  376K  200M   1% /run
/dev/xvda1      7.8G  3.1G  4.3G  42% /
none            4.0K     0  4.0K   0% /sys/fs/cgroup
none            5.0M     0  5.0M   0% /run/lock
none           1001M  120K 1001M   1% /run/shm
none            100M     0  100M   0% /run/user
/dev/xvdf       126G   89G   32G  74% /opt/rwd-data
ubuntu@ip-10-0-3-165:~$ ls -l /opt/rwd-data/
total 292
drwx------    2 root root  16384 Dec 30 21:45 lost+found
drwxr-xr-x    2 root root   4096 Jan  4 14:37 Main_Watershed
drwxr-xr-x 8313 root root 274432 Jan  4 14:38 Subwatershed_ALL
```

:+1:

kdeloach:
Thanks.

PR #1074, opened 2016-01-05
Add RWD service to worker

kdeloach:
To test:
 * Ensure you have the RWD shapefiles mounted at `/opt/rwd-data`.

 * Reload & reprovision worker VM.

```vagrant reload worker --provision```

 * SSH to the worker VM and verify the service returns results.

```curl http://localhost:5000/rwd/-75.276639/39.892986```

Connects #1070

### DIFF
--- a/deployment/ansible/roles/model-my-watershed.rwd/defaults/main.yml
+++ b/deployment/ansible/roles/model-my-watershed.rwd/defaults/main.yml
@@ -1,0 +1,5 @@
+---
+rwd_data_path: "/opt/rwd-data"
+rwd_host: "localhost"
+rwd_port: 5000
+rwd_docker_image: "quay.io/wikiwatershed/rwd:0.2.0"
--- a/deployment/ansible/roles/model-my-watershed.rwd/tasks/main.yml
+++ b/deployment/ansible/roles/model-my-watershed.rwd/tasks/main.yml
@@ -1,0 +1,12 @@
+---
+- name: Pull docker container image
+  command: /usr/bin/docker pull {{ rwd_docker_image }}
+
+- name: Configure upstart service
+  template: src=upstart-mmw-rwd.conf.j2
+            dest=/etc/init/mmw-rwd.conf
+  notify:
+    - Restart mmw-rwd
+
+- name: Ensure service is running
+  service: name=mmw-rwd state=started
--- a/deployment/ansible/roles/model-my-watershed.rwd/templates/upstart-mmw-rwd.conf.j2
+++ b/deployment/ansible/roles/model-my-watershed.rwd/templates/upstart-mmw-rwd.conf.j2
@@ -1,0 +1,32 @@
+description "Rapid Watershed Delineation"
+
+{% if ['development', 'test'] | some_are_in(group_names) -%}
+start on (vagrant-mounted)
+{% else %}
+start on (filesystem and started docker)
+{% endif %}
+stop on stopping docker
+
+kill timeout 20
+kill signal CONT
+respawn
+
+pre-start script
+  /usr/bin/docker kill rwd || true
+  /usr/bin/docker rm rwd || true
+end script
+
+exec /usr/bin/docker run \
+  --name rwd \
+  --publish {{ rwd_port }}:5000 \
+  --volume {{ rwd_data_path }}:{{ rwd_data_path }} \
+  {{ rwd_docker_image }}
+
+post-start script
+  while ! nc -w0 {{ rwd_host }} {{ rwd_port }}; do sleep 1; done
+end script
+
+post-stop script
+  /usr/bin/docker kill rwd
+  /usr/bin/docker rm rwd
+end script
--- a/deployment/ansible/roles/model-my-watershed.rwd/templates/upstart-mmw-rwd.conf.j2
+++ b/deployment/ansible/roles/model-my-watershed.rwd/templates/upstart-mmw-rwd.conf.j2
@@ -19,7 +19,7 @@ end script
 exec /usr/bin/docker run \
   --name rwd \
   --publish {{ rwd_port }}:5000 \
-  --volume {{ rwd_data_path }}:{{ rwd_data_path }} \
+  --volume {{ rwd_data_path }}:{{ rwd_data_path }}:ro \
   {{ rwd_docker_image }}
 
 post-start script
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -93,6 +93,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     worker.vm.synced_folder ".", "/vagrant", disabled: true
     worker.vm.synced_folder "src/mmw", "/opt/app/"
 
+    if ENV["VAGRANT_ENV"].nil? || ENV["VAGRANT_ENV"] != "TEST"
+      worker.vm.synced_folder "/opt/rwd-data", "/opt/rwd-data"
+    end
+
     # Docker
     worker.vm.network "forwarded_port", {
       guest: 2375,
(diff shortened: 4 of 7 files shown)

### CODE REVIEW

**Comment by kdeloach** [comment]
This should mostly work but I have been experiencing problems with Vagrant / VirtualBox yet again.

File: deployment/ansible/roles/model-my-watershed.rwd/tasks/main.yml [hunk 1]
```diff
@@ -0,0 +1,17 @@
+---
+- stat: path={{ rwd_data_path }}
+  register: rwd_data
+- fail: msg="Data folder does not exist. Did you forget to mount /opt/rwd-data?"
```
hectcastro: It seems like this would fail during the AMI creation phase because `/opt/rwd-data` wouldn't exist.

Wouldn't Vagrant throw some kind of failure if one of the defined shares cannot be setup?

deployment/ansible/roles/model-my-watershed.rwd/templates/upstart-mmw-rwd.conf.j2
```diff
@@ -0,0 +1,28 @@
...
+kill signal CONT
+respawn
+
+pre-start script
+  /usr/bin/docker kill docker-rwd || true
+  /usr/bin/docker rm docker-rwd || true
+end script
+
+exec /usr/bin/docker run \
+  --name docker-rwd \
```
hectcastro: Nitpick, but I'd probably just call with `rwd` because the Docker part is kinda implied.

deployment/ansible/roles/model-my-watershed.rwd/tasks/main.yml
kdeloach: I thought we could use a synced folder for local development, and the block device should be mounted here during the packer job (see https://github.com/WikiWatershed/model-my-watershed/pull/1073/files#diff-deda55c5935c68f390c8525df625393dR343). Right?

deployment/ansible/roles/model-my-watershed.rwd/templates/upstart-mmw-rwd.conf.j2
kdeloach: Very true. Updated.

deployment/ansible/roles/model-my-watershed.rwd/tasks/main.yml
hectcastro: If we want the data available on the instance used to create an AMI, we'd have to add a `user_data` or `user_data_file` configuration to Packer. I'd lean toward avoiding that tight a coupling during AMI creation time, if we can.

See also: https://packer.io/docs/builders/amazon-ebs.html

kdeloach:
Going to hold off testing this until the latest docker image is released on quay.

deployment/ansible/roles/model-my-watershed.rwd/templates/upstart-mmw-rwd.conf.j2
```diff
@@ -0,0 +1,28 @@
+description "Rapid Watershed Delineation"
+
+start on (filesystem and started docker)
```
hectcastro: I think that `start on` is going to have to be expanded to something like:

```bash
{% if ['development', 'test'] | some_are_in(group_names) -%}
start on (vagrant-mounted)
{% else %}
start on (filesystem and started docker)
{% endif %}
```

Because we don't want Docker to try and start before the Vagrant share is accessible.

deployment/ansible/roles/model-my-watershed.rwd/templates/upstart-mmw-rwd.conf.j2
```diff
@@ -0,0 +1,28 @@
...
+
+pre-start script
+  /usr/bin/docker kill rwd || true
+  /usr/bin/docker rm rwd || true
+end script
+
+exec /usr/bin/docker run \
+  --name rwd \
+  --publish {{ rwd_port }}:{{ rwd_port }} \
+  --volume {{ rwd_data_path }}:{{ rwd_data_path }} \
```
hectcastro: Seems like we may have to guard this when in the test environment because `rwd_data_path` won't exist within the virtual machine.

kdeloach: I might just disable the RWD role in test mode.

deployment/ansible/roles/model-my-watershed.rwd/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+rwd_data_path: "/opt/rwd-data"
+rwd_host: "localhost"
+rwd_port: 8091
+rwd_docker_image: "quay.io/wikiwatershed/rwd:0.1.0"
```
kdeloach: This should be updated to 0.2.0 when the docker image is finished building.

deployment/ansible/roles/model-my-watershed.rwd/templates/upstart-mmw-rwd.conf.j2
```diff
@@ -0,0 +1,28 @@
...
+respawn
+
+pre-start script
+  /usr/bin/docker kill rwd || true
+  /usr/bin/docker rm rwd || true
+end script
+
+exec /usr/bin/docker run \
+  --name rwd \
+  --publish {{ rwd_port }}:{{ rwd_port }} \
```
lewfish: This should be `--publish {{ rwd_port }}:5000` because Flask server is running on port 5000.

kdeloach: I could set this variable to 5000 so the port used on the guest and host are the same.

kdeloach:
Updated with test instructions.

deployment/ansible/roles/model-my-watershed.rwd/templates/upstart-mmw-rwd.conf.j2
kdeloach: TODO: Hard-code this port.

lewfish:
When following the testing instructions, after running the curl command, I get `{error:''"}` as the response. If I run `sudo service restart mmw-rwd,` then I get the correct response. I also get the correct response if I  manually stop, rm, and start the rwd container. But running `vagrant reload worker` does not work.

lewfish:
I think this is because of the comment Hector made about "we don't want Docker to try and start before the Vagrant share is accessible."

kdeloach:
I still need to update the code to address those comments.

lewfish:
The code looks good aside from that issue.

deployment/ansible/roles/model-my-watershed.rwd/templates/upstart-mmw-rwd.conf.j2
kdeloach: It looks like Jenkins was able to start this service without encountering an error because of this line so I'm inclined to leave as-is for now.

hectcastro: This could also be because a previously unguarded share configuration made the path exist. May not hold true if the box got provisioned from scratch now.

Main thing I'm worried about is an attempt to create a volume with an underlying directory that doesn't exist.

kdeloach: This came as a surprise, but according to the docs...

> If 'host-dir' is missing, then docker automatically creates
       the new volume on the host. This auto-creation of the  host
       path has been deprecated in Release: v1.9.

PR #1075, opened 2016-01-05
Update deployment dependencies; remove unused Packer variables

hectcastro:
Light deployment cleanup by updating dependencies and removing variables that are now threaded into instances via cloud-config.


kdeloach:
+1

PR #1076, opened 2016-01-06
Add details about updating AMI environment tags

hectcastro:
In order to prevent from deleting production promoted AMIs housed in the staging AWS account, ensure that the `Environment` tag is updated to `Production`.


mmcfarland:
+1, good move.

Issue #1071, opened 2015-12-31
Wire in RWD service to UI workflow


mmcfarland:
Based on #1070, the service should be able to be wired into the WIP @lewfish created which includes UI elements and celery tasks.  The goal of this card is to complete a round trip request from the UI to displaying the watershed on the map as an AoI.

PR #1077, opened 2016-01-07
Wire in RWD service to UI workflow

lewfish:
This enables the RWD REST API to be called by the RWD Celery task, and the results to be visualized in the UI.

##### Testing
* Run `./scripts/debugcelery.sh` to make sure the worker is running the latest worker code.
* In the Delineate Watershed menu, click either of the options (they do the same thing for now).
* Click a point near the Delaware River. After polling, it should set the AOI to the computed watershed.
* Reset and then click a point outside the Delaware River Basin. It should result in an error icon and message printed to the console.
* Clicking Reset should work at the following points in time. The original functionality for the other parts of the ToolbarView should be preserved.
 * after clicking the menu item, but before clicking a point on the map
 * after a polling failure
 * while polling (it will cancel the job)
 * after the results are shown on the map

##### Known Issues
* We should probably display the error message to the user when there's a polling failure. Maybe show a popup on the map? Currently, we show an alert box when the user draws an AOI that is too big, which seems kind of clunky. \cc @mmcfarland 

Connects #1071 

Here is the watershed computed for a point in Philly.
screen shot 2016-01-07 at 1 05 03 pm 2
screen shot 2016-01-07 at 1 05 16 pm 2

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -20,7 +20,7 @@ var $ = require('jquery'),
     placeMarkerTmpl = require('./templates/placeMarker.html'),
     settings = require('../core/settings');
 
-var MAX_AREA = 112700; // About the size of a large state
+var MAX_AREA = 112700; // About the size of a large state (in km^2)
 var codeToLayer = {}; // code to layer mapping
 
 function actOnUI(datum, bool) {
@@ -356,7 +356,7 @@ var PlaceMarkerView = Marionette.ItemView.extend({
 
         this.model.disableTools();
         utils.placeMarker(map)
-            .then(function(latlng) {
+             .then(function(latlng) {
                 var point = L.marker(latlng).toGeoJSON(),
                     shape,
                     deferred = $.Deferred();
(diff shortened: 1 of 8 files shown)


mmcfarland:
My response from the server takes the form of an array or coordinates
> JSON.parse(response.result)
> Array [ Array[2], Array[2], Array[2], Array[2] ]

but the code is looking for a `features` property.  Does this indicate a part of my system is not up-to-date?

lewfish:
> Does this indicate a part of my system is not up-to-date?

I think you need to run `./debugcelery.sh` or restart the celery service.

src/mmw/js/src/draw/views.js
```diff
@@ -356,35 +359,35 @@ var PlaceMarkerView = Marionette.ItemView.extend({
 
         this.model.disableTools();
         utils.placeMarker(map)
-            .then(function(latlng) {
+             .then(function(latlng) {
```
mmcfarland: Extra space?

lewfish: This is to make the dot on this line align with the dot on the previous line.

mmcfarland:
Results look good!
screenshot from 2016-01-07 15 57 11

kdeloach:
I can review this today.

lewfish:
I'm making some changes now.

lewfish:
I reverted that refactoring change. What do you think about https://github.com/lewfish/model-my-watershed/commit/8ec66ef49c953f5cad7790bd62e2b7cc58063db6.

kdeloach:
+1

PR #1078, opened 2016-01-07
Fix libpq version reference

hectcastro:
The PostgreSQL APT repositories are preparing to release PostgreSQL 9.5, which is shifting several base packages in and out of the main repository channel.


kdeloach:
+1

PR #1082, opened 2016-01-12
Add Hydrologic Soil Group layer

lewfish:
This adds the Hydrologic Soil Group layer to the layer selector which references the tiles generated in https://github.com/azavea/azavea-data-hub/pull/21

screen shot 2016-01-12 at 10 28 13 am 2

### DIFF
--- a/src/mmw/mmw/settings/layer_settings.py
+++ b/src/mmw/mmw/settings/layer_settings.py
@@ -124,6 +124,23 @@ LAYERS = [
         'has_opacity_slider': True
     },
     {
+        'display': 'Hydrologic Soil Groups',
+        'short_display': 'SSURGO',
+        'helptext': 'Soils are classified by the Natural Resource Conservation '
+                    'Service into four Hydrologic Soil Groups based on the '
+                    'soil\'s runoff potential. The four Hydrologic Soils Groups'
+                    'are A, B, C and D. Where A\'s generally have the smallest '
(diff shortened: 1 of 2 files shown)


mmcfarland:
+1, looks good.

rajadain:
I haven't provisioned MMW in a while, apparently I need `/opt/rwd-data` to get the `worker` up. Is there any documentation for what needs to be in that folder?

kdeloach:
This can be an empty folder. You'll need the RWD shapefiles to run the rapid watershed delineation procedures though.

rajadain:
Okay. I would prefer for that folder to be a local one though. Creating an empty folder in my root just to appease one repo's Vagrant file feels contaminating.

rajadain:
There's a hard boundary in south western Pennsylvania:

image

but I guess that's a problem with the dataset, not with this front-end implementation.

rajadain:
On Chrome on Mac OS X, the "Hydrologic Soil Groups" option is out of the viewable area, and without the scrollbar I would've never known it was there had I not been looking for it specifically:

image

Consider making the layer selector box a little bit taller to accommodate this without having to scroll.

rajadain:
Tested and works well, apart from those two issues. Overzooming works as well. I love that our general setup allows us to add layers like this with configuration only. +1

lewfish:
That border around PA is pretty strange. I increased the height of the layer selector to show the SSURGO layer without scrolling.

PR #1088, opened 2016-01-13
Fix stray reference to latest tag of SJS image

hectcastro:
Don't use the `latest` tag to pull down the initial version of the SJS container image. Instead, use a specific version (0.6.1).

### DIFF
--- a/deployment/ansible/roles/model-my-watershed.spark-jobserver/tasks/main.yml
+++ b/deployment/ansible/roles/model-my-watershed.spark-jobserver/tasks/main.yml
@@ -37,7 +37,7 @@
     - Restart Nginx
 
 - name: Pull Spark Job Server container image
-  command: /usr/bin/docker pull quay.io/azavea/spark-jobserver:latest
+  command: /usr/bin/docker pull quay.io/azavea/spark-jobserver:0.6.1
 
 - name: Configure Spark Job Server service definition
   template: src=upstart-spark-jobserver.conf.j2
--- a/deployment/ansible/roles/model-my-watershed.spark-jobserver/tasks/main.yml
+++ b/deployment/ansible/roles/model-my-watershed.spark-jobserver/tasks/main.yml
@@ -37,7 +37,7 @@
     - Restart Nginx
 
 - name: Pull Spark Job Server container image
-  command: /usr/bin/docker pull quay.io/azavea/spark-jobserver:0.6.1
+  command: "/usr/bin/docker pull {{ sjs_container_image }}"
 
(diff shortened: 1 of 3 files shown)


deployment/ansible/roles/model-my-watershed.spark-jobserver/tasks/main.yml
```diff
@@ -37,7 +37,7 @@
     - Restart Nginx
 
 - name: Pull Spark Job Server container image
-  command: /usr/bin/docker pull quay.io/azavea/spark-jobserver:latest
+  command: /usr/bin/docker pull quay.io/azavea/spark-jobserver:0.6.1
```
mmcfarland: Would this be better placed in `group_vars` with the other version descriptions?

hectcastro: You're right. Addressed in https://github.com/WikiWatershed/model-my-watershed/commit/2b3c6c0c2294502d28879d160d9a46408c49943c.

mmcfarland:
I'm having trouble starting the sjs and mmw-geop services, but I don't think it's related to this PR.  I'm going to try a few more destroy's and provisions to see what happens.  FWIW, SJS service just hangs when starting and the log only seems to reflect me restarting it and looking to kill the container which isn't there.

```
Error response from daemon: Cannot kill container spark-jobserver: no such id: spark-jobserver
Error: failed to kill containers: [spark-jobserver]
Error response from daemon: no such id: spark-jobserver
Error: failed to remove containers: [spark-jobserver]
```

mmcfarland:
I can confirm, at least, that 0.6.1 is being pulled and started

```
vagrant@worker:~$ sudo docker ps
CONTAINER ID        IMAGE                                  COMMAND                  CREATED             STATUS              PORTS                    NAMES
a30a14e7e59d        quay.io/azavea/spark-jobserver:0.6.1   "./docker-entrypoint."   53 minutes ago      Up 53 minutes       0.0.0.0:8090->8090/tcp   spark-jobserver
```

hectcastro:
Yeah, I don't think they are related either. I opened this PR because I noticed that there were two entries for the `quay.io/azavea/spark-jobserver` image in the output of `docker images`. One for `latest` and another for `0.6.1`. The one from `latest` was coming from the Ansible pull.

The error output is expected because Upstart tried to kill off any running containers named `spark-jobserver` before attempting to start another.

Still trying to narrow down the unfavorable SJS behavior.

mmcfarland:
Cool.  +1, image looks correct.

PR #1089, opened 2016-01-13
Enable logs for RWD service

kdeloach:


kdeloach:
We will need to add logging messages to the RWD project to reap the benefits from this change.

caseypt:
+1 based on reviewing https://docs.docker.com/engine/reference/logging/overview/

kdeloach:
I'll follow up with another PR in the RWD project to dump the stack trace to stderr so they appear in Kibana.

Issue #1092, opened 2016-01-14
Add table to Model Output Runoff tab


ajrobbins:
In "Model", Runoff output/results tab, include a table similar to the Water Quality tab or any of the Geospatial Analysis tabs. The table should display runoff type (ET, runoff, infiltration) and depth (cm).

mmcfarland:
Keep in mind #1091, which is related.

Issue #1091, opened 2016-01-14
Add additional values to TR55 display output


mmcfarland:
For the Runoff and Water Quality tables, we would like two columns of data, one that is normalized to AOI area or AOI runoff (to be independent of AOI size) and one that is not. In other words:
* Runoff column 1: Runoff Depth (cm)
* Runoff column 2: Runoff Volume (m3): this is Runoff Depth (cm converted to m) multiplied by AOI area (m2)
* WQ column 1: Average Concentration (g/L)
this is equivalent to kg/m3, where Load (kg) is divided by Runoff Volume adjusted AOI runoff volume (m3)
* WQ column 2: Load (kg)

mmcfarland:
Update:
> WQ column 1: Average Concentration (g/L, which is equivalent to kg/m3), where Load (kg) is divided by adjusted AOI Runoff Volume (m3). One value per row of "Quality Measure"

PR #1102, opened 2016-01-22
TR55 Table Output Columns

mmcfarland:
* Represent the data in the water depth chart also in table form, including
the additional non-modifed column (either 100% Forest or Current
conditions depending on the scenario).
* Include a new column in the runoff table for the volume adjusted to the
AoI size and the Water Quality table for load adjusted by adjusted
runoff volume.

##### To Test
1. Create an AoI and note it's size (`471907` in my screenshots below)
2. Go to the model page and notice a new table for runoff.
  * Check that the values for water depth are consistent with the chart
  * Check that the values for volume are consistent with the formula: `(depth/100) * aoiArea`
3. Go to the water quality tab
  * Check that the new column, Concentration, has values consistent with the formula `load / modified runoff volume (from above)`

screenshot from 2016-01-22 11 36 55

screenshot from 2016-01-22 11 37 11

Connects #1091 
Connects #1092

### DIFF
--- a/src/mmw/js/src/modeling/tr55/quality/views.js
+++ b/src/mmw/js/src/modeling/tr55/quality/views.js
@@ -4,6 +4,7 @@ var $ = require('jquery'),
     _ = require('underscore'),
     Backbone = require('../../../../shim/backbone'),
     Marionette = require('../../../../shim/backbone.marionette'),
+    AoiVolumeModel = require('../models').AoiVolumeModel,
     chart = require('../../../core/chart.js'),
     barChartTmpl = require('../../../core/templates/barChart.html'),
     resultTmpl = require('./templates/result.html'),
@@ -34,6 +35,9 @@ var ResultView = Marionette.LayoutView.extend({
 
     initialize: function(options) {
         this.compareMode = options.compareMode;
+        this.aoiVolumeModel = new AoiVolumeModel({
+            areaOfInterest: this.options.areaOfInterest
+        });
     },
 
     onShow: function() {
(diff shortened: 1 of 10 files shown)


mmcfarland:
@caseypt added you since some changes may be relevant to the sidebar change.

caseypt:
I've tried a couple AoI's of different size, and every one has `NaN` for the "Average Concentration" values on the water quality tab.

caseypt:
Sample response here: https://gist.github.com/caseypt/a40a5f831f24100f708b

src/mmw/js/src/modeling/tr55/quality/views.js
```diff
@@ -64,12 +69,28 @@ var ResultView = Marionette.LayoutView.extend({
...
-    template: tableRowTmpl
+    template: tableRowTmpl,
+
+    templateHelpers: function() {
+        var load = this.model.get('load'),
+            runoff = this.model.get('runoff'),
+            adjustedRunoff = this.options.aoiVolumeModel.adjust(runoff);
+
+        return {
+            concentration: load / adjustedRunoff
```
caseypt: `runoff` is always null for me, which is leading to the `NaN` for concentration.

caseypt:
```
this.model.attributes
Object {load: 4573745.559121256, measure: "Biochemical Oxygen Demand"}
```

caseypt: Restarting Celery fixed this. Thanks.

caseypt:
Consider restructuring the runoff table to look more like this:

screen shot 2016-01-25 at 2 49 15 pm

I think this makes it easier to compare the scenarios.

caseypt:
+1. Tested. Nice job.

mmcfarland:
Thanks - I like that layout better.  Stroud defined the table/columns they wanted so I'd like to present them as is, but will suggest this update.

Issue #1090, opened 2016-01-14
Always display AoI area size


mmcfarland:
Always display [AOI area (m2 or km2)] in all views, including "Geospatial Analysis", "Model" and "Compare"

mmcfarland:
Wait until #1093 is complete

mmcfarland:
Closed as part of #1093

Issue #1093, opened 2016-01-14
Restructure layout for sidebar workflow


mmcfarland:
@jfrankl has developed wireframes for both a revamped workflow and a sidebar oriented display.  We'll start on implementation of the sidebar, this may be split into more detailed issues.

mmcfarland:
Blocked while @jfrankl is working on prototypes for this new feature.

mmcfarland:
Related: https://github.com/WikiWatershed/model-my-watershed/pull/1101

PR #1105, opened 2016-01-25
WIP: Integrate sidebar

caseypt:
Implements the new sidebar design.

TODO:
- [X] Skip over analyze page
- [X] ~~Run models when tab is activated~~ There are some issues with this.
- [X] Merge analyze and modeling route preparation
- [X] Move analyze results to tab
- [x] Make analyze tabs style match modeling tabs
- [x] Fix modeling spinners (currently disappearing before model is finished)
- [x] Fix analyze messages
- [x] ~~Format charts and graphs according to wireframes~~ There are bunch of tasks that need to be done here. I've fixed what I could, and will create a list of the remaning issues for @jfrankl.
- [x] Start analyze when page is loaded
- [x] ~~Fix tests~~ None of them broke!
- [x] Clean-up

Update: Most of the work is done here. There is some UI polishing that needs to happen, but testing and review can begin.

**Testing instructions**
- Select an area of the interest.
- You should be forwarded directly to the modeling page, which is rendered using the new sidebar design.
- Interact with the app, and report and weird behavior or bugs created by moving to this design.

Known issues:

- Minor UI issues described in a comment below.
- The charts code probably use another pass. A lot of times, the ticks are overlapping, and their style doesn't quite match up with the wireframes (in some minor ways): https://marvelapp.com/ad6b6b#8306748
- Some model runs appear to be canceling each other out, but all the results appear to come back. It's not clear to me yet what's going on here. It happened more frequently when we only ran models when a modeling tab was activated. To avoid that, we now run them all at once like we did previously.

Connects to #1093.


rajadain:
When I select the "Model" tab in the sidebar, the selected tab "Runoff" isn't highlighted:

image

But if I click the other tab "Water Quality" and then click back, it is:

image

caseypt:
Thanks @rajadain. The white background wasn't supposed to be there, according to the wireframes. Fixed in 65e4f53.

caseypt:
Here's a list of (mostly minor) design issues with the sidebar for @jfrankl. I can go over each of these in detail if my descriptions don't make sense. I tried to take care of as much as I could, but I got tripped up by some of these.

- Can’t scroll to bottom of analyze tab. There’s an issue with the height of one of the content’s parent elements that is cutting off the bottom of the div.

image

- Scrolling elements can be seen in a small section of the tab header as you scroll

image

- Table styles don’t match those in wireframes (the general bootstrap style is overridden by a plugin we are using, bootstrap-table). This will take a custom class and some tweaking.

- When switching back and for between analyze and model tabs, the horizontal scrollbar for the table appears half way up and at the bottom of the div on the analyze table until you try and click it. Then it moves to the bottom of the div.

image

- Because of the number of columns in the analyze table, there is a vertical scrollbar. (not sure there’s a way around this).

- When viewing the page in the Chrome device emulator, there is a lot of white space around the app that you can scroll through.

image

mmcfarland:
Additional design consideration:  Move the AoI name and size to somewhere visible at all times - it's not just relavent to Analyze and would fix #1090 

screenshot from 2016-01-26 17 09 22

mmcfarland:
Another design thing, if the sidebar could be made to take up more space on wide screens, we wouldn't need to scroll to see the table.  For example, the map is very large on my monitor, but the chart/output area is squished.

mmcfarland:
Looks like "Compare Mode" is broken on this branch:

screenshot from 2016-01-26 17 15 10

caseypt:
Fix for compare mode is in #1115.

jfrankl:
I made all of these fixes except for moving the Aol name (I need to think about that more).

The changes are in this branch: https://github.com/WikiWatershed/model-my-watershed/tree/feature/jfrankl/feature/style-sidebar

@caseypt I didn't know how to make a PR against your fork. Let me know how to do that or if you want to get these commits a different way.

image

caseypt:
Superseded by #1119

PR #1107, opened 2016-01-25
Send BMP area totals to TR55

mmcfarland:
Updates to TR55 require conservation practices to be passed in as sums
of thier effective area so that infiltration is added across the AoI and
not from a specific location of them.

Connects https://github.com/WikiWatershed/tr-55/issues/59

To test:
* Reprovision `worker` with this branch
* Restart celery to get latest task
* Model an AoI and draw a rain garden, porous paver, veg basin or green roof
* Notice a larger than usual decrease in runoff and a commensurate increase in infiltration
  * Precipitation below 2.5cm may produce *no* runoff when a BMP is applied, try increasing the value
screenshot from 2016-01-25 18 04 42
* Draw a very small rain garden and see that it decreases runoff (until the new table is merged in, you may not see a different due to rounding precision).
screenshot from 2016-01-25 18 14 37

### DIFF
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -200,6 +200,19 @@ def run_tr55(censuses, model_input, cached_aoi_census=None):
     modification_censuses = (censuses[1:] if cached_aoi_census is None
                              else censuses[0:])
 
+    # Calculate total areas for each type modification
+    area_sums = {}
+    for piece in modification_pieces:
+        kind = piece['value']
+        area = piece['area']
+
+        if kind in area_sums:
+            area_sums[kind] += area
+        else:
+            area_sums[kind] = area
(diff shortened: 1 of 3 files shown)


caseypt:
Everything looks good, but I'm not seeing a difference when drawing some BMPs (as you pointed out). What is the "new table" that needs to be merged in?

mmcfarland:
Small BMPs  should still affect the runoff/infiltration balance, but the chart only shows 1 digit of precision.  It'd be easier to see it happening with the new runoff results table.  It's now merged into develop so I've rebased this PR.   You should now see small changes in runoff between modified and current conditions when drawing small BMPs in the runoff table.

caseypt:
+1. Tested. 

screen shot 2016-01-26 at 10 10 09 am

Issue #1111, opened 2016-01-26
Soils bar chart colors should match gSSURGO


ajrobbins:
The soil type bar chart should utilize gSSURGO colors, so it also serves as a legend.
![screen shot 2016-01-26 at 4 22 54 pm](https://cloud.githubusercontent.com/assets/12868342/12595418/28993122-c449-11e5-84d4-170048da284b.png)

**gSSURGO Symbology**
Using a 7-class YlOrBr from ColorBrewer:

HSG | Raster Value | Color
----- | ----- | -----
NoData | 0 | transparent
A | 1 |  #ffffd4
B | 2 | #fee391
C | 3 | #fec44f
A/D | 5 | #fe9929
B/D | 6 | #ec7014
C/D | 7 | #cc4c02
D | 4 | #8c2d04

PR #1114, opened 2016-01-27
Soil chart and table updates

mmcfarland:
* The analyze hydrologic soil chart will play double duty as a legend with
the colors of the rendered soil layer matching the bar values.
* Start the chart and table sorted alphabetically by group name

To Test:

* Select an area with many soil types and make sure the original teal color is replaced by the corresponding soil group color value (High Infl => Light; Low Infl => Dark)

screenshot from 2016-01-26 21 18 17

* Check that the initial sort on both table and chart is alphabetic, and that the NLCD table/chart is still by NLCD code (Open Water is first).

Connects #1111


caseypt:
+1. Tested. Looks good.

screen shot 2016-01-27 at 1 42 40 pm

mmcfarland:
Updated with python list change and fixed test

PR #1115, opened 2016-01-27
Allow runoff chart to be displayed

mmcfarland:
Add styling to allow full height display

screenshot from 2016-01-27 14 19 56


caseypt:
Consider splitting out the commit to fix compare mode or updating the existing commit with a message regarding the fix.

caseypt:
Nevermind. I misread the initial message. I thought it was just addressing styling.

+1. Looks good.

screen shot 2016-01-28 at 9 12 31 am

Issue #1108, opened 2016-01-26
Allow overlay panel to function if Vizer API is down


mmcfarland:
The observation API endpoint at www.wikiwatershed-vs.org stopped responding to requests and the overlay panel was unavailable.  The basemap and overlay selector tabs should still show up and only observations should be unavailable.

PR #1116, opened 2016-01-28
Load vizer layers asyncronously

mmcfarland:
Vizer Observation layers were requested async, but the layer control was
not rendered until their successful return.  This caused the control to
disappear when the Vizer API was down.  This commit updates the layer
control with vizer layers as they are returned, without waiting on them
to render.  If there are failures in loading the UI will report a
message on the observation tab.

Connects #1108 

To test:
* First make sure that the observation overlays are showing and loading to the map (if the API is up, it is at the time of this writing)
* Alter the `$.getJSON(vizerUrls.layers)` line in VizerLayers.js to cause an error (appending a random string to the request will work: `$.getJSON(vizerUrls.layers + 'foo')`.  
* Rebundle and reload the page.  The layer control should exist and there should be a basic error message in the observation panel.

### DIFF
--- a/src/mmw/js/src/core/layerControl.js
+++ b/src/mmw/js/src/core/layerControl.js
@@ -11,7 +11,9 @@ module.exports = L.Control.Layers.extend({
 
     // Copied from  https://github.com/Leaflet/Leaflet/blob/master/src/control/Control.Layers.js
     // with modifications for allowing a third layer type (observations)
-    initialize: function (baseLayers, overlays, observationLayers, options) {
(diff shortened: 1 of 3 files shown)


mmcfarland:
@azavea-bot rebuild

caseypt:
+1. Tested. Code looks good. 

screen shot 2016-01-28 at 1 27 12 pm

lewfish:
+1, tested

mmcfarland:
These tests pass locally.  @azavea-bot rebuild

Issue #1117, opened 2016-01-28
Finish sidebar integration


ajrobbins:
Fallout from sidebar

PR #1119, opened 2016-01-31
Integrate sidebar (updated with style fixes)

caseypt:
Supersedes #1105 

Testing instructions:
* Select an area, run models, and report any styling or functionality issues.

Known issues: 
* Some model runs appear to be canceling each other out, but all the results appear to come back. It's not clear to me yet what's going on here. It happened more frequently when I switched models to run only  when that scenario's model tab was activated. To avoid these problems, I'm running all of the models at once when the page is loaded (which is what we are doing now).

Connects to #1093 
Connects to #1117


mmcfarland:
The analyze chart area seems to have an excessive right margin - is it necessary or can the chart be a bit wider in the container?

screenshot from 2016-02-01 09 14 03

mmcfarland:
Also may need to rebase on `develop`, getting that compare view error that was fixed last week.

caseypt:
This should be rebased on the latest `develop`.

caseypt:
> Also may need to rebase on develop, getting that compare view error that was fixed last week.

I think I'm running into this now. This looks like an existing bug with the compare view that wasn't exposed before because we reused the same region for the modeling page. Fixing now.

caseypt:
Update with a fix for the issue reported by @mmcfarland.

caseypt:
Edited a previous commit to expand the chart area. It is even wider now.

mmcfarland:
The scenario tab dropdown is obscured by the header row.  The project dropdown is correctly on top, though.

screenshot from 2016-02-01 14 46 02

caseypt:
Updated with a fix from @jfrankl that addresses the issue @mmcfarland just reported.

caseypt:
Rebased on `develop`. Compare charts are fixed:

screen shot 2016-02-01 at 3 21 17 pm

caseypt:
@azavea-bot rebuild.

caseypt:
(+1 in slack by @mmcfarland).

PR #1121, opened 2016-02-01
Runoff chart display only 1 bar in compare mode

mmcfarland:
A recent refactor removed a reference to the view mode (compare view or
not).  This passes the reference to the newly created chart view which
ultimately restricts the number of bars represented in the chart.

screenshot from 2016-02-01 14 57 18


caseypt:
+1. Tested.

screen shot 2016-02-01 at 3 11 45 pm

caseypt:
Going to merge this.

PR #1122, opened 2016-02-02
Turn of RWD in production

mmcfarland:
The RWD feature is disabled as it's not been approved for release.  This
commit is likely followed up with a commit to re-enable it so that it is
available for additional testing in staging.

Testing:  This will only remove RWD in production/staging environments.


caseypt:
+1 based on examining the history of changes to this file.

hectcastro:
Going to merge to get the release process started.

PR #1124, opened 2016-02-02
Use SQL data access layer vs. file access layer

hectcastro:
The Docker container image assembled by the SJS author makes use of `SqlDAO` and claims it is more robust when compared to the `FileDAO`. When testing the issues we've been seeing in staging with `FileDAO` enabled, I was unable to reproduce them.


caseypt:
+1. After running `vagrant reload worker --provision`, SJS was not running. However, after running `vagrant destroy worker && vagrant provision worker`, SJS was running and successfully running models. There was no need to manually restart the service.

Issue #1127, opened 2016-02-02
Divide by 0 error when precipitation is set to 0


ajrobbins:
When precipitation is set to 0, modeling output disappears.

Test scenario: https://mmw-dev.azavea.com/project/524/scenario/991

![screen shot 2016-02-02 at 5 22 51 pm](https://cloud.githubusercontent.com/assets/12868342/12766409/b1e25732-c9d1-11e5-834e-d87c05a582ec.png)

PR #1130, opened 2016-02-03
Upgrade TR55 for division by 0 fix

mmcfarland:
Fixes in the TR55 library allow runoff to be zero when a BMP is applied.

Connects #1127 

To test: 
* Provision your worker to get the new package
* Model a scenario with BMPs and no precipitation
* You should not get an error thrown, nor empty results.
* Output should resemble:

screenshot from 2016-02-02 19 09 25

Which matches what is currently in production where the error was not present:

screenshot from 2016-02-02 19 09 50


caseypt:
+1. Tested.

image

Issue #1125, opened 2016-02-02
Wrong unit labels in land cover and soil type tables


ajrobbins:
When a smaller AoI is being analyzed, the land cover and soil type areas change units (so that for a 1km square AoI you can see areas in square meters, not fractions of a square km, etc). In the post-sidebar version, the column headers are not changing to reflect this.

Example project: https://mmw-dev.azavea.com/project/524/scenario/990

The column headers should say Area (m^2), not (km^2)
![screen shot 2016-02-02 at 12 01 54 pm](https://cloud.githubusercontent.com/assets/12868342/12765848/893c772a-c9ce-11e5-86f3-e6a13269e732.png)
![screen shot 2016-02-02 at 5 00 36 pm](https://cloud.githubusercontent.com/assets/12868342/12765851/8baba292-c9ce-11e5-9c99-ca6df9dd7195.png)

PR #1132, opened 2016-02-03
Fix analyze table header

caseypt:
* In 2b39fe8, the AoI name and area were added to the analyze results.
Because this change set the shape attribute of the GeoModel for this
view, the model initialization logic overwrote the units
attribute. We manually calculated this value to match the units of
the analyze results.  When the table header was drawn, the units value
matched that of the AoI and not the result units. Here we send
along the manually calculated units as a separate variable to the
analyze table view. We also remove the line to manually set the units
of the GeoModel instance because the initialization logic will take
care of this for us.

**Testing instructions**
- Draw a small AoI (<= 1 km2), and verify that the analysis results table displays `m2` for the area units.
- Draw a large AoI (>= 100 km2), and verify that the analysis results table displays `km2` for the area units.

screen shot 2016-02-02 at 9 19 20 pm
screen shot 2016-02-02 at 9 20 23 pm

Connects to #1125

### DIFF
--- a/src/mmw/js/src/core/utils.js
+++ b/src/mmw/js/src/core/utils.js
@@ -126,9 +126,9 @@ var utils = {
 
     magnitudeOfArea: function(value) {
         if (value >= M2_IN_KM2) {
-            return 'km2';
+            return 'km<sup>2</sup>';
         } else {
-            return 'm2';
+            return 'm<sup>2</sup>';
         }
     },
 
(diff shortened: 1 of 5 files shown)


mmcfarland:
+1, tested.

caseypt:
Updated with a commit to use superscript for square units. cc @ajrobbins.

image

mmcfarland:
+1, retested.  Superscript <sup>all around</sup>

caseypt:
Well done.

Issue #1138, opened 2016-02-04
Reduce whitespace in model tab


mmcfarland:
> Reduce "white space" as much as possible in the model screen - especially on the upper right

Issue #1137, opened 2016-02-04
Remove comparision table rows from model output


mmcfarland:
> In the model/runoff tab, remove the mini-comparison with 100% forest (for the current conditions tab) or the current condition (for the other tabs) - they are nice, but they make the table confusing and are inconsistent with any other tabs

mmcfarland:
This includes the comparison in the 2-bar chart.  The chart and table should both only show the individual scenario.

Issue #1135, opened 2016-02-04
Reverse vertical order of tables/charts


mmcfarland:
In all of the analysis/model tabs with both a bar graph and table, put the bar graph above the table

PR #1139, opened 2016-02-04
Remove model comparisons from output

mmcfarland:
Removes all inter-scenario comparisons in TR55 runoff output, including:

* Bar chart only describes current scenario
* Table only describes current scenario
* Compare mode should still include non-scenario "100% Forest" item.

Connects #1137 
Connects #1135 
Connects #1138 

screenshot from 2016-02-04 13 26 28

### DIFF
--- a/src/mmw/sass/components/_charts.scss
+++ b/src/mmw/sass/components/_charts.scss
@@ -13,6 +13,7 @@
   height: 300px;
   vertical-align: baseline;
   vertical-align: bottom;
+  margin-bottom: 20px;
 
   .nv-bar {
     fill: $brand-primary;
--- a/src/mmw/sass/components/_charts.scss
+++ b/src/mmw/sass/components/_charts.scss
@@ -5,7 +5,6 @@
 .chart-container {
   width: 100%;
   height: 100%;
-  margin-top: 30px;
 }
 
 .bar-chart {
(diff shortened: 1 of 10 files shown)


mmcfarland:
Added whitespace commit to address #1138

src/mmw/sass/components/_charts.scss
```diff
@@ -5,7 +5,6 @@
 .chart-container {
   width: 100%;
   height: 100%;
-  margin-top: 30px;
```
caseypt: Consider switching this to `margin-bottom` to add some space between the chart and graph. It's currently a little tight:

image

caseypt:
Just a reminder to switch the analyze chart and table order too.

caseypt:
+1 besides the comments.

caseypt:
Need another look at this, or are you still adding commits?

mmcfarland:
Ready for another review.  Added analyze reorder and margin in 92441bc & 3ae8a5b

caseypt:
+1. New changes look good. One minor issue is if you select an area with only one soil type, there is excessive space between the chart and table. I believe this is due to a fixed height of the SVG. Fixing it may require a deeper update of the chart.

image

mmcfarland:
@azavea-bot rebuild

Issue #1141, opened 2016-02-05
Make the compare button more prominent


mmcfarland:
From Stroud:
> Make the compare button more prominent

Issue #1136, opened 2016-02-04
Allow sidebar to be collapsed


mmcfarland:
From Stroud:
> Make the analysis/model tabs minimiz-able so the whole screen can be the map (like the bottom bar was before)

SRGDamia1:
Collapsed!  That was the word I'd been looking for!

PR #1142, opened 2016-02-05
Collapse sidebar

mmcfarland:
* Add element to transition visibility of map and sidebar
* Reorient map controls to the left so they don't interfere with sidebar
* Make sidebar narrower
* Make "Compare" button more prominent

Connects #1136 
Connects #1141

Test:
* Make sure new arrow button is not visible on any page but a project
* Test that the button toggles the size of the map/sidebar and re-adjusts the extent and AoI fit in the map
* Make sure that if the sidebar is minimized, the page is transitioned (to draw/compare) and transitioned back - the collapse toggle still behaves correctly
* Test the map controls are still accessible and open controls correctly (Stream slider is moving soon, and is currently a little weird - not active for staging/production)

### DIFF
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -26,6 +26,11 @@ require('leaflet-plugins/layer/tile/Google');
 
 var RootView = Marionette.LayoutView.extend({
     el: 'body',
+    ui: {
+        collapse: '.tab-content-toggle',
+        mapContainer: '.map-container',
+        sidebar: '#sidebar'
+    },
     regions: {
         mainRegion: '#container',
         geocodeSearchRegion: '#geocode-search-region',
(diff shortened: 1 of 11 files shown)


caseypt:
The build is failing because of some lint:

```
js/src/core/views.js: line 53, col 69, Missing semicolon.
js/src/core/views.js: line 54, col 74, Missing semicolon.
```

mmcfarland:
De-linted in 5b1e9a9.

caseypt:
On the analyze > land tab, I get a horizontal scrollbar. 

screen shot 2016-02-08 at 9 48 02 am

caseypt:
The layers control is hidden behind the draw page elements. 

image

It might be easier to push down all the map controls, like so:

image

Though we would also have to it adjust when window shrinks because the draw controls drop into two rows.

src/mmw/js/src/core/views.js
```diff
@@ -33,9 +38,38 @@ var RootView = Marionette.LayoutView.extend({
...
         },
         footerRegion: '#footer'
+    },
+    events: {
+        'click @ui.collapse': 'collapseSidebar',
+        'transitionend @ui.mapContainer': 'onMapResized'
+    },
+
+    collapseSidebar: function() {
+        // Toggle appropriate classes to make show and hide
```
caseypt: Typo. I think you can drop `make` in the first line here.

mmcfarland:
I had it that way first, but then on later screens without those draw controls, there's an unusual amount of space between the map controls and the top of the map container.  I'll try just adjusting the top of the layer panel.

mmcfarland:
Comments have all been addressed.  I ended up moving the layer control down to line up with the control button.  It's not ideal, but it looks the same with and without drawing tools on the map, and also accommodates the 2-row version of the draw tools.

screenshot from 2016-02-09 10 03 34
screenshot from 2016-02-09 10 03 17
screenshot from 2016-02-09 10 03 02

caseypt:
+1. Changes look good to me.

Issue #1134, opened 2016-02-04
Re-implement the work-flow separation between analyze and model


mmcfarland:
- After selecting an AoI, immediately go to geospatial analysis mode, with the results on the right as you have now
- Force user to explicitly click a model button so they understand that there's clearly something different between the analysis of the data and the model's predictions
- Still maintain the analysis results in the model view, just as is now

PR #1145, opened 2016-02-08
Analyze Sidebar

caseypt:
Adds back the analyze page and implements the sidebar design. See commit messages for details.

**Testing instructions**
- Select an AoI.
- You should advance to the analyze page which is sporting the new sidebar design.
- Advance to the modeling page. The modeling tab should be active and the analyze results shouldn't need to be fetched.

screen shot 2016-02-09 at 12 00 58 pm

TODO: 

- [x] Add a new view that wraps the analyze view, similar to the one used in modeling. This will only show an Analyze tab.
- [x] Apply new map sizing for sidebar but no header.
- [x] Advance to modeling page.
- [x] Preserve analyze results from analyze to modeling.
- [x] Review changes from initial PR and bring back any code that may now be needed again.

Connects to #1134


caseypt:
Adding @jfrankl so he can get a look at the "Model this area" button. See the above screenshot.

mmcfarland:
A workflow issue is that you can't now go back from Analyze to Draw, or otherwise reset your AoI.  Not sure where the most appropriate spot for that button would be.  Maybe a floating "Reset" button the map, similar to the one in draw tools?

Additionally, you should try rebasing current release to see if the new collapse/toggle button interferes with the new Model button

mmcfarland:
Code looks good.

mmcfarland:
+1

PR #1153, opened 2016-02-15
Upgrade TR55 for Green Roof fix

mmcfarland:
Uses an updated version of TR55 which includes the correct water storage
unit for Green Roof.

Test:
Reprovision Worker and ensure Green Roofs can be applied to a scenario.


rajadain:
+1 tested

image

Issue #1085, opened 2016-01-12
Preserve points in RWD output


mmcfarland:
For visual display, we should preserve and display these points from the RWD process:

* Point of Click - where the user clicked on the map
* Point of Snap - When doing Stream Network, the point at which the PoC was snapped to a stream feature (may not be directly available)
* Outlet Point - Should be available as a shapefile output in RWD, `New_Outlet.shp`

PR #1154, opened 2016-02-16
Add snapped and clicked outlet point

mmcfarland:
Preserve the originally clicked location and the point which the RWD
service snapped to along with the AoI.

* Upgrade RWD service to that which supports returning outlet points
* Save click and snap points and display on map alongside AoI

Connects #1085 

##### Caveats
* Marker style, marker duration and popup language open to feedback, intending to release on staging for experiments

##### Test
* Provision worker
* Delineate a watershed using Stream Network
* See that there are two points, one at your original click location.
* See that the markers are visible while the AoI is visible, and are removed when the AoI is reset or redrawn.

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -41,8 +41,11 @@ function actOnLayer(datum) {
     }
 }
 
-function validateShape(shape) {
-    var area = coreUtils.changeOfAreaUnits(turfArea(shape), 'm<sup>2</sup>', 'km<sup>2</sup>'),
+// Takes a GeoJSON shape, or if a key is specified, a dict
+// with a GeoJSON object found at shape[shapeKey]
+function validateShape(shape, shapeKey) {
+    var polygon = shapeKey ? shape[shapeKey] : shape,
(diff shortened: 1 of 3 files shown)


mmcfarland:
@azavea-bot rebuild

rajadain:
+1 tested

image

Saw the two map points in various stream networks. Clicking on each describes it in a popup, which is nice. Resetting draw tools removes the points as well as the area of interest. Delineating another one adds new points. Delineating using Hill Slope adds only the snapped point, but does have the popup which is useful:

image

mmcfarland:
@azavea-bot rebuild

src/mmw/js/src/draw/views.js
```diff
@@ -410,10 +412,32 @@ var WatershedDelineationView= Marionette.ItemView.extend({
                 self.rwdTaskModel.start(taskHelper);
                 return deferred;
             })
-            .done(validateShape)
-            .done(function(shape) {
-                addLayer(shape, itemName);
-                navigateToAnalyze();
+            .done(_.partialRight(validateShape, 'watershed'))
```
rajadain: Cool use of `_.partialRight`

rajadain:
Code looks great! Very clean.

Issue #1099, opened 2016-01-18
Snapshot incorrect after panning


ajrobbins:
A bug was reported in which the ITSI snapshot fails to render or displays the incorrect location after the first snapshot is taken. 

Map #1:
![screen shot 2016-01-18 at 12 57 58 pm](https://cloud.githubusercontent.com/assets/12868342/12398676/261aa5aa-bde3-11e5-9f8c-d450bf222d7e.png)

Snapshot:
![screen shot 2016-01-18 at 12 47 42 pm](https://cloud.githubusercontent.com/assets/12868342/12398720/6652f7f8-bde3-11e5-9b6b-9aeb4fa0b942.png)

Map #2 (after panning)
![screen shot 2016-01-18 at 12 48 45 pm](https://cloud.githubusercontent.com/assets/12868342/12398698/474e6784-bde3-11e5-9726-4429b9b2d900.png)

Snapshot #2:
![screen shot 2016-01-18 at 12 49 51 pm](https://cloud.githubusercontent.com/assets/12868342/12398708/574347fe-bde3-11e5-8000-4f43f2f640aa.png)

Snapshot #3 (after zooming out)
![screen shot 2016-01-18 at 12 50 14 pm](https://cloud.githubusercontent.com/assets/12868342/12398726/736abb06-bde3-11e5-9eb0-0cb4a87cca72.png)

**To replicate:**
1. Log into the ITSI portal via MMW or learn.staging.concord.org and run the MMW Staging activity.
2. Log out and then select continue as ITSI to log back in
3. Take a snapshot
4. Pan to a different location and take another snapshot; you should see a screenshot of the first location, not the new one.
5. Zoom and take a snapshot; it should be missing tiles (I could reproduce this error less reliably than the panning issue).

*Note:* The only way to reliably reproduce this is to start from a completely new project by logging in again. A hard refresh will fix the problem, as will returning to the "Classes and Offerings" page and running the activity again.

rajadain:
Is there value in letting people log out if they are in embed mode? If we remove the option to log out, then we can avoid this use case entirely.

ajrobbins:
The log out/ login behavior that was necessary for me to replicate this issue may have been related to me using our general tstudent2 login for learn.staging; it has some saved scenarios and I was only able to replicate this issue by logging out and logging in fresh. So it's possible that it's not related to the logging back in at all.

rajadain:
[...]

PR #1156, opened 2016-02-16
Fix ITSI Screenshots between sessions

rajadain:
## Overview

The approach taken in #978 relies on the presence of `itsi_embed: true` in the `window.clientSettings` hash, which is populated from the session variable `itsi_embed`, which is populated from the URL query string `?itsi_embed=true`. In most cases it worked correctly since an ITSI embed session would always be initialized with `?itsi_embed=true`. However, there are two cases where it didn't work:

 1. New users being registered for the first time. While we did save `itsi_embed` to the session for these users, as soon as we logged them in they would get a new session id, one that did not have these old variables.
 2. Existing users logging out, and then logging with ITSI via the UI. This would redirect them to new page without the `?itsi_embed=true` URL, thus this new logged in session would not have it correctly set.

By saving the session value before logging a new user in, and restoring it after they are logged in, we address 1.
By simulating an initialization with `?itsi_embed=true` when in embed mode, instead of sending them to the classical ITSI login URL, we address 2.

## Testing Instructions

This PR can be indirectly tested by ensuring that `window.clientSettings.itsi_embed` is set in all the correct cases, and not in all the incorrect ones. You can test its current value by typing it into the JavaScript console

    window.clientSettings.itsi_embed

Test Case | Description | `window.clientSettings.itsi_embed`
----------|-------------|-----------------------------------
 1 | Go to home page [http://localhost:8000/](http://localhost:8000/) | `false`
 2 | Log in as ITSI | `false`
 3 | Go to home page in embed mode [http://localhost:8000/?itsi_embed=true](http://localhost:8000/?itsi_embed=true) | `true`
 4 | Log out | `true`
 5 | Log in as ITSI | `true`
 6 | Log out and refresh the page | `false`
 7 | Log out of ITSI, and log in with an account that does not have a corresponding MMW account (or, delete the MMW account corresponding to your ITSI login). Then go to home page in embed mode [http://localhost:8000/?itsi_embed=true](http://localhost:8000/?itsi_embed=true). | `true`

To test this within the ITSI embed environment and the actual screenshots, more preparation is required. Come see me, and I'll set it up.

Connects #1099

### DIFF
--- a/src/mmw/apps/user/views.py
+++ b/src/mmw/apps/user/views.py
@@ -9,6 +9,7 @@ from django.shortcuts import redirect
 from django.contrib.auth.forms import PasswordResetForm
 from django.core.exceptions import ObjectDoesNotExist
 from django.contrib.sites.shortcuts import get_current_site
+from django.conf import settings
 
 from registration.models import RegistrationProfile
 from registration.forms import RegistrationFormUniqueEmail
@@ -20,6 +21,8 @@ from rest_framework.permissions import AllowAny
 
 from apps.user.models import ItsiUser
 from apps.user.itsi import ItsiService
+
+EMBED_FLAG = settings.ITSI['embed_flag']
 
 
 @decorators.api_view(['POST', 'GET'])
@@ -129,8 +132,6 @@ def itsi_auth(request):
         auth_login(request, user)
         return redirect(request.GET.get('next', '/'))
     else:
-        # User did not authenticate. Save their ITSI ID and send to /sign-up
-        request.session['itsi_id'] = itsi_user['id']
         return itsi_create_user(request, itsi_user)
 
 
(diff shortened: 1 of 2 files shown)


rajadain:
I thought about not allowing users to log out while in embed mode,  but that unfortunately traps them in embed mode forever, since they don't log out the session variable is always set. I think this solution is more elegant.

src/mmw/apps/user/views.py
```diff
@@ -129,8 +132,6 @@ def itsi_auth(request):
         auth_login(request, user)
         return redirect(request.GET.get('next', '/'))
     else:
-        # User did not authenticate. Save their ITSI ID and send to /sign-up
-        request.session['itsi_id'] = itsi_user['id']
```
rajadain: Removing this since it is actually never used, and also would be lost as the session changes.

mmcfarland:
+1, test cases all work out and screenshots are correct after using steps posted originally by @ajrobbins

rajadain:
Thanks for reviewing. Also kudos to @ajrobbins for a great and detailed write-up of the original issue, which made fixing this considerably easier.

PR #1152, opened 2016-02-13
Style Opacity Slider

rajadain:
## Overview

The new styling matches other leaflet plugins, works across browsers.

## Demo

image

In Chrome only, when the slider is in an active state, the thumb becomes all black:

image

## Testing Instructions

Test in various browsers.

Connects #1094

Issue #1094, opened 2016-01-14
Fix opacity slider styling


ajrobbins:
Overlay opacity slider needs styling:
![screen shot 2016-01-14 at 12 23 49 pm](https://cloud.githubusercontent.com/assets/12868342/12331944/f1104418-bab9-11e5-84e1-db220ccc059c.png)
![screen shot 2016-01-14 at 12 23 57 pm](https://cloud.githubusercontent.com/assets/12868342/12331949/f51c1e7e-bab9-11e5-8660-ccaaf80f7a60.png)

PR #1157, opened 2016-02-16
Style Opacity Slider

rajadain:
#1152 against `develop`.

Connects #1094

### DIFF
--- a/src/mmw/js/src/core/opacityControl.js
+++ b/src/mmw/js/src/core/opacityControl.js
@@ -87,6 +87,7 @@ var SliderView = Marionette.ItemView.extend({
     tagName: 'input',
     className: 'slider slider-leaflet',
     attributes: {
+        title: 'Drag to change opacity of overlay',
         type: 'range',
         min: 0,
         max: 99,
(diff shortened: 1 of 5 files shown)


mmcfarland:
On my version of Chromium (`Version 48.0.2564.82 Ubuntu`), the scrollbar is horizontal.

screenshot from 2016-02-17 11 03 16

mmcfarland:
Firefox, too.  Does this seem like a bundling problem?  The styles don't look like yours in the screenshots and they _do_ look like the old sliders from Christopher's mockups.

screenshot from 2016-02-17 11 18 23

mmcfarland:
Sliders are building correctly and look good, but it's a little hard to recognize them as a slider for opacity.

* Is there a way to keep a little track line down the center that makes it look more traditionally like a slider?
* Can we add a tool tip or something to the handle to read "Overlay Opacity" or something?

rajadain:
Added tooltip and a visual track, although it renders differently in different browsers:

Chrome on Mac:
image

Safari on Mac:
image

Firefox on Mac:
image

IE 11 on Windows 7:
image

Firefox on Windows 7:
image

Chrome on Windows 7:
image

Perhaps unsurprisingly, IE is the biggest disappointment, because it renders borders _over_ the slider thumb, rather than _under_ it like everybody else. Still, I don't think it is unusable, and probably an improvement. The tooltip does help as well.

image

rajadain:
Just pushed a fix for Firefox:

image

mmcfarland:
Great improvement, +1

screenshot from 2016-02-17 15 26 27

PR #1158, opened 2016-02-17
RWD Enabled

mmcfarland:
* Enables RWD in production and staging environments
* Disables Hill Slope method of RWD

To test, see that Hill Slope is not currently an option.


rajadain:
+1 tested

image

PR #1161, opened 2016-02-19
Remove timestamp from log4j; add multiline support

hectcastro:
Gather up Java exceptions into one message vs. having them split by line. Also, remove the timestamp from the log4j output and instead depend on the timestamp provided by the Kibana UI:

discover_-_kibana_4

---

**Testing**

Provision changes in this pull request and cause a geoprocessing exception. Then, take a look at the Kibana UI.

To cause a geoprocessing exception, introduce a typo into the SJS configuration file (`spark-jobserver.conf.j2`):

```scala
jobdao = spark.jobserver.io.JobSqlDAOjoker
```


mmcfarland:
+1, nice clean exception display for ghost mouse

screenshot from 2016-02-19 16 41 32

PR #1175, opened 2016-03-09
Display alert when RWD does not yield a valid watershed

### LINKED ISSUE #1166: Select by Boundary not working on mmw-dev
The boundary lines are rendering, but when an AoI is clicked the lines disappear and nothing happens. Observed this in Chrome, Firefox, and Safari.

kdeloach:
There are certain conditions where the RWD job completes successfully,
but does not yield a valid watershed. Without this alert, there is no
indication that the job completed or not.

The reason this is a plain alert box instead of something nicer, is
because we intend to address the cause behind the null watersheds soon.

Ref: WikiWatershed/rapid-watershed-delineation#27

Fixes #1166

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -41,11 +41,22 @@ function actOnLayer(datum) {
     }
 }
 
-// Takes a GeoJSON shape, or if a key is specified, a dict
-// with a GeoJSON object found at shape[shapeKey]
-function validateShape(shape, shapeKey) {
-    var polygon = shapeKey ? shape[shapeKey] : shape,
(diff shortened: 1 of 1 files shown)


src/mmw/js/src/draw/views.js
```diff
@@ -41,11 +41,22 @@ function actOnLayer(datum) {
     }
 }
 
-// Takes a GeoJSON shape, or if a key is specified, a dict
-// with a GeoJSON object found at shape[shapeKey]
-function validateShape(shape, shapeKey) {
```
mmcfarland: Removing this will actually fix a newly reported bug https://github.com/WikiWatershed/model-my-watershed/issues/1166

mmcfarland:
+1

kdeloach:
Cool, I added more details in the commit description too.

PR #1182, opened 2016-03-10
Support a second domain for staging and production

### LINKED ISSUE #1165: Point mmw.azavea.com to new URL
Stroud is requesting that we shift to using `app.wikiwatershed.org/model`. They are anticipating that in the future another app would be added at `/monitor`, etc.

hectcastro:
In addition to the existing `{mmw,mmw-dev}.azavea.com` domains, support the following two domains interchangeably:

- `app.wikiwatershed.org`
- `staging.app.wikiwatershed.org`

In addition, change the domain used by the application for outbound e-mail to `app.wikiwatershed.org`.

Helps resolve #1165


mmcfarland:
From what I can parse, this makes sense.  I assume it's the culmination of your testing on staging and is otherwise difficult to exercise.

hectcastro:
Most of the testing here I think we covered with the ITSI Portal integration.

/cc @rajadain

mmcfarland:
Cool, +1 then.

deployment/default.yml.example
```diff
@@ -9,7 +9,8 @@ NATInstanceAMI: ''
 BastionHostInstanceType: 't2.micro'
 BastionHostAMI: ''
 PrivateHostedZoneName: 'mmw.internal'
-PublicHostedZoneName: 'mmw.azavea.com'
+PublicHostedZoneName: 'staging.app.wikiwatershed.org'
+BackwardCompatPublicHostedZoneName: 'mmw-dev.azavea.com'
```
rajadain: :clap:

rajadain:
+1, looks good to me! Confirmed that the ITSI stuff is working with the parallel load balancer approach.

Issue #1179, opened 2016-03-10
Account activation link leads to blank page


ajrobbins:
When the account activation link is clicked, it leads to a blank page. However, the new account is activated on MMW. 

http://staging.app.wikiwatershed.org/accounts/activate/
![screen shot 2016-03-10 at 11 58 02 am](https://cloud.githubusercontent.com/assets/12868342/13677757/4de9b09a-e6b9-11e5-8343-885f64ed8b27.png)

Issue #1178, opened 2016-03-10
Password reset page not working


ajrobbins:
Clicking on the password reset link that MMW sends leads to a blank page:

https://mmw.azavea.com/accounts/password/reset/confirm/NjI5/
![screen shot 2016-03-10 at 11 58 02 am](https://cloud.githubusercontent.com/assets/12868342/13677351/9fac9dfe-e6b7-11e5-9ec6-96b88dc04b77.png)

mmcfarland:
Shares a point with #1179

PR #1183, opened 2016-03-10
Allow display of registration content

mmcfarland:
The previous map container was hidden during the transition to the
sidebar layout (c392ed6c75), but the registration content was not
updated to target the new container.  This restores the visibility of
the previous container, which is now only used for the account pages and
is empty in the main map mode.

Connects #1178 
Connects #1179 

#### To test:
* sign up for an account and use the activation link emailed (to console)
* Request to reset your password, follow the link emailed 
* Quick test of the map and micro apps should not have any altered layout issues.

Both account pages should display content instead of appearing blank with a header.

### DIFF
--- a/src/mmw/apps/core/templates/base.html
+++ b/src/mmw/apps/core/templates/base.html
@@ -33,7 +33,8 @@
         {% endblock map %}
     </div>
 
-    <div class="container-fluid top-nav" style="display: none;">
+    <div class="container-fluid top-nav">
+        <!-- Used on registration/account pages, not on main app -->
         {% block content %}
         {% endblock content %}
     </div>
--- a/src/mmw/apps/core/templates/base.html
+++ b/src/mmw/apps/core/templates/base.html
@@ -31,10 +31,8 @@
     <div class="map-container">
         {% block map %}
         {% endblock map %}
-    </div>
 
-    <div class="container-fluid top-nav">
-        <!-- Used on registration/account pages, not on main app -->
+        {# Used on registration/account pages, not on main app #}
         {% block content %}
         {% endblock content %}
     </div>


rajadain:
While the text boxes _are_ visible now, they are unfortunately not clickable. The `.map-container` is over them and doesn't allow any interaction. Looking into ways to fix this now ...

mmcfarland:
Maybe we should just move `{% block content %}` into the map-container along with `{% block map %}`?

rajadain:
I think you're right. This works well for me:

```diff
rajadain@work:~/dev/mmw (mmcfarland-feature/registration-hidden)
$ git diff
diff --git a/src/mmw/apps/core/templates/base.html b/src/mmw/apps/core/templates/base.html
index c0dddec..04a1611 100644
--- a/src/mmw/apps/core/templates/base.html
+++ b/src/mmw/apps/core/templates/base.html
@@ -31,10 +31,6 @@
     <div class="map-container">
         {% block map %}
         {% endblock map %}
-    </div>
-
-    <div class="container-fluid top-nav">
-        <!-- Used on registration/account pages, not on main app -->
         {% block content %}
         {% endblock content %}
     </div>
```

Also tested the main and micro apps a bit and they work.

mmcfarland:
Updated with fb1638f

src/mmw/apps/core/templates/base.html
```diff
@@ -31,9 +31,8 @@
     <div class="map-container">
         {% block map %}
         {% endblock map %}
-    </div>
 
-    <div class="container-fluid top-nav" style="display: none;">
+        <!-- Used on registration/account pages, not on main app -->
```
rajadain: Consider making this a Python template comment instead of an HTML comment so that it doesn't show up in the UI where it isn't used:

image

rajadain:
+1 tested, works well. Great work!

mmcfarland:
Good call, used a template comment instead.

Issue #1177, opened 2016-03-10
Slider styling issue on microsite


ajrobbins:
Potentially as a result of #1157, the precipitation slider on the microsite is appearing off center, and the title text is wrapping awkwardly on IE.

Chrome & Safari:
![screen shot 2016-03-10 at 10 14 29 am](https://cloud.githubusercontent.com/assets/12868342/13676076/562f4e10-e6b2-11e5-8fb1-929ef27a320f.png)

Internet Explorer:
![screen shot 2016-03-10 at 11 19 55 am](https://cloud.githubusercontent.com/assets/12868342/13676029/230e75ec-e6b2-11e5-8ef9-6126c944dc99.png)

Firefox:
![screen shot 2016-03-10 at 11 21 30 am](https://cloud.githubusercontent.com/assets/12868342/13676062/466c2732-e6b2-11e5-9a5c-1ff451a68706.png)

PR #1184, opened 2016-03-10
Fix Micro Site Slider

rajadain:
## Overview

The micro site's styling had been broken since we updated the opacity slider styling without caring for its specificity. By making the opacity slider styling more specific, and adding back lost micro site styling, we fix the precipitation slider in the micro site.

## Demo

Chrome:

image

Safari:

image

Firefox:

image

Internet Explorer:

image

## Testing Instructions

Test both opacity and precipitation sliders in multiple browsers.

Connects #1177

### DIFF
--- a/src/mmw/sass/components/_slider-native.scss
+++ b/src/mmw/sass/components/_slider-native.scss
@@ -5,6 +5,10 @@
     margin: -7px 0;
     cursor: pointer;
 
+    &:focus {
+        outline: none;
+    }
+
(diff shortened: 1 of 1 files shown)


ajrobbins:
Wrapping issue is still occurring on smaller screens on IE 
screen shot 2016-03-10 at 3 36 53 pm

rajadain:
Noted on #1185.

PR #1186, opened 2016-03-10
Use retires with exponential backoff when choosing workers

### LINKED ISSUE #1080: Protect against NoneType return Celery worker collection
If an attempt to assemble a list of active Celery workers fails, the application throws an exception. It is unclear whether the job still completes or not, but I would assume it does not:

```
Internal Server Error: /api/modeling/start/tr55/
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/views.py", line 452, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/views.py", line 449, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/decorators.py", line 50, in handler
    return func(*args, **kwargs)
  File "/opt/app/apps/modeling/views.py", line 267, in start_tr55
    task_list = _initiate_tr55_job_chain(model_input, job.id)
  File "/opt/app/apps/modeling/views.py", line 278, in _initiate_tr55_job_chain
    job_chain = _construct_tr55_job_chain(model_input, job_id)
  File "/opt/app/apps/modeling/views.py", line 285, in _construct_tr55_job_chain
    routing_key = choose_worker()
  File "/opt/app/apps/modeling/views.py", line 231, in choose_worker
    celery.current_app.control.inspect().ping().keys())
AttributeError: 'NoneType' object has no attribute 'keys'
```

hectcastro:
When the call to ping available workers is made, if no workers come back and the string of calls results in an `AttributeError`, retry the call for workers again.

The current parameters should retry after:

- 0.5s
- 1s (2 x 0.5)
- 2s (2 x 1), and then fail

Attempts to resolve #1080

### DIFF
--- a/src/mmw/apps/modeling/views.py
+++ b/src/mmw/apps/modeling/views.py
@@ -20,6 +20,8 @@ from django.contrib.gis.geos import GEOSGeometry
 import celery
 from celery import chain
 
+from retry import retry
+
 from apps.core.models import Job
 from apps.core.tasks import save_job_error, save_job_result
 from apps.modeling import tasks
@@ -235,8 +237,17 @@ def choose_worker():
     def predicate(worker_name):
         return settings.STACK_COLOR in worker_name or 'debug' in worker_name
 
+    @retry(Exception, delay=0.5, backoff=2, tries=3)
+    def get_list_of_workers():
+        workers = celery.current_app.control.inspect().ping()
+
+        if workers is None:
+            raise Exception('Unable to receive a PONG from any workers')
+
+        return workers.keys()
+
     workers = filter(predicate,
-                     celery.current_app.control.inspect().ping().keys())
+                     get_list_of_workers())
     return random.choice(workers)
 
 
--- a/src/mmw/requirements/base.txt
+++ b/src/mmw/requirements/base.txt
@@ -12,3 +12,4 @@ djangorestframework-gis==0.8.2
 tr55==1.1.3
 requests==2.9.1
 rollbar>=0.11.0,<=0.12.0
+retry==0.9.1


src/mmw/apps/modeling/views.py
```diff
@@ -235,8 +237,12 @@ def choose_worker():
     def predicate(worker_name):
         return settings.STACK_COLOR in worker_name or 'debug' in worker_name
 
+    @retry(AttributeError, delay=0.5, backoff=2, tries=3)
+    def get_list_of_workers():
+        return celery.current_app.control.inspect().ping().keys()
```
kdeloach: For the sake of having descriptive error messages, consider raising an exception if `celery.current_app.control.inspect().ping()` yields no results, instead of relying on `AttributeError`.

hectcastro: Addressed in https://github.com/WikiWatershed/model-my-watershed/commit/50eac071a2d5dd26853c9ea60bf8b1b5cfcfe50c.

src/mmw/apps/modeling/views.py
```diff
@@ -235,8 +237,17 @@ def choose_worker():
     def predicate(worker_name):
         return settings.STACK_COLOR in worker_name or 'debug' in worker_name
 
+    @retry(Exception, delay=0.5, backoff=2, tries=3)
```
kdeloach: I still got an Exception by calling this while restarting the celeryd service. Consider increasing the number of tries. What do you think about making the maximum retry duration around 10-15 seconds?

kdeloach:
+1 Great solution.

src/mmw/apps/modeling/views.py
kdeloach: Consider using 1.618 :cyclone: instead of 2 for a potentially closer fit.

hectcastro: I'm fine with tweaking the parameters, but I think we need to determine at what threshold the call becomes less likely to succeed at. Does waiting 10 seconds give us a greater chance of success?

I don't think there is a clear answer, but given what I've seen of this particular interaction, any more than 2-3 tries it probably too much. Somewhere between 3-5 seconds may be the max as far as overall duration goes. If people with more insight feel differently, I'm open to hearing why.

/cc @jamesmcclain because he's the original author, and might have a better sense.

kdeloach: Without any changes, this is a big improvement. We should wait until the error occurs again before tweaking these.

mmcfarland: I agree - let's see the extent to which we still experience problems before tweaking too much.

jamesmcclain: Late to the party, but that change looks good to me :+1:

Issue #1185, opened 2016-03-10
Changing Precipitation Slider has no effect in Internet Explorer


rajadain:
To reproduce:

 1. Open the micro site in IE
 2. Move the precipitation slider

Actual result:

 * Nothing happens

Expected result:

 * The values of runoff, evapotranspiration, and infiltration should change.

Seen on:

 * IE 11 on Windows 7

rajadain:
Also, there are wrapping issues, as noted by @ajrobbins on https://github.com/WikiWatershed/model-my-watershed/pull/1184#issuecomment-195036452:

> Wrapping issue is still occurring on smaller screens on IE 
> ![screen shot 2016-03-10 at 3 36 53 pm](https://cloud.githubusercontent.com/assets/12868342/13683687/09e7e64c-e6d6-11e5-9e2a-374920e28492.png)

SRGDamia1:
If you change the precip with the slider, and then change the land cover or soil type, it will properly recalculate for the correct precip.

PR #1188, opened 2016-03-11
Fix precipitation slider issues in IE

rajadain:
## Overview

Enables precipitation slider in IE by firing the `change` event, since IE does not support the `input` event. Also fixes the label wrapping issue which happened only when the vertical scrollbar was visible.

## Demo

mmw-ie-precipitation-slider-fix

Connects #1185


mmcfarland:
+1 Looks good in multiple browsers, including IE.

PR #1196, opened 2016-03-14
Remove DRB streams in favor on NHDPlus Flowlines

mmcfarland:
Updates the data loading script:
* Remove partial-DRB coverage of stream network
* Add national coverage of NHDPlus Flowlines

Connects #1194 

To test:

* Reload your services with more RAM (I used 5GB, but didn't try less)
* Load the stream network data via `$ ./scripts/setupdb.sh -s`
* Verify via pgsql or pgweb that the `nhdflowlines` table is created and loaded with `2993513` records and a spatial index on column "geom".

Once merged, I'll run this on our staging RDS instance prior to a PR that tiles this data so it can be tested.

Issue #1194, opened 2016-03-14
Upload NHD+ national stream data to staging


ajrobbins:
Following up on #1148

PR #1197, opened 2016-03-15
Add stream network to overlay layers

mmcfarland:
* Configure app & windshaft to serve and style nhdflowlines
* Consolidate generation of overlay list
* Renamed "stream slider" streams data to be DRB specific.  These may be removed/updated soon.
* Adds new category type to overlay layers

DRB Streams as-is will probably be removed in the near future as the feature
is deprecated.

Connects #1194 

To test, assumes you have `nhdflowlines` loaded (see #1196).

After bundling, on the Layer Selector Overlay panel, you should see a new category for Streams with an option to turn on National Stream Network

screenshot from 2016-03-15 10 47 41


mmcfarland:
Updated with fix for JS unit tests

lewfish:
+1, tested

Issue #1195, opened 2016-03-14
Enable DRB stream visualization


ajrobbins:
Upload TauDEM stream data for DRB (TauDEM 50). We've uploaded some data for this, but are not sure if its for the entire DRB or just a subset. We want to visualize this only when the user is viewing the DRB on the map, so there should be two radio buttons in the Overlays tab of the layer selector, one for NHD+ and one for TauDEM in the DRB (TauDEM would be inactive unless in the DRB).

mmcfarland:
One idea is to add something similar to the zoomLevel attribute for layers, which is "extent".  If the extent of the map window in entirely contained in the DRB shape (simplified & local), it could be enabled and disabled otherwise (and national disabled).

mmcfarland:
Some information from https://github.com/WikiWatershed/model-my-watershed/issues/764 may be useful.

PR #1204, opened 2016-03-18
DRB Stream Layer

lewfish:
This adds a layer for the streams (as predicted by TauDem) in the DRB, and is only selectable when the DRB is visible.

screen shot 2016-03-18 at 5 30 00 pm 2

##### Testing
* Copy this branch's version of `setupdb.sh` to a place accessible to psql (the app VM is good)
* Run `./setupdb.sh -d` to load data
* The DRB streams layer should only be selectable when the map viewport overlaps with the DRB.

Connects #1195


src/mmw/js/src/core/utils.js
```diff
@@ -42,6 +43,43 @@ var utils = {
...
+                viewportPolygon = {
+                    type: 'Feature',
+                    geometry: {
+                        type: 'Polygon',
+                        coordinates: [[
+                            [mapBounds.getNorthEast().lng, mapBounds.getNorthEast().lat],
+                            [mapBounds.getSouthEast().lng, mapBounds.getSouthEast().lat],
+                            [mapBounds.getSouthWest().lng, mapBounds.getSouthWest().lat],
+                            [mapBounds.getNorthWest().lng, mapBounds.getNorthWest().lat],
+                            [mapBounds.getNorthEast().lng, mapBounds.getNorthEast().lat]
```
kdeloach: Does the polygon not self-close if you omit the final point?

kdeloach:
This is really cool. :+1:

src/mmw/js/src/core/utils.js
lewfish: Apparently turf-intersect doesn't work if you omit that final point.

mmcfarland:
Looks good.  The network is so detailed, I think we may actually want to clamp down at `'minZoom': 11` in layer_settings until we can get the caching turned on.  

What's the total row count meant to be, 410,607?

src/mmw/mmw/settings/layer_settings.py
```diff
@@ -119,6 +128,16 @@
         'minZoom': 10
     },
     {
+        'code': 'drb_streams',
+        'display': 'DRB Stream Network',
+        'table_name': 'drb_streams_50',
+        'boundary': False,
+        'stream': True,
+        'overlay': True,
+        'perimeter': drb_perimeter # Layer is only selectable when viewport
```
mmcfarland: Sorry, with more context.  I think we should add a `'minZoom': 11` to this block.

lewfish:
> Looks good. The network is so detailed, I think we may actually want to clamp down at 'minZoom': 11 in layer_settings until we can get the caching turned on.
> 

Will do.

> What's the total row count meant to be, 410,607?

I have 586,285 rows. I ran `SELECT COUNT(*) FROM drb_streams_50;`

mmcfarland:
I had the old DRB streams table and had the count off of that, 586,285 confirms.

src/mmw/mmw/settings/production.py
```diff
@@ -105,7 +105,6 @@
     'LayerAttribution',
     'LayerSelector',
     'LocateMeButton',
-    # 'StreamControl',
```
mmcfarland: There is a similar configuration still in `settings/base.py` and `core/tests.js`

kdeloach:
+1 I see streams.

Issue #1181, opened 2016-03-10
Error popup when you cancel "Delineate Watershed" drawing


kdeloach:
If you press Esc to cancel placing a stream network watershed delineation marker, you'll get an error popup "Unable to delineate watershed at this location".

Caused by #1175.

kellyi:
Will work on this one since it's adjacent to something I was working on in #1160

kdeloach:
After adding the RWD alert message, I realized that we don't have a way to distinguish a failure from a cancellation for drawing actions. 

Normal draw actions (square mile and free draw) can't really "fail", but they can be cancelled with Esc. Delineating a point will fail to return a polygon if RWD throws an error.

I might be able to help with this if you need ideas.

kellyi:
Sounds good. 

I was thinking of maybe having the RWD alert only show if the failed action has sent it a message to display; otherwise it'd just "fail" silently -- which is how you'd want it to fail on Esc or clicking Reset. 

Upside's that the failure messages could be a little more specific to each action; downside's that they'd need to be added for all failed actions that should show a message.

Issue #1160, opened 2016-02-19
Exception: Point located outside the watershed.


hectcastro:
View details in Rollbar: [https://rollbar.com/Azavea/MMW/items/35/](https://rollbar.com/Azavea/MMW/items/35/)

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "apps/modeling/tasks.py", line 42, in start_rwd_job
    raise Exception(response_json['error'])
Exception: Point located outside the watershed.
```

mmcfarland:
This is going to occur when an RWD request is made in an area oustide of coverage, and is currently handled in the UI in a too subtle way:

![screenshot from 2016-02-19 16 17 37](https://cloud.githubusercontent.com/assets/1014341/13189276/665db378-d724-11e5-8778-856d35a94f63.png)

Since this is a known limitation of the RWD project, we should probably handle this in a way that doesn't require raising, even though the error case is expected.

Ideally, this is handled as input validation by doing a point in polygon check that the coordinates are inside the DRB boundary, and not allowing the request to make it all the way to the RWD service in that case.

PR #1206, opened 2016-03-22
Client-side validation that user-selected point is within DRB

kellyi:
This PR adds some client-side validation that the user-selected point for "Delineate Watershed" is within the Delaware River Basin. If the point's not in the DRB, the app will immediate show an alert message to explain the error without making a request to the server. This one depends on @lewfish PR #1204.

I also rewrote how the flow of "Delineate Watershed" handles displaying error message alerts: the general `.fail` case now accepts an optional `message` argument. If there's a message, it will get displayed as an alert; if not, the failure will pass by silently. I added error messages to each possible failure that should show them. For a "failure" like the one mentioned in #1181 where the user has hit Escape or clicked Reset to cancel selecting a point, everything will just reset silently.

**Testing Instructions**

- click Delineate Watershed > Stream Network, then select a point in What Cheer, Iowa (or, really, anywhere not in the DRB). You should see this alert, and you should *not* see a failure icon on the "Delineate Watershed" button:

screen shot 2016-03-22 at 10 29 43 am

- click OK and everything should reset

- click Delineate Watershed > Stream Network and then hit Escape or click Reset. You shouldn't see an alert message and everything should reset.

- turn on Chrome's Network Link Conditioner to simulate being offline altogether, then try to select a point within the DRB. You should see something like:

screen shot 2016-03-22 at 10 34 49 am

& everything should reset.

There are specific messages for a few additional possible failures, and a generic "Unable to delineate watershed at this location" message for others.

Connects #1181 
Connects #1160

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -7,6 +7,7 @@ var $ = require('jquery'),
     turfArea = require('turf-area'),
     turfBboxPolygon = require('turf-bbox-polygon'),
     turfDestination = require('turf-destination'),
+    turfIntersect = require('turf-intersect'),
     router = require('../router').router,
     App = require('../app'),
     utils = require('./utils'),
@@ -69,6 +70,19 @@ function validateShape(polygon) {
         d.reject();
     } else {
         d.resolve(polygon);
+    }
+    return d.promise();
+}
+
+function validateClickedPointWithinDRB(latlng) {
+    var point = L.marker(latlng).toGeoJSON(),
+        d = $.Deferred(),
+        boundaries = settings.get('boundary_layers'),
+        drbPerimeter = _.findWhere(boundaries, {code:'drb_streams'}).perimeter;
+    if (turfIntersect(point, drbPerimeter)) {
+        d.resolve(latlng);
+    } else {
+        d.reject(latlng);
     }
     return d.promise();
 }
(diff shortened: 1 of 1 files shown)


kdeloach:
+1 Great solution.

src/mmw/js/src/draw/views.js
```diff
@@ -73,6 +74,19 @@ function validateShape(polygon) {
     return d.promise();
 }
 
+function validateClickedPointWithinDRB(latlng) {
+    var point = L.marker(latlng).toGeoJSON(),
+        d = $.Deferred();
+    var boundaries = settings.get('boundary_layers');
+    var drbPerimeter = _.findWhere(boundaries, {code: 'drb_streams'}).perimeter;
```
kdeloach: One of our style conventions is to use a single `var` declaration when we need to define multiple variables like this.

src/mmw/js/src/draw/views.js
```diff
@@ -51,7 +51,8 @@ function validateRwdShape(result) {
             })
             .fail(d.reject);
     } else {
-        d.reject(result);
```
lewfish: Consider using a shorter commit title so that it doesn't wrap around.

lewfish:
+1, good job!

kellyi:
Thanks for the notes! I updated the variable declaration.

mmcfarland:
+1

kdeloach:
Consider squashing that last commit.

Issue #1201, opened 2016-03-16
Stub out Celery Workflow for MapShed


rajadain:
Includes the end-point to be hit when starting off the MapShed job

PR #1209, opened 2016-03-23
Stub out Mapshed endpoint

lewfish:
This adds a stubbed out endpoint for starting Mapshed jobs, modeled after the one for RWD, since it isn't clear what the inputs from the UI are yet. To test, run `./debugcelery.sh` and `http -f POST http://localhost:8000/api/modeling/start/mapshed/ input=5`. Then run `http http://localhost:8000/api/modeling/jobs/<job_uuid>/` which should return results containing `{ input: 5 }`.

Connects #1201


kdeloach:
+1

PR #1215, opened 2016-03-30
Add S3 backed persistent tile cache

### LINKED ISSUE #727: Add tilecache bucket config to CloudFormation
The tilecache s3 bucket is manually configured with the following options:

##### Bucket Policy
```json
{
	"Version": "2012-10-17",
	"Statement": [
		{
			"Sid": "AddPerm",
			"Effect": "Allow",
			"Principal": "*",
			"Action": "s3:GetObject",
			"Resource": "arn:aws:s3:::tilecache.mmw-dev.azavea.com/*"
		}
	]
}
```

##### CORS Configuration
```xml
<?xml version="1.0" encoding="UTF-8"?>
<CORSConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
    <CORSRule>
        <AllowedOrigin>*</AllowedOrigin>
        <AllowedMethod>GET</AllowedMethod>
        <MaxAgeSeconds>3000</MaxAgeSeconds>
        <AllowedHeader>*</AllowedHeader>
    </CORSRule>
</CORSConfiguration>
```

##### Static Website Routing Rules
```xml
<RoutingRules>
    <RoutingRule>
        <Condition>
            <HttpErrorCodeReturnedEquals>404</HttpErrorCodeReturnedEquals>
        </Condition>
        <Redirect>
            <HostName>4e139ca.ngrok.com</HostName>
        </Redirect>
    </RoutingRule>
</RoutingRules>
```

These should be scripted within the CloudFormation setup so the resources are available to update on deployment.  Specifically the `HostName` needs to be set to the appropriate color stack tiler.  The VPCID may also be used to create a bucket name that can be generated in the app and be distinct between prod/stage environments within the AWS account.

hectcastro:
In order to reduce the percentage of requests that cause tiles to be rendered on the fly, introduce a lazily built persistent tile cache in Amazon S3.

**Tile requests used to be served in this fashion**:

- Request to CloudFront is served by CloudFront edge cache
- Request to CloudFront passes through to origin (tiler ELB) and response comes back through CloudFront

**New tile request behavior**:

- Request to CloudFront is...
  - Served by CloudFront edge cache (best case scenario)
  - Served by `tile-cache` S3 website, which is the CloudFront origin (second best case scenario)
  - Redirected to `https://color-tiles...` (tiler ELB) and client processes redirect

In the last scenario, 404s against the `tile-cache` S3 website cause the redirects to `https://color-tiles...`. That endpoint is the tile ELB, which is responsible for rendering a tile in the response and to S3 (using the AWS SDK for Node.js).

Resolves #727 

---

**Observed behavior**:

- You try to load tiles that don't exist in the CloudFront or S3 cache
- CloudFront redirects you to `https://color-tiles...`
- Subsequent requests may temporarily continue to redirect if cached by browser or CloudFront (headers give this away)
- After that cache expires, requests should 200 directly from CloudFront (either by its edge cache or the S3 tile cache website)

**Issues**:

- [x] Grid JSON appears to be failing for National Streams, and I haven't figured out why

```json
{
	"error": "Missing interactivity parameter"
}
```


src/tiler/server.js
```diff
@@ -103,9 +103,11 @@ var config = {
             }
 
             if (body) {
-                s3Obj
-                    .upload({Body: body})
-                    .send(function(err, data) { throw (err); });
```
hectcastro: Multiple tile requests at a time was causing this to fail with a `Cannot read property 'message' of null` deep in the AWS SDK. I messed around with updating the SDK in case it was a bug, but that didn't resolve it. After tweaking the way the function was written the errors went away. Was unsure about what `send()` was doing here.

hectcastro:
**Expected behavior**:

- You try to load tiles that don't exist in the CloudFront or S3 cache
- CloudFront redirects you to a `https://color-tiles...` URL; that same process writes tiles to S3
- Subsequent requests may temporarily continue to redirect if cached by browser or CloudFront
- After that cache expires, requests should 200 directly from CloudFront (these are coming from S3)

**Issues**:

- [ ] Grid JSON appears to be failing, and I haven't figured out why

deployment/ansible/group_vars/test
```diff
@@ -22,7 +22,7 @@ celery_processes_per_worker: 1
 itsi_base_url: "https://learn.staging.concord.org/"
 itsi_secret_key: "{{ lookup('env', 'MMW_ITSI_SECRET_KEY') }}"
 
-tilecache_bucket_name: "tilecache.mmw-dev.azavea.com"
+tilecache_bucket_name: "tile-cache.staging.app.wikiwatershed.org"
```
hectcastro: In reality this is dynamic based on overall stack domain, so tile cache buckets for staging and production are separate.

src/tiler/server.js
mmcfarland: Not sure.  I took that nearly verbatim from the SDK example page for uploading a stream (http://docs.aws.amazon.com/AWSJavaScriptSDK/guide/node-examples.html).  If it works better without it, I can't explain.

mmcfarland:
I can confirm that the tiles are loading correctly and are writing to s3.  After disabling the browser cache, tiles appear to be served from cache based on speed, but I can't confirm where they're being served from.  I had a look through the headers but still see `X-Powered-By:Express` which I thought was unusual.  Is there way to confirm a response is coming from windshaft vs cloudfront vs s3?

Also tested that purge setupdb.sh script worked correctly.

src/tiler/server.js
hectcastro: Perhaps the original code was executing differently than the example?

```javascript
s3obj.upload({Body: body}).
  on('httpUploadProgress', function(evt) { console.log(evt); }).
  send(function(err, data) { console.log(err, data) });
```

Looks like that has `on()` chained with `send()`, but the original didn't. Not super familiar with how this chaining behaves, but potentially `send()` was being called in cases where there was no data?

hectcastro:
For anyone else following along, if CloudFront doesn't issue a 302, then the request is either being served by it, or the S3 cache.

- If `X-Cache: Miss from cloudfront` is in the response headers, then it came from the S3 cache
- If `X-Cache: Hit from cloudfront` is in the response headers, then it came from CloudFront's edge cache

In Matt's case, I think he had his Inspector tab on `Img`, which hid the 302s from CloudFront.

mmcfarland:
+1. Not filtering `img` in my dev tools showed the expected responses and redirects.  Everything looks wired up correctly to me.

Issue #1217, opened 2016-03-31
Streams should not be selectable boundaries


mmcfarland:
National and DRB streams are showing as AoI Boundary options, which they are not.  These need to be removed, but retained in the Layer Selector overlay panel.

![screenshot from 2016-03-31 11 25 10](https://cloud.githubusercontent.com/assets/1014341/14181082/9bc483a0-f733-11e5-978e-1af0ace78d6d.png)

PR #1218, opened 2016-03-31
Remove stream layers from boundary selection

mmcfarland:
The code which filters layers based on layer attributes uses the
existance of the key, not the truthiness of it, to determine inclusion.
Remove the `boundary: False` value to remove from the app.

Connects #1217 

##### Test
Start the app and see that Streams and DRB Stream are *not* available in the "Select by Boundary" dropdown, but are still preserved in the Layer Selector Overlay panel.

screenshot from 2016-03-31 13 10 35


kellyi:
Checked and confirmed! +1

Issue #1200, opened 2016-03-16
Add UI Workflow for Mapshed Model


rajadain:
Currently there is only one model in the system. Setup a UI workflow that allows for an additional model, namely MapShed, and allows the user to pick one after the Analyze view. Follow wireframes developed by jfrankl.

This will include reworking the Backbone views and potentially models in the front-end app.

PR #1219, opened 2016-03-31
Select GWLF-E in UI

lewfish:
The user can now select the model package (TR-55 or GWLF-E) for a project. The dropdown will be replaced by a better one after the prototype is finished. 

To test, try with `disabled` set to both `true` and `false` [here](https://github.com/lewfish/model-my-watershed/blob/feature/select-gwlfe/src/mmw/js/src/modeling/models.js#L27). TR-55 should work as before, and GWLF-E should poll the gwlfe endpoint, which will return empty results, leading to blank Runoff and Water Quality views and error messages in the console. Hitting back in the model view should work as before. Also, test while logged in.

screen shot 2016-04-01 at 1 47 37 pm 2

Connects #1200

### DIFF
--- a/src/mmw/apps/home/urls.py
+++ b/src/mmw/apps/home/urls.py
@@ -12,6 +12,7 @@ urlpatterns = patterns(
     url(r'^$', home_page, name='home_page'),
     url(r'^projects/$', projects, name='projects'),
     url(r'^project/$', project, name='project'),
+    url(r'^project/new/', project, name='project'),
     url(r'^project/(?P<proj_id>[0-9]+)/$', project, name='project'),
     url(r'^project/(?P<proj_id>[0-9]+)/clone/?$',
         project_clone, name='project_clone'),
--- a/src/mmw/js/src/analyze/templates/resultsWindow.html
+++ b/src/mmw/js/src/analyze/templates/resultsWindow.html
@@ -7,14 +7,31 @@
             </a>
         </li>
         <li>
-            <a aria-controls="home" aria-expanded="false">
+            <a class="back-button" aria-controls="home" aria-expanded="false">
                 <button type="button" class="btn btn-lg" data-url="/">
                     Back
                 </button>
-                <button type="button" class="btn btn-lg btn-active" data-url="/project">
-                    Model
-                </button>
             </a>
+
+            <div class="dropdown">
+              <button class="btn btn-lg btn-active dropdown-toggle" type="button" data-toggle="dropdown">Model
+              <span class="caret"></span></button>
+              <ul class="dropdown-menu dropdown-menu-right">
+                {% for modelPackage in modelPackages %}
+                    <li>
+                        {% if modelPackage.disabled %}
+                            <a class="disabled" href="#">
+                                {{ modelPackage.display_name }} (Coming Soon)
+                            </a>
+                        {% else %}
+                            <a href="#" data-url="/project/new/{{ modelPackage.name }}">
+                                {{ modelPackage.display_name }}
+                            </a>
+                        {% endif %}
+                    </li>
+                {% endfor %}
+              </ul>
+            </div>
         </li>
     </ul>
 </div>
--- a/src/mmw/js/src/modeling/controllers.js
+++ b/src/mmw/js/src/modeling/controllers.js
@@ -121,13 +121,7 @@ var ModelingController = {
                 var lock = $.Deferred();
 
                 if (!App.currentProject) {
-                    // Only make new project if this is the first time
-                    // hitting the modeling views.
-                    if (!App.projectNumber) {
-                        project = makeProject(lock);
-                    } else {
-                        project = reinstateProject(App.projectNumber, lock);
-                    }
+                    project = reinstateProject(App.projectNumber, lock);
 
                     App.currentProject = project;
                     if (!App.projectNumber) {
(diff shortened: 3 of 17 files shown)


src/mmw/apps/home/urls.py
```diff
@@ -12,8 +12,9 @@
     url(r'^$', home_page, name='home_page'),
     url(r'^projects/$', projects, name='projects'),
     url(r'^project/$', project, name='project'),
+    url(r'^project/new/', project, name='project'),
     url(r'^project/(?P<proj_id>[0-9]+)/$', project, name='project'),
-    url(r'^project/(?P<proj_id>[0-9]+)/clone/?$',
+    url(r'^project/(?P<proj_id>[0-9]+)/clone/?$~',
```
kdeloach: What does `~` do?

lewfish: Good catch. That's a typo.

lewfish:
Hold of on testing this for now. I found a bug when testing it while logged in that I need to fix.

mmcfarland:
Can you add a flag in the settings for whether to enable the GWLFE option in that list (Similar to the erstwhile StreamSlider)?  It could additionally have "(Coming Soon)" added to the disabled text.  Also I think, even to have this displayed temporarily, we should use the actual model names TR55 = "Site Storm Model", GWLF = "Watershed Multi-year Model"

We'll want to be able to merge into `develop` and `release` without maintaining a long running branch for this in-progress code.

lewfish:
@azaveaci rebuild

lewfish:
@azaveabot rebuild

lewfish:
I updated the commits according to the above feedback, and got things to work when logged in.

src/mmw/js/src/analyze/templates/resultsWindow.html
```diff
@@ -7,18 +7,36 @@
...
+                        {% endif %}
+                    </li>
+                {% endfor %}
+              </ul>
+            </div>
         </li>
     </ul>
 </div>
 
+
```
rajadain: Is this extra newline intentional?

src/mmw/js/src/analyze/templates/resultsWindow.html
```diff
@@ -7,18 +7,36 @@
...
+              <ul class="dropdown-menu dropdown-menu-right">
+                {% for modelPackage in modelPackages %}
+                    <li>
+                        {% if modelPackage.disabled %}
+                            <a class="disabled" href="#">
+                                {{ modelPackage.displayName }} (Coming Soon)
+                            </a>
+                        {% else %}
+                                <a href="#" data-url="/project/new/{{ modelPackage.name }}">
+                                    {{ modelPackage.displayName }}
```
rajadain: Inconsistent indentation

src/mmw/apps/home/urls.py
```diff
@@ -12,6 +12,7 @@
     url(r'^$', home_page, name='home_page'),
     url(r'^projects/$', projects, name='projects'),
     url(r'^project/$', project, name='project'),
+    url(r'^project/new/', project, name='project'),
```
rajadain: Is this just a fallback, since there is no actual functionality specific to `/project/new` in the back-end?

lewfish: Yes, it just redirects to /.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -139,37 +133,31 @@ var ModelingController = {
...
 
-                lock.done(function() {
-                    project.on('change:id', updateUrl);
-                    initScenarioEvents(project);
-                    initViews(project);
-                    if (App.projectNumber) {
-                        updateUrl();
-                    }
-                });
+                finishProjectSetup(lock, project);
```
rajadain: In some functions `lock` is the first argument, and in others it is the second. Consider following a consistent convention.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -139,37 +133,31 @@ var ModelingController = {
...
-            // arrow buttons (see issue #690).
-            App.projectNumber = scenarios.at(0).get('project');
-        }
+        projectCleanUp();
+    },
 
-        App.getMapView().updateModifications(null);
-        App.rootView.subHeaderRegion.empty();
-        App.rootView.sidebarRegion.empty();
+    makeNewProjectCleanUp: function() {
```
rajadain: Where is this used?

lewfish: It gets called by the router when exiting the route.

lewfish: The function name is constructed programmatically, so it doesn't show up if you search the source code.

rajadain: Ah, magic

rajadain:
Behaviorally works pretty great. Going back and forth creates a new project every time I pass "Model", which is the right behavior for most cases.

For ITSI, passing "Model" should _not_ create a new project, but delete and replace the existing one. This may be complex and warrant a card of its own. I think the behavior should be:

 * If the user has created a project using Model A, goes back to Analyze, and then clicks on Model A again, they should return to existing project.
 * If the user has created a project using Model A, goes back to Analyze, and then clicks on Model B, they should be given a warning that this will kill their existing work, which is in Model A, and they should choose between preserving the project and proceeding with Model A, or killing it and proceeding with Model B.
 * If the user has created a project using Model A, goes back to Analyze, goes back to Draw, and changes their AoI, then they should be given a warning that their project will be killed, and if they agree, we kill the project and let them proceed with whichever model they like.

src/mmw/js/src/modeling/models.js
```diff
@@ -10,6 +10,24 @@ var Backbone = require('../../shim/backbone'),
...
+    {
+        name: TR55_PACKAGE,
+        displayName: 'Site Storm Model',
+        description: ''
+    },
+    {
+        name: GWLFE,
+        displayName: 'Watershed Multi-Year Model',
+        description: '',
+        disabled: true
```
rajadain: Consider making this a setting somewhere in [`base.py`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/base.py), and threading this through via [`window.clientSettings`](https://github.com/WikiWatershed/model-my-watershed/blob/e1a86e60f3bbc0ae26433f891437414129c281cd/src/mmw/apps/core/templates/base.html#L60).

rajadain:
Other than the comments above, the code looks good and the behavior makes sense. +1!

rajadain:
I would recommend splitting out the ITSI stuff into its own card to be tackled later.

src/mmw/js/src/modeling/models.js
lewfish: I thought about doing this, but the backend doesn't need access to this list of allowed packages, and I would have to define the constants again on the frontend.

rajadain: True, but if this is acting as a feature flag, which we turn on for development but off for staging / production deploys, it makes sense to keep it with all other feature flag-like settings, so there's one place for everything.

lewfish:
Thanks for the review. I pushed a new commit to address your feedback.

rajadain:
+1, sweet fixes!

rajadain:
I think this is causing the tests to fail:

```
$ ./scripts/npm.sh run lint
+ ARGS='run lint'

js/src/analyze/views.js: line 10, col 5, 'modelingModels' is defined but never used.

1 error
```

lewfish:
@azavea-bot rebuild

lewfish:
See https://github.com/WikiWatershed/model-my-watershed/issues/1222

rajadain:
:+1:

PR #1221, opened 2016-04-05
Fix DRB detection for stream delineation feature.

caseypt:
* DRB is not a boundary layer, but a stream layer. This worked
previously because of this bug:
https://github.com/WikiWatershed/model-my-watershed/pull/1218

**Testing**
- Verify that when you click to start an RWD job, an error is not thrown, and that the point is checked to see if it is in the DRB.

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -77,8 +77,8 @@ function validateShape(polygon) {
 function validateClickedPointWithinDRB(latlng) {
     var point = L.marker(latlng).toGeoJSON(),
         d = $.Deferred(),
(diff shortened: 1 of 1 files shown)


rajadain:
Haven't tested, but code looks correct. It should select the right one:

image

rajadain:
+1 tested. No JavaScript errors anymore, although I did get this box afterwards:

image

caseypt:
Thanks. I get that error as well because of https://github.com/WikiWatershed/rapid-watershed-delineation/issues/26

rajadain:
Great. Good to merge then. Nice work finding and fixing this.

Issue #1228, opened 2016-04-07
MapShed Configuration


rajadain:
MapShed uses a number of static variables, whose value has most likely been tuned for certain settings. In order for further tuning to be easy, we should keep all such static values in one place, and structure them in an easily consumable format.

Develop a structure for storing MapShed variables in one place, and thread them through to other services so they can be changed in one place and accessed everywhere.

Issue #1216, opened 2016-03-30
Expand Celery stub to use GWLF-E (Phase 1)


rajadain:
Using the GMS Key, expand the Celery workflow for GWLF-E to create a Python dictionary, populated with Static values, and submit to the model.

rajadain:
-3 to https://github.com/WikiWatershed/mmw-geoprocessing/issues/25

PR #1235, opened 2016-04-08
Basic GWLF-E Static Output

rajadain:
## Overview

 * Organizes all static values for GWLF-E in `gwlfe_settings.py`.
 * Expands celery stub to actually return values
 * Imports GWLF-E module from GitHub (since it is not published to PyPI yet)
 * Instantiates a GWLF-E model and returns it directly (without running the model)

## Testing Instructions

 1. After checking out the repo, either reprovision `app` and `worker`, or run these commands to update the Python dependencies:

        $ vagrant ssh app -c 'sudo pip install -r /opt/app/requirements/development.txt'
        $ vagrant ssh worker -c 'sudo pip install -r /opt/app/requirements/development.txt'

 2. Open MMW. Proceed to Model after Analyze by selecting "Watershed Multi-Year Model"
 3. In your browser's network request inspector tool, see the Celery job return a GWLF-E data model:

    image

## Questions

 * Is 1.11.0 the right version of `numpy` to use?
 * Are the GWLF-E defaults readable and in accordance with MMW conventions?

Connects #1216 
Connects #1228

### DIFF
--- a/src/mmw/mmw/settings/gwlfe_settings.py
+++ b/src/mmw/mmw/settings/gwlfe_settings.py
@@ -341,45 +341,38 @@ GWLFE_DEFAULTS = {
     'AWMSNgPct': 0,  # Animal Waste Management Systems: Poultry (%)
     'RunContPct': 0,  # Runoff Control (%)
     'PhytasePct': 0,  # Phytase in Feed (%),
-    'AnimalName': ['Dairy Cows', 'Beef Cows', 'Broilers', 'Layers',
-                   'Hogs/Swine', 'Sheep', 'Horses', 'Turkeys', 'Other'],
-    'NumAnimals': np.zeros(9),
-    'GrazingAnimal': ['Y', 'Y', 'N', 'N',
-                      'N', 'Y', 'Y', 'N', 'N'],
-    'AvgAnimalWt': np.array([640, 360, 0.9, 1.8,
-                             61, 50, 500, 6.8, 0]),  # Average Animal Weight (kg) # NOQA
-    'AnimalDailyN': np.array([0.44, 0.31, 1.07, 0.85,
-                              0.48, 0.37, 0.28, 0.59, 0]),  # Animal Daily Loads: Nitrogen (kg/AEU) # NOQA
-    'AnimalDailyP': np.array([0.07, 0.09, 0.30, 0.29,
-                              0.15, 0.10, 0.28, 0.59, 0]),  # Animal Daily Loads: Phosphorus (kg/AEU) # NOQA
-    'FCOrgsPerDay': np.array([1.0e+11, 1.0e+11, 1.4e+8, 1.4e+8,
-                              1.1e+10, 1.2e+10, 4.2e+8, 9.5e+7, 0]),  # Fecal Coliforms (orgs/day) # NOQA
-    'Month': ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun',
-              'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'],
-    'NGPctManApp': np.array([0.01, 0.01, 0.15, 0.10, 0.05, 0.03,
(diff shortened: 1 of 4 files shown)


src/mmw/mmw/settings/gwlfe_settings.py
```diff
@@ -0,0 +1,394 @@
...
+    'RetentPLake': 0.29,  # Lake Retention Rate: Phosphorus
+    'RetentSedLake': 0.84,  # Lake Retention Rate: Sediment
+    'AttenFlowDist': 0,  # Attenuation: Flow Distance (km)
+    'AttenFlowVel': 4,  # Attenuation: Flow Velocity (km/hr)
+    'AttenLossRateN': 0.287,  # Attenuation: Loss Rate: Nitrogen
+    'AttenLossRateP': 0.226,  # Attenuation: Loss Rate: Phosphorus
+    'AttenLossRateTSS': 0,  # Attenuation: Loss Rate: Total Suspended Solids
+    'AttenLossRatePath': 0,  # Attenuation: Loss Rate: Pathogens
+    'StreamFlowVolAdj': 1,  # Streamflow Volume Adjustment Factor
+}
```
kdeloach: I would prefer if most of these existed in gwlf-e so we don't have to update these default values in 2 places. There are also values that are more or less static, which can be omitted (ex. NRur/NUrb).

rajadain: I talked a bit with @mmcfarland about it, and the consensus seemed to be to follow the previous GMS file interface. Since these defaults were defined pre-GMS in the original implementation, we should keep them in the corresponding pre-dictionary phase here.

I guess I always imagined the model itself to be pretty much an unconfigured tool, with all intelligence of how to run it kept outside the model itself.

But I think this is an important point and probably warrants more discussion.

mmcfarland: Ultimately, I think it is problematic if these are defined in two projects (even originally), but also agree that _outside_ of the model seems to be the most logical place for them to reside.  If one were to tweak these initial values, it should be as input to the model instead of having to update and re-release the model itself.   Given the large quantity of these values and the potential that they don't 100% overlap with initialized values in gwlf-e, I think we should keep the logic of the original of both projects to not accidentally let a value fall through the cracks.

src/mmw/mmw/settings/gwlfe_settings.py
```diff
@@ -0,0 +1,394 @@
...
+    'NumAnimals': np.zeros(9),
+    'GrazingAnimal': ['Y', 'Y', 'N', 'N',
+                      'N', 'Y', 'Y', 'N', 'N'],
+    'AvgAnimalWt': np.array([640, 360, 0.9, 1.8,
+                             61, 50, 500, 6.8, 0]),  # Average Animal Weight (kg) # NOQA
+    'AnimalDailyN': np.array([0.44, 0.31, 1.07, 0.85,
+                              0.48, 0.37, 0.28, 0.59, 0]),  # Animal Daily Loads: Nitrogen (kg/AEU) # NOQA
+    'AnimalDailyP': np.array([0.07, 0.09, 0.30, 0.29,
+                              0.15, 0.10, 0.28, 0.59, 0]),  # Animal Daily Loads: Phosphorus (kg/AEU) # NOQA
+    'FCOrgsPerDay': np.array([1.0e+11, 1.0e+11, 1.4e+8, 1.4e+8,
```
rajadain: This set of variables (`AnimalName` - `FCOrgsPerDay`) is a little hard to keep track of and validate, since each array contains one slice of data, linked by index. While the model does expect the values in this format, I will look into refactoring this into a structure that keeps linked data together.

lewfish:
+1, tested

src/mmw/mmw/settings/gwlfe_settings.py
kdeloach: Sounds good.

kdeloach:
+1

Issue #1155, opened 2016-02-16
KeyError: u'area'


hectcastro:
View details in Rollbar: [https://rollbar.com/Azavea/MMW/items/33/](https://rollbar.com/Azavea/MMW/items/33/)

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "apps/modeling/tasks.py", line 214, in run_tr55
    area = piece['area']
KeyError: u'area'
```

mmcfarland:
We may want to try and get contextual information about saved `projectId`, if available, for this exception to help debugging.

mmcfarland:
I'm pretty sure this is caused by a BMP drawn over another BMP.  One BMP will be clipped, and I assume the `area` attribute is not preserved in the resulting product, causing this error.  The `area` attribute should be applied post-clipping to ensure this doesn't happen.

mmcfarland:
Not all combinations of LU/BMP will cause this.  This sequence will reliably reproduce the issue:

1. Draw a Forest reclassification
2. Partially intersect that polygon with a Green Roof

PR #1236, opened 2016-04-11
Process overlapping BMPs/land usage polygons

jtarricone:
This fixes a bug that prevented proper model processing for scenarios in which land usage / BMP polygons overlapped. When modelling a scenario that involved 2+ overlapping polygons from both the "Land Cover" and "Conservation Practice" categories, the server will not return any data and the UI component will be empty. This appeared to be the case for any combination of the different land/BMP types.

The problem was a type mismatch, due to the client-side app sending an inconsistently typed 'value', which was eventually caught on the server. The necessary change required the client to send a consistent type, as well as the server to 'expect' a different (but constant) type. Also notable was that the client-side app didn't send a value for 'area' for the overlapping sections, which was also caught on the server and similarly prevented any data being returned to the client.

**To test this as a fix**: create modelling scenarios that have at least 1 overlapping area b/w a land cover area and a conservation area. The original issue #1155  indicated that a land cover of "forest" and a conservation area of "green roof" produced the error, and I found that any random combination I tried produced the error as well.

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -389,6 +389,7 @@ function _alterModifications(rawModifications) {
                         overlapPiece.value.reclass = newPiece.value;
                     }
 
+                    overlapPiece.area = turfArea(overlapPiece.shape);
                     pieces.push(overlapPiece);
 
                     /* remove the overlapping portion from both parents */
@@ -413,7 +414,20 @@ function _alterModifications(rawModifications) {
         pieces = _.filter(pieces, validShape);
     }
 
-    return pieces;
+    return _.map(pieces, function(piece) { //ensures 'value' is uniform type for each piece
+      var p = piece;
+      if ( typeof p.value === 'string') {
+        var v = {};
+        if (p.name === reclass) { 
+          v.reclass = p.value; 
+        }
+        else if (p.name === bmp) { 
+          v.bmp = p.value; 
+        }
+        p.value = v;
+      }
+      return p;
+    });
 }
 
 var alterModifications = _.memoize(_alterModifications, function(_, hash) {return hash;});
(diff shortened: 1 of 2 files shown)


rajadain:
I think the build is failing because of a JSHint issue:

```
$ ./scripts/npm.sh run lint
+ ARGS='run lint'

js/src/modeling/models.js: line 412, col 29, Expected '===' and instead saw '=='.

1 error
```

src/mmw/js/src/modeling/models.js
```diff
@@ -413,7 +414,16 @@ function _alterModifications(rawModifications) {
         pieces = _.filter(pieces, validShape);
     }
 
-    return pieces;
+    return _.map(pieces, function(piece) { //ensures 'value' is uniform type for each piece
+      var p = piece;
+      if ( typeof p.value =='string') {
+        var v = {};
+        if (p.name === reclass) { v.reclass = p.value; }
```
rajadain: Prefer multiline / indented `if`, even for one line clauses, for readability:

```js
if (p.name === reclass) {
    v.reclass = p.value;
} else if (p.name === bmp) {
    v.bmp = p.value;
}
```

jtarricone: Updated based on both comments. I didn't see your second comment until after I'd pushed the first change, so there are 2 commits.

rajadain:
To link pull requests to the original issues which they address, we add a "Connects <Issue Number>" as the last line of the description. This links them in Waffle. For example:

Connects #1155

rajadain:
Also consider adding some instructions on how to verify the fix.

rajadain:
Tested with multiple overlapping Land Cover and Conservation Practice modifications:

image

All model runs completed successfully. +1

rajadain:
Consider squashing the commits before merging.

PR #1239, opened 2016-04-12
Bump Beaver version to avoid Mosquito dependency

hectcastro:
Something happened to a transitive dependency of Beaver, which caused installs to fail. This bumps to a version of Beaver that removes the broken dependency (Mosquito).

See also: https://github.com/python-beaver/python-beaver/pull/399

---

**Testing**

Provision servers with updated dependency and ensure that logs are still flowing to [Kibana](http://localhost:5601/).


mmcfarland:
+1, provisioned and ran site.  Logs are showing up as usual.

Issue #1227, opened 2016-04-07
Derive MapShed values from Vector / Lookup Datasets


rajadain:
Using the GMS File Key, and ingested vector / lookup datasets, derive the values needed for the MapShed calculation and add them to the Celery chain developed in #1216.

The following values are derived exclusively from vector / lookup datasets:

 * KV(m)
 * GrowSeason(m)
 * ErosCoef(m)
 * TotStrmLength
 * AEU
 * RunYrs
 * WxBegYr
 * WxEndYr
 * PcntET(i)
 * PtSrcNitro(m)
 * PtSrcPhos(m)
 * PtSrcFlow(m)
 * n41j
 * n41k
 * n41l
 * Average Temperature
 * Precipitation

In addition, the following values are derived from a combination of Raster and Vector / Lookup datasets:

 * LS(I)
 * SedPhos
 * AgLength
 * UrbLength
 * n42
 * n42b
 * n42c
 * n46e
 * n46f

PR #1241, opened 2016-04-13
Gather data from non-raster datasets

rajadain:
## Overview

Gathers data from non-raster datasets for populating a GWLF-E model.

 - [x] Weather Stations
 - [x] National Weather
 - [x] Streams
 - [x] Animal Types
 - [x] Point Source Discharge

## Testing Instructions

Check out the branch and open mmw. Create a new project with the MapShed model, and inspect the JSON result of the job. It should have a large blob filled with data from the defaults as well as the non-raster datasets.

## Code Review Instructions

We are attempting raw SQL queries for the first time in this project. Comment on the code style used there.

Using the GMS File Key spreadsheet in the MapShed folder on Google Drive, and the Class1.vb source file of the 1.3.0 version, do a quick spot check to ensure that the variables are calculated correctly.

## Output

Sample output from 
[Uploading MapShed.Output.json.txt…]()

## Notes

The raster fields will follow in an upcoming PR.

Connects #1227

### DIFF
--- a/src/mmw/apps/modeling/mapshed.py
+++ b/src/mmw/apps/modeling/mapshed.py
@@ -19,6 +19,7 @@ MONTHDAYS = settings.GWLFE_CONFIG['MonthDays']
 LIVESTOCK = settings.GWLFE_CONFIG['Livestock']
 POULTRY = settings.GWLFE_CONFIG['Poultry']
 LITERS_PER_MGAL = 3785412
+WEATHER_NULL = settings.GWLFE_CONFIG['WeatherNull']
 
 
 def day_lengths(geom):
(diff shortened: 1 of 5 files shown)


src/mmw/mmw/settings/gwlfe_settings.py
```diff
@@ -385,3 +385,11 @@
     'AttenLossRatePath': 0,  # Attenuation: Loss Rate: Pathogens
     'StreamFlowVolAdj': 1,  # Streamflow Volume Adjustment Factor
 }
+
+GWLFE_CONFIG = {
+    'NumWeatherStations': 2,  # Number of weather stations to consider for each polygon # NOQA
+    'Livestock': ['dairy_cows', 'beef_cows', 'hogs', 'sheep', 'horses'],
+    'Poultry': ['broilers', 'layers', 'turkeys'],
+    'MonthDays': [31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31],
```
lewfish: Do we need to worry about leap years?

rajadain: They don't in the original source code. This is only used to calculate the average discharge in liters per square meter for each month:

```python
discharge = np.array([float(mg_d * days * LITERS_PER_MGAL) / area
                      for days in MONTHDAYS])
```

Since it is the average, and not for a specific month in a specific year, we should be okay. I could potentially change it to `28.2425` (the average number of days in February), but that's not what they do in the original source code.

rajadain:
Added lookup fields.

src/mmw/mmw/settings/gwlfe_settings.py
```diff
@@ -16,6 +16,31 @@
...
+                'Open_Land',  # Maps NLCD 21, 71
+                'Bare_Rock',  # Maps NLCD 12, 31
+                'Sandy_Areas',     # Does not map to NLCD
+                'Unpaved_Road',    # Does not map to NLCD
+                'Ld_Mixed',   # Maps NLCD 22
+                'Md_Mixed',   # Maps NLCD 23
+                'Hd_Mixed',   # Maps NLCD 24
+                'Ld_Residential',  # Does not map to NLCD
+                'Md_Residential',  # Does not map to NLCD
+                'Hd_Residential',  # Does not map to NLCD
```
kdeloach: Land use, and several other fields, are referenced throughout the model code by these enum values https://github.com/WikiWatershed/gwlf-e/blob/develop/gwlfe/enums.py#L64. These are more like glorified strings rather than true enums, due to lack of language support. Still, I would recommend using these values where possible so we can write logic against these symbolic "enums" instead of having to sprinkle string matching logic throughout.

rajadain: Good point, will do.

rajadain:
Switched to using enums, removed flake8 checking from the settings file since it isn't tested against that anyway.

src/mmw/apps/modeling/tasks.py
```diff
@@ -34,8 +48,48 @@ def start_gwlfe_job(model_input):
     # TODO remove sleep and call actual GWLFE code
     time.sleep(3)
```
caseypt: Maybe drop this?

rajadain: Haha good idea

src/mmw/mmw/settings/gwlfe_settings.py
```diff
@@ -11,17 +16,42 @@
     'InitSnow': 0,  # Initial Snow Days
     'TileDrainRatio': 0,  # Tile Drain Ratio
     'TileDrainDensity': 0,  # Tile Drain Density
-    'ETFlag': 0,  # ET Flag: 0 for Hammon method, 1 for Blainy-Criddle method
-    'AntMoist': np.zeros(5),  # Antecedent Rain + Melt Moisture Conditions for Days 1 to 5 # NOQA
+    'ETFlag': et.HAMMON_METHOD,  # ET Flag: 0 for Hammon method, 1 for Blainy-Criddle method
```
caseypt: This is now `HAMON_METHOD`.

src/mmw/apps/modeling/tasks.py
```diff
@@ -34,8 +48,48 @@ def start_gwlfe_job(model_input):
...
 
+    aoi_geom = GEOSGeometry(json.dumps(model_input['area_of_interest']),
+                            srid=4326)
+    aoi_area = aoi_geom.transform(5070, clone=True).area  # Square Meters
+
     model = DataModel(settings.GWLFE_DEFAULTS)
```
caseypt: Consider renaming this `z`, which is what is used in the `gwlfe` code.

caseypt:
Code style looks good to me. It's very well formatted and readable.

src/mmw/apps/modeling/mapshed.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+    Given an array of weather stations and a growing season array, returns an
+    array of 12 decimals, one for the erosion coefficient of each month. For
+    months that are in the growing season, we average the `rain_warm` of both
+    the weather stations, and for months outside the growing season, we average
+    `rain_cool` instead.
+    """
+
+    avg_warm = sum([w.rain_warm for w in ws]) / len(ws)
+    avg_cool = sum([w.rain_cool for w in ws]) / len(ws)
```
lewfish: Consider using `mean` instead.

lewfish: I think you can use np.mean on plain arrays, but I'm not sure.

src/mmw/mmw/settings/gwlfe_settings.py
```diff
@@ -33,6 +33,14 @@
                 'Md_Residential',  # Does not map to NLCD
                 'Hd_Residential',  # Does not map to NLCD
                 ],
+    'Imper': np.array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0,          # Impervious surface area percentage # NOQA
```
lewfish: Where do these values come from in the original source?

rajadain: Row 47 of the GMS File Key points me to line 9426 of Class1.vb:

```vb
        ' Default urban landuse lines
        ' - Low Density Mixed
        LuLine(11) = "Ld_Mixed," & LuArea(11).ToString & ",0.15,92,74,60"
        NLoadRate(11) = "0.095,.015,0.33,0"
        PLoadRate(11) = "0.0095,.0021,0.4,0"
        SedLoadRate(11) = "2.8,0.8,0,0"
        ETGrow(11) = 0.85 * (LuArea(11) / BasinAreaHa)
        ETDorm(11) = 0.8 * (LuArea(11) / BasinAreaHa)
```

These are the static values in `LuLine`. Since they are only present in Urban Land uses, I defaulted the rest to 0.

lewfish:
This looks good to me. I'll give it another test when given the go ahead.

rajadain:
Up for one final look before merging pre-RRP

lewfish:
Looking now.

src/mmw/apps/modeling/mapshed.py
```diff
@@ -24,3 +34,70 @@ def day_lengths(geom):
...
+def growing_season(ws):
+    """
+    Given an array of weather stations, returns an array of 12 integers, each 1
+    or 0, indicating whether it is a growing season or not respectively.
+    We adopt a liberal approach, unioning the ranges to get a superset which is
+    a growing season for any weather station.
+    """
+
+    start = min([MONTHS.index(w.grw_start) for w in ws])
+    end = max([MONTHS.index(w.grw_end) for w in ws])
```
kdeloach: This may be a question for BME, but does it make sense to use the average here? Or are the weather stations close enough where it doesn't really matter?

rajadain: My initial approach was conservative, and I asked him about this, but he said to use the max range

> For any given polygon, we find the two nearest weather stations, and use the growing seasons associated with them. How do we calculate the final growing season? Is it the intersection (pick months available for both) or the union (pick months available in either) of their individual growing seasons?

> Use the earliest of the two for "Grw_start" and the latest of the two for "Grw_end".

rajadain: But I think in most cases the values will be the same because the weather stations are so close together

kdeloach:
+1 Looks really solid.

lewfish:
+1, tested. This was a monster PR. Good job!

Issue #1222, opened 2016-04-06
Handle new projects during ITSI workflow


lewfish:
PR #1219 added the ability to select the model package when creating new projects. According to @rajadain (see quoted text below), this behavior is incorrect when in ITSI mode.

For ITSI, passing "Model" should _not_ create a new project, but delete and replace the existing one. This may be complex and warrant a card of its own. I think the behavior should be:

 * If the user has created a project using Model A, goes back to Analyze, and then clicks on Model A again, they should return to existing project.
 * If the user has created a project using Model A, goes back to Analyze, and then clicks on Model B, they should be given a warning that this will kill their existing work, which is in Model A, and they should choose between preserving the project and proceeding with Model A, or killing it and proceeding with Model B.
 * If the user has created a project using Model A, goes back to Analyze, goes back to Draw, and changes their AoI, then they should be given a warning that their project will be killed, and if they agree, we kill the project and let them proceed with whichever model they like.

lewfish:
I think we agreed to the following: in both ITSI and non-ITSI mode, there will a dropdown to create a new model with a specific model package, and if there is an existing project, there will be a (new) button to go back to the modeling view for that project. Is there any other behavior we should implement?

lewfish:
It sounds like we want to do the above with the following additions/modifications: 
* When the user is in the modeling view and then goes back to the main map and tries to hit reset, we should prompt them about clearing the current project regardless of the mode and whether they are logged in (currently it only does this in ITSI mode)
* When the user clicks the new model dropdown, a new project will be created when in non-ITSI mode, and the current project will simply be cleared (scenarios deleted, etc) when in ITSI mode.

lewfish:
> there will be a (new) button to go back to the modeling view for that project

I think we should just implement this using the browser forward button. I also don't know why we have any back buttons in the app... why not just let people use the browser back button?

rajadain:
The back button is useful because when you come from Model to Analyze, the browser back may take you to Model, instead of Draw.

Perhaps instead of Back, it should say Draw.

lewfish:
I can change the button to say Draw. I've decided that the original scheme you came up with, in the first comment in this thread, is going to be easier to implement, so I'm going to do that.

PR #1242, opened 2016-04-18
Fix select model package in ITSI mode

lewfish:
This changes the behavior of selecting a model package when `itsi_embed=true` so that 1) the user will be prompted if they select a different model package after returning from the model view, and 2) will simply return to the same model if selecting the same model package. All other behavior should be the same as before. This follows the approach in the first comment in #1222. Implementing the third point did not require any changes.

screen shot 2016-04-18 at 11 19 30 am 2

##### Testing
* Log in as an ITSI user and append /?itsi_embed=true the URL
* Try creating a new model, then go back to Analyze, then click on the same model package. Note the project # is the same and the results are loaded instantly.
* Do the same except select a different model package. You should be prompted. If you click Cancel, nothing should happen. If you proceed, the same project number will be used but the package and scenarios will have changed.
* The behavior should be the same as before when logged out, and for a non-itsi logged in user.

#1222

### DIFF
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -4,8 +4,11 @@ var $ = require('jquery'),
     _ = require('lodash'),
     Marionette = require('../../shim/backbone.marionette'),
     App = require('../app'),
+    router = require('../router').router,
     models = require('./models'),
     settings = require('../core/settings'),
+    modalModels = require('../core/modals/models'),
+    modalViews = require('../core/modals/views'),
     coreModels = require('../core/models'),
     chart = require('../core/chart'),
     utils = require('../core/utils'),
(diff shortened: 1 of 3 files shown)


rajadain:
I'm seeing some odd behavior in the URL when in the Modeling stage. The Scenario ID doesn't seem to be updating as it should. Replicate this by:

 1. Create a new project in ITSI mode. Create a TR-55 model. Switch between the two default scenarios. Observe the Scenario ID updating correctly in the URL.
 2. Go back to Analyze, then back to the TR-55 Model. You should see the correct initial Scenario ID in the URL, but switching between them no longer updates it.
 3. Go back to Analyze, then create a MapShed Model. Agree to the prompt. Now, even though the scenarios are created and saved, you won't see any in the URL. Switching does nothing.
 4. Go back to Analyze, then back to the MapShed Model. You should see the correct initial Scenario ID in the URL, but switching between them no longer updates it. This is the same behavior as seen in step 2.
 5. Go back to Analyze, then create a TR-55 Model. Agree to the prompt. Now, even though the scenarios are created and saved, you won't see any in the URL. Switching does nothing. This is the same behavior as see in step 3.

Demo:

mmw-itsi-missing-scenario-id

lewfish:
Good catch. I pushed a commit that should fix this, and (with a little luck) doesn't break anything.

rajadain:
Taking another look.

src/mmw/js/src/analyze/views.js
```diff
@@ -30,6 +33,50 @@ var ResultsView = Marionette.LayoutView.extend({
...
+
+        var modelPackages = settings.get('model_packages'),
+            modelPackageName = $(e.target).data('id'),
+            modelPackage = _.find(modelPackages, {name: modelPackageName}),
+            newProjectUrl = '/project/new/' + modelPackageName,
+            projectUrl = '/project';
+
+        if (!modelPackage.disabled) {
+            if (settings.get('itsi_embed') && App.currentProject && !App.currentProject.get('needs_reset')) {
+                var currModelPackageName = App.currentProject && App.currentProject.get('model_package');
```
rajadain: Won't this set `currModelPackageName` to `true` if succesful, instead of the actual name?

lewfish: The `App.currentProject` is redundant, but it does work correctly as is. I'll change it.

rajadain:
+1 tested. Good job with the code, very minimal changes. I agree this is long overdue a refactor, but this is the best that could've been done at this time.

src/mmw/js/src/analyze/views.js
```diff
@@ -30,6 +33,50 @@ var ResultsView = Marionette.LayoutView.extend({
...
+                if (modelPackageName === currModelPackageName) {
+                    // Go to existing project
+                    router.navigate(projectUrl, {trigger: true});
+                } else {
+                    var confirmNewProject = new modalViews.ConfirmView({
+                        model: new modalModels.ConfirmModel({
+                            question: 'If you change the model package you will lose your current work.',
+                            confirmLabel: 'Switch Model Package',
```
kdeloach: Can we rename this to something else besides "model package"? How about just "model"? @ajrobbins

src/mmw/js/src/modeling/controllers.js
```diff
@@ -202,6 +171,50 @@ var ModelingController = {
...
+        'needs_reset': false
+    });
+
+    // Clear current scenarios and start over.
+    // Must convert to an array first to avoid conflicts with
+    // collection events that are disassociating the model during
+    // the loop.
+    var locks = [];
+    _.each(project.get('scenarios').toArray(), function(model) {
+        var $lock = $.Deferred();
```
kdeloach: I think `model.destroy` already returns a promise.

kdeloach:
+1

src/mmw/js/src/analyze/views.js
lewfish: We might have called it modelPackage to avoid confusion with the concept of a Backbone model. I can change it in the dialog though.

kdeloach: That's what I meant. I don't think people will know what "model package" means.

PR #1248, opened 2016-04-22
Remove remaining Ansible 2.0 deprecation warnings

hectcastro:
Stops use of "bare" variables along with `with_dict` and updates Beaver role that was doing something similar.

---

**Testing**

Use Vagrant to re-provision locally and ensure that the modified tasks no longer spit out purple colored deprecation warnings.


mmcfarland:
+1 - re-provisioned and got no deprecation warnings.

Issue #1245, opened 2016-04-21
MapShed Failure on Small AoIs


rajadain:
When selecting small Areas of Interest (such as a square kilometer), if certain datasets return no data (such as the point source discharge), then the MapShed endpoint crashes. Update the code to use defaults when no data is available from non-raster datasets.

PR #1249, opened 2016-04-22
Handle Empty Database Results for Small AoIs

rajadain:
## Overview

For small areas of interest, not all datasets would return values. For example, there may not be any streams, or any point sources of discharge, within a small polygon. Our code now handles these cases by following these rules:

 * If no weather stations are found, then raise an exception, since
   they are necessary for multiple calculations
 * For singleton values, return 0 if nothing is found
 * For arrays, return arrays of 0 if nothing is found

## Testing Instructions

Using the 1 sq km stamp tool, draw an AoI somewhere in central Philadelphia, as shown below:

image

Observe network traffic in either the developer tools, or using `./scripts/debugcelery.sh`. There should be no failures in processing the job.

```
[2016-04-22 14:22:44,736: INFO/MainProcess] Task apps.modeling.tasks.start_gwlfe_job[064aeb88-f895-456d-b6aa-cc9d7df604a9] succeeded in 0.619642704s: '{"NumPondSys": [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0], "RetentNLake": 0.12, "n41k": 391.0, "n38": 0,...
```

## Code Review Instructions

 * Comment on the style of defaulting chosen. Suggest better alternatives if possible.
 * Comment if there are any missing edge cases on query results

Connects #1245

### DIFF
--- a/src/mmw/apps/modeling/mapshed.py
+++ b/src/mmw/apps/modeling/mapshed.py
@@ -57,6 +57,9 @@ def nearest_weather_stations(geom, n=NUM_WEATHER_STATIONS):
     with connection.cursor() as cursor:
         cursor.execute(sql, [geom.wkt, n])
 
+        if cursor.rowcount == 0:
+            raise Exception("No weather stations found.")
+
         # Return all rows from cursor as namedtuple
         weather_station = namedtuple('WeatherStation',
                                      [col[0] for col in cursor.description])
@@ -146,8 +149,8 @@ def animal_energy_units(geom):
         values = cursor.fetchone()  # Only one row since aggregate query
         aeu = dict(zip(columns, values))
 
-        livestock_aeu = round(sum(aeu[animal] for animal in LIVESTOCK))
-        poultry_aeu = round(sum(aeu[animal] for animal in POULTRY))
+        livestock_aeu = round(sum(aeu[animal] or 0 for animal in LIVESTOCK))
+        poultry_aeu = round(sum(aeu[animal] or 0 for animal in POULTRY))
 
         return livestock_aeu, poultry_aeu
 
@@ -197,7 +200,7 @@ def stream_length(geom, drb=False):
     with connection.cursor() as cursor:
         cursor.execute(sql, [geom.wkt, geom.wkt])
 
-        return cursor.fetchone()[0]  # Aggregate query returns single value
+        return cursor.fetchone()[0] or 0  # Aggregate query returns singleton
 
 
 def point_source_discharge(geom, area):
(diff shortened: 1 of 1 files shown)


src/mmw/apps/modeling/mapshed.py
```diff
@@ -57,6 +57,9 @@ def nearest_weather_stations(geom, n=NUM_WEATHER_STATIONS):
     with connection.cursor() as cursor:
         cursor.execute(sql, [geom.wkt, n])
 
+        if cursor.rowcount == 0:
+            raise Exception("No weather stations found.")
+
```
lewfish: I can see how the closest weather station might be far away, but how can there be zero weather stations near a geom?

rajadain: I don't think there can ever be 0, which is why this is such an exceptional case. This would probably indicate a problem in the dataset itself

src/mmw/apps/modeling/mapshed.py
```diff
@@ -146,8 +149,8 @@ def animal_energy_units(geom):
         values = cursor.fetchone()  # Only one row since aggregate query
         aeu = dict(zip(columns, values))
 
-        livestock_aeu = round(sum(aeu[animal] for animal in LIVESTOCK))
-        poultry_aeu = round(sum(aeu[animal] for animal in POULTRY))
+        livestock_aeu = round(sum(aeu[animal] for animal in LIVESTOCK)) or 0
+        poultry_aeu = round(sum(aeu[animal] for animal in POULTRY)) or 0
```
lewfish: If aeu[animal] is None, won't this throw an exception?

src/mmw/apps/modeling/mapshed.py
```diff
@@ -197,7 +200,7 @@ def stream_length(geom, drb=False):
     with connection.cursor() as cursor:
         cursor.execute(sql, [geom.wkt, geom.wkt])
 
-        return cursor.fetchone()[0]  # Aggregate query returns single value
+        return cursor.fetchone()[0] or 0  # Aggregate query returns singleton
```
lewfish: Makes sense.

src/mmw/apps/modeling/mapshed.py
```diff
@@ -219,10 +222,10 @@ def point_source_discharge(geom, area):
         cursor.execute(sql, [geom.wkt])
         mg_d, kgn_month, kgp_month = cursor.fetchone()
 
-        n_load = np.array([kgn_month] * 12)
-        p_load = np.array([kgp_month] * 12)
+        n_load = np.array([kgn_month] * 12) if kgn_month else np.zeros(12)
+        p_load = np.array([kgp_month] * 12) if kgp_month else np.zeros(12)
         discharge = np.array([float(mg_d * days * LITERS_PER_MGAL) / area
-                              for days in MONTHDAYS])
+                              for days in MONTHDAYS]) if mg_d else np.zeros(12)
```
lewfish: Makes sense. Consider extracting np.zeros(12) to its own variable.

rajadain: Well, I want to use three different arrays, all initialized to 0, not pass around the reference to one array, even though they are all identical.

src/mmw/apps/modeling/mapshed.py
rajadain: Good catch (heh), I'll add one

rajadain: Actually, the constants `LIVESTOCK` and `POULTRY` are composed of the column names returned from this query. They should **always** match. If they don't, then we _should_ raise an exception.

However, if the value of one of those fields is `None`, that could interfere with the `sum`. I think moving the `0` inside, like this:

```python
livestock_aeu = round(sum(aeu[animal] or 0 for animal in LIVESTOCK))
```

might be better, so I will switch to that.

lewfish:
+1, tested

PR #1250, opened 2016-04-26
Ensure that Windshaft is restarted after log rotation

hectcastro:
If Windshaft isn't restarted after a log rotation event, it maintains a handle to the previous log file. This prevent new log messages from being written to `/var/log/mmw-tiler.log`, which also prevents them from being slurped up by Beaver.

Depends on https://github.com/azavea/ansible-beaver/pull/6

---

**Testing**

Provision the `tiler` and ensure that the `logrotate` configuration for `mmw-tiler` contains a directive for restarting the service after a log rotation event.


mmcfarland:
`postrotate

mmcfarland:
+1 `postrotate` directive is written to the config after provisioning.

Issue #1246, opened 2016-04-21
Plot GWLF-E Output


mmcfarland:
Mimic the TR55 output structure.  Two tabs, table and plot of values by month, selectable variables.

mmcfarland:
Suggested "Hydrology Output"
![screenshot from 2016-04-25 11 58 47](https://cloud.githubusercontent.com/assets/1014341/14789983/2e321f24-0add-11e6-9adb-8c9921083ab5.png)

Issue #1229, opened 2016-04-07
Using a static GMS input, run GWLF-E model and show results in UI


caseypt:
Follow up to #1216

caseypt:
Taking 1 point for https://github.com/WikiWatershed/gwlf-e/issues/23

PR #1251, opened 2016-04-26
Show GWLF-E runoff results

lewfish:
Run GWLF-E on backend using sample GMS file, and show the runoff results in the UI. 

Before testing, run `scripts/debugcelery.sh` and also run `git pull` in `src/mmw/requirements/src/gwlf-e` to get the latest commits to GWLF-E.

screen shot 2016-04-26 at 12 36 40 pm

Connects #1229 
Connects #1246


kdeloach:
Nice. This is really cool.

kdeloach:
I ran `pip install -r /opt/app/requirements/base.txt` to install the latest version of gwlfe.

kdeloach:
Consider updating requirements to install gwlf-e from PyPi instead of from source.

lewfish:
> Consider updating requirements to install gwlf-e from PyPi instead of from source.

I made a release of GWLF-E, but then realized it was better for now to just grab it from Github, so it's easy to get the latest updates.

kdeloach:
This looks great. We should update the unit of measurements in the next update.

src/mmw/js/src/modeling/gwlfe/runoff/views.js
```diff
@@ -0,0 +1,184 @@
...
+
+    templateHelpers: function() {
+        return {
+            runoffVars: runoffVars
+        };
+    },
+
+    modelEvents: {
+        'change:activeVar': 'render'
+    },
```
kdeloach: Consider moving `modelEvents` closer to the `events` hash.

kdeloach:
+1

kdeloach:
One strange thing I observed while testing this is that it looks like gwlf-e is being executed twice. We should investigate this at some point.

src/mmw/js/src/modeling/gwlfe/runoff/views.js
lewfish: Done.

lewfish:
> One strange thing I observed while testing this is that it looks like gwlf-e is being executed twice. We should investigate this at some point.

This is because there are two scenarios.

kdeloach:
Oh, I didn't think we created another scenario until the "New Scenario" tab is clicked. Working as designed then.

Issue #1252, opened 2016-04-26
Reusable SJS Routines


rajadain:
Currently, the SJS routines are tightly coupled to the data format of the land use census. Decouple them from data formatting so they may be used with any host, port, arguments, and JSON data.

PR #1253, opened 2016-04-26
Make SJS calls more generic

rajadain:
## Overview

Now that we need to support multiple Spark JobServer endpoints for multiple purposes, we decouple the data formatting routines from the SJS communication routines. We also include more detail in the exception messaging, which should make debugging easier.

## Testing Instructions

Check out the branch and observe Celery:

    ./scripts/debugcelery.sh

Open [http://localhost:8000/](http://localhost:8000/), select an area of interest. Observe Celery as the land use analysis is done. Create a "Site Storm Model" (TR-55) project. Observe Celery as the model is run.

All operations should complete successfully.

Connects #1252

### DIFF
--- a/src/mmw/apps/modeling/geoprocessing.py
+++ b/src/mmw/apps/modeling/geoprocessing.py
@@ -52,6 +52,70 @@ SOIL_MAPPING = {
 }
 
 
+@statsd.timer(__name__ + '.sjs_submit')
+def sjs_submit(host, port, args, data):
+    """
+    Submits a job to Spark Job Server. Returns its Job ID, which
+    can be used with sjs_retrieve to get the final result.
+    """
+    url = 'http://{}:{}/jobs?{}'.format(host, port, args)
+    response = requests.post(url, data=json.dumps(data))
+
+    if response.ok:
+        job = response.json()
+    else:
+        raise Exception('Unable to submit job to Spark JobServer.\n'
+                        'Details = {}'.format(response.text))
+
+    if job['status'] == 'STARTED':
+        return job['result']['jobId']
+    else:
+        raise Exception('Submitted job did not start in Spark JobServer.\n'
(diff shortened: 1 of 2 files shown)


src/mmw/apps/modeling/geoprocessing.py
```diff
@@ -52,6 +52,69 @@
...
+    if job['status'] == 'FINISHED':
+        return job['result']
+    elif job['status'] == 'RUNNING':
+        if retry:
+            try:
+                retry()
+            except MaxRetriesExceededError, X:
+                delete = requests.delete(url)  # Job took too long, terminate
+                if delete.ok:
+                    raise X
```
rajadain: I just copied this from the previous implementation, not sure if its the best thing to do here ...

caseypt: Maybe something like this would work:

```py
raise Exception('Job {} timed out, deleted.\n'
                'Details: {}'.format(job_id, delete.text))

rajadain: Found this: https://github.com/WikiWatershed/model-my-watershed/pull/826#discussion_r40717909

This makes sense to me. Will do.

kellyi:
We discussed this yesterday, and I've now got more context for what this is doing. I confirmed that the Celery operations complete successfully for the new MapShed data. The code for the changes works well and it feels pretty legible, too.

I haven't been able to get the streams data on my machine, so it might be worth having somebody else triple check the ops for that data, too. In principle that should work, too, though, as it's just a different data source.

rajadain:
The streams data (or lack thereof) shouldn't affect the work here, since this refactors existing functionality only that to be tested. Streams will be used in the MapShed endpoint which is still under construction.

kellyi:
Sounds good! Giving this a +1.

PR #1255, opened 2016-04-27
Restart gunicorn after exiting the debugserver

kdeloach:
This fixes the problem where you kill your debug server and then you
have to manually restart the gunicorn process on the app VM. Now, the
service should restart automatically, after exiting the debugserver.


rajadain:
+1 tested. This is a very welcome fix!

PR #1257, opened 2016-04-28
Add EPSG:5070 spatial reference to database

rajadain:
## Overview

For a number of MapShed operations, we need to convert data stored as EPSG:4326 (LatLng) to EPSG:5070 (ConusAlbers). To support operations such as ST_Transform(), we need to first add the spatial reference to the system.

## Testing Instructions

Run all migrations.

    ./scripts/manage.sh migrate

Regardless of whether or not it is already in your system, the migration should succeed. If 5070 is already installed, we leave the current installation. Else we add it.

## Notes

PostgreSQL 9.5 supports UPSERT operations using the much nicer syntax:

```sql
INSERT INTO <table>
VALUES <values>
ON CONFLICT DO NOTHING;
```

But since we don't have that, we use the humble `SELECT` instead.

The actual script is taken from http://epsg.io/5070.

### DIFF
--- a/src/mmw/apps/modeling/migrations/0018_postgis_add_EPSG5070.py
+++ b/src/mmw/apps/modeling/migrations/0018_postgis_add_EPSG5070.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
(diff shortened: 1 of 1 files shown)


kdeloach:
+1 Tested by merging into develop and #1251 and the SRID errors went away.

PR #1259, opened 2016-04-29
Meet Paramiko dependency before installing Ansible

hectcastro:
The loose dependency on Paramiko is pulling in a version with a backwards incompatible change, which is causing the Ansible install to fail. Until Ansible addresses this issue, I'd rather explicitly install an older version of Paramiko that satisfies Ansible dependency requirements.

See also: ansible/ansible#15665

---

**Testing**

See if http://civicci01.internal.azavea.com/view/mmw/job/model-my-watershed-packer-tiler/321/ succeeds.


mmcfarland:
+1

Issue #1258, opened 2016-04-28
MapShed Celery Workflow


rajadain:
Design and implement a Celery workflow for assembling the individual pieces of data collected via #1227 and https://github.com/WikiWatershed/mmw-geoprocessing/issues/27 into a GWLF-E model that is run, and returns JSON results.

PR #1265, opened 2016-05-02
MapShed Raster Geoprocessing

rajadain:
## Overview

This PR creates a framework for performing MapShed related geoprocessing calls. The overall structure is as follows:

```
                                                                                  +--+
                                                                                     |
+---------------------------------------------------------------------------------+  |
| start_gwlfe_job                                                                 |  |
|                                                                                 |  |
| +-----------------------------------------------------------------------------+ |  |
| | geop_tasks                                                                  | |  |
| |                                                                             | |  |
| | +----------------+ +----------------+ +----------------+ +----------------+ | |  |
| | | mapshed_start  | | mapshed_start  | | mapshed_start  | |                | | |  |
| | |     -----      | |     -----      | |     -----      | |                | | |  |
| | | mapshed_finish | | mapshed_finish | | mapshed_finish | |      ...       | | |  |
| | |     -----      | |     -----      | |     -----      | |                | | |  |
| | |  nlcd_streams  | | post_process2  | | post_process3  | |                | | |  |
| | +----------------+ +----------------+ +----------------+ +----------------+ | |  |
| |                                                                             | |  |
| +-----------------------------------------------------------------------------+ |  |     +---------------------+
|                                                                                 |  |     |                     |
| +-----------------------------------------------------------------------------+ |  +---> |    save_job_error   |
| |                                                                             | |  |     |                     |
| |                                   identity                                  | |  |     +---------------------+
| |                                                                             | |  |
| +-----------------------------------------------------------------------------+ |  |
|                                                                                 |  |
| +-----------------------------------------------------------------------------+ |  |
| |                                                                             | |  |
| |                                 collect_data                                | |  |
| |                                                                             | |  |
| +-----------------------------------------------------------------------------+ |  |
|                                                                                 |  |
+---------------------------------------------------------------------------------+  |
                                                                                     |
+---------------------------------------------------------------------------------+  |
|                                                                                 |  |
|                                  save_job_result                                |  |
|                                                                                 |  |
+---------------------------------------------------------------------------------+  |
                                                                                     |
                                                                                  +--+
```

There is a group of geoprocessing chains, each of which is a triplet containing `mapshed_start`, `mapshed_finish`, and a post-processing task, which would be different for each call. The results of these are passed to an `identity` task, which simply acts as a workaround to https://github.com/celery/celery/issues/3191, and then fed into `collect_data`, which combines the geoprocessing results with values from the database. `save_job_result` saves the output, and `save_job_errors` handles any errors throughout the chain, just as in previous implementations.

MapShed related code has been refactored into its own module, since there are going to be a large number of Celery tasks related to each geoprocessing value, and it would be difficult to keep track of them in the already large `modeling.tasks`. Thus, the MapShed tasks have been moved from `modeling.tasks` to `modeling.mapshed.tasks`, and the calculation functions from `modeling.mapshed` to `modeling.mapshed.calcs`.

The structure of geoprocessing settings has been changed to accommodate multiple SJS endpoints and payloads.

We add a `streams` method which returns the streams contained in the polygon as GeoJSON. These streams are sent to the geoprocessing service so that we can get a list of NLCD codes for each stream pixel. From this, we calculate the ones categorized as agricultural, and count those as `AgLength`. The rest are counted as `UrbLength`.

See individual commits for details.

## Testing Instructions

Build the geoprocessing service from this PR https://github.com/WikiWatershed/mmw-geoprocessing/pull/32 and place in your worker VM. Restart the `spark-jobserver` service so it picks up the new JAR.

```bash
export MMWG=path/to/mmw-geoprocessing
export MMW=path/to/model-my-watershed

cd $MMWG/
./sbt "project summary" assembly
mv summary/target/scala-2.10/mmw-geoprocessing-assembly-1.0.0.jar $MMW/mmw-geoprocessing-1.0.0.jar
cd $MMW/
vagrant ssh worker -c 'mv -f /vagrant/mmw-geoprocessing-1.0.0.jar /opt/geoprocessing/ && sudo restart spark-jobserver'
```

Debug Celery to open an observable window:

```bash
./scripts/debugcelery.sh
```

In MMW [http://localhost:8000/](http://localhost:8000/), create a MapShed project and observe Celery. Everything should run successfully.

Connects #1258 
ASCII Art created in http://asciiflow.com/


rajadain:
One bug I encountered with @kellyi is that chord unlocks can loop infinitely until the Redis cache is cleared with 

```bash
vagrant ssh services -c 'redis-cli flushall'
```
A potential guard can be to add `max_retries` and `interval` to the chord. See https://github.com/celery/celery/issues/1700#issuecomment-167681486

rajadain:
It would also be useful if the error description in Celery includes the error description Spark JobServer

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -0,0 +1,171 @@
...
+                        srid=4326)
+
+    return (group(geop_tasks(geom)) |
+            identity.s() |
+            collect_data.s(geom.geojson))
+
+
+@shared_task
+def nlcd_streams(result):
+    ag_streams = sum(result.get(nlcd, 0) for nlcd in [81, 82])
```
kellyi: Consider adding a docstring for future reference.

kellyi: Alternatively:

This might be better to do in the next step, but assuming each of the new functions you'll need to write for the next feature have similar structure, would it be worthwhile to make a sort of wrapper function that takes in `result`, some functions for manipulating it in whatever way, and then the key-val pair to return as an object and uses them all to generate the result? 

Then you could document how to use the wrapper fn rather than documenting how to mimic this one to generate other values.

kellyi:
Not sure if this is a bug or expected behavior, but I drew a huge aoi and the operation timed out (as it should) because Celery hit the max limit. However, I was then able to click Model for TR-55 and it kicked off another set of Celery tasks -- which also led to an error the first time I tried it. I tried it again with the same polygon and it led to that infinite loop issue.

kellyi:
Tried this again with this aoi
screen shot 2016-05-03 at 3 41 34 pm

and it looks like I was able to get Celery into that `chord_unlock` loop without having it crash or interrupting the process or anything. I drew the aoi, then immediately tried to model it so I assume everything might have been happening all at once. The initial drawing step ended up creating an error, but the model step got stuck in the loop.

kellyi:
Wonder if it's worth adding more guards on the frontend to sanitize the aoi inputs before they hit Celery.

kellyi:
Trying to model this

screen shot 2016-05-03 at 4 04 07 pm

screen shot 2016-05-03 at 4 05 21 pm


Led to this error:

    [2016-05-03 16:03:50,197: ERROR/MainProcess] Task apps.modeling.mapshed.tasks.nlcd_streams[134ed8b3-361c-4ed3-a656-0b963f905d0d] raised unexpected: ZeroDivisionError('integer division or modulo by zero',)

kellyi:
Tried logging in with a user account and running some models and most reasonably-small aois looked fine. It looks as if the loop thing may be the biggest issue here, along with the divide by zero error for small polygons.

rajadain:
Okay, this is ready for another look. @kellyi see if you can recreate the `chord_unlock` states.

Great testing, by the way! I think it does make sense to add a front-end guard for AoIs that are too small (less than 10 sq m? not sure), but there should probably be a new issue for that.

I've rebased this on latest develop, but only the last four "Celery Errors" commits are new. I would recommend testing with the `develop` of https://github.com/WikiWatershed/mmw-geoprocessing. This should cause an error like the following:

```
[2016-05-04 18:41:28,503: ERROR/Worker-1] Task e4a07644-8ada-400e-a560-fa6dcef1810d run from job 2483 raised exception: [nlcd_streams] Job 116d8a34-dc59-4a86-8ef1-33a8d28338e3 was ERROR (java.lang.Throwable: LineString geometry expected), deleted
```

This tests the error handling for when a geoprocessing task fails. To test the error handling in `collect_data`, change [the error handling](https://github.com/rajadain/model-my-watershed/blob/feature/mapshed-raster-values/src/mmw/apps/modeling/mapshed/tasks.py#L136) to return an empty dictionary instead.

```diff
diff --git a/src/mmw/apps/modeling/mapshed/tasks.py b/src/mmw/apps/modeling/mapshed/tasks.py
index 473f88e..cb30dc7 100644
--- a/src/mmw/apps/modeling/mapshed/tasks.py
+++ b/src/mmw/apps/modeling/mapshed/tasks.py
@@ -133,7 +133,7 @@ def collect_data(geop_result, geojson):
 @shared_task(throws=Exception)
 def nlcd_streams(result):
     if 'error' in result:
-        raise Exception('[nlcd_streams] ' + result['error'])
+        return {}
 
     ag_streams = sum(result.get(nlcd, 0) for nlcd in [81, 82])
     total = sum(result.values())
```

You should see an error like this:

```
[2016-05-04 18:40:23,536: ERROR/Worker-1] Task 4266e6f3-954f-4e49-8c89-8ac2569f5142 run from job 2481 raised exception: u'AgStreamPct'
```

In all cases, when the jobs fail, the front-end should stop polling immediately, and it's `error` field should be populated correctly.

```json
{
  "status": "failed",
  "started": "2016-05-04T22:41:25.827Z",
  "finished": "2016-05-04T22:41:27.510Z",
  "result": "",
  "error": "[nlcd_streams] Job d81f1336-a2d4-4d1f-b922-b2892153328c was ERROR (java.lang.Throwable: LineString geometry expected), deleted",
  "job_uuid": "f5db4b22-155f-45cb-97eb-9e01d68e05f9"
}
```

rajadain:
In addition to testing the failing case, also test with https://github.com/WikiWatershed/mmw-geoprocessing/pull/32 to ensure that the "success" case works properly.

kellyi:
This is beyond the scope of this PR but is it possible and/or worthwhile to increment the version number of the `mmw-geoprocessing....jar` file and then revise the worker VM just to run with the latest version of the the file it sees in `/opt/geoprocessing/`?

kellyi:
Drew this and tried to re-create the `chord_unlock` loop...

screen shot 2016-05-05 at 9 28 10 am

... and it succeeded in throwing an exception and breaking out of the loop.

kellyi:
This one also did not get stuck in the loop:

screen shot 2016-05-05 at 9 32 00 am

rajadain:
>  increment the version number of the `mmw-geoprocessing....jar` file and then revise the worker VM

Good point, created a 1 pointer #1269 for that.

kellyi:
Tested the `collect_data` error handling by commenting out `'AgStreamPct': ag_stream_percent` and it raised the correct exception then halted all the Celery tasks.

kellyi:
Checking again with the jar from mmw-geoprocessing develop with this aoi and I was able to generate an error reading `[2016-05-05 10:16:39,908: ERROR/Worker-1] Bad input data to TR55: u'runoff-vol'` in `debugcelery`, but the Celery tasks did halt.

screen shot 2016-05-05 at 10 18 54 am

kellyi:
OK! I checked this with the jar files from both `develop` and `rajadain:feature/less-specific-endpoints` of mmw-geoprocessing and it looks like it's handling both the failure and success cases properly now. 

Specifically: it looks like the Celery `chord_unlock` loop issue has been remedied, as I wasn't able to generate a loop when deliberately giving input which would force a timeout or generate some other kind of error. I also tried kicking off jobs by drawing a polygon then closing the page, drawing a polygon and then immediately selecting a model, and ctrl-c'ing out of `debugcelery` and then running it again to see what would happen when Redis gave it the prior jobs back. Everything seemed to stop eventually.

+1!

rajadain:
This was a hard one to finish. Thanks for the thorough testing!

kellyi:
No problem! Glad I was able to help.

Issue #1269, opened 2016-05-05
Bump MMW-Geoprocessing Version


rajadain:
There's a new version upcoming which will include the `MapshedJob` methods `rasterLinesJoin`, `rasterLinesJoinSequential`, and `rasterJoin`. Make a release of that version and bump the version in MMW to use that.

PR #1271, opened 2016-05-05
Upgrade Geoprocessing to 1.1.0

rajadain:
## Overview

This adds support for MapshedJob which will be used for calculating various values needed for running GWLF-E.

## Testing Instructions

Checkout the branch, then reprovision the worker:

```
vagrant provision worker
```

Then open [http://localhost:8000/](http://localhost:8000/) and make a MapShed project. You should successfully see an output.

image

Connects #1269


kellyi:
+1. Tested.

Issue #1281, opened 2016-05-05
MapShed: Divide By Zero when calculating AgLength


rajadain:
If there are no streams in an AoI, then we get a divide by zero error here: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/mapshed/tasks.py#L179.

Update the code so that it returns 0 in those cases.

rajadain:
Shares a point with #1272

Issue #1272, opened 2016-05-05
MapShed: Calculate NLCD + Streams Values


rajadain:
Using `RasterLinesJoin` and the example of `nlcd_streams`, calculate the values of

 - [x] n42: Streams in Agricultural Areas (km)
 - [x] n42b: Total Stream Length (km)
 - [x] n42c: Unpaved Road Length (km)
 - [x] n46e: Stream Km in High Density Urban Areas
 - [x] n46f: Stream Km in Low Density Urban Areas

PR #1283, opened 2016-05-06
Add NLCD + Stream values

rajadain:
## Overview

We ignore Unpaved Roads, thus n42c is initialized as 0 with all other static values. The rest are calculated from the NLCD + Streams geoprocessing call.

## Testing Instructions

Check out code and add `print(z.tojson())` [here](https://github.com/rajadain/model-my-watershed/blob/feature/mapshed-nlcd-streams/src/mmw/apps/modeling/mapshed/tasks.py#L134). Choose a polygon in an urban area, and ensure that `n46e` has a non-zero value.

Also ensure that selecting a very small AoI does not result in a Divide By Zero error when there are no streams in the AoI.

Connects #1272 
Connects #1281

### DIFF
--- a/src/mmw/apps/modeling/mapshed/tasks.py
+++ b/src/mmw/apps/modeling/mapshed/tasks.py
@@ -112,8 +112,8 @@ def collect_data(geop_result, geojson):
     z.ManNitr, z.ManPhos = manure_spread(z.AEU)
 
     # Data from Streams dataset
-    z.StreamLength = stream_length(geom)   # Meters
-    z.n42b = round(z.StreamLength / 1000)  # Kilometers
+    z.StreamLength = stream_length(geom)      # Meters
+    z.n42b = round(z.StreamLength / 1000, 1)  # Kilometers
 
     # Data from Point Source Discharge dataset
     n_load, p_load, discharge = point_source_discharge(geom, area)
@@ -126,8 +126,12 @@ def collect_data(geop_result, geojson):
     z.Temp = temps
     z.Prec = prcps
 
-    z.AgLength = geop_result['AgStreamPct'] * z.StreamLength
+    # Begin processing geop_result
+    z.AgLength = geop_result['ag_stream_pct'] * z.StreamLength
     z.UrbLength = z.StreamLength - z.AgLength
+    z.n42 = round(z.AgLength / 1000, 1)
+    z.n46e = geop_result['med_high_urban_stream_pct'] * z.StreamLength / 1000
+    z.n46f = geop_result['low_urban_stream_pct'] * z.StreamLength / 1000
 
     z.CN = np.array(geop_result['cn'])
     z.SedPhos = geop_result['sed_phos']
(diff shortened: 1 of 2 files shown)


kellyi:
Selecting this polygon:

screen shot 2016-05-09 at 2 48 07 pm

Leads `z.n46e` to have a value of 0. No divide by zero error, though.

kellyi:
This polygon's 0 for `n46e` also:
screen shot 2016-05-09 at 2 50 25 pm

kellyi:
This one's `437.405405405`:

screen shot 2016-05-09 at 2 51 56 pm

Did not encounter any divide by zero errors, and `print(z.tojson())` seems to give the correct output.

Assuming the 0 values for `n46e` in some small polygons in urban areas are correct: +1.

rajadain:
`n46e` is "Stream Km in High Density Urban Areas". Since we default to using the NHD dataset, and there were no NHD streams in the areas in your screenshots, those values were correctly 0.

You can turn on streams in the UI from the Overlays menu:

image

rajadain:
However, we probably shouldn't have a value of `437.405405405` km in the screenshot you posted! I think I may have forgotten to convert from m to km somewhere

kellyi:
Ah. -1 then!

rajadain:
Fixed in b69f0ce. Mind checking again?

kellyi:
Last polygon now shows `z.n46e` value of `0.378378378378`

kellyi:
`z.n46e => 1.31514408338` for this:

screen shot 2016-05-09 at 3 26 38 pm

rajadain:
Can you try the one that was 437 m but 0.378 km with this line change?

```diff
diff --git a/src/mmw/apps/modeling/mapshed/tasks.py b/src/mmw/apps/modeling/mapshed/tasks.py
index 06c6fb7..57d823e 100644
--- a/src/mmw/apps/modeling/mapshed/tasks.py
+++ b/src/mmw/apps/modeling/mapshed/tasks.py
@@ -125,7 +125,7 @@ def collect_data(geop_result, geojson):
     z.UrbLength = z.StreamLength - z.AgLength
 
     z.n42 = z.AgLength / 1000
-    z.n46e = geop_result['MedHighUrbanStreamPct'] * z.n42b
+    z.n46e = round(geop_result['MedHighUrbanStreamPct'] * z.StreamLength / 1000)
     z.n46f = geop_result['LowUrbanStreamPct'] * z.n42b
 
     # TODO pass real input to model instead of reading it from gms file
```

kellyi:
For `z.n46e = round(geop_result['MedHighUrbanStreamPct'] * z.StreamLength / 1000)` the result looks like:

`z.n46e => 0.0`

rajadain:
Really? This is the same one as this?

> Last polygon now shows z.n46e value of 0.378378378378

kellyi:
Yeah, with the original line it's `z.n46e => 0.378378378378`

kellyi:
Note: adjusting the function to `z.n46e = geop_result['MedHighUrbanStreamPct'] * float(z.StreamLength) / 1000` outputs a value of `z.n46e => 0.437405405405`

rajadain:
Ah right, good catch with the rounding. Will update to use that instead.

rajadain:
Merging shortly ...

Issue #1284, opened 2016-05-06
Celery fails when querying multiple rasters


rajadain:
```
[2016-05-06 18:20:30,458: ERROR/MainProcess] Task apps.modeling.mapshed.tasks.mapshed_finish[0d90d255-b883-4cbc-b9ce-3a1862406354] raised unexpected: EncodeError(TypeError('keys must be a string',),)
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 280, in trace_task
    signature(callbacks[0], app=app).delay(retval)
  File "/usr/local/lib/python2.7/dist-packages/celery/canvas.py", line 175, in delay
    return self.apply_async(partial_args, partial_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/canvas.py", line 251, in apply_async
    return _apply(args, kwargs, **options)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/task.py", line 559, in apply_async
    **dict(self._get_exec_options(), **options)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/base.py", line 353, in send_task
    reply_to=reply_to or self.oid, **options
  File "/usr/local/lib/python2.7/dist-packages/celery/app/amqp.py", line 305, in publish_task
    **kwargs
  File "/usr/local/lib/python2.7/dist-packages/kombu/messaging.py", line 165, in publish
    compression, headers)
  File "/usr/local/lib/python2.7/dist-packages/kombu/messaging.py", line 241, in _prepare
    body) = dumps(body, serializer=serializer)
  File "/usr/local/lib/python2.7/dist-packages/kombu/serialization.py", line 164, in dumps
    payload = encoder(data)
  File "/usr/lib/python2.7/contextlib.py", line 35, in __exit__
    self.gen.throw(type, value, traceback)
  File "/usr/local/lib/python2.7/dist-packages/kombu/serialization.py", line 59, in _reraise_errors
    reraise(wrapper, wrapper(exc), sys.exc_info()[2])
  File "/usr/local/lib/python2.7/dist-packages/kombu/serialization.py", line 55, in _reraise_errors
    yield
  File "/usr/local/lib/python2.7/dist-packages/kombu/serialization.py", line 164, in dumps
    payload = encoder(data)
  File "/usr/local/lib/python2.7/dist-packages/anyjson/__init__.py", line 141, in dumps
    return implementation.dumps(value)
[...]

PR #1285, opened 2016-05-06
Parse SJS Results in post-processing step

rajadain:
## Overview

Since the SJS results may contain tuples as keys which cannot be transferred between Celery tasks as they are not JSON serializable we switch to parsing them in the post-processing step instead.

## Testing Instructions

Try this with some code that queries more than one raster.

Connects #1284


lewfish:
+1, tested with tuples with > 1 value

Issue #1276, opened 2016-05-05
MapShed: Calculate NLCD + Other Values


rajadain:
These values depend on NLCD and other datasets that are NOT Streams or Slope.

 - [x] SedPhos (NLCD + Soil Group + Soil Phosphorus Concentration)
 - ~~AvgK (K-Factor)~~
 - [x] CN(l) (NLCD + Soil Group)
 - ~~KF(l) (NLCD + K-Factor)~~

rajadain:
For K-Factor use the Raster ID `us-ssugro-kfactor-30m-epsg5070`

Issue #1273, opened 2016-05-05
MapShed: Calculate NLCD Values


rajadain:
These values can be calculated from just the NLCD histogram, and thus can potentially be combined with another simple request.

 - [x] KV(m) (NLCD + Growing Season)
 - [x] Landuse(l)
 - [x] LuArea(l)
 - [x] NormalSys
 - [x] n23: Row Crops Area (Ha)
 - [x] n23b: High Density Urban Area (Ha)
 - [x] n24: Hay/Pasture Area (Ha)
 - [x] n24b: Low Density Urban Area (Ha)

PR #1286, opened 2016-05-09
Calculate Curve Number and Sediment Phosphorus

rajadain:
## Overview

**Curve Number** is defined in a table, mapping each MapShed Non-Urban Land Use Type and a limited subset of Hydrological Soil Group to a to value. MapShed Non-Urban Land Use Types map to NLCD as follows:

| MapShed Name  | NLCD Code     | NLCD Name                                                       |
|---------------|---------------|-----------------------------------------------------------------|
|               |               |                                                                 |
| Hay/Pasture   | 81            | Pasture/Hay                                                     |
| Cropland      | 82,---        | Cultivated Crops,---                                            |
| Forest        | 41,42,(43,52) | Deciduous Forest, Evergreen Forest, (Mixed Forest, Shrub/Scrub) |
| Wetland       | 90,95         | Woody Wetlands, Emergent Herbaceous Wetlands                    |
| Disturbed     | ---           | ---                                                             |
| Turf Grass    | ---           | ---                                                             |
| Open Land     | 21,71         | Developed Open Space, Grassland/Herbaceous                      |
| Bare Rock     | 12,31         | Perennial Ice/Snow, Barren Land                                 |
| Sandy Areas   | ---           | ---                                                             |
| Unpaved Roads | ---           | ---                                                             |

The Soil Group subset we use ranges from `A` to `D`. Other values are reduced to this subset in the `SOIL_GROUP` table. Once we have this pairing, we use the `CURVE_NUMBER` table to look up the final value for each land use type.

We restrict this calculating to Non-Urban Land Use Types only because urban land use types [have default values](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/gwlfe_settings.py#L43-L46).

**Sediment Phosphorus** is the concentration of phosphorus in the top soil, and has values corresponding to different soil textures and land use types stored in the `SOILP` table. For each cell, we check if the land use of that cell is agricultural or not, and use the corresponding value. The final value is the average of each cell.

I've elected to use one geoprocessing call and unpair the results in Python because I expect multiple geoprocessing calls would have been costlier. This does unfortunately complicate the Python code.

Connects #1276

### DIFF
--- a/src/mmw/apps/modeling/mapshed/tasks.py
+++ b/src/mmw/apps/modeling/mapshed/tasks.py
@@ -3,6 +3,8 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import absolute_import
 
+import numpy as np
+
 from os.path import join, dirname, abspath
 from ast import literal_eval as make_tuple
 
@@ -124,6 +126,9 @@ def collect_data(geop_result, geojson):
     z.AgLength = geop_result['AgStreamPct'] * z.StreamLength
     z.UrbLength = z.StreamLength - z.AgLength
 
+    z.CN = np.array(geop_result['CN'])
+    z.SedPhos = geop_result['SedPhos']
+
     # TODO pass real input to model instead of reading it from gms file
     gms_filename = join(dirname(abspath(__file__)), 'data/sample_input.gms')
     gms_file = open(gms_filename, 'r')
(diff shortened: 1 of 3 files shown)


ajrobbins:
I believe some of the MapShed land use types ---> NLCD codes may be off. I consulted with Barry on this late last week and have been maintaining this table https://docs.google.com/spreadsheets/d/1XTJVQI2sSVKnS5U5hsJOq0eznaCULKPgo3Uw4fUzzQQ/edit#gid=0

- NLCD code 52 (shrub/scrub) is not Forest
- Sandy Areas = NLCD Barren Land

rajadain:
Sorry, the table above doesn't include all the values. This is the table we're building off of: https://docs.google.com/spreadsheets/d/12hsklRHwl6sNum0xPk6d5Cw6ONDA3X6BR5-Pc-NwJBc/edit#gid=472348661 and it seems to correspond with yours.

Sandy Areas in yours has MapShed Value of 22, which we do map to NLCD 12 and 31 (Perennial Snow and Barren Land). MapShed Value of 9, which represents both "Wooded Areas" (line 12) and "" (line 14), maps to NLCD 43 in line 12 and NLCD 52 in line 14.

ajrobbins:
You're right, the mappings (and tables) are the same. My mistake!

rajadain:
Actually, hold off on reviewing this. I'm going to see if this work can be combined with something for #1273.

PR #1299, opened 2016-05-11
Calculate NLCD-only dependent values

rajadain:
## Overview

Adds values that only depend on NLCD. Piggy-backs on the `nlcd_soils` geoprocessing call since that is giving us NLCD counts already.

## Testing Instructions

Compare implementation with original VB code referenced in original card and in the Python code.

Connects #1273


jtarricone:
+1 this looks consistent with the VB implementation as far as I can tell

rajadain:
Thanks for the review!

PR #1300, opened 2016-05-11
Upgrade mmw-geoprocessing to 1.2.0

rajadain:
We now have the new `RasterGroupedSum` and `RasterGroupedAverage` endpoints. Calling `RasterGroupedAverage` with an empty `rasters` set will perform an ungrouped average. `RasterJoin` has been renamed to `RasterGroupedCount` for consistency.

Reprovision the worker and test with `./scripts/debugcelery.sh`


lewfish:
+1, tested. You just need to increase the max retries for the chord unlock to something higher, like 10.

Issue #1278, opened 2016-05-05
Add BMPs for GWLF-E


lewfish:
Add UI for adding BMPs for GWLF-E. The existing BMP menus should be modified from TR55 so that clicking a BMP leads to another dialog that allows the user to enter the size. The Draw option should be disabled.

<img width="335" alt="screen shot 2016-05-05 at 1 39 11 pm" src="https://cloud.githubusercontent.com/assets/1896461/15051617/8da8e28e-12c7-11e6-911f-3276b939e4f4.png">

<img width="344" alt="screen shot 2016-05-05 at 1 39 17 pm" src="https://cloud.githubusercontent.com/assets/1896461/15051615/8cae0940-12c7-11e6-9973-e3cb7c4634e1.png">

ajrobbins:
New BMP help text is here: https://docs.google.com/document/d/1aGer7yAqX1HjmVqiFY4yaWfOvJrJi0ckzefOArO7JqA/edit

PR #1306, opened 2016-05-13
Add GWLFE modification dropdown

lewfish:
Add a modification dropdown for GWLFE. This involved a lot of refactoring. 

The UI for entering manually entering BMP details is very rudimentary, and will be completed in another PR for #1307

#### Testing
* For a TR55 model, check that modifications work as before.
* For a GWLFE model, check that there is no modification dropdown for current conditions, and there is one for other scenarios. 
* The dropdown should have a disabled radio button group, which is just for show.
* Clicking the thumbnails should replace the content with a manual entry dialog. 
* In the manual entry dialog, the only thing that works is clicking the back arrow, and clicking convert closes the dialog. 
* Clicking outside the dropdown, or on the dropdown button should close the dropdown.

screen shot 2016-05-13 at 1 04 36 pm

screen shot 2016-05-13 at 1 04 41 pm

Connects #1278


rajadain:
I'm not sure about the readability of the header text when using a modification fill as the background.

lewfish:
> I'm not sure about the readability of the header text when using a modification fill as the background.

I know. We'd have to change the text color as a function of the background to make it look nice. Or put a black outline around white text might work.

rajadain:
> black outline around white text

That's a good idea

ajrobbins:
I'm only working from the screenshots here, but just noting a couple of things:
- There should be 10 BMPs - Animal Waste Management Systems (Livestock), and Animal Waste Management Systems (Poultry). They can have the same description, but they are separate inputs. Also, we should try to preserve more of the input name than "Waste Management", if possible.
- Can the "Draw" and "Manual Entry" text be aligned with the radio buttons? They appear to be slightly shifted up, but maybe that's just the screenshot.
- Can the "Draw" text be greyed out as well as the radio button, to make it super clear that the Draw functionality is not available?
- Are the units of the applied BMP shown anywhere when someone enters a value in the text box? That would probably be good to see.
@mmcfarland Should we be exposing the land cover type that the BMP is applied to? Like "10km of stream length", or something?

mmcfarland:
I'm assuming those issues are meant to be addressed in https://github.com/WikiWatershed/model-my-watershed/issues/1307?

ajrobbins:
Good point, @lewfish disregard my last bullet please!

ajrobbins:
Actually a total of 13 BMPs https://docs.google.com/document/d/1tCRD13ThcJdfY2qJ3mcDK697lJj66ED0p2IfgA7B9hg/edit
The following should have one each for "Urban" and "Rural"
- Vegetated Buffer Strips
- Streambank Stabilization

lewfish:
Draw and Manual entry radio buttons have been removed. The other 3 BMPs and style issues such as button style and text color over fill pattern will be addressed in another PR.

src/mmw/js/src/modeling/controls.js
```diff
@@ -35,79 +34,203 @@ var ControlView = Marionette.LayoutView.extend({
...
+        ControlView.prototype.initialize.apply(this, [options]);
+        var self = this;
+
+        // If clicked outside this view and dropdown is open, then close it.
+        $(document).mouseup(function(e) {
+            var isTargetOutsideDropdown = $(e.target).parents('.dropdown-menu').length === 0;
+
+            if (isTargetOutsideDropdown && self.model.get('dropdownOpen')) {
+                self.model.set('dropdownOpen', false);
+            }
```
caseypt: Consider pulling this out into a named function.

caseypt:
+1. Tested. Nice job refactoring things.

src/mmw/js/src/modeling/controls.js
lewfish: I'll do this in the next PR.

lewfish:
Thanks for the review.

hectcastro:
@azavea-bot rebuild

PR #1289, opened 2016-05-09
Compute values based on nlcd, slope, and kfactor rasters

lewfish:
Compute AgSlope3, AgSlope3To8, n41, AvSlope, AvKF, and KF from the nlcd, slope, and kfactor raster datasets.

I am pulling out SedAFactor into its own issue because it depends on CN values which are computed in #1288.

To test, verify implementation against Class1.vb and look at output in `./scripts/debugcelery.sh` which prints the results to console. I found it strange that all kfactor values are 0.0.

Connects #1274

PR #1308, opened 2016-05-13
Remove accidentally committed changes

rajadain:
## Overview

Visual Studio Code merged #1289 in unexpected fashion. This clears up the accidental changes that were committed.

## Testing Instructions

Ensure there are no print statements. Test with `./scripts/debugcelery.sh` that there are no errors.


kellyi:
Had to reprovision the worker VM but I tested this and I did not see any print statements. I did see one "max retries" error once in Celery, but it only showed up one out of three times.

So: unless that error's not something you anticipated possibly happening: +1.

rajadain:
Great. I think we'll up [`CELERY_CHORD_UNLOCK_MAX_RETRIES`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/base.py#L129) as we go along and our computation becomes more and more complex.

Issue #1277, opened 2016-05-05
MapShed: Calculate LS


rajadain:
LS can be calculated using NLCD, Slope, and Stream Density. A finer method may involve using Flow Accumulation, but unfortunately we don't have a Flow Accumulation raster at 30m resolution.

Use DRB streams if the polygon is wholly contained within the DRB, else use NHD streams.

See if this geoprocessing can be combined with any in #1272 or #1274 to improve performance.

rajadain:
To calculate Average Slope over an entire area, use:

```json
{
  "input": {
    "polygon": [],
    "targetRaster": "us-percent-slope-30m-epsg5070",
    "rasterCRS": "ConusAlbers",
    "polygonCRS": "LatLng",
    "operationType": "RasterAverage",
    "zoom": 0
  }
}
```

jtarricone:
Won't the average slope be calculated as part of #1289 ?

rajadain:
You're right, it should.

jtarricone:
Coolio, I'm at a decent stopping point until that gets merged. The LS factor calculation should be in place, so any additional work after #1289  should be minimal, barring math errors of course.

rajadain:
This should now be unblocked

rajadain:
It'd be great to get a PR for this up soon so I can rebase #1305 over it

PR #1313, opened 2016-05-17
Add slope LS-factor that is calculated from rasters.

jtarricone:
Calculate LS factors for each land use type and add them to the model input.

The primary test for correctness is to compare this implementation to the original VB implementation. The more general test of the implementation is to fold the new files into an existing VM and run ./scripts/debugcelery.sh to see if any errors are encountered; a further step would be to add a print() call on the calculation results to the prior exercise, which would print the computed LS-factors (however, it's not immediately obvious to what to compare those values for validation).

Connects #1277

### DIFF
--- a/src/mmw/apps/modeling/mapshed/calcs.py
+++ b/src/mmw/apps/modeling/mapshed/calcs.py
@@ -24,6 +24,7 @@ POULTRY = settings.GWLFE_CONFIG['Poultry']
 LITERS_PER_MGAL = 3785412
 WEATHER_NULL = settings.GWLFE_CONFIG['WeatherNull']
 AG_NLCD_CODES = settings.GWLFE_CONFIG['AgriculturalNLCDCodes']
+KM_PER_M = 0.001
 
 
 def day_lengths(geom):
@@ -202,6 +203,47 @@ def manure_spread(aeu):
         p_list[lu] = p_spread
 
     return n_list, p_list
+
+
+def ls_factors(lu_strms, total_strm_len, areas, avg_slope):
+    results = [0.0] * len(lu_strms)
+    if 0 <= avg_slope <= 1.0:
+        m = 0.2
+    elif 1.0 < avg_slope <= 3.5:
+        m = 0.3
+    elif 3.5 < avg_slope <= 4.5:
+        m = 0.4
+    else:
+        m = 0.5
+
+    for i in range(len(lu_strms)):
+        results[i] = (ls_factor(lu_strms[i] * total_strm_len * KM_PER_M,
+                      areas[i], avg_slope, m))
+
+    return results
+
+
+def ls_factor(stream_length, area, avg_slope, m):
+    #  units --> stream_length:km, area:hectare, avg_slope:%, m:(n/a)
+    if area == 0.0:
+        return 0.0
+
+    slope_cell_size = 30
+
+    #  ensure a floor on the stream length of 250m
+    if stream_length < (250.0 * KM_PER_M):
+        slope_length = (0.5 * area * 10000) / 250.0
+    else:
+        slope_length = (0.5 * area) / stream_length
+
+    ls = (((slope_length / 22.13) ** m) *
+          (0.065 + (0.043 * avg_slope) + (0.0065 * (avg_slope ** 2))))
+
+    #  keep this for posterity
+    if slope_cell_size < 50:
+        return (0.0793 * ls) + 0.1913
+    else:
+        return ls
 
 
 def stream_length(geom, drb=False):
--- a/src/mmw/apps/modeling/mapshed/tasks.py
+++ b/src/mmw/apps/modeling/mapshed/tasks.py
@@ -25,6 +25,7 @@ from apps.modeling.mapshed.calcs import (day_lengths,
                                          et_adjustment,
                                          kv_coefficient,
                                          animal_energy_units,
+                                         ls_factors,
                                          manure_spread,
                                          streams,
                                          stream_length,
@@ -156,6 +157,9 @@ def collect_data(geop_result, geojson):
 
     z.AvSlope = geop_result['avg_slope']
 
+    z.LS = ls_factors(geop_result['lu_stream_pct'], z.StreamLength,
+                      z.Area, z.AvSlope)
+
     z.AvKF = geop_result['avg_kf']
     z.KF = geop_result['kf']
 
@@ -181,7 +185,6 @@ def collect_data(geop_result, geojson):
 
     # The frontend expects an object with runoff and quality as keys.
     response_json = {'runoff': gwlfe.run(z)}
-
     return response_json
 
 
@@ -237,11 +240,17 @@ def nlcd_streams(sjs_result):
                          for count in (ag_count,
                                        low_urban_count,
                                        med_high_urban_count))
+    lu_stream_pct = [0.0] * 16
+    for nlcd, stream_count in result.iteritems():
+        lu = get_lu_index(nlcd)
+        if lu is not None:
+            lu_stream_pct[lu] += float(stream_count) / total
 
     return {
         'ag_stream_pct': ag,
         'low_urban_stream_pct': low,
-        'med_high_urban_stream_pct': med_high
+        'med_high_urban_stream_pct': med_high,
+        'lu_stream_pct': lu_stream_pct
     }
 
 


rajadain:
Run `check.sh` to see linting errors:

```
$ ./scripts/check.sh 
+ vagrant ssh app -c 'flake8 /opt/app/apps --exclude migrations || echo flake8 check failed'
/opt/app/apps/modeling/mapshed/calcs.py:208:1: E302 expected 2 blank lines, found 1
/opt/app/apps/modeling/mapshed/calcs.py:214:1: E302 expected 2 blank lines, found 1
/opt/app/apps/modeling/mapshed/calcs.py:216:25: E261 at least two spaces before inline comment
/opt/app/apps/modeling/mapshed/calcs.py:216:26: E262 inline comment should start with '# '
/opt/app/apps/modeling/mapshed/calcs.py:234:80: E501 line too long (81 > 79 characters)
/opt/app/apps/modeling/mapshed/calcs.py:237:1: W293 blank line contains whitespace
/opt/app/apps/modeling/mapshed/calcs.py:238:80: E501 line too long (110 > 79 characters)
/opt/app/apps/modeling/mapshed/calcs.py:241:1: E302 expected 2 blank lines, found 1
/opt/app/apps/modeling/mapshed/tasks.py:50:1: E302 expected 2 blank lines, found 1
/opt/app/apps/modeling/mapshed/tasks.py:180:5: E303 too many blank lines (2)
/opt/app/apps/modeling/mapshed/tasks.py:189:1: E302 expected 2 blank lines, found 1
flake8 check failed
```

src/mmw/apps/modeling/mapshed/calcs.py
```diff
@@ -203,6 +205,38 @@ def manure_spread(aeu):
...
+    results = [0.0] * len(lu_strms)
+    for i in range(len(lu_strms)):
+        results[i] = ls_factor(lu_strms[i] * KM_PER_M, areas[i], avg_slope)
+    return results
+
+def ls_factor(stream_length, area, avg_slope):
+    # units --> stream_length:km, area:hectare, avg_slope:%
+    slope_cell_size = 30 #for posterity
+    if 0 <= avg_slope <= 1.0:
+        m = 0.2
```
rajadain: Since the value of `m` depends on `avg_slope` and nothing else, consider putting this in `ls_factors` so that it is calculated only once, and not for each land use type in the loop

src/mmw/apps/modeling/mapshed/calcs.py
```diff
@@ -203,6 +205,38 @@ def manure_spread(aeu):
...
+    elif 1.0 < avg_slope <= 3.5:
+        m = 0.3
+    elif 3.5 < avg_slope <= 4.5:
+        m = 0.4
+    else:
+        m = 0.5
+
+    if slope_cell_size < 50:
+        s = 0.0793
```
rajadain: Same for `s` and `t`

src/mmw/apps/modeling/mapshed/calcs.py
```diff
@@ -203,6 +205,38 @@ def manure_spread(aeu):
...
+        t = 0.1913
+    else:
+        s = 1
+        t = 0
+
+    if stream_length < (250.0 * KM_PER_M):
```
rajadain: Are we checking if `stream_length` is less than 250 km or 250 m?

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -237,11 +238,17 @@ def nlcd_streams(sjs_result):
                          for count in (ag_count,
                                        low_urban_count,
                                        med_high_urban_count))
+    lu_streams = [0.0] * 16
+    for k, v in result.iteritems():
```
rajadain: Consider using `nlcd, streams` instead of `k, v` to clearly describe the contents of result.

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -237,11 +238,17 @@ def nlcd_streams(sjs_result):
...
+        lu = get_lu_index(k)
+        if lu is not None:
+            lu_streams[lu] += float(v) / total
 
     return {
         'ag_stream_pct': ag,
         'low_urban_stream_pct': low,
-        'med_high_urban_stream_pct': med_high
+        'med_high_urban_stream_pct': med_high,
+        'lu_streams': lu_streams
```
rajadain: Hmm, perhaps we should call it `lu_stream_pct` to be consistent with the other keys, and since it does only store the percentage value, not the actual number of streams.

src/mmw/apps/modeling/mapshed/calcs.py
```diff
@@ -203,6 +205,38 @@ def manure_spread(aeu):
...
+        slope_length = (0.5 * area * (1.0 / HECTARE_PER_M2)) / (250.0 * KM_PER_M)
```
rajadain: Consider copying the source more directly instead of assuming what the numbers actually mean (even though I suspect your assumptions are correct, I'm not sure if the original formula is correct):

```python
slope_length = (0.5 * area * 10000) / 250
```

src/mmw/apps/modeling/mapshed/calcs.py
jtarricone: 250 meters. I think the decimal conversion factor convention makes this calculation (even) less clear.

rajadain: Perhaps adding a comment along the lines of the original code would be helpful.

```python
# Floor stream length to 250 meters
```

src/mmw/apps/modeling/mapshed/calcs.py
jtarricone: In that case, would the values have to be passed into the method? That would mildly hurt clarity, but probably not much since ls_factors is only a couple lines away. The expense in cycles is pretty negligible I would imagine, so I guess it's just personal preference where the declarations are, and I don't have a strong opinion either way.

src/mmw/apps/modeling/mapshed/tasks.py
jtarricone: Clarity certainly couldn't hurt, really.

src/mmw/apps/modeling/mapshed/calcs.py
rajadain: Calculating `m` inside the loop implies that it is different in each iteration, which is not the case. I think it would help clarity to move it out.

For `s` and `t`, it may be better to use the numbers directly, and add the explanation in a comment. This should reduce the number of variables to keep track of, as well as be more in sync with the original where they did use straight numbers instead of these variables.

jtarricone:
Otay I think I've addressed all points of feedback.

rajadain:
Taking another look

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -156,6 +157,8 @@ def collect_data(geop_result, geojson):
 
     z.AvSlope = geop_result['avg_slope']
 
+    z.LS = ls_factors(geop_result['lu_streams'], z.Area, z.AvSlope)
```
rajadain: This needs to be updated to `lu_stream_pct` as well.

```
[2016-05-17 17:22:50,730: ERROR/MainProcess] Task apps.modeling.mapshed.tasks.collect_data[ca2fd6d4-d0a0-449f-b0a0-1e56d04842cc] raised unexpected: KeyError(u'lu_streams',)
  File "/opt/app/apps/modeling/mapshed/tasks.py", line 160, in collect_data
    z.LS = ls_factors(geop_result['lu_streams'], z.Area, z.AvSlope)
KeyError: u'lu_streams'
```

src/mmw/apps/modeling/mapshed/calcs.py
```diff
@@ -204,6 +205,43 @@ def manure_spread(aeu):
     return n_list, p_list
 
 
+def ls_factors(lu_strms, areas, avg_slope):
+    results = [0.0] * len(lu_strms)
+    for i in range(len(lu_strms)):
+        if 0 <= avg_slope <= 1.0:
+            m = 0.2
```
rajadain: We can move the calculation of `m` outside the `for`, since it is the same for each iteration

src/mmw/apps/modeling/mapshed/calcs.py
```diff
@@ -204,6 +205,43 @@ def manure_spread(aeu):
...
+        elif 1.0 < avg_slope <= 3.5:
+            m = 0.3
+        elif 3.5 < avg_slope <= 4.5:
+            m = 0.4
+        else:
+            m = 0.5
+
+        results[i] = ls_factor(lu_strms[i] * KM_PER_M, areas[i], avg_slope, m)
```
rajadain: The contents of `lu_strms` here is not the actual length, but only a percentage. To get length in meters you'll need to multiply with `z.StreamLength`, and then convert to kilometers

src/mmw/apps/modeling/mapshed/calcs.py
```diff
@@ -204,6 +205,43 @@ def manure_spread(aeu):
...
+
+    return results
+
+
+def ls_factor(stream_length, area, avg_slope, m):
+    #  units --> stream_length:km, area:hectare, avg_slope:%, m:???
```
rajadain: `m` is unitless, it's a pure number

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -156,6 +157,9 @@ def collect_data(geop_result, geojson):
 
     z.AvSlope = geop_result['avg_slope']
 
+    z.LS = (ls_factors(geop_result['lu_stream_pct'], z.StreamLength,
```
kdeloach: I think the outer parens can be omitted.

jtarricone: Removing them starts a guessing-game with flake of how many indents is correct for a continuation.

rajadain:
There are certain values of `lu_index` which will always be 0 because we don't support them in MMW, such as Low / Med / High Residential, which are the last three. However, for all such indices, we have a value of 0.1913 instead of 0 in the result:

```
[2016-05-17 17:46:05,578: WARNING/Worker-1] ==> LS [0.22391138935905344, 0.20766533512565566, 0.23865775448374904, 0.20609358739895478, 0.1913, 0.1913, 0.2089486601044496, 0.21751368506714341, 0.1913, 0.1913, 0.21690342133316073, 0.23122989620183307, 0.25048590803691023, 0.1913, 0.1913, 0.1913]
```

jtarricone:
That must be because the constant value is being added.

src/mmw/apps/modeling/mapshed/tasks.py
rajadain: This would be the correct indentation:

```python
    z.LS = ls_factors(geop_result['lu_stream_pct'], z.StreamLength,
                      z.Area, z.AvSlope)
```

rajadain:
Looking at a sample GMS file (lines 20-35):

```
Hay/Past,1185,75,0.276,0.846124207294226,0.03,0.52
Cropland,1212,82,0.2895,0.693781528158755,0.42,0.52
Forest,2645,73,0.2173,2.51730543559902,0.002,0.52
Wetland,14,87,0.2919,0.138129109946082,0.01,0.1
Disturbed,65,89,0.32,1.94029540505005,0.08,0.1
Turfgrass,86,71,0.32,1.04784941760857,0.03,0.2
Open_Land,0,0,0,0,0,0
Bare_Rock,0,0,0,0,0,0
Sandy_Areas,0,0,0,0,0,0
Unpaved_Road,1,87,0.266608607322506,0.33230684694556,0.8,1
Ld_Mixed,2,0.15,92,74,60
Md_Mixed,314,0.52,98,79,70
Hd_Mixed,727,0.87,98,79,80
Ld_Residential,227,0.15,92,74,90
Md_Residential,1497,0.52,92,74,100
Hd_Residential,4,0.87,92,74,110
```

For land uses with area 0, the LS value is also 0. I think we should make it so

jtarricone:
Will that preclude having to check if the land use type is one of the unsupported ones?

jtarricone:
Nevermind, I can see the nonzero area for the residential entries.

rajadain:
> Will that preclude having to check if the land use type is one of the unsupported ones?

Yes, I think we can just check if the corresponding value of `z.Area` is 0 or not. This should apply to supported land uses which are simply absent from the area of interest.

src/mmw/apps/modeling/mapshed/tasks.py
kdeloach: Should be good as long as it lines up with the open paren like this:

```
z.LS = ls_factors(geop_result['lu_stream_pct'], z.StreamLength,
                  z.Area, z.AvSlope)
```

jtarricone:
Okay that condition is in place. I'm curious how it was being handled in the VB implementation...

rajadain:
+1 I think this is good to merge. Great work sticking through this!

I think the numbers we're seeing in the GMS file may be coming through the Flow Accumulation path. I still think this is the correct behavior though.

Issue #1307, opened 2016-05-13
Specialize content of dropdown for GWLFE BMPs


lewfish:
Specialize the content of the manual entry dialog for each BMP, do input validation, and actually add the modification when clicking 'Convert'.

For BMP details see: https://drive.google.com/open?id=0B8dzaPNE2DLTUV9EWC1tNXVVWDhUeHhPQzlIVlYyWVVJemxz 

This is a follow-up to #1278.

PR #1314, opened 2016-05-18
Customize BMP manual entry UIs

lewfish:
This PR adds a configuration for each GWLF-E BMP to customize the manual entry UI for each BMP.

When reviewing this PR, ignore style issues including layout and word choices as they will be addressed in https://github.com/WikiWatershed/model-my-watershed/issues/1315.

Currently, this doesn't add a modification to the ScenarioModel, and uses a mock Mapshed data model. This will be addressed in https://github.com/WikiWatershed/model-my-watershed/issues/1316

#### Testing
* Check that BMP UI still works for TR55.
* For the BMP dropdown for GWLF-E models, check that back button and clicking the dropdown button works as expected (should clear things and bring back the thumbnails).
* For each BMP check that everything matches the spec at https://drive.google.com/open?id=0B8dzaPNE2DLTUV9EWC1tNXVVWDhUeHhPQzlIVlYyWVVJemxz
* Clicking Convert will not work if there are error messages.
* Clicking Convert when the input is valid will print an object to the console with the output.
screen shot 2016-05-18 at 11 16 30 am 2
screen shot 2016-05-18 at 11 16 52 am 2
screen shot 2016-05-18 at 11 16 59 am 2

Connects #1307

### DIFF
--- a/src/mmw/js/src/modeling/gwlfeModificationConfig.js
+++ b/src/mmw/js/src/modeling/gwlfeModificationConfig.js
@@ -1,0 +1,286 @@
+"use strict";
+
+var _ = require('lodash');
+
+// These variable names include 'Name' to indicate
+// that they reference the name of the variable, and not the value.
+var n23Name = 'n23',
+    n25Name = 'n25',
+    n26Name = 'n26',
+    n28bName = 'n28b',
+    n41bName = 'n41b',
+    n41dName = 'n41d',
+    n42Name = 'n42',
+    n42bName = 'n42b',
+    n43Name = 'n43',
+    n45Name = 'n45',
+    n46cName = 'n46c',
+    n25bName = 'n25b',
+    n25cName = 'n25c',
+    UrbLengthName = 'UrbLength',
+    FilterWidthName = 'FilterWidth',
+    PctStrmBufName = 'PctStrmBuf',
+    UrbBankStabName = 'UrbBankStab',
+    QretentionName = 'Qretention',
+    PctAreaInfilName = 'PctAreaInfil',
+    UrbAreaTotalName = 'UrbAreaTotal',
+    lengthToConvertName = 'lengthToConvert',
+    areaToConvertName = 'areaToConvert',
+    percentAeuToConvertName = 'percentAeuToConvert',
+    percentAreaToConvertName = 'percentAreaToConvert',
+    AREA = 'area',
+    LENGTH = 'length',
+    FilterWidthDefault = 30,
+    QretentionDefault = 2.54;
+
+// This function converts a list of pairs to objects. We use it to
+// create objects where the keys are the values of variables, which is not
+// possible with object literal syntax. For instance,
+// x = {n23Name: 'y'} would create an object {'n23Name': 'y'} instead of
+// {'n23': 'y'} which is what we want. We could also do {n23: 'y'} but that
+// would defeat the point of using constants here which is to get
+// spell-checking from the interpreter.
+// This function is in lodash 4.12.2. In the future, we might want to upgrade
+// to get it.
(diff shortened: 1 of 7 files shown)


lewfish:
@mmcfarland: For rural vegetated buffer strips, streambank fencing, and streambank stabilization, is the maximum allowable length n42 (total length in ag areas) or n42b (total length in watershed)?

src/mmw/js/src/modeling/gwlfeModificationConfig.js
```diff
@@ -0,0 +1,286 @@
...
+    QretentionDefault = 2.54;
+
+// This function converts a list of pairs to objects. We use it to
+// create objects where the keys are the values of variables, which is not
+// possible with object literal syntax. For instance,
+// x = {n23Name: 'y'} would create an object {'n23Name': 'y'} instead of
+// {'n23': 'y'} which is what we want. We could also do {n23: 'y'} but that
+// would defeat the point of using constants here which is to get
+// spell-checking from the interpreter.
+function toObject(pairs) {
```
caseypt: Does [_.fromPairs](https://lodash.com/docs#fromPairs) accomplish this?

lewfish: Yes! That's what I was looking for.

src/mmw/js/src/modeling/gwlfeModificationConfig.js
```diff
@@ -0,0 +1,286 @@
...
+    (received from the backend) which should be displayed to the user.
+    - userInputNames is a list of non-Mapshed variables used to reference values
+    the user enters into the UI. currently all the BMPs only reference a single
+    user input variable, so using a list is overkill, but it could be useful in
+    the future.
+    - validate is a function which validates the user input using values from
+    the data model.
+    - computeOutput is a function which computes the output of the UI for a BMP,
+    the values for a set of Mapshed variables.
+*/
```
caseypt: Nice comment. I think on future projects we should consider using something like [JSDoc](http://usejsdoc.org/about-getting-started.html)

caseypt:
For validation, consider using [HTML5 form validation](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Forms/Data_form_validation).

caseypt:
Good documentation on the complicated stuff.

caseypt:
Verified that
- TR-55 modification control still works
- You can't enter values if the input is in valid
- Variables output names for each modification line up with the document.

caseypt:
If you set a value for one modification type, then open it again, should we display the current value? There is only one modification be "added" per type, not multiple, right?

caseypt:
This is more for https://github.com/WikiWatershed/model-my-watershed/issues/1315, but if the above assumption is correct, we may want to show the current value for each activate modification on the modification control instead of the list of modifications.

caseypt:
Talked over those last two comments with @lewfish. They will be addressed in #1316

caseypt:
+1. Nice work.

lewfish:
> Talked over those last two comments with @lewfish. They will be addressed in #1316

After talking to @mmcfarland, we decided to just leave things as they are, which means that clicking a BMP that has already been added will let the user reset the value, but will not show them the pre-existing value.

lewfish:
> For validation, consider using HTML5 form validation.

I think it'll be too big of a change for this PR, but it's interesting to know about for next time. It looks like the validation is based on regular expressions which isn't the best fit for this PR.

mmcfarland:
@lewfish regarding max stream length for the three km based rural BMPs, I believe it is n42 (streams in Ag), but I've got an email out to confirm.

Issue #1297, opened 2016-05-11
MapShed Workflow: Save Scenario Properties, Run Model


rajadain:
The bulk output of the MapShed GIS chain will be saved with the Project, but there will be a handful of unique properties that are saved on the Scenario (such as BMPs).

Add an endpoint that, given a Scenario, will combine its properties with the main MapShed dictionary stored in the Project, run the GWLF-E model, store the results in the database on the Scenario, and return them to the caller. Evaluate if this can be a synchronous operation or if it needs to be a Celery job. Add / update database fields as necessary.

rajadain:
Cleared the size because it was counted for Sprint ending on 5/19/2016. Was 2 points.

PR #1317, opened 2016-05-18
Add endpoint to run GWLF-E if passed a full input dataset.

jtarricone:
This implements an endpoint that, when given a data dictionary, runs GWLF-E.

A starting point for testing might be creating a project/scenario and then running the "Watershed Multi-Year" model, but it could also be tested by simply posting the results of the mapshed API call to the GWLF-E one.

Connects #1297


rajadain:
+1 this is working correctly as far as the scope of this card is concerned. However, due to limitations in GWLF-E, I'm getting this error:

```
[2016-05-18 13:43:41,899: ERROR/Worker-4] Task d19060f2-c5af-4fcd-9244-234a0c8fde69 run from job 2821 raised exception: 'DataModel' object has no attribute 'AnnDayHrs'
  File "/opt/app/apps/modeling/tasks.py", line 337, in run_gwlfe
    response_json = {'runoff': gwlfe.run(z)}
  File "/opt/app/requirements/src/gwlf-e/gwlfe/gwlfe.py", line 38, in run
    ReadGwlfDataFile.ReadAllData(z)
  File "/opt/app/requirements/src/gwlf-e/gwlfe/ReadGwlfDataFile.py", line 30, in ReadAllData
    z.AnnDayHrs += z.DayHrs[i]
AttributeError: 'DataModel' object has no attribute 'AnnDayHrs'
```

I'm working out the details, but I almost wonder if we should have an option that returns the values of the sample file if say `model_input` is empty. Something like this would be useful for developing the GUI while we work out the details here.

jtarricone:
Either an option or a separate endpoint, but either way that's a good call, it's pretty maddening to test even w/o throwing errors.

src/mmw/apps/modeling/tasks.py
```diff
@@ -332,13 +332,7 @@ def change_key(modification):
 
 @shared_task
 def run_gwlfe(model_input):
-    # TODO pass real input to model instead of reading it from gms file
-    gms_filename = join(dirname(abspath(__file__)),
-                        'mapshed/data/sample_input.gms')
```
kdeloach: Could we delete this file, if it is no longer needed?

rajadain:
In fact, instead of adding another sample endpoint, we can probably just hold off on merging this until we're ready.

src/mmw/apps/modeling/tasks.py
rajadain: Agreed, we should delete that file as part of this PR. But let's not merge until we're at a ready state.

jtarricone: Deleted, thanks for catching that!

kdeloach:
I'm working on a solution for this.

rajadain:
Closing in favor of #1326

Issue #1323, opened 2016-05-19
MapShed Workflow: UI Polish


rajadain:
- [x] Add spinner / progressbar for when gathering MapShed data
- [x] Add notification / instructions for if it fails

lewfish:
- [ ] Show x% of length or area in manual entry UI
- [ ] Disable thumbnails for BMPs that can't really be applied to the AOI

rajadain:
I would prefer to move the above two to #1298, which has more points and few work items remaining.

lewfish:
Ok, I moved them to #1298

Issue #1310, opened 2016-05-16
MapShed Workflow: Trigger MapShed run on project start


rajadain:
A smaller chunk split off from #1298.

 - [x] When the user creates a new project, we trigger a MapShed GIS chain, and poll until it completes
 - [ ] Consider saving them to the database if the user is logged in
 - [ ] Once the MapShed GIS chain completes, we trigger a GWLF-E model run for Current Conditions (with no special properties) and New Scenario (also with no special properties at first), and poll until it completes

PR #1324, opened 2016-05-19
Trigger MapShed before GWLF-E

rajadain:
## Overview

In case of GWLF-E projects, we need to gather GIS data from MapShed to hand off to the GWLF-E endpoint. This gathering must be done only once per project, since its value does not depend on any modifications.

In this PR we add methods to the `ProjectModel` which support fetching GIS data from the MapShed endpoint if the model package if GWLF-E, or a simple noop if the model package is TR55. In both cases, a promise is returned, only on whose completion does the scenario polling begin. In case of TR55, we return an already resolved promise so that scenario polling can begin immediately.

## Testing Instructions

Run migrations.

```bash
$ ./scripts/manage.sh migrate
```

Test making both TR-55 and GWLF-E projects, both while logged in and while logged out, and ensure that behavior is consistent and unsurprising.

 * Logged in and logged out behavior should be the same as before for TR55.
 * For GWLF-E, making a new project should trigger a MapShed run immediately, with GWLF-E scenarios polling once it completes.
 * Nothing should be saved if you are logged out.
 * If you are logged in, the project should save right after MapShed completes successfully. The scenarios should save before GWLF-E polling begins, and after it ends.
 * If you are logged in and reloading an existing project for which a MapShed run has completed successfully, it should only run GWLF-E after reloading, not the MapShed run again.

## Notes

Currently the GWLF-E endpoint returns static values. However, we do send the true values to the endpoint.

Currently there is no user notification of MapShed polling, and also none in case of failure. These, and other such items, have been relegated to #1323.

Should we also add a `gis_data_hash` field? I can't think of a reason, but maybe someone else can.

Connects #1310

### DIFF
--- a/src/mmw/apps/modeling/migrations/0019_project_gis_data.py
+++ b/src/mmw/apps/modeling/migrations/0019_project_gis_data.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('modeling', '0018_postgis_add_EPSG5070'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='project',
+            name='gis_data',
+            field=models.TextField(help_text='Serialized JSON representation of additional data gathering steps, such as MapShed.', null=True),
+        ),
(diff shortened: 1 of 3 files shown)


lewfish:
+1, tested. Nicely done.

Issue #1315, opened 2016-05-18
Improve style of BMP UI


lewfish:
Some issues include verbiage, layout, and the choice of the fill patterns.

caseypt:
Currently there are some spacing issues for modifications with long names:

![image](https://cloud.githubusercontent.com/assets/1042475/15367209/0525b692-1cf6-11e6-8249-ff880300cbf7.png)

mmcfarland:
We should consider a sub header for Urban | Rural to allow for some visual separation.

lewfish:
> We should consider a sub header for Urban | Rural to allow for some visual separation.

You mean to organize the thumbnails into two groups, with a header on top of each?

lewfish:
If I do that then I won't need to find different fill patterns for the urban versions of the BMP.

PR #1325, opened 2016-05-23
Improve style of GWLF-E BMP UI

lewfish:
This improves the style of the GWLF-E BMP UI. To test, make sure functionality is intact and things are more aesthetically pleasing. Compare to https://github.com/WikiWatershed/model-my-watershed/pull/1314

* Make style of Animal Waste Management (Poultry) slightly different from (Livestock)
* Add Rural and Urban thumbnail groups
* Add more vertical space in thumbnail view so labels are fully visible
* For BMPs with multiple data model variables, change the label of the input to clarify which data model variable is being converted
* In manual entry view: change size and color of info and error messages, improve spacing, and put fill pattern in line under header to improve readability

screen shot 2016-05-23 at 3 31 29 pm
screen shot 2016-05-24 at 5 29 57 pm

Connects #1315

### DIFF
--- a/src/mmw/js/src/modeling/controls.js
+++ b/src/mmw/js/src/modeling/controls.js
@@ -161,12 +161,12 @@ var ManualEntryView = Marionette.CompositeView.extend({
 
     ui: {
         'backButton': '.back-button',
-        'convertButton': '.convert-button'
+        'applyButton': '.apply-button'
     },
 
     events: {
         'click @ui.backButton': 'clearManualMod',
-        'click @ui.convertButton': 'convert',
+        'click @ui.applyButton': 'applyModification',
         'keyup': 'onKeyUp'
     },
 
@@ -191,7 +191,7 @@ var ManualEntryView = Marionette.CompositeView.extend({
 
     onKeyUp: function(e) {
         if (e.keyCode === ENTER_KEYCODE) {
-            this.convert();
+            this.applyModification();
         }
     },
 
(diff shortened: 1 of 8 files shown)


mmcfarland:
I think the visual style of the display bullet points would be better as a table (perhaps with no border) and condense the verbiage.

| Length of streams in Ag areas | 300km |
| ---      | ---       |
| **Length of steams in Watershed** | **1000km** |

Also, "Convert" is a holdover from when we though were were merely converting a land use type into a BMP, although now understanding how it actually works I think "Apply" is better suited.

lewfish:
> I think the visual style of the display bullet points would be better as a table (perhaps with no border) and condense the verbiage.

Done. See updated screenshot.

lewfish:
> Also, "Convert" is a holdover from when we though were were merely converting a land use type into a BMP, although now understanding how it actually works I think "Apply" is better suited.

I think 'Modify' might be a better replacement for 'Convert' because it makes more sense in string names and internal variables. For instance, 'Length of stream to modify' rather than 'Length of stream to apply', and `lengthToModify` rather than `lengthToApply.` We could use 'modify' in those places, but still use 'apply' on the button itself.

caseypt:
+1. Looks more aesthetically pleasing.

src/mmw/js/src/modeling/controls.js
```diff
@@ -111,8 +111,22 @@ var InputInfoView = Marionette.ItemView.extend({
     },
 
     templateHelpers: function() {
+        var output = this.model.get('output'),
+            errorMessage = output && output.errorMessages[this.userInputName],
+            infoMessage = output && output.infoMessages[this.userInputName],
```
caseypt: I did not know that using `&&` in this way would return the correct value here. Pretty interesting.

Issue #1321, opened 2016-05-19
Install GWLF-E from PyPI


kdeloach:
We should link MMW to the latest release of gwlf-e just like we do for tr55.

Ref: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/requirements/base.txt#L16

kdeloach:
The latest release is 0.2.0.

rajadain:
What is the package name in PyPI? I searched for `gwlf`, `gwlfe`, `gwlf-e`, and `gwlf_e` and didn't find anything.

rajadain:
Ah nevermind, found it: https://pypi.python.org/pypi/gwlf-e/

Their search index probably hasn't updated, or it may have trouble searching for names with hyphens in them. Searching for "gwlf-e" doesn't result in our package.

PR #1326, opened 2016-05-23
MapShed Fix Data Formats and Run GWLF-E

rajadain:
## Overview

Fixes a number of data issues noted in #1319. Finally runs GWLF-E on the actual MapShed data, and returns real results instead of sample values. See commit messages for details.

This depends on the latest version of GWLF-E `0.3.0`. We also switch to pulling from PyPI instead of GitHub.

## Testing Instructions

Reprovision app to pull in latest GWLF-E dependency from PyPI

```bash
$ vagrant provision app
```

Rebundle the scripts:

```bash
$ ./scripts/bundle.sh --debug
```

Debug Celery:

```bash
$ ./scripts/debugcelery.sh
```

Create a project using the Watershed Mult-Year Model, and ensure that there are no errors in Celery, and the output is correctly rendered in both Runoff and Water Quality tabs.

Supersedes #1317 
Connects #1319 
Connects #1321


src/mmw/apps/modeling/tasks.py
```diff
@@ -332,10 +332,14 @@ def change_key(modification):
...
-    z = parser.GmsReader(gms_file).read()
+    z = parser.DataModel(model_input)
+    output = StringIO()
+    writer = parser.GmsWriter(output)
+    writer.write(z)
+
+    output.seek(0)
+
+    reader = parser.GmsReader(output)
+    z = reader.read()
```
kdeloach: May want to add a comment explaining the point of this.

It's kind of a hack, but the alternative is spending more time moving defaults from the parser to the DataModel.

rajadain: Will do

kdeloach:
+1 👏

rajadain:
Some clarification added in 731d48f

rajadain:
Added commit that switches to pulling GWLF-E from PyPI instead of GitHub. Amended PR description and testing instructions accordingly.

mmcfarland:
Can you add a commit that enables the Multi-Year model selection on staging?  We can revert this commit prior to a release if we're not going to include the feature in the next release (though the plan is to  include it).

rajadain:
Good call. I think 03d29b2 should do it.

rajadain:
Squashed fixups.

jtarricone:
+1

In consideration of #1323 this looks good. It's worth noting that it's pretty fickle about completion, in that it times out frequently and consistently for certain AoIs. For instance, it completed for Philadelphia County, but timed out for King County (Seattle) and Dallas County. However, it successfully completed for all the school districts within both King & Dallas counties, so I guess it's just an issue of AoI size?

rajadain:
I think it is most directly related to AoI size, and then to the number of streams captured within the AoI. Those with more streams will probably take longer, as it seems that the `nlcd_streams` task takes longest. Still, I keep encountering certain Celery `chord_unlock` lockups whose underlying causes are unclear to me. We have a couple points this sprint for "various issues", so I will likely investigate them later, and merge this in for now, since it gives us a modicum of functionality.

lewfish:
After provisioning for the first time after incorporating this PR, you will need to remove the `src/mmw/requirements/src/gwlfe` directory so that pip can install the new version of gwlfe.

Issue #1298, opened 2016-05-11
MapShed Workflow: Update Front-End Code


rajadain:
Following #1296 and #1297, we will have two endpoints: one to run the MapShed GIS chain for the project, and another to run the GWLF-E model for the scenario. Update the controllers / task runners in the front-end JavaScript code to implement the following interaction model:

 * [x] When the user creates a new project, we trigger a MapShed GIS chain, and poll until it completes #1310 
 * [x] Once the MapShed GIS chain completes, we trigger a GWLF-E model run for Current Conditions (with no special properties) and New Scenario (also with no special properties at first), and poll until it completes #1310 
 * [x] Once the MapShed GIS chain completes, we inspect the output to retrieve properties used to fill in Land Use percent, other quantities #1314 
 * [x] Once the GWLF-E model run completes, ~~we query the database for~~ it should return results of the model run and populate the charts and tables. These results should also be saved to the database so that when the scenario is re-loaded they load instantly. #1329 

Consider adding error handling and retry options for when any of the steps above fail.

lewfish:
> Once the GWLF-E model run completes, we query the database for results of the model run and populate the charts and tables

I'm a little confused by the above statement. The backend should respond with the results on the final poll, so there shouldn't be a separate query for the results.

rajadain:
Good point. The backend should respond with the results, in addition to saving them to the database. When re-loading a saved scenario, we can probably check if the results are already populated or not before running them again, a-la TR-55.

rajadain:
Updated the description to reflect. Feel free to further edit it to fit.

rajadain:
Originally 8 points, -3 to #1310

rajadain:
-2 to #1327

lewfish:
- [ ] Show x% of length or area in manual entry UI
- [ ] Disable thumbnails for BMPs that can't really be applied to the AOI

lewfish:
-1 to #1338

lewfish:
-1 te #1343

rajadain:
-1 to #1345

PR #1329, opened 2016-05-24
MapShed Show Status

rajadain:
## Overview

Adds UI elements that communicate that GIS Data is being gathered for MapShed projects. Only the last two commits are relevant to this PR.

## Demo

Waiting for GIS Data:

image

Waiting for Results:

image

Have Results:

image

Error State:

image

## Testing Instructions

Bundle the scripts.

 * Create a new MapShed project. You should see a spinner as GIS Data is gathered. Once the data is gathered, that screen should be replaced by the Result Type tab headers, each with spinners of their own as they wait for respective results.
 * Reload a saved MapShed project. You should not see a spinner, since GIS Data is already gathered.
 * Create a TR-55 project. You should not see a spinner, as no GIS Data needs to be gathered.
 * Select a very large area. GIS Data step should timeout, and you should see the spinner turn into an error sign.

Builds on top of #1326 
Connects #1298 
Connects #1323


rajadain:
Rebased on `develop`

lewfish:
When going to the model view before the analyze results are ready, the spinner in the analyze tab is not visible.

lewfish:
In the model view, the 'Gathering Data' spinner is visible as expected for the 'Current Conditions' scenario. But after switching to the 'New Scenario' tab, the spinner is not visible.

rajadain:
Multiple POSTs to `/analyze` are a real problem, refactoring that may be too big for this card. Created #1334 to tackle that.

rajadain:
The fixup above should address the lack of spinner when switching between multiple scenarios

rajadain:
Rebased on develop, merged the fixup into the main commit 6644459. Also added 66331fa to de-duplicate GWLF-E runs.

hectcastro:
@azavea-bot rebuild

rajadain:
@azavea-bot rebuild

hectcastro:
@azavea-bot rebuild

lewfish:
When I create a new GWLF-E model, and click the 'New Scenario' tab, I see 'Gathering Data' and then I the spinners on top of the 'Runoff' and 'Quality' tabs go for a second or two, and then the spinners stop and there is nothing for another 5-10 secs before the results appear. This doesn't seem to happen with TR55.

rajadain:
Okay, I've refactored `fetchResultsIfNeeded` to use the same single-promise technique as being used by `fetchGisDataIfNeeded` and I think it is working in all cases now. Can you test one more time?

src/mmw/js/src/core/models.js
```diff
@@ -213,6 +213,21 @@ var TaskModel = Backbone.Model.extend({
     }
 });
 
+var TaskViewModel = Backbone.Model.extend({
```
lewfish: Consider renaming this to `TaskMessageViewModel` to be consistent.

lewfish:
+1, tested

src/mmw/js/src/core/models.js
rajadain: Good call

rajadain:
Reorganized commits. Will merge shortly. Thanks for reviewing.

Issue #1316, opened 2016-05-18
Connect GWLF-E BMP UI to ScenarioModel


lewfish:
Currently, clicking Convert in the BMP UI just prints the variables that need to be added to the Mapshed data model. The Mapshed data model is also a mock object. We need to:
* add a new GwlfeModificationModel and modify the ScenarioModel (esp. the parse method) to handle it 
* (maybe) modify the backend to handle these modifications
* get the UI to add the modification to the ScenarioModel
* get the real Mapshed data model from the ProjectModel
* not show the 'Conservation Practices' dropdown until the data model comes back from the server

rajadain:
Once the MapShed run completes, the output is saved in ProjectModel's `gis_data` field. That data can be used to configure the various modification UIs.

Issue #1280, opened 2016-05-05
Add BMP info box


lewfish:
<img width="336" alt="screen shot 2016-05-05 at 1 48 17 pm" src="https://cloud.githubusercontent.com/assets/1896461/15051882/d8c3b6da-12c8-11e6-8ab5-844107886cf2.png">

@jfrankl Should this box open on hover or click?

Issue #1279, opened 2016-05-05
Show BMPs for GWLF-E


lewfish:
Show list of BMPs in sidebar. Each icon symbolizes the original land type and the new BMP type.

<img width="267" alt="screen shot 2016-05-05 at 1 48 10 pm" src="https://cloud.githubusercontent.com/assets/1896461/15051803/7f06cb50-12c8-11e6-9103-b852210a8592.png">

PR #1336, opened 2016-05-26
Add BMP sidebar and connect to ScenarioModel

lewfish:
This PR adds a sidebar/popup that shows the BMPs that have been added and connects the BMP UI to the ScenarioModel, so that adding BMPs triggers running the GWLF-E model with the modifications. It also makes some other related changes to the UI.

screen shot 2016-05-27 at 1 06 15 pm 2
screen shot 2016-05-27 at 1 06 18 pm 2
screen shot 2016-05-27 at 1 06 23 pm 2

##### Testing
* For a GWLF-E model, try adding BMPs. The sidebar should appear when there are modifications. Check that the last two BMPs work.
* Look in the console for the data that is sent for the GWLF-E run after the BMP is added. (I found that Chrome crashed when trying to look at the payload in the Network tab). Verify that the correct variables are set. For instance, 'Infiltration' should set `PctAreaInfil` to the fraction (0-1) computed from the input and `Qretention` to 2.54.
* Check that the popup show correct data, delete button works (and triggers GWLF-E run), and that it closes when clicking outside or on close button.
* Check that 'Conservation Practices' doesn't appear until Mapshed results are back, and that 'Compare' button is hidden.
* When logged in, check that BMPs are saved and show up after hitting refresh in the model view.
* Check that TR55 and Compare view still work.

Connects #1316 
Connects #1279 
Connects #1280


hectcastro:
@azavea-bot rebuild

rajadain:
The popup

image

shows up on hover, but doesn't go away when I stop hovering. Instead, I have to click the X button. This is not intuitive: usually hover behaviors go away when you stop hovering. If on the other hand we want it to stay, you should have to click to show it.

lewfish:
> shows up on hover, but doesn't go away when I stop hovering. Instead, I have to click the X button. This is not intuitive: usually hover behaviors go away when you stop hovering. If on the other hand we want it to stay, you should have to click to show it.

I see what you mean. I did it this way because the popup has content in it that needs to be clicked. The alternative you suggest, which is to only make the popup stay open if you click the thumbnail, could also be counterintuitive though. I can imagine users thinking there is a bug that is making the popup close when they go to click the delete button.

rajadain:
Perhaps the there could be a tooltip on hover, indicating the name of the BMP, and when you click it the popup shows up? I agree that it should remain active lest you be unable to delete it.

Also, we haven't used the term BMP in any user facing area in the app so far. Is that what we want to call it?

lewfish:
> Perhaps the there could be a tooltip on hover, indicating the name of the BMP, and when you click it the popup shows up? I agree that it should remain active lest you be unable to delete it.

Good idea. I'll do something like that.

lewfish:
> Also, we haven't used the term BMP in any user facing area in the app so far. Is that what we want to call it?

I'm not sure what to call it. The wireframe said 'Manually Add BMPs'. I think I'll call it 'Conservation Practices.'

rajadain:
Maybe "Applied Modifications"? To differentiate from the "Conservation Practices" label above

rajadain:
That may almost make the overlay wide enough to have a name against each icon, removing the need for a tooltip. We could just have the icon and the name then, and when you click it you get the pop up.

mmcfarland:
My recollection of the wireframe is that the BMP bar was to have a symbol on top that had a hover label that said "Manually Added BMPs".  Since we only allow manually added BMPs now, and that's true about the term, I agree "Applied Conservation Practices" seems good.  I think is should still only show on hover as that bar is meant to be slim and aligned to the edge of the screen. 

Is it possible to create the bar along the full edge as in the wireframes?

lewfish:
> My recollection of the wireframe is that the BMP bar was to have a symbol on top that had a hover label that said "Manually Added BMPs". Since we only allow manually added BMPs now, and that's true about the term, I agree "Applied Conservation Practices" seems good. I think is should still only show on hover as that bar is meant to be slim and aligned to the edge of the screen.
> 
> Is it possible to create the bar along the full edge as in the wireframes?

I didn't realize the "Manually added BMPs" label in the wireframe was supposed to be a tooltip, but that makes sense. I can do that and move the bar all the way to the edge. Moving the bar all the way to the edge might not look good though... I was trying to copy the style of the Leaflet controls which have some space between them and the edge of the screen.

mmcfarland:
@lewfish yeah, when I viewed the screenshot of the wireframe on your computer is was all washed out and didn't show the error, though on mine it shows up pretty clearly.  You may want to adjust your contrast?

rajadain:
Tested, works as advertised. Nice work!

We should probably make a card for making a Compare View for this, since as I switch between scenarios the charts reset to the first one, so there's really no way to compare them presently.

rajadain:
I will take another look if there are any more changes. Except for the pop-up behavior, I like it.

rajadain:
Consider encapsulating the functionality that combines Project's `gis_data` with Scenario's modifications into a function on the ScenarioModel. This would help with reuse, for example in #1341.

lewfish:
I made the suggested changes and the screenshots are updated.

rajadain:
Thanks, will take another look

lewfish:
I just noticed a problem. The tooltips still show up when the popup is open. I think I'll turn them off while the popup is open.

screen shot 2016-05-27 at 1 57 45 pm 2

rajadain:
>  I think I'll turn them off while the popup is open.
Good idea.

lewfish:
Fixed it.

rajadain:
Changes look good. +1!

rajadain:
You may also have to adjust the spacing of the Layer Selector now that the buttons are back on the right

image

rajadain:
And the opacity sliders

image

PR #1337, opened 2016-05-26
Convert decimals to floats

rajadain:
## Overview

The database returns Decimal values, however they are not JSON serializable, thus we convert them to floats which are. This is required not just for communicating to the front-end but also between Celery tasks, which use JSON serialization as the intermediate format.

## Testing Instructions

In `collect_data`, add a line to print out values of `z['PointNitr']` and `z['PointPhos']`:

```python
print('==> PointNitr: {}'.format(z['PointNitr'])
print('==> PointPhos: {}'.format(z['PointPhos'])
```

Then run `./scripts/debugcelery.sh` and make a MapShed project and inspect the output. Ensure they are floats.

Connects #1333

### DIFF
--- a/src/mmw/apps/modeling/mapshed/calcs.py
+++ b/src/mmw/apps/modeling/mapshed/calcs.py
@@ -305,8 +305,8 @@ def point_source_discharge(geom, area):
         cursor.execute(sql, [geom.wkt])
         mg_d, kgn_month, kgp_month = cursor.fetchone()
 
-        n_load = [kgn_month] * 12 if kgn_month else [0.0] * 12
-        p_load = [kgp_month] * 12 if kgp_month else [0.0] * 12
+        n_load = [float(kgn_month)] * 12 if kgn_month else [0.0] * 12
+        p_load = [float(kgp_month)] * 12 if kgp_month else [0.0] * 12
         discharge = [float(mg_d * days * LITERS_PER_MGAL) / area
                      for days in MONTHDAYS] if mg_d else [0.0] * 12
 


hectcastro:
@azavea-bot rebuild

rajadain:
Don't pay heed to the failing check, that's been fixed in #1329

jtarricone:
I just tested, the two example data members are float types.

rajadain:
Thanks! (Sorry, thought you were going to be out this morning which is why I reassigned). Do I have a +1?

jtarricone:
Unless @kellyi has any objections, +1
I tried to test it last night while I was here, the environment takes sooooooooo long to start up.

kellyi:
Quick q: I'm not sure whether this is replicating something originally done in NumPy, but if so do we need these to be the `float64` precision used in NumPy? Probably doesn't matter much, but worth asking.

rajadain:
There may be some loss of precision, but the central issue is serializability. `float` is easily serializable in JSON, whereas a `Decimal` is not. Because of the database adapter we are using, all numeric fields are returned as `Decimal` and we have to cast to `int` and `float` before we can serialize them.

Given the values are of the sort `Decimal('480.2500000000000000')` (see original issue), I don't think the loss of precision will harm us here.

rajadain:
@azavea-bot rebuild

kellyi:
It looks like that creates an array of floats, but/and two aois I tried returned an array of 12 0.0s. Here's the second aoi:

screen shot 2016-05-27 at 3 03 17 pm

and here's its array:

```sh
[2016-05-27 15:01:35,622: WARNING/Worker-1] PointNitr: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
[2016-05-27 15:01:35,623: WARNING/Worker-1] PointPhos: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
```

Am I missing a dataset for this maybe?

rajadain:
If no point sources are found within your area of interest, it just returns 0s.

The query is here: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/mapshed/calcs.py#L296-L301

You can run it yourself by just replacing `%s` with a GeoJSON. You can use http://geojson.io/ to make one.

You can run a query against the database by doing

```
./scripts/manage.sh dbshell
```

kellyi:
Got it. Tried again with this aoi -- 

screen shot 2016-05-27 at 3 12 14 pm

and got this output:

```sh
[2016-05-27 15:12:03,603: WARNING/Worker-1] PointNitr: [27475.166666666668, 27475.166666666668, 27475.166666666668, 27475.166666666668, 27475.166666666668, 27475.166666666668, 27475.166666666668, 27475.166666666668, 27475.166666666668, 27475.166666666668, 27475.166666666668, 27475.166666666668]
[2016-05-27 15:12:03,604: WARNING/Worker-1] PointPhos: [3141.9166666666665, 3141.9166666666665, 3141.9166666666665, 3141.9166666666665, 3141.9166666666665, 3141.9166666666665, 3141.9166666666665, 3141.9166666666665, 3141.9166666666665, 3141.9166666666665, 3141.9166666666665, 3141.9166666666665]
```

Those are definitely floats. +1

Issue #1338, opened 2016-05-26
Disable thumbnails for invalid BMPs


lewfish:
Some BMPs don't apply to a given AOI because the area or length to modify is zero. We should disable the thumbnails for these invalid BMPs.

PR #1339, opened 2016-05-26
Disable invalid BMPs 

lewfish:
This PR disables BMPs that are invalid for an AOI. BMPs can be invalid for an AOI when the areas and lengths to modify are zero. If a BMP is disabled, the thumbnail is grayed out, the cursor remains an arrow, and clicking does not work.

Only the last commit is relevant to this PR.

screen shot 2016-05-26 at 4 55 29 pm 2

To test, try out different AOIs. Check that all enabled BMPs have non-zero values for the areas/lengths being modified. In urban AOIs, BMPs involving crops and streams should be disabled. 

Connects #1338

### DIFF
--- a/src/mmw/js/src/modeling/controls.js
+++ b/src/mmw/js/src/modeling/controls.js
@@ -44,7 +44,19 @@ var ThumbSelectView = Marionette.ItemView.extend({
     template: thumbSelectTmpl,
 
     initialize: function(options) {
-        this.model.set('activeMod', null);
+        var modKeys = _.flatten(_.pluck(this.model.get('modRowGroups'), 'rows'), true),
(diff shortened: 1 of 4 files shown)


rajadain:
Not sure if this should be part of this PR or the other one, but it looks like this value should be divided by 1000:

image

rajadain:
Or should be renamed to `m` instead of `km`

rajadain:
The enabling and disabling works though, found no agricultural items were enabled in the above extremely urban area of interest. +1

lewfish:
> Not sure if this should be part of this PR or the other one, but it looks like this value should be divided by 1000:

I'll fix this in #1344 by dividing by 1000, since the instructions from Barry say that the units should be in km.

Issue #1343, opened 2016-05-27
Improve info messages in manual entry UI


lewfish:
When entering a length of area in manual entry BMP UI, the info message says 'x%', but it should say 'x% of agricultural area' or the like.

PR #1344, opened 2016-05-27
Improve info message in manual entry UI

lewfish:
We've improved the info message in the manual entry UI. See below. To test, make sure info message makes sense for all BMPs.

screen shot 2016-05-27 at 3 10 25 pm 2

Connects #1343

### DIFF
--- a/src/mmw/js/src/modeling/controls.js
+++ b/src/mmw/js/src/modeling/controls.js
@@ -432,7 +432,7 @@ var GwlfeConservationPracticeView = ModificationsView.extend({
                     rows: [['urban_buffer_strips', 'urban_streambank_stabilization', 'water_retention', 'infiltration']]
                 }
             ],
-            dataModel: JSON.parse(App.currentProject.get('gis_data')),
+            dataModel: gwlfeConfig.cleanDataModel(JSON.parse(App.currentProject.get('gis_data'))),
             errorMessages: null,
             infoMessages: null
         });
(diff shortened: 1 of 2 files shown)


lewfish:
Added a commit to convert UrbLength to km

screen shot 2016-05-31 at 12 03 37 pm 2

rajadain:
We changed the display to divide by 1000 for KM units, but wouldn't we also have to change the data output to be multiplied by 1000? Or does that always use the percentages?

lewfish:
> We changed the display to divide by 1000 for KM units, but wouldn't we also have to change the data output to be multiplied by 1000? Or does that always use the percentages?

Looking at this again, I think we just forgot to divide UrbLength by 1000. Then we won't need to do the conversion on the frontend.

rajadain:
UrbLength should be in meters according to the [GMS File Key Spreadsheet](https://docs.google.com/spreadsheets/d/12hsklRHwl6sNum0xPk6d5Cw6ONDA3X6BR5-Pc-NwJBc/edit#gid=0) and the [GMS documentation](https://docs.google.com/document/d/1M5GgnLGfJvv3mZUjoVFpkGZZ4dX6oqqq9n0sZgk7KDs/edit)

lewfish:
> UrbLength should be in meters according to the GMS File Key Spreadsheet and the GMS documentation

Then things are as they should be. The output var is either in km or is a %

rajadain:
+1 tested, makes sense!

Issue #1334, opened 2016-05-25
Analyze / GWLF-E Triggered Too Often


rajadain:
Currently, we POST to `/analyze` when the use finishes drawing / selecting an area of interest. If they don't wait for the results but instead proceed to modeling a scenario, two more `/analyze` POSTs are made: one for each Scenario (Current Conditions and New Scenario). If they switch between scenarios while waiting, yet more requests are fired.

In addition, when waiting for the MapShed job to return, if one switches between scenarios, it triggers multiple GWLF-E requests once the MapShed job returns, because [this line](https://github.com/rajadain/model-my-watershed/blob/6644459884bfb8cc0a73d35f4efed971618378d9/src/mmw/js/src/modeling/views.js#L677) is executed multiple times.
All these requests can clog up Celery workers, and affect performance of the site for all users.

One solution may be to add an `fetchAnalysisIfNeeded` function to the `ProjectModel` which returns a promise. If it is called the first time, then it fires a request and returns a promise. If it is called later, it returns the promise of the existing request. If analysis is not needed, it returns an immediately resolved promise. The function may take an optional `force` parameter that would force the request regardless of existing ones.

A similar solution should be applied to `fetchResultsIfNeeded`. See also #451

PR #1347, opened 2016-05-27
Analyze Only When Needed

rajadain:
## Overview

By using the single promise pattern used in `fetchGisDataIfNeeded` and `fetchResultsIfNeeded`, we ensure that if there are multiple calls to `analyze/`, only the first one sends a request to the server. All the rest are given the promise of the first request.

We also improve UX by showing results of Analyze immediately in the Analyze tab in the Modeling view, since the results have (most likely) already been calculated. In case they rush into the Modeling view, or re-load a saved scenario, the Analyze tab will be triggered at the same time as the Results tab, and will render independently. Furthermore, we will only trigger an actual server call only once.

We further improve the UX by remembering the Analyze results when transitioning from Modeling view to Analyze view. Previously going to the Analyze view would always trigger a new request. Now we remember the result, unless coming in from Draw, in which case old results are deliberately forgotten.

## Testing Instructions

 * Going from Draw to Analyze should trigger 1 request to `/analyze`
 * Once the request is finished, going to Modeling should not trigger any requests to `/analyze`. The analysis should be immediately visible, regardless of the status of the modeling results.
 * Adding or removing scenarios should not trigger analysis
 * Reloading a saved project should trigger 1 request to `/analyze`
 * Going from Modeling to Analyze should not trigger any requests to `/analyze`
 * Going from Analyze to Draw, but not changing the Area of Interest, instead just clicking the big green button to go forward to Analyze should not trigger any requests to `/analyze`
 * Going from Analyze to Draw and changing the Area of Interest should trigger 1 new request to `/analyze`

Connects #1334

### DIFF
--- a/src/mmw/js/src/analyze/controllers.js
+++ b/src/mmw/js/src/analyze/controllers.js
@@ -1,7 +1,6 @@
 "use strict";
 
-var _ = require('underscore'),
-    App = require('../app'),
+var App = require('../app'),
     router = require('../router').router,
     views = require('./views'),
     models = require('./models'),
@@ -48,15 +47,11 @@ var AnalyzeController = {
     },
 
     analyze: function() {
-        var aoi = JSON.stringify(App.map.get('areaOfInterest')),
-            analyzeModel = App.analyzeModel || createTaskModel(aoi),
+        var aoi = App.map.get('areaOfInterest'),
+            analyzeModel = models.AnalyzeTaskModel.getSingleton(App, aoi),
             analyzeResults = new views.ResultsView({
                 model: analyzeModel
             });
-
-        if (!App.analyzeModel) {
-            App.analyzeModel = analyzeModel;
-        }
 
         App.state.set('current_page_title', 'Geospatial Analysis');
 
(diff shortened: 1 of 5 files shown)


lewfish:
> Reloading a saved project should trigger 1 request to /analyze

Maybe this is for another issue, but shouldn't it use the saved results instead of triggering another request?

rajadain:
> shouldn't it use the saved results

We actually don't save the results of analyze. I could have added a spot in the `modeling_project` table to do that, but thought it not useful enough. Since we're already lowering the number of analyses from 3-6 per session to 1, I think the benefit of storing it would be marginal at this point.

src/mmw/js/src/analyze/controllers.js
```diff
@@ -49,12 +49,15 @@ var AnalyzeController = {
...
+            analyzeModel = App.analyzeModel || createTaskModel(aoi),
             analyzeResults = new views.ResultsView({
                 model: analyzeModel
             });
 
-        App.analyzeModel = analyzeModel;
+        if (!App.analyzeModel) {
+            App.analyzeModel = analyzeModel;
+        }
+
```
lewfish: Consider encapsulating this logic in a utility function associated with the App.

lewfish:
+1, tested. I'm glad we finally decided to do this!

rajadain:
Added a singleton pattern to reduce code redundancy. Will merge shortly.

Issue #1335, opened 2016-05-26
MapShed: Use DRB if possible


rajadain:
Most stream related methods in MapShed have a `drb` parameter, which when set to true uses DRB streams instead of NHD streams. Add functionality that tests whether or not the area of interest is completely contained within the DRB region, and if so, use DRB instead of NHD.

rajadain:
There is _one_ case where we should always use NHD. I can't remember which one it is, but @ajrobbins should be able to track it down.

mmcfarland:
I believe it is in the stream erosion routine?

rajadain:
The Erosion Coefficient is looked up from the database. "Streambank Erosion" are treated as static values, according to the [GMS File Key](https://docs.google.com/spreadsheets/d/12hsklRHwl6sNum0xPk6d5Cw6ONDA3X6BR5-Pc-NwJBc/edit#gid=0). Not sure where this would fit in.

ajrobbins:
BME says to use NHD streams for the "Stream Bank Erosion routine". If it looks like the routine uses static values, we should check with him to confirm.

rajadain:
Just spoke with BME on the phone. He said to always use the low-res (NHD) streams for everything **except** the LS Factor. The LS Factor should use the DRB stream length if available, else it should use NHD. For everything else, such as `StreamLength`, `AgLength`, `UrbLength`, `n42`, `n42b`, `n46e`, `n46f`, etc we should _always_ use NHD.

jamesmcclain:
That should unburden the geoprocessing service, somewhat.

rajadain:
Alas! All those tests against the DRB dataset were for naught. May I learn about overengineering and too-eager optimizations from this.

jamesmcclain:
Those optimizations should help support larger queries over the NHD dataset, so I think there was still some gain.

rajadain:
This is the DRB GeoJSON: [DRB.geojson.txt](https://github.com/WikiWatershed/model-my-watershed/files/287298/DRB.geojson.txt)

Obtained thusly:

```sql
SELECT ST_AsGeoJSON(ST_ConvexHull(ST_Collect(geom)))
FROM drb_streams_50;
```

mmcfarland:
We also already have a simplified boundary polygon that's used to enable/disable the DRB streams layer selector.

rajadain:
[...]

Issue #1330, opened 2016-05-25
Warning message doesn't show for large AOIs


lewfish:
When creating a very large AOI above the area threshold, the action is canceled, but no message is displayed to the user explaining why this happened.

rajadain:
Consider also adding a message when there are no streams within the area of interest.

PR #1348, opened 2016-05-31
Prefer DRB Stream Length for LS

rajadain:
## Overview

We prefer DRB Stream Length for LS calculations if possible for the selected Area of Interest. If the AoI is outside the DRB, then we fall back to using standard NHD+ stream length.

## Testing Instructions

Add the following line [here](https://github.com/rajadain/model-my-watershed/blob/4132d48d10e418b77f1ce9919c9a2329e36849f4/src/mmw/apps/modeling/mapshed/calcs.py#L255):

```python
print("==> Calculating Stream Length Using {}".format('DRB' if drb else 'NHD+'))
```

And then run `./scripts/debugcelery.sh`. The DRB region looks like this:

image

When you draw a shape within this area, you should see two messages in the Celery log:

```
[WARNING] ==> Calculating Stream Length Using NHD+
[WARNING] ==> Calculating Stream Length Using DRB
```

If you draw a shape outside this area, you should see only one message for NHD+.

Connects #1335

### DIFF
--- a/src/mmw/apps/modeling/mapshed/tasks.py
+++ b/src/mmw/apps/modeling/mapshed/tasks.py
@@ -37,6 +37,7 @@ from apps.modeling.mapshed.calcs import (day_lengths,
 NLU = settings.GWLFE_CONFIG['NLU']
 NRur = settings.GWLFE_DEFAULTS['NRur']
 AG_NLCD_CODES = settings.GWLFE_CONFIG['AgriculturalNLCDCodes']
+DRB = settings.DRB_PERIMETER
 ACRES_PER_SQM = 0.000247105
 HECTARES_PER_SQM = 0.0001
 SQKM_PER_SQM = 0.000001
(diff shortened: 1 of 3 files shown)


kellyi:
+1. Checked and confirmed that this checks whether a polygon's within the DRB and, if so, opts to use that for the calculation.

I ran into two probably unrelated errors while testing this: one's the `float division by zero` thing for AOIs like this one:

screen shot 2016-05-31 at 5 26 11 pm

The other's that Celery times out eventually for large AOIs -- > than say 100-500 sq km. Might be worth adding a frontend check with a guard for AOIs larger than MMW can calculate.

rajadain:
We have https://github.com/WikiWatershed/gwlf-e/issues/61 and #1330 for the two issues you found. Since the functionality of this card has been tested and +1'd, I'm going to go ahead and merge. Thanks!

rajadain:
Would like a confirmation from @mmcfarland that the shape above is the right one to use, rather than this one:

image

mmcfarland:
I confirm, the former is a simplified version of the actual boundary, let's use that to distinguish when to switch stream sources.

Issue #1341, opened 2016-05-27
MapShed: Export Scenario GMS


rajadain:
Currently the GMS file exported by #1327 and #1328 are the project's, and do not reflect any modifications applied in the scenario.

Use an approach similar to #1336 which merges the scenario modifications to the project GMS file, and sends that to the export GMS file endpoint.

Also move the button in the UI from the Project menu to the Scenario menu, since it will now be Scenario specific rather than Project specific.

PR #1349, opened 2016-05-31
Export GMS from Scenario instead of Project

rajadain:
## Overview

Since each scenario can contribute different values to the GMS file given its modifications, we change to exporting from the scenario instead of the project.

Furthermore, we now allow Current Conditions to have a dropdown menu, since we need it to allow GMS Export. While this does allow duplication of the "Share" functionality, since sharing the Current Conditions scenario is the same as sharing the project, consistency in menus is more important, and this duplication is preferable to Current Conditions not having a Share option at all when all other scenarios have it.

## Testing Instructions

Checkout branch and bundle scripts.

Create a MapShed project. Add some modifications to New Scenario. Export a GMS file from Current Conditions and New Scenario. Diff the downloaded files and ensure that they are not identical.

Ensure that Export GMS functionality is not available on TR-55 projects.

Connects #1341

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -826,10 +826,6 @@ var ScenarioModel = Backbone.Model.extend({
                 }
             };
 
-        // TODO remove before merging
-        console.log('gwlfe postData');
-        console.log(JSON.parse(gisData.model_input));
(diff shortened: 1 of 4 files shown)


lewfish:
+1, tested

Issue #1346, opened 2016-05-27
Optimize Map Control Positioning


rajadain:
Originally the Map Controls were in the top right. Then we added the sidebar collapse button, and had to move them to the left #1142. Then we added the modification sidebar and moved them back to the right #1336. Currently this is alright since the sidebar collapse button is missing #1345, but as soon as it is fixed we will need to figure out where to position map controls so that they neither obscure anything nor are obscured by anything.

ajrobbins:
@jfrankl

rajadain:
Shares a point with #1345.

Currently has been set to be like this in #1350:

![image](https://cloud.githubusercontent.com/assets/1430060/15695536/34ade102-2776-11e6-9bd2-ada79ef0e7b3.png)

This card may be revisited if this positioning is not ideal.

Issue #1345, opened 2016-05-27
Regression: Collapse Sidebar in Modeling View


rajadain:
Due to recent changes to `develop`, the Analyze / Modeling sidebar in the Model View is no longer collapsible. Restore this functionality.

rajadain:
Reloading an existing scenario restores this button. It is only when moving from Analyze to Model that it fails.

PR #1350, opened 2016-06-01
Restore sidebar collapse functionality

rajadain:
## Overview

The sidebar collapse functionality would previously be enabled by the pre-handler of the /project route. Since we switched to using the /project/new route to initialize each project for its chosen model package, the original handler was no longer executing, and the button wasn't enabled.

By including that instruction in the new handler, we ensure that the button is made visible. In addition, we push down the map controls a bit so they don't overlap the button.

## Testing Instructions

Checkout branch and bundle scripts. Make one each of GWLF-E and TR-55 projects, and ensure that the sidebar collapse button is visible. Collapse the sidebar. Ensure that the map controls don't obscure or be obscured by the collapse button:

image

Connects #1345 
Connects #1346

### DIFF
--- a/src/mmw/js/src/modeling/controllers.js
+++ b/src/mmw/js/src/modeling/controllers.js
@@ -119,6 +119,7 @@ var ModelingController = {
             finishProjectSetup(project, lock);
             updateUrl();
         }
+        App.rootView.showCollapsable();
     },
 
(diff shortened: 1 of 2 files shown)


lewfish:
+1, tested

Issue #1351, opened 2016-06-01
Convert weather data to metic


mmcfarland:
The weather data (`Prcp` and `TMin/TMax`) were provided in degrees F and inches, but the gwlf-e calculations expect C and cm.  The results need to be converted before being written to the GMS file.

https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/mapshed/calcs.py#L328-L357

The implications are that the model runs hot and dry, leading to high levels of ET and virtually no runoff which in turns reduces the sediment and nutrient loading output.  For a HUC12 in SE Pa, we should expect to see annual precipitation totals of ~100, not ~40 as we are seeing presently.

The numbers generally looked off to BME, but they could all be related to this conversion error.  We'll investigate if the numbers don't realign after this change.  It's worth nothing that the desktop version produces numbers very close to the web app, even though the input data appear off in this case.

PR #1352, opened 2016-06-02
Convert Temperature to Celsius, Precipitation to cm

rajadain:
## Overview

The source data is in Fahrenheit and inches, however GWLF-E needs these values in Celsius and cm.

## Testing Instructions

Create a MapShed project and export a GMS file. Ensure that the values are in Celsius and cm from line 148 onwards.

Connects #1351

### DIFF
--- a/src/mmw/apps/modeling/mapshed/calcs.py
+++ b/src/mmw/apps/modeling/mapshed/calcs.py
@@ -22,6 +22,7 @@ POULTRY = settings.GWLFE_CONFIG['Poultry']
 LITERS_PER_MGAL = 3785412
 AG_NLCD_CODES = settings.GWLFE_CONFIG['AgriculturalNLCDCodes']
 KM_PER_M = 0.001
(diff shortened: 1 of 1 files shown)


mmcfarland:
+1, GMS file appears to have values in metric range and the output precipitation looks more reasonable (including more sensible values for ET).

Annual precipitation in my test watershed was 103.77 which is right in line

screenshot from 2016-06-02 09 24 37

kellyi:
Looks good to me. 

I don't know if this matters, but it looks like the GMS file I exported has CRLF line endings.

rajadain:
@kellyi That's interesting, wonder if @kdeloach put that in on purpose. In the end, it must be opened in Windows, so that's actually a good thing.

kdeloach:
We're using csv writer which defaults to `\r\n` line endings.

Ref: https://docs.python.org/2/library/csv.html#csv.Dialect.lineterminator

Issue #1353, opened 2016-06-02
Use AvGroundwater for Subsurface Flow


mmcfarland:
We're incorrectly using `TileDrain` and reporting it as Subsurface Flow.  Instead, the key should be `AvGroundWater`

https://github.com/WikiWatershed/model-my-watershed/blob/184668955c9d60313480b7e84521952ad7c26309/src/mmw/js/src/modeling/gwlfe/runoff/views.js#L16

PR #1354, opened 2016-06-02
Use AvGroundWater for Subsurface flow display

mmcfarland:
Replaces an incorrect key for display and updates the output tab from Runoff to Hydrology.

Connects #1353 

#### Testing
Run a Multi Year model on this branch and observe that Subsurface Flow results on the Hydrology tab are non-0 and that the ET, R and SsF results _roughly_ add up to Precip (they never will exactly, as ET can pull from water stored in the water table).

screenshot from 2016-06-02 10 21 34


mmcfarland:
Thanks!

Issue #1355, opened 2016-06-02
Convert Point Source Flow from MGD to cm of water depth in GMS


mmcfarland:
> the values written to lines 73-84 of the gms file are actually supposed to be in equivalent cm of water depth over the watershed, and not in "Million Gallons per Day" as noted in our "MapShed GMS File Key".

Guidance for converting MGD (million gals/day) of point source discharge to cm of water depth

1. Convert MGD to MG per month 
     * (MGD * 30 = MGM)
2. Convert MGM to cubic meters/month
    * MGM * 3785 = cubic meters / month
3. Convert  cubic meters/month to equivalent meters/month of water depth
     * Cubic meters per month / area of watershed in square meters = meters of water depth
     * note: 1 ha = 10,000 m^2 = 0.01 km
4. Convert meters/month of water depth to cm of water depth
     * Meters/month * 100 = cm/month

mmcfarland:
Relavent existing VB code:

```vb
' Calculate flow in cm using the formula: PointFlow(i) = PointFlowMGD(i) * (((3785 * DaysPerMonth) / TotArea) * 100)
    For i = 1 To 12
        If (i = 2) Then
            DaysPerMonth = 28
        ElseIf (i = 4) Or (i = 6) Or (i = 9) Or (i = 11) Then
            DaysPerMonth = 30
        Else
            DaysPerMonth = 31
        End If
        
        PointFlowMGD(i) = Val(PtFlow(i - 1).Text)
                
        PointFlow(i) = PointFlowMGD(i) * ((3785.4118 * DaysPerMonth) / (AreaTotal * 10000)) * 100
```


In this case, `AreaTotal` is in hectares.

PR #1356, opened 2016-06-06
Discharge in cm/month, instead of L/m²/month

rajadain:
## Overview

Instead of reporting point source discharge as Liters per Sq M per month, we switch to Centimeters per month.

## Testing Instructions

Inspect the "Point Source Flow" column in the output. For a given shape, the value from this branch should be a factor of ten less than a corresponding value from `develop`.

image

Connects #1355

### DIFF
--- a/src/mmw/apps/modeling/mapshed/calcs.py
+++ b/src/mmw/apps/modeling/mapshed/calcs.py
@@ -19,10 +19,11 @@ MONTHS = settings.GWLFE_DEFAULTS['Month']
 MONTHDAYS = settings.GWLFE_CONFIG['MonthDays']
 LIVESTOCK = settings.GWLFE_CONFIG['Livestock']
 POULTRY = settings.GWLFE_CONFIG['Poultry']
-LITERS_PER_MGAL = 3785412
+M3_PER_MGAL = 3785.41178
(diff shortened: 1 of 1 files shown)


kellyi:
Is develop staged anywhere so I can compare these side by side?

rajadain:
Check https://staging.app.wikiwatershed.org/ for latest `develop`

kellyi:
Thanks!

kellyi:
This isn't exactly the same polygon, but it's close...

Here's what I see on staging:

screen shot 2016-06-06 at 4 23 04 pm

... and here's the output from this PR:

screen shot 2016-06-06 at 4 23 51 pm

Do the new values look like they're in tune with the change?

rajadain:
They do. You can also choose a predefined shape (such as a Congressional District or HUC-12) to get an exact match.

rajadain:
Ah but for those you need to have those shapes available locally, which you may not. I think this is close enough

kellyi:
Ah, ok. I don't think I have any of them set up on my machine. I tried a sq km but it output 0.0 for each value.

kellyi:
+1 for this since they look accurate.

rajadain:
I'm holding off on this pending more info from @mmcfarland and BME

rajadain:
Got a +1 from BME

> OK, the pt src values are now being calculated correctly.

Merging this shortly.

PR #1358, opened 2016-06-08
GWLF-E Should Not Fail on Empty Streams

rajadain:
## Overview

Though the results may be inaccurate, we can add a disclaimer at a later time to notify the user. Crashing is not acceptable.

## Testing Instructions

Checkout the branch and reprovision `worker`. 

    vagrant provision worker

or

    vagrant up worker --provision

Accessing a 1 sq km stamp which does not have streams should no longer fail:

image

Connects https://github.com/WikiWatershed/gwlf-e/issues/61


lewfish:
+1, tested

Issue #1357, opened 2016-06-06
MapShed implementation issues


mmcfarland:
After a review by BME, the following issues in our MapShed implementation need to be resolved:

* [x] NLCD code 21 needs to be mapped to MapShed code 2 (LD Mixed)
* [x] Rural Land use `C` and `P` values are 0.0, should be assigned default values.  These should be in a MapShed routine already.
* [x] Confirm:  We are using Hi-res streams for LS, Low-Res streams for Stream Bank Erosion
* [x] Day Hours are off by one month (our March value shoudl be Feb, etc - see images)
* [x] Stream Bank AEU numbers are wrong order of magnitude.  We should evaluate how we are using County animal data and confirm with BME about the correct formula.

Orig Daylights
![screenshot 2016-06-06 at 2 54 05 pm](https://cloud.githubusercontent.com/assets/1014341/15833897/b6fd4e8e-2bf6-11e6-8aa0-20656d6aba66.png)

Our Daylights
![screenshot 2016-06-06 at 2 53 55 pm](https://cloud.githubusercontent.com/assets/1014341/15833896/b6f9c9c6-2bf6-11e6-8044-17260be5b90d.png)

rajadain:
I have been unable to locate where "Stream Bank AEU" numbers are. If there are no suggestions, I will clarify with BME on a call tomorrow.

PR #1359, opened 2016-06-09
Fix Various MapShed Issues

rajadain:
## Overview

Fixes the following issues:

 * Maps NLCD 21 to Low Density Mixed instead of Open Land
 * Looks up C and P values which were previously missing
 * Corrects day length calculation caused by an index mismatch
 * Corrects LS calculation which was previously mixing NHD stream percentages with DRB stream lengths
 * Corrects KV calculation which was previously using erosion coefficients instead of evapotranspiration coefficients

## Testing Instructions

Checkout the branch. Debug Celery

    ./scripts/debugcelery.sh

Select the **LIttle Neshaminy Creek** HUC12 right above Philadelphia as an area of interest.

image

Create a MapShed project and export "Current Conditions" as a GMS file.

Move the GMS file to a Windows VM. Open the GMS file using GWLF-E 1.4.0 which can be found at smb://fileshare/projects/Stroud_ModelMyWatershed/downloads/mapshed/GWLF-E-1.4.0.exe.

Compare the results of Transport Data and Nutrient Data with that of [`input_365.gms`](https://github.com/WikiWatershed/model-my-watershed/files/308010/input_365.gms.txt). The values will not be exact, since the data sources are different, but they should be within each others ballpark.

Connects #1357

### DIFF
--- a/src/mmw/apps/modeling/mapshed/calcs.py
+++ b/src/mmw/apps/modeling/mapshed/calcs.py
@@ -423,7 +423,7 @@ def curve_number(n_count, ng_count):
         cni_avg([90, 95]),  # Wetland
         0,  # Disturbed
         0,  # Turf Grass
-        cni_avg([21, 71]),  # Open Land
+        cni(71),  # Open Land
         cni_avg([12, 31]),  # Bare Rock
         0,  # Sandy Areas
         0,  # Unpaved Road
@@ -525,11 +525,11 @@ def landuse_pcts(n_count):
         n_pct.get(90, 0) + n_pct.get(95, 0),  # Wetland
         0,  # Disturbed
         0,  # Turf Grass
-        n_pct.get(21, 0) + n_pct.get(71, 0),  # Open Land
+        n_pct.get(71, 0),  # Open Land
         n_pct.get(12, 0) + n_pct.get(31, 0),  # Bare Rock
         0,  # Sandy Areas
         0,  # Unpaved Road
-        n_pct.get(22, 0),  # Low Density Mixed
+        n_pct.get(21, 0) + n_pct.get(22, 0),  # Low Density Mixed
         n_pct.get(23, 0),  # Medium Density Mixed
         n_pct.get(24, 0),  # High Density Mixed
         0,  # Low Density Residential
--- a/src/mmw/apps/modeling/mapshed/calcs.py
+++ b/src/mmw/apps/modeling/mapshed/calcs.py
@@ -616,3 +616,31 @@ def p_factors(avg_slope):
         1.0,   # Unpaved
         0, 0, 0, 0, 0, 0  # Urban Land Use Types
     ]
+
+
+def phosphorus_conc(sed_phos):
+    """
+    Given the average concentration of phosphorus dissolved in sediments for
+    the entire polygon, returns average concentration per land use type based
+    on pre-baked estimates.
+
+    Original at Class1.vb@1.3.0:8975-9001,9350-9359
+    """
+    psed = sed_phos / 1.6
+    stp = 190 * psed / 836
+    prunoff = (1.98 * stp + 79) / 1000
+    prunoff_turf = 2.9 * psed / 836
+
+    return [
+        prunoff,       # Hay/Pasture
+        prunoff,       # Cropland
+        0.01,          # Forest
+        0.01,          # Wetland
+        0.01,          # Disturbed
+        prunoff_turf,  # Turf Grass
+        0.01,          # Open Land
+        0.01,          # Bare Rock
+        0.01,          # Sandy Areas
+        0.01,          # Unpaved
+        0, 0, 0, 0, 0, 0  # Urban Land Use Types
+    ]
--- a/src/mmw/apps/modeling/mapshed/calcs.py
+++ b/src/mmw/apps/modeling/mapshed/calcs.py
@@ -117,21 +117,28 @@ def et_adjustment(ws):
     return [avg_etadj] * 12
 
 
-def kv_coefficient(ecs):
+def kv_coefficient(area_pcts, season):
     """
-    Given an array of erosion coefficients, returns an array of 12 decimals,
-    one for the KV coefficient of each month. The KV of a month is initialized
-    to the corresponding erosion coefficient value times the KV Factor, and
-    then averaged over the preceeding month. January being the first month is
-    not averaged.
+    Given arrays of land use area percentages and growing season, returns an
+    array of 12 floats, one for the KV coefficient of each month. The KV of a
+    month is initialized to the sum of ETs of each land use type weighted by
+    its area percent, using growth or dormant coefficients depending on whether
+    or not the month is in the growing season. The value is then averaged over
+    the preceeding month and multiplied by the KV Factor. January being the
+    first month is not averaged.
 
     Original at Class1.vb@1.3.0:4989-4995
     """
 
-    kv = [ec * KV_FACTOR for ec in ecs]
+    et_grow = sum([et * area_pct for et, area_pct in
+                  zip(settings.GWLFE_CONFIG['ETGrowCoeff'], area_pcts)])
+    et_dorm = sum([et * area_pct for et, area_pct in
+                  zip(settings.GWLFE_CONFIG['ETDormCoeff'], area_pcts)])
 
+    kv = [et_grow if m == GrowFlag.GROWING_SEASON else et_dorm for m in season]
+    kv[0] *= KV_FACTOR
     for m in range(1, 12):
-        kv[m] = (kv[m] + kv[m-1]) / 2
+        kv[m] = KV_FACTOR * (kv[m] + kv[m-1]) / 2
 
     return kv
 
--- a/src/mmw/apps/modeling/mapshed/calcs.py
+++ b/src/mmw/apps/modeling/mapshed/calcs.py
@@ -575,3 +575,35 @@ def sed_a_factor(landuse_pct_vals, cn, AEU, AvKF, AvSlope):
     return ((0.00467 * urban_pct) + (0.000863 * AEU) +
             (0.000001 * avg_cn) + (0.000425 * AvKF) +
             (0.000001 * AvSlope) - 0.000036)
+
+
+def p_factors(avg_slope):
+    """
+    Given the average slope, calculates the P Factor for rural land use types.
+
+    Original at Class1.vb@1.3.0:4393-4470
+    """
+    if 0 <= avg_slope < 2.1:
+        ag_p = 0.52
+    elif 2.1 <= avg_slope < 7.1:
+        ag_p = 0.45
+    elif 7.1 <= avg_slope < 12.1:
+        ag_p = 0.52
+    elif 12.1 <= avg_slope < 18.1:
+        ag_p = 0.66
+    else:
+        ag_p = 0.74
+
+    return [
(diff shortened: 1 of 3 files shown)


rajadain:
Here are my comparisons:

Little Neshaminy Transport Data
little_neshaminy_kvfix_forreal_td

BME Transport Data
input_365_td

Little Neshaminy Nutrient Data
little_neshaminy_kvfix_forreal_nd

BME Nutrient Data
input_365_nd

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -252,6 +256,29 @@ def nlcd_streams(sjs_result):
...
+    This callback is run when the geometry falls within the DRB. We calculate
+    the percentage of DRB streams in each land use type.
+    """
+    if 'error' in sjs_result:
+        raise Exception('[nlcd_streams_drb] {}'.format(sjs_result['error']))
+
+    result = parse_sjs_result(sjs_result)
+    total = sum(result.values())
+
+    lu_stream_pct_drb = [0.0] * 16
```
caseypt: Consider adding a comment to describe the significance of `16`.

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -252,6 +256,29 @@ def nlcd_streams(sjs_result):
...
+    for nlcd, stream_count in result.iteritems():
+        lu = get_lu_index(nlcd)
+        if lu is not None:
+            lu_stream_pct_drb[lu] += float(stream_count) / total
```
caseypt: Should we protect against a divide by zero error here?

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -437,6 +464,12 @@ def geop_tasks(geom, errback):
          nlcd_kfactor)
     ]
 
+    if geom.within(DRB):
+        definitions.append(('nlcd_streams',
+                            {'polygon': [geom.geojson],
+                             'vector': streams(geom, drb=True)},
+                            nlcd_streams_drb))
```
caseypt: If an AoI is within the DRB, it looks like the `nlcd_streams` task will be run twice; once in the beginning, and once at the end.

src/mmw/apps/modeling/mapshed/tasks.py
rajadain: Good catch!

rajadain: Although we probably don't need to, because if `total` has a value it is because some item in result has a value. Thus, when `total` is 0, we simply won't enter the `for nlcd, stream_count in result.iteritems()` loop.

src/mmw/apps/modeling/mapshed/tasks.py
rajadain: It's just the total number of land use types, but I will add a comment / variable to make that more clear.

src/mmw/apps/modeling/mapshed/tasks.py
caseypt: Is it possible for `result` to look like `{ nlcd_value: 0, nlcd_value: 0, nlcd_value: 0, etc.. }`?

src/mmw/apps/modeling/mapshed/calcs.py
```diff
@@ -575,3 +575,35 @@ def sed_a_factor(landuse_pct_vals, cn, AEU, AvKF, AvSlope):
...
+    return [
+        ag_p,  # Hay/Pasture
+        ag_p,  # Cropland
+        ag_p,  # Forest
+        0.1,   # Wetland
+        0.1,   # Disturbed
+        0.8,   # Turf Grass
+        ag_p,  # Open Land
+        ag_p,  # Bare Rock
+        0.0,   # Sandy Areas
```
lewfish: In `Class1.vb`, it looks like `PFactor(9)` is equal to `P14` which is equal to `Pf1-Pf5`, so Sandy Areas should also be `ag_p`.

src/mmw/mmw/settings/gwlfe_settings.py
```diff
@@ -463,6 +463,42 @@
     'SSLDR': 4.4,
     'SSLDM': 2.2,
     'WeatherNull': -99999,  # This value is used to indicate NULL in ms_weather dataset
+    'ETGrowCoeff': [
```
caseypt: Can you point me to the source of these so I can double check them?

src/mmw/apps/modeling/mapshed/tasks.py
rajadain: That's correct. While we prefer to use the DRB streams for calculating LS if they are available, we _always_ use the NHD streams for Stream Bank Erosion, which are calculated here: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/mapshed/tasks.py#L137-L142. Thus, we need the NHD streams in all cases, and the DRB streams only in the cases when they are available.

src/mmw/mmw/settings/gwlfe_settings.py
rajadain: Class1.vb@1.3.0:4393-4470

src/mmw/apps/modeling/mapshed/tasks.py
caseypt: Got it. Is it okay that the `opname`s are the same?

rajadain: That's a bit confusing, probably warrants a comment, but yes. That actually points to the JSON block that is sent to the geoprocessing service, which is defined here: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/base.py#L386-L399

Both requests use the same JSON block, just different values in the `vector` field.

src/mmw/mmw/settings/gwlfe_settings.py
rajadain: Sorry, that's for P Factor.

For ET take a look at Class1.vb:4989-4995 `TransMonthData`, and see where it is called from (sorry don't have the file with me right now)

caseypt: Looks good.

src/mmw/apps/modeling/mapshed/calcs.py
```diff
@@ -575,3 +575,35 @@ def sed_a_factor(landuse_pct_vals, cn, AEU, AvKF, AvSlope):
...
+    else:
+        ag_p = 0.74
+
+    return [
+        ag_p,  # Hay/Pasture
+        ag_p,  # Cropland
+        ag_p,  # Forest
+        0.1,   # Wetland
+        0.1,   # Disturbed
+        0.8,   # Turf Grass
```
lewfish: In `Class1.vb`, `PFactor(6)` = `P16` which can only ever be 0 or 0.2

src/mmw/apps/modeling/mapshed/calcs.py
```diff
@@ -575,3 +575,35 @@ def sed_a_factor(landuse_pct_vals, cn, AEU, AvKF, AvSlope):
...
+        ag_p,  # Open Land
+        ag_p,  # Bare Rock
+        0.0,   # Sandy Areas
+        1.0,   # Unpaved
```
lewfish: `PFactor(10)` is 0 or 1 depending on the value of `Lu`. Why do we ignore `Lu` here?

src/mmw/mmw/settings/gwlfe_settings.py
caseypt: Yeah I ended up finding them. Thanks.

src/mmw/apps/modeling/mapshed/tasks.py
rajadain: No, the geoprocessing service does not have any idea of what the "full set" of NLCD values is like. In its response it only reports the cell values it finds. Thus, if a certain polygon has only 3 types of land uses, the output will also only have 3.

src/mmw/mmw/settings/gwlfe_settings.py
```diff
@@ -40,6 +40,8 @@
                 ],
     'Imper': [0, 0, 0, 0, 0, 0, 0, 0, 0, 0,         # Impervious surface area percentage
               0.15, 0.52, 0.87, 0.15, 0.52, 0.87],  # only defined for urban land use types
+    'C': [0.03, 0.42, 0.002, 0.01, 0.1, 0.03, 0.04, 0.001, 0.01, 0.08,  # C Factor Defaults
```
lewfish: I don't see where C is calculated in the original code.

src/mmw/apps/modeling/mapshed/calcs.py
```diff
@@ -31,18 +31,20 @@ def day_lengths(geom):
...
-                                             math.tan(latitude * 0.017453) *
-                                             math.cos(0.0172 *
-                                                      (month * 30.4375 - 5)))
+        lengths[m] = 7.63942 * math.acos(0.43481 *
+                                         math.tan(0.017453 * latitude) *
+                                         math.cos(0.0172 *
+                                                  ((m + 1) * 30.4375 - 5)))
 
-    return lengths
+    return [round(l, 1) for l in lengths]
```
lewfish: Looks good.

src/mmw/apps/modeling/mapshed/tasks.py
caseypt: Got it.

caseypt:
Last two commits look good.

src/mmw/mmw/settings/gwlfe_settings.py
lewfish: If these numbers are supposed to match C1, C2, .. they appear to be off.

mmcfarland:
Corrected reference GMS from BME, with screenshots:

[input_253.gms.txt](https://github.com/WikiWatershed/model-my-watershed/files/309285/input_253.gms.txt)

screenshot from 2016-06-10 11 44 07
screenshot from 2016-06-10 11 43 40

mmcfarland:
+1 Screenshots show pretty good tracking of the desktop GIS numbers.  I think there are some outstanding dissolved P discrepancies that won't likely get resolved immediately because they'll require additional datasets that we are holding off implementing.

src/mmw/mmw/settings/gwlfe_settings.py
rajadain: They map lines 4393-4436 by land use type

lewfish: I think the 5th value (disturbed) and 10th (unpaved) might be off.  If the comment below is accurate, then disturbed should should be 0.08 and unpaved should be 0.8.
```
        If Lu = 12 Or Lu = 13 Or Lu = 11 Or Lu = 15 Then ' Disturbed, Bare Rock, Unpaved Roads, & Open Land
            C12 = 0.08 : C13 = 0.1 : C11 = 0.8 : C15 = 0.08
```

rajadain: Thanks, I think I used a word document that BME sent. This is more correct, though.

lewfish:
Looks good otherwise.

rajadain:
Thanks all. I've addressed all the comments and added a final fix that corrects dissolved phosphorus concentration values:

untitled_project__current_conditions_nd

Issue #1360, opened 2016-06-10
Update County Animal Dataset


mmcfarland:
New dataset has been provided with additional attributes.  Update the sql tar hosted on s3.

mmcfarland:
Loaded new shapefile into PostGIS and exported to sql, updated the s3 file.

#### Test
Load the updated data via setup script from your app VM 
-  `./setupdb.sh -f ms_county_animals.sql.gz`
- Verify the new attributes are present:

 ```
totalaeu     | double precision
grazingaeu   | double precision
totaeu_ha    | numeric(7,2)
grzaeu_ha    | numeric(7,2)
```

- The new table should have the same number of records as the old (4409)
 ```
mmw=# select count(*) from ms_county_animals;
 count 
-------
  4409
(1 row)
```

##### Steps to update
I used these two commands to produce the new sql file

```
ogr2ogr -t_srs EPSG:4326 -f "PostgreSQL" PG:"host=localhost user=mmw dbname=mmw password=mmw" -skipfailures ~/shapes/cty_animals.shp -nln ms_county_animals
```
```
pg_dump -c --if-exists -O -t ms_county_animals -h localhost -U mmw mmw > ms_county_animals.sql
```

mmcfarland:
This will have to be run on staging/production when the associated commits are released.

rajadain:
Thanks! Confirmed that I have the new columns. I'll make a PR to use these new values in the AEU calculation.

rajadain:
Interestingly, the `geom` column has been renamed to `wkb_geometry` in the new dataset

mmcfarland:
Shoot.  I'll update the column & indexes....

mmcfarland:
s3 file is updated to `geom` and `ms_county_animals_geom_idx" gist (geom)`.

rajadain:
Ran the script on Staging.

PR #1361, opened 2016-06-10
Fix AEU Calculation

rajadain:
## Overview

Previously we were not multiplying the animal populations by weight and then dividing by 1000, as done in the source:

```vb
' Obtain AEU for each farm animal
AEU1 = ((Chicks / 2) * (BroilerWt) / 1000) + ((Chicks / 2) * (LayerWt) / 1000) ' Chickens
AEU2 = (Turkeys * TurkeyWt) / 1000 ' Turkeys
AEU3 = (Sheep * SheepWt) / 1000 ' Sheep
AEU4 = (Hogs * HogWt) / 1000 ' Hogs
AEU5 = (Horses * HorseWt) / 1000 ' Horses
AEU6 = (DairyCows * DairyWt) / 1000 ' Dairy Cows
AEU7 = (BeefCows * BeefWt) / 1000 ' Beef Cows

' Get the total AEU, Total livestock and poultry AEU
TotAEU = AEU1 + AEU2 + AEU3 + AEU4 + AEU5 + AEU6 + AEU7
TotLAEU = AEU3 + AEU4 + AEU5 + AEU6 + AEU7
TotPAEU = AEU1 + AEU2

' If the total livestock AEU > 0 then recalculate the AEU based on the basin area
If TotLAEU > 0 Then
    AEU = TotLAEU / (BasinAreaHa * 2.471)
End If
```

This adds that technique. ~~In addition, we also have a [new dataset from BME](https://github.com/WikiWatershed/model-my-watershed/issues/1360) which includes a pre-calculated `grzaeu` column which can be area-weighted and used directly.~~

~~I am in conversation with BME on whether to use the new kg weighted value or the `grzaeu` column. Will hold off on merging this until I have an answer.~~

BME says to use the technique which correctly mimics the VB code, rather than the `grzaeu` column.

## Testing Instructions

Compile and run MMW. Create a MapShed project, and export a GMS file from Current Conditions. The value of AEU in line 2 column 15 should be of the order of 0.05559663102442437 for a watershed of ~ 100 sq km.

Connects #1357 
Connects #1360

### DIFF
--- a/src/mmw/apps/modeling/mapshed/calcs.py
+++ b/src/mmw/apps/modeling/mapshed/calcs.py
@@ -17,6 +17,7 @@ NUM_WEATHER_STATIONS = settings.GWLFE_CONFIG['NumWeatherStations']
 KV_FACTOR = settings.GWLFE_CONFIG['KvFactor']
 MONTHS = settings.GWLFE_DEFAULTS['Month']
 MONTHDAYS = settings.GWLFE_CONFIG['MonthDays']
+WEIGHTOF = settings.GWLFE_CONFIG['AvgAnimalWt']
 LIVESTOCK = settings.GWLFE_CONFIG['Livestock']
 POULTRY = settings.GWLFE_CONFIG['Poultry']
 M3_PER_MGAL = 3785.41178
@@ -146,6 +147,8 @@ def kv_coefficient(area_pcts, season):
 def animal_energy_units(geom):
     """
     Given a geometry, returns the total livestock and poultry AEUs within it
+
+    Original at Class1.vb@1.3.0:9230-9247
     """
     sql = '''
           WITH clipped_counties AS (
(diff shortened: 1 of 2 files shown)


rajadain:
Will remove the `grzaeu` commit before merging.

src/mmw/apps/modeling/mapshed/calcs.py
```diff
@@ -182,8 +185,10 @@ def animal_energy_units(geom):
         values = cursor.fetchone()  # Only one row since aggregate query
         aeu = dict(zip(columns, values))
```
lewfish: Consider changing `aeu` here to `population` since it's not an AEU until you multiply by weight and divide by 1000.

lewfish:
+1, tested

src/mmw/apps/modeling/mapshed/calcs.py
rajadain: Good call.

rajadain:
Removed extra commits, updated variable naming. 6415d33

PR #1362, opened 2016-06-10
Various UI Fixes

rajadain:
## Overview

Addresses various UI fixes listed in https://github.com/azavea/civic-apps/issues/116

## Testing Instructions

Checkout the branch and rebundle the scripts. Ensure that:

 1. Column headers in the Hydrology table with GWLF-E results have units
 2. The Project menu is always on top of everything else
 3. Renaming a project and exporting a GMS file results in the correctly named file being downloaded

Connects https://github.com/azavea/civic-apps/issues/116

### DIFF
--- a/src/mmw/js/src/modeling/gwlfe/runoff/templates/table.html
+++ b/src/mmw/js/src/modeling/gwlfe/runoff/templates/table.html
@@ -7,7 +7,7 @@
                         {{ columnName }}
                     </th>
                 {% else %}
-                    <th data-sortable="true" data-sorter="window.numericSort">{{ columnName }}</th>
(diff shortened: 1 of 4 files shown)


lewfish:
The z-index for this menu also needs to be updated:

screen shot 2016-06-13 at 9 26 39 am

lewfish:
^ Actually, I'm not sure what menu should take precedence in this case.

lewfish:
+1, tested

rajadain:
Good catch. I think the scenario selection pop-up menu should be on top, because it isn't sticky like the modification selection one.

image

PR #1367, opened 2016-06-13
Add Simple Error Messages

rajadain:
## Overview

Adds two error messages. If you draw an area of interest that the web app cancels, it no longer cancels it silently but instead tells you why:

image

If you select an area of interest that is too big for MapShed, you will see an error there as well:

image

## Testing Instructions

 1. Draw an area of interest around a large state like Pennsylvania. You should see an alert message before your drawing is canceled.
 2. Select a HUC-8 and proceed to MapShed. The "Gathering Data" step should fail with a timeout message.

Connects #1330

### DIFF
--- a/src/mmw/js/src/core/models.js
+++ b/src/mmw/js/src/core/models.js
@@ -183,7 +183,7 @@ var TaskModel = Backbone.Model.extend({
         // pollInterval has elapsed.
         var getResults = function() {
             if (elapsed >= self.get('timeout')) {
-                defer.reject();
+                defer.reject({timeout: true});
                 return;
             }
 
@@ -219,6 +219,14 @@ var TaskMessageViewModel = Backbone.Model.extend({
 
     setError: function() {
         this.set('message', 'Error');
+        this.set('iconClasses', 'fa fa-exclamation-triangle');
+    },
+
+    setTimeoutError: function() {
+        var message = 'Operation took too long <br />' +
+                      'Consider trying a smaller area of interest.';
+
+        this.set('message', message);
         this.set('iconClasses', 'fa fa-exclamation-triangle');
     },
 
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -62,11 +62,10 @@ function validateShape(polygon) {
         d = new $.Deferred();
 
     if (area > MAX_AREA) {
-        var message = '';
-        message += 'Sorry, your Area of Interest is too large.\n\n';
-        message += Math.floor(area) + ' square km were selected, ';
-        message += 'but the maximum supported size is currently ';
-        message += MAX_AREA + ' square km.';
+        var message = 'Sorry, your Area of Interest is too large.\n\n' +
+                      Math.floor(area).toLocaleString() + ' km² were selected, ' +
+                      'but the maximum supported size is currently ' +
+                      MAX_AREA.toLocaleString() + ' km².';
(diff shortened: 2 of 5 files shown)


src/mmw/js/src/draw/views.js
```diff
@@ -67,6 +67,7 @@ function validateShape(polygon) {
         message += Math.floor(area) + ' square km were selected, ';
         message += 'but the maximum supported size is currently ';
         message += MAX_AREA + ' square km.';
+        window.alert(message);
```
caseypt: Interesting. We were constructing the message, but not showing it? I thought I remember seeing it at one point.

src/mmw/js/src/core/models.js
```diff
@@ -222,6 +222,14 @@ var TaskMessageViewModel = Backbone.Model.extend({
         this.set('iconClasses', 'fa fa-exclamation-triangle');
     },
 
+    setTimeout: function() {
```
caseypt: Consider renaming this to `setTimeoutError` or something similar to indicate it's an error state and that we aren't setting some timeout number/amount.

caseypt:
If we know that a very large AoI causes a timeout for MapShed, can we lower the AoI maximum size?

rajadain:
> If we know that a very large AoI causes a timeout for MapShed, can we lower the AoI maximum size?

Unfortunately at the stage when the user is selecting the AoI, we don't know if they will proceed to MapShed or TR-55. We may make a policy decision to set the bound to the lower of the two, but I don't know how this could affect existing use cases for TR-55 analysis. I also remember some discussion about how MapShed was more accurate for large AoIs than TR-55, so perhaps setting a lower bound to MapShed would lead to more accurate results.

We could also make it so that MapShed is unavailable as an option if the AoI is bigger than some limit (which we'll have to deduce by trial and error, although I suspect it'll be around the 1,000 km² range).

However, losing the ability to work with large AoIs (most HUC-8s would be excluded) may be seen as a loss of functionality, even though it was only apparent and not accurate.

At some point it was mentioned that we want to do the least possible at this stage to get the release out, and revisit this when we have more funding available. Tagging @ajrobbins and @mmcfarland for their input.

src/mmw/js/src/core/models.js
rajadain: Good catch.

src/mmw/js/src/draw/views.js
rajadain: I do too. Let me do a git history inspection to see if I can pinpoint when we lost it

rajadain: I think it was accidentally lost here https://github.com/WikiWatershed/model-my-watershed/commit/210891039ea37cbd16c684660d62d9a9f7a94890#diff-72c102beb3924e611bfc083303b1f8d9L69

rajadain:
Also added a commit to make the error message friendlier:

image

rajadain:
So do I have a +1?

caseypt:
+1. Code looks good. I was only able to test the first error message when drawing a large AoI.

kellyi:
Drew a polygon around Connecticut and saw the error:

screen shot 2016-06-14 at 3 10 12 pm

kellyi:
Tried to model it with MapShed and saw the error again:

screen shot 2016-06-14 at 3 12 20 pm

kellyi:
Looks good to me. +1

PR #1374, opened 2016-06-15
Use datahub CDN endpoint for NLCD and SSURGO tiles

hectcastro:
Instead of serving tiles directly from Amazon S3, use the new CDN enabled datahub endpoint.

See also: https://github.com/azavea/azavea-data-hub/pull/37

---

**Testing**

Ensure that the application makes use of the new tile endpoint, and inspect that the tiles render within the application correctly. I am in the middle of copying all of the tiles over to the origin of this CDN, so not all of the tiles may be available right away (at zoom level 13 for NLCD as I type this).

### DIFF
--- a/src/mmw/mmw/settings/layer_settings.py
+++ b/src/mmw/mmw/settings/layer_settings.py
@@ -143,8 +143,7 @@ LAYERS = [
         'short_display': 'NLCD',
         'helptext': 'National Land Cover Database defines'
                     'land use across the U.S.',
-        'url': 'https://s3.amazonaws.com/com.azavea.datahub.tms/'
-               'nlcd/{z}/{x}/{y}.png',
+        'url': 'https://{s}.tiles.azavea.com/nlcd/{z}/{x}/{y}.png',
         'raster': True,
         'overlay': True,
         'maxNativeZoom': 13,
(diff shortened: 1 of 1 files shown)


rajadain:
I'm seeing NLCD from zoom 10 and deeper. Soils are 404ing at all levels right now. But they're all being requested from the right endpoint. Sweet!

rajadain:
The NLCD that does render does so correctly and matches the basemap.

rajadain:
Will take another look once Soil layers have been copied.

hectcastro:
This is accurate. I'm at level 13 for NLCD and SSURGO hasn't started yet.

hectcastro:
Both of those layers should be available now.

rajadain:
Taking another look

rajadain:
+1 tested, compared with Staging. All tiles loaded correctly, even felt faster.

caseypt:
I ran the iCRIS extract unit tests after I shutdown LR09 and installed the new version of the ULRS which is pointed at LR13. 

All the tests passed: http://jenkins01.internal.azavea.com/view/DOR/job/DOR_iCRIS_Extracts_Develop_Unit_Tests/290/console

rajadain:
+1

Issue #1041, opened 2015-12-01
Update GMaps API Key domains


mmcfarland:
`mmw-dev` & `mmw.azavea.com` are reporting:

```
Google Maps API warning: InvalidKey: 
https://developers.google.com/maps/documentation/javascript/error-messages
```

caseypt:
It looks as if we aren't setting the key at all. On staging and production, the script include looks like this:

![screen shot 2016-01-26 at 9 54 15 am](https://cloud.githubusercontent.com/assets/1042475/12583898/fcf4b68a-c412-11e5-8168-ecf1437853a8.png)

caseypt:
Could be a problem with this line: https://github.com/WikiWatershed/model-my-watershed/blob/6d8eee5fca771cc88762c2a86f861455b0d306e9/src/mmw/apps/core/templates/base.html#L51 or the settings file.

ajrobbins:
Google maps API is no longer supporting keyless access: http://googlegeodevelopers.blogspot.com/2016/06/building-for-scale-updates-to-google.html. We should address this before the next release https://github.com/WikiWatershed/model-my-watershed/issues/1376

hectcastro:
Some settings may need to be updated by those who have access to Google Maps API admin. Looks like we need to whitelist `staging.app.wikiwatershed.org` and `app.wikiwatershed.org` (MMW URLs too, or can we finally deprecate them?):

```
js:32 Google Maps API error: RefererNotAllowedMapError https://developers.google.com/maps/documentation/javascript/error-messages#referer-not-allowed-map-error
Your site URL to be authorized: https://staging.app.wikiwatershed.org/
```

PR #1378, opened 2016-06-24
Fix Google Maps Key

rajadain:
## Overview

Previously it was embedded inside a `json.dumps` call. Since it is never used by any JavaScript code, we extract it from client_settings and include it directly.

## Testing Instructions

Checkout the branch and open up [http://localhost:8000/](http://localhost:8000/). Open up developer tools. They should **NOT** look like this:

image

The `maps.googleapis.com` URL should have a real key, and the two Google Maps API warnings should not be present.

Connects #1041

### DIFF
--- a/src/mmw/apps/core/templates/base.html
+++ b/src/mmw/apps/core/templates/base.html
@@ -54,7 +54,7 @@
 
     {% block javascript %}
         <script type="text/javascript"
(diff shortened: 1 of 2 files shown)


rajadain:
Closing this because all new PRs should target the release branch instead.

PR #1382, opened 2016-06-28
Scale Celery across multiple workers correctly

rajadain:
## Overview

Geoprocessing tasks were spreading indiscriminately between workers, causing `mapshed_finish` to attempt to retrieve jobs from one worker that were started on another, thus failing.

By pinning `mapshed_start` and `mapshed_finish` to work on same worker, we ensure that the Spark jobs are always available.

This does create a small bottleneck since all geoprocessing tasks now occur on a single worker instead of spreading out, but that is left for later optimization if deemed necessary.

## Validating Instructions

The same code has been deployed manually to staging which has been upgraded to have multiple workers.

Once this is deployed, ensure that MapShed works on staging with multiple workers.

### DIFF
--- a/src/mmw/apps/modeling/mapshed/tasks.py
+++ b/src/mmw/apps/modeling/mapshed/tasks.py
@@ -445,7 +445,7 @@ def nlcd_kfactor(result):
     return output
 
 
-def geop_tasks(geom, errback):
+def geop_tasks(geom, errback, exchange, routing_key):
     # List of tuples of (opname, data, callback) for each geop task
     definitions = [
         ('nlcd_streams',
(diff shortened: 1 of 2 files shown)


hectcastro:
👍

mmcfarland:
> all geoprocessing tasks now occur on a single worker 

This means that a _single_ task will always occur on a single worker, but that new tasks are still distributed across the workers (to remain on one after it is scheduled), right?

rajadain:
It means that all geoprocessing tasks (`nlcd_streams`, `nlcd_soils`, `gwn`, `avg_awc`, `nlcd_slope`, `slope`, `nlcd_kfactor`, `nlcd_streams_drb`) for one hit of `/start/mapshed` will occur on a single worker. A second hit may happen on the same worker or another worker. Non-geoprocessing tasks (`combine`, `collect_data`, and `run_gwlfe`) will be spread over all workers. These are safe to spread because they are pure functions, and depend only on the value given to them when calling.

PR #1384, opened 2016-06-29
Scope to workers of same color

rajadain:
Ensure that they don't cross over stack boundaries when multiple stacks exist at once.

Also increase Celery Chord Unlock Max Retries. In some cases, having multiple workers exhausts this limit quicker than anticipated. The setting exists purely to protect against an infinite loop, so having a slightly larger limit will not affect performance.


hectcastro:
👍

Issue #1390, opened 2016-07-14
Add 2 new DEP layers to Layer Selector


mmcfarland:
The new layers from #1389 should be added to the application layer selector:
- PA Municipal Boundares: Boundary
- PA Urban Areas: Coverage.

Issue #1389, opened 2016-07-14
Setup windshaft to tile 2 new DEP layers


mmcfarland:
Once loaded from #1388, configure windshaft to produce tiles for these layers.
- Municipal Boundaries can share styling with other boundary types
- Urban areas should have a simple fill color and border

Issue #1388, opened 2016-07-14
Add data scripts for 2 new DEP reference layers


mmcfarland:
Two new shapefiles are on the project share drive:
- PA Municipal Boundaries
- PA Urban Area

Prepare sql tars of these with the standard geom/index/srs/naming conventions and add to the the `setupdb.sh` script

PR #1391, opened 2016-07-19
Add DEP Layers

mmcfarland:
#### Overview
To support a use case for municipalities in PA to conform to DEP reporting guidelines, we've added two new reference layers to the map.  Within PA, Municipalities and Urban Areas can be toggled on simultaneously to guide users in delineating a watershed with certain requirements within their municipality.  The ultimate workflow intends for the user to then generate a GMS file from the app which will be used in a special desktop version of GWLF-E with these specific reporting capabilities built in.  

Commits include:
- Updated data loading script to access sql files on s3
- Windshaft tile rendering
- Layers added to layer control
- Layer availability restriction to PA

Connects #1388 
Connects #1389 
Connects #1390 

#### Testing
- Prior to testing, ensure your app vm is recently provisioned and bundle your scripts `./scripts/bundle.sh --debug`
- On your app server, run `setupdb.sh -p` to load the new DEP layers from s3 into your local database
  - If you do not have aws-cli installed, ignore the error at the end of the script, which is trying to purge the cache for this layer
  - Verify that the two tables are in your database, with spatial indexes, in `psql`
    - `\d dep_municipalities`
    - `\d dep_urban_areas`
- Open app and pan to an area within PA, verify that two new layers are enabled DEP Muncipalities (boundary) and DEP Urban Areas (coverage)
- Zoom away from PA so it is not in your viewport, verify that those two layers are disabled.
- Zoom back to PA, and turn on both layers.  Verify that that they render and that the boundary is rendered above the coverage, despite the order in which they are added
- Verify that Urban Areas has an opacity slider

screenshot 2016-07-19 at 1 56 11 pm

Simple PA perimeter used (drawn from geojson.io)
screenshot 2016-07-19 at 2 08 07 pm


src/mmw/mmw/settings/data/pa_perimeter.json
```diff
@@ -0,0 +1 @@
+{"type":"Feature","geometry":{"type":"Polygon","coordinates":[[[-75.11352539062499,42.0125705565935],[-79.7607421875,42.01665183556825],[-79.78271484375,42.27730877423709],[-80.5352783203125,41.97174336327968],[-80.52978515625,39.690280594818034],[-75.43212890625,39.740986355883564],[-74.7509765625,40.16208338164619],[-75.04486083984375,40.84913799774759],[-74.68505859374999,41.42625319507272],[-75.0146484375,41.68932225997044],[-75.11352539062499,42.0125705565935]]]}}
```
caseypt: Consider updating this commit message with a note about the layer source.

caseypt:
The radio button is disabled when you are outside of PA for "PA DEP Urban Areas", but the name is not grayed out like disabled "Boundary" layers.

image

image

caseypt:
+1, tested.

image

mmcfarland:
Updated Coverage layer name style for disabled text in  	9e86544

screenshot from 2016-07-19 15 42 28

mmcfarland:
I'll update the commit message for the geojson provenance prior to merging.

src/mmw/mmw/settings/data/pa_perimeter.json
mmcfarland: Done

PR #1392, opened 2016-07-20
Fix layer selector text from being always disabled

mmcfarland:
Fixes a bug introduced in 5655f726 that resulted in all Boundary layer
text in being disabled "looking" even when they were in fact enabled.

#### Test
Ensure that when zooming high to low, and in and out of PA that the Boundary layers disabled styling updates accordingly.

Before: 
screenshot from 2016-07-20 11 36 55

After:
screenshot from 2016-07-20 11 37 37

### DIFF
--- a/src/mmw/sass/base/_map.scss
+++ b/src/mmw/sass/base/_map.scss
@@ -77,7 +77,7 @@
   height: 100%;
 }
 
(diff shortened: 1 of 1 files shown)


caseypt:
+1, tested.

image

image

Issue #1238, opened 2016-04-11
Map attribution cut off on "satellite with roads" and "terrain" basemaps


ajrobbins:
Observed on Chrome, Firefox, and Safari. 

The top of an additional line of attribution appears to be visible, so potentially there is a wrapping issue.
![screen shot 2016-04-11 at 1 33 11 pm](https://cloud.githubusercontent.com/assets/12868342/14436327/27e28ad0-ffea-11e5-8cb1-4f9d2165c63f.png)

caseypt:
Shares a point with #1237

Issue #1237, opened 2016-04-11
Basemaps appear to be misaligned


ajrobbins:
The "satellite with roads" and "terrain" basemaps appear to be shifted slightly south from the "satellite" and "streets" basemaps. This is a bit difficult to observe, but is helped by turning the NLCD overlay on. You can see that the roads on the streets and satellite basemaps align with the NLCD land use classification, but the other two basemaps do not.

It's unclear if this issue is actually zoom-dependent, but it's pretty much only perceptible around zoom levels 14-15.

Streets (zoom level 15)
![screen shot 2016-04-11 at 1 17 54 pm](https://cloud.githubusercontent.com/assets/12868342/14436079/01980b4e-ffe9-11e5-80d9-911fca74e070.png)

Terrain (zoom level 15)
![screen shot 2016-04-11 at 1 17 59 pm](https://cloud.githubusercontent.com/assets/12868342/14436085/07bbc3c6-ffe9-11e5-91e1-4c7bee3db869.png)

mmcfarland:
This actually appears to not be zoom dependent or projection related.  When zoomed out and displaying a DEP municipal boundary, you can see the offset around the DE/PA border.  The offset is not spatially consistent when zooming in, but does appear to be pixel/viewport consistent with a gutter observed at the top of the map when these google layers are turned out:

![screenshot from 2016-07-19 11 17 20](https://cloud.githubusercontent.com/assets/1014341/16955470/bdd5bc82-4da2-11e6-8582-dcfc11e9818e.png)

I believe this is the same root cause of #1238 as well.  

The culprit seems to be this [css rule](https://github.com/WikiWatershed/model-my-watershed/blob/f13c3c444c902daf4ae591d3e3a2fad59980a3e5/src/mmw/sass/base/_map.scss#L105-L107) which is adding the gutter.  Disabling that rule results in the correct display:

![screenshot from 2016-07-19 11 24 42](https://cloud.githubusercontent.com/assets/1014341/16955640/6765e4c0-4da3-11e6-9880-b5bb12032c13.png)

Apparently the class `leaflet-top` only gets added to the map container for Google base maps, it does not appear to be used when other layer sources are used.

caseypt:
It looks like the Google leaflet plugin adds this class https://github.com/shramov/leaflet-plugins/blob/master/layer/tile/Google.js#L91.

I'll try and make our rule more specific.

PR #1396, opened 2016-07-25
Make .leaflet-top CSS override rule more specific

caseypt:
Only apply the modified .leaflet-top rule to elements in
.leaflet-control-container. The rule was originally added to target
these elements. Since it was applied more broadly, it caused issues
with the Google Leaflet plugin, which adds the leaflet-top class
to the container used to house the Google tiles.

**Testing**
- Turn on the NLCD layer.
- Switch between the various basemaps and ensure the NLCD overlay is generally in the same place for each layer.
- Ensure there is no gray bar across the top of the map and the attribution element is visible when the "Satellite with Roads" layer is active.

image

Connects to #1237
Connects to #1238

### DIFF
--- a/src/mmw/sass/base/_map.scss
+++ b/src/mmw/sass/base/_map.scss
@@ -102,7 +102,7 @@
     float: right;
 }
 
-.leaflet-top {
+.leaflet-control-container .leaflet-top {
   top: 45px;
 }
 


ajrobbins:
Checked this out on @caseypt's machine, basemap alignment looks good:
Streets
screen shot 2016-07-25 at 1 27 48 pm

Satellite
screen shot 2016-07-25 at 1 27 56 pm

Satellite with Roads
screen shot 2016-07-25 at 1 28 02 pm

Still seeing the attribution cut off (#1238) though - not sure if this PR is supposed to address that as well.

caseypt:
@ajrobbins Sorry, can you check it again?

ajrobbins:
@caseypt Seeing cut off attribution: 
screen shot 2016-07-25 at 3 45 48 pm

ajrobbins:
Just checked on Casey's machine and the Google attribution is there for satellite with streets and terrain - It's just on a different line. +1

kellyi:
Taking a look now.

kellyi:
Looks good to me. It'd be nice to make the Google attribution more visible even beneath the layer, but I guess there may not be a class to style for it with CSS since it's baked into the map tile? If it is possible, it may be worth making it even more legible.

+1

caseypt:
The style rules making the Google attribution hard to read are all inline rules created by the Google Maps API (I think), so they'll be tough to override without some hacks.

kellyi:
That makes sense. It may be worth creating another issue just for that in case we want to consider it somewhere down the line. Rest of this looks good, though!

caseypt:
Created https://github.com/WikiWatershed/model-my-watershed/issues/1401.

Thanks!

Issue #1123, opened 2016-02-02
Soil Group Colors for Micro Site Storm Model


SRGDamia1:
Can the selection highlights for the soil groups in the micro site model be changed to match the soils legend, just like the land covers do?  This is *very low priority*, but if it is a very simple change, it would be nice to see.

mmcfarland:
Yep, this is easy and was an oversight, we plan on keeping that useful link between apps - thanks for the reminder.

PR #1399, opened 2016-07-25
Update micro site with soil colors

caseypt:
Apply the same colors used to symbolize the various soil types on the main site in the micro site. 

**Testing**
- Visit the micro site.
- Select the various soil types and verify that the tile checkmark appears in the correct color for the active soil type.
- Hover over each soil type and verify that the popover header appears in the correct color for the soil type that is being hovered on.

image

Connects to #1123

### DIFF
--- a/src/mmw/sass/pages/_water-balance.scss
+++ b/src/mmw/sass/pages/_water-balance.scss
@@ -440,6 +440,38 @@
               background-color: $nlcd-90;
             }
           }
+
+          &.soil-a {
+            border-color: $soil-a;
+
+            &:after {
+              background-color: $soil-a;
+            }
+          }
+
+          &.soil-b {
+            border-color: $soil-b;
+
+            &:after {
+              background-color: $soil-b;
+            }
+          }
+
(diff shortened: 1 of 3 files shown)


src/mmw/sass/pages/_water-balance.scss
```diff
@@ -440,6 +440,38 @@
               background-color: $nlcd-90;
             }
           }
+
+          &.soil-a {
+            border-color: $soil-a;
+
+            &:after {
+              background-color: $soil-a;
```
rajadain: The white check-mark is not visible with these light colors:

image

Consider adding `color: #333` in the `&:after` block for all blocks that have dark text over a light background in the pop-up, namely High and Moderate Infiltration in the Soils block, and Wetlands, Grasslands, and Pasture / Hay in the Land Cover block.

caseypt:
Added a commit with your suggestions.

rajadain:
+1 tested! Looks so much more consistent now!

caseypt:
Thanks!

PR #1400, opened 2016-07-25
Remove biochemical oxygen demand component

kellyi:
This PR removes the "biochemical oxygen demand" factor from TR-55 visualizations. 

- remove oxygen demand from water quality graph and chart
- remove oxygen demand from "Compare" bar chart

**Testing**
- grab this branch, `vagrant up`, and `./scripts/bundle.sh`
- visit `localhost:8000` in the browser and open the developer console
- choose a Square Kilometer draw area and select a spot on the map
- when the "Analyze" sidebar box appears, choose the "Site Storm Model" option to run TR-55
- verify that "Biochemical Oxygen Demand" no longer appears in either the "Runoff" or "Water Quality" visualizations
- click on "Compare". On that page, use the select boxes to toggle between "Runoff" and "Water Quality" for each scenario and verify that "Biochemical Oxygen Demand" no longer appears in those visualization
- verify that there weren't any errors in the console.

Connects #1174

### DIFF
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -149,11 +149,10 @@ def aoi_resolution(area_of_interest):
 
 
 def format_quality(model_output):
-    measures = ['Biochemical Oxygen Demand',
-                'Total Suspended Solids',
+    measures = ['Total Suspended Solids',
                 'Total Nitrogen',
                 'Total Phosphorus']
-    codes = ['bod', 'tss', 'tn', 'tp']
+    codes = ['tss', 'tn', 'tp']
 
     def fn(input):
         measure, code = input
--- a/src/mmw/js/src/core/utils.js
+++ b/src/mmw/js/src/core/utils.js
@@ -234,6 +234,10 @@ var utils = {
         }
     },
 
+    filterOutOxygenDemand: function(element) {
+        return element.measure !== "Biochemical Oxygen Demand";
+    },
+
     // Reverse sorting of a Backbone Collection.
     // Taken from http://stackoverflow.com/a/12220415/2053314
     reverseSortBy: function(sortByFunction) {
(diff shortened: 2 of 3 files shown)


lewfish:
This now has merge conflicts.

kellyi:
Fixing now.

kellyi:
Should be good now!

lewfish:
Biochemical Oxygen Demand still shows up in the chart and table, except for the Compare view, where it has been removed.

kellyi:
Hmmm. Let me look again -- I may have neglected to commit an additional file.

kellyi:
Ok, this should work now. This version filters the "Biochemical Oxygen Demand" factor out of the `this.model.get('result').quality` array so that it's not present to be visualized. I added a utility function to handle the repeated logic, and placed it in ...`/core/utils` then imported it in since it was being used in two Marionette components in `/tr55/quality/views.js`. (I'm happy to move this function to the top of the `views.js` file -- I just didn't know whether having a utility function there  was idiomatic in Marionette.)

Testing instructions are the same as above, but this time you should 

- place this line at lines 59 and 174 of the `/tr55/quality/views.js` file so that you can compare that the values labeled in the visualization are the same as the ones in the data array:

```
window.console.log(this.model.get('result').quality.filter(utils.filterOutOxygenDemand));
```

- draw a square KM AOI on the map, then run the TR-55 model
- when the results return, click on the "Water Quality" tab. Verify that "Biochemical Oxygen Demand" is no longer present in the table or bar chart, and verify that the label-value pairs presented in the chart match the `console.log`ged objects' labels and values
- click "Compare" and select "Water Quality" for one of the panes, and verify that the label-value pairs presented there match the objects' label-values in the console.

lewfish:
+1, tested

kellyi:
Thanks for taking another look!

Issue #1407, opened 2016-08-09
Use “lower resolution” NHD streams for LS calculation


mmcfarland:
For LS calculations, disable high resolution DRB code paths to prefer "low res" NHD streams always.

PR #1413, opened 2016-08-10
Use NHD streams for `ls_factors` calculation

kellyi:
This PR updates MapShed's `tasks.py` and `calcs.py` files so that they'll always use `nhdflowline` data when calculating `ls_factors`. 

The changes remove a check in `tasks.py` which would update the `ls_factors` input data if the geom was in the DRB and also remove -- since it wasn't being used anymore -- some logic from the `stream_length` method which would use `drb_streams` if called.

**Testing**
- get this branch
- add the following line to https://github.com/kellyi/model-my-watershed/blob/8ce422a36aaeefa5f9545ae2dc4da5031977eadf/src/mmw/apps/modeling/mapshed/tasks.py#L182

```python
print('`ls_factors` result is => {}`'.format(z['LS']))
```

- `vagrant up`, `./scripts/bundle.sh && ./scripts/debugserver.sh`, then open a new terminal window and run `./scripts/debugcelery.sh`
- draw an AOI on the map, then select a "Watershed Multi-Year Model" to run MapShed
- switch back to the `debugcelery` window and verify that you see a printed warning showing you that the `ls_factors` result has a value.
- verify that the results appear in the browser and that there aren't any celery errors or browser console errors

Connects #1407

### DIFF
--- a/src/mmw/apps/modeling/mapshed/tasks.py
+++ b/src/mmw/apps/modeling/mapshed/tasks.py
@@ -39,7 +39,6 @@ from apps.modeling.mapshed.calcs import (day_lengths,
 NLU = settings.GWLFE_CONFIG['NLU']
 NRur = settings.GWLFE_DEFAULTS['NRur']
 AG_NLCD_CODES = settings.GWLFE_CONFIG['AgriculturalNLCDCodes']
-DRB = settings.DRB_PERIMETER
 ACRES_PER_SQM = 0.000247105
 HECTARES_PER_SQM = 0.0001
 SQKM_PER_SQM = 0.000001
@@ -177,14 +176,7 @@ def collect_data(geop_result, geojson):
     z['SedAFactor'] = sed_a_factor(geop_result['landuse_pcts'],
                                    z['CN'], z['AEU'], z['AvKF'], z['AvSlope'])
 
-    if geom.within(DRB):
-        ls_stream_length = stream_length(geom, drb=True)
-        ls_stream_pcts = geop_result['lu_stream_pct_drb']
-    else:
-        ls_stream_length = z['StreamLength']
-        ls_stream_pcts = geop_result['lu_stream_pct']
-
-    z['LS'] = ls_factors(ls_stream_pcts, ls_stream_length,
+    z['LS'] = ls_factors(geop_result['lu_stream_pct'], z['StreamLength'],
                          z['Area'], z['AvSlope'])
 
     z['P'] = p_factors(z['AvSlope'])
@@ -468,12 +460,6 @@ def geop_tasks(geom, errback, exchange, choose_worker):
          {'polygon': [geom.geojson]},
          nlcd_kfactor)
     ]
-
-    if geom.within(DRB):
-        definitions.append(('nlcd_streams',
-                            {'polygon': [geom.geojson],
-                             'vector': streams(geom, drb=True)},
-                            nlcd_streams_drb))
 
     return [(mapshed_start.s(opname, data).set(exchange=exchange,
                                                routing_key=geop_worker) |


src/mmw/apps/modeling/mapshed/calcs.py
```diff
@@ -254,11 +254,9 @@ def ls_factor(stream_length, area, avg_slope, m):
         return ls
 
 
-def stream_length(geom, drb=False):
```
mmcfarland: Was `LS` the only calc that had been using the switch?  I seem to recall another calculation that made the same evaluation on location.

kellyi: Let me double check. I searched through the whole project and don't think I saw any other invocations of the `stream_length` method.

There's a `streams(geom, drb=False)` method with similar logic that's called elsewhere.

kellyi: Yeah, looks like just two calls to `stream_length` in `tasks.py`:

This one sets the value for `z['StreamLength']`: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/mapshed/tasks.py#L123

And this one's in the DRB logic I cut: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/mapshed/tasks.py#L181

Here's the call to `streams`, which has a similar signature: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/mapshed/tasks.py#L475

rajadain:
I don't think we should take DRB calculation out completely: we should simply stop using it. I think something more along the lines of this:

```diff
diff --git a/src/mmw/apps/modeling/mapshed/tasks.py b/src/mmw/apps/modeling/mapshed/tasks.py
index 8082e3e..218adf3 100644
--- a/src/mmw/apps/modeling/mapshed/tasks.py
+++ b/src/mmw/apps/modeling/mapshed/tasks.py
@@ -39,7 +39,6 @@ from apps.modeling.mapshed.calcs import (day_lengths,
 NLU = settings.GWLFE_CONFIG['NLU']
 NRur = settings.GWLFE_DEFAULTS['NRur']
 AG_NLCD_CODES = settings.GWLFE_CONFIG['AgriculturalNLCDCodes']
-DRB = settings.DRB_PERIMETER
 ACRES_PER_SQM = 0.000247105
 HECTARES_PER_SQM = 0.0001
 SQKM_PER_SQM = 0.000001
@@ -177,12 +176,8 @@ def collect_data(geop_result, geojson):
     z['SedAFactor'] = sed_a_factor(geop_result['landuse_pcts'],
                                    z['CN'], z['AEU'], z['AvKF'], z['AvSlope'])
 
-    if geom.within(DRB):
-        ls_stream_length = stream_length(geom, drb=True)
-        ls_stream_pcts = geop_result['lu_stream_pct_drb']
-    else:
-        ls_stream_length = z['StreamLength']
-        ls_stream_pcts = geop_result['lu_stream_pct']
+    ls_stream_length = z['StreamLength']
+    ls_stream_pcts = geop_result['lu_stream_pct']
 
     z['LS'] = ls_factors(ls_stream_pcts, ls_stream_length,
                          z['Area'], z['AvSlope'])
@@ -469,12 +464,6 @@ def geop_tasks(geom, errback, exchange, choose_worker):
          nlcd_kfactor)
     ]
 
-    if geom.within(DRB):
-        definitions.append(('nlcd_streams',
-                            {'polygon': [geom.geojson],
-                             'vector': streams(geom, drb=True)},
-                            nlcd_streams_drb))
-
     return [(mapshed_start.s(opname, data).set(exchange=exchange,
                                                routing_key=geop_worker) |
              mapshed_finish.s().set(exchange=exchange,
```

would excise the usage, while preserving the capability should we ever need to return to it.

kellyi:
Sounds good -- I'll add the logic back to `calcs.py` so we'll have it there. I *think* the issue this addresses wanted the DRB calc removed only from creating `ls_factors`, but not necessarily from other uses like `'vector': streams(geom, drb=True)},`.

kellyi:
Pushed an updated commit which includes only changes to `tasks.py`.

rajadain:
The `nlcd_streams_drb` task [outputs](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/mapshed/tasks.py#L280) `lu_stream_pct_drb` which is only used for the LS calculation. Since we're not using that anymore, we should not call it. It is by far the most time taking GeoTrellis call, and should speed up MapShed considerably.

kellyi:
Ah, got it. I'll cut those lines too. Thanks!

mmcfarland:
> which is only used for the LS calculation. Since we're not using that anymore, we should not call it.

Good catch.

rajadain:
Tested with DRB factoring in LS (`develop`) and without (this branch), and saw these differences in the GMS file:

```diff
--- Little_Nishemany with DRB__Current_Conditions.gms	2016-08-10 18:11:10.000000000 -0400
+++ Little_Nishemany without DRB__Current_Conditions.gms	2016-08-10 18:15:35.000000000 -0400
@@ -17,14 +17,14 @@
 Oct,1.0457270621091717,10.8,0,0.091,0,0,1.385
 Nov,1.0087081300568108,9.7,0,0.091,0,0,1.385
 Dec,0.9872371494664416,9.1,0,0.091,0,0,1.385
-Hay/Past,803.0651091556414,75,0.2714132968948771,0.20970380934911945,0.03,0.45
-Cropland,571.1105265178822,82,0.27604142433856804,0.20820229288712816,0.42,0.45
-Forest,2071.0294885185103,73,0.27157785493664266,0.2058387525439958,0.002,0.45
-Wetland,364.718653026077,88.5,0.31307519406197576,0.20205938680533908,0.01,0.1
+Hay/Past,803.0651091556414,75,0.2714132968948771,0.22836631257618079,0.03,0.45
+Cropland,571.1105265178822,82,0.27604142433856804,0.22735773644384383,0.42,0.45
+Forest,2071.0294885185103,73,0.27157785493664266,0.2151564164245495,0.002,0.45
+Wetland,364.718653026077,88.5,0.31307519406197576,0.20441617849311078,0.01,0.1
 Disturbed,0,0,0,0,0.08,0.1
 Turfgrass,0,0,0,0,0.03,0.2
-Open_Land,61.56652484448092,87,0.26802137526366693,0.21142463498011502,0.04,0.45
-Bare_Rock,32.76347228565944,89,0.27985624785802954,0.20874317883392307,0.001,0.45
+Open_Land,61.56652484448092,87,0.26802137526366693,0.25482037505979976,0.04,0.45
+Bare_Rock,32.76347228565944,89,0.27985624785802954,0.24386862750396412,0.001,0.45
 Sandy_Areas,0,0,0,0,0.01,0.45
 Unpaved_Road,0,0,0,0,0.8,1
 Ld_Mixed,5921.547567936711,0.15,92,74,60
```

As can be seen in a [highlighted diff](https://github.com/so-fancy/diff-so-fancy), the changed fields are in Line 20-29 comma position 5, which corresponds to LS as specified in the [GMS File Key](https://docs.google.com/spreadsheets/d/12hsklRHwl6sNum0xPk6d5Cw6ONDA3X6BR5-Pc-NwJBc/edit#gid=0):

image

Curiously I did not see any differences in the Hydrology table's values though. Perhaps the difference isn't big enough to show up?

The processing is markedly faster, though.

rajadain:
Given the above I'm not entirely sure if this would satisfy the client. I, on the other hand, am completely satisfied. +1

mmcfarland:
@ajrobbins can you follow up with BME to see if the changes in the GMS seem appropriate or what he was anticipating, and how they don't seem to influence the hydrology tab?  @rajadain what approximate area/shape did you use to generate that GMS?

rajadain:
> what approximate area/shape did you use to generate that GMS?

I used BME's favorite Little Neshaminy Creek HUC-12:

image

I will add the GMS files here as soon as I get in.

rajadain:
The files I used above:

 * [Little_Neshaminy with DRB__Current_Conditions.gms.txt](https://github.com/WikiWatershed/model-my-watershed/files/413541/Little_Nishemany.with.DRB__Current_Conditions.gms.txt)
 * [Little_Neshaminy without DRB__Current_Conditions.gms.txt](https://github.com/WikiWatershed/model-my-watershed/files/413540/Little_Nishemany.without.DRB__Current_Conditions.gms.txt)

The Little Neshaminy HUC-12 is just slightly north of Philadelphia:

image

ajrobbins:
Let's hold off on merging this, as we may need to go with an alternate method for calculating LS anyway. I sent Barry the diff and he said:
>The LS factors are about a third less than what we calculate in MapShed with our "flow accumulation" routine (which we originally thought we could do in MMW).

ajrobbins:
@rajadain I also just sent BME the gms files to double check that it's still not right.

mmcfarland:
I think we should consider merging this, despite the additional changes that may come for LS calculation.  I think if we're not going to be using the stream calc (esp DRB), the removal of that geoprocessing step is still highly beneficial.

ajrobbins:
Good point, if it speeds up the geoprocessing we should merge it. We will have to revisit this LS calculation, though.

rajadain:
> We will have to revisit this LS calculation, though.

True. Let's make a new card for that.

kellyi:
Made a new issue for `ls_factors` here: https://github.com/WikiWatershed/model-my-watershed/issues/1418

Issue #1371, opened 2016-06-15
Add a "totals" row to hydrology table


ajrobbins:
Add a row at the bottom of the hydrology table with the variable totals. 

![screen shot 2016-06-15 at 10 31 19 am](https://cloud.githubusercontent.com/assets/12868342/16083796/59833a4e-32e4-11e6-936d-bcac71f26549.png)

rajadain:
Would there be value in adding an average row as well?

mmcfarland:
> Would there be value in adding an average row as well?

It wasn't requested, let's just keep to a totals row.

PR #1414, opened 2016-08-10
Add Hydrology Totals

rajadain:
## Overview

Adds a footer element to the Hydrology table, and populates it with the total sum of each column.

## Demo

mmw-hydrology-total

## Testing Instructions

Checkout the branch and update packages to get the latest bootstrap table.

```bash
./scripts/npm.sh install
./scripts/bundle.sh --vendor
```

Load or create a MapShed project. Wait for the calculations to complete, then observe the Hydrology table. It should have a new "Total" row at the bottom, which should scroll horizontally with the rest of the table. Sorting the columns should not affect the row, which should stay at the bottom. The sum should be accurate for each column.

Ensure that other instances of bootstrap table (in the Analyze, Water Quality, and TR-55 tabs) continue to work as they should.

## Caveats

The bootstrap table's footer element is not well implemented, and has a number of slightly breaking edge cases. Resizing the window seems to disalign the columns ever so slightly. Overscrolling (on macOS or iOS for example) has the bounce effect only on the main table, not the header or the footer. Finally, sorting the columns triggers a recalculation of the totals which is why they flicker. I'm not quite sure how to address that.

Connects #1371


src/mmw/js/src/modeling/gwlfe/runoff/templates/table.html
```diff
@@ -1,13 +1,13 @@
-<table class="table custom-hover" data-toggle="table">
+<table class="table custom-hover" data-toggle="table" data-height="auto" data-show-footer="true">
```
rajadain: We need `data-height="auto"` otherwise the footer doesn't scroll horizontally with the rest of the table 😞

src/mmw/js/src/modeling/gwlfe/runoff/templates/table.html
```diff
@@ -1,13 +1,13 @@
...
         <tr>
             {% for columnName in columnNames %}
                 {% if loop.index0 == 0 %}
-                    <th data-sortable="true" data-field="date" data-sort-name="_date_data" data-sorter="window.monthSorter">
+                    <th data-sortable="true" data-field="date" data-sort-name="_date_data" data-sorter="window.monthSorter" data-footer-formatter="Total">
                         {{ columnName }}
                     </th>
                 {% else %}
-                    <th data-sortable="true" data-sorter="window.numericSort">{{ columnName }} (cm)</th>
+                    <th data-sortable="true" data-sorter="window.numericSort" data-footer-formatter="window.sumFormatter" data-align="right">{{ columnName }} (cm)</th>
```
rajadain: It seemed easiest to use a `window.` function here, just like we do for `numericSort`. Not sure if there's better ways to do this.

src/mmw/js/src/modeling/gwlfe/runoff/templates/table.html
```diff
@@ -19,7 +19,7 @@
                     {% if loop.index0 == 0 %}
                         <td class="text-left" data-month="{{ value }}">{{ monthNames[value] }}</td>
                     {% else %}
-                        <td class="strong text-right">{{ value|round(2)|toLocaleString(2) }}</td>
```
rajadain: Switched from using a `text-right` class to `data-align="right"` on the `th` so that the right alignment carries through to the footer (which is in another table).

src/mmw/sass/components/_tables.scss
```diff
@@ -45,7 +45,19 @@
   }
 }
 
-.fixed-table-container thead th .th-inner {
-  white-space: normal;
-  line-height: 1.25;
+.fixed-table-container {
+  padding-bottom: 0px !important;
```
rajadain: Setting `data-height="auto"` above adds a large amount of padding, which we have to remove here.

src/mmw/sass/components/_tables.scss
```diff
@@ -45,7 +45,19 @@
...
+    white-space: normal;
+    line-height: 1.25;
+  }
+}
+
+.fixed-table-footer tr {
+    font-weight: bold;
+
+    td:nth-child(n+2) {
+        color: $brand-primary;
```
rajadain: Makes sure that the total values are highlighted in the same color as the main values.

src/mmw/js/src/modeling/gwlfe/runoff/views.js
```diff
@@ -154,6 +154,16 @@ function monthSorter(a, b) {
 
 window.monthSorter = monthSorter;
 
+function SumFormatter(data) {
```
rajadain: This is defined as capitalized `SumFormatter` instead of the regular `sumFormatter` because otherwise jslint gives the following error:

```
> kj npm run lint
+ ARGS='run lint'

js/src/modeling/gwlfe/runoff/views.js: line 158, col 17, Possible strict violation.

1 error
```

for this line:

```js
var field = this.field,
```

See here for details: http://stackoverflow.com/a/27117322/2053314

src/mmw/js/src/modeling/gwlfe/runoff/views.js
```diff
@@ -154,6 +154,16 @@ function monthSorter(a, b) {
...
+    var field = this.field,
+        total = data.reduce(function(sum, row) {
+            return sum + (parseFloat(row[field]) || 0);
+        }, 0);
```
lewfish: Consider turning this into a map and a sum operation, which might be easier to read.

lewfish:
+1, tested. Good job finding these little hacks to make it work.

rajadain:
Updated to use `_.sum(_.map(...))`. Thanks for taking a look. Merging shortly.

Issue #1370, opened 2016-06-15
No value for Septic Systems


mmcfarland:
In gwlfe hydrology output, septic system source is always zero.  Check that we are estimating the values as prescribed below:

> Also check to see if you have the following code in your front-end pre-processor:

```vb
If SeptFlag = 1 And EstSS = False Then

            ' Calculate septic system values

            CalcSepticSystems()

        ElseIf EstSS Then

            ' Estimate septic system values

            SewPublic = 0

            ShortSys = 0

            SewSeptic = (SSLDR * LuArea(14)) + (SSLDM * LuArea(11)) ' Set number on direct/public systems

        End If

        NormalSys = SewSeptic ' Set number on normal systems

        Ponding = 0 ' Set number on pond systems

        ShortSys = SewOther ' Set number of short circuit systems

```

> If you do, just set SSLDR and SSLDM equal to "3.3", and you should be ok with your septic system calculations (because right now, they are probably both set to "0").

![screenshot 2016-06-15 at 10 18 15 am](https://cloud.githubusercontent.com/assets/1014341/16083456/0079f5a6-32e3-11e6-97a5-00840de1095c.png)

ajrobbins:
Further info from BME:
- In MapShed we now use “low-density” urban land to estimate septic system populations.
- For MMW, multiply hectares of “low-density mixed” by 3.3 persons per ha to get total population of people on septic systems in the AOI
- Assign this value to the “Normal” category, which is written to position 1 on lines 86-97 of the gms file

kellyi:
`SSLDR` and `SSLDM` are currently set to 4.4 and 2.2, respectively:

https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/gwlfe_settings.py#L471

kellyi:
Ran `debugcelery` and it seems like this method is generating a non-zero value -- https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/mapshed/calcs.py#L572 -- so I'm going to check whether it's not getting assigned or displayed properly.

kellyi:
`z['NormalSys']` has a value right before it's sent to the GWLF-E GMS parser to create the input file here -- https://github.com/kellyi/model-my-watershed/blob/develop/src/mmw/apps/modeling/tasks.py#L360 -- so I think that value may not be being read in GWLF-E.

kellyi:
[...]

PR #1420, opened 2016-08-15
Adjust `NormalSys` values in pre-processor

kellyi:
This PR updates some of the code by which MMW creates its MapShed dictionary for GWLF-E by changing the names of `NormalSys` values to match GWLF-E's expected key -- `NumNormalSys`. It also changes the value of that key to an array of 12 ints, which is what GWLF-E expects to receive for calculating the "Total Nitrogen" value for Septic Systems.

screen shot 2016-08-15 at 11 40 37 am

The result so far only addresses the Total Nitrogen value: "Sediment" and "Total Phosphorus" remain 0 because we are using hard-coded 0 values for each.

The GWLF-E code returns a 0 for Septic/Sediment here: 
https://github.com/WikiWatershed/gwlf-e/blob/develop/gwlfe/WriteOutputFiles.py#L887

MMW hard codes the values necessary to generate "Total Phosphorus" to an array of 12 zeroes here:
https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/gwlfe_settings.py#L99

**Testing**
- grab this branch, `vagrant up`, `./scripts/bundle.sh` then visit `localhost:8000`
- select an area of interest to analyze, then run the MapShed model
- when modeling's done, select the "Water Quality" tab and scroll down to verify that there's a value for "Total Nitrogen"
- create a new scenario, then export a gms file
- from the command line run `cat <your-gms-file>.gms | sed '85,97!d'` and verify that position 1 at each of the lines after the first has a value

Connects #1370

### DIFF
--- a/src/mmw/apps/modeling/mapshed/calcs.py
+++ b/src/mmw/apps/modeling/mapshed/calcs.py
@@ -554,7 +554,7 @@ def landuse_pcts(n_count):
     ]
 
 
-def normal_sys(lu_area):
+def num_normal_sys(lu_area):
     """
     Given the land use area in hectares, estimates the number of normal septic
     systems based on the constants SSLDR and SSLDM for Residential and Mixed
@@ -563,13 +563,18 @@ def normal_sys(lu_area):
     effectively dependent only on the area of medium density mixed land use.
     However, we replicate the original formula for consistency.
 
+    Returns an array with an integer value for each month of the
+    year as input for GWLF-E.
+
     Original at Class1.vb@1.3.0:9577-9579
     """
 
     SSLDR = settings.GWLFE_CONFIG['SSLDR']
     SSLDM = settings.GWLFE_CONFIG['SSLDM']
 
-    return SSLDR * lu_area[14] + SSLDM * lu_area[11]
+    normal_sys_estimate = SSLDR * lu_area[14] + SSLDM * lu_area[11]
+    normal_sys_int = int(round(normal_sys_estimate))
+    return [normal_sys_int for n in xrange(12)]
 
 
 def sed_a_factor(landuse_pct_vals, cn, AEU, AvKF, AvSlope):
--- a/src/mmw/apps/modeling/mapshed/tasks.py
+++ b/src/mmw/apps/modeling/mapshed/tasks.py
@@ -32,7 +32,7 @@ from apps.modeling.mapshed.calcs import (day_lengths,
                                          groundwater_nitrogen_conc,
                                          sediment_delivery_ratio,
                                          landuse_pcts,
-                                         normal_sys,
+                                         num_normal_sys,
                                          sed_a_factor
                                          )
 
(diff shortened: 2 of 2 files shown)


src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -148,7 +148,8 @@ def collect_data(geop_result, geojson):
     z['UrbAreaTotal'] = sum(z['Area'][NRur:])
     z['PhosConc'] = phosphorus_conc(z['SedPhos'])
 
-    z['NormalSys'] = normal_sys(z['Area'])
+    normal_sys_estimate = int(num_normal_sys(z['Area']))
```
kellyi: Casting these as `int`s because that's what GWLF-E expects to receive.

rajadain: Consider moving this and the array formation within `num_normal_sys`, so that here we have only

```python
z['NumNormalSys'] = num_normal_sys(z['Area'])
```

Every other line in this method populates `z` directly.

rajadain:
I just tested with my standard testing HUC-12 (Little Neshaminy) and didn't see any values for Septic Systems in the Water Quality tab or in the GMS file:

image

kellyi:
Oof. Let me see what went wrong. Does the GMS output from that have values in the right places?

rajadain:
Doesn't look like it: [Untitled_Project__Current_Conditions.gms.txt](https://github.com/WikiWatershed/model-my-watershed/files/419000/Untitled_Project__Current_Conditions.gms.txt)


```
> cat Untitled_Project__Current_Conditions.gms | sed '85,97!d'
1
0,0,0,0,0
0,0,0,0,0
0,0,0,0,0
0,0,0,0,0
0,0,0,0,0
0,0,0,0,0
0,0,0,0,0
0,0,0,0,0
0,0,0,0,0
0,0,0,0,0
0,0,0,0,0
0,0,0,0,0
```

rajadain:
Sorry, my fault. Hadn't restarted Celery after switching to this branch. Working now.

kellyi:
Oh! I was testing it out and trying to see if I was accidentally rounding down. However, that made me think that I should probably round the float up or down rather than just rounding it down to an int.

rajadain:
The values for Little Neshaminy:

```
> cat Little_Neshaminy_with_Septic_Systems__Current_Conditions.gms | sed '86,97!d'
2173,0,0,0,0
2173,0,0,0,0
2173,0,0,0,0
2173,0,0,0,0
2173,0,0,0,0
2173,0,0,0,0
2173,0,0,0,0
2173,0,0,0,0
2173,0,0,0,0
2173,0,0,0,0
2173,0,0,0,0
2173,0,0,0,0
```

image

rajadain:
+1 tested

src/mmw/apps/modeling/mapshed/tasks.py
kellyi: Done! Also updated the docstring for that method to explain why it was returning an array of ints.

src/mmw/apps/modeling/mapshed/calcs.py
```diff
@@ -563,13 +563,17 @@ def normal_sys(lu_area):
     effectively dependent only on the area of medium density mixed land use.
     However, we replicate the original formula for consistency.
 
+    Returns an array of 12 ints as input for GWLF-E.
```
rajadain: Instead of saying "12 ints" (which describes the form) consider using something that describes the function, such as "Return a value for each month of the year."

kellyi: Makes sense -- I'll push a commit with that update.

Issue #649, opened 2015-07-31
Windows Touch Screen support


mmcfarland:
With work on #502 we established some parameters of devices that we were willing not to support due to problems in our L.Draw fork.  A workshop user was unable to close a polygon on the draw screen using an HP Intel ProBook, which is claiming _not_ to be a touch device, although the screengrab of the settings for that computer indicate that there is some touch support (see `System` -> `Pen and Touch` below).

![img_0542](https://cloud.githubusercontent.com/assets/1014341/9007774/aab1e448-3762-11e5-82ca-b2bf461f39f0.jpg)

rajadain:
Just tested on a Windows touchscreen with Leaflet.Draw [0.3.0](https://github.com/Leaflet/Leaflet.draw/releases/tag/v0.3.0) (latest release version) and [0.3.2](https://github.com/Leaflet/Leaflet.draw/releases/tag/v0.3.2) (latest available tag), with not great results.

Chrome 52 still doesn't close shapes when using the mouse. Touch works correctly.
Firefox 42 closes shapes correctly using the mouse, but Touch is _really_ flaky.
Edge and IE11 seem to be working perfectly in both cases, which is an improvement over the last version where the site wouldn't render in IE.

So for our Windows users, we should recommend Edge.

I'm going to put up a PR that upgrades to 0.3.0, the latest stable version. There is https://github.com/Leaflet/Leaflet.draw/issues/548, which I've subscribed to. Unfortunately, if we want to investigate the guts of Leaflet.draw to see if this can be figured out, we'll need more points.

Issue #1190, opened 2016-03-14
Unable to close shape - 'Draw Area' - 'free draw'


greytilde:
Creating a simple 4 sided shape,   when clicking back on the first corner, the shape would not close. Click radius might be too small. No amount of clicking would cause the shape to close. Only additional points were created (see screen shot)
<img width="960" alt="unable to close shape - free draw" src="https://cloud.githubusercontent.com/assets/8659605/13732665/1394bc30-e95e-11e5-8cde-0f9945649998.png">

rajadain:
Duplicate of #649. Also see #892 (closed).

SRGDamia1:
We confirmed that this is a problem on a Surface 4 running Windows 10.

I was also looking over the series of these issues wrt touch on leaflet draw and found that the newest release of leaflet.draw (0.3.0, released March 9, 2016) is now supposed to support touch devices: https://github.com/Leaflet/Leaflet.draw/blob/master/CHANGELOG.md#030-march-09-2016

rcheetham:
I noticed the same thing in Chrome 50 on Windows 10. I do think it's likely a scale dependent issue. The cursor changes when you are over the vertex (in other words, the ability to detect cursor location intersection with the vertex seems to be fine), but I was only able to successfully close a shape perhaps 1 in 4 times.

SRGDamia1:
I just got another email of someone reporting the same bug.

rajadain:
I just put up #1421 to upgrade Leaflet.Draw to 0.3.0, but that doesn't fix the issues in Chrome and Firefox unfortunately. IE and Edge now work correctly though, which may hold our Windows users over for a bit. See #649 (comment) for details.

arottersman:
I was able to fix the issue for Chrome and Firefox on Windows by forking `Leaflet.Draw` and updating how the library handles its touch events: https://github.com/arottersman/Leaflet.draw 

Any thoughts on the best way to incorporate the update into Model My Watershed if we don't want to wait for `Leaflet.draw` proper to accept a PR of my change?

mmcfarland:
I think the best way to move forward on this is to create a PR which points the MMW project to a branch on your `L.Draw` fork, which we could then go about testing.  I'm curious, for instance, to see if your changes are compatible with a touch-only device like an iPad.

We used a leaflet draw fork for a while and only recently switched back when they incorporated some draw fixes - we can switch to a new fork again for the app and work through getting a PR accepted to the main project.

[...]

PR #1421, opened 2016-08-15
Upgrade Leaflet Draw to 0.3.0

rajadain:
## Overview

Improves touch behavior on Windows touchscreens in Edge and Internet Explorer 11. Chrome still has issues when using the mouse, and Firefox when using the touchscreen.

## Testing Instructions

Test drawing a shape on various browsers on a Windows touchscreen device, using both the mouse and touchscreen.

Connects #649

### DIFF
--- a/src/mmw/npm-shrinkwrap.json
+++ b/src/mmw/npm-shrinkwrap.json
@@ -1506,9 +1506,9 @@
       "resolved": "https://registry.npmjs.org/leaflet/-/leaflet-0.7.3.tgz"
     },
     "leaflet-draw": {
-      "version": "0.2.3",
-      "from": "../../tmp/npm-20250-abe6eb52/1438892615591-0.05360055458731949/e753b7d0eb9d1de8864ca85d5dafbf9ac955759c",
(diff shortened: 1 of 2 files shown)


mmcfarland:
> Chrome still has issues when using the mouse, and Firefox when using the touchscreen.

Sounds like we at least have a workaround for those browsers.  We should recommend Chrome-> touch; FF -> mouse ?

rajadain:
Yes, but I think that was already working. What is most worrisome are the systems described in #649, where the user's computer seems to have enough touch support to mess up Chrome, but not enough for them to actually use their fingers.

rajadain:
Testing on a Windows Touch device, it is clear that Edge works correctly in both cases. And even Firefox works correctly in both cases. It's just that the vertex boxes are much smaller in Firefox than they are in Edge and Chrome, which makes it a harder target to hit with a finger. Somehow Edge and Chrome can detect if the device is a touch screen and make the vertex boxes bigger. If one is careful and intentional in Firefox, one can use both mouse and touch correctly.

So this only leaves mouse misbehavior in Chrome. I'm going to give this a quick shot and see if we can fix it ourselves.

rajadain:
This work has brought us 80% of the way there, with the only missing use case being Chrome on Windows. I say we merge this, and point #1190 in the next sprint for investigating that specific case.

Issue #1415, opened 2016-08-11
Release gwlfe and upgrade MMW


mmcfarland:
After gwlfe bugfixes are merged in, cut a release and upgrade MMW so that client can evaluate changes in model run and GMS creation.

PR #1422, opened 2016-08-15
Upgrade to GWLF-E 0.6.0

kellyi:
This PR updates the app's requirements file to use GWLF-E 0.6.0, up from 0.5.0.

**Testing**
- grab this branch, `vagrant provision worker`, then `vagrant up`
- visit `localhost:8000` and run some mapshed jobs and verify that everything works as it should.

Connects #1415


rajadain:
+1 tested, seems to work correctly.

Also note that one should reprovision `app` in addition to `worker`, since GWLF-E is used by `app` to generate the GMS file.

kellyi:
Ah, that makes sense. Thanks!

PR #1423, opened 2016-08-16
Use Precalculated LS, C, and P for Agricultural Land Uses

rajadain:
## Overview

Original MapShed uses a Flow Accumulation method to calculate the LS factor for a given area of interest. However, due to missing data, we have been using the Stream Flow method. Unfortunately, this is less accurate, especially for agricultural land uses such as Hay/Pasture and Cropland.

To alleviate this, Dr. Barry Evans has provided us with an updated county data file, which we import to the `ms_county_animals` table, that includes these new columns:

+    `crop_ls` numeric(6,3),
+    `hp_ls` numeric(6,3),
+    `crop_c` numeric(6,3),
+    `crop_p` numeric(6,3),
+    `hp_c` numeric(6,3),
+    `hp_p` numeric(6,3)

Which correspond to the LS, C and P factor values for Hay/Pasture and Cropland for each county.

Our new approach is to calculated the area-weighted average of each of these values for all counties in an area of interest, and use them instead of the regular calculations. Regular calculations are still used for all other land use types.

## Testing Instructions

Choose a standard shape, such as a HUC-12, on `develop`, create a MapShed project and export a GMS file from the Current Conditions scenario.

Check out this branch. Update the `ms_county_animals` dataset to include the new columns:

```bash
$ vagrant ssh app -c 'cd /vagrant && ./scripts/aws/setupdb.sh -f ms_county_animals.sql.gz'
```

Run Celery in the debugger to make sure it picks up new changes:

```bash
$ ./scripts/debugcelery.sh
```

Create a new MapShed project with the same standard shape as before, and export another GMS file from the Current Conditions scenario. You should see differences in the LS, C, and P values of Hay/Pasture and Cropland:

```diff
--- Little_Neshaminy_develop_20160816__Current_Conditions.gms	2016-08-16 13:32:14.000000000 -0400
+++ Little_Neshaminy_with_Corrected_LS_C_P__Current_Conditions.gms	2016-08-16 15:51:56.000000000 -0400
@@ -17,8 +17,8 @@
 Oct,1.0457270621091717,10.8,0,0.091,0,0,1.385
 Nov,1.0087081300568108,9.7,0,0.091,0,0,1.385
 Dec,0.9872371494664416,9.1,0,0.091,0,0,1.385
-Hay/Past,803.1,75,0.2714132968948771,0.22836631257618079,0.03,0.45
-Cropland,571.1,82,0.27604142433856804,0.22735773644384383,0.42,0.45
+Hay/Past,803.1,75,0.2714132968948771,0.849484581102515,0.01,1
+Cropland,571.1,82,0.27604142433856804,0.738799109654387,0.258101173614922,0.773079185225836
 Forest,2071.0,73,0.27157785493664266,0.2151564164245495,0.002,0.45
 Wetland,364.7,88.5,0.31307519406197576,0.20441617849311078,0.01,0.1
 Disturbed,0.0,0,0,0,0.08,0.1
```

## Deployment Notes

The `ms_county_animals` table will have to be reloaded on Staging and Production whenever this is deployed. This should be very straightforward as nothing links to that table, and the `DROP` will be handled by the script itself. Run the deployment equivalent of this Vagrant command:

```bash
$ vagrant ssh app -c 'cd /vagrant && ./scripts/aws/setupdb.sh -f ms_county_animals.sql.gz'
```

Connects #1418

### DIFF
--- a/src/mmw/apps/modeling/mapshed/calcs.py
+++ b/src/mmw/apps/modeling/mapshed/calcs.py
@@ -254,7 +254,7 @@ def ag_ls_c_p(geom):
         return ag_lscp(*(cursor.fetchone()))
 
 
-def ls_factors(lu_strms, total_strm_len, areas, avg_slope):
+def ls_factors(lu_strms, total_strm_len, areas, avg_slope, ag_lscp):
     results = [0.0] * len(lu_strms)
     if 0 <= avg_slope <= 1.0:
         m = 0.2
@@ -265,7 +265,10 @@ def ls_factors(lu_strms, total_strm_len, areas, avg_slope):
     else:
         m = 0.5
 
-    for i in range(len(lu_strms)):
+    results[0] = ag_lscp.hp_ls
+    results[1] = ag_lscp.crop_ls
+
+    for i in xrange(2, 16):
         results[i] = (ls_factor(lu_strms[i] * total_strm_len * KM_PER_M,
                       areas[i], avg_slope, m))
 
@@ -639,7 +642,7 @@ def sed_a_factor(landuse_pct_vals, cn, AEU, AvKF, AvSlope):
             (0.000001 * AvSlope) - 0.000036)
 
 
-def p_factors(avg_slope):
+def p_factors(avg_slope, ag_lscp):
     """
     Given the average slope, calculates the P Factor for rural land use types.
 
@@ -657,8 +660,8 @@ def p_factors(avg_slope):
         ag_p = 0.74
 
     return [
-        ag_p,  # Hay/Pasture
-        ag_p,  # Cropland
+        ag_lscp.hp_p,    # Hay/Pasture
+        ag_lscp.crop_p,  # Cropland
         ag_p,  # Forest
         0.1,   # Wetland
         0.1,   # Disturbed
--- a/src/mmw/apps/modeling/mapshed/calcs.py
+++ b/src/mmw/apps/modeling/mapshed/calcs.py
@@ -213,6 +213,47 @@ def manure_spread(aeu):
     return [n_spread] * num_land_uses, [p_spread] * num_land_uses
 
 
+def ag_ls_c_p(geom):
+    """
+    Given a geometry, calculates the area-weighted average value of LS, C, and
+    P factors for agriculatural land use tyeps within the geometry, namely
+    Hay/Pasture and Cropland.
+    """
+    sql = '''
+          WITH clipped_counties AS (
+              SELECT ST_Intersection(geom,
+                                     ST_SetSRID(ST_GeomFromText(%s),
+                                                4326)) AS geom_clipped,
+                     ms_county_animals.*
+              FROM ms_county_animals
+              WHERE ST_Intersects(geom,
+                                  ST_SetSRID(ST_GeomFromText(%s),
+                                             4326))
+          ), clipped_counties_with_area AS (
+              SELECT ST_Area(geom_clipped) /
+                     ST_Area(ST_SetSRID(ST_GeomFromText(%s),
+                                        4326)) AS clip_percent,
+                     clipped_counties.*
+              FROM clipped_counties
+          )
+          SELECT SUM(hp_ls * clip_percent) AS hp_ls,
+                 SUM(hp_c * clip_percent) AS hp_c,
+                 SUM(hp_p * clip_percent) AS hp_p,
+                 SUM(crop_ls * clip_percent) AS crop_ls,
+                 SUM(crop_c * clip_percent) AS crop_c,
+                 SUM(crop_p * clip_percent) AS crop_p
+          FROM clipped_counties_with_area;
+          '''
+
+    with connection.cursor() as cursor:
+        cursor.execute(sql, [geom.wkt, geom.wkt, geom.wkt])
+
+        ag_lscp = namedtuple('Ag_LS_C_P',
+                             [col[0] for col in cursor.description])
+
+        return ag_lscp(*(cursor.fetchone()))
+
+
 def ls_factors(lu_strms, total_strm_len, areas, avg_slope):
     results = [0.0] * len(lu_strms)
     if 0 <= avg_slope <= 1.0:
--- a/src/mmw/apps/modeling/mapshed/tasks.py
+++ b/src/mmw/apps/modeling/mapshed/tasks.py
@@ -19,6 +19,7 @@ from apps.modeling.mapshed.calcs import (day_lengths,
                                          et_adjustment,
                                          kv_coefficient,
                                          animal_energy_units,
+                                         ag_ls_c_p,
                                          ls_factors,
                                          p_factors,
                                          manure_spread,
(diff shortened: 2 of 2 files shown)


kellyi:
Loading the boundaries on my machine now so I can make a proper comparison. 

Here are values for these roughly similar AOIs, though:

this is from staging:

screen shot 2016-08-16 at 5 08 10 pm

```txt
Hay/Past,0.0,0,0,0,0.03,0.52
Cropland,0.0,0,0,0,0.42,0.52
```

and this is from this branch:

screen shot 2016-08-16 at 5 09 00 pm

```txt
Hay/Past,0.0,0,0,0.871,0.03,0.99
Cropland,0.0,0,0,0.447,0.25,0.9
```

rajadain:
These numbers correspond to the values in the database, since the area of interest is in only one county:

```sql
WITH clipped_counties AS (
    SELECT ST_Intersection(geom,
                           ST_SetSRID(ST_GeomFromGeoJSON('{"type":"MultiPolygon","coordinates":[[[[-75.19951694119622,40.00178862327482],[-75.18777973543952,40.00178862327482],[-75.18777973543952,40.01077901704747],[-75.19951694119622,40.01077901704747],[-75.19951694119622,40.00178862327482]]]]}'),
                                      4326)) AS geom_clipped,
           ms_county_animals.*
    FROM ms_county_animals
    WHERE ST_Intersects(geom,
                        ST_SetSRID(ST_GeomFromGeoJSON('{"type":"MultiPolygon","coordinates":[[[[-75.19951694119622,40.00178862327482],[-75.18777973543952,40.00178862327482],[-75.18777973543952,40.01077901704747],[-75.19951694119622,40.01077901704747],[-75.19951694119622,40.00178862327482]]]]}'),
                                   4326))
), clipped_counties_with_area AS (
    SELECT ST_Area(geom_clipped) / ST_Area(ST_SetSRID(ST_GeomFromGeoJSON('{"type":"MultiPolygon","coordinates":[[[[-75.19951694119622,40.00178862327482],[-75.18777973543952,40.00178862327482],[-75.18777973543952,40.01077901704747],[-75.19951694119622,40.01077901704747],[-75.19951694119622,40.00178862327482]]]]}'),
                                                      4326)) AS clip_percent,
           clipped_counties.*
    FROM clipped_counties
)
SELECT cnty_name, clip_percent, hp_ls, hp_c, hp_p, crop_ls, crop_c, crop_p
FROM clipped_counties_with_area;
```

```
  cnty_name   | clip_percent | hp_ls | hp_c  | hp_p  | crop_ls | crop_c | crop_p
--------------+--------------+-------+-------+-------+---------+--------+--------
 Philadelphia |            1 | 0.871 | 0.030 | 0.990 |   0.447 |  0.250 |  0.900
(1 row)
```

kellyi:
Cool. Here's a comparison with PA's 2nd Congressional District:

screen shot 2016-08-16 at 5 16 13 pm


```sh
kinnes@allegheny:~/Downloads    $ diff pull-request-pa-2.txt staging-pa-2.txt
20,21c20,21
< Hay/Past,53.2,63,0.20224982603800848,1.07168152419464,0.0240889094493478,0.992955545275328
< Cropland,7.7,64,0.13614218530710787,0.628470479905024,0.247044454724674,0.805422551189567
---
> Hay/Past,53.2,63,0.20224982603800848,0.2674402175271712,0.03,0.45
> Cropland,7.7,64,0.13614218530710787,0.3024407709542921,0.42,0.45
```

rajadain:
I don't have great intuition for these numbers, so can't say if they seem right or wrong. But they seem to be doing what was asked on paper at least. https://docs.google.com/document/d/16TQjFav-ygEMwqFg4h1zT-SC9Lf-mlJiwBEoiIyv8o0/edit

src/mmw/apps/modeling/mapshed/calcs.py
```diff
@@ -213,6 +213,47 @@ def manure_spread(aeu):
...
+          FROM clipped_counties_with_area;
+          '''
+
+    with connection.cursor() as cursor:
+        cursor.execute(sql, [geom.wkt, geom.wkt, geom.wkt])
+
+        ag_lscp = namedtuple('Ag_LS_C_P',
+                             [col[0] for col in cursor.description])
+
+        return ag_lscp(*(cursor.fetchone()))
```
kellyi: Python q: where do the names `hp_p` and `crop_p` get assigned for this tuple? I see them getting used in the other commit, but wasn't sure whether they were assigned there or here.

rajadain: They are assigned here. The `namedtuple` line above makes essentially a `struct` with the field names corresponding to query, and this line populates them with the right values.

I did this rather than make a dictionary because I preferred to use `ag_lscp.crop_c` rather than `ag_lscp['crop_c']`. We do the same thing with [`WeatherStation`](https://github.com/WikiWatershed/model-my-watershed/blob/41d94d26e51502fa723eadb9edba8585e32a244a/src/mmw/apps/modeling/mapshed/calcs.py#L70-L73) above.

kellyi: Ok, cool. Would it be worth updating the docstring to specify what key-value pairs come out of this function?

rajadain: Sure I'll add it.

rajadain: I can't think of what to write.

```diff
diff --git a/src/mmw/apps/modeling/mapshed/calcs.py b/src/mmw/apps/modeling/mapshed/calcs.py
index ff8d562..bfab354 100644
--- a/src/mmw/apps/modeling/mapshed/calcs.py
+++ b/src/mmw/apps/modeling/mapshed/calcs.py
@@ -218,6 +218,8 @@ def ag_ls_c_p(geom):
     Given a geometry, calculates the area-weighted average value of LS, C, and
     P factors for agriculatural land use tyeps within the geometry, namely
     Hay/Pasture and Cropland.
+
+    Returns a named tuple for easy access.
     """
     sql = '''
           WITH clipped_counties AS (
```

kellyi:
These changes look good to me. The logic in the `ag_ls_c_p` seems to implement what was described in the google doc above, and the output value from the two AOIs above has changed as a result. +1 from me.

I think you mentioned asking BME whether the Little Neshaminy values appeared to be accurate? Should we send values from another known AOI to get another point of comparison?

src/mmw/apps/modeling/mapshed/calcs.py
kellyi: Maybe something like `Returns a named tuple with keys as specified in lines 239-245` (or whatever the new lines are)?

rajadain:
BME did corroborate the values of Little Neshaminy. I think we should get this on Staging so he can play around with it, and then offer more specific suggestions.

src/mmw/apps/modeling/mapshed/calcs.py
rajadain: That should be evident from the code, especially since we're using it in another function nearby ... trying to think of a way to describe it that focuses on the why rather than the what or how.

rajadain:
Someone will have to SSH in to staging and run the script mentioned above to get this to work correctly. I can't do it from Pittsburgh.

Issue #1424, opened 2016-08-18
Include Animal Population summary in analysis results


ajrobbins:
Basic summary statistics for Area of Interest from national county animal population data (by animal type), including total Animal Equivalent Units (AEU), will be included in the Analyze output.

We need to:
- Add a new tab to the Analysis tab, alongside "Land" and "Soil", to read "Livestock". 
- The tab will include the AoI name/title and size, as the land and soil tabs do
- This tab will include a table of: Animal type, AEU amount, and proportion for the area of interest

Questions for @mmcfarland : should the tab also include a bar chart like we have for land/soil:
![screen shot 2016-08-18 at 1 52 34 pm](https://cloud.githubusercontent.com/assets/12868342/17784635/3890404a-654b-11e6-837c-3edbfadbc700.png)

 with proportion of AEUs by animal type?
Also, should the tab be called "Livestock" or something else, like "Animal distribution" or something?

mmcfarland:
We probably have to proportionally weight the county to the portion that exists in the AoI (similarly to what we do for MapShed).

mmcfarland:
Also, this should be unified as much as possible with the existing mapshed query getting animal counts.  If that is not area weighted, check out in-progress changes that deal with new attributes on that dataset which do similar things:
https://github.com/WikiWatershed/model-my-watershed/pull/1423/files

kellyi:
Sounds good. I'll take another look at that PR to see how it's implemented.

mmcfarland:
And just for clarification, the PR is calculating _LS_ from values on `ms_county_animals`, the actual map shed AEU stuff may already have what we want for analysis output.

kellyi:
For reference the MapShed animal count method is here -- https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/mapshed/calcs.py#L149

kellyi:
Dropping this here for reference, too, from a description of the USDA county animals shapefile:

> It provides information that will allow the estimation basin-level populations for 9 different types of farm animals on a per unit area basis. The key fields to be used by MapShed include “Fips_code”, “Beef_ha”, “Dairy_ha”, “Broiler_ha”, “Layer_ha”, “Hogs_ha”, “Goat_ha”, “Horse_ha”, Sheep_ha”, and “Turkey_ha”.

PR #1301, opened 2016-05-12
Include goats in sheep count

rajadain:
From BME:

> Yes, include both goats and sheep in the "Sheep" category. Thanks.

Before:

```
    beef_cows     |     broilers     |    dairy_cows    |      sheep       |       hogs       |      horses      |      layers      |     turkeys      
------------------+------------------+------------------+------------------+------------------+------------------+------------------+------------------
 299.779487357229 | 102110.982701653 | 207.148229899953 | 656.656073848423 | 613.369111513422 | 28837.9543177855 | 10985.3384931695 | 18.8222402678298
(1 row)
```

After:

```
    beef_cows     |     broilers     |    dairy_cows    |      sheep       |       hogs       |      horses      |      layers      |     turkeys      
------------------+------------------+------------------+------------------+------------------+------------------+------------------+------------------
 299.779487357229 | 102110.982701653 | 207.148229899953 | 1271.72268876914 | 613.369111513422 | 28837.9543177855 | 10985.3384931695 | 18.8222402678298
(1 row)
```

PR #1425, opened 2016-08-19
Retrieve and display weighted animal population

kellyi:
This PR updates the "Analyze" tab results to include a livestock tab which displays the weighted animal population presented in terms of "AEU per hectare" and "AEU Proportion" -- e.g., what percentage each animal type comprises of the total animal AEU per hectare.

This works by making a database query with the input aoi to retrieve data from `ms_county_animals`, then appending it to the outgoing analyze results. There are a few commits here:

- the first commit sets up the backend to retrieve and parse the animal data, then append it to the outgoing json results
- the second updates some of the CSS to make the style rules amenable to displaying an analyze tab with just a table and no bar chart
- third handles setting up a model, view, and template for the livestock tab in the front-end code
- fourth is a fixup to remove code which retrieved what I believe is some unnecessary data pulled from the db in the first commit (I'll squash before merging)

The results currently look like this:

screen shot 2016-08-22 at 12 25 54 pm

screen shot 2016-08-22 at 12 26 19 pm

**Testing**
- get this branch, `vagrant up`, `./scripts/bundle.sh`, then open the browser console and visit `localhost:8000`
- select an aoi to analyze
- verify that the results now includes a "Livestock" tab which displays AEU per hectare in one column and a percentage value in the other column
- repeat with a few more aois
- try looking at some of the other pre-existing analyze tabs and also modeling an aoi to verify that nothing has changed with them as a result of these changes
- verify that you haven't seen any errors in the console.

Connects #1424

### DIFF
--- a/src/mmw/apps/modeling/calcs.py
+++ b/src/mmw/apps/modeling/calcs.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import absolute_import
+
+from django.contrib.gis.geos import GEOSGeometry
+
+from django.conf import settings
+
+from apps.modeling.mapshed.calcs import animal_energy_units
+
+ANIMAL_KEYS = settings.GWLFE_CONFIG['AnimalKeys']
+ANIMAL_NAMES = settings.GWLFE_DEFAULTS['AnimalName']
+
+ANIMAL_DISPLAY_NAMES = dict(zip(ANIMAL_KEYS, ANIMAL_NAMES))
+
+
+def animal_population(geojson):
+    """
+    Given a GeoJSON shape, call MapShed's `animal_energy_units` method
+    to calculate the area-weighted county animal population. Returns a
+    dictionary to append to the outgoing JSON for analysis results.
+    """
+    geom = GEOSGeometry(geojson, srid=4326)
+    aeu_for_geom = animal_energy_units(geom)[2]
+    aeu_return_values = []
+
+    for animal, aeu_value in aeu_for_geom.iteritems():
+        aeu_return_values.append({
+            'type': ANIMAL_DISPLAY_NAMES[animal],
+            'aeu': int(round(aeu_value)),
+        })
+
+    return {
+        'displayName': 'Animals',
+        'name': 'animals',
+        'categories': aeu_return_values
+    }
--- a/src/mmw/apps/modeling/mapshed/calcs.py
+++ b/src/mmw/apps/modeling/mapshed/calcs.py
@@ -165,15 +165,15 @@ def animal_energy_units(geom):
                      clipped_counties.*
               FROM clipped_counties
           )
-          SELECT SUM(beef_ha * totalha * clip_percent) AS beef_cows,
-                 SUM(broiler_ha * totalha * clip_percent) AS broilers,
-                 SUM(dairy_ha * totalha * clip_percent) AS dairy_cows,
-                 SUM(goat_ha * totalha * clip_percent) +
-                 SUM(sheep_ha * totalha * clip_percent) AS sheep,
-                 SUM(hog_ha * totalha * clip_percent) AS hogs,
-                 SUM(horse_ha * totalha * clip_percent) AS horses,
-                 SUM(layer_ha * totalha * clip_percent) AS layers,
-                 SUM(turkey_ha * totalha * clip_percent) AS turkeys
+          SELECT SUM(beef_ha * ag_ha * clip_percent) AS beef_cows,
+                 SUM(broiler_ha * ag_ha * clip_percent) AS broilers,
+                 SUM(dairy_ha * ag_ha * clip_percent) AS dairy_cows,
+                 SUM(goat_ha * ag_ha * clip_percent) +
+                 SUM(sheep_ha * ag_ha * clip_percent) AS sheep,
+                 SUM(hog_ha * ag_ha * clip_percent) AS hogs,
+                 SUM(horse_ha * ag_ha * clip_percent) AS horses,
+                 SUM(layer_ha * ag_ha * clip_percent) AS layers,
+                 SUM(turkey_ha * ag_ha * clip_percent) AS turkeys
           FROM clipped_counties_with_area;
           '''
 
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -15,6 +15,8 @@ from celery import shared_task
 from apps.modeling.geoprocessing import histogram_start, histogram_finish, \
     data_to_survey, data_to_censuses
 
+from apps.modeling.calcs import animal_population
+
 from tr55.model import simulate_day
 from gwlfe import gwlfe, parser
 
@@ -102,7 +104,7 @@ def get_histogram_job_results(self, incoming):
 
 
 @shared_task
-def histogram_to_survey(incoming):
+def histogram_to_survey(incoming, aoi):
     """
     Converts the histogram results (aka analyze results)
     to a survey of land use, which are rendered in the UI.
@@ -111,6 +113,7 @@ def histogram_to_survey(incoming):
     data = incoming['histogram'][0]
     results = data_to_survey(data)
     convert_result_areas(pixel_width, results)
+    results.append(animal_population(aoi))
 
     return results
 
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -18,6 +18,8 @@ var $ = require('jquery'),
     aoiHeaderTmpl = require('./templates/aoiHeader.html'),
     tableTmpl = require('./templates/table.html'),
     tableRowTmpl = require('./templates/tableRow.html'),
+    animalTableTmpl = require('./templates/animalTable.html'),
+    animalTableRowTmpl = require('./templates/animalTableRow.html'),
     tabPanelTmpl = require('../modeling/templates/resultsTabPanel.html'),
     tabContentTmpl = require('./templates/tabContent.html'),
     barChartTmpl = require('../core/templates/barChart.html'),
@@ -244,7 +246,9 @@ var TabContentView = Marionette.LayoutView.extend({
             units = utils.magnitudeOfArea(largestArea),
             census = this.model.get('name') === 'land' ?
                 new coreModels.LandUseCensusCollection(categories) :
-                new coreModels.SoilCensusCollection(categories);
+                this.model.get('name') === 'soil' ?
+                new coreModels.SoilCensusCollection(categories) :
+                new coreModels.AnimalCensusCollection(categories);
 
         this.aoiRegion.show(new AoiView({
             model: new coreModels.GeoModel({
@@ -253,15 +257,22 @@ var TabContentView = Marionette.LayoutView.extend({
             })
         }));
 
-        this.tableRegion.show(new TableView({
-            units: units,
-            collection: census
-        }));
+        if (this.model.get('name') === 'animals') {
+            this.tableRegion.show(new AnimalTableView({
+                units: units,
+                collection: census
+            }));
+        } else {
+            this.tableRegion.show(new TableView({
+                units: units,
+                collection: census
+            }));
 
-        this.chartRegion.show(new ChartView({
-            model: this.model,
-            collection: census
-        }));
+            this.chartRegion.show(new ChartView({
+                model: this.model,
+                collection: census
+            }));
+        }
     }
 });
 
@@ -313,6 +324,35 @@ var TableView = Marionette.CompositeView.extend({
     }
 });
 
+var AnimalTableRowView = Marionette.ItemView.extend({
+    tagName: 'tr',
+    template: animalTableRowTmpl,
+    templateHelpers: function() {
+        return {
+            aeu: this.model.get('aeu'),
+        };
+    }
+});
+
+var AnimalTableView = Marionette.CompositeView.extend({
+    childView: AnimalTableRowView,
+    childViewOptions: function() {
+        return {
+            units: this.options.units
+        };
+    },
+    templateHelpers: function() {
+        return {
+            headerUnits: this.options.units
+        };
+    },
+    childViewContainer: 'tbody',
+    template: animalTableTmpl,
+
+    onAttach: function() {
+        $('[data-toggle="table"]').bootstrapTable();
+    }
+});
 
 var ChartView = Marionette.ItemView.extend({
     template: barChartTmpl,
--- a/src/mmw/apps/modeling/views.py
+++ b/src/mmw/apps/modeling/views.py
@@ -364,7 +364,7 @@ def _initiate_analyze_job_chain(area_of_interest, job_id, testing=False):
                  .set(exchange=exchange, routing_key=routing_key),
                  tasks.get_histogram_job_results.s()
                  .set(exchange=exchange, routing_key=routing_key),
-                 tasks.histogram_to_survey.s()
+                 tasks.histogram_to_survey.s(area_of_interest)
                  .set(exchange=exchange, routing_key=choose_worker()),
                  save_job_result.s(job_id, area_of_interest)
                  .set(exchange=exchange, routing_key=choose_worker())) \
--- a/src/mmw/js/src/analyze/templates/animalTable.html
+++ b/src/mmw/js/src/analyze/templates/animalTable.html
@@ -1,0 +1,12 @@
+<table class="table custom-hover" data-toggle="table">
+    <thead>
+        <tr>
+            <th data-sortable="true">Animal</th>
+            <th class="text-right" data-sortable="true" data-sorter="window.numericSort">
+                Count
+            </th>
+        </tr>
+    </thead>
+    <tbody>
+    </tbody>
+</table>
--- a/src/mmw/js/src/analyze/templates/animalTableRow.html
+++ b/src/mmw/js/src/analyze/templates/animalTableRow.html
@@ -1,0 +1,2 @@
+<td>{{ type }}</td>
+<td class="strong text-right">{{ aeu|toLocaleString() }}</td>
--- a/src/mmw/js/src/core/models.js
+++ b/src/mmw/js/src/core/models.js
@@ -245,6 +245,10 @@ var SoilCensusCollection = Backbone.Collection.extend({
     comparator: 'code'
 });
 
+var AnimalCensusCollection = Backbone.Collection.extend({
+    comparator: 'type'
+});
+
 var GeoModel = Backbone.Model.extend({
     M_IN_KM: 1000000,
 
@@ -298,6 +302,7 @@ module.exports = {
     TaskMessageViewModel: TaskMessageViewModel,
     LandUseCensusCollection: LandUseCensusCollection,
     SoilCensusCollection: SoilCensusCollection,
+    AnimalCensusCollection: AnimalCensusCollection,
     GeoModel: GeoModel,
     AreaOfInterestModel: AreaOfInterestModel,
     AppStateModel: AppStateModel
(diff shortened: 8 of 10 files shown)


src/mmw/apps/modeling/tasks.py
```diff
@@ -94,10 +98,12 @@ def get_histogram_job_results(self, incoming):
     """
     pixel_width = incoming['pixel_width']
     histogram = histogram_finish(incoming['sjs_job_id'], self.retry)
+    aoi = incoming['aoi']
 
     return {
         'pixel_width': pixel_width,
-        'histogram': histogram
+        'histogram': histogram,
+        'aoi': aoi
```
rajadain: Instead of thread the area of interest through `start_histogram_job` and `get_histogram_job_results`, just to have them available in `histogram_to_survey`, consider giving it directly to that function:

```diff
diff --git a/src/mmw/apps/modeling/tasks.py b/src/mmw/apps/modeling/tasks.py
index 0ac6b7e..beb3c89 100644
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -108,7 +108,7 @@ def get_histogram_job_results(self, incoming):
 
 
 @shared_task
-def histogram_to_survey(incoming):
+def histogram_to_survey(incoming, aoi):
     """
     Converts the histogram results (aka analyze results)
     to a survey of land use, which are rendered in the UI.
@@ -117,7 +117,7 @@ def histogram_to_survey(incoming):
     data = incoming['histogram'][0]
     results = data_to_survey(data)
     convert_result_areas(pixel_width, results)
-    results.append(animal_population(incoming['aoi']))
+    results.append(animal_population(aoi))
 
     return results
 
diff --git a/src/mmw/apps/modeling/views.py b/src/mmw/apps/modeling/views.py
index 36bd598..5f1390e 100644
--- a/src/mmw/apps/modeling/views.py
+++ b/src/mmw/apps/modeling/views.py
@@ -364,7 +364,7 @@ def _initiate_analyze_job_chain(area_of_interest, job_id, testing=False):
                  .set(exchange=exchange, routing_key=routing_key),
                  tasks.get_histogram_job_results.s()
                  .set(exchange=exchange, routing_key=routing_key),
-                 tasks.histogram_to_survey.s()
+                 tasks.histogram_to_survey.s(area_of_interest)
                  .set(exchange=exchange, routing_key=choose_worker()),
                  save_job_result.s(job_id, area_of_interest)
                  .set(exchange=exchange, routing_key=choose_worker())) \
```

src/mmw/js/src/analyze/views.js
```diff
@@ -253,15 +257,22 @@ var TabContentView = Marionette.LayoutView.extend({
             })
         }));
 
-        this.tableRegion.show(new TableView({
-            units: units,
-            collection: census
-        }));
+        if (this.model.get('name') !== 'livestock') {
```
rajadain: Consider reversing this `if` block so there is a more direct relationship between the condition and the body:

```javascript
if (this.model.get('name') === 'livestock') {
  // new LivestockTableView
} else {
  // new ChartView
  // new TableView
}
```

mmcfarland:
I'll have some feedback on the output format for this stuff this afternoon, but I don't think the numbers we have in this table are what we'll end up needing.

src/mmw/js/src/analyze/views.js
```diff
@@ -313,6 +324,37 @@ var TableView = Marionette.CompositeView.extend({
     }
 });
 
+var LivestockTableRowView = Marionette.ItemView.extend({
+    tagName: 'tr',
+    template: livestockTableRowTmpl,
+    templateHelpers: function() {
+        return {
+            // Convert proportion to percentage for display.
+            proportion: (this.model.get('proportion') * 100),
```
rajadain: Is it possible to use [`toLocaleString`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) here instead? Like we do in Program Analysis https://github.com/azavea/usace-program-analysis/blob/f9d891aa8af408998e44f7bd7709250ccc3e512e/src/usace_program_analysis/app/js/src/helpers/index.js#L69-L73

Although it should be noted that `toLocaleString` is not supported in Safari, so perhaps we're better off using this.

src/mmw/apps/modeling/calcs.py
```diff
@@ -0,0 +1,107 @@
...
+    with connection.cursor() as cursor:
+        cursor.execute(sql, [geom.wkt, geom.wkt, geom.wkt])
+
+        animal_pop = namedtuple('AnimalPopulation',
+                                [col[0] for col in cursor.description])
+
+        animal_data = {}
+
+        for k, v in animal_pop(*(cursor.fetchone()))._asdict().iteritems():
+            animal_data[k.decode('utf-8')] = v
```
rajadain: Instead of transforming the query results into a named tuple and then into a dictionary, consider transforming directly to a dictionary as done here: https://github.com/WikiWatershed/model-my-watershed/blob/9cc43eff1dc8fdebf7a07c7ee6c76ecaf8dd2ace/src/mmw/apps/modeling/mapshed/calcs.py#L183-L186

src/mmw/apps/modeling/calcs.py
```diff
@@ -0,0 +1,107 @@
...
+        animal_fields = {
+            u'beef_ha': u'Beef',
+            u'dairy_ha': u'Dairy',
+            u'broiler_ha': u'Broilers',
+            u'goat_ha': u'Goats',
+            u'hog_ha': u'Hogs',
+            u'horse_ha': u'Horses',
+            u'layer_ha': u'Layers',
+            u'sheep_ha': u'Sheep',
+            u'turkey_ha': u'Turkeys'
```
rajadain: Instead of using this dictionary mapping, use the same keys as here: https://github.com/WikiWatershed/model-my-watershed/blob/9cc43eff1dc8fdebf7a07c7ee6c76ecaf8dd2ace/src/mmw/apps/modeling/mapshed/calcs.py#L168-L176 in your SQL query, and [`ANIMAL_KEYS`](https://github.com/WikiWatershed/model-my-watershed/blob/77ca09747886a09dd8d07113d68d63d7dd115727/src/mmw/mmw/settings/gwlfe_settings.py#L459-L460) and [`AnimalName`](https://github.com/WikiWatershed/model-my-watershed/blob/77ca09747886a09dd8d07113d68d63d7dd115727/src/mmw/mmw/settings/gwlfe_settings.py#L406) to map from the keys to display names. This way they'll remain standard without duplication.

src/mmw/apps/modeling/calcs.py
```diff
@@ -0,0 +1,107 @@
...
+                u'aeu': animal_data[code]
+            }
+
+        animal_coverage = []
+
+        for code, name in animal_fields.iteritems():
+            animal_coverage.append(coverage_for_animal(code, name))
+
+        return {
+            u'displayName': 'Livestock',
```
rajadain: I don't think we need to include the `u` prefix if our keys don't have any special unicode characters.

rajadain:
Also consider adding some sort of total to the Livestock tab in the Analyze sidebar. Will take another look once we have more instruction on what numbers are expected here.

ajrobbins:
I guessed wrong, the tab should be called "Animals" :(
@mmcfarland has further details on the AEU figures.

kellyi:
Cool. Thanks for checking! Going to make the changes @rajadain suggested and prepare it to be updated with the new values.

kellyi:
I pushed an update which revises this to use the existing MapShed implementation of `animal_energy_units` to retrieve the population data, then uses a wrapper function to prepare that data to be sent back to the front-end.

The table now looks like this on the front-end:

screen shot 2016-08-23 at 5 49 48 pm

src/mmw/apps/modeling/mapshed/calcs.py
```diff
@@ -165,15 +165,15 @@ def animal_energy_units(geom):
...
-                 SUM(goat_ha * totalha * clip_percent) +
-                 SUM(sheep_ha * totalha * clip_percent) AS sheep,
-                 SUM(hog_ha * totalha * clip_percent) AS hogs,
-                 SUM(horse_ha * totalha * clip_percent) AS horses,
-                 SUM(layer_ha * totalha * clip_percent) AS layers,
-                 SUM(turkey_ha * totalha * clip_percent) AS turkeys
+          SELECT SUM(beef_ha * ag_ha * clip_percent) AS beef_cows,
+                 SUM(broiler_ha * ag_ha * clip_percent) AS broilers,
+                 SUM(dairy_ha * ag_ha * clip_percent) AS dairy_cows,
+                 SUM(goat_ha * ag_ha * clip_percent) +
```
kellyi: Was wondering whether this `+` here is a typo, but it looks like the app does not specify `goats` in GWLF-E settings -- https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/gwlfe_settings.py#L467

rajadain: See #1301

kellyi:
For reference, here is the difference in animal values for two known AOIs when using `totalha` and `ag_ha`, respectively:

`totalha`:
screen shot 2016-08-23 at 6 12 21 pm

`ag_ha`:
screen shot 2016-08-23 at 6 10 50 pm

src/mmw/apps/modeling/mapshed/calcs.py
kellyi: :goat:

src/mmw/apps/modeling/tasks.py
rajadain: Since `area_of_interest` has been added to `histogram_to_survey`, all changes to `get_histogram_job_results` and `start_histogram_job` can be reverted, as they are no longer needed.

src/mmw/apps/modeling/calcs.py
```diff
@@ -0,0 +1,43 @@
...
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import absolute_import
+
+from django.contrib.gis.geos import GEOSGeometry
+
+from apps.modeling.mapshed.calcs import animal_energy_units
+
+ANIMAL_DISPLAY_NAMES = {
+    'sheep': 'Sheep',
```
rajadain: Instead of recreating the mapping here, prefer reusing existing definitions:

```python
from django.conf import settings


ANIMAL_KEYS = settings.GWLFE_CONFIG['AnimalKeys']
ANIMAL_NAMES = settings.GWLFE_DEFAULTS['AnimalName']

ANIMAL_DISPLAY_NAMES = dict(zip(ANIMAL_KEYS, ANIMAL_NAMES))
```

src/mmw/apps/modeling/calcs.py
```diff
@@ -0,0 +1,43 @@
...
+    for animal, aeu_value in aeu_for_geom.iteritems():
+        animal_dictionary = {
+            'type': ANIMAL_DISPLAY_NAMES[animal],
+            'aeu': aeu_value,
+        }
+        aeu_return_values.append(animal_dictionary)
+
+    return {
+        u'displayName': 'Animals',
+        u'name': 'livestock',
```
rajadain: Consider renaming `livestock` to `animals` here too, so that they name and display name stay in sync and are easier to reason about. Best do this now before it has a chance to permeate the system.

rajadain:
Here's a difference in GMS file between `TotalHa` (current behavior) and `AgHa` (new behavior):

```diff
> diff -u Little_Neshaminy_with_TotalHa__Current_Conditions.gms Little_Neshaminy_with_AgHa__Current_Conditions.gms
--- Little_Neshaminy_with_TotalHa__Current_Conditions.gms	2016-08-23 18:55:47.000000000 -0400
+++ Little_Neshaminy_with_AgHa__Current_Conditions.gms	2016-08-23 18:49:07.000000000 -0400
@@ -1,5 +1,5 @@
 10,6,0
-1.4.0,0.06,0,10,0,0,0.11065935189270064,9.937211446569115,73797,3495.1217861975642,70301.87821380244,227.27408659695075,0,2.844365174567346,0.05559663102442437,30,1961,1990,0.0032336808039321995,11158.662599444428,0,0,0,0.25355267653906527
+1.4.0,0.06,0,10,0,0,0.11065935189270064,9.937211446569115,73797,3495.1217861975642,70301.87821380244,227.27408659695075,0,2.844365174567346,0.0072895778446896925,30,1961,1990,0.0031919918170380886,11158.662599444428,0,0,0,0.25355267653906527
 0
 0
 0
@@ -107,7 +107,7 @@
 0,0,0,0,0,0,0,0,0,0,0,0,0
 0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0
 0,0,0,0,0,0,0,0,0
-0.16538940197812274,0,0,0,0,0,0,0,0,1533,2,1535,3.5,73.8,0,0,0,0,0,0,0,0,0,0,0,0
+0.16538940197812274,0,0,0,0,0,0,0,0,201,0,201,3.5,73.8,0,0,0,0,0,0,0,0,0,0,0,0
 0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0
 0,0,0,0,1.165040595399188,18.440928281461435,0,0,0,0,0,0,0,0,0,0
 0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0
@@ -145,14 +145,14 @@
 0,0,0,0,0,0,0,0,0,0,0
 2373,785,3380000000,0.55,0.28,0.83,2373,785,3380000000,0.52,0.18,1,1
 500000000,25,0.9,9600,2000000000,0,200,0.5,0,0,0,0
-Dairy Cows,658,Y,640,0.44,0.07,100000000000
-Beef Cows,514,Y,360,0.31,0.09,100000000000
-Broilers,586,N,0.9,1.07,0.3,140000000
+Dairy Cows,91,Y,640,0.44,0.07,100000000000
+Beef Cows,64,Y,360,0.31,0.09,100000000000
+Broilers,80,N,0.9,1.07,0.3,140000000
 Layers,0,N,1.8,0.85,0.29,140000000
-Hogs/Swine,1171,N,61,0.48,0.15,11000000000
-Sheep,1223,Y,50,0.37,0.1,12000000000
-Horses,1588,Y,500,0.28,0.06,420000000
-Turkeys,248,N,6.8,0.59,0.2,95000000
+Hogs/Swine,125,N,61,0.48,0.15,11000000000
+Sheep,162,Y,50,0.37,0.1,12000000000
+Horses,207,Y,500,0.28,0.06,420000000
+Turkeys,37,N,6.8,0.59,0.2,95000000
 Other,0,N,0,0,0,0
 Jan,0.01,0.05,0.07,0.12,0,0.2,0.2,0.12
 Feb,0.01,0.05,0.07,0.12,0,0.2,0.2,0.12
```

 * [Little_Neshaminy_with_TotalHa__Current_Conditions.gms.txt](https://github.com/WikiWatershed/model-my-watershed/files/433718/Little_Neshaminy_with_TotalHa__Current_Conditions.gms.txt)
 * [Little_Neshaminy_with_AgHa__Current_Conditions.gms.txt](https://github.com/WikiWatershed/model-my-watershed/files/433719/Little_Neshaminy_with_AgHa__Current_Conditions.gms.txt)

mmcfarland:
Screenshot looks good, a few changes:
* Change "Type" to "Animal" and "AEU" to "Count" (we're no longer showing equivilent units)
* Round the Count column to an integer so we're not implying partial animals

src/mmw/apps/modeling/calcs.py
kellyi: Done! Also changed `livestock` -> `animals` in the frontend code added here.

kellyi:
Sounds good! I just pushed fixup commits for the backend and the frontend. 

Backend changes remove the superfluous passing of `aoi` around in `tasks.py`, use GWLFE defaults for the animal names, round the outgoing animal count for this display then convert to an int, and rename `livestock` -> `animals`. Frontend changes also rename `livestock` -> `animals` and update the table headers to read "Animal" and "Count" respectively. Results now look like this:

screen shot 2016-08-24 at 9 42 49 am

src/mmw/apps/modeling/calcs.py
```diff
@@ -0,0 +1,39 @@
...
+
+    for animal, aeu_value in aeu_for_geom.iteritems():
+        animal_dictionary = {
+            'type': ANIMAL_DISPLAY_NAMES[animal],
+            'aeu': int(round(aeu_value)),
+        }
+        aeu_return_values.append(animal_dictionary)
+
+    return {
+        u'displayName': 'Animals',
```
rajadain: I don't think we need the `u` prefix here. It'll still work without this.

src/mmw/apps/modeling/tasks.py
```diff
@@ -56,15 +58,17 @@ def start_histogram_job(self, json_polygon):
     kick off the SJS job to generate a histogram
     of the provided polygon (i.e. AoI).
     Returns the id of the job so we can poll later
-    for the results, as well as the pixel width of
+    for the results, the pixel width of
     the polygon so that areas can be calculated from
-    the results.
+    the results, and the input json_polygon to use to
```
rajadain: This comment should also be reverted.

src/mmw/apps/modeling/tasks.py
```diff
@@ -56,15 +58,17 @@ def start_histogram_job(self, json_polygon):
...
+    retrieve animal population data from the db.
     """
     polygon = json.loads(json_polygon)
 
     return {
         'pixel_width': aoi_resolution(polygon),
-        'sjs_job_id': histogram_start([json_polygon], self.retry)
+        'sjs_job_id': histogram_start([json_polygon], self.retry),
+        'aoi': json_polygon
```
rajadain: I'm still seeing `json_polygon` being returned here, when it doesn't need to be.

mmcfarland:
Another table formatting issue, can we add comma's to the count values?

kellyi:
Pushed another fixup which reverts the docstring, drops the unicode prefix for the outgoing `animal_population` results, and removes the `aoi` from the `start_histogram_job` return dictionary.

kellyi:
> Another table formatting issue, can we add comma's to the count values?

Sure!

kellyi:
Output formatted with commas:

screen shot 2016-08-24 at 10 04 46 am

src/mmw/apps/modeling/tasks.py
```diff
@@ -64,7 +66,7 @@ def start_histogram_job(self, json_polygon):
 
     return {
         'pixel_width': aoi_resolution(polygon),
-        'sjs_job_id': histogram_start([json_polygon], self.retry)
+        'sjs_job_id': histogram_start([json_polygon], self.retry),
```
rajadain: Sorry to be so persistent here, but we don't need this extra comma here.

Essentially what I'm trying to do is look at all the lines changed in this PR and ask if this line _must_ change for the purpose of this PR. If not, even though it might be a good idea (like having trailing commas in objects, which is objectively a good idea), I would prefer to not include it. This allows PRs to be atomic and isolated, because they have all the changes they need and none of the changes they don't.

src/mmw/apps/modeling/calcs.py
```diff
@@ -0,0 +1,39 @@
...
+    geom = GEOSGeometry(geojson, srid=4326)
+    aeu_for_geom = animal_energy_units(geom)[2]
+    aeu_return_values = []
+
+    for animal, aeu_value in aeu_for_geom.iteritems():
+        animal_dictionary = {
+            'type': ANIMAL_DISPLAY_NAMES[animal],
+            'aeu': int(round(aeu_value)),
+        }
+        aeu_return_values.append(animal_dictionary)
```
rajadain: Consider inlining `animal_dictionary`.

```python
aeu_return_values.append({
  'type': ...
  'aeu': ...
})
```

kellyi: Done!

rajadain:
+1 code looks very neat. Thanks for sticking through this.

kellyi:
Thanks for all your help with this! Will squash before merging.

Issue #1411, opened 2016-08-09
Create zoom dependent DRB stream styling


mmcfarland:
This is related to #1409, but may or may not have the same rules since the layer has restricted zoom levels.  The DRB streams have an `order` attribute which will allow stream order dependent rendering.

Issue #1409, opened 2016-08-09
Create zoom dependent NHD stream styling


mmcfarland:
Exact zoom levels TBD, but when #1408 is available, render streams appropriate for that the order and zoom level.

PR #1427, opened 2016-08-23
Enable zoom/attribute dependent stream rendering

mmcfarland:
#### Overview
For both NHD Plus and DRB streams, configure the cartographic styling
to limit streams of a high Stream Order (scale is 10-0, large to small)
to be visible at low zoom levels (high altitude).  As the zoom level
increases, more streams of lower stream order are shown and the stroke
width of the rendered line corresponds to the relative stream order.
The effect should be a gradual increase in detail of streams as the
user zooms in, and streams of a higher Stream Order should appear
larger than those of a smaller order.

The intent isn't to make the stream order attribute apparent from
viewing the stream, but to give a relative indication of size and limit
the number of streams shown when zoomed out.

Since the DRB and NHD streams vary considerably in density of streams
at every zoom level, some of the sizes were adjusted to accommodate a
reasonable appearance for both datasets while keeping the width of the
stream the same, per stream order, per dataset.

Connects #1409 
Connects #1411 

#### Testing
* Restart your tile and app server to ensure that the latest styling changes and zoom configurations are present
  * `vagrant ssh tiler -c "sudo service mmw-tiler restart"`
  * `vagrant ssh app -c "sudo service mmw-app restart"`
* Starting at zoom 7, enable the National Stream overlay (this is not fast, since it's doing a large geographic query, but will be cached in stage/prod)
* Zoom in, noticing that more "smaller" streams appear the closer you get
* Do the same in the DRB, starting at zoom 10

#### Code evaluation
It will be helpful to understand that in CartoCSS `[][]` is a logical `AND` and `[], []` is a logical `OR`.

This may not be the final set of zoom/attribute breaks, I'm sure the client will have feedback for a desired distribution.

### DIFF
--- a/src/tiler/server.js
+++ b/src/tiler/server.js
@@ -33,8 +33,8 @@ var interactivity = {
         huc8: 'boundary_huc08',
         huc10: 'boundary_huc10',
         huc12: 'boundary_huc12',
-        'drb_streams': 'drb_streams_50',
-        stream: 'nhdflowline',
+        drb_streams_v1: 'drb_streams_50',
+        nhd_streams_v1: 'nhdflowline',
         municipalities: 'dep_municipalities',
         urban_areas: 'dep_urban_areas'
     },
--- a/src/tiler/server.js
+++ b/src/tiler/server.js
@@ -42,6 +42,29 @@ var interactivity = {
         // Caching can happen if the bucket to write to is defined
         // and the request is not coming from localhost.
         return req.headers.host!== 'localhost' && tileCacheBucket;
+    },
+    getSqlForStreamByReq = function(req) {
+        /* Limit the number of stream features returned at certain zoom levels.
+         * Stream rendering is a function of stream_order in styles.mss.  Lower
+         * number zoom levels only render higher order stream_orders.  Changing
+         * these filters may impact the style definitions. Reducing the number
+         * of features returned by the query greatly increase overall request
+         * performance.
+        */
+        zoom = req.params.z;
+        tableName = req.params.table;
+        stream_order = 0;  // All streams
+
+        if (zoom <= 5) {
+            stream_order = 7;
+        } else if (zoom <= 8) {
+            stream_order = 6;
+        } else if (zoom <= 10) {
+            stream_order = 5;
+        }
+
+        return '(SELECT geom, stream_order FROM ' + tableName +
+          ' WHERE stream_order >= ' + stream_order + ') as q';
     };
 
 var config = {
@@ -125,6 +148,12 @@ var config = {
                 tableName = tables[tableId];
 
             req.params.table = tableName;
+
+            // Streams have special performance optimized SQL queries
+            if (tableId.indexOf('streams') >= 0) {
+                req.params.sql = getSqlForStreamByReq(req);
+            }
+
             req.params.dbname = dbName;
             req.params.style = styles;
             req.params.interactivity = interactivity[tableName];
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -78,7 +78,7 @@ function validateClickedPointWithinDRB(latlng) {
     var point = L.marker(latlng).toGeoJSON(),
         d = $.Deferred(),
         streamLayers = settings.get('stream_layers'),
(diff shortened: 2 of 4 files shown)


mmcfarland:
#### DRB Streams
drb_streams

mmcfarland:
#### NHD Streams
nhd_streams_sm

mmcfarland:
Updated with a commit that renames both steam layers.  This is to avoid the CloudFront/s3 caching we have in place in staging.  Each PR that adjusts these styles can rename the layer so that changes are visible immediately with waiting for caches to expire or needing to purge them on staging/production.

rajadain:
I restarted the `mmw-tiler`service, reloaded the `tiler` VM, then reprovisioned it, but I'm still getting a 400 on all the requests:

image

rajadain:
Here's a response detail:

image

rajadain:
Ah, looking at the tiler log I see the following:

```
[TILE RENDER ERROR] -- 400: Error: Postgis Plugin: ERROR:  column "stream_order" does not exist
LINE 1: SELECT ST_AsBinary("geom") AS geom,"stream_order" FROM drb_s...
                                           ^

in executeQuery Full sql was: 'SELECT ST_AsBinary("geom") AS geom,"stream_order" FROM drb_streams_50 WHERE "geom" && ST_SetSRID('BOX3D(-75.14923095703125 39.94133068417753,-75.09979248046875 39.97922477515792)'::box3d, 4326)'
```

I probably need to update my local streams datasets. Doing so now.

mmcfarland:
Oh sorry - yeah, that work didn't coincide with this PR so it slipped my mind, but this does require an updated NHD stream set.  The old DRB already had a stream order attribute, but i renamed it in a data update to match NHD so that we could use the same rendering rules for both.  You'll need to update both.

rajadain:
Consider renaming `drb_streams` here as well: https://github.com/WikiWatershed/model-my-watershed/blob/c4891a56401cce469c634e1330a2ba5c0511c492/src/mmw/js/src/draw/views.js#L81

rajadain:
If we're moving to a convention of always using new IDs for the caching layer, can we excise the `purge_tile_cache` command here https://github.com/WikiWatershed/model-my-watershed/blob/77982cc669568de3c8c0658f5440de0e76619e62/scripts/aws/setupdb.sh#L68-L72 ? The [specified](https://github.com/WikiWatershed/model-my-watershed/blob/77982cc669568de3c8c0658f5440de0e76619e62/scripts/aws/setupdb.sh#L101) [paths](https://github.com/WikiWatershed/model-my-watershed/blob/77982cc669568de3c8c0658f5440de0e76619e62/scripts/aws/setupdb.sh#L110) are no longer valid.

rajadain:
+1 tested, they look much nicer now, and I didn't have to wait very long for them to load either.

`setupdb` with `nhdflowline` was taking forever within the `app` VM, so I had to manually do the steps by saving the 1.2 GB file to disk, then unzipping it to ~3.8 GB, and finally running `psql` on that, which was relatively quicker. I think my `app` VM may not have had enough RAM to be able to stream that via pipes very quickly.

mmcfarland:
Good catch with the missed rename, addressed in 03ce620.  

Also added a commit that improves rendering performance at very low zoom levels to satisfy a request at showing high stream_orders at a high altitude view.  Slightly tweaked the rendering rules, but the testing should now include viewing both stream sets at high levels.

nhd
drb

mmcfarland:
Re: the cache purge paths, I was planning on changing the names back to their original names in a subsequent PR, after however many number of revisions of style tweaks.  I'm going to keep the old purge names for now and manually clean up the s3 buckets for these new ones unless this cache bypass ends up working out better than the cache purge.

rajadain:
Cool.

It's really interesting seeing the visualizations at the national scale. I don't think I ever had a sense of the major rivers in the country until now.

src/tiler/server.js
```diff
@@ -124,7 +147,14 @@ var config = {
             var tableId = req.params.tableId,
                 tableName = tables[tableId];
 
+
```
mmcfarland: I'll take out this extra line.

rajadain:
+1 tested the fixups. Works well. Great job!

Issue #1429, opened 2016-08-25
Change DRB to Delaware River Basin in layer selector


ajrobbins:
![screen shot 2016-08-25 at 9 49 44 am](https://cloud.githubusercontent.com/assets/12868342/17971670/ccc4d52e-6aa9-11e6-9ed5-8252af370fa8.png)

Should read "Delaware River Basin Stream Network".

kellyi:
Shares a point with #1431

PR #1439, opened 2016-08-26
Rename `DRB` -> `Delaware River Basin`

kellyi:
This PR renames "DRB Stream Network" to "Delaware River Basin Stream Network" in the frontend layer selector control:

screen shot 2016-08-26 at 9 56 35 am

**Testing**
- get this branch, `vagrant up`, `'./scripts/bundle.sh`, then visit `localhost:8000` and open the browser console
- click on the layer selector control, select the "Overlays" tab, and verify that "DRB Stream Network" now reads "Delaware River Basin Stream Network", that the selector still works to toggle the layer, and that there are no errors in the console

Connects #1429

### DIFF
--- a/src/mmw/mmw/settings/layer_settings.py
+++ b/src/mmw/mmw/settings/layer_settings.py
@@ -135,7 +135,8 @@ LAYERS = [
     },
     {
         'code': 'drb_streams_v1',
-        'display': 'DRB Stream Network',
+        'display': ('Delaware River Basin High Resolution' +
+            '<br />&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Stream Network'),
         'table_name': 'drb_streams_50',
         'stream': True,
         'overlay': True,


ajrobbins:
Sorry for the late-breaking request (just came in from Stroud) - can we add "high resolution" to this? So it should read "Delaware River Basin High Resolution Stream Network".

kellyi:
Sure!

rajadain:
Is that going to fit?

kellyi:
Looks like it fits:

screen shot 2016-08-26 at 2 08 21 pm

rajadain:
Ah it just stretches out

ajrobbins:
Yah I was worried that it wouldn't fit well - can we wrap "Stream Network" to the next line?

kellyi:
Yeah -- I'll try to have it wrap.

kellyi:
Updated:

screen shot 2016-08-26 at 3 22 14 pm

ajrobbins:
I like this better, what do you think @rajadain ? The layer selector styling in general will likely be altered in the near-ish future, so I don't think we should stress about the general look too much. But this wrapping brings it more in line with the other layer names.

kellyi:
The function I added will try to wrap all layer names > 35 char long like that, of which this is currently the only one. Testing instructions should probably include verifying that none of the other layer names have been affected by the change.

rajadain:
Tested at different widths:

image

image

src/mmw/mmw/settings/layer_settings.py
```diff
@@ -135,7 +135,7 @@
     },
     {
         'code': 'drb_streams_v1',
-        'display': 'DRB Stream Network',
+        'display': 'Delaware River Basin High Resolution Stream Network',
```
rajadain: Would it not have been easier to just add a `<br />` in this field?

kellyi: Ah! Likely true. Let me try that and verify that it works, as it's much simpler than my alternative here.

src/mmw/js/src/core/layerControl.js
```diff
@@ -123,6 +123,7 @@ module.exports = L.Control.Layers.extend({
         L.DomEvent.on(input, 'click', this._onInputClick, this);
 
         var name = document.createElement('span');
+
```
rajadain: Extra newline

rajadain:
+1 tested! Sweet and simple. Friendly reminder to squash before merging.

kellyi:
Thanks! I cut the function, the whitespace, and the extra commits, and I'll merge this in once the tests pass.

Issue #1431, opened 2016-08-25
Align rounding rules for Animal count


ajrobbins:
We are rounding the animal count up on the front end and down in the GMS file.

kellyi:
For reference: we are rounding the population counts down for the GMS file here:

https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/mapshed/tasks.py#L124

...and rounding them up for the frontend Analyze tab here:

https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/calcs.py#L31

I'm going to round the frontend counts down to match the work already done for MapShed.

PR #1440, opened 2016-08-26
Align animal population counts rounding

kellyi:
This PR adjusts the rounding of the animal population count estimate displayed in the "Analyze" tab to round the float down rather than up so that it matches how MapShed rounds animal population counts for the GMS file.

**Testing**
- get this branch, `vagrant up`, then visit `localhost:8000` and open the browser console
- select an AOI and let the Analyze tab return results
- run a MapShed job on the AOI, then export a GMS file
- open the GMS file and search for "Dairy Cows"
- click back on the "Analyze" tab and verify that the integer count for Dairy Cows is the same as the number immediately to the right of "Dairy Cows" in the GMS file
- verify the same for the other animal population counts

Connects #1431


rajadain:
+1 tested. The numbers in the UI and the GMS file are the same:

image

kellyi:
Thanks!

Issue #1437, opened 2016-08-25
Include Point Source pollution data in Analysis summary results


ajrobbins:
Basic summary statistics for the point source pollution data that we currently use for Mapshed should be displayed in the Analysis tab, as was done with animal data in https://github.com/WikiWatershed/model-my-watershed/pull/1425

The tab should be called "Point Source". According to the contract, the statistics for display should be "limited to aggregations of the existing attributes in the table". I'm not familiar with the data attributes, so there will have to be a little back and forth with BME and Stroud as to which attributes to display and how.

Note: We will be receiving DRB-specific Point Source data. When an analysis is run on an AoI within the DRB, we should use the DRB-specific data. This DRB check will also be implemented in the Mapshed model run https://github.com/WikiWatershed/model-my-watershed/issues/1436 
For now, we can mock the DRB-specific data with an identical table, since it will have the same schema.

kellyi:
I'm going to generate the Analyze results using the existing MapShed method, so when we're ready to switch to the using DRB data for DRB AOIs it can all be handled there.

kellyi:
These are the columns currently available in the `ms_pointsource` table:

AVG_N_CONC, 
LBSN_YR,
MGD,
AVGPCONC,
LBSP_YR,
KGN_YR,
KGP_YR

For MapShed, it looks like we're currently using MGD, KGN_YR, KGP_YR and transforming those values into estimates of monthly nitrogen load, monthly phosphorus load, and monthly discharge.

@ajrobbins can you ask if these are the values we should output in the Analyze tab or if there are others to include as well?

ajrobbins:
BME has provided recommended output instructions here: https://docs.google.com/document/d/1Kg93peuPF5OTQUzKCiWtzgFJY-kMegCPc0jljXI6BP4/edit

Let me know if you have any trouble viewing the document; I tried to share it with you.

kellyi:
Thanks! I was able to see the document, so I'll adjust the code to display like that.

ajrobbins:
@aufdenkampe requested that we include an additional column for point source name/code. So I think the table should show: Code | Location| Discharge (MGD) | TN Load (Kg/yr) | TP Load (kg/yr)

PR #1442, opened 2016-08-26
Display Point Source Data in "Analyze" Tab

kellyi:
This PR adds Point Source data from the `ms_pointsource` table to MMW's "Analyze" tab:

screen shot 2016-08-29 at 4 56 30 pm

Since some of the columns in that data set contain null values, I've replaced the null value with "no data" and I've added a comparator method to handle sorting the HTML table. The PR also adds a method to reduce the values in each column to a single value for the Total row in the table footer.

**Testing**
- grab this branch, `vagrant up`, `./scripts/bundle.sh` then visit `localhost:8000` and open the browser console
- hard refresh the page to load js code from this branch
- search for an AOI and verify that when the results return the Analyze tab now includes a Point Source sub-tab which presents the point source data in a table with sortable columns.

Connects #1437

### DIFF
--- a/src/mmw/apps/modeling/calcs.py
+++ b/src/mmw/apps/modeling/calcs.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import
 from django.contrib.gis.geos import GEOSGeometry
 
 from django.conf import settings
+from django.db import connection
 
 from apps.modeling.mapshed.calcs import animal_energy_units
 
@@ -35,4 +36,36 @@ def animal_population(geojson):
         'displayName': 'Animals',
         'name': 'animals',
         'categories': aeu_return_values
+    }
+
+
+def point_source_pollution(geojson):
+    """
+    Given a GeoJSON shape, retrieve point source pollution data
+    from the `ms_pointsource` table to display in the Analyze tab.
+    Returns a dictionary to append to the outgoing JSON for analysis
+    results.
+    """
+    geom = GEOSGeometry(geojson, srid=4326)
+
+    sql = '''
+          SELECT city, npdes_id, mgd, kgn_yr, kgp_yr
+          FROM ms_pointsource
+          WHERE ST_Intersects(geom, ST_SetSRID(ST_GeomFromText(%s), 4326))
+          '''
+
+    with connection.cursor() as cursor:
+        cursor.execute(sql, [geom.wkt])
+
+        if cursor.rowcount != 0:
+            columns = [col[0] for col in cursor.description]
+            point_source_results = [dict(zip(columns, row)) for row in
+                                    cursor.fetchall()]
+        else:
+            point_source_results = []
+
+    return {
+        'displayName': 'Point Source',
+        'name': 'pointsource',
+        'categories': point_source_results
     }
--- a/src/mmw/js/src/core/utils.js
+++ b/src/mmw/js/src/core/utils.js
@@ -1,6 +1,7 @@
 "use strict";
 
 var _ = require('underscore'),
+    lodash = require('lodash'),
     md5 = require('blueimp-md5').md5,
     intersect = require('turf-intersect');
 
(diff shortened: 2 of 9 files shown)


src/mmw/js/src/core/utils.js
```diff
@@ -94,6 +94,18 @@ var utils = {
         }
     },
 
+    noDataSort: function(x, y) {
+        var noData = 'no data';
+        if (x === noData && y !== noData) {
+            return -1;
+        } else if (x === noData && y === noData) {
+            return 0;
+        } else if (y === noData && x !== noData) {
```
rajadain: prefer to keep the order of `x` and `y` the same throughout:

```js
} else if (x !== noData && y === noData) {
  return 1;
}
```

src/mmw/js/src/core/utils.js
```diff
@@ -254,6 +266,12 @@ var utils = {
 
             return l < r ? 1 : l > r ? -1 : 0;
         };
+    },
+
+    totalForPointSourceCollection: function(collection, key) {
+        return _.reduce(collection, function(accumulator, element) {
```
rajadain: This may be made easier to read as a combination of `_.sum` and `_.map`:

```js
return _.sum(_.map(collection, function(element) {
  return element.attributes[key] || 0;
}));
```

src/mmw/apps/modeling/calcs.py
```diff
@@ -36,3 +39,42 @@ def animal_population(geojson):
         'name': 'animals',
         'categories': aeu_return_values
     }
+
+
+def point_source_pollution(geojson):
+    """
+    Given a GeoJSON shape, call MapShed's `point_source_discharge` method
```
rajadain: It doesn't look like we are calling MapShed's [`point_source_discharge`](https://github.com/WikiWatershed/model-my-watershed/blob/c2eacd3383753a51badfed30190dab33e3e4e03b/src/mmw/apps/modeling/mapshed/calcs.py#L351) method here?

src/mmw/js/src/core/utils.js
kellyi: I checked and it didn't look like underscore had a `sum` method.

src/mmw/apps/modeling/calcs.py
kellyi: Oh, good catch -- docstring was never updated from the first pass at this.

src/mmw/js/src/core/utils.js
rajadain: Switch to `lodash`. It's already included as a dependency, and has [`_.sum`](https://lodash.com/docs#sum)

src/mmw/apps/modeling/calcs.py
```diff
@@ -36,3 +39,42 @@ def animal_population(geojson):
...
+          SELECT city, npdes_id, mgd, kgn_yr, kgp_yr
+          FROM ms_pointsource
+          WHERE ST_Intersects(geom, ST_SetSRID(ST_GeomFromText(%s), 4326))
+          '''
+
+    with connection.cursor() as cursor:
+        cursor.execute(sql, [geom.wkt])
+
+        if cursor.rowcount != 0:
+            point_source = namedtuple('Point_Source_Data',
```
rajadain: Rather than creating a named tuple out of query results and then translating that into a dictionary, consider directly creating a dictionary from the query results:

```python
columns = [col[0] for col in cursor.description]
point_source_results = [dict(zip(columns, row)) for row in cursor.fetchall()]
```

rajadain:
Notably, I had to restart Celery to pick up the changes in this PR

```bash
$ vagrant ssh worker -c 'sudo service celeryd restart'
celery init v10.1.
Using config script: /etc/default/celeryd
celery multi v3.1.18 (Cipater)
> Stopping nodes...
	> Black-worker1@worker: TERM -> 8562
	> Black-worker0@worker: TERM -> 8544
> Waiting for 2 nodes -> 8562, 8544......
	> Black-worker1@worker: OK
> Restarting node Black-worker1@worker: OK
> Waiting for 1 node -> 8544....
	> Black-worker0@worker: OK
> Restarting node Black-worker0@worker: OK

```

rajadain:
+1 tested, works as advertised:

image

Let me know if you need any more clarifications on the code recommendations above. Otherwise this is good to go!

kellyi:
Thanks for taking a look at this! I'll make the changes mentioned above, then verify that it works before merging.

src/mmw/js/src/core/utils.js
kellyi: Will do!

PR #1447, opened 2016-08-30
Convert point source data from decimals to floats

kellyi:
Fixes a bug whereby Decimals returned from the point source db query were not serializable as JSON by converting them to floats.

**Testing**
- get this branch, `vagrant up && ./scripts/bundle.sh`
- visit `localhost:8000`, select an AOI, and verify that the Analysis results return including point source data
- draw a sq km AOI and verify that the analysis results return when the point source tab includes no data

Connects #1437

### DIFF
--- a/src/mmw/apps/modeling/calcs.py
+++ b/src/mmw/apps/modeling/calcs.py
@@ -59,8 +59,14 @@ def point_source_pollution(geojson):
 
         if cursor.rowcount != 0:
             columns = [col[0] for col in cursor.description]
-            point_source_results = [dict(zip(columns, row)) for row in
-                                    cursor.fetchall()]
+            point_source_results = [
+                dict(zip(columns,
+                         [row[0], row[1],
+                          float(row[2]) if row[2] else None,
+                          float(row[3]) if row[3] else None,
+                          float(row[4]) if row[4] else None]))
+                for row in cursor.fetchall()
+            ]
         else:
             point_source_results = []
 


kellyi:
Also may be worth adding `print('point_source_results => {}'.format(point_source_results)` here -- https://github.com/kellyi/model-my-watershed/blob/fa27e243f2afdc638382e9614dae739ac1b0b9bd/src/mmw/apps/modeling/calcs.py#L73 -- and then monitoring `./scripts/debugcelery.sh` output to see that they are no longer decimals.

src/mmw/apps/modeling/calcs.py
```diff
@@ -59,8 +61,13 @@ def point_source_pollution(geojson):
 
         if cursor.rowcount != 0:
             columns = [col[0] for col in cursor.description]
-            point_source_results = [dict(zip(columns, row)) for row in
-                                    cursor.fetchall()]
+            point_source_results = []
+            for row in cursor.fetchall():
+                point_source_row = dict(zip(columns, row))
+                for k, v in point_source_row.iteritems():
+                    if isinstance(v, decimal.Decimal):
```
rajadain: Since we _know_ which fields could be decimals, this check is unnecessary. This may be a matter of taste, but I would prefer the more Pythonic:

```diff
diff --git a/src/mmw/apps/modeling/calcs.py b/src/mmw/apps/modeling/calcs.py
index 6a21eb4..1799292 100644
--- a/src/mmw/apps/modeling/calcs.py
+++ b/src/mmw/apps/modeling/calcs.py
@@ -61,13 +61,14 @@ def point_source_pollution(geojson):
 
         if cursor.rowcount != 0:
             columns = [col[0] for col in cursor.description]
-            point_source_results = []
-            for row in cursor.fetchall():
-                point_source_row = dict(zip(columns, row))
-                for k, v in point_source_row.iteritems():
-                    if isinstance(v, decimal.Decimal):
-                        point_source_row[k] = float(v)
-                point_source_results.append(point_source_row)
+            point_source_results = [
+                dict(zip(columns,
+                         [row[0], row[1],
+                          float(row[2]) if row[2] else None,
+                          float(row[3]) if row[3] else None,
+                          float(row[4]) if row[4] else None]))
+                for row in cursor.fetchall()
+            ]
         else:
             point_source_results = []
 
```

We need those three `if` checks in there since `float(None)` raises an error.

kellyi: Sounds good to me!

rajadain:
Tested locally and on staging. Good job!

rajadain:
+1

kellyi:
Thanks for the quick review! I adjusted the `point_source_results` code, removed the unnecessary `import decimal` line, and checked again on my local. Will merge once the CI tests pass.

PR #1450, opened 2016-08-31
Fix mismatched HTML tag

rajadain:
Wish we had a linter for this.


kellyi:
+1. Checked this in the browser and it's got the proper closing tag now:

screen shot 2016-08-31 at 1 13 42 pm

Issue #1438, opened 2016-08-25
Asynchronize Analysis Requests


rajadain:
As the analysis screen grows more complicated, the longest-to-calculate dataset holds back all data from the server.

Address this by having different access points for each data set. One solution may be to have different endpoints for each dataset, which would allow easy separation on both request and response side.

Another solution may be to have a single endpoint for submitting the request (since each endpoint would use the same input: area of interest), which returns two arrays: one of results and one of job ids. The results are used immediately, and the job ids can be used to poll independently.

Or use some combination of these approaches.

This should also include a way to submit multiple requests from the front-end, as well as have different spinners  indicating loading for each analysis tab.

rajadain:
The effort to make the Analyze view asynchronous to accomodate more, longer requests is constrained by the following assumptions:

 * All endpoints require the same input data, namely the shape of the area of interest.
 * The input data can potentially be heavy, of the order of 100s of KBs.
 * Responses may be immediate (`{results: {...}}`) or asynchronous (`{job_id: 1234}`), and the request does not know ahead of time which it is going to be. Thus, the front-end should be able to handle both.
   - This is advantageous because it allows us to change the response-type from the server side without having to change the client-side code.
   - It also allows us to have one piece of code on the client-side responsible for handling all types of Analyze responses.
 * We may have a different set of Analyze requests based on configuration (MMW vs BigCZ)
 * We may have a different set of Analyze requests based on application context (inside or outside the Delware River Basin)
 * One request may populate more than one tab (Land Cover / Soil tabs are populated from a single asynchronous request)

Given the above, we ponder the following questions:

 * Should client-side requests be combined into one request, since they all require the same data? Or should they remain independent endpoints for easy addition / removal?
 * Should the list of Analyze endpoints and tabs now be represented through configuration? In some JSON object in `clientSettings`, to be used by meta code? Or should it remain a set of views and models that is used directly in hand-written JavaScript?
 * Since there are now to be a number of endpoints for Analyze, should these be pulled in to their own Django app, out of `modeling`?
 * Should the "immediately responding" endpoints return a direct response in a new structure, or return a job_id anyway and go through Celery anyway, just so that the client-side handling can be simplified and always expect the same structure? A simpler Celery job would finish quicker, although not instantaneously.

My leanings are towards:

[...]

PR #1451, opened 2016-09-01
Asynchronize Analyze

rajadain:
## Overview

Given that the **Analyze** tab is getting more content, we switch to an asynchronous fetching pattern so that quicker results may return sooner than slower ones.

The central difference is that instead of one analyze endpoint which returns all results, we now have multiple analyze endpoints which return multiple results. This manifests as follows in our code:

 * The main `/api/modeling/start/analyze/` endpoint now returns only the results of the original Analyze chain, which includes counts from NLCD and Soil rasters. The results of `animal_population` and `point_source_pollution` have been removed from its output.
 * We now have two new endpoints `/api/modeling/start/analyze/animals/` and `api/modeling/start/analyze/pointsources/` for Animals and Point Sources respectively. Both operate identically to the `analyze` endpoint, inasmuch they respond with a job id which can be polled to get results. Since both are quick jobs, the results are usually available on the first poll.
 * Instead of having one `App.analyzeModel` which was the task runner for the `analyze` endpoint, we now have `App.analyzeCollection` which contains task runners for `analyze`, `analyze/animals`, and `analyze/pointsources`.
 * Instead of a number of `if`s and `switch`es in `TabContentView`, we now show a different result view for each result type. This separation allows us to more easily make changes between result views.
 * Instead of the entire `AnalyzeWindow` handling the polling messaging, that is now handled by `TabContentView` so we can show different polling states for each task runner. The polling is also reflected in a spinner on the tab panel.
 * Each result view is populated by a view model, which points to exactly one task runner. The `land` and `soil` view models point to the same task runner, since they get their results from the same endpoint, and thus are updated at the same time. `animals` and `pointsource` view models point to their own task runners, and are updated independently. These can finish or fail independently without affecting one another.

### Demo

http://gfycat.com/BasicElaborateGopher

### Notes

~~The tests for the Analyze view were never very elaborate. Now they are testing views that are no longer used. I tried to come up with better ones, but that's a big task on its own. We should probably create a card for that.~~

The new tests exercise only the success workflow. We should add tests that exercise waiting, failing workflows as well.

## Testing Instructions

 * Bundle scripts `./scripts/bundle.sh --debug`
 * Debug Celery so it picks up new tasks `./scripts/debugcelery.sh`
 * Open the app. Draw / Select an area of interest.
 * In the Analyze sidebar, instead of a simple "Analyzing" message, you should see four tabs corresponding to each result type, each with a spinner and its own "Analyzing" message. The Animals and Point Source tabs should populate first, then Land and Soil should populate together.
 * Going to Model stage should preserve the Analyze tab and it should continue to work as before.
 * Going to Draw and going back to Analyze without changing the area of interest should not trigger new requests.
 * Loading a saved project should trigger Analyze requests, and they should behave identically as they did when coming in from Draw.

Connects #1438


src/mmw/js/src/analyze/views.js
```diff
@@ -215,14 +187,6 @@ var TabPanelsView = Marionette.CollectionView.extend({
     },
     childView: TabPanelView,
 
-    events: {
-        'shown.bs.tab li a ': 'triggerBarChartRefresh'
-    },
-
-    triggerBarChartRefresh: function() {
-        $('#analyze-output-wrapper .bar-chart').trigger('bar-chart:refresh');
```
rajadain: I'm not really sure what this code was supposed to do. Deleting it has had no effect.

src/mmw/js/src/analyze/views.js
```diff
@@ -238,60 +202,71 @@ var TabContentView = Marionette.LayoutView.extend({
...
-        } else if (this.model.get('name') === 'pointsource') {
-            this.tableRegion.show(new PointSourceTableView({
-                units: units,
-                collection: census
-            }));
+        this.showAnalyzingMessage();
+
+        this.model.get('taskRunner').fetchAnalysisIfNeeded()
+            .done(_.bind(this.showResultsIfNotDestroyed, this))
+            .fail(_.bind(this.showErrorIfNotDestroyed, this));
```
rajadain: We need the `IfNotDestroyed` versions for the case when loading an existing project, and the Analyze requests are fired but the Analyze view is not shown. Otherwise, it tries to show views in regions that don't exist and blows up.

src/mmw/apps/modeling/views.py
```diff
@@ -312,6 +312,54 @@ def start_analyze(request, format=None):
...
+def start_analyze_pointsource(request, format=None):
+    user = request.user if request.user.is_authenticated() else None
+    created = now()
+    area_of_interest = request.POST['area_of_interest']
+    job = Job.objects.create(created_at=created, result='', error='',
+                             traceback='', user=user, status='started')
+
+    chain = (tasks.analyze_pointsource.s(area_of_interest) |
+             save_job_result.s(job.id, area_of_interest))
+    task_list = chain.apply_async(link_error=save_job_error.s(job.id))
```
rajadain: There's a lot of rhyming in this pattern here. We should perhaps one day try to refactor this into something reusable. A lot of these variables are interdependent though, so it would be non-trivial to disentangle them.

rajadain:
Hmm looks like the tests aren't as unaffected as I thought. Alright will give this another look.

src/mmw/apps/modeling/views.py
lewfish: I could be wrong, but it looks like the only difference between these two `start...` functions is the task that is run. Maybe you could extract a single function that just takes the task as an argument.

src/mmw/js/src/analyze/models.js
```diff
@@ -54,6 +54,33 @@ var AnalyzeTaskModel = coreModels.TaskModel.extend({
     }
 });
 
+var AnalyzeTaskCollection = Backbone.Collection.extend({
+    model: AnalyzeTaskModel
+});
+
+AnalyzeTaskCollection.getSingleton = function(App, aoi) {
```
lewfish: This is just stylistic, but couldn't you include this and the next function in the object passed to `Collection.extend` above?

src/mmw/js/src/analyze/views.js
```diff
@@ -452,6 +453,87 @@ var ChartView = Marionette.ItemView.extend({
...
+        this.tableRegion.show(new AnimalTableView({
+            units: units,
+            collection: census
+        }));
+    }
+});
+
+var PointSourceResultView = AnalyzeResultView.extend({
+    onShow: function() {
+        var categories = this.model.get('categories'),
```
lewfish: Consider extracting out a single function from the `onShow` functions for the four `ResultView` classes since they only differ in the `CensusCollection` and `TableView` classes that are used.

src/mmw/js/src/analyze/models.js
```diff
@@ -54,6 +54,33 @@ var AnalyzeTaskModel = coreModels.TaskModel.extend({
...
+    if (!App.analyzeCollection) {
```
lewfish: This might work correctly, but I was somewhat surprised that this only checks that the `analyzeCollection` doesn't exist. It seems like it should also check to see if `aoi` is equal to the `aoi` for the current `analyzeCollection` if it exists.

lewfish:
When you go to Analyze, wait for polling to stop, and draw another AOI, the analysis is not triggered again.

src/mmw/js/src/analyze/models.js
rajadain: Hmm you're right. Good catch! Although previously we were just checking if `analyzeModel` exists, so I just assumed it would work, but apparently I must have missed something.

src/mmw/js/src/analyze/views.js
rajadain: One of the reasons to separate these was that even though their code is similar, things may be more different / advanced in upcoming variations like BigCZ. Thus I thought it best to keep each `AnalyzeResultView` as independent as possible, even at the expense of some redundancy. If this feels like too early an optimization, I can look into grouping theses together.

src/mmw/js/src/analyze/models.js
rajadain: I think `getSingleton` needs to be defined outside of `Backbone.Collection.extend()` because it is instantiating its own collection.

`createViewModelCollection` doesn't need to be associated with `AnalyzeTaskCollection` at all, so I'll pull that out.

src/mmw/js/src/analyze/models.js
rajadain: Ah, we were deleting `analyzeModel` in the Draw controller. Updated that to deleting `analyzeCollection` in 028ed43

src/mmw/js/src/analyze/views.js
lewfish: Yea, I think that sounds like a premature optimization.

src/mmw/js/src/analyze/models.js
rajadain: Done in the latest version: ab8b9cd

src/mmw/js/src/analyze/models.js
```diff
@@ -54,27 +54,37 @@ var AnalyzeTaskModel = coreModels.TaskModel.extend({
...
-// Pass in the serialized Area of Interest for
-// caching purposes (_.memoize returns the same
-// results for any object), and deserialize
-// the AoI for use on the model.
-var createTaskModel = _.memoize(function(aoi) {
-    return new AnalyzeTaskModel({
-        area_of_interest: JSON.parse(aoi)
-    });
-});
+function createAnalyzeResultViewModelCollection(App, aoi) {
```
kdeloach: Considering the first argument is `App`, does it make sense to move this logic to the `App` class?

src/mmw/js/src/draw/controllers.js
```diff
@@ -53,7 +53,7 @@ var DrawController = {
     },
 
     drawCleanUp: function() {
-        delete App.analyzeModel;
+        delete App.analyzeCollection;
```
kdeloach: Consider adding functions to assign/delete `analyzeCollection` inside the `App` class.

kdeloach:
This looks great so far.

src/mmw/js/src/analyze/models.js
rajadain: If I move this logic to the file that describes the `App` class, I will have to `require` `AnalyzeTaskCollection` to be able to call `getSingleton` from here, which could lead to cyclic dependencies.

If I define this on `App` here, by `require`ing `App` and then doing `App.createAnalyzeResultViewModelCollection = function(aoi)`, I think that is more confusing because then it is less clear where the function gets defined and attached.

rajadain:
The tests seem to require considerable redesign. Am going to address the issues brought up here first, then move on to them.

src/mmw/js/src/analyze/models.js
kdeloach: Sounds like there is a flaw in our dependency chain. One would expect that `App` should be able to include models and collections since it is a high level controller.

rajadain: True. I just tried it and it worked without any dependency issues. _However_, do we then move this kind of logic:

```javascript
return new Backbone.Collection([
    { name: 'land', displayName: 'Land', taskRunner: runners.findWhere({ taskName: 'analyze' }) },
    { name: 'soil', displayName: 'Soil', taskRunner: runners.findWhere({ taskName: 'analyze' }) },
    { name: 'animals', displayName: 'Animals', taskRunner: runners.findWhere({ taskName: 'analyze/animals' }) },
    { name: 'pointsource', displayName: 'Point Sources', taskRunner: runners.findWhere({ taskName: 'analyze/pointsource' }) },
]);
```

to `App`? This seems very specific to Analyze, and probably should remain there.

rajadain: The only reason the first parameter is `App` is because we want to get a singleton copy of the task runners, which query and store the data from the API. These are attached to the App, which is our global scope. The logic of this function does not require App at all, and we could just as easily hand it a list of task runners. The reason I haven't done that is then we'd have to get that list outside of this function _every time_, and so we might as well make that transparent by doing it inside.

kdeloach: I think I would keep this logic here, but I would move the actual singleton getter/setter logic to `App`.

The rationale for this is to encapsulate implementation logic inside `App` and avoid mutating it from an external function.

So instead of

```
var App = { };

function getSingleton(App) {
    if (!App.analyzeCollection) {
        App.analyzeCollection = { ... };
    }
    return App.analyzeCollection;
}

getSingleton(App);
delete App.analyzeCollection;
```

We could do

```
var App = {
    getAnalyzeCollection: function() {
        if (!this.analyzeCollection) {
            this.analyzeCollection = createAnalyzeCollection();
        }
        return this.analyzeCollection;
    },
    clearAnalyzeCollection: function() {
        delete this.analyzeCollection;
    }
};

function createAnalyzeCollection() {
    return { ... };
}

App.getAnalyzeCollection();
App.clearAnalyzeCollection();
```

kdeloach: (This is just an idea)

rajadain: Ah! I see, yes that makes total sense. Sorry that took a while to get through to me. Let me switch to that real quick ...

rajadain: Done in the latest version.

src/mmw/js/src/draw/controllers.js
rajadain: Done. Also moved the code from here, which gets executed every time `draw` is exited, to `ResetDrawView.resetDrawingState`. Now, the analysis results will be preserved if the user goes back to Draw but does not reset the drawing state, and will only be refreshed when the existing area of interest is cleared.

rajadain:
I've updated the tests to ensure that the tables render the right values given successful results, which should be enough to make the tests pass. However, I don't think that this is testing enough, or even testing the right thing. I'm not sure what that is. But I think we should test (somehow) that the analyses are independent, and also test error conditions (timeouts, bad responses). Will create an additional card for that.

rajadain:
Just added a new refactored version here fcba5bd which reduces redundancy in starting Celery jobs.

rajadain:
Okay this is finally ready for another review. Thanks for the patience, all.

kellyi:
Heads up that there's a merge conflict.

rajadain:
Urgh. Fixing shortly.

src/mmw/js/src/analyze/views.js
caseypt: I think this was necessary when we had the chart and table in the bottom bar. We needed to trigger this so that the bar chart would resize itself after the size of the bottom bar changed, I think.

rajadain:
Rebased on `develop`, there are no more merge conflicts.

caseypt:
Code looks good. Attempting to test this out, but I'm having some Vagrant mounting issues with the  worker VM.

caseypt:
Fixed the mounting issue. Testing now.

caseypt:
If I select an AoI, let analyze complete, click back in the browser, then select a different AoI, the analysis isn't rerun and the results from the previous AoI are shown.

caseypt:
Selecting an AoI that is completely made up of water causes the animal endpoint to fail:

```
{"status":"failed","started":"2016-09-08T15:46:32.656Z","finished":"2016-09-08T15:46:35.535Z","result":"","error":"unsupported operand type(s) for *: 'NoneType' and 'float'","job_uuid":"d510c772-46c8-4c4e-b8b1-cf7a24b98381"}
```

rajadain:
> Selecting an AoI that is completely made up of water causes the animal endpoint to fail

We may want to make a separate card for that, but what's so great about this card is that the other endpoints won't necessarily fail just because this one does.

rajadain:
> click back in the browser

Ah I hadn't tested this. Let me see what can be done.

caseypt:
> We may want to make a separate card for that, but what's so great about this card is that the other endpoints won't necessarily fail just because this one does.

Cool. Will do.

rajadain:
The existing Analyze Collection should be cleared properly now.

caseypt:
Taking another look

caseypt:
Verified that selecting a new AoI generates new analyses.

+1. Nice work.

Issue #1444, opened 2016-08-29
Add point source locations to map


ajrobbins:
We are currently using point source pollution vector data for Mapshed, and in the Analysis step of MMW (https://github.com/WikiWatershed/model-my-watershed/pull/1442). We need to visualize these points as locations on the map, similar to how we are currently visualizing Vizer observations:

![screen shot 2016-08-29 at 4 10 07 pm](https://cloud.githubusercontent.com/assets/12868342/18065717/233fa17a-6e03-11e6-9043-be15cc12bcce.png)

The point source data visualization should be activated by selecting a new "Point Source" option in the observations tab of the layer selector. Clicking on a point should open a tooltip with the following data fields: Name (we do not currently have this in our data, it will be coming soon), Code, Location, Discharge,  TN Load, TP Load. This is the output that we will be showing in the "Point Source" Analysis tab.

PR #1452, opened 2016-09-06
Add Point Source locations layer to Observations tab

kellyi:
This PR adds a DRB Point Source locations marker layer to the Observations tab. To retrieve the DRB point source data, the PR creates an API endpoint at `/api/modeling/point-source` which queries `ms_pointsource_drb` to retrieve the relevant columns, then sends the data back as a GeoJSON FeatureCollection. The endpoint response includes a header to cache the results for a week on the client side, since the data presumably will not be updated very often.

On the client side, the changes involved creating the DRB Point Source marker layer, then adding it as an option to the Observations tab via the `layerControl.js` file. Creating the markers happens in a `pointSourceLayer.js` file, and the PR adds a new template for the popups in `pointSourcePopup.html` which looks like this:

screen shot 2016-09-07 at 3 50 24 pm

The markers currently resize contextually based on the zoom level in order to keep them from clustering together: each zoom level from 0-18 has a corresponding number indexed in a `markerSizesForZoomLevels` array. That means that the markers look like this when zoomed in closely:

screen shot 2016-09-07 at 3 52 26 pm

...and like this when you zoom out:

screen shot 2016-09-07 at 3 43 30 pm

...

screen shot 2016-09-07 at 3 43 46 pm

The first commit here upgrades `django-debug-toolbar` from version 1.3 to 1.5 to incorporate a bug fix addressing a bug which otherwise crashed the api endpoint in development. You'll have to reprovision your app VM to upgrade the library.

**Testing**
- get this branch, `vagrant provision app`, `./scripts/bundle.sh`, then visit `localhost:8000` and open the dev console
- open the layer selector and click on the Observations tab
- toggle the "DRB Point Source" layer on and verify that you see orange markers appear on the map
- click on a few point source markers and verify that you see the pop ups appear and that they're formatted properly (note that some of the fields will read "no data" if the database has null values for them
- zoom out and verify that the markers gradually decrease in size as you move further out, and that they increase in size as you zoom back in
- toggle some of the other layers in the control on and off to verify that they still work
- verify that you haven't seen any errors in the console

Connects #1444

### DIFF
--- a/src/mmw/apps/modeling/views.py
+++ b/src/mmw/apps/modeling/views.py
@@ -494,3 +494,43 @@ def boundary_layer_detail(request, table_code, obj_id):
             return Response(geojson)
         else:
             return Response(status=status.HTTP_400_BAD_REQUEST)
+
+
+@decorators.api_view(['GET'])
+@decorators.permission_classes((AllowAny, ))
+def drb_point_sources(request):
+    query = '''
+          SELECT ST_X(geom) as lon, ST_Y(geom) as lat, city, state, npdes_id,
+                 mgd, kgn_yr, kgp_yr, facilityname
+          FROM ms_pointsource_drb
+          '''
+
+    point_source_results = {u'type': u'FeatureCollection', u'features': []}
+
+    with connection.cursor() as cursor:
+        cursor.execute(query)
+
+        point_source_array = [
+            {
+                'type': 'Feature',
+                'geometry': {
+                    'type': 'Point',
+                    'coordinates': [row[0], row[1]],
+                },
+                'properties': {
+                    'city': row[2],
+                    'state': row[3],
+                    'npdes_id': row[4],
+                    'mgd': float(row[5]) if row[5] else None,
+                    'kgn_yr': float(row[6]) if row[6] else None,
+                    'kgp_yr': float(row[7]) if row[7] else None,
+                    'facilityname': row[8]
+                }
+
+            } for row in cursor.fetchall()
+        ]
+
+    point_source_results['features'] = point_source_array
+
+    return Response(json.dumps(point_source_results),
+                    headers={'Cache-Control': 'max-age: 604800'})
--- a/src/mmw/js/src/core/pointSourceLayer.js
+++ b/src/mmw/js/src/core/pointSourceLayer.js
@@ -1,0 +1,47 @@
+"use strict";
+
+var L = require('leaflet'),
+    $ = require('jquery'),
+    utils = require('./utils'),
+    Backbone = require('../../shim/backbone'),
+    Marionette = require('../../shim/backbone.marionette'),
+    pointSourcePopupTmpl = require('./templates/pointSourcePopup.html');
+
+// Increase or decrease the marker size based on the map zoom level
+var markerSizesForZoomLevels = [0.05, 0.1, 0.2, 0.25, 0.5, 0.75, 1, 1.25, 1.75,
+    2, 2.25, 3, 6, 8, 10, 10, 10, 10, 10];
+
+module.exports = {
+    createLayer: function(geojsonFeatureCollection, leafletMap) {
+        return L.geoJson($.parseJSON(geojsonFeatureCollection), {
+            pointToLayer: function (feature, latlng) {
+                return L.circleMarker(latlng, {
+                    fillColor: "#ff7800",
+                    weight: 0,
+                    fillOpacity: 0.75
+                });
+            },
+            onEachFeature: function(feature, marker) {
+                var model = new Backbone.Model(feature.properties),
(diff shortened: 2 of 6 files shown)


src/mmw/apps/modeling/views.py
```diff
@@ -494,3 +494,58 @@ def boundary_layer_detail(request, table_code, obj_id):
             return Response(geojson)
         else:
             return Response(status=status.HTTP_400_BAD_REQUEST)
+
+
+@decorators.api_view(['GET'])
+@decorators.permission_classes((AllowAny, ))
+def plot_point_source(request):
```
mmcfarland: This dataset is updated rarely, let's set aggressive cache headers on this endpoint.

src/mmw/apps/modeling/views.py
```diff
@@ -494,3 +494,58 @@ def boundary_layer_detail(request, table_code, obj_id):
...
+        query = '''
+              SELECT ST_AsGeoJSON(geom) as geometry, city, state, npdes_id,
+                     mgd, kgn_yr, kgp_yr
+              FROM ms_pointsource_drb
+              UNION
+              SELECT ST_AsGeoJSON(geom) as geometry, city, state, npdes_id,
+                     mgd, kgn_yr, kgp_yr
+              FROM ms_pointsource
+              '''
+    except (KeyError, IndexError):
```
kdeloach: I don't think the statement within this try/catch block can trigger these exceptions.

src/mmw/apps/modeling/views.py
```diff
@@ -494,3 +494,58 @@ def boundary_layer_detail(request, table_code, obj_id):
...
+
+        point_source_array = [
+            dict(zip(columns,
+                     [row[0],
+                      row[1],
+                      row[2],
+                      row[3],
+                      float(row[4]) if row[4] else None,
+                      float(row[5]) if row[5] else None,
+                      float(row[6]) if row[6] else None]))
```
kdeloach: Since the columns aren't dynamic, consider hard-coding the column->value mapping. This might make it easier to see at a glance that everything lines up correctly.

Ex.

```py
    point_source_array = [
        {
            'geometry': row[0],
            'city': row[1],
            # ...
        }
        for row in cursor.fetchall()
    ]
```

kellyi: Sounds good to me!

mmcfarland:
At 23k points, we're pretty far beyond the capability of the current vector point rendering system.  @ajrobbins your issue doesn't specify, but is this restricted to a particular area, like DRB only?

ajrobbins:
@kellyi I was incorrect when we chatted earlier - this supposed to be just for the DRB according to the contract. How many points is just DRB?

kellyi:
1055 for just DRB:

screen shot 2016-09-06 at 5 44 58 pm

ajrobbins:
@kellyi It's been requested that we include the site name in the tooltip - I see from your first screenshot that there is a "Point source name" in the tooltip. Is that placeholder at this point? I believe we should have a "name" field in the DRB data (though it wasn't in the national dataset).

kellyi:
@ajrobbins oh yeah, that is a placeholder, but since the column exists in the DRB data I'll just pull the actual value from there!

src/mmw/apps/modeling/views.py
kellyi: I set the header to cache for a week by returning this from the endpoint:

```python
return Response(json.dumps(point_source_results),
    headers={'Cache-Control': 'max-age: 604800'})
```

kellyi:
Realized the way I had set up the point source addition to the observations layer control tab was not guaranteed to update with the other observations if the point source layer returned first, so I pushed a fixup to wrap both of them in a `$.when` so that the app will call `self._update` after everything has returned properly.

src/mmw/apps/modeling/views.py
lewfish: Consider renaming this function to `get_drb_point_sources` or the like since we're not actually plotting anything here.

src/mmw/js/src/core/pointSourceLayer.js
```diff
@@ -0,0 +1,47 @@
...
+                        radius: markerSizesForZoomLevels[newZoomLevel]
+                    });
+                });
+                marker.bindPopup(view.render().el);
+            },
+        });
+    }
+};
+
+var PointSourcePopupView = Marionette.LayoutView.extend({
```
lewfish: I think this should be an ItemView since it doesn't compose a bunch of other views.

lewfish:
+1, tested. Nice job!

kellyi:
Thanks for taking a look! Going to make the changes you suggested before merging this in.

PR #1454, opened 2016-09-06
Fix Draw Tool For Windows Touch+Mouse Devices Using Chrome

arottersman:
We use `Leaflet.Draw` for our Draw Area --> Free Draw functionality. The library, however, has a bug where if a device has both touch and mouse input, an `onMouseUp` event will fire twice when the mouse is used. When this happens one click will draw two vertices one on top of the other, and it will be impossible to close the drawn polygon.

I've created a [fork of `Leaflet.Draw`](https://github.com/arottersman/Leaflet.draw/tree/arr/master) that addresses this problem (thanks @rajadain !), and updated the dependencies of Model My Watershed to point to it so we can test the fix.

## Testing
Pull this branch
```
vagrant ssh app
cd /src/mmw
// just incase npm doesn't update...
rm -f -r node_modules/leaflet-draw/
npm cache clean
npm install
// outside the VM
./scripts/bundle.sh --vendor
```

- In the browser, go to `localhost:8000`
- Draw Area --> Free Draw
- Confirm that you can create a polygon that closes and takes you to the analyse tab

The original bug was for a Windows touch+mouse device in Chrome, but we should also test IE, and Firefox, as well as performance on desktop, iOS, and android devices.

Connects #1190

### DIFF
--- a/src/mmw/npm-shrinkwrap.json
+++ b/src/mmw/npm-shrinkwrap.json
@@ -1507,8 +1507,8 @@
     },
     "leaflet-draw": {
       "version": "0.3.0",
-      "from": "leaflet-draw@0.3.0",
-      "resolved": "https://registry.npmjs.org/leaflet-draw/-/leaflet-draw-0.3.0.tgz"
+      "from": "git://github.com/azavea/Leaflet.draw",
+      "resolved": "git://github.com/azavea/Leaflet.draw"
     },
     "leaflet-plugins": {
       "version": "1.3.8",
--- a/src/mmw/package.json
+++ b/src/mmw/package.json
@@ -33,7 +33,7 @@
     "jshint": "2.8.0",
     "jstify": "0.9.0",
     "leaflet": "0.7.3",
-    "leaflet-draw": "0.3.0",
+    "leaflet-draw": "git://github.com/azavea/Leaflet.draw",
     "leaflet-plugins": "git://github.com/azavea/leaflet-plugins#feature/browserify",
     "leaflet.locatecontrol": "0.43.0",
     "livereload": "0.3.7",


mmcfarland:
This is working well for me.  On the Windows Touch device, I can use the mouse on all 3 browsers to draw and complete a polygon, which is the baseline we need.  Touch drawing works well, although it is generally harder to hit the last node - just because of fingers.  Regular desktop and Android touch-only also work as expected. 

I'm getting greedy touch events on Windows-touch Chrome, meaning that non-Leaflet elements no longer respond to touch.  I want to investigate that a bit more (see if it happens when someone else builds, it's not happening on @arottersman machine) and also test one more device.

Nice job!

rajadain:
I remember the greedy touch issue on Chrome, I thought @arottersman fixed it.

mmcfarland:
After testing more, it only seems my build is affected by greedy touch.  I'm happy with the behavior resulting from this PR, however, I think we should update the fork from your personal account to an azavea based one before we commit this.

arottersman:
There's already an azavea based Leaflet.draw fork. Should I create a new branch on it and apply my changes?

rajadain:
I would save the current changes there on a branch, and then make `master` on the Azavea fork reflect your work.

mmcfarland:
@rajadain I think that's a good plan.

rajadain:
Tested and found working in these environments:

 * [x] Chrome on Mac OS X El Capitan
 * [x] Firefox on Mac OS X El Capitan
 * [x] Safari on Mac OS X El Capitan
 * [x] Chrome on Windows 7 (non-touch device)
 * [x] Firefox on Windows 7 (non-touch device)
 * [x] Internet Explorer on Windows 7 (non-touch device)
 * [x] Chrome on Windows 10 (touch device) (mouse)
 * [x] Firefox on Windows 10 (touch device) (mouse)
 * [x] Internet Explorer on Windows 10 (touch device) (mouse)
 * [x] Edge on Windows 10 (touch device) (mouse)
 * [x] Chrome on Windows 10 (touch device) (finger)
 * [x] Firefox on Windows 10 (touch device) (finger)
 * [x] Internet Explorer on Windows 10 (touch device) (finger)
 * [x] Edge on Windows 10 (touch device) (finger)
 * [x] Safari on iPad 2
 * [x] Chrome on Android "Ice Cream Sandwich" Tablet
 * [x] Chrome on Android "Lollipop" Phone

Great work! 👏  I like the bigger hit targets in all environments, but I'm unsure if Leaflet.draw would like that in a PR. In any case, I think we should go ahead and use this as our fork.

rajadain:
And I did not observe the greedy touch behavior in any of these environments. That may have been a fluke on @mmcfarland's machine.

arottersman:
Great! Thanks! I've updated the project to point to the azavea fork which now has my changes on `master`, and its old state saved as branch.

Issue #1443, opened 2016-08-29
Stream rendering refinement


mmcfarland:
Increase the level at which new stream orders appear, generally following these guidelines:

```
At coast to coast scale, blue line density is perfect.
+ 1 zoom in (Kansas to Maine) renders more blue lines. This is also perfect.
+ 1 zoom in (Ohio to Mass) renders nothing new. This is also good.
+ 1 zoom in (PA to NJ/Long Island) renders nothing new. This is where I would like to see some new lines.
+ 1 zoom in (State College PA to NY City) renders new lines. It would be good to render these one zoom up (PA to NJ) and possibly render a few additional lines at this level.
+ 1 zoom in (Reading PA to New Brunswik NJ), renders nothing new. It would definitely be good to render more here.
+ 1 zoom in (Pottstown PA to Trenton NJ), renders new streams. I think this rendering level could be split into two levels, one rendering two levels up and another rendering one level up.
+ 1 zoom in (Pheonixville PA to Trenton NJ), renders nothing new. This density is not enough.
+ 1 zoom in (Valley Forge PA to Byn Athyn PA), renders nothing new. This density is not enough.
+ 1 zoom in (West Norriton to Whitemarsh), renders the last of the NHD medium resolution flowlines.
```

PR #1456, opened 2016-09-07
Adjust stream network tile rendering

caseypt:
Attempts to satisfy the revisions specified in #1443.

**Testing**
- Open a tab with this branch and another with staging.
- Turn on the stream network overlay.
- Zoom in on each, compare how the streams are rendered, and check that they satisfy the revisions set in #1443.

Connects to #1443

### DIFF
--- a/src/mmw/mmw/settings/layer_settings.py
+++ b/src/mmw/mmw/settings/layer_settings.py
@@ -126,7 +126,7 @@ LAYERS = [
         'minZoom': 9,
     },
     {
-        'code': 'nhd_streams_v1',
+        'code': 'nhd_streams_v2',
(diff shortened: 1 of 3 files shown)


mmcfarland:
Rendering looks good to me.  Could you update the names from `nhd_streams_v1 | drb_streams_v1` in the tiler and app configs to `***_streams_v2`?  That way we don't have to wait for the cloud front cache to expire before the client can see the changes on staging.

caseypt:
Sure. Will do.

caseypt:
Added a commit to bust the cache.

mmcfarland:
+1

Issue #1455, opened 2016-09-07
Rename "Code" to "NPDES Code" in Point Source Analysis tab


ajrobbins:
Renaming to "NPDES Code" will indicate the data source to users. The text should wrap as it does in the last three column headers.

![screen shot 2016-09-07 at 9 05 13 am](https://cloud.githubusercontent.com/assets/12868342/18312973/4dd3fbd8-74da-11e6-8e06-f84405d27e84.png)

PR #1461, opened 2016-09-09
Rename 'Code' -> `NPDES Code' in Point Source analysis tab.

kellyi:
This PR renames 'Code' to 'NPDES Code' in the Point Source analysis tab:

screen shot 2016-09-09 at 9 15 11 am

**Testing**
- get this branch, `vagrant up`, `./scripts/bundle.sh` then visit `localhost:8000` in the browser and open the dev console
- draw an AOI
- once the Point Source analyze results return, verify that the code column's header now reads "NPDES Code"
- verify that there aren't any errors in the console

Connects #1455

### DIFF
--- a/src/mmw/js/src/analyze/templates/pointSourceTable.html
+++ b/src/mmw/js/src/analyze/templates/pointSourceTable.html
@@ -1,7 +1,7 @@
 <table class="table custom-hover" data-toggle="table">
     <thead>
         <tr>
-            <th data-sortable="true">Code</th>
+            <th data-sortable="true">NPDES Code</th>
             <th data-sortable="true">City</th>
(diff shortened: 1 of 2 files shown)


rajadain:
As of #1451 we test that the tables show up correctly in the UI. You'll also have to change the test here: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/analyze/tests.js#L143

kellyi:
Ah, just found that! I'll dump the fixup and make the change to the test!

rajadain:
+1 tested. Sorting works correctly too.

Thanks to the test, this was one where we technically didn't need code review! If the test is passing that means that it is rendered correctly.

kellyi:
Thanks for taking a look!

Issue #1459, opened 2016-09-08
Remove Django Debug Toolbar


caseypt:
It appears that many people don't use this, and that it just gets in the way.

kdeloach:
:+1

PR #1462, opened 2016-09-09
Remove Django Debug Toolbar

mmcfarland:
The toolbar was underutilized by developers and started showing some
maintenance requirements so it was decided to remove it from the
project.

Connects #1459 

To test:
* `vagrant provision app`
* Load site and verify the toolbar has not been added to the page


kellyi:
+1. Toolbar is gone.

Here's where it sits on `develop`:

screen shot 2016-09-09 at 2 55 51 pm

and it's gone on this branch:

screen shot 2016-09-09 at 2 56 15 pm

Issue #1453, opened 2016-09-06
Analysis collapse button is missing


mmcfarland:
The button to toggle the sidebar is not visible as of 7ca369a6967bc8b5bd6887ea9611db292d206b13

![screenshot 2016-09-06 at 5 12 28 pm](https://cloud.githubusercontent.com/assets/1014341/18291336/445ca7c4-7455-11e6-8fd3-aea6b84ad157.png)

rajadain:
It would be really simple to just add the same gray button from the modeling view to this one. Would it be okay if I also did some styling adjustments? I was thinking something more like this:

![image](https://cloud.githubusercontent.com/assets/1430060/18391580/66286cc0-767c-11e6-94ee-2fa5b0dcff14.png)

![image](https://cloud.githubusercontent.com/assets/1430060/18391608/791fac9e-767c-11e6-8dfe-8e93cd00e526.png)

And the same for the Modeling view:

![image](https://cloud.githubusercontent.com/assets/1430060/18391664/b4e01188-767c-11e6-9f22-e3b35e6e6509.png)

![image](https://cloud.githubusercontent.com/assets/1430060/18391644/9fdf9dc6-767c-11e6-9cff-1adb631441a5.png)

Note that while this does improve styling overall, and places the button where I feel it should've been in the first place, this will change any "muscle memory" our users have developed for the unfortunate existing placement.

I am currently using `chevron-right` and `chevron-left`. Alternatively, we may use one of these:

![image](https://cloud.githubusercontent.com/assets/1430060/18391789/2af06ed6-767d-11e6-9fad-af4ce6ae3614.png)

rajadain:
In the "Draw" view that button will simply be absent.

PR #1463, opened 2016-09-09
Improved Sidebar Toggle Control

rajadain:
## Overview

This changes the sidebar collapse button to be a Leaflet control which conforms better to the app's visual style than the grey button previously. The button is also updated in the modeling view.

### Demo

| View | Before | After |
|-------------------|--------|-------|
| Analyze (Open) | image | image |
| Analyze (Closed) | - | image |
| Modeling (Open) | image | image |
| Modeling (Closed) | image | image |

### Notes

Since this moves the placement of the collapse button in the Modeling (Open) state, our users may have to retrain their muscle memory.

## Testing Instructions

Check out this branch and bundle dependencies:

```bash
./scripts/bundle.sh --debug
```

 * Open the app. Ensure there is no sidebar collapse control in the Draw view
 * Draw / Select an area of interest to move to the Analyze view. Ensure there _is_ a sidebar collapse button, that it works as expected, and has the right tooltip in both states.
 * Move back to Draw and ensure that the sidebar toggle button goes away.
 * Move back to Analyze and go forward to Modeling. Ensure that the button is still there and works as expected.
 * Hide the sidebar and move back to Analyze. Ensure that the sidebar stays collapsed and the button is in the right state.
 * Load an existing project (or go to a previously saved project's URL) to start directly in the Modeling view. Ensure that the button still works.

Connects #1453

### DIFF
--- a/src/mmw/apps/core/templates/base.html
+++ b/src/mmw/apps/core/templates/base.html
@@ -38,9 +38,6 @@
     </div>
 
     <div id="sidebar">
-        <button class="btn btn-primary tab-content-toggle" type="button" title="Show/Hide results sidebar">
-            <i class="fa fa-angle-right"></i>
-        </button>
         <div id="sidebar-content"></div>
     </div>
 
--- a/src/mmw/js/src/analyze/controllers.js
+++ b/src/mmw/js/src/analyze/controllers.js
@@ -44,6 +44,7 @@ var AnalyzeController = {
             // Modelling Controller.
             App.currentProject = null;
         }
+        App.getMapView().addSidebarToggleControl();
     },
 
     analyze: function() {
@@ -60,6 +61,7 @@ var AnalyzeController = {
 
     analyzeCleanUp: function() {
         App.rootView.sidebarRegion.empty();
+        App.getMapView().removeSidebarToggleControl();
     }
 };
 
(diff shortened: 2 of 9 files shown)


rajadain:
Also tagging @ajrobbins for feedback on design and button placement.

caseypt:
+1. Tested. Nice work.

Might be worth waiting for some design input before merging.

jfrankl:
Thanks for taking a first pass at this. Here is what I would like to see:

On any screen where a sidebar is present, add a button with the [fa-expand](http://fontawesome.io/icon/expand/) icon on the bottom right of the map. Clicking on this button will "expand the map", which really means moving the sidebar and expanding the map to take up the full screen. I think this icon makes more sense than the arrow when placed on the map because it is an action that is relative to the map (making it bigger).

image

On any screen where the map is expanded (and the sidebar is collapsed), change the icon on the button to [fa-compress](http://fontawesome.io/icon/compress/).

image

rajadain:
Of course! This makes more sense. Will change and notify.

rajadain:
Posted a fixup to change to expand / compress

image

jfrankl:
+1 looks great!

rajadain:
Squashed and rebased. Will merge once CI succeeds.

Issue #1445, opened 2016-08-29
Activate Point Source tooltip on table click


ajrobbins:
Following https://github.com/WikiWatershed/model-my-watershed/issues/1444, the following behavior has been requested: When a user clicks on an item in the Point Source analysis table, highlight the appropriate point on the map and open the tooltip.

caseypt:
If the user doesn't have the point-source layer on, should this turn on the layer and then open the tooltip?

PR #1465, opened 2016-09-12
Show point source analysis results on map

caseypt:
Hovering over the point source location id in the point source analysis results table will add a marker at the location of the point source, pan the map to that location, and open a popup with the information from the table for that source.

To make this possible, the latitude and longitude of each location are returned in the analysis results (the state was also added too).

Some modifications were made to the popup template to make the text cleaner in cases where there is no data for some of the header information. For example, I thought showing "Philadelphia" was better than showing "Philadelphia, No Data" for cases where the state is missing. Similarly, showing "No data" by itself on a line when there is no facility name was confusing. Instead, we render now render an empty string.

**Notes**
- For reasons that remain unknown, I was unable to attach an event handler in `PointSourceTableRowView`. I tried many different approaches, but nothing worked. I ended up attaching the events in the parent class (and using data attributes as opposed to directly using the model), which does have benefit of making it easier to manage one marker on the map.

**Testing**
- Draw an AoI that will include a few point sources outside of the DRB.
- When the analysis completes, activate the "Point Source" tab.
- Hovering over the NPDES code should add a marker to the map, pan to that location, and open a pop up with the information that is displayed in the table for that particular source.
- Repeat the above process with an AoI that is within the DRB. The behavior should be the same.

Connects to #1445

### DIFF
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -335,6 +335,7 @@ var AnimalTableView = Marionette.CompositeView.extend({
 
 var PointSourceTableRowView = Marionette.ItemView.extend({
     tagName: 'tr',
+    className: 'point-source',
     template: pointSourceTableRowTmpl,
 
     templateHelpers: function() {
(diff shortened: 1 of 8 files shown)


src/mmw/js/src/analyze/views.js
```diff
@@ -364,6 +368,41 @@ var PointSourceTableView = Marionette.CompositeView.extend({
...
+        this.marker = L.circleMarker(latLng, {
+            fillColor: "#ff7800",
+            weight: 0,
+            fillOpacity: 0.75
+        }).bindPopup(new pointSourceLayer.PointSourcePopupView({
+          model: new Backbone.Model(data)
+        }).render().el);
+
+        this.marker.addTo(map);
+        map.panTo(this.marker.getLatLng());
```
mmcfarland: I think we should ditch this and let the map only show the point if it happens to be in the viewport.  I found it pretty jarring to have the map zoom around both at high and low zoom levels.  Perhaps if the link is actually clicked, we then `panTo`?

src/mmw/js/src/analyze/views.js
```diff
@@ -364,6 +368,41 @@ var PointSourceTableView = Marionette.CompositeView.extend({
...
     onAttach: function() {
         $('[data-toggle="table"]').bootstrapTable();
+    },
+
+    ui: {
+        'pointSourceId': '.point-source-id'
+    },
+
+    events: {
+        'mouseover @ui.pointSourceId': 'showPointSourceMarker',
```
mmcfarland: I found it a little difficult to actually keep my mouse hovered over the link causing the point to flash on/off.  Given the suggestion about click/pan, how would you feel about the hover event being applied to the whole row to add it to the map (without pan)?

src/mmw/js/src/analyze/views.js
caseypt: Yeah, that makes sense to me. I'll make those changes.

src/mmw/js/src/analyze/views.js
caseypt: I'll give that a go.

caseypt:
Pushed a fixup commit to modify the behavior of the events.

caseypt:
@azaveaci  rebuild

Issue #1268, opened 2016-05-04
Prevent Drawing Very Small AoIs


rajadain:
See https://github.com/WikiWatershed/model-my-watershed/pull/1265#issuecomment-216648897

Similar to how we have a warning to users when the AoI is too big, similarly we may consider having a warning for AoIs that are too small. Investigate how small  is too small and implement that guard.

kellyi:
Wonder if it's worth also reducing the guard against very large aois, too. I think there's already a guard in place against that, but if we now have a sense of what area values are too large to use to get results from the model before Celery times out, we can make the other guard more specific.

Re: the small aoi guard: the issue I saw was a divide by zero error, so I imagine there's a way to check what's the lower bound for an aoi that would result in that error happening elsewhere.

Issue #1187, opened 2016-03-11
No warning or error message when analyze/model fail for very large watersheds


SRGDamia1:
When returning a very large watershed with RWD, the analyze or model may fail without warning.  I suspect it's because the area is just too big, but it would be nice to get a warning message to that effect if it is the case.

Example:  at 40.402026103094194, -74.97677564620972 analyze succeeds and gives a watershed of 17,275 km2, but modeling fails.  This is a huge chunk of the Delaware; I was surprised the analysis worked at all.  (It was slow, ~40 seconds.)  Clicking just slightly downstream, the analysis fails.
![screenshot](https://cloud.githubusercontent.com/assets/5374350/13709982/bd5cc590-e784-11e5-9d6a-938cb7755ab8.jpg)

SRGDamia1:
it also fails without warning if I select too large an area with the free-draw:
![screenshot](https://cloud.githubusercontent.com/assets/5374350/13710034/0c993152-e785-11e5-9ad1-8972689bebcf.jpg)
I do get an error if I try to draw an area bigger than 112700 km2.

mmcfarland:
We should tweak the `MAX_AREA` to be more consistent with the current
infrastructure's ability.

https://github.com/WikiWatershed/model-my-watershed/blob/3c28a492ec4151cf2a94a1e3847837556bb59db5/src/mmw/js/src/draw/views.js#L23

ajrobbins:
We've added front-end guards against too-large AoIs for analyze and model.

PR #1466, opened 2016-09-13
Guard against AOIs without enough land cover

kellyi:
This PR adds some backend guards and a specific frontend error message to guard against AOIs without enough land cover. Such AOIs include both shapes drawn entirely over water, like this one --

screen shot 2016-09-13 at 1 23 47 pm

-- as well as AOIs too small to have any land cover types, like this one:

screen shot 2016-09-13 at 1 25 08 pm

To accomplish this, I added two Exceptions in `tasks.py` thrown when either `z['Area']` is entirely zeroes or when the `nt_count` in the `nlcd_soils` method doesn't include any values. (This one had caused an error on small AOIs because it's used elsewhere as a denominator.

I've set these guards up to fail mapshed immediately and throw a specific error message; when the frontend sees this new error message, the code will now display "Selected area of interest doesn't include enough land cover" in place of "Error" in the error message.

One consequence of these exceptions is that the GMS file will fail to be created for these AOIs as well. In theory I think we could move these exceptions to the end of the processing steps and have it fail after a GMS file can be created, but I don't know whether the values in that file would be worthwhile. (I put up this PR for GWLF-E which might also handle them both -- https://github.com/WikiWatershed/gwlf-e/pull/71 -- but for the water aoi at least I think it would display results which may not be worthwhile/accurate).

**Testing**
- get this branch, `vagrant up`, `./scripts/bundle.sh` then visit localhost:8000 and open the browser console
- draw an AOI over water, then run a MapShed job and confirm that you see the specific error message
- draw a tiny AOI, then run MapShed and confirm that you see the same error message
- draw a larger AOI, then run MapShed and confirm that the results return and that you're able to export a GMS file

Connects #1268

### DIFF
--- a/src/mmw/apps/modeling/mapshed/tasks.py
+++ b/src/mmw/apps/modeling/mapshed/tasks.py
@@ -45,6 +45,7 @@ ANIMAL_KEYS = settings.GWLFE_CONFIG['AnimalKeys']
 ACRES_PER_SQM = 0.000247105
 HECTARES_PER_SQM = 0.0001
 SQKM_PER_SQM = 0.000001
+NO_LAND_COVER = 'NO_LAND_COVER'
 
 
 @shared_task(bind=True, default_retry_delay=1, max_retries=42)
@@ -155,6 +156,11 @@ def collect_data(geop_result, geojson):
     z['SedPhos'] = geop_result['sed_phos']
     z['Area'] = [percent * area * HECTARES_PER_SQM
                  for percent in geop_result['landuse_pcts']]
+
+    # Immediately return an error if z['Area'] is a list of 0s
+    if sum(z['Area']) == 0:
+        raise Exception(NO_LAND_COVER)
+
     z['UrbAreaTotal'] = sum(z['Area'][NRur:])
     z['PhosConc'] = phosphorus_conc(z['SedPhos'])
 
@@ -298,6 +304,10 @@ def nlcd_soils(sjs_result):
         raise Exception('[nlcd_soils] {}'.format(sjs_result['error']))
 
     ngt_count = parse_sjs_result(sjs_result)
+
+    # Raise exception if no NLCD values
+    if len(ngt_count.values()) == 0:
+        raise Exception(NO_LAND_COVER)
 
     # Split combined counts into separate ones for processing
     # Reduce [(n, g, t): c] to
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -217,7 +217,7 @@ var TabContentView = Marionette.LayoutView.extend({
         if (err && err.timeout) {
             tmvModel.setTimeoutError();
         } else {
-            tmvModel.setError();
+            tmvModel.setError('Error');
         }
 
         this.resultRegion.show(new coreViews.TaskMessageView({
--- a/src/mmw/js/src/core/models.js
+++ b/src/mmw/js/src/core/models.js
@@ -217,8 +217,8 @@ var TaskMessageViewModel = Backbone.Model.extend({
     message: null,
     iconClass: null,
 
-    setError: function() {
-        this.set('message', 'Error');
+    setError: function(message) {
+        this.set('message', message);
         this.set('iconClasses', 'fa fa-exclamation-triangle');
     },
 
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -803,9 +803,12 @@ var ResultsView = Marionette.LayoutView.extend({
             tmvModel = new coreModels.TaskMessageViewModel(),
             errorHandler = function(err) {
                 if (err && err.timeout) {
-                  tmvModel.setTimeoutError();
+                    tmvModel.setTimeoutError();
                 } else {
-                  tmvModel.setError();
+                    var message = err.error === 'NO_LAND_COVER' ?
+                        'Selected area of interest doesn\'t include any land ' +
+                        'cover to run the model' : 'Error';
+                    tmvModel.setError(message);
                 }
                 self.modelingRegion.show(new coreViews.TaskMessageView({ model: tmvModel }));
             };


src/mmw/js/src/core/models.js
```diff
@@ -217,8 +217,10 @@ var TaskMessageViewModel = Backbone.Model.extend({
     message: null,
     iconClass: null,
 
-    setError: function() {
-        this.set('message', 'Error');
+    setError: function(err) {
+        var message = err === 'No land cover' ? 'Selected area of interest ' +
```
mmcfarland: I like the specificity, but I think "...enough land cover data to run the model." shows that's it's a data problem and not a _land cover_ problem, per se.

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -45,6 +45,7 @@
 ACRES_PER_SQM = 0.000247105
 HECTARES_PER_SQM = 0.0001
 SQKM_PER_SQM = 0.000001
+NO_LAND_COVER = 'No land cover'
```
mmcfarland: If we're doing conditionals off of this text, better to make it look like a code I think (`no_land_cover`), so as not to be interpreted as a label and casually changed.

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -155,6 +156,11 @@ def collect_data(geop_result, geojson):
     z['SedPhos'] = geop_result['sed_phos']
     z['Area'] = [percent * area * HECTARES_PER_SQM
                  for percent in geop_result['landuse_pcts']]
+
+    # Immediately return an error for AOIs without enough land cover
+    if len(filter(lambda x: x != 0.0, z['Area'])) == 0:
```
rajadain: Stylistic choice, and might be inconsequential since these will never have negative values, but whenever checking that the value is not 0, I prefer `x > 0` to `x != 0`.

src/mmw/js/src/core/models.js
```diff
@@ -217,8 +217,10 @@ var TaskMessageViewModel = Backbone.Model.extend({
     message: null,
     iconClass: null,
 
-    setError: function() {
-        this.set('message', 'Error');
+    setError: function(err) {
+        var message = err === 'NO_LAND_COVER' ? 'Selected area of interest ' +
```
rajadain: Consider moving this logic outside `TaskMessageViewModel` so that it stays generic and unaware of GWLF-E business logic. Prefer something like:

```js
setError: function(message) {
  this.set('message', message || 'Error');
  this.set('iconClasses', 'fa fa-exclamation-triangle');
},
```

And supplying the detailed message in `ResultsView.onShow`

src/mmw/js/src/modeling/models.js
```diff
@@ -351,7 +351,7 @@ var ProjectModel = Backbone.Model.extend({
 
                     pollFailure: function(err) {
                         console.log('Failed to gather data required for MAPSHED');
-                        promise.reject(err);
+                        promise.reject(taskModel.get('error'));
```
rajadain: Consider reverting this change. The promise should faithfully reject with the original error, and any presentational labeling should happen at the very end. There may be other listeners depending on this promise's rejection, for example.

src/mmw/js/src/modeling/views.js
```diff
@@ -805,7 +805,7 @@ var ResultsView = Marionette.LayoutView.extend({
                 if (err && err.timeout) {
                   tmvModel.setTimeoutError();
                 } else {
-                  tmvModel.setError();
+                  tmvModel.setError(err);
```
rajadain: Consider moving the labeling logic to here.

src/mmw/js/src/core/models.js
```diff
@@ -217,8 +217,10 @@ var TaskMessageViewModel = Backbone.Model.extend({
...
+            'doesn\'t include enough land cover to run the model' : 'Error';
```
rajadain: Instead of saying "enough" (which implies some unstated magic threshold that the user must guess), we should say "any" because that is what we are checking for. The caught areas of interest don't have _any_ land cover.

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -312,6 +318,10 @@ def nlcd_soils(sjs_result):
         g2 = settings.SOIL_GROUP[g]
         ng_count[(n, g2)] = count + ng_count.get((n, g2), 0)
 
+    # Immediately return an error for AOIs without enough land cover
+    if len(nt_count.values()) == 0:
```
rajadain: Use `n_count` instead of `nt_count`, since we are looking for NLCD values only, not how they pair with soil texture.

Can we test `ngt_count` itself? It might be quicker to test that instead of until after destructuring it.

Also, try phrasing the comment without "AOI". This function simply processes a dictionary, and is not aware of concepts like area of interest. Perhaps something like "Raise exception if no NLCD values".

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -155,6 +156,11 @@ def collect_data(geop_result, geojson):
     z['SedPhos'] = geop_result['sed_phos']
     z['Area'] = [percent * area * HECTARES_PER_SQM
                  for percent in geop_result['landuse_pcts']]
+
+    # Immediately return an error for AOIs without enough land cover
+    if len(filter(lambda x: x > 0, z['Area'])) == 0:
```
rajadain: This is simpler:

```python
if sum(z['Area']) == 0:
  raise Exception(NO_LAND_COVER)
```

Is it equivalent? I think so ...

src/mmw/apps/modeling/mapshed/tasks.py
kellyi: Switched this to check `ngt_count.values()` and return immediately if the length of that's 0.

kellyi:
Pushed some fixups to incorporate the changes suggested above.

src/mmw/js/src/modeling/views.js
```diff
@@ -805,7 +805,10 @@ var ResultsView = Marionette.LayoutView.extend({
                 if (err && err.timeout) {
                   tmvModel.setTimeoutError();
                 } else {
-                  tmvModel.setError(err);
+                    var message = err.error === 'NO_LAND_COVER' ?
```
kellyi: I wonder if we might want to pass along the specific error message for other error/exceptions thrown from Django. I believe many of them currently have error messages set, but I don't know whether these were written to be meaningful for end users.

We could probably address https://github.com/WikiWatershed/model-my-watershed/issues/1187 by sending back a similar error message.

rajadain: Very large areas of interest should time out, which should result in this: https://github.com/kellyi/model-my-watershed/blob/f80396fcb52d7f421421d7060db179cc0ac44aa9/src/mmw/js/src/core/models.js#L225-L231. In case that is not happening, maybe because Spark Job Server times out before the front-end polling (unlikely but possible?), perhaps we should return a more descriptive error.

src/mmw/js/src/core/models.js
```diff
@@ -217,8 +217,8 @@ var TaskMessageViewModel = Backbone.Model.extend({
     message: null,
     iconClass: null,
 
-    setError: function() {
-        this.set('message', 'Error');
+    setError: function(message) {
+        this.set('message', message);
```
kellyi: Sounds good. I just pushed a fixup to pass the error message as an arg there.

rajadain:
Taking another look

src/mmw/js/src/analyze/views.js
```diff
@@ -217,7 +217,8 @@ var TabContentView = Marionette.LayoutView.extend({
         if (err && err.timeout) {
             tmvModel.setTimeoutError();
         } else {
-            tmvModel.setError();
+            var errorMessage = 'Error';
+            tmvModel.setError(errorMessage);
```
rajadain: Prefer inlining `'Error'`

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -155,6 +156,11 @@ def collect_data(geop_result, geojson):
     z['SedPhos'] = geop_result['sed_phos']
     z['Area'] = [percent * area * HECTARES_PER_SQM
                  for percent in geop_result['landuse_pcts']]
+
+    # Immediately return an error for AOIs without enough land cover
+    if sum(z['Area']) == 0:
+        raise Exception(NO_LAND_COVER)
```
kdeloach: Exceptional.

src/mmw/js/src/modeling/views.js
```diff
@@ -803,9 +803,12 @@ var ResultsView = Marionette.LayoutView.extend({
             tmvModel = new coreModels.TaskMessageViewModel(),
             errorHandler = function(err) {
                 if (err && err.timeout) {
-                  tmvModel.setTimeoutError();
+                    tmvModel.setTimeoutError();
```
kdeloach: Show that whitespace who's boss.

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -299,6 +305,10 @@ def nlcd_soils(sjs_result):
 
     ngt_count = parse_sjs_result(sjs_result)
 
+    # Raise exception if no NLCD values
+    if len(ngt_count.values()) == 0:
+        raise Exception(NO_LAND_COVER)
```
kdeloach: Hate when that happens.

kellyi:
Pushed a fixup to inline the Analyze error message and another to remove the "AOI" reference from the other comment I added to tasks.py.

rajadain:
+1! Friendly reminder to squash before merging.

PR #1467, opened 2016-09-15
Fix RWD bug by changing `drb_streams_v1` -> `v2`

kellyi:
We recently updated the code of `drb_streams_v1` to `drb_streams_v2`, but missed updating the code similarly in the `validateClickedPointWithinDRB` function. The outcome's that RWD is currently broken on staging; this PR fixes the error so that RWD will work again.

**Testing**
- get this branch, `vagrant up`, `./scripts/bundle.sh` and then visit `localhost:8000` and open the console
- use the RWD feature and verify that it works and that there are no errors in the console.


kellyi:
Thanks!

PR #1472, opened 2016-09-19
Add RWD data mounting instructions to README

kellyi:
This PR updates the README to include info and instructions for setting up RWD data, including instructions for mounting the data using an external drive on OS X by updating the Vagrantfile to adjust the Worker VM's synced dirs.

**Testing**
- get this branch, follow the instructions for mounting the RWD data via an external volume on OS X
- load the app and verify that the RWD feature works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,26 @@ $ ./scripts/manage.sh test
 envdir: fatal: unable to switch to directory /etc/mmw.d/env: access denied
 ```
 
+### Mounting RWD Data
+
+The worker VM mounts the `/opt/rwd-data` directory for use in the Rapid Watershed Delineation feature. You can set up a working development environment without the RWD data by making an empty `/opt/rwd-data` dir on your local machine.
+
+Linux users can mount RWD data from an external drive by symlinking the data from that drive to `/opt/rwd-data`. On OS X, developers can mount the data from an external drive by updating [the Worker's `synced_folder` in the Vagrantfile](https://github.com/WikiWatershed/model-my-watershed/blob/develop/Vagrantfile#L94) from
+
+```
+worker.vm.synced_folder "/opt/rwd-data", "/opt/rwd-data"
+```
+
+to
+
+```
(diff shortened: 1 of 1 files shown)


kdeloach:
@lewfish @rajadain Do you guys also need to do this?

rajadain:
I've never mounted the RWD data, so can't say.

lewfish:
I don't think I have the RWD data loaded on my computer. I never needed it since I wiped the drive in March. However, I do remember having a problem with using symlinks and Vagrant on OS X. Or maybe it was Docker.

README.md
```diff
@@ -101,6 +101,26 @@ $ ./scripts/manage.sh test
...
+to
+
+```
+worker.vm.synced_folder "/Volumes/<EXTERNAL-DRIVE-NAME>/<RWD-DATA-DIR>", "/opt/rwd-data"
+```
+
+where `<EXTERNAL-DRIVE-NAME>` is the name of the external drive and `<RWD-DATA-DIR>` is the name of the directory storing the `Main_Watershed` and `Subwatershed_ALL` directories.
+
+If you've already created a Worker VM, you'll need to run `vagrant reload --provision worker` to have the change take effect.
+
```
kdeloach: The command I use to symlink RWD data is:

```
sudo ln -s /media/passport/rwd /opt/rwd-data
```

Does this work for you?

In any case, consider updating `Vagrantfile` to read this path from an environment variable instead of suggesting that Mac users manually edit this file. Reading paths and port numbers from environment variables will become the standard approach for dealing with issues like this in the long term.

kellyi: Yeah tried that and I couldn't get it to work. Agreed on using an environment var, though. I'll set it up and push a follow up commit to this PR, then update the README again.

kellyi:
Going to close this PR to clear it out of the queue. If it becomes a hurdle again, we can create a new PR which includes the environment variable suggestion above.

Issue #1473, opened 2016-09-19
Upgrade GWLF-E


kellyi:
Following https://github.com/WikiWatershed/gwlf-e/issues/74 we should upgrade MMW to use the new version of GWLF-E

kellyi:
Shares a point with https://github.com/WikiWatershed/gwlf-e/issues/74

PR #1475, opened 2016-09-20
Upgrade gwlfe

kellyi:
This PR upgrades GWLF-E to use 0.6.2, the most recent release.

**Testing**
- grab this branch, `vagrant up`, then `vagrant reload --provision app`, `vagrant reload --provision worker`
- `vagrant ssh app`, then run `pip show gwlf-e` to see the install info and verify that it's version 0.6.2
- `vagrant ssh worker` then run `pip show gwlf-e` to see the install info and verify that it's version 0.6.2

To see it in action:
- pan to South Florida, turn on Congressional boundaries or similar, and click to analyze, then run MapShed
- confirm that you see something like this:

screen shot 2016-09-20 at 3 26 29 pm

or this:

screen shot 2016-09-20 at 3 27 55 pm

for the selected watershed when the MapShed job completes.

Connects #1473

### DIFF
--- a/src/mmw/requirements/base.txt
+++ b/src/mmw/requirements/base.txt
@@ -10,7 +10,7 @@ python-omgeo==1.7.2
 rauth==0.7.1
 djangorestframework-gis==0.8.2
 tr55==1.2.0
-gwlf-e==0.6.0
+gwlf-e==0.6.2
 requests==2.9.1
 rollbar==0.12.1
 retry==0.9.1


rajadain:
+1 tested. Can now see results in Miami:

image

rajadain:
Can you add link to the PR in the GWLF-E repo that fixes this?

kellyi:
Sure -- this is the GWLF-E fix:

https://github.com/WikiWatershed/gwlf-e/pull/73

rajadain:
Do we have any idea why this may happen for this area?

kellyi:
Something about the input GMS data led `z.Water` to trigger the `BasinWater` method before the `AdjUrbanQTotal` got set. If it skips that codepath then the value gets set a few lines later, so it may have been a bug only encountered with whatever data those South Florida watersheds were generating.

Issue #1470, opened 2016-09-16
Exception: Job xxx was ERROR (java.lang.Throwable: empty collection), deleted


mmcfarland:
View details in Rollbar: [https://rollbar.com/Azavea/MMW/items/96/](https://rollbar.com/Azavea/MMW/items/96/)

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "apps/modeling/tasks.py", line 96, in get_histogram_job_results
    histogram = histogram_finish(incoming['sjs_job_id'], self.retry)
  File "/usr/local/lib/python2.7/dist-packages/statsd/client.py", line 26, in wrapper
    return f(*args, **kw)
  File "apps/modeling/geoprocessing.py", line 164, in histogram_finish
    data = sjs_retrieve(host, port, job_id, retry)
  File "/usr/local/lib/python2.7/dist-packages/statsd/client.py", line 26, in wrapper
    return f(*args, **kw)
  File "apps/modeling/geoprocessing.py", line 124, in sjs_retrieve
    raise Exception('Job {} was {}, deleted'.format(job_id, status))
Exception: Job faf47062-6253-4ccb-a30e-e70548cbbbfe was ERROR (java.lang.Throwable: empty collection), deleted
```

jtarricone:
I think this might be due to spark quietly running into resource constraints. I started testing different changes to the configuration that would increase the number of partitions per CPU, the amount of memory available to the spark driver, etc. but actually getting the new configuration reflected in the environment was tedious so I didn't get to actually test whether performance was affected.

kellyi:
Some additional notes:

First occurrence was on July 4, 2016: https://rollbar.com/Azavea/MMW/items/96/occurrences/14353190288/

Looks like we released MapShed on June 29: https://github.com/WikiWatershed/model-my-watershed/releases/tag/1.13.0

kellyi:
I am able to recreate this error -- https://rollbar.com/Azavea/MMW/items/96/ -- on production by drawing a shape like this:

[...]

PR #1485, opened 2016-09-21
Prevent errors related to self-intersecting shapes by rejecting them on the frontend

kellyi:
This PR tries to avoid a few Rollbar errors by preventing users from drawing self-intersecting shapes on the front-end. To do so, it adds the [turf-kinks](http://turfjs.org/examples/turf-kinks/) library, then uses that to detect whether a drawn shape has any points where it intersects itself. If it does, then the frontend shows and error message and does not commence a call to the backend.

The specific Rollbar errors are these:

- https://rollbar.com/Azavea/MMW/items/96/
- https://rollbar.com/Azavea/MMW/items/119/
- https://rollbar.com/Azavea/MMW/items/96/

which I was able (and only able) to re-create by drawing self-intersecting shapes as described in my notes to #1470. It seems as if drawing a shape on production like this one:

screen shot 2016-09-21 at 4 28 24 pm

triggers the following error in Rollbar (which just appeared on trying to analyze that shape): https://rollbar.com/Azavea/MMW/items/96/occurrences/17010866915/

Drawing a similar shape on staging -- 

screen shot 2016-09-21 at 4 30 12 pm

-- immediately led to:

https://rollbar.com/Azavea/MMW/items/119/occurrences/17010903983/
&
https://rollbar.com/Azavea/MMW/items/68/occurrences/17010905660/

Now when a someone attempts to close a self-intersecting shape which would lead to an error on the API side, the front-end will show an alert like this:

screen shot 2016-09-21 at 4 24 33 pm

**Testing**
- grab this branch, `vagrant up`, `./scripts/bundle.sh --vendor`, then visit `localhost:8000`
- attempt to draw a self-intersecting shape and verify that you're unable to do so
- draw a non-self-intersecting shape and verify that it starts the Analyze process and that you're able to get results and don't see an error message.

Connects #1470

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -8,6 +8,7 @@ var $ = require('jquery'),
     turfBboxPolygon = require('turf-bbox-polygon'),
     turfDestination = require('turf-destination'),
     turfIntersect = require('turf-intersect'),
+    turfKinks = require('turf-kinks'),
     router = require('../router').router,
     App = require('../app'),
     utils = require('./utils'),
@@ -63,8 +64,15 @@ function validateRwdShape(result) {
 function validateShape(polygon) {
     var area = coreUtils.changeOfAreaUnits(turfArea(polygon), 'm<sup>2</sup>', 'km<sup>2</sup>'),
         d = new $.Deferred();
+    var selfIntersectingShape = turfKinks(polygon).features.length > 0;
 
-    if (area > MAX_AREA) {
+    if (selfIntersectingShape) {
+        var errorMsg = 'This watershed shape is invalid because it intersects ' +
+                       'itself. Try drawing the shape again without crossing ' +
+                       'over its own border.';
+        window.alert(errorMsg);
+        d.reject(errorMsg);
+    } else if (area > MAX_AREA) {
         var message = 'Sorry, your Area of Interest is too large.\n\n' +
                       Math.floor(area).toLocaleString() + ' km² were selected, ' +
                       'but the maximum supported size is currently ' +
--- a/src/mmw/package.json
+++ b/src/mmw/package.json
@@ -52,6 +52,7 @@
     "turf-destination": "1.2.1",
     "turf-erase": "1.3.2",
     "turf-intersect": "1.4.2",
+    "turf-kinks": "3.0.12",
     "underscore": "1.8.3",
     "watchify": "3.1.0",
     "zeroclipboard": "2.2.0"
--- a/src/mmw/bundle.sh
+++ b/src/mmw/bundle.sh
@@ -130,6 +130,7 @@ JS_DEPS=(backbone
          turf-destination
          turf-erase
          turf-intersect
+         turf-kinks
          underscore
          zeroclipboard)
 
--- a/src/mmw/js/src/draw/utils.js
+++ b/src/mmw/js/src/draw/utils.js
@@ -12,6 +12,7 @@ var polygonDefaults = {
 function drawPolygon(map, drawOpts) {
     var defer = $.Deferred(),
         tool = new L.Draw.Polygon(map, {
+            allowIntersection: false,
             shapeOptions: _.defaults(drawOpts || {}, polygonDefaults)
         }),
         clearEvents = function() {
--- a/src/mmw/npm-shrinkwrap.json
+++ b/src/mmw/npm-shrinkwrap.json
@@ -41,12 +41,12 @@
     },
     "bootstrap-select": {
       "version": "1.6.4",
-      "from": "../../tmp/npm-31030-7f32f0c4/1436971651298-0.12548802117817104/f1755efa102a41e43fc367ba36ce905b8f2b90e1",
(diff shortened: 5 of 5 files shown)


kellyi:
I was not sure what protocol was for committing changes to the `npm-shrinkwrap` file, so I left that as a commit on its own for now to potentially unstage/restage if it's better just to add the piece relevant to the single package.

src/mmw/js/src/draw/views.js
```diff
@@ -64,7 +65,13 @@ function validateShape(polygon) {
     var area = coreUtils.changeOfAreaUnits(turfArea(polygon), 'm<sup>2</sup>', 'km<sup>2</sup>'),
         d = new $.Deferred();
 
-    if (area > MAX_AREA) {
+    var selfIntersectingShape = turfKinks(polygon).features.length > 0;
```
mmcfarland: I like that this is small and self contained, but consider looking into the `L.Draw::allowIntersection` options which may prevent a user from drawing a self intersected polygon at all - it may be more intuitive than accepting the shape, but then rejecting it.

https://github.com/Leaflet/Leaflet.draw#draw-handler-options

src/mmw/js/src/draw/views.js
```diff
@@ -64,7 +65,13 @@ function validateShape(polygon) {
...
+
+    if (selfIntersectingShape) {
```
mmcfarland: RWD can produce this kind of polygon as well, based on your previous testing right?  If so, we may want to apply this validation to both paths.

src/mmw/js/src/draw/views.js
```diff
@@ -64,7 +65,13 @@ function validateShape(polygon) {
...
+        var message = 'Unable to analyze a self-intersecting shape.'
```
mmcfarland: A general language comment, I think this could be improved for the user with something like:  "This watershed shape is invalid because it intersects itself.  Try drawing the shape again without crossing over its own border."

src/mmw/js/src/draw/views.js
kellyi: Nice -- using that would prevent needing to import the additional package.

kellyi:
I dropped the extra commits and just used the the `allowIntersection: false` option for L.Draw. The default error message it provides is:

screen shot 2016-09-21 at 5 26 20 pm

And the docs suggest that you can customize the message with a handler: https://github.com/Leaflet/Leaflet.draw

Using the L.Draw option does not currently address the RWD shape issue, but I think those may have be prevented already because in cases when RWD creates self-intersecting shapes, I believe it generates `MultiPolygon`s which we're rejecting in another front-end check.

kellyi:
Oh, also just checked and with some persistent clicking, it *is* possible to circumvent the L.Draw error and create a shape with an intersection, so I'll re-add the commits to bring `turf-kinks` back in and then we can decide what we want to do with them.

screen shot 2016-09-21 at 5 32 31 pm

kellyi:
Ok: window alert will now show up if you manage you circumvent the L.Draw guard:

screen shot 2016-09-21 at 5 41 03 pm

src/mmw/js/src/draw/views.js
```diff
@@ -63,8 +64,14 @@ function validateRwdShape(result) {
 function validateShape(polygon) {
     var area = coreUtils.changeOfAreaUnits(turfArea(polygon), 'm<sup>2</sup>', 'km<sup>2</sup>'),
         d = new $.Deferred();
-
-    if (area > MAX_AREA) {
+    var selfIntersectingShape = turfKinks(polygon).features.length > 0
+
+    if (selfIntersectingShape) {
+        var errorMsg = 'This watershed shape is invalid because it intersects ' +
+            'itself. Try drawing the shape again without crossing over its own ' + 'border.'
```
rajadain: Consider formatting this so the message is left aligned, as done for `var message = 'Sorry, your Area of Interest is too large.'` below.

src/mmw/package.json
```diff
@@ -52,6 +52,7 @@
     "turf-destination": "1.2.1",
     "turf-erase": "1.3.2",
     "turf-intersect": "1.4.2",
+    "turf-kinks": "^3.0.12",
```
rajadain: By convention we pin to a specific version in this project.

src/mmw/js/src/draw/views.js
```diff
@@ -63,8 +64,14 @@ function validateRwdShape(result) {
 function validateShape(polygon) {
     var area = coreUtils.changeOfAreaUnits(turfArea(polygon), 'm<sup>2</sup>', 'km<sup>2</sup>'),
         d = new $.Deferred();
-
-    if (area > MAX_AREA) {
+    var selfIntersectingShape = turfKinks(polygon).features.length > 0
```
rajadain: missing semicolon here is causing lint to fail

mmcfarland:
Wasn't able to exercise the code, but I think this is a good solution. +1

src/mmw/package.json
kellyi: Updated this to remove the `^`.

Issue #956, opened 2015-10-15
Extra Geoprocessing Calls


jamesmcclain:
The same data requested from the geoprocessing endpoint which are used to construct the "survey" for the `/analyze` page could also be used to construct the `aoi_census` when the user transitions from the analyze page to the modeling page.

Since there are are two (identical) geoprocessing queries over the AoI done when the modeling page is entered (one for current conditions and one for the new scenario [in spite of caching because the `aoi_census` does not exist upon entry]), this would reduce the number of queries over the entire AoI from 3 to 1.

mmcfarland:
Is it truly the same survey since the analyze doesn't currently account for soils per query?  Not sure if that's technically true or not.

jamesmcclain:
There may be some information missing such that it is not possible to reconstruct the census from the survey, but I believe that the actual calls to the backend are the same (I could be mistaken).

mmcfarland:
Cool, definitely worth investigating.

kdeloach:
Is this done?

jamesmcclain:
No, solution got rejected.

mmcfarland:
I think we wanted to do it, but would like to invest in a refactor of project creation first, since this gets attached (or should) to it.  

If it doesn't look like we're going to get to that refactor, we should consider if having this reduced load on geop in the interim should override the initial decision.

mmcfarland:
See previous discussion here: https://github.com/WikiWatershed/model-my-watershed/pull/973

PR #1492, opened 2016-09-27
Reduce geoprocessor calls

jtarricone:
This is a minor refactor that reduces the number of calls to the geoprocessor by bundling the AoI census into the JSON blob returned by the `analyze` endpoint, and subsequently using it when instantiating projects and scenarios so that it can be POSTed to the API modeling endpoints. The durations of the blocking geoprocessor calls are not trivial, especially for large AoI's.

To test:
* clone the branch and bring up the environment
* go through typical user actions, e.g. select an AoI for analysis, create a TR55 model, add some scenarios
* keeping the test version of the app open, switch branches to `develop` or navigate the to the staging site and perform the same user actions, trying to keep all the parameters as close to identical as possible
* compare the two to make sure the output is identical, or if the parameters of the model are slightly different, that the differences can reasonably be ignored
* in general, the test version should be faster than the baseline, especially if the AoI is somewhat large (>10k sq-km)

Connects #956

### DIFF
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -13,7 +13,7 @@ from StringIO import StringIO
 from celery import shared_task
 
 from apps.modeling.geoprocessing import histogram_start, histogram_finish, \
-    data_to_survey, data_to_censuses
+    data_to_survey, data_to_census, data_to_censuses
 
 from apps.modeling.calcs import (animal_population,
                                  point_source_pollution)
@@ -108,17 +108,18 @@ def get_histogram_job_results(self, incoming):
 
 
 @shared_task
-def histogram_to_survey(incoming, aoi):
+def histogram_to_survey_census(incoming):
     """
     Converts the histogram results (aka analyze results)
-    to a survey of land use, which are rendered in the UI.
+    to a survey & census of land use, which are rendered in the UI.
     """
     pixel_width = incoming['pixel_width']
     data = incoming['histogram'][0]
-    results = data_to_survey(data)
-    convert_result_areas(pixel_width, results)
+    census = data_to_census(data)
+    survey = data_to_survey(data)
+    convert_result_areas(pixel_width, survey)
 
-    return results
+    return {'survey': survey, 'census': census}
 
 
 @shared_task
@@ -138,7 +139,7 @@ def analyze_animals(area_of_interest):
     """
     Given an area of interest, returns the animal population within it.
     """
-    return [animal_population(area_of_interest)]
+    return {'survey': [animal_population(area_of_interest)]}
 
 
 @shared_task
@@ -146,7 +147,7 @@ def analyze_pointsource(area_of_interest):
     """
     Given an area of interest, returns point sources of pollution within it.
     """
-    return [point_source_pollution(area_of_interest)]
+    return {'survey': [point_source_pollution(area_of_interest)]}
 
 
 def parse_single_ring_multipolygon(area_of_interest):
--- a/src/mmw/apps/modeling/views.py
+++ b/src/mmw/apps/modeling/views.py
@@ -303,7 +303,7 @@ def start_analyze(request, format=None):
              .set(exchange=exchange, routing_key=routing_key),
         tasks.get_histogram_job_results.s()
              .set(exchange=exchange, routing_key=routing_key),
-        tasks.histogram_to_survey.s(area_of_interest)
+        tasks.histogram_to_survey_census.s()
              .set(exchange=exchange, routing_key=choose_worker())
     ], area_of_interest, user)
 
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -277,7 +277,7 @@ var TabContentView = Marionette.LayoutView.extend({
 
     showResults: function() {
         var name = this.model.get('name'),
-            results = JSON.parse(this.model.get('taskRunner').get('result')),
+            results = JSON.parse(this.model.get('taskRunner').get('result')).survey,
             result = _.find(results, { name: name }),
             resultModel = new models.LayerModel(result),
             ResultView = AnalyzeResultViews[name];
--- a/src/mmw/js/src/modeling/controllers.js
+++ b/src/mmw/js/src/modeling/controllers.js
@@ -114,17 +114,44 @@ var ModelingController = {
             project = App.currentProject;
             project.set('model_package', modelPackage);
             itsiResetProject(project);
+            App.getMapView().addSidebarToggleControl();
+            setPageTitle();
         } else {
             var lock = $.Deferred();
-            project = makeProject(modelPackage, lock);
+            project = new models.ProjectModel({
+                name: 'Untitled Project',
+                created_at: Date.now(),
+                area_of_interest: App.map.get('areaOfInterest'),
+                area_of_interest_name: App.map.get('areaOfInterestName'),
+                model_package: modelPackage,
+                scenarios: new models.ScenariosCollection()
+            });
+
+            var analyzeTask = App.getAnalyzeCollection().findWhere({taskName:'analyze'});
+
             App.currentProject = project;
+            lock.resolve();
+
+            analyzeTask
+                .fetchAnalysisIfNeeded()
+                .done(function() {
+                    App.currentProject.set(
+                        'aoi_census',
+                        JSON.parse(
+                            App.getAnalyzeCollection()
+                                .findWhere({taskName:'analyze'})
+                                .get('result')
+                        ).census
+                    );
+                })
+                .fail(projectCleanUp);
 
             setupNewProjectScenarios(project);
             finishProjectSetup(project, lock);
             updateUrl();
+            App.getMapView().addSidebarToggleControl();
+            setPageTitle();
         }
-        App.getMapView().addSidebarToggleControl();
-        setPageTitle();
     },
 
     projectCleanUp: function() {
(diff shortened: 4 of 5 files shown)


src/mmw/js/src/modeling/controllers.js
```diff
@@ -94,7 +94,7 @@ var ModelingController = {
 
                     App.currentProject = project;
                     if (!App.projectNumber) {
-                        setupNewProjectScenarios(project);
+                        setupNewProjectScenarios(project, project.get('aoi_census'));
```
rajadain: Since the information is contained within `project`, we don't need to send `project.get('aoi_census')` as a second parameter here, we can just send `project` and have `setupNewProjectScenarios` extract it. Unless there are cases when this is supplied from outside `project`?

kellyi:
`./scripts/debugcelery.sh` is showing an error when trying to Analyze this AOI:

screen shot 2016-09-27 at 3 46 25 pm

Here's the error:

```sh
[2016-09-27 15:46:22,842: ERROR/MainProcess] Task apps.modeling.tasks.histogram_to_survey_census[a3b72e39-0a3d-4e35-94f1-c5360c56b151] raised unexpected: TypeError('histogram_to_survey_census() takes exactly 1 argument (2 given)',)
TypeError: histogram_to_survey_census() takes exactly 1 argument (2 given)
[2016-09-27 15:46:22,847: ERROR/Worker-1] Task a3b72e39-0a3d-4e35-94f1-c5360c56b151 run from job 1596 raised exception: histogram_to_survey_census() takes exactly 1 argument (2 given)
TypeError: histogram_to_survey_census() takes exactly 1 argument (2 given)
```

kellyi:
(Looks like I'm encountering the same area for other AOIs, too.)

src/mmw/apps/modeling/tasks.py
```diff
@@ -108,17 +108,18 @@ def get_histogram_job_results(self, incoming):
 
 
 @shared_task
-def histogram_to_survey(incoming, aoi):
+def histogram_to_survey_census(incoming):
```
rajadain: You need to keep two parameters here. The first parameter is automatically supplied by the previous task in the Celery chain. The second argument is the area of interest specified in views.py:306 below.

src/mmw/apps/modeling/views.py
```diff
@@ -303,7 +303,7 @@ def start_analyze(request, format=None):
              .set(exchange=exchange, routing_key=routing_key),
         tasks.get_histogram_job_results.s()
              .set(exchange=exchange, routing_key=routing_key),
-        tasks.histogram_to_survey.s(area_of_interest)
+        tasks.histogram_to_survey_census.s(area_of_interest)
```
rajadain: Or rather, remove the `area_of_interest` parameter from here.

jtarricone:
My bad, it was a last-minute fix to satisfy the linter, only half the change was committed.

src/mmw/js/src/analyze/views.js
```diff
@@ -237,7 +237,7 @@ var TabContentView = Marionette.LayoutView.extend({
 
     showResults: function() {
         var name = this.model.get('name'),
-            results = JSON.parse(this.model.get('taskRunner').get('result')),
+            results = JSON.parse(this.model.get('taskRunner').get('result')).survey,
```
rajadain: While the `analyze` endpoint now returns both census and survey, the other endpoints `analyze/animals` and `analyze/pointsources` don't, causing those tables in the Analyze tab to be empty even when there is real data.

rajadain:
If I try and start TR-55 before the Analyze phase completes, the app crashes:

image

This is behavior we have long supported.

src/mmw/js/src/modeling/models.js
```diff
@@ -135,7 +135,8 @@ var ProjectModel = Backbone.Model.extend({
         is_activity: false,             // Project that persists across routes
         gis_data: null,                 // Additionally gathered data, such as MapShed for GWLF-E
         needs_reset: false,             // Should we overwrite project data on next save?
-        allow_save: true                // Is allowed to save to the server - false in compare mode
+        allow_save: true,               // Is allowed to save to the server - false in compare mode
+        aoi_census: null                // JSON blob
```
rajadain: Try to phrase the comment to describe what is being stored, rather than how.

src/mmw/js/src/modeling/controllers.js
jtarricone: Good point.

jtarricone:
I committed a couple fixups that should address all points. I'm not sure how necessary it ultimately is, but I also did the round of tests as a logged-in user to make sure project state persisted and launched correctly, which appeared to be the case.

rajadain:
Taking another look.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -114,15 +114,37 @@ var ModelingController = {
             project = App.currentProject;
             project.set('model_package', modelPackage);
             itsiResetProject(project);
+            App.getMapView().addSidebarToggleControl();
```
rajadain: Why is this needed here when it also exists on line 148?

src/mmw/js/src/modeling/controllers.js
```diff
@@ -114,15 +114,37 @@ var ModelingController = {
...
+                area_of_interest: App.map.get('areaOfInterest'),
+                area_of_interest_name: App.map.get('areaOfInterestName'),
+                model_package: modelPackage,
+                scenarios: new models.ScenariosCollection()
+            });
+
             App.currentProject = project;
+            lock.resolve();
+
+            App.getAnalyzeCollection().findWhere({taskName:'analyze'}).fetchAnalysisIfNeeded()
```
rajadain: Since `App.getAnalyzeCollection().findWhere({ taskName: 'analyze' })` is used multiple times, consider assigning it to a variable like `landSoilAnalyzeTask`.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -114,15 +114,37 @@ var ModelingController = {
...
+            .done(function() {
```
rajadain: Consider indenting the `.done` and `.fail` blocks one level to improve readability.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -114,15 +114,37 @@ var ModelingController = {
...
+                App.currentProject.set(
+                    'aoi_census',
+                    JSON.parse(
+                        App.getAnalyzeCollection().findWhere({taskName:'analyze'}).get('result')
+                    ).census
+                );
+            })
+            .fail(function() { projectCleanUp(); });
```
rajadain: Prefer fully indented blocks, even when the body is a single statement:

```js
.fail(function() {
    projectCleanup();
});
```

Or even better, simply pass the handler directly rather than wrapping it in a function:

```js
.fail(projectCleanup);
```

src/mmw/js/src/modeling/controllers.js
jtarricone: Good catch, I think that's redundant; I tried to totally separate ITSI/non-ITSI blocks once the project state calisthenics started getting awkward, this is a leftover from that.

kellyi:
This still seems to be causing an error returning the data for the analysis results. For AOIs I've checked I'm getting back no data for all the analysis tabs:

screen shot 2016-09-29 at 1 04 05 pm

kellyi:
HM, the above seems to be an issue with my local because it's happening on `develop` too.

kellyi:
Ok, turns out I had not run `./scripts/bundle.sh` on this branch, and I had merged this one into my local `develop`. Checked again on this one after running `bundle` and the results are outputting correctly.

src/mmw/js/src/modeling/controllers.js
```diff
@@ -114,15 +114,37 @@ var ModelingController = {
...
+
             App.currentProject = project;
+            lock.resolve();
+
+            App.getAnalyzeCollection().findWhere({taskName:'analyze'}).fetchAnalysisIfNeeded()
+            .done(function() {
+                App.currentProject.set(
+                    'aoi_census',
+                    JSON.parse(
+                        App.getAnalyzeCollection().findWhere({taskName:'analyze'}).get('result')
```
kellyi: Would it be worthwhile to indent this like:

```js
App.getAnalyzeCollection()
       .findWhere({...})
       .get(...)
```

kellyi:
+1. Tested this with a variety of AOIs and it worked. 

Looks like the failing CI build's related to the Postgres update; there's a fix for it on `release` but not on `develop` yet, so it might be worth waiting til the release has been made and merged into `develop`, then rebasing these changes on that so that the tests will pass. Here's the error:

```
TASK [azavea.postgresql-support : Install client API libraries for PostgreSQL] *
Thursday 29 September 2016  15:42:13 -0400 (0:00:00.424)       0:00:06.772 **** 
fatal: [services]: FAILED! => {"cache_update_time": 0, "cache_updated": false, "changed": false, "failed": true, "msg": "'/usr/bin/apt-get -y -o \"Dpkg::Options::=--force-confdef\" -o \"Dpkg::Options::=--force-confold\"     install 'libpq-dev=9.5.*.pgdg14.04+1'' failed: E: Version '9.5.*.pgdg14.04+1' for 'libpq-dev' was not found\n", "stderr": "E: Version '9.5.*.pgdg14.04+1' for 'libpq-dev' was not found\n", "stdout": "Reading package lists...\nBuilding dependency tree...\nReading state information...\n", "stdout_lines": ["Reading package lists...", "Building dependency tree...", "Reading state information..."]}
```

hectcastro:
@azavea-bot rebuild

Issue #1486, opened 2016-09-22
Reject too-small AoIs on the front end


ajrobbins:
https://github.com/WikiWatershed/model-my-watershed/pull/1466 Adds a front end error message when a too-small AoI s modeled with Mapshed. We should throw this error earlier in the process (during the analyze step, or as soon as Mapshed is selected for modeling).

ajrobbins:
Reject these AoIs on the front end to avoid https://rollbar.com/Azavea/MMW/items/122/

PR #1493, opened 2016-09-27
Guard frontend AOIs without land cover

kellyi:
This PR adds a frontend guard to show an error alert and prevent running MapShed when the analysis results have indicated that the area doesn't include any land cover:

screen shot 2016-09-27 at 2 52 38 pm

This works by performing a check on the analysis results when the "Watershed Multi Year Model" button has been clicked: if there are analysis results and if the total land cover sums to 0 in them, then instead of running MapShed the app will now show an alert and return.

The guard will only work when the analysis results have returned with data; if the analysis is still underway, it's possible to click to run a MapShed job (and the code path will skip the guard block). I decided against creating a block on running MapShed until the analysis results had returned, as it's likely there are situations when one would want to skip the analyze step, and there wouldn't be anything to check for the guard anyway. In other words, we assume that the shape does include land cover until it's known that it doesn't, and any shapes with no land cover which slip through will be rejected by MapShed on the backend.

**Testing**
- get this branch, `vagrant up`, `./scripts/bundle.sh` then visit localhost:8000 and open the dev console
- select a 1 sq km AOI in Delaware Bay to analyze and wait for the results to return. Once they're in, try to kick off a MapShed job and confirm that the alert window pops up -- and that the MapShed job doesn't start
- select a different 1 sq km AOI in Delaware Bay to analyze. This time don't wait for the results to return, but instead click to start a MapShed job. Confirm that it has started and that you eventually get an error message issued from MapShed.
- zoom in very closely and draw the smallest AOI you can ( < 150 sq m if possible). Wait for the analysis results to return and confirm that if the land cover total is 0, attempting to run a MapShed job will lead to the error alert message (and not to running the MapShed job)
- select a usual AOI, like a congressional district, a sq km over land, or a HUC-12, then when the analysis results return, confirm that you can start a MapShed job

Connects #1486

### DIFF
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -55,7 +55,29 @@ var ResultsView = Marionette.LayoutView.extend({
             modelPackageName = $(e.target).data('id'),
             modelPackage = _.find(modelPackages, {name: modelPackageName}),
             newProjectUrl = '/project/new/' + modelPackageName,
-            projectUrl = '/project';
+            projectUrl = '/project',
+            analysisResults = JSON.parse(App.getAnalyzeCollection()
+                                            .findWhere({taskName: 'analyze'})
+                                            .get('result')),
+            landResults = _.find(analysisResults, function(element) {
+                    return element.name === 'land';
+            });
+
+        if (landResults) {
+            var landCoverTotal = _.sum(_.map(landResults.categories,
+                    function(category) {
+                        if (category.type === 'Open Water') {
+                            return 0;
+                        }
+                        return category.area;
+                    }));
+
+            if (landCoverTotal === 0) {
+                window.alert("The selected Area of Interest doesn't " +
+                             "include any land cover to run the model.");
+                return;
+            }
+        }
 
         if (!modelPackage.disabled) {
             if (settings.get('itsi_embed') && App.currentProject && !App.currentProject.get('needs_reset')) {


src/mmw/js/src/analyze/views.js
```diff
@@ -57,6 +57,33 @@ var ResultsView = Marionette.LayoutView.extend({
             newProjectUrl = '/project/new/' + modelPackageName,
             projectUrl = '/project';
 
+        if (modelPackageName === 'gwlfe') {
+            var analysisResults = JSON.parse(App.getAnalyzeCollection()
+                                                .findWhere({taskName: 'analyze'})
+                                                .get('result'));
+
+            var landResults = _.find(analysisResults, function(element) {
```
rajadain: In MMW there is a convention of grouping variable declarations together in JavaScript. Thus, prefer:

```js
var analysisResults = JSON.parse(App.getAnalyzeCollection()
                                    .findWhere({taskName: 'analyze'})
                                    .get('result')),
    landResults = _.find(analysisResults, function(element) {
        return element.name === 'land';
    });
```

src/mmw/js/src/analyze/views.js
```diff
@@ -57,6 +57,33 @@ var ResultsView = Marionette.LayoutView.extend({
...
+                return element.name === 'land';
+            });
+
+            if (landResults) {
+                var landAreaResults = landResults.categories;
+
+                var landCoverTotal = _.reduce(_.filter(landAreaResults,
```
rajadain: Same here, consider grouping these variable declarations into one statement.

rajadain: Prefer `_.sum(_.map())` over `_.reduce(_.filter())` which would be easier to follow.

src/mmw/js/src/analyze/views.js
```diff
@@ -57,6 +57,33 @@ var ResultsView = Marionette.LayoutView.extend({
...
+                    function(category) {
+                        return category.type !== 'Open Water';
+                    }), function(sum, category) {
+                        return sum + category.area;
+                    }, 0);
+
+                if (landCoverTotal === 0) {
+                    window.alert('The selected Area of Interest doesn\'t ' +
```
rajadain: You wouldn't have to escape the apostrophe if you quoted the string in double-quotes. Just sayin'

src/mmw/js/src/analyze/views.js
```diff
@@ -57,6 +57,33 @@ var ResultsView = Marionette.LayoutView.extend({
             newProjectUrl = '/project/new/' + modelPackageName,
             projectUrl = '/project';
 
+        if (modelPackageName === 'gwlfe') {
```
rajadain: If there is no land cover, should we also stop in case of TR-55? @ajrobbins @mmcfarland

rajadain:
+1 works as advertised

src/mmw/js/src/analyze/views.js
mmcfarland: Good question.  I was a bit confused by the terminology used.  "No Land Cover" seemed to me to mean `NODATA` for that particular raster set, but it instead the code seems to mean, "more than just Open Water land cover type".  It appears that gwlfe requires more than 100% Open Water, if TR55 is the same, we should probably treat it the same.  It's possible that it just reports 100% runoff in that scenario.

kellyi: Looks like TR-55 fails on very small AOIs too. Trying to model this one:

screen shot 2016-09-28 at 3 41 23 pm

led to this error:

```sh
[2016-09-28 15:40:44,123: ERROR/Worker-1] Bad input data to TR55: u'runoff-vol'
```

kellyi: A 1 sq km AOI in the middle of Delaware Bay returns this:

screen shot 2016-09-28 at 3 43 13 pm

&

screen shot 2016-09-28 at 3 43 38 pm

kellyi:
Cool. I pushed a fixup to adjust the variable assignment formatting, replace `reduce/filter` with `sum/map`, and remove the `gwlfe` guard so that the check will also be applicable to TR-55.

src/mmw/js/src/analyze/views.js
```diff
@@ -55,7 +55,30 @@ var ResultsView = Marionette.LayoutView.extend({
...
+            analysisResults = JSON.parse(App.getAnalyzeCollection()
+                                            .findWhere({taskName: 'analyze'})
+                                            .get('result')),
+            landResults = _.find(analysisResults, function(element) {
+                    return element.name === 'land';
+            });
+
+        if (landResults) {
+            var landAreaResults = landResults.categories,
+                landCoverTotal = _.sum(_.map(landAreaResults,
```
rajadain: Since `landAreaResults` is used only once here, and we're using `category` below, consider just using `landResult.categories` here and save an indirection.

kellyi:
I made that `landResults.categories` change noted above, squashed the fixup commits, and pushed the updated single commit. Thanks for taking a look at this!

Issue #1476, opened 2016-09-21
Exception: Unable to submit job to Spark JobServer.
Details = {
  "status": "ERROR",
  "result": "context geoprocessing not found"
}


ajrobbins:
View details in Rollbar: [https://rollbar.com/Azavea/MMW/items/118/](https://rollbar.com/Azavea/MMW/items/118/)

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "apps/modeling/tasks.py", line 73, in start_histogram_job
    'sjs_job_id': histogram_start([json.dumps(polygon)], self.retry)
  File "/usr/local/lib/python2.7/dist-packages/statsd/client.py", line 26, in wrapper
    return f(*args, **kw)
  File "apps/modeling/geoprocessing.py", line 146, in histogram_start
    return sjs_submit(host, port, args, data, retry)
  File "/usr/local/lib/python2.7/dist-packages/statsd/client.py", line 26, in wrapper
    return f(*args, **kw)
  File "apps/modeling/geoprocessing.py", line 76, in sjs_submit
    'Details = {}'.format(response.text))
Exception: Unable to submit job to Spark JobServer.
Details = {
  "status": "ERROR",
  "result": "context geoprocessing not found"
}
```

ajrobbins:
This seems to be happening the first time something is requested after a new deployment.

hectcastro:
Attempts at mitigating this could consist of:

- Checking to see if the geoprocessing context exists before submitting a task
- Enhancing the health check used for the workers

kellyi:
I was able to reproduce this error on my local by deleting the `geoprocessing` context with `httpie`:

```sh
http DELETE :8090/contexts/geoprocessing
```

Visiting http://localhost:8090/contexts/ to see the list of available contexts then just showed an empty array, and when I tried to analyze an AOI I got this error message in `debugcelery.sh`:

```sh
[...]

PR #1496, opened 2016-09-28
Guard missing Spark Job Server geoprocessing context

kellyi:
This PR addresses an error whereby a job submitted to Spark Job Server would occasionally fail because it could not find the "geoprocessing" context. The changes here add a check to `sjs_submit` to see whether a post request to SJS has failed with a message of 'context geoprocessing not found'. If so, we now make a put request to reset the geoprocessing service's contexts and -- if it's successful -- we try re-submitting the job.

**Testing**
- get this branch, `vagrant up`, then use httpie or curl to delete the geoprocessing context from SJS:

```sh
http DELETE :8090/contexts/geoprocessing
```

- check the service's contexts by making a get request to `:8090/contexts/` and verify that it's an empty array

- add this print statement at line 76 of `src/mmw/apps/modeling/geoprocessing.py`:

```py
print('error => {}'.format(error))
```
- run `./scripts/debugcelery.sh`
- open `localhost:8000` in the browser and select a 1 sq km area of interest
- switch back to the `debugcelery.sh` window and confirm that you see the print statement reporting the error but that the job continues and completes instead of throwing an Exception
- once the job's complete, check the frontend to see that results have been returned
- delete the geoprocessing context again with httpie or curl:

```sh
http DELETE :8090/contexts/geoprocessing
```

- run a MapShed job on your AOI, verify that you see the print statement above again in `debugcelery`, and verify that the job eventually completes and doesn't create an error

Connects #1476

### DIFF
--- a/src/mmw/apps/modeling/geoprocessing.py
+++ b/src/mmw/apps/modeling/geoprocessing.py
@@ -58,8 +58,9 @@ def sjs_submit(host, port, args, data, retry=None):
     Submits a job to Spark Job Server. Returns its Job ID, which
     can be used with sjs_retrieve to get the final result.
     """
-    url = 'http://{}:{}/jobs?{}'.format(host, port, args)
-    response = requests.post(url, data=json.dumps(data))
+    base_url = 'http://{}:{}'.format(host, port)
+    jobs_url = '{}/jobs?{}'.format(base_url, args)
+    response = requests.post(jobs_url, data=json.dumps(data))
 
     if response.ok:
         job = response.json()
@@ -71,6 +72,21 @@ def sjs_submit(host, port, args, data, retry=None):
             except MaxRetriesExceededError:
                 raise Exception('No slots available in Spark JobServer.\n'
                                 'Details = {}'.format(response.text))
+        elif error['result'] == 'context geoprocessing not found':
+            reboot_sjs_url = '{}/contexts?reset=reboot'.format(base_url)
+            context_response = requests.put(reboot_sjs_url)
+
+            if context_response.ok:
+                if retry:
+                    retry()
+                else:
+                    raise Exception('Geoprocessing context missing in Spark'
+                                    'JobServer, but no retry was set.\n'
+                                    'Details = {}'.format())
+            else:
+                raise Exception('Unable to create missing geoprocessing '
+                                'context in Spark JobServer.\n'
+                                'Details = {}'.format(context_response.text))
         else:
             raise Exception('Unable to submit job to Spark JobServer.\n'
                             'Details = {}'.format(response.text))


src/mmw/apps/modeling/geoprocessing.py
```diff
@@ -71,6 +72,16 @@ def sjs_submit(host, port, args, data, retry=None):
             except MaxRetriesExceededError:
                 raise Exception('No slots available in Spark JobServer.\n'
                                 'Details = {}'.format(response.text))
+        elif error['result'] == 'context geoprocessing not found':
+            geop_context_url = '{}/contexts?reset=reboot'.format(base_url)
+            context_response = requests.put(geop_context_url)
```
kellyi: An alternate possibility here would be to send this to Spark Job Server to create the "geoprocessing" context:

```py
geop_context_url = '{}/contexts/geoprocessing'.format(base_url)
context_response = requests.post(geop_context_url, data='')
```

Instead of resetting the contexts back to the initial config, this would create the "geoprocessing" context.

rajadain: Rebooting might be safer, because we're reusing the default context definition, so if we ever change its name we won't have to change it in multiple places. This also protects against other errors that could have happened: it is generally cleaner to reboot the whole thing than to try and piece it back together.

src/mmw/apps/modeling/geoprocessing.py
```diff
@@ -71,6 +72,16 @@ def sjs_submit(host, port, args, data, retry=None):
...
+
+            if context_response.ok:
+                retry()
+            else:
+                raise Exception('Unable to create missing geoprocessing\n'
```
rajadain: Consider leaving out the `\n` in between the message, so the message is the first line and the details the second. The output should look like:

```
Unable to create missing geoprocessing context in Spark JobServer.
Details = [...]
```

src/mmw/apps/modeling/geoprocessing.py
```diff
@@ -71,6 +72,16 @@ def sjs_submit(host, port, args, data, retry=None):
             except MaxRetriesExceededError:
                 raise Exception('No slots available in Spark JobServer.\n'
                                 'Details = {}'.format(response.text))
+        elif error['result'] == 'context geoprocessing not found':
+            geop_context_url = '{}/contexts?reset=reboot'.format(base_url)
+            context_response = requests.put(geop_context_url)
+
+            if context_response.ok:
+                retry()
```
rajadain: Consider checking that `retry` exists. It defaults to `None`.

rajadain:
I just tried deleting the geoprocessing context and then triggering a MapShed job. I was afraid that rebooting would take too long, and since many calls to `sjs_submit` are made in quick succession (almost in parallel), they would all try to reboot SJS. But in my tests the first one reboots it and the rest pick it up almost immediately, which is great.

I'll see if I can think of any more edge cases, but this is looking really good. Good job!

src/mmw/apps/modeling/geoprocessing.py
```diff
@@ -71,6 +72,21 @@ def sjs_submit(host, port, args, data, retry=None):
...
+
+            if context_response.ok:
+                if retry:
+                    retry()
+                else:
+                    raise Exception('Geoprocessing context missing in Spark'
+                                    'JobServer, but no retry was set.\n'
+                                    'Details = {}'.format())
+            else:
+                raise Exception('Unable to create missing geoprocessing'
```
rajadain: You may need a space at the end of `geoprocessing` else it will be joined with `context` on the next line:

```python
In [1]: print('Unabe to create missing geoprocessing'
   ...: 'context in Spark JobServer.\n')
Unabe to create missing geoprocessingcontext in Spark JobServer.

```

rajadain:
This looks good to me. 👍

src/mmw/apps/modeling/geoprocessing.py
rajadain: Consider renaming `geop_context_url` to `reboot_sjs_url` or similar to make it more apparent what the code is trying to do here.

kellyi: Sounds good. I'll make that change to include in the final commit.

kellyi:
This already got put on the `release/1.15.0` branch, then was merged back into `develop` in https://github.com/WikiWatershed/model-my-watershed/commit/90edb189e99848ce262f778996648a212ee603e3

Going to close this PR out.

Issue #1495, opened 2016-09-28
Add site name to Point Source popup


ajrobbins:
When a NPDES code is clicked in the point source table, the popup appears without a site name. When the same point is clicked on the map, the site name is included. We should include the site name in the first case popup as well.

Table click:
![screen shot 2016-09-28 at 10 57 16 am](https://cloud.githubusercontent.com/assets/12868342/18919017/af24abd0-856a-11e6-804c-ed761041078e.png)

Map point click:
![screen shot 2016-09-28 at 10 57 41 am](https://cloud.githubusercontent.com/assets/12868342/18919026/b743db92-856a-11e6-8c1e-54ae6d873210.png)

ajrobbins:
Shares a point with https://github.com/WikiWatershed/model-my-watershed/issues/1494

PR #1497, opened 2016-09-28
Add `facilityname` to dynamic point source popups within DRB

kellyi:
This PR adds the `facilityname` to the point source pop ups created dynamically when pointing the mouse over one of the point source table rows for an AOI within the DRB. It turned out we were not pulling in the facilityname column's value from the endpoint the analysis was using, so I added it to the the results for AOIs within the DRB -- and added a null column as a placeholder for AOIs outside the DRB, since the `ms_pointsource` table does not currently include a `facilityname` column.

This means that the markers for point source locations within the DRB will now display the facility name on mouseover-ing the row:

screen shot 2016-09-28 at 3 18 44 pm

... and those outside the DRB should render correctly without displaying the facility name:

screen shot 2016-09-28 at 3 21 16 pm

**Testing**
- get this branch, `vagrant up`, `./scripts/bundle.sh` then visit `localhost:8000` and open the browser console.
- select an AOI inside the DRB and when the point source analysis results return, verify that pointing the mouse over the row creates a popup which includes the facility name
- select an AOI outside the DRB and when the point source analysis results return, verify that pointing the mouse over the row creates a popup which does not include the facility name
- verify that you haven't seen any errors in the app or the console.

Connects #1495

### DIFF
--- a/src/mmw/apps/modeling/calcs.py
+++ b/src/mmw/apps/modeling/calcs.py
@@ -44,7 +44,9 @@ def animal_population(geojson):
 def point_source_pollution(geojson):
     """
     Given a GeoJSON shape, retrieve point source pollution data
-    from the `ms_pointsource` table to display in the Analyze tab.
+    from the `ms_pointsource` or `ms_pointsource_drb table to display
+    in the Analyze tab.
+
     Returns a dictionary to append to the outgoing JSON for analysis
     results.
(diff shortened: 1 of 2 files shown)


caseypt:
+1, tested.

kellyi:
Thanks!

Issue #1488, opened 2016-09-26
Show fewer DRB streams at zoom level 11


ajrobbins:
Request from Stroud - DRB stream rendering is too noisy at too high an altitude. For example, at zoom level 11 over Philly:

![screen shot 2016-09-26 at 12 49 12 pm](https://cloud.githubusercontent.com/assets/12868342/18853336/964fe7b8-8412-11e6-972f-b60dda85f44d.png)

More input:
>I would say that for any given zoom level that the rendered blue line density for TauDEM DRB streams should be just a little bit more than (or even equivalent to) the rendered blue line density for NHDplus medium res flowlines. The full resolution (and value) of the DRB streams should only become obvious when zoomed into the closest views.The more I think about it, I think the blue line rendering density could between the two datasets could be very nearly set to be the same when zoomed out, and the differences between the two datasets will be mostly distinguished by the extra vertices.

We should adjust the rendering to show only larger streams at level 11 (also investigate whether it looks too busy at level 12).

PR #1498, opened 2016-09-28
Update animal display names for analysis results

kellyi:
This PR updates the animal display names for the frontend only to match the names specified in #1494:

```
Beef Cows --> Cows, Beef
Broilers --> Chickens, Broilers
Dairy Cows --> Cows, Dairy
Hogs/Swine --> Pigs/Hogs/Swine
Horses
Layers --> Chickens, Layers
Sheep
Turkeys
```
screen shot 2016-09-28 at 2 07 54 pm

The changes will only affect the frontend; the MapShed GMS files will still have the prior names set in `gwlfe_settings.py`. To make the change from the frontend, I replaced the imported values from `gwlfe_settings` used for the `animal_population` method with a set of hard coded `ANIMAL_DISPLAY_NAMES` representing the values above.

**Testing**
- get this branch, `vagrant up`, then visit `localhost:8000` and open the browser console.
- pick an area of interest to analyze, then check the "Animals" tab and verify that the animals are named as above.

Connects #1494

PR #1499, opened 2016-09-29
Show fewer streams

jtarricone:
This change modifies the tile server configuration to provide a more specific mapping between map zoom levels and the stream order used to make rendered tiles. More to the point, that mapping reflects a more gradual adjustment in stream orders to prevent the DRB stream overlay from severely cluttering the map.

Additionally, I noticed that the rendered stream tiles sometimes had a crayon-like texture, like so -->
mmw_staging_drb_tile

I changed the CartoCSS style to try to smooth the texture out, which resulted in this -->
mmw_test_drb_tile

Despite the obvious difference in the number of streams rendered, I think the difference is apparent and the new styling is an improvement.

To test:
* toggle the DRB stream overlay and go through various zoom levels (it may be worth doing the same on a control version of the tile server, the easiest way being to do the same process on the staging app)

Connects #1488

### DIFF
--- a/src/tiler/server.js
+++ b/src/tiler/server.js
@@ -55,12 +55,22 @@ var interactivity = {
         tableName = req.params.table;
         stream_order = 0;  // All streams
 
-        if (zoom <= 5) {
-            stream_order = 7;
-        } else if (zoom <= 6) {
-            stream_order = 6;
-        } else if (zoom <= 8) {
-            stream_order = 5;
+        if (tableName === tables.drb_streams_v2) {
+            // drb_zoom_levels: { zoomLevel : stream_order }
+            drb_zoom_levels = {
+                1:7, 2:7, 3:7, 4:7, 5:7, 6:6, 7:6, 8:5, 9:5, 10:4,
+                11:3, 12:2, 13:0, 14:0, 15:0, 16:0, 17:0, 18: 0
+            }; 
+
+            stream_order = zoom in drb_zoom_levels ? drb_zoom_levels[zoom] : 0;
+        } else {
+            if (zoom <= 5) {
+                stream_order = 7;
+            } else if (zoom <= 6) {
+                stream_order = 6;
+            } else if (zoom <= 8) {
+                stream_order = 5;
+            }
         }
 
         return '(SELECT geom, stream_order FROM ' + tableName +
--- a/src/tiler/styles.mss
+++ b/src/tiler/styles.mss
@@ -44,18 +44,28 @@
    Ensure that any changes to which stream_order + zoom levels are rendered
    here have appropriate coinciding filters in server.js
 */
+
+#drb_streams_50 {
+    line-color: @streamColor;
+    line-join: round;
+    line-cap: round;
+}
+
+#nhdflowline {
+    line-color: @streamColor;
+    line-join: round;
+    line-cap: round;
+}
+
 #drb_streams_50[zoom<=4],
 #nhdflowline[zoom<=4] {
   [stream_order=10] {
-    line-color: @streamColor;
     line-width: 5.0 * @zoomBase;
   }
   [stream_order=9] {
-    line-color: @streamColor;
     line-width: 3.0 * @zoomBase;
   }
   [stream_order<=8] {
-    line-color: @streamColor;
     line-width: 2.0 * @zoomBase;
   }
 }
@@ -63,15 +73,12 @@
 #drb_streams_50[zoom>=5][zoom<=6],
 #nhdflowline[zoom>=5][zoom<=6] {
   [stream_order=10] {
-    line-color: @streamColor;
     line-width: 7.0 * @zoomBase;
   }
   [stream_order=9] {
-    line-color: @streamColor;
     line-width: 4.0 * @zoomBase;
   }
   [stream_order<=8] {
-    line-color: @streamColor;
     line-width: 3.0 * @zoomBase;
   }
 }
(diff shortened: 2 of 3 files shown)


kdeloach:
+1 Nice

src/tiler/styles.mss
```diff
@@ -65,14 +65,20 @@
   [stream_order=10] {
     line-color: @streamColor;
     line-width: 7.0 * @zoomBase;
+    line-join: round;
```
caseypt: Good idea. Consider moving these into `#drb_streams_50, #nhdflowline`, so that they don't have to be repeated for each zoom level.

src/tiler/server.js
```diff
@@ -55,12 +55,20 @@ var interactivity = {
...
 
-        if (zoom <= 5) {
-            stream_order = 7;
-        } else if (zoom <= 6) {
-            stream_order = 6;
-        } else if (zoom <= 8) {
-            stream_order = 5;
+        if (tableName === tables.drb_streams_v2) {
+            drb_zoom_levels = {
+                1:7, 2:7, 3:7, 4:7, 5:7, 6:6, 7:6, 8:5, 9:5,
```
caseypt: Consider adding a comment or adding a commit message to describe this mapping

kellyi:
Not sure if this is something wrong with my local, but at max zoom level it looks like my tiler is showing some 400 errors for the streams:

screen shot 2016-09-29 at 12 32 11 pm

kellyi:
I reprovisioned and reloaded my tiler and still got those 400 errors at the max zoom level. The streams work at zoom levels below (above?) that, though, and the entire DRB layer's thinned out as you zoom out.

mmcfarland:
@ajrobbins was this request meant to be included in the upcoming release?

src/tiler/server.js
```diff
@@ -55,12 +55,20 @@ var interactivity = {
...
         stream_order = 0;  // All streams
 
-        if (zoom <= 5) {
-            stream_order = 7;
-        } else if (zoom <= 6) {
-            stream_order = 6;
-        } else if (zoom <= 8) {
-            stream_order = 5;
+        if (tableName === tables.drb_streams_v2) {
+            drb_zoom_levels = {
```
jtarricone: I incorrectly assumed 17 was the max zoom; I added a mapping for 18 as well as a check for the zoom level.

ajrobbins:
@mmcfarland Yes, this PR as well as #1498 and #1497 were in response to feedback on the release branch on staging; barring any blocking or large issues we should include the fixes in Tuesday's release.

src/tiler/server.js
```diff
@@ -55,12 +55,22 @@ var interactivity = {
...
-            stream_order = 7;
-        } else if (zoom <= 6) {
-            stream_order = 6;
-        } else if (zoom <= 8) {
-            stream_order = 5;
+        if (tableName === tables.drb_streams_v2) {
+            // drb_zoom_levels: { zoomLevel : stream_order }
+            drb_zoom_levels = {
+                1:7, 2:7, 3:7, 4:7, 5:7, 6:6, 7:6, 8:5, 9:5, 10:4,
+                11:3, 12:2, 13:0, 14:0, 15:0, 16:0, 17:0, 18: 0
```
mmcfarland: Our zoom levels actually go up to 22, with the Google basemaps.

jtarricone: Since the stream order is 0 past 13 anyway, there shouldn't be any visible effect.

kellyi: I checked this again and I'm no longer getting the 400 error from the tiler on zooming in all the way for the default basemap. 

When I switch basemaps and zoom in further, I don't see a the 400 error but I also don't see any streams. I checked on staging and it looks like this is how it behaves there, too (maybe the tiler doesn't go past 18 for this overlay?). Just wanted to make sure this is the desired behavior?

mmcfarland: Good catch, it's definitely the desired behavior to have all streams showing out beyond 18 if the basemap supports it.  There shouldn't be any restriction in the tiler per se, although it's probably related to a rendering rule.

jtarricone:
I found the bug that prevented tiles from being fetched past zoom 18, and committed a fix for it.

kellyi:
Taking another look!

kellyi:
The zoom control has some slightly wonky behavior now when the DRB layer is turned on:

- with the "Streets" and "Satellite" basemaps, I am able to zoom past the maximum zoom for the tileset while still seeing the streams
- with the "Satellite with Roads" and "Terrain" basemaps, the increase zoom control doesn't disable at max zoom level, so I can continue clicking it and the DRB streams layer will temporarily seem to zoom in, and then everything snaps back to the max zoom for the basemap.

src/mmw/mmw/settings/layer_settings.py
```diff
@@ -141,6 +141,7 @@
         'stream': True,
         'overlay': True,
         'minZoom': 5,
+        'maxZoom': 22,
```
kellyi: Is it possible to set `maxZoom` for this layer dynamically on the client side to === the `maxZoom` of whatever is the currently selected basemap?

mmcfarland: There's some code we have specifically for the dynamic nature of Google basemap max zoom:
https://github.com/WikiWatershed/model-my-watershed/blob/1a836bb5f8281b7d7f2b0bedb2a2f7bf9131794a/src/mmw/js/src/core/views.js#L359-L373

jtarricone:
I just pushed an update that mostly addresses the zoom issue; I noticed that if the base layer is set to satellite or something, magnified past 18, and then the DRB layer enabled, leaflet won't request the tiles.

kellyi:
Tested this again and zoom control now works as expected. 

The DRB stream overlay is back to not rendering above the max zoom of the default "Streets" basemap for any of the other basemaps, but I think we should make another issue to address that since that issue already exists on both staging and production already and a fix may be too involved to work out prior to tomorrow's release.

+1 from me.

jtarricone:
>The DRB stream overlay is back to not rendering above the max zoom of the default "Streets" basemap for any of the other basemaps

I only observed that with the following steps: switch to non-Streets basemap -> zoom to 19+ -> activate the DRB layer.

If the DRB layer was active while zoomed below 18, leaflet requested the tiles for 18+ zoom levels. The zoom settings seems to be mutated somewhere that isn't immediately obvious.

At any rate, I agree that this issue should be handled separately, but I think this PR gets us part of the way there at least.

PR #1503, opened 2016-09-30
Feature/add alert modal

jtarricone:
This adds a modal form for alerts, and replaces the existing generic `window.alert` calls with the modal form.

To test:
* perform user actions that should result in an error: draw a self-intersecting AoI polygon; rename a model scenario with an existing name; draw a very large AoI polygon; cause a login error
* the modal forms should show and have relevant messages

Connects #645

### DIFF
--- a/src/mmw/js/src/core/modals/models.js
+++ b/src/mmw/js/src/core/modals/models.js
@@ -33,17 +33,17 @@ var ShareModel = Backbone.Model.extend({
 var AlertTypes = {
     info: {
         alertHeader: 'Information',
-        alertIcon: '\"fa fa-info-circle\"',
+        alertIcon: 'fa-info-circle',
         dismissLabel: 'Okay'
     },
     warn: {
         alertHeader: 'Warning',
-        alertIcon: '\"fa fa-exclamation-triangle\"',
+        alertIcon: 'fa-exclamation-triangle',
         dismissLabel: 'Okay'
     },
     error: {
         alertHeader: 'Error',
-        alertIcon: '\"fa fa-ban\"',
+        alertIcon: 'fa-ban',
         dismissLabel: 'Okay'
     }
 };
--- a/src/mmw/js/src/core/modals/models.js
+++ b/src/mmw/js/src/core/modals/models.js
@@ -32,17 +32,17 @@ var ShareModel = Backbone.Model.extend({
 
 var AlertTypes = {
     info: {
-        header: 'Information',
+        alertHeader: 'Information',
         alertIcon: '\"fa fa-info-circle\"',
         dismissLabel: 'Okay'
     },
     warn: {
-        header: 'Warning',
+        alertHeader: 'Warning',
         alertIcon: '\"fa fa-exclamation-triangle\"',
         dismissLabel: 'Okay'
     },
     error: {
-        header: 'Error',
+        alertHeader: 'Error',
         alertIcon: '\"fa fa-ban\"',
         dismissLabel: 'Okay'
     }
@@ -50,7 +50,7 @@ var AlertTypes = {
 
 var AlertModel = Backbone.Model.extend({
     defaults: {
-        message: '',
+        alertMessage: '',
         alertType: AlertTypes.info,
     },
 });
--- a/src/mmw/js/src/core/modals/models.js
+++ b/src/mmw/js/src/core/modals/models.js
@@ -30,8 +30,35 @@ var ShareModel = Backbone.Model.extend({
     }
 });
 
+var AlertTypes = {
+    info: {
+        header: 'Information',
+        alertIcon: '\"fa fa-info-circle\"',
+        dismissLabel: 'Okay'
+    },
+    warn: {
+        header: 'Warning',
+        alertIcon: '\"fa fa-exclamation-triangle\"',
+        dismissLabel: 'Okay'
+    },
+    error: {
+        header: 'Error',
+        alertIcon: '\"fa fa-ban\"',
+        dismissLabel: 'Okay'
+    }
+};
+
+var AlertModel = Backbone.Model.extend({
+    defaults: {
+        message: '',
+        alertType: AlertTypes.info,
+    },
+});
+
 module.exports = {
     ConfirmModel: ConfirmModel,
     InputModel: InputModel,
-    ShareModel: ShareModel
+    ShareModel: ShareModel,
+    AlertTypes: AlertTypes,
+    AlertModel: AlertModel
 };
--- a/src/mmw/js/src/core/modals/views.js
+++ b/src/mmw/js/src/core/modals/views.js
@@ -12,6 +12,7 @@ var _ = require('underscore'),
     modalInputTmpl = require('./templates/inputModal.html'),
     modalPlotTmpl = require('./templates/plotModal.html'),
     modalShareTmpl = require('./templates/shareModal.html'),
+    modalAlertTmpl = require('./templates/alertModal.html'),
     vizerUrls = require('../settings').get('vizer_urls'),
 
     ENTER_KEYCODE = 13,
@@ -198,6 +199,31 @@ var ShareView = ModalBaseView.extend({
     }
 });
 
+var AlertView = ModalBaseView.extend({
+    template: modalAlertTmpl,
+
+    ui: {
+        dismiss: '.btn-default'
+    },
+
+    events: _.defaults({
+        'click @ui.dismiss': 'dismissAction'
+    }, ModalBaseView.prototype.events),
+
+    dismissAction: function() {
+        this.triggerMethod('dismiss');
+    },
+
+    templateHelpers: function() {
+        return {
+            alertMessage: this.model.get('message'),
+            alertHeader: this.model.get('alertType').header,
+            alertIcon: this.model.get('alertType').alertIcon,
+            dismissLabel: this.model.get('alertType').dismissLabel
+        };
+    }
+});
+
 var PlotView = ModalBaseView.extend({
     template: modalPlotTmpl,
 
@@ -371,5 +397,6 @@ module.exports = {
     ShareView: ShareView,
     InputView: InputView,
     ConfirmView: ConfirmView,
-    PlotView: PlotView
+    PlotView: PlotView,
+    AlertView: AlertView
 };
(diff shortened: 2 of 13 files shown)


rajadain:
Sorry seemed to have missed this (didn't get an email notification? weird). Will get on it right away.

src/mmw/sass/components/_modals.scss
```diff
@@ -138,3 +138,20 @@
   top: 1rem;
   left: 2rem;
 }
+
+.fa {
+  font-size: 1.0rem;
+  padding: 0 0.1rem 0 0.1rem;
+}
```
kellyi: These changes are scoped broadly enough to impact some of the other app icons:

screen shot 2016-10-05 at 2 37 34 pm

kellyi: Seems to the analyze tab while that's running, too:

screen shot 2016-10-05 at 2 41 21 pm

kellyi:
Trying this out and, got this error alert modal when trying to close a shape which would self-intersect:

screen shot 2016-10-05 at 2 36 41 pm

Saw this message for the RWD multipolygon failure:

screen shot 2016-10-05 at 2 46 14 pm


This popup for a too-small AOI seems not to use the modal:

screen shot 2016-10-05 at 2 40 04 pm

src/mmw/sass/components/_modals.scss
kellyi: Also the "Delineate Watershed" tab:

screen shot 2016-10-05 at 2 47 35 pm

jtarricone:
This branch doesn't appear to include #1493 so I'll have to rebase it.

src/mmw/sass/components/_modals.scss
jtarricone: I might have the wrong idea about how the sass rules are scoped. Do those rules apply to components other than modal elements?

kellyi:
Sounds good. `develop` should also now have the fix for the Postgres bindings which was causing failures on the CI server.

src/mmw/js/src/modeling/models.js
```diff
@@ -888,6 +889,7 @@ var ScenarioModel = Backbone.Model.extend({
 var ScenariosCollection = Backbone.Collection.extend({
     model: ScenarioModel,
     comparator: 'created_at',
+    alerts: [],
```
rajadain: Why use an `alerts` array?

src/mmw/js/src/core/modals/views.js
```diff
@@ -198,6 +199,31 @@ var ShareView = ModalBaseView.extend({
...
+    }, ModalBaseView.prototype.events),
+
+    dismissAction: function() {
+        this.triggerMethod('dismiss');
+    },
+
+    templateHelpers: function() {
+        return {
+            alertMessage: this.model.get('message'),
+            alertHeader: this.model.get('alertType').header,
```
rajadain: If we're reusing the same names, we could also use something like `_.extend`:

```js
return _.extend(this.model.get('alertType'), {
    alertMessage: this.model.get('message')
};
```

src/mmw/js/src/core/modals/models.js
```diff
@@ -30,8 +30,35 @@ var ShareModel = Backbone.Model.extend({
...
+    },
+    warn: {
+        header: 'Warning',
+        alertIcon: '\"fa fa-exclamation-triangle\"',
+        dismissLabel: 'Okay'
+    },
+    error: {
+        header: 'Error',
+        alertIcon: '\"fa fa-ban\"',
+        dismissLabel: 'Okay'
```
rajadain: Consider extracting `dismissLabel` into `AlertModal` since it has the same value in all three cases. Also, having a default there would allow us to not having to specify it in `modeling/models.js:965` below.

src/mmw/js/src/modeling/models.js
jtarricone: I recall there was some odd issue I encountered when I initially tried using a single object, but in any event this just seemed like a slightly cleaner approach. Since it's attached to a collection type, a "mailbox" seemed more idiomatic. More tangibly, I didn't like the the idea of touching the state in the view that this binds to, e.g. clearing the value if it were just a string, so using list constructs seemed preferable.

rajadain: So the assumption is that since this is called from within `ScenarioCollection.updateScenarioName`, it could be called my multiple `ScenarioModel`s in quick succession, leading to unmanaged overlapping modals?

This will never happen. The only reason `updateScenarioName` is on the collection rather than on the model itself is we want to keep scenario names unique, and since a model has no reference to the collection to which it belongs we set it on the collection.

And since nothing from `this` is passed to the alert modal (only a string, and a warning config), we can simply use the pattern of creating an `modalViews.AlertView` and calling `render()` on it inline.

rajadain:
You may also have to update this test: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/tests.js#L901-L947 which listens for `window.alert` to be called on `updateScenarioName`.

src/mmw/js/src/core/views.js
```diff
@@ -124,11 +124,17 @@ var HeaderView = Marionette.ItemView.extend({
...
                 })
                 .fail(function() {
-                    window.alert(
-                        'There was an error trying to clone that project. ' +
-                        'Please ensure that the project ID is valid and ' +
-                        'that you have permission to view the project.'
-                    );
+                    var alertView = new modalViews.AlertView({
+                        model: new modalModels.AlertModel({
+                            alertMessage:'There was an error trying to clone that project. ' +
```
rajadain: Prefer having a space after the colon, and left aligning the strings:

```js
model: new modalModels.AlertModel({
    alertMessage: 'There was an error trying to clone that project. ' +
                  'Please ensure tha ttheh project ID is valid and ' +
                  'that you have permission to view the project.',
    alertType: 'Error',
    dismissLabel: 'Okay',
})
```

src/mmw/js/src/core/views.js
```diff
@@ -124,11 +124,17 @@ var HeaderView = Marionette.ItemView.extend({
...
+                                    'Please ensure that the project ID is valid and ' +
+                                    'that you have permission to view the project.',
+                            alertType:'Error',
+                            dismissLabel:'Okay'
```
rajadain: Since the `dismissLabel` is _always_ `'Okay'`, consider giving it a default value in `AlertModel` and only overriding it when different.

src/mmw/js/src/core/modals/models.js
```diff
@@ -30,8 +30,35 @@ var ShareModel = Backbone.Model.extend({
...
+    },
+    warn: {
+        alertHeader: 'Warning',
+        alertIcon: '\"fa fa-exclamation-triangle\"',
+        dismissLabel: 'Okay'
+    },
+    error: {
+        alertHeader: 'Error',
+        alertIcon: '\"fa fa-ban\"',
```
rajadain: Prefer not escaping strings here, and quoting them in the template. Also, since they all have `fa`, consider putting that in the template for less redundancy:

```js
error: {
    alertHeader: 'Error',
    alertIcon: 'fa-ban',
    dismissLabel: 'Okay',
}
```

```jinja
<i class="fa {{ alertIcon }}"></i>
```

hectcastro:
@azavea-bot rebuild

src/mmw/js/src/core/modals/models.js
jtarricone: I didn't know that quoting `{{ alertIcon }}` put literal quotes in the markup, I kept wondering why it wasn't working so I escaped the strings in the model.

src/mmw/js/src/core/views.js
```diff
@@ -124,11 +124,17 @@ var HeaderView = Marionette.ItemView.extend({
...
-                        'There was an error trying to clone that project. ' +
-                        'Please ensure that the project ID is valid and ' +
-                        'that you have permission to view the project.'
-                    );
+                    var alertView = new modalViews.AlertView({
+                        model: new modalModels.AlertModel({
+                            alertMessage:'There was an error trying to clone that project. ' +
+                                    'Please ensure that the project ID is valid and ' +
+                                    'that you have permission to view the project.',
+                            alertType:'Error',
```
arottersman: Should this be `alertType: modalModels.AlertTypes.error` now?

src/mmw/js/src/modeling/views.js
```diff
@@ -385,11 +392,35 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
...
+                var match = self.model.collection.find(function(model) {
+                    return model.get('name').toLowerCase() === newName.toLowerCase();
+                });
+
+                if (match) {
+                    console.log('This name is already in use.');
+                    var alertView = new modalViews.AlertView({
+                        model: new modalModels.AlertModel({
+                            alertMessage: 'There is another scenario with the same name.' +
+                                            'Please choose a unique name for this scenario.',
```
arottersman: There should be a space before "Please"

src/mmw/js/src/projects/views.js
```diff
@@ -152,7 +152,15 @@ var ProjectRowView = Marionette.ItemView.extend({
             self.model
                 .destroy({ wait: true })
                 .fail(function() {
-                    window.alert('Could not delete this project.');
+                    var alertView = new modalViews.AlertView({
+                        model: new modalModels.AlertModel({
+                            alertMessage: 'Could not delete this project.', 
+                            alertType: 'error',
```
arottersman: I believe this should also be `alertType: modalModels.AlertTypes.error`

src/mmw/sass/components/_modals.scss
```diff
@@ -138,3 +138,22 @@
...
 }
+
+.modal-popup {
+    .fa {
+      font-size: 1.0rem;
+      padding: 0 0.1rem 0 0.1rem;
+    }
+
+    .fa-info-circle {
+      color: green;
```
arottersman: Looks fine in this case, but does bring me back to this comment from @jfrankl 
https://github.com/azavea/pwd-parcel-viewer/issues/359#issuecomment-250157013

arottersman:
Tests are all good:
screen shot 2016-10-12 at 12 29 56 pm

jtarricone:
>There are a few spots though where you're still declaring the alertType to be a string. 

I guess I fixup commit hadn't been pushed, because in my local repo each modal instance had the proper type declared.

arottersman:
Cool, looks like it's all good to go now. +1

PR #1506, opened 2016-10-04
Don't run MapShed for AoIs larger than 1000 sq km

rajadain:
## Overview

Large jobs fail, but also clog up our infrastructure causing cascading effects. By limiting the size of allowed areas of interest for MapShed, we lower the probability of getting very large jobs that affect other jobs.

### Notes

I pulled in #1493 as well so there aren't any merge conflicts when this gets merged back to `develop`.

## Testing Instructions

Try running MapShed on HUC-10, ensure it works. Try running MapShed on a large County. Ensure you get this error message:

image

### DIFF
--- a/src/mmw/apps/home/views.py
+++ b/src/mmw/apps/home/views.py
@@ -135,6 +135,7 @@ def get_client_settings(request):
             'map_controls': settings.MAP_CONTROLS,
             'vizer_urls': settings.VIZER_URLS,
             'model_packages': get_model_packages(),
+            'mapshed_max_area': settings.GWLFE_CONFIG['MaxAoIArea']
         }),
         'google_maps_api_key': settings.GOOGLE_MAPS_API_KEY,
     }
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -13,6 +13,7 @@ var $ = require('jquery'),
     modalViews = require('../core/modals/views'),
     coreModels = require('../core/models'),
     coreViews = require('../core/views'),
+    coreUtils = require('../core/utils'),
     chart = require('../core/chart'),
     utils = require('../core/utils'),
     pointSourceLayer = require('../core/pointSourceLayer'),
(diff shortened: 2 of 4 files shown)


kellyi:
Tried to model a 6,800km aoi and:

screen shot 2016-10-04 at 1 07 54 pm

kellyi:
As it should, this 960km one slips past but still times out on my local:

screen shot 2016-10-04 at 1 12 46 pm

kellyi:
If I try running MapShed *after* the Point Source and Animals Analyze tabs have returned with data, I get an error in the console and not the error message popup:

screen shot 2016-10-04 at 1 15 04 pm

kellyi:
That error seems to go away and the popup works, btw, once the Land and Soil tabs have returned data. 

In other words, it seems to happen between the time when the Animals/Point Source data returns and when the Land and Soil data returns.

kellyi:
That error also seems to happen between Animals/Point Source -> Land/Soil when trying to run MapShed on AOIs < 1000 km^2.

kellyi:
Checked out `develop` and it seems to be happening there, too. Going to look at `release`.

kellyi:
It isn't happening on `release`, so I assume it's something to do with the check in this commit, because it's also happening when I try to run TR-55 in between Animals/Point Source -> Land/Soil:

https://github.com/WikiWatershed/model-my-watershed/pull/1506/commits/41d637435c76ed65e5b8d2eae5edbb7aecfbd5a4

rajadain:
Hmm let me try that out on my local. Good catch.

kellyi:
I think it's the variable assignment on line 63 for `analysisResults`. Commenting that and everything new after out prevented the error; uncommenting that and leaving everything else new commented-out showed the error again.

rajadain:
It seems like we initialize `results` to `""` rather than `"{}"`. When we run `JSON.parse("")` we get this error. Let me think if we can alleviate this ...

kellyi:
I was trying this and it seemed to work. Basically: don't parse `analysisResults` or initialize `landResults` until we know for sure that `analysisResults` is not an empty string:

```js
analysisResults = App.getAnalyzeCollection()
                                 .findWhere({taskName: 'analyze'})
                                 .get('result'),
            landResults = false;

        if (analysisResults !== '') {
            landResults = _.find(JSON.parse(analysisResults),
                function(element) {
                    return element.name === 'land';
                }
            )
        }

        if (modelPackageName === 'gwlfe' && settings.get('mapshed_max_area') &&
            (aoiModel.get('area') > settings.get('mapshed_max_area'))) {
            window.alert("The selected Area of Interest is too big for " +
                         "the Watershed Multi-Year Model. The currently " +
                         "maximum supported size is 1000 km².");
            return;
        }

        if (landResults) {
            var landCoverTotal = _.sum(_.map(landResults.categories,
                    function(category) {
                        if (category.type === 'Open Water') {
                            return 0;
                        }
                        return category.area;
                    }));

            if (landCoverTotal === 0) {
                window.alert("The selected Area of Interest doesn't " +
                             "include any land cover to run the model.");
                return;
            }
        }
```

rajadain:
I posted a simple fixup, that defaults to string of empty object if `result` is empty string. Could you take another look?

kellyi:
Looking at it now.

kellyi:
numeric check for `> 1000` is applying to m^2 also

rajadain:
Should be updated in the latest.

rajadain:
Sorry, force pushed the last commit to fix some formatting.

kellyi:
+1. Tested this with a variety of AOIs. I was able to run Mapshed on AOIs < 1000 km^2, saw it fail for those above, and saw it work for AOIs with values of 20,000 m^2. The frontend check for too-small AOIs also works.

Feel free to drop the `self.modelingRegion` commit, and we can handle that and similar errors in another card.

rajadain:
Thanks for all your help on this! All the tests are passing on my local, so I'm merging now ...

Issue #1479, opened 2016-09-21
Visualize NHD Stream Water Quality attributes


mmcfarland:
Working from #1478 & #1471, create a Windshaft tile endpoint which executes a joined query on NHD + water quality table and does custom styling rules to apply the defined styles based on the TN, TSS or TP attribute.  These are three distinct tiling endpoints, one for each endpoint.  The join results should include _all_ `nhdflowlines` and have nulls for the TP, TN & TSS values where those records don't exist.  In this way, we will ultimately render streams which have no water quality data as "gray", no matter where they occure (in or outside of the DRB) and records _with_ water quality attributes will be symoblized according to that attributes class breaks. 

Create Stream layer selector entries for these three new endpoints.  Ensure that they are routed through the same s3/CF caching layer as the other tiles and that they use the same attribute based zoom and width rendering rules.

mmcfarland:
#1478 has the class & color breaks defined.

caseypt:
Blocked on #1478 and #1471

Issue #1478, opened 2016-09-21
Create class break scheme for stream water quality


mmcfarland:
Based on #1471, the values for each TN, TP & TSS have to be broken into class breaks or color ramps for which we can generate styling rules.  This may include "high - med - low" categories, or discrete clusters of value binning.  Coordinate with Stroud to come up with the most appropriate breaks.

ajrobbins:
BME is recommending the following break points:

|          |(kg/ha)       |(mg/l)          |
|------|-----------|------------|
|TSS    | 0-250      |     0-50       |
|           | 250-500 |     50-100    |
|           | 500-750  |    100-150  |
|           | 750-1000|    150-200  |
|           | >1000      | >200          |
|           |                  |                    |
|TN      | 0-5           | 0-1              |
|           | 5-10         | 1-2              |
|           |10-15         |2-3              |
|           | 15-20       | 3-4             |
|           | >20           | >4               |
|           |                   |                    | 
|TP       | 0-0.30      |0-0.03         |
|            | 0.30-0.60| 0.03-0.06   |
|            | 0.60-0.90| 0.06-0.09   |
|            | 0.90-1.20 | 0.09-0.12   |
|             | >1.20       | >0.12          |

ajrobbins:
We should use a greyscale (low loading rate = light, high loading rate = dark) for the catchements (kg/ha), and a scale from green (lowest quartile) to yellow to red (highest quartile) for the stream data (mg/l).

mmcfarland:
Here is the color brewer R-Y-G quintile break colors:
`['#d7191c','#fdae61','#ffffbf','#a6d96a','#1a9641']`

caseypt:
Blocked on #1471

jtarricone:
It seems like this issue can be totally absorbed by #1479 now that the discrete bins have been defined, right?

Issue #1471, opened 2016-09-19
Import water quality data for DRB area NHD+ streams


ajrobbins:
This should be a table with ComID, TN, TSS, and TP. Data is on the fileshare in the project folder at /data/srat

mmcfarland:
This new table joins `ComID` with our existing `nhdflowline` table, but only represents features in the national dataset which occur in the DRB and have had this water quality modeling applied.  We should import this data as a new table (`nhd_water_quality`?) and verify that we can successfully join on `ComID` between the two tables.  If there are significant records not included in `nhdflowlines` that are related in this new table, we may have to seek advice.  Anticipate that only a fraction of the entire national set will be represented in the new table, though.

The end goal is a table which with we can write a query for a new tiler endpoint to render nhd streams based off of the new TN, TP & TSS attributes.

mmcfarland:
The data author anticipates that around 10k stream in the DRB area will have values, and approximately 6k streams in the DRB areas will not.  All streams outside of the DRB will not have any values.

mmcfarland:
Additionally, the units for these values are all concentrations in mg/l.

ajrobbins:
Sorry @arottersman but the data author just provided a slightly updated table for this. I put it on the fileshare; it's the 9_29_16 backup file.

Issue #1482, opened 2016-09-21
Visualize DRB Catchment water quality measurments


mmcfarland:
Based on #1481 & #1480, create 3 tile endpoints which visualize each TN, TP & TSS attribute for DRB Catchments.  Add the three layers to the layer selector and allow them to be cached by the existing cache infrastructure.

ajrobbins:
This card also includes assigning color breaks for catchment visualization:

Use a greyscale scheme with the following breaks:
BME is recommending the following break points (the mg/l column is the relevant one for this card):

|          |(kg/ha)       |
|------|-----------|
|TSS    | 0-250      |
|           | 250-500 | 
|           | 500-750  |    
|           | 750-1000|    
|           | >1000      | 
|           |                  |
|TN      | 0-5           |
|           | 5-10         |
|           |10-15         |
|           | 15-20       |
|           | >20           |
|           |                   |
|TP       | 0-0.30      |
|            | 0.30-0.60|
|            | 0.60-0.90| 
|            | 0.90-1.20 |
|             | >1.20       |

PR #1507, opened 2016-10-04
Feature/water quality tiles

jtarricone:
This builds on #1471 to add rendered tile overlays to the map.

To test:
* clone this branch and bring up an environment (if you have one running, `tiler` will probably have to be reprovisioned to pick up `underscore`, or else you can just `npm install underscore` on the VM
* try enabling the new layers from the `Overlay` tab in the layer selector; the new ones are "Deleware River Basin TP Concentration", "Deleware River Basin TN Concentration", & "Deleware River Basin TSS Concentration"; cycle through different zoom levels for each, and try switching base maps at some point
* the stream layers should render and the Delaware basin tiles should have different colors per the data binding for each layer

This will probably have merge conflicts with `develop` once #1499 gets merged in, but I think I replicated most of the parts of `styles.mss` to avoid anything gnarly. Additionally, it's probably worth taking a critical look at the color scheme provided in #1478, since I could understand how the rendered colors might be hard to see on certain base maps at certain zoom levels.

Connects #1479 & #1478

### DIFF
--- a/scripts/aws/setupdb.sh
+++ b/scripts/aws/setupdb.sh
@@ -130,8 +130,5 @@ if [ "$load_water_quality" = "true" ] ; then
     # Fetch water quality data
     FILES=("nhd_water_quality.sql.gz" "drb_catchment_water_quality.sql.gz")
     PATHS=("drb_catchment_water_quality_tn" "drb_catchment_water_quality_tp"
-           "drb_catchment_water_quality_tss")
-
-    download_and_load $FILES
-    purge_tile_cache $PATHS
+            "drb_catchment_water_quality_tss" "nhd_quality_tp" "nhd_quality_tn")
 fi
--- a/src/mmw/mmw/settings/layer_settings.py
+++ b/src/mmw/mmw/settings/layer_settings.py
@@ -145,6 +145,30 @@ LAYERS = [
         # overlaps with perimeter polygon.
     },
     {
+        'code': 'nhd_quality_tp',
+        'display': 'Delaware River Basin TP Concentration',
+        'table_name': 'nhd_quality_tp',
+        'stream': True,
+        'overlay': True,
+        'minZoom': 3
+    },
+    {
+        'code': 'nhd_quality_tn',
+        'display': 'Delaware River Basin TN Concentration',
+        'table_name': 'nhd_quality_tn',
+        'stream': True,
+        'overlay': True,
+        'minZoom': 3
+    },
+    {
+        'code': 'nhd_quality_tss',
+        'display': 'Delaware River Basin TSS Concentration',
+        'table_name': 'nhd_quality_tss',
+        'stream': True,
+        'overlay': True,
+        'minZoom': 3
+    },
+    {
         'display': 'National Land Cover Database',
         'short_display': 'NLCD',
         'helptext': 'National Land Cover Database defines'
--- a/src/tiler/server.js
+++ b/src/tiler/server.js
@@ -15,7 +15,7 @@ var dbUser = process.env.MMW_DB_USER,
     redisPort = process.env.MMW_CACHE_PORT,
     tileCacheBucket = process.env.MMW_TILECACHE_BUCKET,
     stackType = process.env.MMW_STACK_TYPE,
-        rollbarAccessToken = process.env.ROLLBAR_SERVER_SIDE_ACCESS_TOKEN;
+    rollbarAccessToken = process.env.ROLLBAR_SERVER_SIDE_ACCESS_TOKEN;
 
 var NHD_QUALITY_TSS_MAP = {
     'L1': [0,50],
@@ -56,6 +56,9 @@ var interactivity = {
         huc12: 'boundary_huc12',
         drb_streams_v2: 'drb_streams_50',
         nhd_streams_v2: 'nhdflowline',
+        nhd_quality_tn: 'nhd_quality_tn',
+        nhd_quality_tp: 'nhd_quality_tp',
+        nhd_quality_tss: 'nhd_quality_tss',
         municipalities: 'dep_municipalities',
         urban_areas: 'dep_urban_areas',
         // The DRB Catchment tables here use aliases to match data from different
@@ -66,6 +69,7 @@ var interactivity = {
         drb_catchment_water_quality_tss: 'drb_catchment_water_quality_tss'
     },
     drbCatchmentWaterQualityTable = 'drb_catchment_water_quality';
+    nhdQualityTable = 'nhd_water_quality',
     shouldCacheRequest = function(req) {
         // Caching can happen if the bucket to write to is defined
         // and the request is not coming from localhost.
@@ -188,7 +192,7 @@ var config = {
     },
     redis: {
         host: redisHost,
-        eort: redisPort
+        port: redisPort
     },
     log_format: '{ "timestamp": ":date[iso]", "@fields": { "remote_addr": ":remote-addr", "body_bytes_sent": ":res[content-length]", "request_time": ":response-time", "status": ":status", "request": ":method :url HTTP/:http-version", "request_method": ":method", "http_referrer": ":referrer", "http_user_agent": ":user-agent" } }',
 
@@ -251,14 +255,13 @@ var config = {
                 req.params.sql = getSqlForStreamByReq(req);
             }
 
-<<<<<<< HEAD
             if (tableId.indexOf('drb_catchment') >= 0) {
                 req.params.sql = getSqlForDRBCatchmentByTableId(tableId);
-=======
+            }
+
             if (tableId.indexOf('nhd_quality') >= 0) {
                 req.params.table = tables[tableId];
                 req.params.sql = getSqlForStreamByReq(req);
->>>>>>> Add rendering & styling for water quality data
             }
 
             req.params.dbname = dbName;
--- a/src/tiler/server.js
+++ b/src/tiler/server.js
@@ -15,7 +15,28 @@ var dbUser = process.env.MMW_DB_USER,
     redisPort = process.env.MMW_CACHE_PORT,
     tileCacheBucket = process.env.MMW_TILECACHE_BUCKET,
     stackType = process.env.MMW_STACK_TYPE,
-    rollbarAccessToken = process.env.ROLLBAR_SERVER_SIDE_ACCESS_TOKEN;
+        rollbarAccessToken = process.env.ROLLBAR_SERVER_SIDE_ACCESS_TOKEN;
+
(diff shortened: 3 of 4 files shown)


src/tiler/styles.mss
```diff
@@ -44,126 +51,195 @@
    Ensure that any changes to which stream_order + zoom levels are rendered
    here have appropriate coinciding filters in server.js
 */
+
+#drb_streams_50 {
```
jtarricone: As I mentioned in the comment, most of this file is modified to mirror the changes from #1499, so a lot of these diffs are irrelevant.

src/mmw/mmw/settings/layer_settings.py
```diff
@@ -145,6 +145,30 @@
         # overlaps with perimeter polygon.
     },
     {
+        'code': 'nhd_quality_tp',
+        'display': 'Delaware River Basin TP Concentration',
```
kellyi: I believe these should be labeled as for the "National" rather than "Delaware River Basin" data.

kellyi:
I'm working on #1482 which is similar to this but for DRB data, so I'm going to rebase my branch on this one.

src/tiler/server.js
```diff
@@ -63,8 +90,44 @@ var interactivity = {
...
+        } else if (tableId === 'nhd_quality_tp') {
+            return '(SELECT geom, stream_order, ' + 
+                caseSql(NHD_QUALITY_TP_MAP,'tp_yr_avg_concmgl') + ' AS nhd_qual_grp ' +
+                'FROM ' + tables.nhd_streams_v2 + ' ' +
+                'LEFT OUTER JOIN ' + nhdQualityTable + ' ' +
+                'ON nhd_water_quality.comid=nhdflowline.comid ' +
+                'WHERE stream_order >= ' + stream_order + ') as q';
+        } else if (tableId === 'nhd_quality_tss') {
+            return '(SELECT geom, stream_order, ' + 
+                caseSql(NHD_QUALITY_TP_MAP,'tss_concmgl') + ' AS nhd_qual_grp ' +
```
mmcfarland: `NHD_QUALITY_TP_MAP` should be `TSS` I believe

src/tiler/server.js
```diff
@@ -63,8 +90,44 @@ var interactivity = {
...
+            bounds[0] = 'CASE ' + bounds[0];
+            bounds.push('WHEN ' + field + ' > ' + mapping.L4[1] + ' THEN \'L5\' ');
+            bounds.push('ELSE \'NA\' ');
+            bounds.push('END');
+
+            return _.reduce(bounds, function(acc, x) { return acc + ' ' + x; });
+        }
+
+        if (tableId === 'nhd_quality_tn') {
+            return '(SELECT geom, stream_order, ' +
```
mmcfarland: These three queries look identical save for some parameters, I think they could be DRY'd up so that the query is constructed only once with the relevant parameters passed in.

src/tiler/server.js
jtarricone: Ack, good catch.

src/tiler/server.js
jtarricone: True, which I did already with `caseSql()` to handle the most annoying part of the query. A case might reasonably be made for using a stored proc here.

src/mmw/mmw/settings/layer_settings.py
kellyi: re: Layer Selector text: this isn't ideal, but we're currently handling the label for the "Delaware River Basin High Resolution Stream Network" by including a `<br />` tag and some spaces as the value for `display` here.

src/tiler/server.js
```diff
@@ -63,8 +90,44 @@ var interactivity = {
             stream_order = 5;
         }
 
-        return '(SELECT geom, stream_order FROM ' + tableName +
-          ' WHERE stream_order >= ' + stream_order + ') as q';
+        var caseSql = function(mapping, field) {
```
arottersman: Consider renaming this to `waterQualityCaseSql` to specify what it's for/doing

arottersman:
Just tested this....
- I wasn't able to get `underscore` to install with `vagrant reload tiler --provision`, but was able to install it by going into the `tiler` VM and running `npm install` in `/vagrant/src/tiler`
- The coloring on the streams looks all right on most basemaps, except the `terrain` one makes the bright yellow pretty difficult to see

screen shot 2016-10-04 at 6 09 33 pm

- Should this be all red? 

screen shot 2016-10-04 at 6 08 09 pm

- These colors make it look like we bleed google

screen shot 2016-10-04 at 6 09 10 pm

hectcastro:
`underscore` may not have installed cleanly because the NPM dependencies are shrinkwrapped. There is a process for updating the shrinkwrap manifest:

> 1. Run npm install in the package root to install the current versions of all dependencies.
> 2. Add or update dependencies. npm install --save each new or updated package individually to update the package.json and the shrinkwrap. Note that they must be explicitly named in order to be installed: running npm install with no arguments will merely reproduce the existing shrinkwrap.
> 3. Validate that the package works as expected with the new dependencies.
> 4. Commit the new npm-shrinkwrap.json, and publish your package.

See: https://docs.npmjs.com/cli/shrinkwrap#building-shrinkwrapped-packages

src/mmw/mmw/settings/layer_settings.py
jtarricone: @mmcfarland is it a viable option to quietly show two layers (NHD and then water quality) when the water quality overlay is enabled? It avoids the left join and might still be somewhat fast if the NHD tiles are cached.

jtarricone:
>Should this be all red? 

The TSS and TP color mappings were reversed.

It's probably worth creating an issue to add some kind of legend to the map.

src/tiler/server.js
```diff
@@ -154,6 +217,11 @@ var config = {
                 req.params.sql = getSqlForStreamByReq(req);
             }
 
+            if (tableId.indexOf('quality') >= 0) {
```
kellyi: Consider renaming `quality` to `nhd_quality` or `nhd_quality_t` to narrow down the possible matches for this check.

src/tiler/styles.mss
```diff
@@ -44,126 +51,195 @@
...
+  line-cap: round;
+}
+
+#nhdflowline {
+  line-color: @streamColor;
+  line-join: round;
+  line-cap: round;
+}
+
+#nhd_quality_tp {
```
mmcfarland: These style declarations look to be repeated since you did the work of coalescing the values into groups in SQL.  You can render all three with one block if you join the selectors:
```css
#nhd_quality_tp, 
#nhd_quality_tn, 
#nhd_quality_tss {
   [nhd_qual_grp="L1"] {
     line-color: @wtrQualColor0;
   }
  ...
}

**Comment by ajrobbins** [comment]
@jfrankl Can you weigh in on the colors here? Client requested red-yellow-green ramp, I believe we're using these colors https://github.com/WikiWatershed/model-my-watershed/issues/1478#issuecomment-250470268

File: src/tiler/server.js [hunk 8]
```diff
@@ -63,8 +90,44 @@ var interactivity = {
...
+            bounds = _.map(Object.keys(mapping), function(category) {
+                return 'WHEN ' + field + ' >= ' + mapping[category][0] + ' AND ' + field +' < ' + mapping[category][1] + ' THEN \'' + category +'\' ';
+            });
+
+            bounds[0] = 'CASE ' + bounds[0];
+            bounds.push('WHEN ' + field + ' > ' + mapping.L4[1] + ' THEN \'L5\' ');
+            bounds.push('ELSE \'NA\' ');
+            bounds.push('END');
+
+            return _.reduce(bounds, function(acc, x) { return acc + ' ' + x; });
```
mmcfarland: I believe you're reducing this array to join the string elements inside.  You can achieve the same much more simply with `bounds.join(' ')` and we can remove the `underscore` dependency.

src/tiler/styles.mss
```diff
@@ -36,6 +36,13 @@
 }
 
 @streamColor: #1562A9;
+@wtrQualColor0: #d7191c;
+@wtrQualColor1: #fdae61;
+@wtrQualColor2: #ffffbf;
+@wtrQualColor3: #a6d96a;
+@wtrQualColor4: #1a9641;
+@wtrQualColorNA: #648db9;
```
mmcfarland: The blue for `NA` here seems a little too in-line with the symbology for actual data and close to the regular stream color.  I'd suggest a dark gray or something to indicate more clearly that there is no water quality data.

src/mmw/mmw/settings/layer_settings.py
mmcfarland: I considered that as well, but that would still leave us in a situation of rendering blue lines for DRB streams that don't have water quality records, which I don't think they want.

mmcfarland:
You'll also want to update the `setupdb.sh` script so that the cached tiles for these layers are purged when we reload data: https://github.com/WikiWatershed/model-my-watershed/blob/develop/scripts/aws/setupdb.sh#L128-L131  (Add the `layerIds` as `PATHS` and add a call to `purge_tile_cache $PATHS`)

jfrankl:
Thanks for consulting me about the colors. I will need to know more about this feature before I can think about how it needs to work and how it will fit into the application. I will not have time to provide recommended colors during this sprint, but I can get that for you early this coming sprint, which begins tomorrow.

jtarricone:
I refactored to cut underscore out, and while I was changing `setupdb.sh` I noticed that the `PATH`s for the NHD & DRB layers appeared to be wrong, which I would guess would prevent the tile cache from being purged, so I updated those as well.

src/tiler/styles.mss
jtarricone: I updated it to a greyer grey.

jtarricone:
I think `setupdb.sh` is missing a path for school district boundary tiles to purge, is that intentional?

scripts/aws/setupdb.sh
```diff
@@ -102,7 +102,7 @@ fi
 if [ "$load_stream" = "true" ] ; then
     # Fetch stream network layer sql files
     FILES=("nhdflowline.sql.gz")
-    PATHS=("stream")
+    PATHS=("nhd_streams_v2")
```
mmcfarland: These stream updates already have a commit waiting in #1510

src/tiler/server.js
jtarricone: Ahh good look, I'll update it.

arottersman:
These changes look good to me +1

hectcastro:
@azavea-bot rebuild

kellyi:
Heads up that there's a merge conflict.

PR #1508, opened 2016-10-04
Increase SJS JVM stack size to avoid stack overflow

hectcastro:
The serializers Kryo provides use the call stack when serializing nested objects. This change increases the JVM stack size to 5MB so that Kryo serialization can squeeze through.

The stack size increase here is for all threads, so it is possible that this change has a negative impact in other areas if the number of threads grows large.

See: https://github.com/EsotericSoftware/kryo#stack-size

**Testing**

Provision the `worker`, then try to analyze the NJ-04 congressional district. When that's complete, attempt to the Watershed Multi-Year Model. It should complete successfully.

### DIFF
--- a/deployment/ansible/roles/model-my-watershed.spark-jobserver/templates/upstart-spark-jobserver.conf.j2
+++ b/deployment/ansible/roles/model-my-watershed.spark-jobserver/templates/upstart-spark-jobserver.conf.j2
@@ -24,7 +24,7 @@ exec /usr/bin/docker run \
   --volume {{ geop_home }}:{{ geop_home }} \
   --volume {{ sjs_data_dir }}:{{ sjs_data_dir }} \
   --log-driver syslog \
-  {{ sjs_container_image }} --driver-memory {{ sjs_driver_memory }}
+  {{ sjs_container_image }} --driver-memory {{ sjs_driver_memory }} --driver-java-options "-Xss5m -Dlog4j.configuration=file:log4j.properties"
 
 post-stop script
   /usr/bin/docker kill spark-jobserver || true


rajadain:
Before this the NJ-04 Congressional District was giving me a stack overflow error. Now I simply get a "job XXX timed out, deleted." error:

```
[2016-10-04 16:44:51,688: ERROR/Worker-1] Task bb1a394d-673e-43a8-aa83-5315123e5452 run from job 4692 raised exception: [nlcd_soils] Job 58dc6c21-86f6-4d23-8b58-c6802f948962 timed out, deleted.
  File "/opt/app/apps/modeling/mapshed/tasks.py", line 304, in nlcd_soils
    raise Exception('[nlcd_soils] {}'.format(sjs_result['error']))
Exception: [nlcd_soils] Job 58dc6c21-86f6-4d23-8b58-c6802f948962 timed out, deleted.
```

The user feels no difference for this job, since that timeout is still 45 seconds.

I'm not sure what changing 42 to 60 here does, since it seems like we are still deleting a job before it is complete.

I like that we're not getting a stack overflow anymore. Hard to judge how this would affect the system at large, given the caveat above.

hectcastro:
The change from 42 to 60 gives Spark enough time to complete the job before Celery issues a `DELETE` operation. It was an arbitrary value larger than 42 because 42 was reliably terminating the job before I could see if the stack overflow error was resolved.

If the UI timeout needs to be increased to help a job success surface to the user, I think we should consider making it, I'm just not sure where that change needs to be made at.

rajadain:
We'd make that change here: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/models.js#L107

It doesn't seem like Spark is getting a chance before Celery `DELETE`s on my machine, even at 60 seconds of timeout.

PR #1505, opened 2016-10-03
Import `drb_catchment_water_quality` table

kellyi:
This PR updates the `setupdb.sh` script to get the `drb_catchment_water_quality` table from S3. 

I imported the Shapefile mentioned in #1480, added a spatial index, `pg_dump`ed it from the instance of Postgres running on my Services VM, then uploaded it to S3. This update adds the filename to the list of files to get when using the `-q` option to get water quality data.

**Testing**
- get this branch, `vagrant up`, then `vagrant ssh app` and `cd /vagrant/scripts/aws`
- run `./setupdb.sh -q` to import the DRB catchment water quality data
- inside the services VM, run `psql -h localhost -U mmw` to login
- run `SELECT COUNT(*) FROM drb_catchment_water_quality;` and verify that you see

```sql
 count 
-------
 15004
(1 row)
```

- run `\d drb_catchment_water_quality` and verify that you see this output:

```sql
 
                                      Table "public.drb_catchment_water_quality"
   Column   |            Type             |                                 Modifiers                                 
------------+-----------------------------+---------------------------------------------------------------------------
 gid        | integer                     | not null default nextval('drb_catchment_water_quality_gid_seq'::regclass)
 nord       | numeric(10,0)               | 
 areaha     | numeric                     | 
 pop10      | numeric(10,0)               | 
 tn_tot_kgy | numeric                     | 
 tn_urban_k | numeric                     | 
 tn_riparia | numeric                     | 
 tn_ag_kgyr | numeric                     | 
 tn_natural | numeric                     | 
 tn_pt_kgyr | numeric                     | 
 tp_tot_kgy | numeric                     | 
 tp_urban_k | numeric                     | 
 tp_riparia | numeric                     | 
 tp_ag_kgyr | numeric                     | 
 tp_natural | numeric                     | 
 tp_pt_kgyr | numeric                     | 
 tss_tot_kg | numeric                     | 
 tss_urban_ | numeric                     | 
 tss_ag_kgy | numeric                     | 
 tss_natura | numeric                     | 
 tss_rip_kg | numeric                     | 
 geom       | geometry(MultiPolygon,4326) | 
Indexes:
    "drb_catchment_water_quality_pkey" PRIMARY KEY, btree (gid)
    "drb_catchment_water_quality_geom_idx" gist (geom)
```

Connects #1480

PR #1509, opened 2016-10-05
Render DRB Catchment Water Quality tiles

kellyi:
This PR updates the layer settings and tiler to render TN, TP, and TSS loading rates for DRB Catchments from the `drb_catchment_water_quality` table added in #1505. It also adds styling rules to establish a five-step gray color-break scheme to render the values according to the intervals noted in #1482. In addition, I've updated the `setupdb.sh` aws script to purge the tiles for these layers on reloading the table.

The result currently looks like this:

DRB Catchment TN:

screen shot 2016-10-05 at 12 03 10 pm

DRB Catchment TP:

screen shot 2016-10-05 at 12 03 17 pm

and DRB Catchment TSS:

screen shot 2016-10-05 at 12 03 24 pm

**Testing**
- get this branch, `vagrant up`, `vagrant ssh app`, `cd /vagrant/scripts/aws` and `./setupdb.sh -q` to get the `drb_catchment_water_quality_table`
- visit `localhost:8000` in the browser, pan to the DRB area, and open the layer selector's overlays tab
- open the browser console, network tab, and "img" to see the tiles as they're downloaded
- the new layers are currently in the "Boundary" section. Toggle each of "DRB Catchment... TN", "TP", and "TSS" on and verify that you see the tiles rendering, that they're being requested from endpoints concluding with `_tn`, `_tp`, and `_tss` respectively, and that the values change when you toggle from one layer to another
- verify that you haven't seen any errors in the browser console

Connects #1482

### DIFF
--- a/src/mmw/mmw/settings/layer_settings.py
+++ b/src/mmw/mmw/settings/layer_settings.py
@@ -233,6 +233,36 @@ LAYERS = [
         'opacity': 0.618,
         'has_opacity_slider': True,
         'perimeter': pa_perimeter,
+    },
+    {
+        'code': 'drb_catchment_water_quality_tn',
+        'display': ('DRB Catchment Water Quality Data' +
+                    '<br />&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;TN Loading Rates'),
+        'table_name': 'drb_catchment_water_quality_tn',
+        'vector': True,
+        'overlay': True,
+        'minZoom': 3,
+        'perimeter': drb_perimeter
+    },
+    {
+        'code': 'drb_catchment_water_quality_tp',
+        'display': ('DRB Catchment Water Quality Data' +
+                    '<br />&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;TP Loading Rates'),
+        'table_name': 'drb_catchment_water_quality_tp',
+        'vector': True,
+        'overlay': True,
+        'minZoom': 3,
+        'perimeter': drb_perimeter
+    },
+    {
+        'code': 'drb_catchment_water_quality_tss',
+        'display': ('DRB Catchment Water Quality Data' +
+                    '<br />&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;TSS Loading Rates'),
+        'table_name': 'drb_catchment_water_quality_tss',
+        'vector': True,
+        'overlay': True,
+        'minZoom': 3,
+        'perimeter': drb_perimeter
     }
 ]
 
--- a/src/tiler/server.js
+++ b/src/tiler/server.js
@@ -36,8 +36,15 @@ var interactivity = {
         drb_streams_v2: 'drb_streams_50',
         nhd_streams_v2: 'nhdflowline',
         municipalities: 'dep_municipalities',
-        urban_areas: 'dep_urban_areas'
+        urban_areas: 'dep_urban_areas',
+        // The DRB Catchment tables here use aliases to match data from different
+        // columns in the `drb_catchment_water_quality` table to different
+        // rendering rules.
+        drb_catchment_water_quality_tn: 'drb_catchment_water_quality_tn',
+        drb_catchment_water_quality_tp: 'drb_catchment_water_quality_tp',
+        drb_catchment_water_quality_tss: 'drb_catchment_water_quality_tss'
     },
+    drbCatchmentWaterQualityTable = 'drb_catchment_water_quality';
     shouldCacheRequest = function(req) {
         // Caching can happen if the bucket to write to is defined
         // and the request is not coming from localhost.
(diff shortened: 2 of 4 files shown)


arottersman:
I was able to get the catchment data and view it on the map! 

I had some trouble though with the update for purging tiles. When I ran `setupdb.sh -q` I got this after all the SQL statements:

```
vagrant@app:/vagrant/scripts/aws$ ./setupdb.sh -q
...
+ purge_tile_cache drb_catchment_water_quality_tn
+ for path in '"${PATHS[@]}"'
+ aws s3 rm --recursive s3://tile-cache./drb_catchment_water_quality_tn/
./setupdb.sh: line 74: aws: command not found
```

Then while running the app, I got the same error that Rollbar was spitting out earlier today (https://rollbar.com/Azavea/MMW/items/131/) :
screen shot 2016-10-05 at 3 36 04 pm
screen shot 2016-10-05 at 3 35 56 pm

kellyi:
Oh! Can you try running `./setupdb.sh -m` to get the MapShed data if you haven't yet? That should fix the `ms_pointsource_drb` issue. The app makes an initial request to that table to create the DRB Point Source overlay.

I *think* the `command not found` error will always happen on local instances of this since there may be nothing to clear.

mmcfarland:
@arottersman - you'll want to run `setupdb.sh -m` to load up the mapshed data which the `ms_point_source` error refers to.  Also, the `aws` errors are expected since we don't want to purge the staging tile cache when someone runs a data update on their dev box.

src/tiler/server.js
```diff
@@ -164,6 +168,26 @@ var config = {
                 req.params.sql = getSqlForStreamByReq(req);
             }
 
+            if (tableId.indexOf('drb_catchment') >= 0) {
+                switch (tableId) {
```
mmcfarland: Consider removing this to its own function like `getSqlForStreamsByReq` to keep this function small

arottersman:
Oh, yeah, loading the Mapshed data did the trick! No more errors.

src/tiler/server.js
```diff
@@ -36,8 +36,12 @@ var interactivity = {
         drb_streams_v2: 'drb_streams_50',
         nhd_streams_v2: 'nhdflowline',
         municipalities: 'dep_municipalities',
-        urban_areas: 'dep_urban_areas'
+        urban_areas: 'dep_urban_areas',
+        drb_catchment_water_quality_tn: 'drb_catchment_water_quality_tn',
```
mmcfarland: I'm wondering if a comment is appropriate here.  In other cases this table map actually does map to a db table, where as this maps to an alias, which is referenced in the style sheet.  The actual table is then defined later.  Since it breaks the pattern, it might good to point that out specifically.

src/mmw/mmw/settings/layer_settings.py
```diff
@@ -233,6 +233,33 @@
...
+        'vector': True,
+        'overlay': True,
+        'minZoom': 3,
+        'perimeter': drb_perimeter
+    },
+    {
+        'code': 'drb_catchment_water_quality_tss',
+        'display': 'DRB Catchment Water Quality Data TSS Loading Rates',
+        'table_name': 'drb_catchment_water_quality_tss',
+        'vector': True,
```
mmcfarland: On the fence if these should be added as "Boundary" or "Coverage".  My instinct is to say "Coverage" since it isn't just a boundary outline like the others in that category.  It's more similar to the coverages which symbolize values as filled areas.   Not a strong opinion, but worth considering.

kellyi: That makes sense. Another consideration here would be which other layers we might want to allow/prevent from being turned on when these layers are on: would it be preferable to allow these layers to be on in tandem with the boundary layers or in tandem with the coverage layers.

src/tiler/server.js
kellyi: Sounds good -- I'll update with this change.

src/mmw/mmw/settings/layer_settings.py
arottersman: Not sure how people actually use these layers, but this looks really neat:

screen shot 2016-10-05 at 4 09 42 pm

arottersman:
Should the PA Urbanized area Coverage layer be working?

kellyi:
Yeah, if you run `./setupdb.sh -p` to get the DEP tables they should work.

src/tiler/server.js
kellyi: Good point. Going to add this comment to explain why the disharmony:

```js
// The DRB Catchment tables use aliases here to match data from different
// columns in the `drb_catchment_water_quality` table to different
// rendering rules.
```

hectcastro:
@azavea-bot rebuild

PR #1510, opened 2016-10-05
Update stream paths for cache purge

mmcfarland:
### Overview
In order to see styling updates immediately and not wait for a TTL to
expire on CloudFront, these paths were updated periodically.  Now that
it's settled, syncing up correct purge paths so the cache is cleared
when the data is reloaded in the future.

### Testing
Verify that the paths associated with the datasets are the same as those used by the tiler
https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/tiler/server.js#L36-L37

### DIFF
--- a/scripts/aws/setupdb.sh
+++ b/scripts/aws/setupdb.sh
@@ -102,7 +102,7 @@ fi
 if [ "$load_stream" = "true" ] ; then
     # Fetch stream network layer sql files
     FILES=("nhdflowline.sql.gz")
-    PATHS=("stream")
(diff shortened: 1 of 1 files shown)


kellyi:
+1

Stream changes from this PR:

screen shot 2016-10-05 at 3 40 50 pm

School districts layer name in this PR:

screen shot 2016-10-05 at 3 42 12 pm

NHD & DRB Stream + School District names in the tiler:

screen shot 2016-10-05 at 3 40 59 pm

rajadain:
+1

Issue #1525, opened 2016-10-10
Address tab overflow issues


jfrankl:
![image](https://cloud.githubusercontent.com/assets/1809908/19249909/d4a33346-8f05-11e6-93b2-8d9f03f6b5a8.png)

As we add additional tabs to the sidebar, we are beginning to experience spacing issues. In the short term, it might be sufficient to reduce the space between tabs. In the long term, we will need to consider more complex fixes.

1. While tab content loads, spinners are displayed inside of tabs, which causes them to take up more space; we could find another way to communicate about loading which will not impact the size of tabs
2. At six or seven tabs, we might want to begin to consider a different UI; after a certain point, a dropdown or other interface might be more effective than tabs

PR #1526, opened 2016-10-10
WIP: Add Catchment Water Quality Analysis Tab

arottersman:
Adds a new tab to the analysis view with the NHD Catchment Water Quality Data. So far, we only have data for the DRB. Like with the point source data, the corresponding catchment will highlight on the map on mouse over of cells in the water quality table. When you click the row's id, a popup appears with more info about that catchment. 

screen shot 2016-10-10 at 5 55 12 pm

Connects #1518


src/mmw/js/src/analyze/views.js
```diff
@@ -466,6 +469,104 @@ var PointSourceTableView = Marionette.CompositeView.extend({
     }
 });
 
+var CatchmentWaterQualityTableRowView = Marionette.ItemView.extend({
```
arottersman: With this fifth tab, I found this file to be awkwardly large - maybe we should be splitting off each view into its own file now?

kellyi:
About to take a look at this! Looks like the CI server caught some JS lint.

kellyi:
Water quality tab looks to get pushed to a second row and hidden behind the total area text when I view this on my MacBook screen at 100% size:

screen shot 2016-10-12 at 2 12 02 pm

kellyi:
Happening on a wider viewport, too:

screen shot 2016-10-12 at 2 14 21 pm

kellyi:
Here are some example water quality values; looks like it must return all the catchment shapes which intersect with the drawn aoi:

[water_quality_values.json.txt](https://github.com/WikiWatershed/model-my-watershed/files/525336/water_quality_values.json.txt)

src/mmw/js/src/analyze/tests.js
```diff
@@ -129,19 +129,35 @@ function pointsourceTableFormatter(categories) {
...
 
+function catchmentWaterQualityTableFormatter(categories) {
+    var collection = new coreModels.CatchmentWaterQualityCensusCollection(categories);
+
+    return collection.map(function(category) {
+        var nord = category.get('nord').toString(),
+            areaha = category.get('areaha') === null ? 'No Data' : category.get('areaha').toLocaleString(),
+            tn_tot_kgy = category.get('tn_tot_kgy') === null ? 'No Data' : (category.get('tn_tot_kgy')/category.get('areaha')).toLocaleString(),
+            tp_tot_kgy = category.get('tp_tot_kgy') === null ? 'No Data' : (category.get('tp_tot_kgy')/category.get('areaha')).toLocaleString(),
+            tss_tot_kg = category.get('tss_tot_kg') === null ? 'No Data' : (category.get('tss_tot_kg')/category.get('areaha')).toLocaleString();
```
kellyi: If `areaha` is `null` or 0, should this skip the other checks? I checked in the Node repl and the browser console and it looks like dividing an actual number by 0 or `null` returns infinity.

src/mmw/js/src/analyze/tests.js
```diff
@@ -129,19 +129,35 @@ function pointsourceTableFormatter(categories) {
...
 };
 
 var tableHeaders = {
     land: ['Type', 'Area (km2)', 'Coverage (%)'],
     soil: ['Type', 'Area (km2)', 'Coverage (%)'],
     animals: ['Animal', 'Count'],
     pointsource: ['NPDES Code', 'City', 'Discharge (MGD)', 'TN Load (kg/yr)', 'TP Load (kg/yr)'],
-};
+    catchment_water_quality: ['Id', 'Area (ha)', 'Total N (kg/ha)', 'Total P (kg/ha)', 'Total TSS (kg/ha)'],
+}
```
kellyi: I *think* the missing semi-colon here is one of the linter errors.

src/mmw/js/src/analyze/views.js
kellyi: I feel like it makes sense to do the splitting of this file in tandem with the splitting of the UI in #1525

src/mmw/apps/modeling/calcs.py
```diff
@@ -92,3 +92,42 @@ def point_source_pollution(geojson):
...
+    from the `drb_catchment_water_quality` table to display
+    in the Analyze tab.
+
+    Returns a dictionary to append to the outgoing JSON for analysis
+    result
+    """
+    geom = GEOSGeometry(geojson, srid=4326)
+    table_name = 'drb_catchment_water_quality'
+    sql = '''
+    SELECT nord, areaha, tn_tot_kgy, tp_tot_kgy, tss_tot_kg,
```
kellyi: Consider indenting the sql statement here to match the point source sql's indentation.

src/mmw/apps/modeling/calcs.py
```diff
@@ -92,3 +92,42 @@ def point_source_pollution(geojson):
...
+    WHERE ST_Intersects(geom, ST_SetSRID(ST_GeomFromText(%s), 4326))
+    '''.format(table_name=table_name)
+
+    with connection.cursor() as cursor:
+        cursor.execute(sql, [geom.wkt])
+
+        if cursor.rowcount != 0:
+            columns = [col[0] for col in cursor.description]
+            catchment_water_quality_results = [
+                dict(zip(columns, row))
```
kellyi: It looks like the point source function does some coalescing of the return values to replace null values with `None`: https://github.com/WikiWatershed/model-my-watershed/blob/d927dbb712265254982c87b33d8308e9d0eac5e7/src/mmw/apps/modeling/calcs.py#L82

I think you've got some checks for that on the frontend, but you could consider coalescing the values here, too, and perhaps reduce the number of frontend checks.

kellyi:
I manually removed the point source tab html element and was able to see the catchment water quality tab:

screen shot 2016-10-12 at 2 49 07 pm

Here it is for a smaller AOI:

screen shot 2016-10-12 at 2 51 26 pm

kellyi:
This is not the best way to do this but I was able to get all the tabs to fit on the same line by giving the `.nav-tabs > li` css class a rule for `margin-right: -5px;`:

screen shot 2016-10-12 at 2 54 01 pm

I think if we're going to make a change like that in the near term, it should be targeted only to these specific tab elements -- which probably means giving the tabs themselves a very specific class to add a rule too which wouldn't impact anything else.

src/mmw/apps/modeling/calcs.py
arottersman: I thought the coalescing was to cast some of the values to `float` (which I should probably also be doing). If there's no value for a column in the database, won't they come back as `None` without coalescing?

kellyi: Oh, right -- I think thinks actually come back as `Decimal` values from the database, which is why we adjusted the types there.

arottersman:
- Added the casts to `float` in `modeling/calcs.py` (it doesn't look quite right...thank goodness for atom's multiline editing!)
- Added missing semi and fixed some styling
- added a `filterNoData` filter to `nunjunk` that cleans up some of the many checks we do for our tables

Should I address the tab spacing in this PR or leave it to what you've done in #1538?

src/mmw/apps/modeling/calcs.py
```diff
@@ -92,3 +92,64 @@ def point_source_pollution(geojson):
...
+        cursor.execute(sql, [geom.wkt])
+
+        if cursor.rowcount != 0:
+            columns = [col[0] for col in cursor.description]
+            catchment_water_quality_results = [
+                # The TN, TP, and TSS values return as type Decimal,
+                # but we want floats.
+                dict(zip(columns,
+                         [row[0],
+                          float(row[1]) if row[1] else None,
```
arottersman: 🤔

kellyi:
About to take another look at this!

arottersman:
Sorry about the linting errors - correcting them now. I keep expecting them to appear in the output of `./scripts/bundle.sh --debug --watch`

kellyi:
re: Tab Spacing, I think we should address it here as the last commit in this PR since it seems logically to connect with this work. I'm going to test the CSS a bit to see what might work best and I'll report back if I find something. I believe the shim I wrote just puts a `margin-right` of -10px for one of the list item classes, but I don't think that's a good solution as I don't know what else it might impact.

Best solution would be to find/create a single specific class for the tab header labels to minimize the width of their columns. @jfrankl might have some suggestions about existing classes which would target that, or other classes to add.

kellyi:
This seems to work very well, btw! 

One additional quirk I noticed is that when the analyze tab is shown, the zoom level of the map depends on the drawn AOI. In this case, some of the catchment boundaries extend outside the viewport containing the map if the zoom level's very high and the catchment intersects with an edge then extends outward, like when you select a 1 sq km aoi. This also means that when the popup shows, it sometimes pans the map such that the AOI is no longer in the viewport.

Not quite sure what would be the best fix for this, as I don't know that we want to adjust the zoom level of the map when panning. May be worth checking with @mmcfarland and @ajrobbins to see what they suggest.

Here's an example for reference:

screen shot 2016-10-13 at 10 27 44 am

->

screen shot 2016-10-13 at 10 27 51 am

arottersman:
Just pushed the lint fixes. Maybe the build with finally pass now!

RE: Popup Outside the AOI
I'd be a fan of zooming out to show the full catchment, and then zooming back into the AOI once the popup is gone. Would that make people motionsick? I think if you are inspecting a particular catchment in the table, you might want to see it in full on the map. Related to this - should the popup appear centered on the catchment, or is it okay it's on the border of the shape?

arottersman:
This is ready for another look now
- @jfrankl is A+ and fixed the tab spacing/moved the loader spinnies. Thanks, Jeff!
screen shot 2016-10-13 at 4 36 55 pm

- Kelly helped me work out how to deal with the popup placement - we're now using `turf-intersect` and `turf-centroid` to find the center of the intersection of the AOI and the catchment. Now it won't pan you away from the shape you drew:
screen shot 2016-10-13 at 4 25 43 pm

kellyi:
Cool -- going to take another look.

kellyi:
+1. This works really well! 

I tested it out with a few AOIs including congressional districts and sq kms. The popups still now show up in the intersection even when the catchment extends beyond the original shape.

Nice work!

arottersman:
Thanks for looking at this, Kelly!

Issue #1491, opened 2016-09-27
Mapshed celery job setup time


mmcfarland:
The process of initiating a MapShed model run involves an initial POST to `/mapshed` which sets up and submits the jobs, returning the jobId which is later polled against.  This POST can be from 10-20 seconds in my testing, almost entirely consumed by the [setup and chaining of celery tasks](https://github.com/WikiWatershed/model-my-watershed/blob/5d04999b60edfbcccc88df5878dd29a1d4a3ee1a/src/mmw/apps/modeling/views.py#L263-L270).  This submission time is included in the overall job time as reported by the polling endpoint, and can account for up to 50% of the total MapShed job time.

The executing line is dense, it may warrant breaking up into more discrete chunks in order to assess how that time is allocated across the different methods and functions executed.  This may reveal areas where the job submission can be sped up.

There may be some timing issues shared with #1490, but in my testing the timing of job submissions was more substantial.

PR #1529, opened 2016-10-11
Split up & speed up  Mapshed job setup

jtarricone:
This addresses a couple issues with the setup process for Mapshed jobs. First, it breaks up the setup task into smaller tasks, which, in general, are easier to debug and profile both for purposes of this issue as well as in future. Second, it makes some minor changes to the implementation with the aim of speeding the process up in general, primarily by removing a notable blocking method invocation as well as parallelizing the process where possible.

To test:
* clone this branch
* it may be worth changing `Vagrantfile` to bump up the settings for `worker`; I was testing it with 2 CPUs and 2048 GB memory
* bring up the environment
* go through typical user actions with the app, an ensure that the app behavior is consistent with expectations

The `POST` to `/api/modeling/start/mapshed` should be faster than it was before or currently is on staging (the call in staging should take 18-20 seconds, this implementation could take 2-8 seconds depending on the VM resources). The overall model run was also faster in my testing, but there are several mitigating factors if a local version is compared to staging, so it's probably not worth spending much time evaluating.

Connects #1491


src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -456,31 +457,56 @@ def nlcd_kfactor(result):
...
-
-    return [(mapshed_start.s(opname, data).set(exchange=exchange,
-                                               routing_key=geop_worker) |
-             mapshed_finish.s().set(exchange=exchange,
-                                    routing_key=geop_worker) |
-             callback.s().set(link_error=errback,
-                              exchange=exchange,
-                              routing_key=choose_worker()))
-            for ((opname, data, callback), geop_worker) in tasks]
+@shared_task(bind=True, default_retry_delay=1, max_retries=42)
```
rajadain: I don't think we should be retrying the entire chain of `geop_task`. The `mapshed_start` and `mapshed_finish` tasks have retries specified on them and that should be enough.

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -456,31 +457,56 @@ def nlcd_kfactor(result):
 
     return output
 
+def geop_tasks():
+    return geop_task_defs().keys()
+
+def geop_task_defs():
```
rajadain: Can this be stored as a dictionary instead of a function?

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -456,31 +457,56 @@ def nlcd_kfactor(result):
...
+    try:
+        (opname, data, callback) = geop_task_defs()[taskName](geom)
+        geop_worker = choose_worker()
+        return (mapshed_start.s(opname, data).set(exchange=exchange,
+                                                   routing_key=geop_worker) |
+                 mapshed_finish.s().set(exchange=exchange,
+                                        routing_key=geop_worker) |
+                 callback.s().set(link_error=errback,
+                                  exchange=exchange,
+                                  routing_key=geop_worker))
```
rajadain: The callback doesn't need to happen on the same `geop_worker` as `mapshed_start` or `mapshed_finish`. We can safely use `choose_worker()` here.

src/mmw/apps/modeling/views.py
```diff
@@ -254,22 +256,29 @@ def start_mapshed(request, format=None):
...
 
-    chain = (group(geop_tasks(geom, errback, MAGIC_EXCHANGE, choose_worker)) |
-             combine.s().set(exchange=MAGIC_EXCHANGE,
-                             routing_key=choose_worker()) |
-             collect_data.s(geom.geojson).set(link_error=errback,
-                                              exchange=MAGIC_EXCHANGE,
-                                              routing_key=choose_worker()) |
-             save_job_result.s(job_id, mapshed_input)
-             .set(exchange=MAGIC_EXCHANGE, routing_key=choose_worker()))
+    job_chain = []
```
rajadain: Is this switch from the pipe-syntax to creating an array and chaining at the end the main thing that provides the speed up?

src/mmw/apps/modeling/views.py
```diff
@@ -27,13 +27,15 @@
 from apps.core.models import Job
 from apps.core.tasks import save_job_error, save_job_result
 from apps.modeling import tasks
-from apps.modeling.mapshed.tasks import geop_tasks, collect_data, combine
+from apps.modeling.mapshed.tasks import (geop_tasks, collect_data, combine,
+                                         geop_task_defs, geop_task, geop_tasks)
```
rajadain: `geop_tasks` is imported twice, and `geop_task_defs` is never used.

src/mmw/apps/modeling/mapshed/tasks.py
jtarricone: Yeah, it would flatten things a bit as well.

src/mmw/apps/modeling/views.py
```diff
@@ -254,22 +256,29 @@ def start_mapshed(request, format=None):
...
+    workers = choose_workers()
+    get_worker = lambda: random.choice(workers)
     errback = save_job_error.s(job_id).set(exchange=MAGIC_EXCHANGE,
-                                           routing_key=choose_worker())
+                                           routing_key=get_worker())
 
     geom = GEOSGeometry(json.dumps(mapshed_input['area_of_interest']),
                         srid=4326)
 
-    chain = (group(geop_tasks(geom, errback, MAGIC_EXCHANGE, choose_worker)) |
```
kdeloach: Interesting that this used to shadow the celery chain function, we must have overlooked that.

src/mmw/apps/modeling/views.py
```diff
@@ -254,22 +256,29 @@ def start_mapshed(request, format=None):
 
 
 def _initiate_mapshed_job_chain(mapshed_input, job_id):
+    workers = choose_workers()
+    get_worker = lambda: random.choice(workers)
```
kdeloach: This seems to be the most significant difference between the old version. Instead of making a query for a celery worker for each task in the job chain, you query the list of workers once for the whole job chain.

I'd be curious to see how much of an impact this change has by itself, without any of the other changes.

src/mmw/apps/modeling/views.py
jtarricone: I think there was a small benefit to doing that, but the main block was `choose_worker`. I wasn't able to get a solid grasp on when parts of chained tasks ended up being evaluated, nor exactly how parallel the operations ended up being.

src/mmw/apps/modeling/views.py
jtarricone: Pooling the workers for each job made the biggest difference, but I did notice a small change just from splitting the task up as well, although I don't have as clear an idea on why exactly that might be the case. I think that the former implementation would be equivalent with enough workers or higher degrees of concurrency, but I'm similarly unsure about what happens in the even of failure in each case.

kdeloach:
For gathering performance benchmarks you can use [flower](https://pypi.python.org/pypi/flower). We had it installed at one point but it was [removed](https://github.com/WikiWatershed/model-my-watershed/commit/86a64aa48b6794c52fc22c8668efc5d9791d5c72).

Quick start guide:

```
vagrant ssh worker
sudo pip install flower
cd /opt/app
# Note: You'll need to update your Vagrantfile to setup port forwarding
envdir /etc/mmw.d/env/ celery flower -A mmw.celery:app --address=0.0.0.0 --port=5555
```

screenshot 2016-10-11 at 16 29 25

kdeloach:
Are there any additional steps to test this? I get this error when running mapshed with `./scripts/debugserver.sh`:

```
Internal Server Error: /api/modeling/start/mapshed/
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/views.py", line 452, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/views.py", line 449, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/decorators.py", line 50, in handler
    return func(*args, **kwargs)
  File "/opt/app/apps/modeling/views.py", line 245, in start_mapshed
    task_list = _initiate_mapshed_job_chain(mapshed_input, job.id)
  File "/opt/app/apps/modeling/views.py", line 281, in _initiate_mapshed_job_chain
    return chain(job_chain).apply_async(link_error=errback)
  File "/usr/local/lib/python2.7/dist-packages/celery/canvas.py", line 251, in apply_async
    return _apply(args, kwargs, **options)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/builtins.py", line 277, in apply_async
    tasks, results = self.prepare_steps(args, kwargs['tasks'])
  File "/usr/local/lib/python2.7/dist-packages/celery/app/builtins.py", line 237, in prepare_steps
    res = task.freeze()
  File "/usr/local/lib/python2.7/dist-packages/celery/canvas.py", line 539, in freeze
    task = maybe_signature(task, app=self._app).clone()
  File "/usr/local/lib/python2.7/dist-packages/celery/canvas.py", line 197, in clone
    s = Signature.from_dict({'task': self.task, 'args': tuple(args),
TypeError: 'NoneType' object is not iterable
```

jtarricone:
>For gathering performance benchmarks you can use flower. We had it installed at one point but it was removed.

I actually did install it but I didn't get to try it out.

kdeloach:
It's difficult to review this without benchmarks. We don't have celery configured in such a way to easily collect and compare metrics. 

My instinct is that the changes to how workers are selected per task is the biggest significant improvement, and that's a change that we should definitely keep. The rest of the changes seem to be a slightly different way of doing the same thing.

I would recommend splitting these changes into several commits and pull requests, so that we can test each change against the staging environment instead of speculating.

jtarricone:
>The rest of the changes seem to be a slightly different way of doing the same thing.

Breaking up the setup was actually (I thought) the main point of the issue, so that it would be easier to see what was happening.  I incidentally noticed that the worker selection was a major source of blockage while I was trying to increase the parallelism.

>My instinct is that the changes to how workers are selected per task is the biggest significant improvement, and that's a change that we should definitely keep.

I had previously mentioned that that was the major problem, it was about 1200ms per invocation. I'm not familiar enough with celery to know for sure whether the prior implementation of `geop_task`  actually executed each piece in parallel, but I think this one does.

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -457,30 +458,58 @@ def nlcd_kfactor(result):
...
+                                       routing_key=geop_worker) |
+                callback.s().set(link_error=errback,
+                                 exchange=exchange,
+                                 routing_key=geop_worker))
+    except Retry as r:
+        raise r
+    except Exception as x:
+        return {
+            'error': x.message
+        }
```
rajadain: Regarding swallowing exceptions: this was done in the first two tasks in the three-task chains here:

mmw celery workflow

Because if there was an exception in the inner chain on any task except the last task, the chord would never unlock and Celery would be stuck in a loop forever.

Here's an excerpt from my upcoming blog post about the matter:

> #### Error Handling in Nested Chains
> Even though the MapShed workflow shown above is complex with multiple specialized tasks, the error handling is very general. Regardless of the task that raises an exception or encounters an error, we simply want to log it to the database and return to the client. We assumed that given its general applicability, we could specify the error handler once on the parent chain and have it be catch errors raised anywhere in the constituent tasks. Unfortunately this is not true, and we must attach the error handler separately to every task in every subchain. We do this by [defining the error handling task once](https://github.com/WikiWatershed/model-my-watershed/blob/3c4f7e2123481d9909d18d1dad9699ed3c24aebe/src/mmw/apps/modeling/views.py#L257-L258), and [passing it as a parameter](https://github.com/WikiWatershed/model-my-watershed/blob/3c4f7e2123481d9909d18d1dad9699ed3c24aebe/src/mmw/apps/modeling/views.py#L263) to the function that [creates all geoprocessing task chains](https://github.com/WikiWatershed/model-my-watershed/blob/3c4f7e2123481d9909d18d1dad9699ed3c24aebe/src/mmw/apps/modeling/mapshed/tasks.py#L488).

rajadain: I think the issue was refactoring in this task. The forwarded error needs to be thrown by the next step as done [here](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/mapshed/tasks.py#L78-L79).

```python
def mapshed_finish(self, incoming):
    if 'error' in incoming:
        return incoming
```

We really don't need to be doing this here though.

rajadain: And it needs to be caught in the end, as done [here](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/mapshed/tasks.py#L275-L276) and [here](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/mapshed/tasks.py#L303-L304), etc.

src/mmw/apps/modeling/views.py
```diff
@@ -260,16 +262,20 @@ def _initiate_mapshed_job_chain(mapshed_input, job_id):
...
+                     routing_key=choose_worker()) |
+                 collect_data.s(geom.geojson).set(
+                     link_error=errback,
+                     exchange=MAGIC_EXCHANGE,
+                     routing_key=choose_worker()) |
+                 save_job_result.s(job_id, mapshed_input).set(
+                     exchange=MAGIC_EXCHANGE, routing_key=choose_worker()))
 
-    return chain.apply_async(link_error=errback)
+    return chain(job_chain).apply_async(link_error=errback)
```
kdeloach: I suspect the only reason this was slow originally is because each of those geop tasks calls `choose_worker`. Since you fixed that, we may no longer need to refactor how the job chain is composed.

src/mmw/mmw/settings/base.py
```diff
@@ -130,6 +130,7 @@ def get_env_setting(setting):
 CELERY_CREATE_MISSING_QUEUES = True
 CELERY_CHORD_PROPAGATES = True
 CELERY_CHORD_UNLOCK_MAX_RETRIES = 60
+# CELERYD_CONCURRENCY = 2
```
hectcastro: We are currently running 2 Celery workers per machine, with a worker concurrency of 1. See `celery_number_of_workers` and `celery_processes_per_worker` [here](https://github.com/WikiWatershed/model-my-watershed/blob/develop/deployment/ansible/roles/model-my-watershed.celery-worker/templates/celeryd-defaults.j2).

If we plan to tweak these settings, I think it would be good to keep the changes within Ansible so that they can live closer to the process that drives the instance type Celery resides on. Also, tweaking this setting will require monitoring Spark carefully, as the increase in workers can increase the number of things submitted to Spark at a time.

src/mmw/apps/modeling/views.py
```diff
@@ -373,7 +382,11 @@ def get_list_of_workers():
 
     workers = filter(predicate,
                      get_list_of_workers())
-    return random.choice(workers)
+    return workers
+
+
+def choose_worker():
```
hectcastro: One potential downside to changing the way this works is that the list of workers `choose_worker()` operates on grows stale as work is being done. I don't know that that is super likely, but are we currently asking Celery for available workers enough that this optimization gets us a lot?

jtarricone: Correct me if I misunderstand your point, but within the context of `choose_worker`'s use I think the list of available workers will always be up-to-date at the time of invocation since it's just a thin selector over `get_living_workers`, unless the python interpreter does some type of caching or memoization internally. From the perspective of all consuming methods, I *think* the change to `choose_worker` is invisible; I split `get_living_workers` into a separate declaration so that I could bind the result for re-use specifically in `_initiate_mapshed_job` and leave every other use unaffected.

To your question, pinging celery for each job was a bottleneck with noticeable effects, especially for job chains with more than a couple jobs and where the jobs themselves are not especially long-running. You're correct that the list of workers in the mapshed chain could grow stale, but the chain itself isn't very long running, and qualitatively I think there's some confidence in the stability of the workers in the +10 second time frame.

I also created #1535 to address the concern of leaving changes in celery workers unhandled.

src/mmw/mmw/settings/base.py
jtarricone: Good point about Spark. I definitely don't advocate tweaking it until we have some kind of test in place that can mimic each type of potential resource or concurrency problem, but given the time/money required it may prove to be a worthy trade at some point.

Issue #1514, opened 2016-10-06
Defer point-source observation call


mmcfarland:
When the app loads, an api call to /point-source is made to get point source observations for the layer selector.  The request is fast and has aggressive cache headers, but still effects the page load time.  Consider an alternate event to load this data, such as opening the layer selector.

PR #1530, opened 2016-10-11
Fetch point-source data after layer control is opened

lewfish:
The fetch to `/point-source` is slowing down the initial page loading time. Therefore, we now do the fetch after the layer control is opened. While the data is loading, a loading message is displayed.

#### Testing
* In the network tab, when the page loads, check that `/point-source` is not loaded.
* Throttle the network connection so it is slow. Then open the layer selector and check that it displays a 'Loading...' message in the observations tab.
* Eventually, the data should load and the observations layers should appear and function as before.
* Close the layer selector and open it again. There should not be another request to `/point-source`.

Connects #1514


caseypt:
+1. Tested. Works well.

Issue #1533, opened 2016-10-11
Add back call to load files and purge cache to water quality data


caseypt:
Accidentally removed in https://github.com/WikiWatershed/model-my-watershed/commit/780415d73ab19f4cb21881a775010c3f3c3f4f49

```diff
diff --git a/scripts/aws/setupdb.sh b/scripts/aws/setupdb.sh
index 8f8e16a..eb322ea 100755
--- a/scripts/aws/setupdb.sh
+++ b/scripts/aws/setupdb.sh
@@ -131,4 +131,7 @@ if [ "$load_water_quality" = "true" ] ; then
     FILES=("nhd_water_quality.sql.gz" "drb_catchment_water_quality.sql.gz")
     PATHS=("drb_catchment_water_quality_tn" "drb_catchment_water_quality_tp"
             "drb_catchment_water_quality_tss" "nhd_quality_tp" "nhd_quality_tn")
+
+    download_and_load $FILES
+    purge_tile_cache $PATHS
 fi
```

kellyi:
Fixing this as part of #1531

PR #1534, opened 2016-10-11
Restore missing lines in setupdb.sh script

kellyi:
This PR restores the missing lines needed to download the water quality tables and purge the tile cache which got cut inadvertently.

**Testing**
- run the `setupdb.sh` script with the `-q` flag to load the water quality data. Verify that the operation completes successfully and that it the database tables on your local. 

`drb_catchment_water_quality` should have 15004 rows
`nhd_water_quality` should have 10957 rows

Connects #1533

### DIFF
--- a/scripts/aws/setupdb.sh
+++ b/scripts/aws/setupdb.sh
@@ -131,4 +131,7 @@ if [ "$load_water_quality" = "true" ] ; then
     FILES=("nhd_water_quality.sql.gz" "drb_catchment_water_quality.sql.gz")
     PATHS=("drb_catchment_water_quality_tn" "drb_catchment_water_quality_tp"
             "drb_catchment_water_quality_tss" "nhd_quality_tp" "nhd_quality_tn")
+
+    download_and_load $FILES
+    purge_tile_cache $PATHS
 fi


scripts/aws/setupdb.sh
```diff
@@ -131,4 +131,7 @@ if [ "$load_water_quality" = "true" ] ; then
     FILES=("nhd_water_quality.sql.gz" "drb_catchment_water_quality.sql.gz")
     PATHS=("drb_catchment_water_quality_tn" "drb_catchment_water_quality_tp"
             "drb_catchment_water_quality_tss" "nhd_quality_tp" "nhd_quality_tn")
+
+    download_and_load $FILES
+    purge_tile_cache $PATHS
```
kdeloach: Does this function actually work for other people? And should this ever execute in the context of a dev env? I always need to comment it out.

kellyi: I think the `purge_tile_cache` function will always fail on local. We should update the script as a whole at some point to have it execute conditionally.

kdeloach: Thanks, that makes sense.

caseypt:
+1. Tested. Thanks.

scripts/aws/setupdb.sh
mmcfarland: The call fails on local, as designed, but doesn't affect the general function of the script.  Does it actually fail to load data for you?  I agree it could be cleaned up so as not to report that error.

hectcastro: The AWS CLI only gets installed when the production `requirements.txt` file is used. Might be able to do something like this to side-step the errors locally:

```bash
if which aws > /dev/null; then
    aws ...
fi
```

Issue #1516, opened 2016-10-06
Investigate time out issues


mmcfarland:
https://rollbar.com/Azavea/MMW/items/58/ and similar errors are coming in after the new release.

arottersman:
**Here's a recap of some digging we did into this issue:**

There's not much info on rollbar about what might be causing these errors. We thought we'd inspect the `core_job` table in the production db to see if there was any more relevant info about the timed out jobs. Rollbar provides the `uuid` of a job, but when we queried the table for each occurrence's `uuid` we found nothing- no jobs with those `uuid`s. At the point of failure, the `uuid` hasn't been set/saved to the `task_chain.id` which seems to be what rollbar associates with it. This is all to say that when you want to look up a job with this error from the database, you have to look it up like this:
```SELECT * FROM core_job WHERE error ~ '^Job <whatever id the rollbar message has in its title>';``` 

Currently though, the attributes of the timed out jobs aren't too helpful for figuring out why they timed out. It'd be helpful to see the the shapes that are causing these timeouts. For reference, this is all going down in `mmw/apps/modeling/views.py start_celery_job` which takes a param `job_input` - in this case the AOI. On success we save the `job_input` as `model_input`, but on error we don't save it all.  https://github.com/WikiWatershed/model-my-watershed/blob/2c18298688d857f33d43513f49e4909dce923a36/src/mmw/apps/modeling/views.py#L564

hectcastro:
This information may already be well-known, but my suspicion is that these jobs are being stopped by Celery after [enough retries occur](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/geoprocessing.py#L119-L130) to check the status of an SJS job.

Changes were made to the Spark configuration in the last release to give Spark more runway to complete a job that would have previously failed. I think we thought that guarding on a maximum AoI size would avoid timeouts like these from popping up, but perhaps our rough ratio of AoI size to timeout duration is off.

mmcfarland:
Yeah, I agree.  The goal is now to start capturing some information about specific AoIs that fail so we can tighten up that guard based on their attributes.  Thanks for the info.

PR #1536, opened 2016-10-12
Persist Input Of Failed Jobs

arottersman:
We have been getting these mysterious timed out analyze histograms jobs on staging and production: https://rollbar.com/Azavea/MMW/items/58/ . We tried to do some sleuthing on the jobs in the production db, and found that the most useful field for sleuthing `model_input` (the AOI) isn't persisted when a job fails. 

This PR makes sure that the `model_input` persists whether success or failure for all the analysis related jobs. The plan of attack after we merge this is to wait around until the error occurs again on staging, and then go back and inspect. There could be some too-big/bad shape making its way through.

#### Testing
Pull this branch and reload.
You can: `./scripts/debugcelery.sh`
Draw a shape and confirm the analyze jobs are fine. 
Now confirm the job fails when you run something like: 
```
http --form POST http://localhost:8000/api/modeling/start/analyze/ area_of_interest=%7B%22type%22%3A%22MultiPolygon%22%2C%22coordinates%22%3A%5B%5B%5B%5B-75.87982177734375%2C41.11246878918086%5D%2C%5B-75.98007202148438%2C41.05967965145965%5D%2C%5B-75.93887329101562%2C41.009957022453015%5D%2C%5B-75.78643798828125%2C41.05967965145965%5D%2C%5B-75.80703735351562%2C41.07313973329343%5D%2C%5B-75.81390380859375%2C41.08763212467916%5D%2C%5B-75.87982177734375%2C41.11246878918086%5D%5D%5D%5D%7
```
or any other malformed analyze request of your choosing. 

Now:
``` 
ssh vagrant services
psql -h localhost -U mmw

\o model_inputs.txt
SELECT model_input, error FROM core_job ORDER BY timestamp DESC LIMIT 2;
\q

cat model_inputs.txt
```
You should see something like the following, where the first row is the malformed request which now has a `model_input`, and the second row is the good request also still with a `model_input`:
```
                                                                                                                                                                                                      model_input                                                                                                                                                                                                      |              error              
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------------------
 %7B%22type%22%3A%22MultiPolygon%22%2C%22coordinates%22%3A%5B%5B%5B%5B-75.87982177734375%2C41.11246878918086%5D%2C%5B-75.98007202148438%2C41.05967965145965%5D%2C%5B-75.93887329101562%2C41.009957022453015%5D%2C%5B-75.78643798828125%2C41.05967965145965%5D%2C%5B-75.80703735351562%2C41.07313973329343%5D%2C%5B-75.81390380859375%2C41.08763212467916%5D%2C%5B-75.87982177734375%2C41.11246878918086%5D%5D%5D%5D%7D | No JSON object could be decoded
 {"type":"MultiPolygon","coordinates":[[[[-75.87982177734375,41.11246878918086],[-75.98007202148438,41.05967965145965],[-75.93887329101562,41.009957022453015],[-75.78643798828125,41.05967965145965],[-75.80703735351562,41.07313973329343],[-75.81390380859375,41.08763212467916],[-75.87982177734375,41.11246878918086]]]]}                                                                                         | 
(2 rows)
```
Connects #1516

### DIFF
--- a/src/mmw/apps/modeling/views.py
+++ b/src/mmw/apps/modeling/views.py
@@ -557,7 +557,8 @@ def start_celery_job(task_list, job_input, user=None,
     """
     created = now()
     job = Job.objects.create(created_at=created, result='', error='',
-                             traceback='', user=user, status='started')
+                             traceback='', user=user, status='started',
+                             model_input=job_input)
     routing_key = routing_key if routing_key else choose_worker()
     success = save_job_result.s(job.id, job_input).set(exchange=exchange,
                                                        routing_key=routing_key)


arottersman:
Hold up, I broke a test. Fixing it now.

src/mmw/apps/modeling/views.py
```diff
@@ -557,7 +557,8 @@ def start_celery_job(task_list, job_input, user=None,
     """
     created = now()
     job = Job.objects.create(created_at=created, result='', error='',
-                             traceback='', user=user, status='started')
+                             traceback='', user=user, status='started',
+                             model_input=job_input)
```
arottersman: Adding `model_input` on the job here so it's saved with this task:
```
error = save_job_error.s(job.id).set(exchange=exchange,
                                         routing_key=routing_key)
```

This introduces a little redundancy at our call for `save_job_result(job.id, job_input)`. I think it's better for now to make this change as targeted as possible, and not introduce the more wide-reaching change of removing the `model_input` param of `save_job_result`...we use the function a lot throughout the project, so when I had changed the parameters, it was breaking tests

mmcfarland:
Since the `uuid` isn't saved until after the failure, it doesn't seem like we'll have a way to link the Rollbar reported locals with anything to query the `job` table with.  Does it seem reasonable that the `created_at` attribute would suffice?

arottersman:
I've actually been using the `error` attribute to query. It's the same as the title of the error on rollbar

arottersman:
Thanks for taking a look!

PR #1543, opened 2016-10-14
Cast Water Quality `nord` from Decimal -> Int before serializing

kellyi:
This PR addresses a rollbar failure on staging whereby the water quality `nord` fails because it hasn't been cast to a JSON serializable type:

https://rollbar.com/Azavea/MMW/items/74/occurrences/17640657020/

The update casts it to an int before returning, and streamlines casting the other values to floats using a list comprehension over a segment of the array.

**Testing**
- get this branch, `vagrant up`, `./scripts/bundle.sh` then visit `localhost:8000` and open the browser console.
- select a congressional district within the DRB
- when the analysis results return, verify that the water quality data returns too and that there's data presented in the table and popup

### DIFF
--- a/src/mmw/apps/modeling/calcs.py
+++ b/src/mmw/apps/modeling/calcs.py
@@ -124,29 +124,10 @@ def catchment_water_quality(geojson):
                 # The TN, TP, and TSS values return as type Decimal,
                 # but we want floats.
                 dict(zip(columns,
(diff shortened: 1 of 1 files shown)


arottersman:
Tested- water quality analysis data still loads. +1

kellyi:
Thanks!

Issue #1489, opened 2016-09-27
Delete dangling Celery / SJS jobs


rajadain:
Currently, when processing a large area of interest (such as a HUC-10 or HUC-8 in MapShed), the front-end polls the endpoint 90 times (twice a second) before giving up. At this point, if the user wishes to continue, they must resubmit the job.

In the meantime, this old "dangling" job continues to be processed by Celery / SJS in the back-end. This clogs up the pipeline for other jobs, even the ones that the user may immediately re-submit.

Find a way to cancel and delete Celery and SJS jobs once they have reached a similar timeout. Consider defining this timeout value in standard configuration so it can be easily shared between components.

mmcfarland:
Closing as we're not doing SJS calls anymore.  This could conceivably still be appropriate for just the Celery jobs based on our new setup, but lets re-evaluate when the refactored geoprocessing pipeline is stable, and open a new issue if necessary.

PR #1544, opened 2016-10-17
Feature/delete jobs

jtarricone:
This feature introduces an avenue to attempt to stop running jobs, and also modifies the app to send requests for termination where appropriate and sensible to do so.

The change to the API involves adding an endpoint at `.../jobs/[job id]/kill/` that accepts requests and attempts to stop running jobs; the celery settings have also been updated to add default timeouts. The change to the app places requests to the `kill` endpoint where it made sense to do so, and also tags expiration/timeout values onto job submission requests.

To test:
- clone this branch and bring up the environment
- with the browser console, network monitor, and `debugcelery.sh` and `debugserver.sh` output watchable: 
-  navigate to `http://localhost:8000` and go through standard user actions
-  go through user actions but back the app out of them at the earliest opportunity
-  for good measure, intersperse the two, such that you back out of an action but then proceed to let it run to completion
- check that requests are being made to the `kill` endpoint where it makes sense to do so, such as when you back out of a project action, and check if the celery output indicates that it is revoking or actively terminating jobs

Connects #1489

### DIFF
--- a/src/mmw/apps/modeling/mapshed/tasks.py
+++ b/src/mmw/apps/modeling/mapshed/tasks.py
@@ -3,10 +3,11 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import absolute_import
 
-from ast import literal_eval as make_tuple
+import requests
 
+from ast import literal_eval as make_tuple
 from celery import shared_task
-from celery.exceptions import Retry
+from celery.exceptions import Retry, SoftTimeLimitExceeded
 
 from django.conf import settings
 from django.contrib.gis.geos import GEOSGeometry
@@ -81,6 +82,27 @@ def mapshed_finish(self, incoming):
 
     try:
         return sjs_retrieve(retry=self.retry, **incoming)
+    except SoftTimeLimitExceeded:
+        sjs_job_id = self.request.args[0]['job_id']
+        url = 'http://{}:{}/jobs/{}'.format(settings.GEOP['host'],
+                                            settings.GEOP['port'],
+                                            sjs_job_id)
+
+        response = requests.get(url)
+        if response.ok:
+            sjs_result = response.json()
+        else:
+            raise Exception('Unable to retrieve job {} from Spark JobServer.\n'
+                            'Details = {}'.format(sjs_job_id))
+
+        if sjs_result['status'] == 'RUNNING':
+            delete = requests.delete(url)
+            if delete.ok:
+                return {'message':
+                        'Job {0} successfully terminated.'.format(sjs_job_id)}
+            else:
+                raise Exception('Job {} timed out, unable to delete.\n'
+                                'Details: {}'.format(sjs_job_id, delete.text))
     except Retry as r:
         # Celery throws a Retry exception when self.retry is called to stop
         # the execution of any further code, and to indicate to the worker
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -11,6 +11,7 @@ from math import sqrt
 from StringIO import StringIO
 
 from celery import shared_task
+from celery.exceptions import SoftTimeLimitExceeded
 
 from apps.modeling.geoprocessing import histogram_start, histogram_finish, \
     data_to_survey, data_to_census, data_to_censuses
@@ -21,6 +22,8 @@ from apps.modeling.calcs import (animal_population,
 
 from tr55.model import simulate_day
 from gwlfe import gwlfe, parser
+
+from django.conf import settings
 
 logger = logging.getLogger(__name__)
 
@@ -99,13 +102,34 @@ def get_histogram_job_results(self, incoming):
     """ Calls a function that polls SJS for the results
     of the given Job. Self here is Celery.
     """
-    pixel_width = incoming['pixel_width']
-    histogram = histogram_finish(incoming['sjs_job_id'], self.retry)
+    try:
+        pixel_width = incoming['pixel_width']
+        histogram = histogram_finish(incoming['sjs_job_id'], self.retry)
+        return {
+            'pixel_width': pixel_width,
+            'histogram': histogram
+        }
+    except SoftTimeLimitExceeded:
+        sjs_job_id = self.request.args[0]['sjs_job_id']
+        url = 'http://{}:{}/jobs/{}'.format(settings.GEOP['host'],
+                                            settings.GEOP['port'],
+                                            sjs_job_id)
 
-    return {
-        'pixel_width': pixel_width,
-        'histogram': histogram
-    }
+        response = requests.get(url)
+        if response.ok:
+            sjs_result = response.json()
+        else:
+            raise Exception('Unable to retrieve job {} from Spark JobServer.\n'
+                            'Details = {}'.format(sjs_job_id))
+
+        if sjs_result['status'] == 'RUNNING':
+            delete = requests.delete(url)
+            if delete.ok:
+                return {'message':
+                        'Job {0} successfully terminated.'.format(sjs_job_id)}
+            else:
+                raise Exception('Job {} timed out, unable to delete.\n'
+                                'Details: {}'.format(sjs_job_id, delete.text))
 
 
 @shared_task
--- a/src/mmw/apps/modeling/views.py
+++ b/src/mmw/apps/modeling/views.py
@@ -567,6 +567,25 @@ def drb_point_sources(request):
                     headers={'Cache-Control': 'max-age: 604800'})
 
 
+@decorators.api_view(['GET'])
+@decorators.permission_classes((AllowAny, ))
+def kill_job(request, job_uuid, format=None):
+    try:
+
+        celery.current_app.control.revoke(job_uuid,
+                                          terminate=True,
+                                          signal='SIGUSR1')
+        return Response(
+            'Job {0} was successfully terminated.'.format(
+                job_uuid))
+    except Exception as ex:
+        response_data = {
+            'errors': ['Job {0} could not be killed.'.format(job_uuid),
+                       ex.message]}
+        return Response(data=response_data,
+                        status=status.HTTP_500_INTERNAL_SERVER_ERROR)
+
+
 def start_celery_job(task_list, job_input, user=None,
                      exchange=MAGIC_EXCHANGE, routing_key=None):
     """
--- a/src/mmw/mmw/settings/base.py
+++ b/src/mmw/mmw/settings/base.py
@@ -130,6 +130,9 @@ CELERY_WORKER_DIRECT = True
 CELERY_CREATE_MISSING_QUEUES = True
 CELERY_CHORD_PROPAGATES = True
 CELERY_CHORD_UNLOCK_MAX_RETRIES = 60
+# the time limit won't affect the polling tasks since they are short-running
+# the value of 42 coincides with the implied time limit on the polling tasks
+CELERYD_TASK_TIME_LIMIT = 42
 # CELERYD_CONCURRENCY = 2
 # END CELERY CONFIGURATION
 
--- a/src/mmw/apps/modeling/urls.py
+++ b/src/mmw/apps/modeling/urls.py
@@ -13,7 +13,7 @@ from apps.modeling import views
 # Third set of characters must start with a '4'.
 # Fourth set of characters must start with one of 'a,b,8,9'.
 uuid_regex = '(?P<job_uuid>[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-4[0-9A-Fa-f]{3}-' \
-             + '[89abAB][0-9A-Fa-f]{3}-[0-9A-Fa-f]{12})/$'
+             + '[89abAB][0-9A-Fa-f]{3}-[0-9A-Fa-f]{12})'
 
 urlpatterns = patterns(
     '',
@@ -30,7 +30,8 @@ urlpatterns = patterns(
         views.start_analyze_catchment_water_quality,
         name='start_analyze_catchment_water_quality'),
     url(r'start/mapshed/$', views.start_mapshed, name='start_mapshed'),
-    url(r'jobs/' + uuid_regex, views.get_job, name='get_job'),
+    url(r'jobs/' + uuid_regex + r'/$', views.get_job, name='get_job'),
+    url(r'jobs/' + uuid_regex + r'/kill/$', views.kill_job, name='kill_job'),
     url(r'start/tr55/$', views.start_tr55, name='start_tr55'),
     url(r'start/rwd/$', views.start_rwd, name='start_rwd'),
     url(r'start/gwlfe/$', views.start_gwlfe, name='start_gwlfe'),


kellyi:
Taking a look at this.

src/mmw/apps/modeling/geoprocessing.py
```diff
@@ -11,6 +11,8 @@
 import json
 import re
 
+import time
```
kellyi: Consider dropping the whitespace above this import to place it in a block with the imports above.

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -37,6 +37,7 @@
                                          sed_a_factor
                                          )
 
+import time
```
kellyi: Probably worth moving this import toward the top of the file, above the imports from the application.

src/mmw/apps/modeling/tasks.py
```diff
@@ -22,6 +22,8 @@
 from tr55.model import simulate_day
 from gwlfe import gwlfe, parser
 
+import time
```
kellyi: Consider moving this import to the top of the file, too.

src/mmw/apps/modeling/views.py
```diff
@@ -239,10 +239,15 @@ def start_mapshed(request, format=None):
     user = request.user if request.user.is_authenticated() else None
     created = now()
     mapshed_input = json.loads(request.POST['mapshed_input'])
+
+    # the client app convention is measuring time in ms, conver to s here
```
kellyi: I think `conver` should be `convert` here.

src/mmw/.jshintrc
```diff
@@ -6,6 +6,7 @@
     "browser": true,
     "curly": true,
     "eqeqeq": true,
+    "esnext": true,
```
kellyi: I'm not 100% sure we'd want to update this project to use ES6 syntax because I think some bits of it still may have to be transpiled to work across all browsers, and I think everything else here is ES5.

kellyi:
What's a sure way to get the `/kill` endpoint called via the frontend? I've tried starting MapShed jobs, then moving back to draw and selecting a different AOI, but I don't think I've seen the `/kill` endpoint hit yet.

jtarricone:
>What's a sure way to get the /kill endpoint called via the frontend?

The sequence you described should send a kill request. There should be a pause if you back out of an action immediately because the app has to block until getting a job ID back, but you should still see a request go through after a few seconds. There aren't any requests being sent at all?

src/mmw/.jshintrc
jtarricone: Ugh, I was a little wary about it but I hoped it wouldn't be a problem. There have been multiple time with this app where just a basic `let` would have made something easier and/or more readable.

Is there a list of browsers this app is required to support? I checked the compatibility of ES6/`let` and it looked reasonable at first glance.

ajrobbins:
@jtarricone 
>Is there a list of browsers this app is required to support?

Latest versions of Safari and IE, two most recent major versions of Firefox and Chrome. Edge in the future.

jtarricone:
In that case `let` looks okay, but @kellyi has a point about flagging JSHint to be quiet about ES6 in general. I'll check if specific ES6 features can be ignored rather than the whole thing, unless there's another reason to excise the whole thing.

kellyi:
Could most of the uses of `let` get replaced with `var` here? I think `var`'s now idiomatic to the project, and it seems like the uses of `let` in this patch behave the same way a `var` assignment would. 

If we're just trying to restrict the scoping, consider adjusting the variable names in those assignments to something longer + more unique to avoid collisions.

jtarricone:
>Could most of the uses of let get replaced with var here?

Sure, but having single-use `var`'s or conditionally-bound `var`'s floating around has made sections of this app difficult to follow without watching it in a debugger.

src/mmw/.jshintrc
jtarricone: Why?

jtarricone:
I pushed a fixup to move the `import`s around and correct the spelling error. @kellyi I'm still curious whether you see any `kill` requests either client- or server-side.

kellyi:
Looking at this again now and it looks like the `/kill` endpoint is being called when I click the back button while a MapShed job's running.

kellyi:
I haven't been able to run and complete a MapShed job successfully, however. It looks like it typically gets hung up on `ncld_soils` or `nlcd_kfactor` and then times out on the frontend. It also looks like the frontend sends a kill signal when it times out.

kellyi:
Seems like MapShed keeps failing on my local on `develop`, too. It reaches the `nlcd_soils` or `nlcd_kfactor` bits and then it hangs temporarily, long enough for the frontend to show the "Operation took too long" message.

`debugcelery.sh` shows that the job eventually does complete, but it's after the frontend has stopped polling for results. Is there a change I should make to my local worker VM to help it return the results more quickly?

jtarricone:
>debugcelery.sh shows that the job eventually does complete, but it's after the frontend has stopped polling for results.

Does the app send a `kill` request? If so, does anything happen in celery?

kellyi:
That was on `develop`, so no kill req. On this branch it does dispatch a kill request when the op times out.

jtarricone:
>Is there a change I should make to my local worker VM to help it return the results more quickly?

I changed my vagrantfile to provision 4 CPUs and 2048gb of memory. That may be excessive but I guess just generally try to make the VM beefier?

kellyi:
Ok, going to reprovision. I think we should consider committing some worker VM changes to the Vagrantfile if they're required to enable MapShed to run.

jtarricone:
>That was on develop, so no kill req. On this branch it does dispatch a kill request when the op times out.

Sorry, I forgot you had mentioned it wasn't on this branch. That's the desired behavior though, so at least that result is somewhat promising.

kellyi:
When a job times out, it looks like the frontend hits `/kill` 2x:

screen shot 2016-10-19 at 2 20 49 pm

The first one goes to the endpoint with the job id, and shows a "Job xyz successfully terminated." message.

The second one tries to hit http://localhost:8000/api/modeling/jobs/null/kill/ and returns a 404.

kellyi:
Hm, I upped the worker VM RAM to 4096 and gave it 2 CPUs and it's still hanging on `nlcd_soils`, then timing out.

kellyi:
When the `/kill` endpoint's hit, Celery reports that it's `discarding revoked task` and halts after the next step in the task returns:

```sh
[2016-10-19 14:32:15,430: INFO/MainProcess] Discarding revoked task: apps.core.tasks.save_job_result[468be829-a55c-4336-9bca-dd533cec882f]
[2016-10-19 14:32:15,443: INFO/MainProcess] Task apps.modeling.mapshed.tasks.collect_data[b17fe767-bebc-4f6e-b974-e0606cb3a51a] succeeded in 0.31474007s: {'RetentNLake': 0.12, u'MaxWaterCap': 0.0, u'n41k': 0.0, 'n38': 0, 'n39': 0, 'n30': 0, 'n31': 0, 'n32': 0, 'n33': 0, 'n34':...
```

kellyi:
Another run, from a job which timed out:

```sh
[2016-10-19 14:37:47,222: INFO/MainProcess] Task apps.modeling.mapshed.tasks.nlcd_soils[d903497b-6db7-4996-8bb7-4403e7f95f70] succeeded in 0.00383781800008s: {u'landuse_pcts': [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0.07726864330637916, 0.4474393530997305, 0.4752920035938904, 0, 0, 0],...
[2016-10-19 14:38:29,778: DEBUG/MainProcess] pidbox received method revoke(signal=u'SIGTERM', terminate=True, task_id=u'a87b8d7a-05f8-4a4c-990a-5d6e9fc8664d') [reply_to:None ticket:None]
```

mmcfarland:
Generally speaking, this looks like `kill` requests are being sent at appropriate places.  I was monitoring a `debugcelery` output while these requests were received and reported successfully in the console.  The jobs continued to process, but did end at `Discarding revoked task: apps.core.tasks.save_job_result[7dee9323-1084-43e2-b800-51c87d6c9994]`

I'm wondering what the kill is killing, just the one job that saves the results leaving the rest or is there a time disconnect before the jobs are actually killed?  Additionally, the connected issue refers to killing both celery and SJS jobs - I see an expiration passed into SJS jobs, but it doesn't look like this is killed with the celery job.  Is that the case and if so, is it possible to kill a job in SJS, my understanding would be that that  is the resource which we're really trying to free up by these "dangling" jobs.

jtarricone:
I pushed a couple fixups that I think address most points, plus I fixed a bug that put the UI in a bad state if a user moved directly from selecting an AoI to one of the model options (i.e. didn't wait for "Analyze" to return).

jtarricone:
> Additionally, the connected issue refers to killing both celery and SJS jobs - I see an expiration passed into SJS jobs, but it doesn't look like this is killed with the celery job. Is that the case and if so, is it possible to kill a job in SJS?

It is possible, however it would be pretty gnarly because I think the UUID that a spark job is known by is never known outside the celery chain. If true, the client wouldn't know which job to kill. I imagine it's feasible to kill a spark job within the celery chain but it wasn't immediately obvious how to do so. Having a settable expiration seemed to be a reasonable approach until more direct manipulation can be implemented.

mmcfarland:
> It is possible, however it would be pretty gnarly because I think the UUID that a spark job is known by is never known outside the celery chain.

Got it.  Can you peel off that into a new issue so we don't lose track of it?

kellyi:
Trying this out and it looks like the frontend is still calling `/kill` 2x, although now the second one is pointing at the correct URL so doesn't result in a 404. Is the duplication required here?

jtarricone:
>I can get the application in an error state with the following workflow:
>    Draw an area, let analyze succeed
>    Select Site Storm Model, but use the in-app back button to go back before the model completes
>    A single kill request is sent, but there is no confirmation of success in the console
>    See an error in the console: main.js:15417 Uncaught TypeError: Cannot read property 'show' of undefined
>    Hitting Draw to go back again one more, or otherwise using the app seems to break the UI

I think that should now be fixed, but I didn't observe an error using that sequence; I only saw the UI break if I immediately went to TR55/GWLFE before letting the analysis call return, and then kept it on the model view until there was a timeout.

kellyi:
Tried out the instructions for breaking the UI described above and the UI still works, but when I run MapShed, then back out of it and click Draw to return to the main map, I see these two errors, which I think are caused by the error message trying to populate an element which doesn't exist:

```
main.js:9223 Uncaught TypeError: Cannot read property 'set' of null
```

&&

```
main.js:15409 Uncaught TypeError: Cannot read property 'show' of undefined
```

mmcfarland:
For future reference, it would be desirable to have that unrelated bug fixed either in its own PR or at least its own commit so it wouldn't be so coupled to this feature.  Especially with a bug fix, it reduces our flexibility of choosing what goes in a release or a bug fix release.

jtarricone:
>Is the duplication required here?

I'm having trouble replicating that. When do you see multiple requests?

jtarricone:
I can back out the bug fix, although the fix and this feature somewhat overlap. Additionally, the primary issue still exists, which is that project state is being mutated as the result of jQuery calls on objects that are no longer bound.

kellyi:
re: duplication:

1. Select an area to analyze. While the analysis is going on, click to start a MapShed job. 
2. Click the back button to back out of the MapShed job, but stay on the analyze tab
3. It looks like it's calling `/kill` 2x:

screen shot 2016-10-20 at 1 42 24 pm

Additionally, it also seems like killing a MapShed job like this will cause the analysis part to get hung up, so that if you stay on the screen with the analyze tab the raster tasks seem never to complete, and `debugcelery.sh` doesn't show any new actions.

mmcfarland:
Yeah, it would be ideal to pop it out since we can't do a release without the bugfix, but I'd like to be able to have this kill stuff in develop for a while (and skip this release) since it hits a bunch of important features, so we have time to watch and catch side effects.

jtarricone:
>Additionally, it also seems like killing a MapShed job like this will cause the analysis part to get hung up, so that if you stay on the screen with the analyze tab the raster tasks seem never to complete, and debugcelery.sh doesn't show any new actions.

That does seem to happen; I think it might be prudent to just remove most of the front-end changes here, since obviously there's some wonky state business happening that will probably take some time to address. We can leave the server-side changes, which will kill long-running tasks.

jtarricone:
I pushed a combination of excisions and refactors to the previous code. All the front-end code has been removed, while the server-side code is essentially the same but a little leaner and more idiomatic with respect to the use of celery. The `/job/{id}/kill/` endpoint remains because it is the only way to terminate polling jobs and chord unlocks, but until it gets wired up to the UI or some other mechanism it will require direct invocation.

src/mmw/apps/modeling/views.py
```diff
@@ -567,6 +567,25 @@ def drb_point_sources(request):
...
 
 
+@decorators.api_view(['GET'])
+@decorators.permission_classes((AllowAny, ))
+def kill_job(request, job_uuid, format=None):
+    try:
+
+        celery.current_app.control.revoke(job_uuid,
+                                          terminate=True,
+                                          signal='SIGUSR1')
```
jtarricone: The reason to send the `SIGUSR1` signal instead of `SIGTERM` or `SIGKILL` is that celery will raise a `SoftTimeLimitExceeded` for the former, whereas `SIGTERM` and `SIGKILL` both appear to immediately kill the process. I would have expected celery to raise a `HardTimeLimitExceeded` exception upon a `SIGTERM` signal but it doesn't appear to be the case.

The reason for use of `SIGUSR1` is evident through reading the docs, but I can add a comment here if it would be useful.

kellyi:
Taking another look at this.

src/mmw/mmw/settings/base.py
```diff
@@ -130,6 +130,9 @@ def get_env_setting(setting):
 CELERY_CREATE_MISSING_QUEUES = True
 CELERY_CHORD_PROPAGATES = True
 CELERY_CHORD_UNLOCK_MAX_RETRIES = 60
+# the time limit won't affect the polling tasks since they are short-running
+# the value of 42 coincides with the implied time limit on the polling tasks
+CELERYD_TASK_TIME_LIMIT = 42
```
kellyi: Q: is the implied time limit actually specified somewhere on the client side? Or is it implied elsewhere?

kellyi:
Tried running a MapShed job in the frontend, then took the URI listed in the console to which the client was polling and sent a request using `httpie` to 

```
http localhost:8000/api/modeling/jobs/fd5cb76a-6b43-4266-8dd8-3f85fa385467/kill/
```

I can see some task revoked errors in the `debugcelery` output, but MapShed seems to keep running which makes me wonder whether it's not adequately killing the entire chain:

```
[2016-10-27 14:06:53,418: INFO/MainProcess] Received task: celery.chord_unlock[493f8786-7383-46d5-8270-e21fc67dcad8] eta:[2016-10-27 14:06:54.291020-04:00]
[2016-10-27 14:06:53,424: ERROR/MainProcess] Task apps.modeling.mapshed.tasks.mapshed_finish[db3365ea-6e57-48f8-a2c9-bd9700f106e7] raised unexpected: TaskRevokedError(u'Job derp successfully terminated.',)
  File "/opt/app/apps/modeling/mapshed/tasks.py", line 83, in mapshed_finish
    raise TaskRevokedError('Job {0} successfully terminated.'.format('derp'))
TaskRevokedError: Job derp successfully terminated.
[2016-10-27 14:06:53,434: INFO/MainProcess] Received task: apps.modeling.mapshed.tasks.mapshed_start[6f58a68f-6166-496c-a221-578c0a33eac8]
[2016-10-27 14:06:53,545: INFO/MainProcess] Received task: apps.modeling.mapshed.tasks.mapshed_start[380b603b-0a89-4483-96fe-a8056c8d4b1e]
[2016-10-27 14:06:53,553: DEBUG/MainProcess] Task accepted: apps.modeling.mapshed.tasks.mapshed_start[6f58a68f-6166-496c-a221-578c0a33eac8] pid:3120
[2016-10-27 14:06:53,591: INFO/MainProcess] Task apps.modeling.mapshed.tasks.mapshed_start[6f58a68f-6166-496c-a221-578c0a33eac8] succeeded in 0.151281227s: {u'host': 'localhost', u'port': '8090', u'job_id': u'8190e74e-740d-4e3e-a408-a561bfcad6c6'}
[2016-10-27 14:06:53,636: INFO/MainProcess] Received task: apps.modeling.mapshed.tasks.mapshed_start[e14ff6bf-b7ab-49a4-bad4-b8823f7e9fda]
[2016-10-27 14:06:53,734: DEBUG/MainProcess] Task accepted: apps.modeling.mapshed.tasks.mapshed_start[e36118e2-a781-4b65-b2b2-553ba7271ac7] pid:3121
[2016-10-27 14:06:53,782: INFO/MainProcess] Received task: apps.modeling.mapshed.tasks.mapshed_finish[37cc84a8-158a-4e6d-af0a-aa072e24a9cb]
[2016-10-27 14:06:53,810: INFO/MainProcess] Received task: apps.modeling.mapshed.tasks.mapshed_finish[667d4f15-d234-4a90-9049-ae03dcb0b8ef]
[2016-10-27 14:06:53,825: INFO/MainProcess] Received task: apps.modeling.mapshed.tasks.mapshed_finish[248a67ff-dcc6-49dd-a1df-f8579be11fb4]
[2016-10-27 14:06:53,842: INFO/MainProcess] Task apps.modeling.mapshed.tasks.mapshed_start[e36118e2-a781-4b65-b2b2-553ba7271ac7] succeeded in 0.281826519s: {u'host': 'localhost', u'port': '8090', u'job_id': u'd468dd83-d7c1-445e-b1b8-5768571543bd'}
[2016-10-27 14:06:53,871: INFO/MainProcess] Received task: apps.modeling.mapshed.tasks.mapshed_finish[2117e6be-5fea-419f-8f76-8c49d3e5e3bb]
[2016-10-27 14:06:53,908: DEBUG/MainProcess] Task accepted: apps.modeling.mapshed.tasks.mapshed_start[380b603b-0a89-4483-96fe-a8056c8d4b1e] pid:3121
[2016-10-27 14:06:54,066: DEBUG/MainProcess] Task accepted: apps.modeling.mapshed.tasks.mapshed_start[e14ff6bf-b7ab-49a4-bad4-b8823f7e9fda] pid:3120
[2016-10-27 14:06:54,176: INFO/MainProcess] Received task: apps.modeling.mapshed.tasks.mapshed_finish[0405f6ee-6673-490e-bde5-6e3d11d37a44]
[2016-10-27 14:06:54,191: INFO/MainProcess] Task apps.modeling.mapshed.tasks.mapshed_start[380b603b-0a89-4483-96fe-a8056c8d4b1e] succeeded in 0.362248463s: {u'host': 'localhost', u'port': '8090', u'job_id': u'193dd9bd-e74b-4a95-81a3-ee0951655667'}
[2016-10-27 14:06:54,206: DEBUG/MainProcess] Task accepted: apps.modeling.mapshed.tasks.mapshed_finish[37cc84a8-158a-4e6d-af0a-aa072e24a9cb] pid:3121
[2016-10-27 14:06:54,230: ERROR/MainProcess] Task apps.modeling.mapshed.tasks.mapshed_finish[37cc84a8-158a-4e6d-af0a-aa072e24a9cb] raised unexpected: TaskRevokedError(u'Job derp successfully terminated.',)
  File "/opt/app/apps/modeling/mapshed/tasks.py", line 83, in mapshed_finish
    raise TaskRevokedError('Job {0} successfully terminated.'.format('derp'))
TaskRevokedError: Job derp successfully terminated.
[2016-10-27 14:06:54,379: INFO/MainProcess] Received task: apps.modeling.mapshed.tasks.mapshed_finish[9fe1ab44-d7bc-4b81-b6af-d9bb579a5637]
[2016-10-27 14:06:54,398: INFO/MainProcess] Task apps.modeling.mapshed.tasks.mapshed_start[e14ff6bf-b7ab-49a4-bad4-b8823f7e9fda] succeeded in 0.329473853s: {u'host': 'localhost', u'port': '8090', u'job_id': u'a45a28da-5655-4f5f-92f8-73058ab7e396'}
[2016-10-27 14:06:54,413: DEBUG/MainProcess] Task accepted: apps.modeling.mapshed.tasks.mapshed_finish[667d4f15-d234-4a90-9049-ae03dcb0b8ef] pid:3120
[2016-10-27 14:06:54,422: ERROR/MainProcess] Task apps.modeling.mapshed.tasks.mapshed_finish[667d4f15-d234-4a90-9049-ae03dcb0b8ef] raised unexpected: TaskRevokedError(u'Job derp successfully terminated.',)
  File "/opt/app/apps/modeling/mapshed/tasks.py", line 83, in mapshed_finish
    raise TaskRevokedError('Job {0} successfully terminated.'.format('derp'))
TaskRevokedError: Job derp successfully terminated.
[2016-10-27 14:06:54,451: DEBUG/MainProcess] Task accepted: apps.modeling.mapshed.tasks.mapshed_finish[248a67ff-dcc6-49dd-a1df-f8579be11fb4] pid:3121
[2016-10-27 14:06:54,474: ERROR/MainProcess] Task apps.modeling.mapshed.tasks.mapshed_finish[248a67ff-dcc6-49dd-a1df-f8579be11fb4] raised unexpected: TaskRevokedError(u'Job derp successfully terminated.',)
  File "/opt/app/apps/modeling/mapshed/tasks.py", line 83, in mapshed_finish
    raise TaskRevokedError('Job {0} successfully terminated.'.format('derp'))
TaskRevokedError: Job derp successfully terminated.
[2016-10-27 14:06:54,529: DEBUG/MainProcess] Task accepted: apps.modeling.mapshed.tasks.mapshed_finish[2117e6be-5fea-419f-8f76-8c49d3e5e3bb] pid:3120
[2016-10-27 14:06:54,542: ERROR/MainProcess] Task apps.modeling.mapshed.tasks.mapshed_finish[2117e6be-5fea-419f-8f76-8c49d3e5e3bb] raised unexpected: TaskRevokedError(u'Job derp successfully terminated.',)
  File "/opt/app/apps/modeling/mapshed/tasks.py", line 83, in mapshed_finish
    raise TaskRevokedError('Job {0} successfully terminated.'.format('derp'))
TaskRevokedError: Job derp successfully terminated.
[2016-10-27 14:06:54,561: DEBUG/MainProcess] Task accepted: apps.modeling.mapshed.tasks.mapshed_finish[0405f6ee-6673-490e-bde5-6e3d11d37a44] pid:3121
[2016-10-27 14:06:54,564: ERROR/MainProcess] Task apps.modeling.mapshed.tasks.mapshed_finish[0405f6ee-6673-490e-bde5-6e3d11d37a44] raised unexpected: TaskRevokedError(u'Job derp successfully terminated.',)
  File "/opt/app/apps/modeling/mapshed/tasks.py", line 83, in mapshed_finish
    raise TaskRevokedError('Job {0} successfully terminated.'.format('derp'))
TaskRevokedError: Job derp successfully terminated.
[2016-10-27 14:06:54,650: DEBUG/MainProcess] Task accepted: apps.modeling.mapshed.tasks.mapshed_finish[9fe1ab44-d7bc-4b81-b6af-d9bb579a5637] pid:3121
[2016-10-27 14:06:54,662: DEBUG/MainProcess] Task accepted: celery.chord_unlock[493f8786-7383-46d5-8270-e21fc67dcad8] pid:3120
[2016-10-27 14:06:54,664: ERROR/MainProcess] Task apps.modeling.mapshed.tasks.mapshed_finish[9fe1ab44-d7bc-4b81-b6af-d9bb579a5637] raised unexpected: TaskRevokedError(u'Job derp successfully terminated.',)
  File "/opt/app/apps/modeling/mapshed/tasks.py", line 83, in mapshed_finish
    raise TaskRevokedError('Job {0} successfully terminated.'.format('derp'))
TaskRevokedError: Job derp successfully terminated.
[2016-10-27 14:06:54,682: INFO/MainProcess] Received task: celery.chord_unlock[493f8786-7383-46d5-8270-e21fc67dcad8] eta:[2016-10-27 14:06:55.572692-04:00]
[2016-10-27 14:06:54,685: INFO/MainProcess] Task celery.chord_unlock[493f8786-7383-46d5-8270-e21fc67dcad8] retry: Retry in 1s
[2016-10-27 14:06:55,801: DEBUG/MainProcess] Task accepted: celery.chord_unlock[493f8786-7383-46d5-8270-e21fc67dcad8] pid:3120
[2016-10-27 14:06:55,811: INFO/MainProcess] Received task: celery.chord_unlock[493f8786-7383-46d5-8270-e21fc67dcad8] eta:[2016-10-27 14:06:56.780918-04:00]
[2016-10-27 14:06:55,816: INFO/MainProcess] Task celery.chord_unlock[493f8786-7383-46d5-8270-e21fc67dcad8] retry: Retry in 1s
[2016-10-27 14:06:57,155: DEBUG/MainProcess] Task accepted: celery.chord_unlock[493f8786-7383-46d5-8270-e21fc67dcad8] pid:3121
```

(& then MapShed tasks continue).

kellyi:
It seems like the MapShed jobs finally stop, btw, when they hit a `MaxRetriesExceeded` error.

jtarricone:
I had to push a fixup, that `TaskRevokedError` was something I added for debugging so I wouldn't have to keep trying to catch the job ID in time to send a kill request.

jtarricone:
Testing this in "human" time is pretty tricky and honestly quite tedious. I think it might be a better way to go to write a script to do it. I think that it works as intended, but it's hard to determine that with total certainty because of the nature of the problem.

kdeloach:
Is this still in progress?

kdeloach:
No activity on this PR in 3 months. Let's consider closing this for now. We can keep the branch alive.

rajadain:
I agree, there doesn't seem to be a burning need for this right now.

Issue #1500, opened 2016-09-29
Touchscreen issues with ACER laptop


ajrobbins:
Reported by Stroud:
>Just got a couple videos sent to me by Nanettte. She was working with two teachers this morning in Kennett Square, PA (DRB). Their school has ACER laptops with touchscreens and Chrome as the browser. They were using a mouse but still had the following issues.
 
1.      Could not click on Overlays or Observations in the Layers menu. Could not enable any of these.

2.      Delineate watershed does not work.

ajrobbins:
More details - teachers were running Windows and using the production site. Some of these issues may be fixed with https://github.com/WikiWatershed/model-my-watershed/pull/1454

ajrobbins:
I just tested the production site on our ASUS windows touch test device (testing using Chrome), and I couldn't click on anything in the layer selector (including other tabs). I can double tap on the layer selector and zoom into the basemap underneath as well as drag to pan, so I think the layer selector is not registering touch events correctly and they're being captured by the basemap beneath.

ajrobbins:
I can't reproduce the RWD issue.

arottersman:
What we found with #1454 was that even if you're clicking, chrome will fire both touch and mouse events when its running on a touch/mouse machine like new windows laptops. 

Also, another data point, on my android phone running chrome, the same thing happens - clicking the tabs on the layer selector pans/zooms the maps instead of selecting a new tab. Maybe we should be handling touch events explicitly here https://github.com/WikiWatershed/model-my-watershed/blob/a464bd213a92e3d04080e0af26b0af65e47d4a16/src/mmw/js/src/core/layerControl.js#L285 ?

PR #1548, opened 2016-10-19
Fix Layer Controls On Narrow Screens and Touch+Mouse Devices

arottersman:
Connects #1500

We weren't able to use the layer controls on Windows devices that have both touch and mouse input, or on Android tablets/phones in portrait mode. 

- For the Windows devices the issue was that calling `L.Browser.touch` in `core/layerControl.js` returns true in Chrome for devices that have touch capability (even if you're using the mouse), and we were checking for truthiness so that we could call: 
```
 L.DomEvent.on(container, 'click', L.DomEvent.stopPropagation);
 L.DomEvent.on(listContainer, 'click', L.DomEvent.stopPropagation);
```
The code now more closely matches watch the leaflet parent control library does (disabling scroll propagation if a device has touch input.) 

- For the android tablets and phones in portrait mode, the problem was just that the draw toolbar div was covering the leaflet controls, and taking all the clicks. 

### Testing
Pull this branch, and `./scripts/bundle.sh`
- Confirm that you can still use the layer controls/draw toolbar on your desktop browser
Then, accessing your build at `<your computer's name>.local:8000`
- Take our test ASUS laptop/tablet and verify you can use the draw toolbar, the layer controls, and the geolocator. 
- You should be able to close and reopen the layer controls, switch tabs, etc (switching tabs is kind of slow)

- Take our test Nexus tablet or a suite of devices in Chrome developer tools and go through the same steps as above in portrait and landscape mode

### DIFF
--- a/src/mmw/js/src/core/layerControl.js
+++ b/src/mmw/js/src/core/layerControl.js
@@ -69,18 +69,12 @@ module.exports = L.Control.Layers.extend({
         // from firing a mouseout event when the touch is released
         container.setAttribute('aria-haspopup', true);
 
-        // Copied directly from the parent class.
+        // Copied directly from the parent class, with addition of listContainer
+        L.DomEvent.disableClickPropagation(container);
+        L.DomEvent.disableClickPropagation(listContainer);
         if (!L.Browser.touch) {
-            L.DomEvent
-                .disableClickPropagation(container)
-                .disableScrollPropagation(container);
-
-            L.DomEvent
-                .disableClickPropagation(listContainer)
-                .disableScrollPropagation(listContainer);
-        } else {
-            L.DomEvent.on(container, 'click', L.DomEvent.stopPropagation);
-            L.DomEvent.on(listContainer, 'click', L.DomEvent.stopPropagation);
+            L.DomEvent.disableScrollPropagation(container);
+            L.DomEvent.disableScrollPropagation(listContainer);
         }
 
         // Expand the layer control so that Leaflet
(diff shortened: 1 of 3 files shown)


ajrobbins:
Can y'all test this on our tester iPad and on an iPad mini simulator, if possible?

kellyi:
Layout generally looks good. On a Nexus 7 width device on portrait mode, it looks like the "Reset" button does cover the Layer Control, so we might want to force it onto another line or something:

screen shot 2016-10-19 at 1 01 21 pm

This doesn't seem to be an issue with iPad Mini portrait mode:

screen shot 2016-10-19 at 1 02 48 pm

kellyi:
Is an issue on iPhone 6 portrait mode, but I think we'd want to do a more involved redesign if we decide to support this screen width, as there are a few other noticeable problems:

screen shot 2016-10-19 at 1 04 00 pm

ajrobbins:
@kellyi We are not supporting MMW for phones, just tablets

arottersman:
I'll see what I can do about the reset button.

kellyi:
#1550 was a preexisting style issue which was probably never noticed until this was in.

Tested this and it seems to work! I'm able to use the layer selector on tablets. I tested the iPad and the Windows tablet on devices, plus an iPhone 6. I also tested the simulated screen sizes for a Nexus 7 and an iPad Mini in Chrome.

I also checked the desktop version in Chrome with these changes and the layer selector still works there, too.

+1 for this. It'd be nice if we could get the Reset button adjusted on Nexus 7 sized devices (which I assume have the narrowest viewport of anything we support?), but if it's very involved, it may be worth leaving that for another issue.

arottersman:
Now if the device is less than or equal to about Nexus portrait width (710px), it sets the toolbar to be `display: block` so that the reset button is on a new line.

Thanks for all the testing!

Issue #1546, opened 2016-10-18
Change "agrarian" to "agriculture" in catchment popup


ajrobbins:
Popup when catchment ID is clicked under "Water Quality" tab of analyze:

![screen shot 2016-10-18 at 5 15 55 pm](https://cloud.githubusercontent.com/assets/12868342/19496701/b52a6eec-9556-11e6-89a6-64d8c88d8a39.png)

lewfish:
Shares a point with #1537

Issue #1537, opened 2016-10-12
Outlet points for invalid watershed are displayed


kdeloach:
Delineating an invalid (MultiPolygon) watershed, after delineating a valid watershed, results in the following:

1. The previous valid watershed AOI is displayed
2. The outlet point markers from the current (invalid) watershed are displayed

Steps to reproduce:

1. Zoom to "Bug's Airport" using geolocator
2. Delineate a watershed North West of the airport (above Hollo Rd., East of Nazareth Pike)

    ![image](https://cloud.githubusercontent.com/assets/43062/19324890/b2ec962e-9091-11e6-86f2-a4ac05a75670.png)

3. Delineate a watershed directly North of Hollo Rd. from the airport

    ![image](https://cloud.githubusercontent.com/assets/43062/19324964/01abf714-9092-11e6-95a3-d5424699df6b.png)

4. Observe that the previous AOI is restored, and the outlet points for the current (invalid) watershed are displayed

See discussion from #1524 for added context.

PR #1549, opened 2016-10-19
Clear old AOI and markers when RWD fails

lewfish:
This PR makes it so that the AOI and RWD markers are cleared when an RWD run fails after a successful one. It also changes 'Agrarian' to 'Agriculture'.

#### Testing
* Search for Bug's Airport using geocoder
* Delineate a watershed at a point as in screenshot below
screen shot 2016-10-19 at 12 59 27 pm
* Click back from analyze, and then click Delineate Watershed and click the same point as in the screenshot below (don't click Reset first). This should cause an error message to popup, but no AOI or markers should be shown.
screen shot 2016-10-19 at 12 59 47 pm 2
* The drawing functionality should still work as before. Specifically, drawing an AOI, then clicking back, then drawing part of a polygon and hitting escape should revert to the previous AOI.
* In the Water Quality tab, click a catchment ID in the Water Quality tab and check that Agrarian has been changed to Agriculture.

Connects #1537 
Connects #1546

### DIFF
--- a/src/mmw/js/src/core/templates/catchmentWaterQualityPopup.html
+++ b/src/mmw/js/src/core/templates/catchmentWaterQualityPopup.html
@@ -10,7 +10,7 @@
     </thead>
     <tbody>
         <tr>
-            <td>Agrarian</td>
+            <td>Agriculture</td>
             <td class="text-right">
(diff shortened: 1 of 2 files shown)


kdeloach:
+1

Issue #1545, opened 2016-10-18
Simplify catchment geometry for Analysis


mmcfarland:
The analysis results for WQ catchments return all of the geometries for intersected polygons.  These polygons are quite complex given their size and the resulting payload for a large AoI can take a long time to download.  The geometries should be simplified in the query before they are returned to the client as vectors.

![screenshot 2016-10-18 at 4 26 56 pm](https://cloud.githubusercontent.com/assets/1014341/19495286/bfb132ac-9550-11e6-8bbe-55ae8e29bc08.png)

PR #1551, opened 2016-10-19
Simplify catchment polygon in water quality analysis

lewfish:
The water quality endpoint sends back the catchment polygons which can be large, so we simplify them as part of the DB query. Below, you can see the same catchment without and with the simplification.

screen shot 2016-10-19 at 2 21 50 pm 2
screen shot 2016-10-19 at 2 19 25 pm 2

#### Testing
* Draw an AOI near the DRB and go to the Water Quality tab. When hovering over the rows, you should see polygons that are simplified relative to those when running the `develop` branch.

Connects #1545

### DIFF
--- a/src/mmw/apps/modeling/calcs.py
+++ b/src/mmw/apps/modeling/calcs.py
@@ -110,7 +110,8 @@ def catchment_water_quality(geojson):
           tn_urban_k, tn_riparia, tn_ag_kgyr, tn_natural, tn_pt_kgyr,
           tp_urban_k, tp_riparia, tp_ag_kgyr, tp_natural, tp_pt_kgyr,
           tss_urban_, tss_rip_kg, tss_ag_kgy, tss_natura,
-          tn_yr_avg_, tp_yr_avg_, tss_concmg, ST_AsGeoJSON(geom) as geom
+          tn_yr_avg_, tp_yr_avg_, tss_concmg,
+          ST_AsGeoJSON(ST_Simplify(geom, 0.0003)) as geom
           FROM {table_name}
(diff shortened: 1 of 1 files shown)


src/mmw/apps/modeling/calcs.py
```diff
@@ -110,7 +110,8 @@ def catchment_water_quality(geojson):
           tn_urban_k, tn_riparia, tn_ag_kgyr, tn_natural, tn_pt_kgyr,
           tp_urban_k, tp_riparia, tp_ag_kgyr, tp_natural, tp_pt_kgyr,
           tss_urban_, tss_rip_kg, tss_ag_kgy, tss_natura,
-          tn_yr_avg_, tp_yr_avg_, tss_concmg, ST_AsGeoJSON(geom) as geom
+          tn_yr_avg_, tp_yr_avg_, tss_concmg,
+          ST_AsGeoJSON(ST_Simplify(geom, 0.001)) as geom
```
mmcfarland: The sample polygon in the image does seem very simplified, is there a balance in this threshold that could get us a little more detail without getting a lot more nodes?

lewfish:
This is with the threshold set to 0.0003.

screen shot 2016-10-19 at 3 48 03 pm

mmcfarland:
That looks great and still seems to have many fewer vertices.  Does it reduce the analyze payload for WQ significantly?

lewfish:
> Does it reduce the analyze payload for WQ significantly?

The payload is 3.5x the size without simplification.

kdeloach:
+1

PR #1562, opened 2016-10-24
Update the embedded micro link

mmcfarland:
The app will redirect from /micro but this puts the explicit new link
in the url.

#### Testing
Run the Site Storm Model and verify that the new static micro site URL is in link at the bottom of the results page.

screenshot from 2016-10-24 10 48 17


arottersman:
Tested! +1

PR #1564, opened 2016-10-24
Fix/ui error on mapshed timeout

jtarricone:
This fixes a bug that resulted in a bad UI state if a modeling job timed out. As a corollary, it adds a minor refactor to address the issue of null references in jQuery callbacks, which would sometimes have UI effects but were mostly just reported to the console.

To test:
* open the browser console
* perform various user actions, including but not limited to:
* perform sequences of analysis & modelling actions without waiting for the asynchronous calls to return;
* as in the last item, perform sequences of actions and then back out, e.g. pick an AoI, start a tr-55 / multiyear model run before the analysis job completes, back out of the model run view once it starts polling, activate the other model type and let it proceed;
* start analysis/modelling actions that will time out, e.g. select half of Kansas as an AoI and then immediately kick off a tr55 run, or select a ~1,000 sq-km AoI and start a multiyear run; 
* with a completed model run, add some scenario modifications
* with a completed model run, rename a scenario to a taken name to ensure it raises an error
* save and reload a project with nontrivial scenarios

In the course of testing, there should not be any UI errors and the app should behave as expected. In the console, there should be a handful of messages to the effect of "`Caught TypeError exception that implies a callback encountered a null reference`" followed by the specific member that was found to be null.

Connects #1559

### DIFF
--- a/src/mmw/js/src/modeling/controllers.js
+++ b/src/mmw/js/src/modeling/controllers.js
@@ -144,7 +144,7 @@ var ModelingController = {
                         ).census
                     );
                 })
-                .fail(projectCleanUp);
+                .fail(projectErrorState);
 
             setupNewProjectScenarios(project);
             finishProjectSetup(project, lock);
@@ -156,6 +156,10 @@ var ModelingController = {
 
     projectCleanUp: function() {
         projectCleanUp();
+    },
+
+    projectErrorState: function() {
+        projectErrorState();
     },
 
     makeNewProjectCleanUp: function() {
@@ -288,6 +292,18 @@ function projectCleanUp() {
     App.getMapView().updateModifications(null);
     App.rootView.subHeaderRegion.empty();
     App.rootView.sidebarRegion.empty();
+}
+
+function projectErrorState() {
+    if (App.currentProject) {
+        var scenarios = App.currentProject.get('scenarios');
+        App.currentProject.off('change:id', updateUrl);
+        scenarios.off('change:activeScenario change:id', updateScenario);
+        App.currentProject.set('scenarios_events_initialized', false);
+        App.projectNumber = scenarios.at(0).get('project');
+    }
+
+    App.getMapView().updateModifications(null);
 }
 
 function updateItsiFromEmbedMode() {


src/mmw/js/src/core/models.js
```diff
@@ -136,7 +150,7 @@ var TaskModel = Backbone.Model.extend({
             }),
             pollingDefer = $.Deferred();
 
-            startDefer.done(function() {
+            startDefer.done(handleNullInCallback(function() {
```
arottersman: Replace `done` with `then`. If `done` fails I think nothing happens, but if `then` fails it gets caught and goes to `fail`

src/mmw/js/src/core/models.js
```diff
@@ -6,6 +6,20 @@ var Backbone = require('../../shim/backbone'),
     turfArea = require('turf-area'),
     utils = require('./utils');
 
+var handleNullInCallback = function(callbackFn) {
+    return _.wrap(callbackFn, function(fn, args) {
```
kdeloach: Using `_.wrap` may be unnecessary since you can just return the wrapped function directly.

Ex.

```
return function() {
    callbackFn.apply(this, arguments);  
};
```

src/mmw/js/src/core/models.js
arottersman: Where do these null references come from in the first place?

caseypt:
I was trying to reproduce this, I was only able to trigger and catch the `Cannot read property 'show' of undefined` error. I did this by switching back and forth between the two models before they had a chance to complete, or switching to compare view before a model completed. Is there another error that you were able to trigger and catch? If not, I think we should consider trying to fix this error directly.

mmcfarland:
Can you retarget this PR to the `release/1.16.0` branch?

src/mmw/js/src/core/models.js
jtarricone: >Where do these null references come from in the first place?

They occur because UI objects that existed when the async process starts have been destroyed by the time the callbacks are invoked.

>Replace done with then. If done fails I think nothing happens, but if then fails it gets caught and goes to fail

I think you're probably right that this could be done within a `then()` but it may not look that different. I think the two options in that case are either `then()` would chain the exception handler into the promise and the `done` and `fail` callbacks are chained unmodified; or, specify the `done` and `fail` (wrapped) callbacks in `then` and then have a single resolution. I'm not sure if an exception raised within a callback within `then()` would result in the chained `fail()` being invoked, but assuming it would, it wouldn't be the desired behavior because `fail` would then be invoked *either* because the promise returned successfully but the callback failed, or because the promise failed to begin with.

jtarricone:
The connected issue was about a bug that would cause the UI to go into a bad state if the modeling task timed out. To replicate that (not on this branch), start a long-running modeling task, such as a TR55 run for all of Pennsylvania, immediately following the AoI selection, and then just wait for it to time out.

The second commit addresses a problem that was made apparent by fixing the connected bug. I can remove that commit if it causes confusion, but the console will show errors and there could be less-serious UI effects as a result.

src/mmw/js/src/core/models.js
```diff
@@ -6,6 +6,20 @@ var Backbone = require('../../shim/backbone'),
...
+        try {
+            fn(args);
+        } catch (e) {
+            if (e instanceof TypeError) {
+                console.log('Caught TypeError exception that implies a callback encountered a null reference --> ' + e.message);
```
jtarricone: I think that this null reference problem *did* occur before, but it was much more frequent after I fixed the connected bug, so it seemed appropriate to address it. The most noticeable effect prior to this fix was that the console showed errors in place of this message, and I recall there was a minor UI problem as well.

jtarricone:
>Can you retarget this PR to the release/1.16.0 branch?

Done.

caseypt:
> @caseypt the connected issue was about a bug that would cause the UI to go into a bad state if the modeling task timed out. To replicate that (not on this branch), start a long-running modeling task, such as a TR55 run for all of Pennsylvania, immediately following the AoI selection, and then just wait for it to time out.

Thanks. I'm able to get the analyze request to time out (the land and soil results show an error, through the animals and point source results always succeed), but I'm not seeing anything odd in the UI on staging/production. What should I be looking for?

I think this is outside of the scope of this issue, but even though the analyze request has timed out, I'm able still able to select and run a model. Should we disable model running if there are no analyze results?

jtarricone:
The sixth bullet point in the testing blurb should replicate the original issue; you have to select a large AoI and then **immediately** move to the TR55 modelling run. Once  the analysis task times out, the model view will get thrown into a bad state.

caseypt:
> The sixth bullet point in the testing blurb should replicate the original issue; you have to select a large AoI and then immediately move to the TR55 modelling run. Once the analysis task times out, the model view will get thrown into a bad state.

Thanks for the clarification. I compared the behavior on this PR with staging and verified that the weird UI state is fixed.

As for the second commit, consider removing it (but maybe keeping it around on a separate branch for reference?). I think we should create a new issue documenting the problem and address it separately from this PR.

jtarricone:
I stashed the second commit and pushed the updated branch; the browser console will show the exceptions, but I didn't notice any other differences in my testing.

kdeloach:
Created issue #1572 for the other null exception errors.

Issue #1555, opened 2016-10-20
itsi_embed option doesn't handle anonymous users


scytacki:
We recently started using the `itsi_embed` option. This fixes the snapshots, but it broke anonymous access.

The problem seems to be that MMW automatically tries to log in with ITSI if the `itsi_embed` option is used.  But if the activity being used anonymously then this fails. Here is an example activity: https://authoring.concord.org/activities/5667  If you make sure you are not logged in and then run that activity, scroll to the MMW interactive and click "click here to start the interactive", you can see the issue.

There are two solutions, I can think of:
1. Don't try to auto login the user when itsi embed is used.
2. Use the lara-interactive-api to figure out if the model is being run anonymously or not.

Option 2 is better but a bit more complicated. The way you do it is:
Send the `getAuthInfo` message to LARA. This documented here: http://concord-consortium.github.io/lara-interactive-api/docs/#getauthinfo
Then wait for the `authInfo` response. The authInfo response has a `loggedIn` property. If that is false then MMW should log the user in as a guest, if it is true, then it should log the user in through ITSI.  Or if you want to make it more flexible, then MMW should use the `provider` property to figure out which provider it should try to log the user in with.  MMW should also handle the case where no response is sent after a reasonable amount of time. Perhaps 30s is good here. If no `authInfo` response is received MMW could show the usual login dialog.

rajadain:
Hi Scott,

I've reviewed notes from a recent meeting, and done some preliminary investigation, and this is what I've found:

 * We can't query `auth/concord_id/user.json` from the Django middleware because that server doesn't know who the user is logged in as. That GET request needs the `cc_auth_token` cookie to return the right value.
 * We can't XHR `auth/concord_id/user.json` from the front-end either, because of CORS limitations
 * `getAuthInfo` will only return `loggedIn` as `true` or `false`, and doesn't tell us who the user is logged in as. 

The best we can do is, as you suggested above, disable the auto-login and use LARA to see if the user is logged in outside. If they are, we can trigger the OAuth dance. Else, we continue as guest. They can always click "Login" in the interim and log themselves in.

The tricky part is going to be dealing with saved state. Imagine the following scenario:

 * Logged in user opens activity.
 * LARA communicates the last saved URL (via `loadInteractive`), which is a private project belonging to the user
 * The project is not visible because the user isn't logged in yet
 * We do the redirect dance to log the user in, but they come back to the home page instead of the project
 * They are not taken to their original project

[...]

PR #1570, opened 2016-10-26
Log ITSI Embed Users in as Guests

arottersman:
Model My Watershed embedded in an ITSI activity currently assumes that the user is an ITSI user, and either signs them in with the ITSI credentials available or automatically takes them to the ITSI login. There's no way to sign in as a guest. 

There are a few cases where the ITSI activity's user will be an "Anonymous" user, in which case they won't be able to use the MMW interactive at all. Because no one really uses the ITSI/logged-in user specific features of MMW in the embedded interactive anyway, and we need to allow anonymous users access to the interactive, we are now going to default all MMW `itsi_embed` interactives to auto sign in users as guests.

This PR...
- Removes the auto ITSI login from the backend
- Defaults to guest login if `itsi_embed=true`
- Removes other ITSI embed specific logic
- Removes the notion of `activityMode` and the boolean on a project `is_activity` -- there's nothing special about modeling in an ITSI activity anymore
### Testing

Pull this branch. You will need to:

```
MMW_ITSI_SECRET_KEY="<the secret key>" vagrant reload
./scripts/bundle.sh
./scripts/manage.sh migrate
```

Please follow the "Prepare An Activity Section" directions from [this PR](https://github.com/WikiWatershed/model-my-watershed/pull/675) to set up an ITSI embed viewer pointing to your local.
- While not logged into ITSI, confirm that MMW works and doesn't prop you to sign in
- Go through the steps of modeling and confirm there's no reference to saving anything
- Take a screen cap
- Confirm there are no errors in the console as you create scenarios, etc

In regular MWW you should still be able to log into ITSI

Connects #1555


arottersman:
Rebased on `release/1.16.0`

src/mmw/apps/modeling/migrations/0020_remove_project_is_activity.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('modeling', '0019_project_gis_data'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
hectcastro: Trying to determine the implications of this schema change. In a scenario where the old release attempts to interact with the database that is in a state after this migration is applied, it'll fail when a new project is created?

arottersman: I don't think it would fail - the field wasn't required, and we don't specify it in the Rest Framework serializer. Any old releases that are ITSI embed viewers might have some unexpected behavior though (being able to create more than one project in an activity).

Happy to remove the migration/model change if you think that'd be best.

hectcastro: I don't think we need to remove it, but we may consider applying this migration after the 1.16.0 release is live.

For a brief period, the previous release may have to interact with the new database schema as we apply migrations before cutting over completely to the new release. The other approach is to cut over to the new release, and then apply migration (implying that the code in 1.16.0 would work OK with the current schema).

Given that we're removing a field, it seems like applying the relation after 1.16.0 is deployed seems like the way to go. Agreed?

arottersman: Yup, that sounds good!

jtarricone:
This looks good, but the app seems to choke if you try to go back the the basic "Draw" mode from the Analyze view. The console shows the familiar `App.currentProject is null` and the app seizes up as shown here -->
20161031_mmw

arottersman:
Do we need to show that model at all anymore? @ajrobbins

ajrobbins:
@arottersman I think it's still good to give them an alert that they are navigating away from their project, though there's no way to actually save it. They might want to take a snapshot before doing so.

arottersman:
I've just pushed a fixup that adds back in the concept of a "current project" for the embed view. Now you'll see that modal if you try to reset an AOI and it _won't_ freeze, and you'll see your activity initialize as "New Activity" instead of "Untitled Project"

arottersman:
Reopening targeted to `develop`

caseypt:
The issue I mentioned above was due to the fact that the ITSI env var wasn't being set. I had to manually edit the env file to set it, and that fixed my issue.

arottersman:
Thanks! I had the same issue with the env var - maybe `MMW_ITSI_SECRET_KEY="<the secret key>" vagrant reload` doesn't do what I'd expect it to.

hectcastro:
Here I think:

```bash
$ MMW_ITSI_SECRET_KEY="<the secret key>" vagrant reload
```

Needs to be something like:

```bash
$ MMW_ITSI_SECRET_KEY="<the secret key>" vagrant reload --provision
```

Ansible picks up the `MMW_ITSI_SECRET_KEY` environment variable and tries to put it in the right place, but a `reload` without `--provision` doesn't trigger Ansible.

arottersman:
Ah, that's right. Thanks, @hectcastro!

Issue #1554, opened 2016-10-20
Update Google Analytics to track "virtual" pageloads


mmcfarland:
Our google analytics setup records the initial pageload for `/`, but then is unable to detect user behavior within the app.  Google offers features that help alleviate that issue for SPAs.  Update our setup to better track these client initiated URL changes and record if anything must be done on the Stroud GA account to  enable them.

https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications

PR #1571, opened 2016-10-26
Track virtual pageloads with Google Analytics

lewfish:
Since MMW is a single page app, it doesn't have many 'real' page loads as the user navigates the app. This PR makes it so that virtual page loads are recorded by Google Analytic. No changes need to be made to the account setup for Stroud. This follows the approach in https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications

screen shot 2016-10-26 at 12 56 07 pm

#### Testing
* Login to Google Analytics using azaveadev@azavea.com login. The password is on the wiki.
* Click MMW - All Data
* Go to Real Time -> Content in panel on the left.
* Navigate around the app. Every time the URL updates, it should appear in Google Analytics. (Although sometimes there's a lag of a second or two).

Connects #1554

### DIFF
--- a/src/mmw/apps/core/templates/base.html
+++ b/src/mmw/apps/core/templates/base.html
@@ -1,5 +1,31 @@
 {% include 'head.html' %}
 {% load staticfiles %}
+
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    {% block metatitle %}
+    <title>Model My Watershed</title>
+    {% endblock metatitle %}
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link rel="icon" type="image/png" href="{% static 'favicon.png' %}" sizes="16x16">
+    <link rel="icon" type="image/png" href="{% static 'favicon@2x.png' %}" sizes="32x32">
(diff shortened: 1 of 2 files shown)


lewfish:
Was this what you were looking for @mmcfarland ?

lewfish:
One potential issue is that the URLs are going to contain id numbers eg. `project/10/...` and Stroud probably wants to collapse these into `/project` for analysis purposes. I think you can set that up using regular expressions in GA Dashboard, or we could just send the root of the URL from the client.

mmcfarland:
Yep, looks like it - I'll do a review

mmcfarland:
Curiously, it sends two ga requests per page view:

screenshot from 2016-10-26 15 00 18

But when I remove the `ga('send'` it sends 0.  The configuration looks correct based on the documentation.  This looks good to me, but we should check GA after it's been on the staging environment to ensure it's not double counting anything.  The code is getting executed once per transition, correctly.

+1

mmcfarland:
Heads up that the new `ga` global is causing lint failures

lewfish:
@azavea-bot rebuild

Issue #1565, opened 2016-10-24
Fix: hover tooltip persists after view re-render


jtarricone:
![20161024_mmw_bug](https://cloud.githubusercontent.com/assets/18290666/19661787/0d3aa95c-9a03-11e6-944e-496dae758352.png)

It only appears to happen if the cursor remains on the chart region, making the tooltip for the chart active, when the asynchronous call returns and the view re-renders. The tooltip will then persist through all app states until a full refresh occurs.

mmcfarland:
Nice job noting the steps to reproduce.  I've had this intermittently, but can get it every time with those steps.  

Tagging this as NSF since it was reported by classroom groups.

PR #1574, opened 2016-10-27
Remove tooltip when chart is removed

lewfish:
Previously, when new results would show up in the vertical bar chart, an extra copy of the tooltip would appear. This was because new results would trigger the creation of a new bar chart. The old chart would get cleared out, but the old tooltip would persist because it was not removed from the DOM along with the old chart because the tooltips are in a different part of the DOM. This PR fixes that problem by manually removing the tooltip when the old chart elements are removed. Note that there are a bunch of tooltip related issues in the NVD3 repo. https://github.com/novus/nvd3/issues/1312 seems closest to our issue, but I tried the latest release and the problem was still present.

While working on this PR, I discovered that the chart is being recreated three times after the precipitation value is changed. Ideally, this wouldn't happen, and the chart would be merely updated with new data. I still think the fix in this PR is necessary though for handling switching from the modeling view to the compare view. In this case, we actually do want to destroy the chart. On the other hand, I suppose you could keep the modeling view in the DOM while the compare view is being displayed, but this would be a more radical departure from how Backbone apps normally work.

#### Testing
* Go to the TR55 modeling view runoff chart.
* Change the precipitation value via the slider. While it is loading, hover the mouse over the chart to show the tooltip. When the spinner stops and new results appear in the chart, the tooltip should disappear. After moving the mouse slightly, the tooltip should reappear and there should only be one copy.
* Try this in the compare view.

Connects #1565

### DIFF
--- a/src/mmw/js/src/core/chart.js
+++ b/src/mmw/js/src/core/chart.js
@@ -7,6 +7,30 @@ var d3 = require('d3'),
 
 var widthCutoff = 400;
 
+// Make jQuery handle destroyed event.
+// http://stackoverflow.com/questions/2200494/
+// jquery-trigger-event-when-an-element-is-removed-from-the-dom
+(function($) {
+    $.event.special.destroyed = {
+        remove: function(o) {
+            if (o.handler) {
+                o.handler();
+            }
+        }
+    };
+})($);
+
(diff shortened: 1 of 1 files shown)


jtarricone:
+1 Tested. Good job figuring this out, especially noting that the charts are being needlessly rendered.

Issue #1569, opened 2016-10-26
Reduced rendered row count for Analysis results


mmcfarland:
Within Analysis result tables, determine a number of rows that has suitable rendering times and conform the existing tables to a paged layout scheme to keep the number of rows rendered at any time to that minimum.

A potential library option: https://github.com/backbone-paginator/backbone.paginator#id2

In reference to performance issues reported here: #1561 #1568

PR #1579, opened 2016-10-28
Feature/add table pagination

jtarricone:
This adds pageable tables, specifically intended for Backbone projects, to the "Point Source" and "Water Quality" views, to address performance issues stemming from rendering the records in larger AoI's.

To test:
- clone this branch and bring up and environment, checking that `bootstrap.paginator` successfully installs
- select a larger AoI and wait for the analysis task to complete
- look at the "Point Source" and "Water Quality" tabs and make sure that the tables are paged
- test the paging buttons to ensure the table updates as expected
- test the table sorting, where approriate
- repeat the process at different viewport sizes to check that the tables are usable and that styling is reasonably consistent

To test performance, select the same AoI's on the locally hosted app and on the staging app; for both, start the profiler, then activate the point source and water quality tabs, then start a TR55 model run and back out once the model view renders. Once the analysis view renders again after the model->analysis transition, stop the profiler. Compare the sequence of actions between the two profiler sessions. For a large AoI with several records, the render time should be measurably lower in the local app.

Connects #1569

### DIFF
--- a/src/mmw/js/src/analyze/models.js
+++ b/src/mmw/js/src/analyze/models.js
@@ -1,7 +1,7 @@
 "use strict";
 
 var $ = require('jquery'),
-    _ = require('lodash'),
+    lodash = require('lodash'),
     Backbone = require('../../shim/backbone'),
     coreModels = require('../core/models');
 
@@ -20,7 +20,7 @@ var LayerCategoryCollection = Backbone.Collection.extend({
 });
 
 var AnalyzeTaskModel = coreModels.TaskModel.extend({
-    defaults: _.extend( {
+    defaults: lodash.extend( {
             area_of_interest: null,
             taskName: 'analyze',
             taskType: 'modeling'
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -1,7 +1,7 @@
 "use strict";
 
 var $ = require('jquery'),
-    _ = require('lodash'),
+    lodash = require('lodash'),
     L = require('leaflet'),
     Marionette = require('../../shim/backbone.marionette'),
     Backbone = require('../../shim/backbone'),
@@ -58,7 +58,7 @@ var ResultsView = Marionette.LayoutView.extend({
 
         var modelPackages = settings.get('model_packages'),
             modelPackageName = $(e.target).data('id'),
-            modelPackage = _.find(modelPackages, {name: modelPackageName}),
+            modelPackage = lodash.find(modelPackages, {name: modelPackageName}),
             newProjectUrl = '/project/new/' + modelPackageName,
             projectUrl = '/project',
             alertView,
@@ -69,7 +69,7 @@ var ResultsView = Marionette.LayoutView.extend({
             analysisResults = JSON.parse(App.getAnalyzeCollection()
                                             .findWhere({taskName: 'analyze'})
                                             .get('result') || "{}"),
-            landResults = _.find(analysisResults, function(element) {
+            landResults = lodash.find(analysisResults, function(element) {
                     return element.name === 'land';
             });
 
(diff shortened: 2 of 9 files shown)


jtarricone:
I had to make some style changes to support the paginated tables; @jfrankl can you or someone from design look over this branch and recommend any changes?

src/mmw/js/src/analyze/models.js
```diff
@@ -1,7 +1,7 @@
 "use strict";
 
 var $ = require('jquery'),
-    _ = require('lodash'),
+    lodash = require('lodash'),
```
jtarricone: Having lodash aliased to `_` caused an odd scoping problem that I think was due to a naming conflict with underscore (also aliased to `_`), which resulted in anything in the lodash namespace not being in scope.

src/mmw/js/src/analyze/views.js
```diff
@@ -457,6 +461,22 @@ var PointSourceTableView = Marionette.CompositeView.extend({
       }).render().el, { closeButton: false });
     },
 
+    nextPage: function() {
+        if (this.collection.hasNextPage()) {
+            this.collection.getNextPage();
+            this.render();
```
jtarricone: Forcing a render was the only way I found to solve an issue that caused the first set of records to essentially be "stuck." The paging would still work as a means to cycle through the records, but the paged records would render below the initial set, no matter how many pages were cycled.

ajrobbins:
Would you mind posting a screenshot of the style changes?

jtarricone:
20161028_mmw_1
20161028_mmw_2
20161028_mmw_3
20161028_mmw_4
20161028_mmw_5
20161028_mmw_6
20161028_mmw_7
20161028_mmw_8

jtarricone:
Sorry, looks like the image got cropped somehow.

jtarricone:
20161028_mmw_7
20161028_mmw_8

jtarricone:
The next set show the app at a smaller view port size. The primary difference is that the paging controls are not in a sticky footer, but rather at the very bottom of the tab. I circled record 15019 to show the effect of scrolling.

kellyi:
About to take a look.

kellyi:
The pagination controls are there and seem to work, but the table contains too many records on each page to view an entire page without scrolling:

screen shot 2016-10-31 at 9 32 14 am

screen shot 2016-10-31 at 9 33 44 am

kellyi:
They also seem to appear as active even if there's only one page of results:

screen shot 2016-10-31 at 9 35 15 am

jtarricone:
I don't think the goal was to avoid any scrolling, but rather to reduce the number of records. I would think that having perfectly sized tables would require either a conservative constant number of records or else doing some math and styling behind the scenes.

jtarricone:
>They also seem to appear as active even if there's only one page of results:

I think that would be fairly straightforward to hide if desired, but I had assumed that it was unnecessary at this point.

kellyi:
Code seems to work well! 

Going to tag @jfrankl and @ajrobbins to ask about the scrolling/pagination. The other tabs do still scroll, but their tables appear beneath another chart and they don't have numerous records. I think it feels a little weird to have a paginated table that's still so tall that it requires scrolling through a ton of records. For the water quality catchments, 50 records is about 2x as many as will fit on a single page on my external display, which means I have to scroll the top set of results entirely off the screen to get to the bottom:

screen shot 2016-10-31 at 9 47 41 am

&

screen shot 2016-10-31 at 9 47 51 am

jtarricone:
>For the water quality catchments, 50 records is about 2x as many as will fit on a single page on my external display, which means I have to scroll the top set of results entirely off the screen to get to the bottom:

The easiest approach might be to just map page size to a handful of viewport sizes, but I'm not sure if that might get wonky if the viewport changes after the app renders. Similarly, hiding the footer controls based on page size might be problematic in the same cases.

jtarricone:
I took general guidance from @mmcfarland about setting the page size, but it was based more on how the different sizes affected the rendering time. If there's a case for changing that based on usability or style concerns I can probably push a fix in time for the release.

ajrobbins:
@mmcfarland and I think having to scroll to see the 50 records/pagination at the bottom is fine, since you had to scroll before anyways.

kellyi:
Sounds good. Going to take one last look at this.

kellyi:
+1. Works well!

Issue #1573, opened 2016-10-26
Retry SJS Celery Job on requests.exceptions.ConnectionError


rajadain:
In some cases when there is a connection error between Celery and SJS, [this line](https://github.com/WikiWatershed/model-my-watershed/blob/90edb189e99848ce262f778996648a212ee603e3/src/mmw/apps/modeling/geoprocessing.py#L63) fails and throws a `requests.exceptions.ConnectionError` exception, which causes the entire Celery chain to fail.

Add a catch handler for that exception which retries the job instead. A-la:

``` python
try:
    response = requests.post(jobs_url, data=json.dumps(data))
except ConnectionError:
    if (retry):
        retry()
```

hectcastro:
Only concern with doing it at the Celery level vs. Requests level is that we may not want to retry the HTTP request as many times as we might already be retrying the entire job.

An overview of retries at the Requests level is summarized nicely [here](http://www.coglib.com/~icordasc/blog/2014/12/retries-in-requests.html).

rajadain:
We can also split this into it's own Celery task with a different number of retries.

PR #1582, opened 2016-10-31
Retry on Requests ConnectionError against SJS

hectcastro:
When the Requests HTTP client is used to interact with Spark JobServer, ensure that the Celery job is retried when a ConnectionError exception is raised.

The `retry()` calls make use of the `exc` keyword argument. This is used to supply a custom exception when the max restart limit is exceeded.

Attempts to resolve https://github.com/WikiWatershed/model-my-watershed/issues/1573

---

**Testing**

Use `debugserver.sh` to be able to monitor the Celery worker process:

``` bash
$ ./scripts/debugcelery.sh
```

From there, SSH into the `worker` and delete the `geoprocessing` Spark context:

``` bash
vagrant@worker:~$ curl localhost:8090/contexts/
["geoprocessing"]
vagrant@worker:~$ curl -XDELETE localhost:8090/contexts/geoprocessing
OK
vagrant@worker:~$ curl localhost:8090/contexts/
[]
```

Then, submit an analyze job via the UI and ensure that it completes after a retry.

Next, shut down Spark JobServer and submit an analyze job via the UI. As it begins retrying, start Spark JobServer to see if it completes:

``` bash
vagrant@worker:~$ sudo stop spark-jobserver
vagrant@worker:~$ sudo start spark-jobserver
```

### DIFF
--- a/src/mmw/apps/modeling/geoprocessing.py
+++ b/src/mmw/apps/modeling/geoprocessing.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import
 from django.conf import settings
 from django_statsd.clients import statsd
 from celery.exceptions import MaxRetriesExceededError
+from requests.exceptions import ConnectionError
 
 import requests
 import json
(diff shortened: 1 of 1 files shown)


src/mmw/apps/modeling/geoprocessing.py
```diff
@@ -60,29 +61,36 @@ def sjs_submit(host, port, args, data, retry=None):
...
         error = response.json()
-        if error['status'] == 'NO SLOTS AVAILABLE' and retry:
-            try:
-                retry()
-            except MaxRetriesExceededError:
-                raise Exception('No slots available in Spark JobServer.\n'
-                                'Details = {}'.format(response.text))
+
+        if error['status'] == 'NO SLOTS AVAILABLE' and retry is not None:
+            retry(exc=Exception('No slots available in Spark JobServer.\n'
```
rajadain: `exc` is a good find.

src/mmw/apps/modeling/geoprocessing.py
```diff
@@ -60,29 +61,36 @@ def sjs_submit(host, port, args, data, retry=None):
...
+                                'Details = {}'.format(response.text)))
         elif error['result'] == 'context geoprocessing not found':
             reboot_sjs_url = '{}/contexts?reset=reboot'.format(base_url)
             context_response = requests.put(reboot_sjs_url)
 
             if context_response.ok:
-                if retry:
-                    retry()
+                if retry is not None:
+                    retry(exc=Exception('Geoprocessing context missing in '
```
rajadain: So with this, if retry is caused and ultimately exceeds max retries, we will see this exception instead of the `MaxRetriesExceededError`? Sweet.

hectcastro: Yeah. A caveat I encountered is:

> If this argument is set and retry is called while an exception was raised (sys.exc_info() is set) it will attempt to reraise the current exception.
> If no exception was raised it will raise the exc argument provided.

jtarricone:
+1 Tested per the instructions, everything appeared to work as expected.

Issue #1586, opened 2016-10-31
Analyze Table UI issues


mmcfarland:
Two quick updates for the analyze table:

* [x] Update Water Quality description, missed the word "Delaware" in description.  Should read:

> Delaware River Basin only: Stream Reach Assessment Tool model estimates

![screenshot from 2016-10-31 19 34 50](https://cloud.githubusercontent.com/assets/1014341/19875248/97f72a9a-9fa1-11e6-835b-b3e831a4bcad.png)

* [ ] Fix number of decimals in table columns.  Number should be consistent so values visually sort when column is sorted.

![screenshot from 2016-10-31 19 35 22](https://cloud.githubusercontent.com/assets/1014341/19875273/d7c3e50a-9fa1-11e6-9d06-cd99f75331d7.png)

PR #1589, opened 2016-11-01
Update water quality analysis tab description

kellyi:
This PR adds "Delaware" to the Water Quality analysis tab description. It's the first part of #1586, and I'm pushing it here to separate it off from the work for the second part of that card.

**Testing**
- get this branch, bring up the dev env, run `./scripts/bundle.sh`, then open the browser and console and analyze an aoi within the drb
- verify that the water quality analysis tab now reads "Delaware River Basin..." and that everything else works as it should and that there aren't any errors in the console.

Connects #1586

### DIFF
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -744,7 +744,7 @@ var PointSourceResultView = AnalyzeResultView.extend({
 
 var CatchmentWaterQualityResultView = AnalyzeResultView.extend({
     onShow: function() {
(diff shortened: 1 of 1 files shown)


lewfish:
+1. I'm having trouble getting MMW running right now, so can't test this, but am fairly confident it will work.

kellyi:
Cool. Thanks for taking a look!

kellyi:
Thanks for taking a look @mmcfarland!

PR #1591, opened 2016-11-01
Show point source and water quality data to three digits

kellyi:
This PR updates the point source and water quality analysis tabs and popups to display the data to three digits. Previously we'd been displaying it haphazardly, which made it difficult to compare at a glance. With these changes, we force each value in these tables to be three digits (and do the same with the associated popups):

screen shot 2016-11-01 at 4 48 04 pm

&

screen shot 2016-11-01 at 4 48 04 pm

and

screen shot 2016-11-01 at 4 49 24 pm

&

screen shot 2016-11-01 at 4 49 29 pm

The second commit here adds a method to sort the area column of the water quality table; the third updates the tests.

**Testing**
- get this branch, `vagrant up` and `./scripts/bundle.sh`, then open the browser and console and visit `localhost:8000`.
- select a known AOI to analyze within the DRB and verify that when the results return the point source and water quality tabs have three digits for each column but otherwise resemble the data for the same AOI on staging
- click to open a few popups from each tab and verify that the popups show three digits, too
- verify that you haven't seen any errors in the console

Connects #1586

### DIFF
--- a/src/mmw/js/src/analyze/templates/catchmentWaterQualityTable.html
+++ b/src/mmw/js/src/analyze/templates/catchmentWaterQualityTable.html
@@ -3,7 +3,9 @@
       <thead>
           <tr>
               <th data-sortable="true">Id</th>
-              <th data-sortable="true">Area (ha)</th>
+              <th data-sortable="true" data-sorter="window.noDataSort">
+                  Area (ha)
+              </th>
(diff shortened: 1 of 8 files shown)


jtarricone:
+1 Tested

kellyi:
Thanks for taking a look!

Issue #1583, opened 2016-10-31
RWD not working on border of DRB


ajrobbins:
RWD appears to not be working on the boundary edge, though Stroud reports that the point tested is indeed in the DRB and works on desktop GIS.

**To replicate:**
1. Search for "Avon Grove Intermediate School" and select the PA location
2. Zoom in, turn on the DRB streams, and select a point near the stream, between Avon Grove and Penn London schools
3. You should get an error message: Selected point is outside the DRB

**More info:**
>I’m working in the live version of MMW and have found an issue delineating a watershed on the western border of the DRB. Attached is a screenshot. It appears that the delineation function is not working on the boundary edge. It is working further within the DRB. We must be getting some kind of boundary effects. The location I’ve selected is within the DRB but near the edge (Avon Grove Intermediate School) in Chester County a little south of Rt. 1.

![delineateissue](https://cloud.githubusercontent.com/assets/12868342/19855500/4df930c4-9f4b-11e6-8366-1bd6d685f47f.PNG)

From David Tarboten:
> I experimented with this a bit to see if there was a problem with the underlying data for watershed delineation.  I was able to reproduce the problem on the live website, but when I go into a local GIS and enter the coordinates of a point as close as I could get to where Steve seems to be clicking, I am able to delineate a watershed.  The watershed delineated is below (yellow).
 
>Here is some technical information that may be helpful to Azavea in tracking this down.
 
>The longitude and latitude of point I am using is -75.883125  39.798818.  This lies within subwatershed number 2348.  This is what is highlighted.  If you look at the shapefile in \Main_Watershed\delaware_gw_5000_diss.shp from http://hydrology.usu.edu/mmw2/PreProcessed_Product_RWD/  this point should lie in shape with GRIDCODE 2348.  This should be used by the code to identify the subwatershed folder to start the watershed delineation.

![image002](https://cloud.githubusercontent.com/assets/12868342/19855520/72df7466-9f4b-11e6-8ba7-7e32c02f2435.png)

ajrobbins:
This is likely due to the simplified DRB polygon we use to validate RWD requests on the front end.

mmcfarland:
I think our best bet will be to either:
* simplify the polygon less, 
* buffer the polygon slightly (make it larger), then simplify to the same degree,
* leave as is, but do a distance check for the point to an edge and allow if it's within some tolerance.

Mostly we don't want the geojson size to increase substantially and be a large payload for page load.

PR #1596, opened 2016-11-02
Make RWD work near border of DRB

lewfish:
The DRB perimeter is used on the frontend to decide if a point can have RWD run on it. Previously, we used a simplified version to have a smaller payload, but this excluded some places that are inside the DRB. This PR makes it so the frontend uses a buffered (expanded by 20 miles) and simplified DRB (Delaware River Basin) perimeter to avoid this issue. It also switches to using the full DRB perimeter on the backend to maintain accuracy. 

As part of an effort to make the data wrangling process more repeatable, I've included a script and some instructions on how these new files were generated as part of the source code.

#### Testing
* Run `./scripts/debugserver.sh` and `./scripts/debugcelery.sh`.
* Try running RWD on a point that is clearly inside the DRB (eg. Philly). It should work.
* Try it on a point that is clearly outside the DRB (eg. Pittsburgh). The frontend should immediately reject it.
* Try it on Avon Grove Intermediate School, which was reported in the issue. It should work.
* Try it on a point that is inside the buffered polygon, but outside the true DRB perimeter (eg. Lancaster). It should start RWD, but then server should respond with an error.

Connects #1583


caseypt:
Just to be clear, da520e5 is unrelated to the DRB issue, right?

lewfish:
> Just to be clear, da520e5 is unrelated to the DRB issue, right?

Right.

caseypt:
This doesn't seem to be problem in the app, but just FYI both of the DRB GeoJSON files are invalid GeoJSON because they do not have a `properties` key.

caseypt:
Consider decreasing the size of the buffer (I'm not sure what the right number is, 2 miles?). Even though the back-end is using the real shape, it seems like we are leaving a lot of room for RWD to pass the initial test, but fail to return results.

Original:
image

Buffer
image

lewfish:
> This doesn't seem to be problem in the app, but just FYI both of the DRB GeoJSON files are invalid GeoJSON because they do not have a properties key.

I'll add that and find a better buffer distance for a snugger fit.

lewfish:
With a 3 mile buffer:

screen shot 2016-11-03 at 11 45 25 am

caseypt:
Cool. That looks good to me.

caseypt:
Tested things out with the new buffer. Looks good.

lewfish:
Thanks for the review.

Issue #1594, opened 2016-11-02
Paged table sorting


mmcfarland:
The paging control on the Analyze tables now increment pages of 50 records at a time.  The sorting mechanism (from bootstrap-table) sorts the actual contents of the rendered table, resulting in sorting _within_ a page.  The sort behavior should still sort the entire collection, rendering the first page of the newly sorted collection/table.

arottersman:
In case it's helpful -- my first thought on this was to sort the Backbone paginated collection on the bootstrap table event `onSort`. I tried this out by calling the `renderBootstrap()` function below wherever we'd usually call `$('[data-toggle="table"]').bootstrapTable()` in the `CatchmentWaterQualityTableView` in `analyze/views.js`

```
    renderBootstrap: function() {
        $('[data-toggle="table"]').bootstrapTable({
            onSort: function(name, order) { // name will be the column's `data-field-name`
                this.collection.setSorting(name, order === 'asc' ? 1 : -1, {full: true});
                this.collection.comparator = this.buildComparator(name);
                this.collection.sort()

                this.collection.setSorting(null);

                this.render()
                this.renderBootstrap();
            }.bind(this));
    },

    buildComparator: function(attributeName) {
        return function(x, y) {
            utils.noDataSort(x.attributes[attributeName],
                y.attributes[attributeName])
        }
    },
```

This did reorder the table, but not in any way I could figure out the logic behind. It definitely didn't sort it.

arottersman:
Why was this issue closed? It doesn't look like it was fixed.

rajadain:
Ugh 🤦‍♂️  I think this line in the #1849 description autoclosed this:

> Now that we're relying more on the pagination, we might have to really fix #1594.

arottersman:
Weird. Do you think it was the word `fix` or any time we mention another issue in a PR it does that? I  feel like I mention  _a lot_ of issues...

PR #1597, opened 2016-11-02
Switch To Bootstrap Table's Pagination

arottersman:
When we were letting Backbone handle the pagination, the bootstrap table was only able to sort the set of data on the current page instead of the expected full collection. This PR removes the Backbone paginator in favor of using the bootstrap table's own (thanks @caseypt for figuring this out!) :

screen shot 2016-11-02 at 1 31 46 pm

### Testing

Pull this branch and follow the steps in the `README` to update the dependencies/rebundle the vendor.

Then, `./scripts/bundle.sh` and...
- Select an AoI in the DRB about county sized. Confirm in the catchment water quality/point source analyze tabs you can
  - go between pages smoothly
  - render time is still dow
  - the column sort sorts the whole set
- Select an small AoI with <50 water quality rows and confirm no pagination shows up

Connects #1594

### DIFF
--- a/src/mmw/js/src/core/models.js
+++ b/src/mmw/js/src/core/models.js
@@ -239,14 +239,10 @@ var AnimalCensusCollection = Backbone.Collection.extend({
 
 var PointSourceCensusCollection = Backbone.PageableCollection.extend({
     comparator: 'city',
-    mode: 'client',
-    state: { pageSize: 50, firstPage: 1 }
 });
 
 var CatchmentWaterQualityCensusCollection = Backbone.PageableCollection.extend({
     comparator: 'nord',
-    mode: 'client',
-    state: { pageSize: 50, firstPage: 1 }
 });
 
 var GeoModel = Backbone.Model.extend({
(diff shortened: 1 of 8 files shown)


src/mmw/js/src/core/models.js
```diff
@@ -239,14 +239,10 @@ var AnimalCensusCollection = Backbone.Collection.extend({
 
 var PointSourceCensusCollection = Backbone.PageableCollection.extend({
```
mmcfarland: These `Backbone.PageableCollection` calls are throwing errors now since the library has been removed.

arottersman: Sorry, just pushed a fixup.

mmcfarland:
Unfortunately, this seem to have extremely long rendering times for large watersheds.  On this branch, a watershed with ~2000 rows takes about 10 seconds to transition to model & between tabs on model.  An equivalent watershed on production (with the previous paging) takes just a second or two.

For reference, I'm using HUC-8 Subbasin: Schuylkill (the bunny rabbit HUC)

screenshot from 2016-11-02 14 21 07

arottersman:
Closing this until we either...
- fix the underlying re-rendering the analyze tabs in the Model view
- actually need to fix the sorting

Issue #1590, opened 2016-11-01
Open New Scenario tab by default


ajrobbins:
Currently, clicking the "model" button from analyze opens the "Current Conditions" tab in the modeling interface:
![screen shot 2016-11-01 at 3 56 40 pm](https://cloud.githubusercontent.com/assets/12868342/19905137/d913b5ac-a04b-11e6-9a16-ef57f80a5cd9.png)

This is confusing because users don't see the land cover and conservation dropdowns, which are on the New Scenario tab. We should land users on the New Scenario tab by default:

![screen shot 2016-11-01 at 3 56 50 pm](https://cloud.githubusercontent.com/assets/12868342/19905142/df18723a-a04b-11e6-9e67-8f174d945c75.png)

rajadain:
For history, see #364 and #628

arottersman:
Shares a point with...
https://github.com/azavea/pwd-stormdrain-marking/issues/246
https://github.com/WikiWatershed/model-my-watershed/issues/1595
https://github.com/WikiWatershed/model-my-watershed/issues/1577

PR #1598, opened 2016-11-02
Make New Scenario Model Open On Model Start up

arottersman:
Sets the initial `active` flag on the New Scenario scenario instead of current conditions when the user enters the Model view.

screen shot 2016-11-02 at 3 13 23 pm

## Testing
- Pull this branch
- `./scripts/bundle.sh`
- Select an AoI -> Model (either one)
- Note that the initial active scenario tab is "New Scenario"

Connects #1590


kellyi:
+1: tab highlighted and open on start:

screen shot 2016-11-02 at 5 57 08 pm

Issue #1603, opened 2016-11-03
Point source table not showing full AOI totals


ajrobbins:
The "total for area of interest" row appears to only be totaling that page of the table:

![screen shot 2016-11-03 at 2 21 23 pm](https://cloud.githubusercontent.com/assets/12868342/19979535/34657892-a1d1-11e6-974c-9f461eefe8fe.png)
![screen shot 2016-11-03 at 2 21 29 pm](https://cloud.githubusercontent.com/assets/12868342/19979538/35bf8516-a1d1-11e6-8334-622ad9aa8a48.png)

Though it appears to be working correctly for water quality.

PR #1592, opened 2016-11-01
Fixup Analyze Tab Paginated Tables

arottersman:
- Show footer with info for whole collection
- Show number of pages
- Disable pagination if no (more) pages
- Show tooltip on hover of arrow pagination button

screen shot 2016-11-01 at 5 20 24 pm
screen shot 2016-11-01 at 5 20 34 pm

## Testing

Pull this branch
In the DRB,
- Select a 1km AoI (should have only 1 page) - confirm there are no pagination buttons on the Water Quality tab
- Select a big AoI (should have multiple pages) -
On the water quality tab
-  confirm there is no back button on the first page
-  no forward button on the last
-  tooltip on button hover
-  totals don't change as you flip page to page

Connects #1585

PR #1606, opened 2016-11-07
Show Total Averages For Whole Table On Point Source Analyze Tab

arottersman:
While working on #1592 I addressed showing an AOI's total averages on the Water Quality analysis tab, and forgot that we have to do the same for the Point Source tab.

With the `Backbone.paginator` `this.collection.models` refers to all the models of the page, and to get all the models in the collection you have to use `this.collection.fullCollection.models`

## Testing Instructions
Pull this branch and rebundle: `./scripts/bundle.sh`
On `localhost:8000`: 
- Select By Boundary -> HUC-8 Subbasin -> Lower Susquehanna
- Go to the Point Sources analyze tab (the table should be two pages long)
- Flip between pages and confirm none of the total averages change between pages (ie they are the totals for the whole AOI
screen shot 2016-11-07 at 9 59 35 am
screen shot 2016-11-07 at 9 59 40 am

Connects #1603

### DIFF
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -419,11 +419,11 @@ var PointSourceTableView = Marionette.CompositeView.extend({
         return {
             headerUnits: this.options.units,
             totalMGD: utils.totalForPointSourceCollection(
-                this.collection.models, 'mgd'),
+                this.collection.fullCollection.models, 'mgd'),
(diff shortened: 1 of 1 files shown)


caseypt:
Consider updating the commit message with the explanation you added to the PR.

arottersman:
Thanks! I've updated the commit message.

Issue #1607, opened 2016-11-07
Smarter caching of JS/CSS assets


mmcfarland:
*NSF feature request*

Presently, the app uses a combination of headers (`Cache-Control`, `ETag`, `Last Modified` and `Expires`) to control the caching of JS & CSS assets (vendor and bundle).  In practice, after a deployment the user must manually refresh the browser several times to see consistent changes - and if that doesn't happen, they will often report a bug, the solution to which is to refresh the browser.

I do see `If-None-Matched` getting sent on the request and a `304` being returned as result.  Investigate if other cache headers are conflicting with the intended behavior, or if we should implement more drastic features like adding a hash to the built assets and having the references in `index` templated at build time.

hectcastro:
Is it easy enough to append a checksum of the static asset bundles? That would allow us to continue to be aggressive on the caching front, but also avoid staleness issues on the frontend.

lewfish:
It wouldn't be too difficult. I just think we should go with the builtin etag mechanism which checks the server to see if it the content needs to be sent, unless the network overhead of the check is really a problem.

PR #1609, opened 2016-11-09
Turn off time-based caching on production

lewfish:
The production nginx configuration template makes static assets have a cache that expires way in the future due to the `expires max;` line. When static assets are updated, and the website is loaded without a refresh, the client never bothers sending a request to see if the etag is out of date, and doesn't get the update. This PR preserves etag based caching and reduces the expires timeout to 1hr.

It may be helpful to read https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching

#### Notes
* We might still want to set an expiration time for the static assets, but just make it shorter. We might also want to have a different cache policy for things like images and fonts which don't change as much. @mmcfarland 

#### Testing
* Stay on the `develop` branch.
* Run `vagrant ssh app` and `sudo vi /etc/nginx/sites-available/mmw-app.conf`
* Remove the `{% if ['packer'] | is_in(group_names) -%}` block. This is to simulate the production environment.
* Run `sudo restart nginx` 
* Go to `localhost:8000`
* Make change to some javascript file (with bundler running) and then go to website again without hitting refresh.
* The updates will not appear (you can also see in the network tab that the files are from the cache).
* Hit reload. The changes should appear.
* Edit the nginx file so that it doesn't have the `expires max;` line. Run `sudo restart nginx`. In reality the server will be using `expires 1h;` but you don't want to wait 1hr for the cache to expire while testing this.
* Go to `localhost:8000` and then hit reload. Repeat the above routine with making a change to a file. Now you shouldn't need to reload again in order to get the change because the browser will check the etag and see that it's invalid and send the new file.

Connects #1607

### DIFF
--- a/deployment/ansible/roles/model-my-watershed.app/templates/nginx-app.conf.j2
+++ b/deployment/ansible/roles/model-my-watershed.app/templates/nginx-app.conf.j2
@@ -31,7 +31,7 @@ server {
   location /static/ {
     {% if ['packer'] | is_in(group_names) -%}
     etag on;
-    expires max;
+    expires 1h;
     {% endif %}
 
     alias {{ app_static_root }};


mmcfarland:
My first instinct was to say we should maybe have more specific location stanzas for the paths that we wanted to drop the max age header from, but on inspection of the `/static` directory, there's actually little else there - and it remains behind an E-Tag header.  I think a fair tradeoff may be to leave the max age header, but reduce its time to `1h` (like `/media`) so that during a typical session, the requests aren't made on refresh, but in between sessions it would pick up file changes via E-Tag.

lewfish:
> think a fair tradeoff may be to leave the max age header, but reduce its time to 1h (like /media) so that during a typical session, the requests aren't made on refresh, but in between sessions it would pick up file changes via E-Tag.

Ok, made the update.

Issue #1616, opened 2016-11-18
Add parameterized ASG scheduled actions


hectcastro:
Add scheduled actions for the following ASGs:

- `Application`
- `Tiler`
- `Worker`

For staging, configure the windows to go down to zero outside of 8AM-6PM ET. For production, scale up at 7AM and back down at 6PM ET.

PR #1617, opened 2016-11-18
Add ASG scheduled actions for all services

hectcastro:
For each of the following services, add scheduled action resources that update the autoscaling group's capacity based on a cron-like schedule:

- `App`
- `Tiler`
- `Worker`

The idea is that parameters like `AppServerAutoScalingScheduleStartRecurrence` can be used to change an ASGs capacity at the beginning of a day to `AppServerAutoScalingScheduleStartCapacity`. Likewise, that change can be undone at the end of the day with a schedule defined via `AppServerAutoScalingScheduleEndRecurrence` and `AppServerAutoScalingScheduleEndCapacity`.

Resolves https://github.com/WikiWatershed/model-my-watershed/issues/1616

---

**Testing**

- Review staging to ensure that the scheduled actions on each ASG align with the settings in `default.yml.example
- Review the **Activity History** for each ASG and ensure that it scaled down and stayed down on Sunday, but then scaled up again this morning


deployment/default.yml.example
```diff
@@ -35,6 +41,12 @@ TileServerInstanceProfile: 'TileServerInstanceProfile'
 TileServerAutoScalingDesired: '1'
 TileServerAutoScalingMin: '1'
 TileServerAutoScalingMax: '1'
+TileServerAutoScalingScheduleStartCapacity: '1'
+# 8AM ET
+TileServerAutoScalingScheduleStartRecurrence: '0 13 * * *'
```
hectcastro: This needs to take into account the weekend.

Issue #1618, opened 2016-11-21
Add Stroud center logo to header bar


ajrobbins:
Per Stroud: 
>Add a small Stroud Center logo to the black bar at the top of every page of the MMW app

All-white logo is attached (delete the .txt extension). May need @jfrankl consult to see if this looks ok/take it out of AI format

[stroud_logo_white_text_white_waves.ai.txt](https://github.com/WikiWatershed/model-my-watershed/files/604388/stroud_logo_white_text_white_waves.ai.txt)

jfrankl:
Please use this png with transparency: http://sandbox.azavea.com/stroud-logo.png

Please apply the following css to the image, through a class or other means.

```css
height: 24px;
position: relative;
top: -2px;
```

caseypt:
The link to the image above doesn't resolve anymore, so I created a new version of the AI file, which is attached here for reference.

![stroud-logo](https://cloud.githubusercontent.com/assets/1042475/20633657/122794ac-b317-11e6-8700-29232f7216c6.png)

PR #1622, opened 2016-11-25
Add Stroud logo to header

caseypt:
## Overview

Per Stroud, their logo is added to the app header. This PR also removes some images that weren't used and adds the source image directory to the asset pipeline. Prior to this PR, we hadn't used any original images in the app.

### Demo

image

## Testing Instructions

- Proceed through the app and verify that the logo is always visible in the header.

Connects #1618


caseypt:
@jfrankl I had to drop `float: left;` from the header link to get the image to display properly. This didn't appear to cause any problems.

arottersman:
Tested, and the logo is there.

I wonder if it should be clickable like the header label? (although now that I click on the header label, I realize it doesn't do much...)

Also, should we add a `margin-right` on the icon?
screen shot 2016-11-28 at 11 26 29 am

jfrankl:
Thanks for getting the ball rolling. Can you please move the logo to the right side of the bar, next to WikiWatershed? The top left should be reserved for the MMW identity.

image

caseypt:
Pushed a fixup commit to move the logo to the right, link to Stroud's site, and to update the registration page header to have better parity with the app header.

image

image

arottersman:
+1, nice catch on the reset password page!

caseypt:
Thanks!

Issue #1625, opened 2016-11-28
Total N value is low


ajrobbins:
In the Water Quality tab, the values for "Total N (kg/ha) differ from the `tn_tot_kgy` field in the shapefile. I'm thinking it might be a unit conversion problem?

For example, check out the value for catchment #1994, located just southwest of center city, west of the Schuylkill:

MMW - 
![screen shot 2016-11-28 at 5 39 16 pm](https://cloud.githubusercontent.com/assets/12868342/20690533/2b9f3186-b599-11e6-864c-ddaf8c5cd5f7.png)
![total_n](https://cloud.githubusercontent.com/assets/12868342/20690561/5602fc1e-b599-11e6-84d2-b699a5347f35.png)

final_lateralshed_11_28_16 file - 
![screen shot 2016-11-28 at 6 37 00 pm](https://cloud.githubusercontent.com/assets/12868342/20690639/cfcacb6c-b599-11e6-84fd-94d6e3e2bf7a.png)

jtarricone:
It's because the displayed value is on a per-hectare basis. "Total N" might not be the clearest name but the units indicate how the value is derived.

jtarricone:
It also may be notable to hydrologists that the basis of measurement is kilogram-years, I can't speak to that.

ajrobbins:
I thought the units in both cases (source data and the data we are displaying) were kg/yr/ha. Are we using the `tn_tot_kgy` field for our Total N field in the app? If so, are we doing any conversions on it?

jtarricone:
The app is dividing `tn_tot_kgy` by `areaha`, which presumes that the UoM of `tn_tot_kgy` is kg-years.

mmcfarland:
I agree that it's working as requested - we're dividing kg by area.  They requested results to be shown as kg/ha.  It could be that the source data `tn_tot_kyy` is not the unit the name implies, but the conversion looks correct based on the labels.

jtarricone:
Based on this --> http://ec.europa.eu/eurostat/statistics-explained/index.php/Agri-environmental_indicator_-_gross_nitrogen_balance

It seems like the UoM of `tn_tot_kgy` may actually be kg-yr/HA (or kg / HA /year, assuming equivalence), since it looks like the standard way of measuring nitrogen surplus.

mmcfarland:
[...]

PR #1627, opened 2016-11-30
Fix calculation of displayed water quality data

jtarricone:
This fixes an error in the front-end that resulted in an unnecessary UoM adjustment for the water
quality totals displayed. The data were already represented in units of kg-year / hectare, so the
app did a superfluous conversion.

The prior version -->
download 8

This version -->
download 9

To test:
* open both the local version and a staging/prod version in a browser
* pick a similar/identical AoI in both, and compare the water quality totals; the local version should show values several times (the value of the "Area" field, specifically) larger than the staging/prod version

Connects #1625

### DIFF
--- a/src/mmw/js/src/analyze/templates/catchmentWaterQualityTable.html
+++ b/src/mmw/js/src/analyze/templates/catchmentWaterQualityTable.html
@@ -27,29 +27,6 @@
       <tbody>
       </tbody>
       <tfoot>
-          <tr class="catchment-water-quality-table-footer">
-              <td class="strong text-right" colspan="2">
-                  Total For Area of Interest (kg/ha)
-              </td>
-              <td class="catchment-water-quality-footer-item strong text-right">
-                  {{ totalTN|filterNoData()|toLocaleString(3) }}
-              </td>
-              <td class="catchment-water-quality-footer-item strong text-right">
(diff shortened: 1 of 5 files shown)


arottersman:
We are also dividing the total for the collection by area: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/utils.js#L295

That's not correct either, right?

jtarricone:
Good catch.

arottersman:
+1, looks right to me

Table Totals are updated now too:

screen shot 2016-11-30 at 1 57 24 pm

screen shot 2016-11-30 at 1 57 33 pm

arottersman:
And nice catch remembering the tests!

ajrobbins:
@jtarricone We recently got a request from Stroud to remove the totals row, actually - summing loading rates and averages doesn't make a ton of sense. Would you be able to take it out without a lot of extra work?

jtarricone:
It shouldn't be a problem.

jtarricone:
I just pushed an update that removes the aggregate row and the corollary parts of the view and utility files.

arottersman:
I'll take a peek - still have it set up.

arottersman:
It is indeed gone. 
screen shot 2016-11-30 at 3 15 12 pm

PR #1646, opened 2016-12-21
Support Anonymous ITSI Sessions

rajadain:
## Overview

The original ITSI embed implementation assumed that it would only ever be used in a logged in environment, prompting design decisions to automatically log in users with their ITSI credentials if they are in that environment. This assumption is no longer valid as there are now anonymous users that use / preview the app without logging in first. They are taken to the ITSI login page when activating the app and can not access it at all.

This PR makes the following changes to allow anonymous users to use the app when embedded inside ITSI:

 * Users are no longer made to log in with their ITSI credentials on start.
 * Users are no longer shown the log in popup on start.
 * The app asks the parent frame for login status using the LARA Interactive API. If the response indicates that the user is logged in, we trigger the ITSI login procedure. If they already have an account in the app, they will be logged in as that. If they don't, a new one will be created for them.
 * We defer resuming old sessions until the user is logged in to the app, since without logging in they may not have access to the project in the old session.

There was initial concern about the time it would take to communicate with LARA, and if the user starts working before that is done they may lose that work. Initial testing indicates that communication to be pretty snappy, so we may not have anything to worry about.

## Testing Instructions

 * Pull the branch and bundle the scripts.
 * Go to [http://localhost:8000/?itsi_embed=true](http://localhost:8000/?itsi_embed=true). Ensure that you are not redirected to log in via ITSI, and that the login modal doesn't show up automatically.
 * Expose your local on `ngrok` and follow ITSI embed setup instructions from #675.
 * Ensure that you aren't auto logged in while editing the activity as `mmwterence`.
 * Log in as `tstudent2` to ITSI, go to the activity. Ensure that you are logged in automatically, and that the session resumes to the modeling stage of some pre-saved project.

Connects #1555

### DIFF
--- a/src/mmw/apps/user/middleware.py
+++ b/src/mmw/apps/user/middleware.py
@@ -1,34 +1,18 @@
 # -*- coding: utf-8 -*-
 
 from django.conf import settings
-from django.core.urlresolvers import reverse
-from django.shortcuts import redirect
 
-from apps.user.views import itsi_login
 
 EMBED_FLAG = settings.ITSI['embed_flag']
-LOGIN_URL = reverse(itsi_login)
 
 
 class ItsiAuthenticationMiddleware(object):
     """
-    Middleware for automatically logging in ITSI users
-    and setting relevant flags.
+    Middleware for setting relevant ITSI flags.
     """
 
     def process_request(self, request):
-        """
-        Check if ITSI EMBED FLAG is set, and if so attempt to log
-        the user in with their ITSI credentials and redirect them
-        to current page.
-        """
-
-        # If flag is not set then return None so Django can proceed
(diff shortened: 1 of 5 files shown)


arottersman:
Taking a look now!

rajadain:
Thanks for taking a look. The only really tricky part is when you visit the site in embed mode, and then by itself. Since the `itsi_embed` flag is set as a session variable, it persists, which can be confusing. You usually have to log out and log back in to clear it (that also clears the entire session). But the only way around this would be to clear the flag if not present in the URL, which defeats the purpose of having it entirely (otherwise we would always just use the query string). But this seems to be an extremely rare (if ever) use case, so we're okay.

Issue #1639, opened 2016-12-09
Snapshot has offset map in some cases


scytacki:
You can test this by
1. logging into the [ITSI portal](https://itsi.portal.concord.org). 
2. Then open this URL: https://authoring.concord.org/activities/5644
3. Scroll down until you see the MMW interactive.
4. Click the interactive to open it
5. Change the Basemap to "Satellite with Roads"
6. Select a square kilometer somewhere not in the center of the map.
7. Take a snapshot.

You should see something like this:
![screen shot 2016-12-09 at 12 39 57 pm](https://cloud.githubusercontent.com/assets/86016/21058590/abf5bfd2-be0c-11e6-9da0-93898d004113.png)

Look in the "Editing snapshot" window. Notice the grey bar on the left and notice the map tiles are shifted right and no longer line up with the selection box.

If you switch the Basemap to something else. And then switch it back to "Satellite with Roads" then the problem goes away. This type of problem occured before the `itsi_embed=true` parameter was added.  I believe it has to do with leaflet using 3d css transforms to move the tiles around.  The snapshot service doesn't handle 3d css transforms. So `itsi_embed=true` is supposed to disable these transforms.  It seems the "Satellite with Roads" is still using them.

mmcfarland:
Related to #1640, most likely same issue.  Shares points.

Issue #1640, opened 2016-12-09
Firefox has problem snapshotting "Satellite with Roads" basemap


scytacki:
To reproduce
1. In Firefox, log into the [ITSI portal](https://itsi.portal.concord.org)
2. Then open this URL: https://authoring.concord.org/activities/5644
3. Scroll down until you see the MMW interactive.
4. Click the interactive to open it
5. Change the Basemap to "Satellite with Roads"
6. Draw a Square Km Area
7. Take a snapshot (in the top right)

You should see something like:
![screen shot 2016-12-09 at 12 51 02 pm](https://cloud.githubusercontent.com/assets/86016/21058926/3233d9e8-be0e-11e6-835c-1e9c20df4f37.png)

Look at the image in the Editing Snapshot window.  Notice the bulleted list at the top and the grey area on the right.
If the Basemap is set to "Streets" this doesn't happen. If the Basemap is set to "Streets" and then set back to "Satellite with Roads" then problem goes away.

So this might be the same issue as #1639

ajrobbins:
Shares points with #1639

rajadain:
After some investigation, this is where I am:

I can reproduce this issue by taking a screenshot upon immediate load. The screenshot looks like this:

![image](https://cloud.githubusercontent.com/assets/1430060/21506093/83933460-cc3b-11e6-8156-baabfa268b32.png)

if I open the layer selector and switch to "Satellite" and back to "Streets", the screenshot looks like this:

![image](https://cloud.githubusercontent.com/assets/1430060/21506100/98f2f39a-cc3b-11e6-91c3-d53e2c3ddcfd.png)

These are the respective HTML shutterbug caches of these images: https://ccshutterbugtest.s3.amazonaws.com/347c0facfab.html and https://ccshutterbugtest.s3.amazonaws.com/bfdfd609401.html. (Interestingly neither page opens correctly in Firefox, but both do in Chrome).

If I run PhantomJS with [rasterize.js](https://github.com/ariya/phantomjs/blob/master/examples/rasterize.js) on these HTML pages locally, I get two correctly rendered images:

![firefox-before-347c0facfab](https://cloud.githubusercontent.com/assets/1430060/21506139/d8ccf286-cc3b-11e6-97a4-2a8993f9e30a.png)

![firefox-after-bfdfd609401](https://cloud.githubusercontent.com/assets/1430060/21506144/de07062e-cc3b-11e6-8e52-fd8eb25b6890.png)

[...]

PR #1648, opened 2016-12-22
Fix Google Tile Layer screenshots

rajadain:
## Overview

The Google Maps JavaScript API always uses CSS transforms, which are not supported by PhantomJS. We convert the CSS transform values to CSS left and top values before each screenshot, and restore them immediately afterwards. This allows the screenshot to capture the map accurately.

Got the idea from here: https://github.com/niklasvh/html2canvas/issues/345#issuecomment-71774787

## Testing Instructions

Check out the branch and bundle the scripts. `ngrok` your local app.

 * Open Firefox.
 * Go to https://authoring.concord.org/activities/5644.
 * Open the dev tools, and replace the iframe `src` to your ngrok rather than `app.wikiwatershed.org`. Preserve the `itsi_embed` flag. It should look like something like this: `https://b011c760.ngrok.io?itsi_embed=true`.
 * Take screenshots with all four basemap types. Ensure that they render correctly.

Do the same test in Chrome.

Connects #1639

### DIFF
--- a/src/mmw/js/src/app.js
+++ b/src/mmw/js/src/app.js
@@ -122,18 +122,55 @@ function RestAPI() {
 }
 
 function initializeShutterbug() {
+    var googleTileLayerSelector = '#map > .leaflet-google-layer > div > div > div:nth-child(1) > div:nth-child(1)';
+
     $(window)
         .on('shutterbug-saycheese', function() {
             // Set fixed width before screenshot to constrain width to viewport
             $('#model-output-wrapper, body > .map-container').css({
                 'width': window.innerWidth
             });
+
+            var mapView = App.getMapView(),
+                activeBaseLayer = mapView.baseLayers[mapView.getActiveBaseLayerName()],
+                googleLayerVisible = !!activeBaseLayer._google;
+
+            if (googleLayerVisible) {
+                // Convert Google Maps CSS Transforms to Left / Right
+                var $googleTileLayer = $(googleTileLayerSelector),
+                    transform = $googleTileLayer.css('transform').split(','),
+                    left = parseFloat(transform[4]),
+                    top = parseFloat(transform[5]);
+
+                $googleTileLayer.css({
+                    transform: 'none',
+                    left: left,
+                    top: top,
+                });
+            }
         })
         .on('shutterbug-asyouwere', function() {
             // Reset after screenshot has been taken
             $('#model-output-wrapper, body > .map-container').css({
                 'width': ''
             });
+
+            var mapView = App.getMapView(),
+                activeBaseLayer = mapView.baseLayers[mapView.getActiveBaseLayerName()],
+                googleLayerVisible = !!activeBaseLayer._google;
+
+            if (googleLayerVisible) {
+                var $googleTileLayer = $(googleTileLayerSelector),
+                    left = parseFloat($googleTileLayer.css('left')),
+                    top = parseFloat($googleTileLayer.css('top')),
+                    transform = 'matrix(1, 0, 0, 1, ' + left + ', ' + top + ')';
+
+                $googleTileLayer.css({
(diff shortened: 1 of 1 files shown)


arottersman:
Nice find with this! The initial problem of the offset-by-grey map when screenshotting Satellites with Roads is gone. Unfortunately there are some new weird problems. I think a somewhat reliable way to reproduce the following is to
- take a screenshot with one of the google layers turned on
- reset the aoi + draw another
- take a screenshot with one of the non-google layers

screen shot 2016-12-22 at 10 27 51 am

Some other bugs I haven't been able to consistently reproduce but come from generally switching on and off layers, redrawing areas and taking screenshots:

screen shot 2016-12-22 at 10 53 48 am

No auto-center to AOI
screen shot 2016-12-22 at 10 29 03 am

src/mmw/js/src/app.js
```diff
@@ -122,18 +122,41 @@ function RestAPI() {
...
+    var googleTileLayerSelector = '#map > .leaflet-google-layer > div > div > div:nth-child(1) > div:nth-child(1)';
+
     $(window)
         .on('shutterbug-saycheese', function() {
             // Set fixed width before screenshot to constrain width to viewport
             $('#model-output-wrapper, body > .map-container').css({
                 'width': window.innerWidth
             });
+            // Convert Google Maps CSS Transforms to Left / Right
+            var $googleTileLayer = $(googleTileLayerSelector),
```
arottersman: Should this only happen if a google tilelayer is on?

rajadain: The `googleTileLayerSelector` is tuned to select only Google tile layers. If they are not visible, it returns an empty array, and none of the following operations will have any effect.

arottersman: Ah, okay, that's what I'd expect

rajadain:
I'm not quite sure why, but adding a guard for only doing the switch when the Google tile layer is visible seems to have worked. This is ready for another look.

src/mmw/js/src/app.js
```diff
@@ -122,18 +122,55 @@ function RestAPI() {
...
+            }
         })
         .on('shutterbug-asyouwere', function() {
             // Reset after screenshot has been taken
             $('#model-output-wrapper, body > .map-container').css({
                 'width': ''
             });
+
+            var activeBaseLayer = App.getMapView().getActiveBaseLayerName(),
+                googleLayerVisible = activeBaseLayer === 'Satellite with Roads' ||
```
arottersman: Was going to write "Consider using the[ `type` on the layer](https://github.com/WikiWatershed/model-my-watershed/blob/f2c0cbe8177280923a7745da6335c3d07a91b2be/src/mmw/mmw/settings/layer_settings.py#L244)", but when I tried to access it in the console, it looked like it was only available on the mapbox and esri layers. 

Another way we could test if a basemap is google though is to do : 
```
var mapView = App.getMapView(),
      activeLayer = mapView.baseLayers[mapView.getActiveBaseLayerName()],
      googleLayerVisible = !!(activeLayer._google)
```

rajadain: Good idea, will do

arottersman:
The current state of affairs: 
Screenshots are correct when they work, but on firefox they sometimes end up like this: 
screen shot 2016-12-22 at 5 03 09 pm

To reproduce:
1.  Turn one of the google layers on
2. Draw an AOI and take a screenshot in Analyze view
3.  Repeat 1 and 2 for the Satellite layer

rajadain:
Good job testing. I think we can keep the stuff in this PR for #1639, and investigate #1640 separately.

rajadain:
I've squashed the fixups. I'm in the mood to merge this and look at the Firefox issues independently.

Issue #1662, opened 2017-01-16
Display current lat lng on map


kdeloach:
Create a new map control which displays the current lat lng under the cursor for debugging purposes.

rajadain:
I think it would be kinda cool if this was inside a function that we could trigger. So it wouldn't show up automatically, but if we were to open the console and type in `MMW.debug();` it would. This would allow us to debug even production.

kdeloach:
While implementing this, I realized that it might make more sense to just attach the leaflet map to `window`. Then you could easily call `getCenter` or `setView` etc. in staging/production.

rajadain:
We could also do

```js
var map = MMW.getLeafletMap();
map.getCenter();
```

Since `MMW` is attached to `window` already. But references are cheap, and making debugging easier is useful.

kdeloach:
Oh, I didn't realize that. Very cool.

PR #1663, opened 2017-01-16
Display current lat lng on map

kdeloach:
This adds a control in the corner of the map to display the current lat lng for debugging purposes. Clicking on the map will also log the current lat lng in the console.

image

Connects #1662

### DIFF
--- a/src/mmw/js/src/core/latLngControl.js
+++ b/src/mmw/js/src/core/latLngControl.js
@@ -1,0 +1,50 @@
+'use strict';
+
+var L = require('leaflet');
+
+var PRECISION = 5;
+
+module.exports = L.Control.extend({
+    options: {
+        position: 'bottomleft'
+    },
+
+    onAdd: function(map) {
+        var el = L.DomUtil.create('span');
+        el.className = 'leaflet-latlng-control';
+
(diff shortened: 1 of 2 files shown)


kdeloach:
@mmcfarland @ajrobbins @jfrankl :+1:  / :-1: ?

ajrobbins:
To clarify: this only happens when a point is selected for RWD, right? It seems really useful for debugging on staging, but I'm not sure it should end up on production.

kdeloach:
Right now it's always on. I could update it to only show while dragging a marker (in draw mode). Or only enable it in debug mode (staging).

ajrobbins:
I meant, that is is the lat/lon of the clicked RWD point, right? It doesn't show when you draw a shape or select by boundary, right?

kdeloach:
Demo: http://algiers:8000/

ajrobbins:
I see! I was misunderstanding. I think this should be for staging only. It would be helpful for debugging, but kind of distracting in production. @mmcfarland ?

mmcfarland:
Rather than wire it into an environment, let's add this now for debugging and revert the commit when we're done having external folks debug.

Issue #1658, opened 2017-01-12
RWD Feedback


mmcfarland:
* Help debug RWD issues on staging
* Rename mid-atlantic RWD region
This should probably read "National Medium Resolution" and be the first in the list.

![screenshot 2017-01-12 at 2 28 00 pm](https://cloud.githubusercontent.com/assets/1014341/21904947/7d0c57bc-d8d3-11e6-96ab-2e13d98801b5.png)

@SRGDamia1 or @aufdenkampe, do you have input into what the best label for National RWD is (to replace Mid-Atlantic)?

aufdenkampe:
In his 12/3/2016 email response to the "MMW Release 1.17" thread, @dtarb suggested this:

> I suggest the following wording:
>  
> In the overlay labels
> -          Delaware high resolution stream network
> -          Continental US medium resolution stream network
>  
> In the delineate watershed tool
> -          Snap to Delaware high resolution streams
> -          Snap to Continental US medium resolution streams
>  
> In the help text
> This moves downhill from the point you click to snap onto the nearest point on the Delaware high resolution stream network and calculates the watershed upstream of this point using a 1/3 arc second (10) m resolution digital elevation model for the Delaware River Basin.  The stream network snapped to was delineated using TauDEM (http://hydrology.usu.edu/taudem).
>  
> This moves downhill from the point you click to snap onto the nearest point on the medium resolution National Hydrography Dataset (NHDPlus flowline) and calculates the watershed upstream of this point using the 30 m resolution NHDPlus flow direction grid for the continental US.  If the point you click does not have a NHDPlus flowline downstream of it (e.g. is in an internally draining area) the watershed is calculated from the point you click.  Watersheds calculated may have holes in them due to internally draining areas that do not connect to the outlet in the NHDPlus flow direction dataset.  To learn more about NHDPlus see  http://www.horizon-systems.com/nhdplus/.  The model my watershed watershed delineation uses NHDPlus version 2.1 data model with latest content version accessed 11/22/16.
>  
> Rationale
[...]

PR #1664, opened 2017-01-17
Rename NHD watershed delineation dropdown item

kdeloach:
Connects #1658


ajrobbins:
[Just from looking at the diff] - beautiful

Issue #1657, opened 2017-01-12
Allow viewing large watersheds that cannot be analyzed


mmcfarland:
With the recent addition of National RWD, it is possible to generate watershed which are too large to be analyzed or modeled.  Currently, it results in a warning and then removes the polygon in the map.  The behavior should instead be that the polygon is viewable, but the user is unable to move to the analyze step.  The message should still be displayed, somehow.  Consult design to agree on a conceptual workflow.

kdeloach:
We should also gracefully handle errors if watershed is null.

![image](https://cloud.githubusercontent.com/assets/43062/21994219/95c0a3f0-dbec-11e6-9d3e-cbc2aa24fd84.png)

![image](https://cloud.githubusercontent.com/assets/43062/21994200/8211d77a-dbec-11e6-9be2-205875279954.png)

PR #1669, opened 2017-01-18
Allow display of very large and invalid watersheds

kdeloach:
#### Changes
* Graceful RWD error handling, and better messages
* Display invalid watersheds
* Hide "Analyze" button for invalid watersheds

#### Demo

We now handle errors much more consistently instead of having the UI stuck in a bad state.

image

For invalid (multipolygon) watersheds, we now display a better error message. Since we can't calculate the area, it is hidden in the analyze footer.

image

For valid watersheds that are too large (I have set the limit to 5km^2 for testing purposes) we display the generated watershed along with a warning, and hide the "Analyze" button in the footer.

image

Connects #1657

### DIFF
--- a/src/mmw/js/src/core/models.js
+++ b/src/mmw/js/src/core/models.js
@@ -4,7 +4,9 @@ var Backbone = require('../../shim/backbone'),
     $ = require('jquery'),
     _ = require('lodash'),
     turfArea = require('turf-area'),
-    utils = require('./utils');
+    utils = require('./utils'),
+    drawUtils = require('../draw/utils');
+
 
 var MapModel = Backbone.Model.extend({
     defaults: {
@@ -257,6 +259,7 @@ var GeoModel = Backbone.Model.extend({
         shape: null,        // GeoJSON
         area: '0',
         units: 'm<sup>2</sup>',
+        isValidForAnalysis: true
     },
 
     initialize: function() {
@@ -266,6 +269,7 @@ var GeoModel = Backbone.Model.extend({
 
     update: function() {
         this.setDisplayArea();
+        this.setValidForAnalysis();
     },
 
     setDisplayArea: function(shapeAttr, areaAttr, unitsAttr) {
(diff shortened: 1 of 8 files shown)


kdeloach:
@ajrobbins Feedback on text changes?

For very large watersheds:

image

For multipolygons (analyze endpoints do not support these):

image

mmcfarland:
A bit of duplication in the first and last sentence, maybe combine to:

> Sorry, your Area of Interest is too large to analyze or model.

Then the second sentence as you have it, presently.

I also wonder if we should use the term Area of Interest or "Watershed".

kdeloach:
There may appear to be a lot more changes than necessary, but I encountered several bugs with syncing state between the views and models that have gone unnoticed for some time. I'm looking into ways to reduce the number of changes needed for this feature.

ajrobbins:
If this message is only for RWD, I'd say:

```
Sorry, the watershed you have delineated is too large to analyze or model. xxx km2 were selected, but the maximum supported size is currently yyy km2.
```

Second one looks good!

kdeloach:
> I also wonder if we should use the term Area of Interest or "Watershed".

I think the first error message is used on both RWD and Free Draw mode.

ajrobbins:
>I think the first error message is used on both RWD and Free Draw mode.
If that's the case, I think we should say "the area you have delineated" instead of "Area of Interest" which is not super familiar to people.

kdeloach:
I will remove b7bd983fe4ac4bdc6dd29f91d30dd4ead5c86bbb before merging.

kdeloach:
Updated text and added screenshots.

kdeloach:
Fixing lint and tests now.

ajrobbins:
+1 on the new messages!

kdeloach:
Unit tests pass successfully for me, but fail on Jenkins.

image

kdeloach:
I was able to reproduce these test failures by upgrading my version of Firefox.

kdeloach:
Fixed unit tests.

PR #1681, opened 2017-02-01
Update RWD

kdeloach:
### Overview

Changes:
* Upgrade to latest version of RWD
* Add script to warm up EBS volume

### DIFF
--- a/deployment/cfn/worker.py
+++ b/deployment/cfn/worker.py
@@ -417,7 +417,10 @@ class Worker(StackNode):
                 '  - path: /etc/mmw.d/env/ROLLBAR_SERVER_SIDE_ACCESS_TOKEN\n',
                 '    permissions: 0750\n',
                 '    owner: root:mmw\n',
-                '    content: ', self.get_input('RollbarServerSideAccessToken')]  # NOQA
+                '    content: ', self.get_input('RollbarServerSideAccessToken'),  # NOQA
+                '\n',
+                'runcmd:\n',
+                '  - /opt/model-my-watershed/scripts/aws/ebs-warmer.sh']
 
     def create_cloud_watch_resources(self, worker_auto_scaling_group):
         self.add_resource(cw.Alarm(
--- a/scripts/aws/ebs-warmer.sh
+++ b/scripts/aws/ebs-warmer.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+# Ref: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-initialize.html
+
+function log() {
+    echo "[$(date --rfc-3339=seconds)] $1"
+}
+
+log "Warming up RWD EBS volume..."
+
+find /opt/rwd-data/nhd/Main_Watershed/* -type f -print0 \
+    | xargs -0 -P0 -L1 -t cat >/dev/null
+
+find /opt/rwd-data/drb/Main_Watershed/* -type f -print0 \
+    | xargs -0 -P0 -L1 -t cat >/dev/null
+
+log "Done"
--- a/deployment/ansible/roles/model-my-watershed.rwd/defaults/main.yml
+++ b/deployment/ansible/roles/model-my-watershed.rwd/defaults/main.yml
@@ -2,7 +2,7 @@
 rwd_data_path: "/opt/rwd-data"
 rwd_host: "localhost"
 rwd_port: 5000
-rwd_docker_image: "quay.io/wikiwatershed/rwd:1.2.0"
+rwd_docker_image: "quay.io/wikiwatershed/rwd:1.2.1"
 
 app_config:
     RWD_HOST: "{{ rwd_host }}"
--- a/deployment/packer/template.js
+++ b/deployment/packer/template.js
@@ -63,7 +63,7 @@
             "ami_block_device_mappings": [
                 {
                     "device_name": "/dev/sdf",
-                    "snapshot_id": "snap-08b88a0fe2960cc3f",
+                    "snapshot_id": "snap-05eb65ba5d3f6151b",
                     "volume_type": "gp2",
                     "delete_on_termination": true
                 }


scripts/ebs-warmer.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# Source: https://gist.github.com/muhqu/3293988
+# Ref: https://forums.aws.amazon.com/message.jspa?messageID=198413
+
+if [ -z "$1" ]; then
+    echo "Usage: sudo $0 /dev/sdh1"
+    exit 1;
+fi
+
+dd if=$1 of=/dev/null & pid=$!
```
mmcfarland: The [AWS docs recommend](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-initialize.html) adding a blocksize argument at 1M `bs=1M`

kdeloach: Good catch. By the way, I decided to use this helper script instead of just running `dd` inside the cloud config, so that we can monitor the progress and benchmark how long the process takes.

kdeloach:
This is ready for testing.

deployment/cfn/worker.py
```diff
@@ -417,7 +417,10 @@ def get_cloud_config(self):
                 '  - path: /etc/mmw.d/env/ROLLBAR_SERVER_SIDE_ACCESS_TOKEN\n',
                 '    permissions: 0750\n',
                 '    owner: root:mmw\n',
-                '    content: ', self.get_input('RollbarServerSideAccessToken')]  # NOQA
+                '    content: ', self.get_input('RollbarServerSideAccessToken'),  # NOQA
+                '\n',
+                'runcmd:\n',
+                '  - /opt/model-my-watershed/scripts/ebs-warmer.sh /dev/xvdf']
```
kdeloach: We should monitor the output of this command during testing to get a benchmark for how long this process takes.

kdeloach: The logs should be located here: `/var/log/cloud-init-output.log`

scripts/ebs-warmer.sh
```diff
@@ -0,0 +1,20 @@
...
+# Source: https://gist.github.com/muhqu/3293988
+# Ref: https://forums.aws.amazon.com/message.jspa?messageID=198413
+# Ref: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-initialize.html
+
+if [ -z "$1" ]; then
+    echo "Usage: sudo $0 /dev/sdh1"
+    exit 1;
+fi
+
+dd if=$1 of=/dev/null bs=1M & pid=$!
```
hectcastro: This appears to be making progress at ~3.5MB/s on a fresh EBS volume. At that rate, it might be better to try and identify the subset of files that RWD uses and `cat` those to `/dev/null`. That'll force just those blocks to be warm.

Matt's comments on which files to target:

> That seems like it might be preferable.  There are a few large files that exist in `Main_Watershed` that all RWD requests are going to hit, then the branch out into a selection of several 10,000s directories within `Subwatershed_All` depending on the geographic point.  The contents of those files are relatively small, and can probably tolerate not being warmed.  In aggregate their a lot of data, but in practice I don't think an individual call hits many of them.

kdeloach: Updated.

kdeloach:
Changed the EBS warm up script to `cat` the contents of the `Main_Watershed` files for drb and nhd to `/dev/null`.

The output should look like this:

```
[2017-02-06 18:00:00+00:00] Warming up RWD EBS volume...
cat /opt/rwd-data/nhd/Main_Watershed/domainfromregions.dbf
cat /opt/rwd-data/nhd/Main_Watershed/domainfromregions.prj
cat /opt/rwd-data/nhd/Main_Watershed/domainfromregions.sbn
cat /opt/rwd-data/nhd/Main_Watershed/domainfromregions.sbx
cat /opt/rwd-data/nhd/Main_Watershed/domainfromregions.shp
cat /opt/rwd-data/nhd/Main_Watershed/domainfromregions.shx
cat /opt/rwd-data/nhd/Main_Watershed/gwgrid.tfw
...
[2017-02-06 20:00:00+00:00] Done
```

scripts/aws/ebs-warmer.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# Ref: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-initialize.html
+
+function log() {
+    echo "[$(date --rfc-3339=seconds)] $1"
+}
+
+log "Warming up RWD EBS volume..."
+find /opt/rwd-data/nhd/Main_Watershed/* -type f | xargs -L1 -t cat >/dev/null
+find /opt/rwd-data/drb/Main_Watershed/* -type f | xargs -L1 -t cat >/dev/null
```
kdeloach: Shellcheck complains about these lines:

```
In scripts/aws/ebs-warmer.sh line 10:
find /opt/rwd-data/nhd/Main_Watershed/* -type f | xargs -L1 -t cat >/dev/null
^-- SC2038: Use either 'find .. -print0 | xargs -0 ..' or 'find .. -exec .. +' to allow for non-alphanumeric filenames.
```

However, I wanted to echo the commands being executed as they happen (using the `xargs` `-t` flag) to give a sense of progress but I couldn't find an elegant way to do this by following the `shellcheck` suggestions.

hectcastro: I think that `shellcheck` just wants you to add `-print0` to `find` and `-0` to `xargs` so that it can more confidently parse the file names that are output:

```bash
$ find /opt/rwd-data/drb/Main_Watershed/* -type f -print0 | xargs -0 -L1 -t cat >/dev/null
```

kdeloach: Oh, thanks! That did the trick.

scripts/aws/ebs-warmer.sh
```diff
@@ -0,0 +1,17 @@
...
+# Ref: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-initialize.html
+
+function log() {
+    echo "[$(date --rfc-3339=seconds)] $1"
+}
+
+log "Warming up RWD EBS volume..."
+
+find /opt/rwd-data/nhd/Main_Watershed/* -type f -print0 \
+    | xargs -0 -L1 -t cat >/dev/null
```
hectcastro: I saw an increase in throughput by adding `-P0` to `xargs`. It allows `xargs` to spawn more instances of `cat` at a time.

kdeloach: Great tip. Updated.

PR #1684, opened 2017-02-07
Postpone ASG scheduled shutdown for staging

mmcfarland:
Push back the shutdown time for `AppServer`, `Worker` and `Tiler` instances in `staging`. This will enable testers in western timezones to use the staging service later into their workday.  Previous value was translated as 6PM EST (11PM UTC) and the new value is 8PM EST (1AM UTC).

This is not intended to make any alterations for the ASG scheduled actions on `production`.

### DIFF
--- a/deployment/default.yml.example
+++ b/deployment/default.yml.example
@@ -29,8 +29,8 @@ AppServerAutoScalingScheduleStartCapacity: '1'
 # 8AM ET
 AppServerAutoScalingScheduleStartRecurrence: '0 13 * * 1-5'
 AppServerAutoScalingScheduleEndCapacity: '0'
-# 6PM ET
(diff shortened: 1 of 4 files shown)


deployment/default.yml.example
```diff
@@ -60,7 +60,7 @@ WorkerAutoScalingScheduleStartCapacity: '2'
 WorkerAutoScalingScheduleStartRecurrence: '0 13 * * 1-5'
 WorkerAutoScalingScheduleEndCapacity: '0'
 # 6PM ET
-WorkerAutoScalingScheduleEndRecurrence: '0 23 * * *'
+WorkerAutoScalingScheduleEndRecurrence: '0 1 * * *'
```
hectcastro: The comments just above these still say 6PM.

mmcfarland:
Thanks, updated commit to include changing the comments.

PR #1691, opened 2017-02-16
Mount RWD default folder which exists on all platforms

rajadain:
## Overview

`/dev/null` could not be mounted as a folder on macOS because it is a character file. `/tmp` should be available on all platforms.

### Notes

I'm assigning one macOS person and one Linux person to test this on all platforms we use.

## Testing Instructions

Check out this branch. Ensure `$RWD_DATA` environment variable is **not** set. Run `vagrant up`. Ensure there are no errors and the app loads correctly at [:8000/](http://localhost:8000/)


kdeloach:
Thanks.

kellyi:
I'm setting up MMW now, and I'll re-provision my worker VM when that's done. Can confirm that the `/tmp` dir exists on macOS though

Issue #1687, opened 2017-02-09
Add "original clicked point" when RWD is initiated


mmcfarland:
A visual cue that the process is underway.

kellyi:
#1693 creates the marker and it should work on RWD success & failure, but I'm still setting up the RWD data on my local so I've only checked the failure case so far.

kellyi:
Got RWD data set up and tested the success case. It worked, so I assigned #1693 for review.

PR #1693, opened 2017-02-17
Mark original clicked point on starting RWD

kellyi:
## Overview

This PR displays the RWD "Original Clicked Outlet Point" marker on starting RWD, then removes it when the chain of deferreds completes in success or failure. If RWD succeeds, the marker's removed to be replaced by the existing pair of markers for "Original Clicked Outlet Point" and "Outlet Point Snapped to Nearest Stream". If RWD fails, an error message shows and the marker is cleared from the map.

I set the marker up to use the same style `divIcon` and popup as the existing "Original" RWD marker. To standardize it, I moved the function which made that icon into `draw/utils.js`. I also bound a popup to it with the "Original Clicked Outlet Point" message.

Connects #1687 

## Screenshot

screen shot 2017-02-17 at 10 49 42 am

## Testing

* get this branch, then run `bundle.sh`
* visit MMW in the browser and run RWD for both DRB and continental US. Verify that the marker clears in both the success and failure cases. (To cause a failure, I think you can try the points noted to cause an error in https://github.com/WikiWatershed/rapid-watershed-delineation/issues/65)

### DIFF
--- a/src/mmw/js/src/draw/utils.js
+++ b/src/mmw/js/src/draw/utils.js
@@ -72,6 +72,13 @@ function placeMarker(map, drawOpts) {
     return defer.promise();
 }
 
+function createRwdMarkerIcon(iconName) {
+    return L.divIcon({
+        className: 'marker-rwd marker-rwd-' + iconName,
+        iconSize: [16,16]
+    });
+}
+
 // Cancel any previous draw action in progress.
 function cancelDrawing(map) {
     map.fire('draw:drawstop');
@@ -94,6 +101,7 @@ function isValidForAnalysis(shape) {
 module.exports = {
     drawPolygon: drawPolygon,
     placeMarker: placeMarker,
+    createRwdMarkerIcon: createRwdMarkerIcon,
     cancelDrawing: cancelDrawing,
     polygonDefaults: polygonDefaults,
     shapeArea: shapeArea,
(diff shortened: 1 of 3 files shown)


arottersman:
Testing this now!

src/mmw/js/src/draw/utils.js
```diff
@@ -72,6 +72,13 @@ function placeMarker(map, drawOpts) {
     return defer.promise();
 }
 
+function createRwdMarkerIcon(iconName) {
```
arottersman: Consider refactoring this further to remove the duplicate `L.marker` and `bindPopup` code
```
function createRwdMarker(latlng, iconName, popupMessage) {
   return L.marker(latlng, {
        icon : L.divIcon({
            className: 'marker-rwd marker-rwd-' + iconName,
            iconSize: [16,16]
        })
    }).bindPopup(popupMessage);
}
```

kellyi: Ah, good idea. We can probably skip passing in the message as we'll know which one to show based on the name.

kellyi: Tried this and it looks like binding the popup prevents the `addTo` method from getting called correctly for the initial clicked point (and the popups seem to get bound in a slightly different way for the resulting points. Going to leave this as is for now.

kellyi:
Thanks for taking a look!

Issue #1690, opened 2017-02-13
Description text at top of model output tabs


aufdenkampe:
@ajrobbins, last week we had a question come in from a MMW.

> How do I determine the lbs/year load reduction form the water quality data? Is the 'Load (kg)' already in kg/yr? If so, I can convert it to lbs.

This helped me realize the need for some text at the top of each of the Model output tabs, similar to what we have at the top of the Analyze output tabs. Here are some suggestions.

For the Site Storm Model:
* Hydrology: "Water fluxes in a single, 24-hour hypothetical storm event as simulated by SLAMM & TR55 model algorithms. "
* Water Quality: "Total loads delivered in a single, 24-hour hypothetical storm event as simulated by EPAs STEP-L model algorithms. "

For the Watershed Multi-Year Model:
* Hydrology: "Average monthly water fluxes in cm (= 10 L per square meter) from 30-years of simulated daily water balance calculated by the GWLF-E (MapShed) model."
* Water Quality: "Average annual loads from 30-years of simulated daily fluxes calculated by the GWLF-E (MapShed) model."

Please suggest edits. Let's put this in the queue to implement sooner than later.

ajrobbins:
@aufdenkampe I agree, some explanatory text for the model tabs makes sense. Tagging @jfrankl to consider styling for this and the analysis description text for future UI updates.

For the text:

**Site Storm Model**
- Hydrology (runoff): What about something like:
"Results of a 24-hour hypothetical storm event as simulated by SLAMM & TR55 model algorithms. "
- Water quality: looks good, I'd remove "single" for brevity

**Watershed Multi-Year Model**
- Hydrology: I'm confused by the unit conversion. To clarify, does 1cm water depth = 10L per square meter?
- Water Quality: how about "Average annual loads from 30-years of daily fluxes simulated by the GWLF-E (MapShed) model."

aufdenkampe:
@ajrobbins, thanks for the feedback. I like your wording suggestions. 

Is it possible to include these short text explanations in the present UI? I'm thinking they could go in the top of each of the Model tabs, just like we do with the Analyze tabs. I assume the machinery is all there, and it would be very nice to include this text sooner than later (as in the next release). Let me know if this might be possible. I'm getting questions about these details and minor differences several times per week.

[...]

PR #1699, opened 2017-02-20
Add introductory text to model results

rajadain:
## Overview

Adds introductory text to each type of result: Runoff and Water Quality for TR55, and Hydrology and Water Quality for GWLF-E.

Connects #1690 

### Demo

image

image

image

image

### Notes

In this current iteration, I've copied the text as exactly specified by @ajrobbins on #1690, but would recommend considering the following changes:

 * Replace "SLAMM & TR55" with "SLAMM and TR55". The ampersand should be reserved for when the pairing is a proper noun, such as Fox & Hound or Lord & Taylor.
 * Is it "TR55" or "TR-55"? I'm not sure if we've ever referred to it in body copy before.
 * Replace "in cm (= 10 L per square meter)" with "in cm (equal to 10 L/m²)", to reuse unit conventions applied elsewhere in the application, and avoid mathematical operators in sentences.
 * This is the first time we're using "MapShed" in body copy. Should we call it that, or leave it at GWLF-E, which is the "official" name?

## Testing Instructions

 * Check out this branch and bundle scripts `./scripts/bundle.sh --debug`
 * Make a new TR55 project or load an existing one. Ensure text is above each result type.
 * Do the same for GWLF-E.
 * Ensure there are no typos.

Also tagging @ajrobbins to take a look.

### DIFF
--- a/src/mmw/js/src/modeling/gwlfe/quality/templates/result.html
+++ b/src/mmw/js/src/modeling/gwlfe/quality/templates/result.html
@@ -1,1 +1,6 @@
+<p class="result-text text-muted">
+    Average annual loads
+    from 30-years of daily fluxes
+    simulated by the GWLF-E (MapShed) model.
+</p>
 <div class="quality-table-region"></div>
--- a/src/mmw/js/src/modeling/gwlfe/runoff/templates/result.html
+++ b/src/mmw/js/src/modeling/gwlfe/runoff/templates/result.html
@@ -1,3 +1,8 @@
+<p class="result-text text-muted">
+    Average monthly water fluxes in centimeters
+    from 30-years of daily water balance
+    simulated by the GWLF-E (MapShed) model.
+</p>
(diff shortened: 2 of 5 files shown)


ajrobbins:
@rajadain Can you change the text for the Multi-Year model hydrology tab to: "Average monthly water fluxes in centimeters from 30-years of daily water balance simulated by the GWLF-E (MapShed) model."

Based on https://github.com/WikiWatershed/model-my-watershed/issues/1690#issuecomment-281187773
Also addresses your note on the unit conversion.

ajrobbins:
Good idea to replace the & and make it "TR-55". Let's keep the "MapShed" because users who get to that tab may be more familiar with the colloquial name rather than the model name.

rajadain:
Updated values:

image

image

Not right now, but it might be a good idea to hyperlink the model names to more detailed documentation.

PR #1700, opened 2017-02-21
Add global navigation static components

jfrankl:
## Overview

Add static components for breadcrumb/global navigation.

Connects #XXX

### Demo

Here's a screenshot:
image

### Notes

When a new project is started, only "Select Area" should be visible. Once the user enters the "Analyze" screen, that breadcrumb should become visible as well. When the user researches "Model", they should see "Select Area", "Analyze", and "Model". Add the `.active` class on the active step. So if the user is on the "Analyze" screen, "Analyze" should be active.

### DIFF
--- a/src/mmw/img/global-navigation-arrow.svg
+++ b/src/mmw/img/global-navigation-arrow.svg
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 16.0.4, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 width="6px" height="16px" viewBox="0 0 6 16" style="enable-background:new 0 0 6 16;" xml:space="preserve">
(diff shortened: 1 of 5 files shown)


ajrobbins:
Without seeing how the interaction works, nav looks good from the screenshot!

rajadain:
Current labels are:

 * Choose Area of Interest (for Select Area)
 * Geospatial Analysis (for Analysis)
 * Depending on the Model, one of the following (for Model):
    * Site Storm Model (TR-55)
    * Watershed Multi-year Model (GWLF-E)

Would the breadcrumb show these longer labels, or do we intend to drop them for the simpler names?

rajadain:
Also notable, a very similar effect can be achieved without adding the SVG, using existing Font Awesome and a CSS Transform:

```html
<i class="fa fa-angle-right" style="
    color: #888;
    transform: matrix(0.80,0.00,0.00,1.5,0,1);
"></i>
```

image

jfrankl:
@rajadain let's use the simplified labels. 

I don't have a preference between the two techniques for the breadcrumbs/right-angles. Does your solution work on IE and mobile browsers? If so, I'll leave it up to you.

rajadain:
SVGs are better supported. I do think we'd need some differentiation for the two models though.

jfrankl:
I don't think the two models differentiation belongs in the breadcrumbs, but perhaps that can be displayed elsewhere.

rajadain:
That's a good point. Alright then, this looks good to go!

rajadain:
@azavea-bot rebuild

arottersman:
Will these act for actual navigation? If you click `Select Area` will you go back to Draw mode?

If you've gone to Analyze and navigated back to Select Area, will you still see the Analyze breadcrumb?

Issue #1708, opened 2017-02-27
RWD Draw Tool Issues


kdeloach:
Noticed 2 issues on staging when delineating a DRB watershed:

1. Null reference exception happens when you delineate a point outside of DRB bounds. I believe there is a condition where we try to remove `rwdClickedPoint` from the map before it gets created.

2. Reset button does not clear the `rwdClickedPoint` marker.

kellyi:
Possibly part of the same bug but it looks like clicking "Reset" doesn't actually stop RWD from running. On staging, for instance, you can click to start an RWD job, then click "Reset" and everything will appear to have cleared except the marker. However, it seems like the polling continues and the RWD shape will eventually load if the app's left alone after clicking reset.

I'm guessing that the deferred's not rejecting on clicking "Reset", which would explain why the fail block doesn't trigger immediately for the original clicked point marker.

kdeloach:
Good point. If we can easily reject the deferred when the reset button is pressed, I think it's worth doing. But if it's too much of a hassle then maybe we should leave it as-is. More important is fixing the null reference error.

kellyi:
Going to check whether other kinds of polling results have the same behavior.

kellyi:
Investigated a little more and it does seem like the deferred's not rejecting on clicking the reset button/calling `reset()` on the `rwdTaskModel`. I got a hacky version of rejecting this to work by giving the taskModel a reference to the deferred object for whatever polling is currently taking place, then doing something like

```js
if (this.rwdTaskModel.defaults.defer) {
    window.console.log('rejecting');
    this.rwdTaskModel.defaults.defer.reject('rejected');
}
```

in the function called on clicking the "Reset" button; rejecting the deferred here causes the polling to cease after about 5 more tries, then displays the standard failure error message.

I'm tempted to spin handling that off into another issue, as I think an actual fix involves a more through update to the task model. Incidentally, I *think* the non-rejected-deferred may also be what causes these errors when running TR-55 then immediately clicking the back arrow and waiting a bit for the results to trigger an error:

<img width="917" alt="screen shot 2017-02-28 at 11 21 51 am" src="https://cloud.githubusercontent.com/assets/4165523/23415016/e19fc978-fdab-11e6-8c33-2e0bb06e2635.png">

[...]

PR #1709, opened 2017-02-27
Fix point null ref exception when rwdClickedPoint's undefined

kellyi:
## Overview

This PR fixes a null reference exception described in #1708 whereby the app would try to remove a null original clicked RWD point marker from a map.

To fix it, I added a guard to check that the marker had actually been created before removing it in the `.always` clause.

Connects #1708 

## Testing Instructions

 * rebuild this branch then follow the steps described in the first issue reported in #1708 to verify that this fixes the bug

Going to work on another branch to fix the second issue noted there.

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -423,7 +423,7 @@ var WatershedDelineationView = Marionette.ItemView.extend({
             itemName = $item.text(),
             snappingOn = !!$item.data('snapping-on'),
             dataSource = $item.data('data-source'),
-            rwdClickedPoint;
+            rwdClickedPoint = null;
 
         clearAoiLayer();
         this.model.set('pollError', false);
@@ -456,7 +456,9 @@ var WatershedDelineationView = Marionette.ItemView.extend({
                 displayAlert(message, modalModels.AlertTypes.warn);
             })
             .always(function() {
-                map.removeLayer(rwdClickedPoint);
+                if (rwdClickedPoint) {
+                    map.removeLayer(rwdClickedPoint);
+                }
                 self.model.enableTools();
             });
     },


kdeloach:
Looks good, but I don't think I will have time to test it this sprint.

kellyi:
No worries -- I'll reassign it.

kellyi:
Also realized that this should maybe be retargeted at `release`? cc: @mmcfarland @ajrobbins

src/mmw/js/src/draw/views.js
```diff
@@ -436,10 +436,12 @@ var WatershedDelineationView = Marionette.ItemView.extend({
             .then(function(latlng) {
                 // Assign `rwdClickedPoint` so it's in scope to be removed from
                 // the map when the deferred chain completes.
-                rwdClickedPoint = L.marker(latlng, {
-                    icon: drawUtils.createRwdMarkerIcon('original-point')
-                }).bindPopup('Original clicked outlet point');
-                rwdClickedPoint.addTo(map);
+                if (map) {
```
mmcfarland: Is this guard on `map` necessary?  In recreating the error, `map` was not undefined and given that it's an argument to the function which creates this promise, this function would never execute.  If `App.getLeafletMap()` is returning a falsey value, then we'd need to add the guard at `utils.placeMarker` which is where an exception would originate.  However, I suspect that we couldn't get to this method at all without a valid map reference.  Consider removing this nesting and guard, or if it is needed, moving it to before the first usage of `map`.

src/mmw/js/src/draw/views.js
```diff
@@ -456,7 +458,9 @@ var WatershedDelineationView = Marionette.ItemView.extend({
                 displayAlert(message, modalModels.AlertTypes.warn);
             })
             .always(function() {
-                map.removeLayer(rwdClickedPoint);
+                if (map && rwdClickedPoint) {
```
mmcfarland: Same comment as above, `rwdClickedPoint` seems to be the only reference that might be `null`.

mmcfarland:
Yep, let's include this in the release as part of the RWD overhaul.

kellyi:
Oh, good point. I pushed a fixup which cuts the `if (map)` checks so now this change just checks whether `rwdClickedPoint` exists before attempting to remove it from the map. Also retargeted this PR at `release`.

kellyi:
Thanks! Squashed the fixup; will merge once the tests pass.

PR #1710, opened 2017-02-28
Add purge cache only option to setupdb.sh

mmcfarland:
## Add purge cache only option to setupdb.sh

`-x` option to script will accept `PATH` argument to delete s3 tile cache

### Demo

`./setupdb.sh -x drb_streams_v2`

### Notes

I wanted to purge a particular tile set without having to reload the entire streams table, so I added this option.

## Testing Instructions

From a staging app instance, run `./setupdb.sh -x drb_streams_v2` and verify the aws command generated is correct.  I've run this already on `drb_streams_v2` so it's safe to do on that dataset (although the directory will be empty).

### DIFF
--- a/scripts/aws/setupdb.sh
+++ b/scripts/aws/setupdb.sh
@@ -15,6 +15,7 @@ where: \n
     -m  load/reload mapshed data\n
     -p  load/reload DEP data\n
     -q  load/reload water quality data\n
+    -x  purge s3 cache for given path\n
 "
 
 # HTTP accessible storage for initial app data
@@ -25,7 +26,7 @@ load_stream=false
 load_mapshed=false
 load_water_quality=false
 
-while getopts ":hbsdpmqf:" opt; do
+while getopts ":hbsdpmqf:x:" opt; do
     case $opt in
         h)
             echo -e $usage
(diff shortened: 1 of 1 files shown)


scripts/aws/setupdb.sh
```diff
@@ -44,6 +45,8 @@ while getopts ":hbsdpmqf:" opt; do
             load_water_quality=true ;;
         f)
             file_to_load=$OPTARG ;;
+        x)
```
hectcastro: I think that `x:` needs to be added to `getopts` above or else I get the following error message:

```
invalid option: -x
```

mmcfarland: You're right, I had an amended commit that I failed to push up.  Updated with newer commit with `getopts`.

scripts/aws/setupdb.sh
```diff
@@ -25,7 +26,7 @@ load_stream=false
 load_mapshed=false
 load_water_quality=false
 
-while getopts ":hbsdpmqf:" opt; do
+while getopts ":hbsdpmqfx:" opt; do
```
hectcastro: Since `f` also takes a value, doesn't it need to be `f:x:`? That's my takeaway from the docs at least.

>When you want getopts to expect an argument for an option, just place a : (colon) after the proper option flag. If you want -A to expect an argument (i.e. to become -A SOMETHING) just do:
>
>```
>getopts fA:x VARNAME
>```

mmcfarland:
You're right, I broke `-f` with the missing `:`, updated in fixup	7a234f4

PR #1713, opened 2017-02-28
Clear RWD marker on clicking reset

kellyi:
## Overview

This is a UI fix to address an issue noted in #1708 whereby the RWD original clicked point marker won't clear on clicking reset. This one's just a quick UI fix which finds and clears the marker when reset's clicked.

I think the underlying issue warrants some changes to enable rejecting Deferred objects issued by `TaskModels`; it seems as if clicking "Reset" doesn't always call `.reject()`, so RWD jobs continue running even after reset.

Going to investigate that issue a bit more, but thought I'd put this up to address the UI issues.

Connects #1708 

## Testing Instructions

 * rebuild this branch and view in the browser, then open the console
 * start RWD then click reset and verify that the marker disappears. Repeat a few times to ensure everything stays in sync
 * start RWD and let it complete, then verify that the resulting shape has only two markers: one for the origin point, one for the outlet point

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -422,8 +422,7 @@ var WatershedDelineationView = Marionette.ItemView.extend({
             $item = $(e.currentTarget),
             itemName = $item.text(),
             snappingOn = !!$item.data('snapping-on'),
-            dataSource = $item.data('data-source'),
-            rwdClickedPoint = null;
+            dataSource = $item.data('data-source');
 
         clearAoiLayer();
         this.model.set('pollError', false);
@@ -434,11 +433,12 @@ var WatershedDelineationView = Marionette.ItemView.extend({
                 return validatePointWithinDataSourceBounds(latlng, dataSource);
             })
             .then(function(latlng) {
-                // Assign `rwdClickedPoint` so it's in scope to be removed from
-                // the map when the deferred chain completes.
-                rwdClickedPoint = L.marker(latlng, {
+                // Set `rwd-original-point` in the model so it can be
+                // removed when the deferred chain completes.
+                var rwdClickedPoint = L.marker(latlng, {
                     icon: drawUtils.createRwdMarkerIcon('original-point')
                 }).bindPopup('Original clicked outlet point');
+                self.model.set('rwd-original-point', rwdClickedPoint);
                 rwdClickedPoint.addTo(map);
                 return latlng;
             })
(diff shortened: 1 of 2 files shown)


src/mmw/js/src/draw/views.js
```diff
@@ -568,6 +573,11 @@ var ResetDrawView = Marionette.ItemView.extend({
     },
 
     resetDrawingState: function() {
+        App.getLeafletMap().eachLayer((l) => {
+            if (l.id && l.id === 'rwd-origin-point') {
+                App.getLeafletMap().removeLayer(l);
+            }
+        });
```
kdeloach: Kind of curious why these points aren't handled like other drawn layers. By updating model state.

kellyi: Oh, I must have skipped placing it in state in the first place. I'll see about adding it to the `rwdTaskModel` state, which would obviate needing to do the find by id stuff.

kdeloach: Just pointing out that the machinery to show/hide layers depending on different state may already exist. I'd have to verify that though.

kellyi: Makes sense. I amended that commit to store the marker in the model so that `get` and `unset` can be used in the reset button function.

src/mmw/js/src/draw/views.js
```diff
@@ -456,8 +456,9 @@ var WatershedDelineationView = Marionette.ItemView.extend({
                 displayAlert(message, modalModels.AlertTypes.warn);
             })
             .always(function() {
-                if (rwdClickedPoint) {
-                    map.removeLayer(rwdClickedPoint);
+                if (self.model.has('rwd-original-point')) {
```
mmcfarland: Since this process is repeated and requires multiple steps to be successful, consider putting it inside of a model instance method, so that you might be able to say `model.clearRwdMarker()` or similar, here and below.  If the model doesn't have a reference to `App`, it could accept a `map` argument for clearing the layer.

kellyi: Sounds good to me!

src/mmw/js/src/draw/views.js
kdeloach: :+1:

kellyi:
Pushed an update which adds a `clearRwdClickedPoint` method to the Toolbar model, taking the map as an argument. The method performs the if checks internally -- if map/if rwd-marker -- before attempting to remove the marker from the map.

kellyi:
Thanks! I'll make a new issue for the deferred/reject stuff, and I'll squash the fixups on this one before merging.

kellyi:
Made #1716 for the reject/deferred work.

Issue #1719, opened 2017-03-03
Remove Model descriptive text from Compare view


kellyi:
As a result of adding the descriptive model text in #1699 the precipitation slider on the compare view sits above the bar chart, I think because the bar chart got pushed down but the slider has some sort of absolute positioning:

![screen_shot_2017-03-03_at_10 56 07_am](https://cloud.githubusercontent.com/assets/4165523/23559003/b73a5900-0003-11e7-94e1-593aea6b85cc.png)

We should update the compare view template so that it does not display the descriptive text, and possible figure out how to fix the position of the slider as relative to the bar chart.

kellyi:
Target this at the hotfix/1.18.1 branch.

PR #1720, opened 2017-03-03
Remove TR-55 model descriptions from Compare view

kellyi:
## Overview

This PR fixes a bug introduced in #1699 whereby the TR-55 model descriptions would push the bar chart underneath the precipitation slider on the compare view. The result was that the precipitation slider hovered over the bar chart.

The fix here adds some checks to show the model descriptions only when the app's not in Compare view. This restores the functionality for compare view prior to adding the descriptions, and it keeps the descriptions for the non-Compare view.

Connects #1719 

### Demo

screen shot 2017-03-03 at 1 32 36 pm

->

screen shot 2017-03-03 at 1 32 40 pm

->

screen shot 2017-03-03 at 1 33 00 pm

->

screen shot 2017-03-03 at 1 33 05 pm

### Notes

A more thorough fix would position the precipitation slider to be fixed beneath the bar chart in all cases on compare view, but if we want to pursue that line I think we should kick it to another card as it may entail more changes to the layout and style of the Compare view.

## Testing Instructions

 * get this branch, then `bundle`
 * visit the site in the browser and hard refresh to ensure you've got the latest bundle
 * select an AOI and run TR-55. When the results return verify that you see the model descriptions for both runoff and water quality
 * click "Compare" and when the results return, verify that you do not see the model descriptions and that the slider remains positioned beneath the bar chart

### DIFF
--- a/src/mmw/js/src/modeling/tr55/quality/templates/result.html
+++ b/src/mmw/js/src/modeling/tr55/quality/templates/result.html
@@ -1,6 +1,8 @@
+{% if showModelDescription %}
 <p class="result-text text-muted">
     Total loads delivered in a 24-hour hypothetical storm event
     as simulated by EPA's STEP-L model algorithms.
 </p>
+{% endif %}
 <div class="quality-chart-region"></div>
 <div class="quality-table-region"></div>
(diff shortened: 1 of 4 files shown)


kellyi:
Thanks! Just realized there may be some extra commas in the `templateHelper` return objects, so I'll remove those before merging.

kellyi:
Linted, tested, and squashed the fixup. Will merge this once the CI tests pass.

PR #1727, opened 2017-03-08
Remove back buttons

caseycesari:
## Overview

The various back buttons in the app have been superseded by the new
breadcrumbs feature, and are no longer needed.

The back button DOM elements, and related style rules and code, are
removed here.

Connects #1726

### Demo

image
*Returning to Draw from Analyze*

image
*Modeling page*

image
*Compare page*

## Testing Instructions

* Start the app server and rebundle all of the JS bundles.
* Navigate back and forth through the app, and verify that all of the back buttons have been removed, and that the main area of the app is always sized correctly (i.e. there isn't a blank space where the analyze -> draw footer used to be)


src/mmw/js/src/compare/templates/compareWindow.html
```diff
@@ -1,8 +1,5 @@
 <div id="compare-scenarios-region"></div>
 <div id="compare-footer">
-    <div class="back"> <!-- Back -->
-      <a data-url="/project" class="btn btn-md btn-icon"><i class="fa fa-arrow-left"></i></a>
-    </div> <!-- End Back -->
     <h3>Navigate Scenarios</h3>
```
arottersman: Should we remove this or is it referring to the arrows on the far right of the page? Either way I find it confusing.

screen shot 2017-03-08 at 12 05 26 pm

caseycesari: I think it is referring to the arrows on the far right of the page. It is indeed confusing. I'll open an issue.

caseycesari: https://github.com/WikiWatershed/model-my-watershed/issues/1729

caseycesari:
Good catch. I missed the biggest back button. I'll remove it.

caseycesari:
Added a fixup commit to remove the back button on the analyze page.

Note the style changes.

Before:
image

After:
image

The updated text comes from the latest wireframes, and makes the bigger button not look as awkward.

PR #1730, opened 2017-03-08
Update README.md

caseycesari:
Add note about how to run JS linter

## Testing Instructions

- Verify that this is the correct command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,12 @@ $ ./scripts/test.sh
 
 ##### Python
 
+To check for Python lint:
+
+```bash
+$ ./scripts/check.sh
+```
+
 To run all the tests on the Django app:
 
 ```bash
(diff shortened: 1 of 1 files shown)


rajadain:
Consider adding a mention to [`check.sh`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/scripts/check.sh) in the [Python section](https://github.com/WikiWatershed/model-my-watershed#python) as well.

kellyi:
Would it be worth adding a `lint.sh` script that just called both of these?

rajadain:
I would be up for renaming `check.sh` to `lint.sh` and adding JavaScript linting to it.

kellyi:
Since this is still open: I noticed that the `test.sh` script has a command in it to run the linter as part of the tests, but the line's commented out:

https://github.com/WikiWatershed/model-my-watershed/blob/develop/scripts/test.sh#L18

Not sure why that was commented out in the first place, but if it wouldn't screw anything up I'd be in favor of un-commenting the line so that `test.sh` would do the linting too.

caseycesari:
Going to merge this in, but the other suggestions are worthwhile.

Issue #1736, opened 2017-03-21
Download analyze and model results as CSV


mmcfarland:
Each panel of analysis and model results should get a "Download Data" treatment as displayed on p1 of the wireframes (p9, shows additional panel). These datasets are smallish and already loaded into the browser.  Rather than generate additional load on the server, or cache them there, we should attempt to assemble the downloadable CSV from the content the browser has already assembled for display.  Are there disadvantages from doing this?

I've used and have had good experience with http://papaparse.com/.  It can be evaluated for use in this feature. 

https://drive.google.com/drive/folders/0B0hs2mX4ApaLN0xLSjY0cVhDckk

kellyi:
Q to clarify the desired behavior for this:

Each panel should get a download button which creates a CSV using *only* the data visible on that panel -- and *not* the data visible on the panel + the other data?

mmcfarland:
Yes, this is intended to be a per-dataset download, we won't be making a zip file of multiple results at once.

PR #1740, opened 2017-03-21
Download Analyze & Model data in CSV format

kellyi:
## Overview

This PR enables downloading the data from the analyze and model tabs in CSV format. The first commit here adds the `papaparse` library to handle parsing the CSV; the second adds a utility function to use the library then calls it from a link on each of the analyze tabs plus the TR-55 and MapShed tabs.

Connects #1736 

### Demo

screen shot 2017-03-21 at 5 46 21 pm

[nlcd_land_cover_2017-03-21T21-46-25.378Z.csv.txt](https://github.com/WikiWatershed/model-my-watershed/files/859750/nlcd_land_cover_2017-03-21T21-46-25.378Z.csv.txt)

### Notes

I used an `fa-download` icon which was already available in the app for the "Download..." div, but I'll double check the wireframes to ensure that it's the correct style and import a new one from Fontello if not.

## Testing Instructions

 * get this branch, `npm.sh install` and `bundle`
 * visit the site in the browser and choose an AOI
 * for each analysis tab, click the "Download..." link at the bottom and verify that it outputs a CSV file with the data represented on the tab
 * run TR-55 and click the download links on both tabs. Verify that the data in the CSV represents what's displayed in the tab
 * repeat the above step for MapShed

### DIFF
--- a/src/mmw/js/src/analyze/constants.js
+++ b/src/mmw/js/src/analyze/constants.js
@@ -1,0 +1,51 @@
+"use strict";
+
+var constants = {
+    csvColumnMaps: {
+        nlcd: {
+            'nlcd': 'nlcd_id',
+            'area': 'area_sq_m',
+            'coverage': 'coverage_pct',
+        },
+        soil: {
+            'code': 'soil_code',
+            'area': 'area_sq_m',
+            'coverage': 'coverage_pct'
+        },
+        animals: {
+            'aeu': 'animal_count',
+            'type': 'animal_type'
+        },
+        pointSource: {
+            'kgp_yr': 'phosphorus_load_kg_per_yr',
+            'mgd': 'discharge_million_gallon_per_day',
+            'kgn_yr': 'nitrogen_load_kg_per_yr'
+        },
+        waterQuality: {
+            'nord': 'nord',
+            'areaha': 'area_hectares',
+            'tss_tot_kg': 'total_suspended_solids_kg_yr_per_ha',
+            'tp_tot_kgy': 'total_phosphorus_kg_yr_per_ha',
+            'tn_tot_kgy': 'total_nitrogen_kg_yr_per_ha',
+            'tss_urban_': 'total_suspended_solids_urban_kg_yr_per_ha',
+            'tp_urban_k': 'total_phosphorus_urban_kg_yr_per_ha',
+            'tn_urban_k': 'total_nitrogen_urban_kg_yr_per_ha',
+            'tss_rip_kg': 'total_suspended_solids_riparian_kg_yr_per_ha',
+            'tn_riparia': 'total_nitrogen_riparian_kg_yr_per_ha',
+            'tp_riparia': 'total_phosphorus_riparian_kg_yr_per_ha',
+            'tp_pt_kgyr': 'total_phosphorus_point_source_kg_yr_per_ha',
+            'tn_pt_kgyr': 'total_nitrogen_point_source_kg_yr_per_ha',
+            'tss_natura': 'total_suspended_solids_natural_kg_yr_per_ha',
+            'tn_natural': 'total_nitrogen_natural_kg_yr_per_ha',
+            'tp_natural': 'total_phosphorus_natural_kg_yr_per_ha',
+            'tss_ag_kgy': 'total_suspended_solids_agricultural_kg_yr_per_ha',
+            'tn_ag_kgyr': 'total_nitrogen_agricultural_kg_yr_per_ha',
+            'tp_ag_kgyr': 'total_phosphorus_agricultural_kg_yr_per_ha',
+            'tss_concmg': 'total_suspended_solids_concentration_mg_per_l',
+            'tp_yr_avg_': 'total_phosphorus_concentration_mg_per_l',
+            'tn_yr_avg_': 'total_nitrogen_concentration_mg_per_l'
+        }
+    }
+};
+
+module.exports = constants;
--- a/src/mmw/js/src/analyze/templates/analyzeResults.html
+++ b/src/mmw/js/src/analyze/templates/analyzeResults.html
@@ -1,3 +1,6 @@
 <div class="desc-region"></div>
 <div class="chart-region"></div>
 <div class="table-region"></div>
+<div class="downloadcsv-link" data-action="download-csv">
+    <i class="fa fa-download"></i> Download this data
+</div>
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -1,6 +1,7 @@
 "use strict";
 
 var $ = require('jquery'),
+    _ = require('underscore'),
     lodash = require('lodash'),
     L = require('leaflet'),
     Marionette = require('../../shim/backbone.marionette'),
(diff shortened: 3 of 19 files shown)


mmcfarland:
From the screenshot, can you use a `font-size` that is consistent with other text on the panels?

kellyi:
Sure! I'll push a fixup to update that.

kellyi:
Looks like the table text is set to `0.7rem`, so the fixup uses that value for the download link:

screen shot 2017-03-21 at 5 59 49 pm

kellyi:
@jfrankl Is the `fa-download` icon used above in tune with what you wanted? Or should I add another one?

jfrankl:
Looks great. Thanks @kellyi!

kellyi:
@azavea-bot rebuild

src/mmw/js/src/analyze/templates/analyzeResults.html
```diff
@@ -1,3 +1,6 @@
 <div class="desc-region"></div>
 <div class="chart-region"></div>
 <div class="table-region"></div>
+<div class="downloadcsv-link" data-action="download-csv">
+    <i class="fa fa-download"></i> Download this data
```
kellyi: I *think* this template only gets rendered once the results have returned for a tab.

arottersman: Oh, good point. It works fine with no data actually...just downloads an empty file.

kellyi:
re: Column names. I'm not too sure how these CSVs are going to be used. @mmcfarland do you have any suggestions on whether to attempt to modify the column names or value names at all? Or should I just leave them alone and let the data be output in a format like what's in the app state?

mmcfarland:
I think it would be somewhat preferable to have better names mapped over the keys we use, although in some cases they might be difficult to come up with something concise.  I think this is an "ounce of prevention" situation where it might be faster now to add better names rather than having to do data dictionary documentation later.

kellyi:
Sounds good to me. I just pushed two fixups: one to use `this.model.get`; the other to split the download operations off into their own methods in each class. I'll make & push another update to replace the key names with something more legible.

kellyi:
Most of these name changes might be pretty straightforward; the catchment water quality analysis result will be the most involved because it has the most columns of any of the data sets:

```csv
tss_urban_,tn_riparia,tn_pt_kgyr,tp_urban_k,tss_tot_kg,geom,nord,tss_concmg,tp_ag_kgyr,tp_yr_avg_,tn_yr_avg_,tn_ag_kgyr,tss_natura,tp_pt_kgyr,tn_natural,areaha,tp_tot_kgy,tn_urban_k,tp_natural,tp_riparia,tn_tot_kgy,tss_ag_kgy,tss_rip_kg
```

kellyi:
Pushed another fixup which handles renaming some of the columns for the CSV files. To make this happen I created a utility function which generates a new object which renames the keys provided in a `nameMap`, while leaving the non-named keys intact. In tandem, I updated some of the downloadCSV methods to include a `nameMap` object which pairs a key representing the key to replace with a value representing the desired new column name.

I think in theory we could put these new column names in a constants file somewhere, but practically speaking it might be nicer to have them declared closer to where they're actually going to be used for debugging purposes.

I generally tried to keep the column names in the same case as what was already being output: some are PascalCase; others are snake_case, and since the code doesn't change the names of all the keys, it seemed to make sense to keep the new column names harmonious with the existing ones.

Also: I still need to do a bit more investigation to figure out whether the names for the water quality columns are accurate.

arottersman:
Taking a look now

src/mmw/js/src/analyze/views.js
```diff
@@ -681,6 +681,86 @@ var AnalyzeResultView = Marionette.LayoutView.extend({
...
+                nameMap = {
+                    'tss_urban_': 'total_suspended_solids_urban',
+                    'tn_riparia': 'total_nitrogen_riparian',
+                    'tn_pt_kgyr': 'total_nitrogen_point_source_kilograms_per_year',
+                    'tp_urban_k': 'total_phosphorus_urban_kilograms_per_year',
+                    'tss_tot_kg': 'total_suspended_solids_kilograms_per_year',
+                    'geom': 'geometry',
+                    'nord': 'nord',
+                    'tss_concmg': 'total_suspended_solids_concentration_milligrams',
+                    'tp_ag_kgyr': 'total_phosphorus_agricultural_kilgrams_per_year',
```
kellyi: Sounds good to me!

kellyi:
re: TR-55 Water Quality: Looks like runoff's used as part of calculating average concentration on the  client side, so I'll update the CSV output for that tab to echo what's displayed in the chart.

kellyi:
Looks like this is the place where we parsed out the water quality table names most thoroughly, so I'll update the csv output to roughly match:

screen shot 2017-03-23 at 11 52 21 am

src/mmw/js/src/modeling/tr55/quality/views.js
```diff
@@ -86,11 +86,28 @@ var ResultView = Marionette.LayoutView.extend({
...
     downloadCSV: function() {
         var data = this.model.get('result').quality,
+            aoiVolumeModel = new AoiVolumeModel({
+                areaOfInterest: this.options.areaOfInterest }),
             prefix = 'tr55_water_quality_',
             timestamp = new Date().toISOString(),
             filename = prefix + timestamp;
 
-        utils.downloadDataCSV(data, filename);
+        var adjustedData =  _.map(data, function(row) {
```
kellyi: I kept this fn here rather than moving it to a constants or utils file since the calculations need some methods from the `aoiVolumeModel` available in this file.

kellyi:
I squashed everything down to 2 commits; I *think* this is ready for another look.

src/mmw/js/src/analyze/constants.js
```diff
@@ -0,0 +1,52 @@
...
+        pointSource: {
+            'kgp_yr': 'phosphorus_load_kg_per_yr',
+            'mgd': 'discharge_million_gallon_per_day',
+            'kgn_yr': 'nitrogen_load_kg_per_yr'
+        },
+        waterQuality: {
+            'nord': 'nord',
+            'areaha': 'area_hectares',
+            'geom': 'geometry',
+            'tss_tot_kg': 'total_suspended_solids_kg_per_yr',
```
arottersman: Sorry for the confusion: these all should be `kg_yr_per_ha`

src/mmw/js/src/analyze/constants.js
```diff
@@ -0,0 +1,52 @@
...
+        },
+        pointSource: {
+            'kgp_yr': 'phosphorus_load_kg_per_yr',
+            'mgd': 'discharge_million_gallon_per_day',
+            'kgn_yr': 'nitrogen_load_kg_per_yr'
+        },
+        waterQuality: {
+            'nord': 'nord',
+            'areaha': 'area_hectares',
+            'geom': 'geometry',
```
arottersman: We should remove this column

kellyi:
Going to rebase this on `develop`.

src/mmw/js/src/analyze/constants.js
kellyi: Good catch; I removed it from the data before it gets run through these keys, but I'll cut it here too.

arottersman:
Ready for another look?

kellyi:
Yep!

kellyi:
Thanks your help with this! I'll wait til Monday to merge.

Issue #1743, opened 2017-03-22
Adjust map tools location


mmcfarland:
Move map controls to bottom right, align with layer selector.

As seen on Wireframes.

![screenshot from 2017-03-22 16 19 04](https://cloud.githubusercontent.com/assets/1014341/24218899/60ceba9e-0f1b-11e7-91d3-56fe808d7e88.png)

Issue #1739, opened 2017-03-21
Add ability to upload shapefile or geojson for AoI


caseycesari:
Users should be able to upload a geojson or shapefile (hopefully, just the `.shp` file since we don't need the attributes and the file should only contain a single feature, which lets us drop the index file)    to create an AoI.

Initially, we'll assume the uploaded file is valid (validation will be performed in a later card).

The UI for this will be crude until the new sidebar is implemented.

PR #1758, opened 2017-03-23
Add support for creating AoI via uploaded Shapefile and GeoJSON

caseycesari:
## Overview

Adds support for creating AoIs from uploaded shapefiles and geojson.

The browser DataTransfer and FileReader APIs are used to upload and read
in the file.

At the moment, only the first feature from the uploaded file is used.
This may change in the future.

The UI for this feature is temporary. It will be moved once #1743
is implemented

Tested in Chrome, Safari, Firefox, and IE11.

Connects #1739

### Demo

mmw1

### Notes

There are a couple follow up issues:

- Move the UI to the sidebar and add a file selector input (per the wireframes).
- Add more validation (check that coordinates are geographic, check that the shape is in the U.S., implement a better file type check than looking at the extension). It may be a good idea to create a Backbone model for uploaded data and move the validation there.

## Testing Instructions

I didn't provide any test data because I thought it would be better to test it with more datasets than the ones I created.

Try to upload the following geometries (either GeoJSON or a shapefile), and check that you get the expected results:

- A small polygon; should proceed to analyze
- A very large polygon; should add the shape to the map but display an error reporting that it can't be analyzed because it's too large
- A file that isn't a shapefile or geojson; should get an error that the file type is not supported.
- A file that is larger than 200 MB; should get an error that the file is too large.
- A self-intersecting geometry; should add the shape to the map but display an error reporting that it can't be analyzed because it's self-intersecting.
- A file with multiple features; the first one should be added to the map.

### DIFF
--- a/src/mmw/bundle.sh
+++ b/src/mmw/bundle.sh
@@ -127,6 +127,7 @@ JS_DEPS=(backbone
          lodash
          moment
          nunjucks
+         shapefile
          turf-area
          turf-bbox-polygon
          turf-centroid
--- a/src/mmw/js/src/draw/templates/aoiUpload.html
+++ b/src/mmw/js/src/draw/templates/aoiUpload.html
@@ -1,0 +1,1 @@
+<span id="dropbox">Drop a .shp file here.</span>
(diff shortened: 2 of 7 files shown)


mmcfarland:
Heads up that CI flagged some lint.

caseycesari:
Thanks. Fixing it now.

caseycesari:
CI is passing now.

Issue #1767, opened 2017-03-24
RWD click/snap points persisting across projects


ajrobbins:
If I create a project with an RWD AoI, then launch a different project, the click & snap points from the first project persist in the second. To replicate:
- Log in to MMW
- [If you don't have any project saved] create and name a new project, can be anything
- Create a project using RWD
- Navigate to your first project and zoom/pan so that the click & snap points from your RWD are in view

As you can see here, this is a project based on a County AoI but RWD click & snap points are visible:
![screen shot 2017-03-24 at 10 19 52 am](https://cloud.githubusercontent.com/assets/12868342/24298148/81bddc92-107b-11e7-9796-30ddad201015.png)

kellyi:
I was able to replicate this via the steps described above:

<img width="1021" alt="screen shot 2017-03-24 at 11 44 06 am" src="https://cloud.githubusercontent.com/assets/4165523/24301918/3b2323d0-1087-11e7-909c-147bde651b38.png">

kellyi:
Investigating this a bit more and it looks as if the RWD points are cleared when selecting a new aoi via the draw & boundary selection tools, so I assume that switching between projects skips past a clearing step that we're performing there.

kellyi:
More detail: the points are cleared when the a new drawing option, boundary option, or reset is clicked.

kellyi:
This is the fn getting called there: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/draw/views.js#L825

PR #1771, opened 2017-03-24
Clear prior AOI on opening new project

kellyi:
## Overview

This PR fixes a bug described in #1767 whereby loading a previously saved project immediately after creating an RWD project would not remove the markers placed to designate the RWD clicked point and RWD outlet point

Connects #1767 

## Testing Instructions

 * get and bundle this branch, then view it in the browser
 * login, then create and save new non-RWD project as described in #1767
 * start over and create and save new RWD project, verifying that you see the clicked points
 * click on "My Projects" to open the project menu, then select the first saved non-RWD project
 * when that project loads, verify that the new AOI appears and everything loads as normal & that the RWD points from the previous RWD project are saved on the map
 * verify that you can switch back and forth between projects as before and that you don't see any errors in the console

### DIFF
--- a/src/mmw/js/src/projects/views.js
+++ b/src/mmw/js/src/projects/views.js
@@ -168,6 +168,10 @@ var ProjectRowView = Marionette.ItemView.extend({
 
     openProject: function() {
         App.clearAnalyzeCollection();
+        App.map.set({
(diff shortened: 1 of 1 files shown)


kellyi:
Thanks for taking a look!

Issue #1763, opened 2017-03-23
Feature flags


kdeloach:
Develop strategy to support feature flags. We should be able to override these flags for multiple versions of the application.

Determine if there are any existing packages that we can use. Ideally, we would be able to deploy feature flags via Django settings, so that we could have 2 running stacks with different features enabled (for blue/green deployment).

PR #1774, opened 2017-03-27
ADR 001 Feature Flags

kdeloach:
## Overview

This ADR describes which library we will use for feature gating, the reasoning behind that decision, and the consequences.

Connects #1763

### DIFF
--- a/doc/arch/adr-001-feature-flags.md
+++ b/doc/arch/adr-001-feature-flags.md
@@ -1,0 +1,57 @@
+# 001 - Feature Flags
+
+## Context
+
+We need to add a mechanism to toggle features in MMW so that we can deploy
+multiple implementations from the same codebase. For example, in BiG CZ,
+we'll need to disable Analyze mode and other MMW-specific features such as
+ITSI integration.
+
+Primary deciding factors:
+
+* How will this impact development?
+* How easy will it be to switch from MMW to BiGCZ?
+* How will this impact deployments?
+
+Here are the options that I considered:
+
+### [gutter-django](https://github.com/disqus/gutter-django)
+* Too complicated
+* Not actively maintained
+
+### [gargoyle-yplan](https://gargoyle-yplan.readthedocs.io/en/latest/)
+* Actively maintained fork
+* Documentation not great
+* Can add default values in `settings.py`
+* Flags are stored in database
+
+### [django-waffle](https://waffle.readthedocs.io/en/v0.11.1/)
+* Popular
+* Good documentation
+* Not very active
+* Has JavaScript interface
+* Flags are stored in database
+
+### None of the above (custom code)
+* No database required
+* We already have some conventions in place for custom settings
+
+## Decision
+
+We have decided to follow existing conventions to add custom settings
+to MMW to support BiG CZ instead of adding a feature flipper library.
+
+Because flags will rarely change, and we don't need any advanced features
+beyond swiching boolean flags, the overhead of adding and maintaining another
+dependency may not be worthwhile.
+By storing custom settings in Python instead of the database, we'll avoid
+the need to enhance our deployment process to support loading data
+fixtures. Switching between MMW and BiG CZ will be done by using
+an environment variable.
+
+## Consequences
+
+There may be long-running consequences to disabling entire features using
+flags, no matter which implementation we choose. We'll need to be
+careful to make sure that the work we do on BiG CZ doesn't negatively
+impact the functionality of MMW and vice versa.


doc/arch/adr-001-feature-flags.md
```diff
@@ -0,0 +1,59 @@
...
+* Too complicated
+* Not actively maintained
+
+### gargoyle-yplan
+* Actively maintained fork
+* Documentation not great
+* Can add default values in `settings.py`
+* Flags must be created in Django admin
+
+### django-waffle
```
rajadain: Consider adding links to the projects

kdeloach: Links added.

rajadain:
Had a brief IRL conversation with @kdeloach, where we brainstormed if this could be done with hand-code rather than a library.

Arguments for using a library:

 * More standard, might be useful to learn one because feature flags are a common software engineering use case that we may run in to in other apps.

Arguments against using a library:

 * No clear winner, on the attributes of documentation, active maintenance, and complexity
 * We don't need to support granular feature flagging (only certain roles, only 10% of users, etc)
 * Feature flags are so far unlikely to change for a given project, and are used more to distinguish the different flavors of MMW
 * In the end each library provides a boolean value. The work after that (adding if conditions, decorators) is the same in each case.

Arguments against `django-waffle`:

 * It stores flags in the database. While this can make flipping features very fast, since this doesn't require re-deploying, it can also make it hard to track, since there is no record of the flag being flipped.
 * Database storage also prevents blue-green deployments. Our flags may not be that granular, but this is a useful functionality we may want to keep for the future.

We do something very similar using [`client_settings`](https://github.com/WikiWatershed/model-my-watershed/blob/49a319212abc8a16681ce5e5d1c8b5213d7e5060/src/mmw/apps/home/views.py#L115) for MMW.

If this can be implemented using something as simple as a `BIGCZ_SETTINGS` dictionary in `settings/base.py` that only gets loaded if an environment variable is set, then I would be in favor of that. Switching between MMW and BIGCZ could be as simple as setting or unsetting that environment variable, and we can add that environment variable to our deployment scripts to ensure it is always present for BIGCZ deployments.

kdeloach:
Good summary. I'll create a new revision of the ADR which incorporates these points.

kdeloach:
Thanks again for the feedback. I've updated this ADR.

doc/arch/adr-001-feature-flags.md
```diff
@@ -0,0 +1,57 @@
...
+We have decided to follow existing conventions to add custom settings
+to MMW to support BiG CZ instead of adding a feature flipper library.
+
+Because flags will rarely change, and we don't need any advanced features
+beyond swiching boolean flags, the overhead of adding and maintaining another
+dependency may not be worthwhile.
+By storing custom settings in Python instead of the database, we'll avoid
+the need to enhance our deployment process to support loading data
+fixtures. Switching between MMW and BiG CZ will be done by using
+an environment variable.
```
hectcastro: Here, the environment variable is being used to branch down one set of MMW or BiG CZ logic?

doc/arch/adr-001-feature-flags.md
```diff
@@ -0,0 +1,57 @@
...
+We need to add a mechanism to toggle features in MMW so that we can deploy
+multiple implementations from the same codebase. For example, in BiG CZ,
+we'll need to disable Analyze mode and other MMW-specific features such as
+ITSI integration.
+
+Primary deciding factors:
+
+* How will this impact development?
+* How easy will it be to switch from MMW to BiGCZ?
+* How will this impact deployments?
```
hectcastro: Is the idea that all development for MMW or BiG CZ will occur within this same repository, i.e., no repository forking?

doc/arch/adr-001-feature-flags.md
```diff
@@ -0,0 +1,57 @@
...
+### None of the above (custom code)
+* No database required
+* We already have some conventions in place for custom settings
+
+## Decision
+
+We have decided to follow existing conventions to add custom settings
+to MMW to support BiG CZ instead of adding a feature flipper library.
+
+Because flags will rarely change, and we don't need any advanced features
```
hectcastro: Are there any examples of things that we don't anticipate will change (other than analyze mode and ITSI)?

doc/arch/adr-001-feature-flags.md
kdeloach: Yup, exactly.

doc/arch/adr-001-feature-flags.md
kdeloach: BiG CZ will have a custom sidebar to explore data sources from external APIs. See https://marvelapp.com/2bj76ci.

doc/arch/adr-001-feature-flags.md
kdeloach: Yes. See initial implementation here: https://github.com/WikiWatershed/model-my-watershed/pull/1775

kdeloach:
Thanks.

PR #1775, opened 2017-03-28
Feature gate ITSI and Analyze mode for BiG CZ

kdeloach:
## Overview

Adds the `MMW_BIGCZ_MODE` environment variable to enable/disable certain
MMW features when enabled.

Connects #1763

### Demo

The ITSI login button should not be visible when BiG CZ mode is enabled.

BiG CZ:

screenshot 2017-03-28 at 14 09 56

Drawing a shape should not redirect to Analyze mode in BiG CZ mode:

screenshot 2017-03-28 at 14 21 28

## Testing Instructions

* Enable BiG CZ:
```
MMW_BIGCZ_MODE=1 vagrant provision app
```
* Verify that the ITSI login button is not visible
* Verfy that Draw mode doesn't navigate to Analyze mode after producing a shape
* Disable BiGCZ:
```
MMW_BIGCZ_MODE= vagrant provision app
```

### DIFF
--- a/src/mmw/js/src/core/settings.js
+++ b/src/mmw/js/src/core/settings.js
@@ -4,6 +4,8 @@ var _ = require('lodash');
 
 var defaultSettings = {
     itsi_embed: false,
+    itsi_enabled: true,
+    analyze_enabled: true,
     base_layers: {},
     stream_layers: {},
     coverage_layers: {},
--- a/deployment/ansible/group_vars/all
+++ b/deployment/ansible/group_vars/all
@@ -50,3 +50,5 @@ geop_version: "1.2.0"
 
 nginx_cache_dir: "/var/cache/nginx"
 observation_api_url: "http://www.wikiwatershed-vs.org/"
+
+bigcz_mode: "{{ lookup('env', 'MMW_BIGCZ_MODE') }}"
--- a/deployment/ansible/roles/model-my-watershed.base/defaults/main.yml
+++ b/deployment/ansible/roles/model-my-watershed.base/defaults/main.yml
@@ -19,4 +19,6 @@ envdir_config:
   MMW_STACK_COLOR: "{{ stack_color }}"
   MMW_TILECACHE_BUCKET: "{{ tilecache_bucket_name }}"
   MMW_STACK_TYPE: "{{ stack_type }}"
+  MMW_BIGCZ_MODE: "{{ bigcz_mode }}"
+
   ROLLBAR_SERVER_SIDE_ACCESS_TOKEN: "Unknown"
--- a/scripts/mode.sh
+++ b/scripts/mode.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+# Switch between MMW and BiG CZ modes
+
+set -e
+set -u
+set -x
+
+usage() {
+    set +x
+    echo -n "$(basename "${0}") [OPTION]
+
+Set development mode to MMW or BiG CZ.
+
+Options:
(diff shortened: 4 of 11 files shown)


rajadain:
I was hoping for a `./scripts/mode.sh` like this:

```bash
#!/bin/bash

# Switch between MMW and BIGCZ modes

set -e
set -x

usage() {
    echo -n "$(basename "${0}") [OPTION]

Set development mode to MMW or BIGCZ.

Options:
    mmw     Model My Watershed
    bigcz   BIG Crisis Zone
"
}

case $1 in
    mmw)    vagrant ssh app -c 'echo "" > /etc/mmw.d/env/MMW_BIGCZ_MODE' ;;
    bigcz)  vagrant ssh app -c 'echo 1  > /etc/mmw.d/env/MMW_BIGCZ_MODE' ;;
    help|*) usage; exit 1 ;;
esac
```

which we could use like:

```bash
$ kj mode mmw
```

without having to reprovision the VM to switch between modes. Unfortunately, it doesn't work because the environment variable file is read-only in the VM:

```bash
vagrant@app:~$ ll /etc/mmw.d/env/MMW_BIGCZ_MODE
-rwxr-x--- 1 root mmw 0 Apr 13 14:51 /etc/mmw.d/env/MMW_BIGCZ_MODE*
```

and `sudo` doesn't seem to have an effect either.

rajadain:
In addition to the ITSI flag in the login screen, consider adding a check for the `itsi_enabled` [here](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/app.js#L23) so that ITSI gets disabled throughout the app, not just in the login screen.

```diff
diff --git a/src/mmw/js/src/app.js b/src/mmw/js/src/app.js
index af6b9ce..7a2540b 100644
--- a/src/mmw/js/src/app.js
+++ b/src/mmw/js/src/app.js
@@ -20,7 +20,7 @@ var App = new Marionette.Application({
         this.state = new models.AppStateModel();
 
         // If in embed mode we are by default in activity mode.
-        var activityMode = settings.get('itsi_embed');
+        var activityMode = settings.get('itsi_enabled') && settings.get('itsi_embed');
         settings.set('activityMode', activityMode);
 
         // Initialize embed interface if in activity mode
```

kdeloach:
We could easily add a script to change modes. I was just demonstrating how to do it by using an environment variable, which is what we'll likely use for CI. The trick is to use `tee` to write to the file.

Thanks for the heads up about ITSI. I have no doubt that we'll find many more places that need to be feature gated.

src/mmw/js/src/core/settings.js
```diff
@@ -4,6 +4,8 @@ var _ = require('lodash');
 
 var defaultSettings = {
     itsi_embed: false,
+    itsi_enabled: true,
+    analyze_enabled: true,
```
rajadain: On one hand I like that these are separate top level variables, because there can be multiple reasons for disabling ITSI or Analyze. On the other hand, I'd prefer it if all the BIGCZ variables were grouped together in an object. Curious what others think.

kdeloach:
Added `mode.sh` helper script and disabled ITSI mode throughout app if `itsi_enabled` is set to false.

src/mmw/js/src/core/settings.js
kdeloach: I prefer to keep things flat rather than nested but I'm open to suggestions.

kdeloach:
Thanks.

Issue #1776, opened 2017-03-28
Layerpicker missing styling for open tab


arottersman:
As @jfrankl noticed, the background of the open tab should be white to distinguish it from the tab that has layers active.
![screen shot 2017-03-28 at 2 16 48 pm](https://cloud.githubusercontent.com/assets/7633670/24421868/967ee872-13c5-11e7-9ea5-9e026b230729.png)

Issue #1745, opened 2017-03-23
Allow overlay layer association from Analyze


mmcfarland:
As per p16 of the wireframes, a section is added to analyze pages that have a corresponding layer in the layer control. For example, the `Land` tab should allow toggling on the NLCD.  

* Some tabs don't have a layer (Animals)
* Some tabs may have multiple layers.
    * Water quality is NHD catchment, which can be visualized on TN, TSS or TP
* Some tabs are "obeservations" (Point Source)

https://drive.google.com/drive/folders/0B0hs2mX4ApaLN0xLSjY0cVhDckk

mmcfarland:
@jfrankl I missed this in earlier wireframe reviews.  Can you advise on approach for the point above where there are multiple layer associations for an Analyze tab?  In this case, there are 3 layers for the "Water Quality" tab.

arottersman:
For more context, here are the layers, and what they look like all stacked on top of one another:

![screen shot 2017-03-28 at 10 16 34 am](https://cloud.githubusercontent.com/assets/7633670/24409930/aa23de1c-139f-11e7-943a-8a501ec8ee8a.png)

arottersman:
Also @jfrankl what shade of blue should the "Turn on/Turn off" be?

PR #1777, opened 2017-03-29
Add layer toggles to analyze sidebar

arottersman:
## Overview

The MMW experience is greatly enhanced when you use the overlays. To call them out, we're adding related overlay toggles to each analyze tab (except animals).

#### Wireframe
screen shot 2017-03-29 at 2 52 01 pm

(Note in the above, Jeff gave sign off that the Turn On blue should actually match the link color in the rest of the app)

Connects #1745 
Connects #1776 

### Demo

suj3f3azty

### Notes

- The "Total Area" info at the top of each tab will move eventually, so don't worry about that not matching the wireframe

- For the water quality tab @jfrankl is working on a mockup of a collapsable list of layers (there are three: TP, TSS, TN). Will add a commit to style them accordingly once that's ready 

## Testing Instructions

 * Bundle
 * Test the layerpicker as its model has been refactored
     - DRB layers should be disabled when outside their extent
      - Go through each tab
 * Draw an AOI
 * Go through each tab analyze view and confirm you see the relevant layers for each
* The main layerpicker should be synced with the toggles
 * Pointsource should have a loading message while loading

### DIFF
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -23,6 +23,7 @@ var $ = require('jquery'),
     windowTmpl = require('./templates/window.html'),
     AnalyzeDescriptionTmpl = require('./templates/analyzeDescription.html'),
     analyzeResultsTmpl = require('./templates/analyzeResults.html'),
+    analyzeLayerToggleTmpl = require('./templates/analyzeLayerToggle.html'),
     aoiHeaderTmpl = require('./templates/aoiHeader.html'),
     tableTmpl = require('./templates/table.html'),
     tableRowTmpl = require('./templates/tableRow.html'),
@@ -316,7 +317,7 @@ var TabContentView = Marionette.LayoutView.extend({
 });
 
 var TabContentsView = Marionette.CollectionView.extend({
-    className: 'tab-content',
+    className: 'tab-content analyze',
     childView: TabContentView,
     onRender: function() {
         this.$el.find('.tab-pane:first').addClass('active');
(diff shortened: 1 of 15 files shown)


kellyi:
This seems to be working well! Question about how the point source layer should work.

For all the other layers, it looks like we're only allowing one of them to be on at a time: if you toggle soils on, water quality toggles off, and vice versa and so on. However, this isn't the case for the point source layer, which stays on even if one selects another layer:

screen shot 2017-03-30 at 10 15 30 am

Is this the intended behavior?

src/mmw/js/src/core/layerPicker.js
```diff
@@ -312,63 +241,11 @@ var LayerPickerView = Marionette.LayoutView.extend({
     },
 
     collectionEvents: {
-        'change': 'render'
+        'change': 'render',
+        'change:layer': 'render'
```
kellyi: I'm still working out my mental model of how these events work in Marionette but: would `change:layer` trigger render 2x, as there's a generic change happening then the specific change event? Does the order here matter?

src/mmw/js/src/analyze/views.js
```diff
@@ -344,7 +422,7 @@ var TableRowView = Marionette.ItemView.extend({
             // Scale the area to display units.
             scaledArea: utils.changeOfAreaUnits(area, 'm<sup>2</sup>', units)
         };
-    }
+    },
```
kellyi: Stray comma

src/mmw/js/src/analyze/views.js
```diff
@@ -791,11 +880,13 @@ var AnimalsResultView = AnalyzeResultView.extend({
 });
 
 var PointSourceResultView = AnalyzeResultView.extend({
+
```
kellyi: Extra whitespace

ajrobbins:
@kellyi 
>For all the other layers, it looks like we're only allowing one of them to be on at a time: if you toggle soils on, water quality toggles off, and vice versa and so on. However, this isn't the case for the point source layer, which stays on even if one selects another layer

Yes, I think this is the intended behavior. For example, it might be useful to visualize point source pollution points over water quality loading data, to see potential correlations between the two.

src/mmw/js/src/analyze/views.js
```diff
@@ -331,6 +332,83 @@ var AnalyzeDescriptionView = Marionette.ItemView.extend({
...
+        App.getLayerTabCollection().trigger('change:layer');
+    },
+
+    templateHelpers: function() {
+        if (this.message) {
+            return {
+                message: this.message,
+            };
+        }
+        else if(this.model) {
```
kellyi: Missing space between `if` and `(this.model)`.

arottersman:
Re: Pointsource

That's the intended behavior assuming it should act the same as the layerpicker. All the analyze tabs except Pointsource are showing layers from the coverage layer tab. From the layerpicker we only allow one coverage layer on the map at a time, so turning on one turns off the others. The point source layer is from the observations tab, so it can be on in addition to any coverage layers.

kellyi:
Got it, that makes sense. It seemed odd because the only other layers toggleable in the analysis tab are mutually exclusive, but the layer selector has other non-exclusive options like congressional district boundaries which don't show up in the analysis results.

src/mmw/js/src/core/layerPicker.js
arottersman: `render` should only trigger once because `change:layer` is actually a (poorly named) custom event. Because the layers themselves are nested so deeply in the `LayerTabCollection`, toggling a single layer doesn't fire a regular `change` event on the collection, we have to trigger it "manually".

I'll rename it to something like `toggle:layer`

kellyi:
Encountered a bug: if you move to model, then move back to analyze, and try to click on a layer option in the analyze tab, the console shows this error when you try to toggle it on in either the analyze tab or the layer selector:

screen shot 2017-03-30 at 10 48 56 am

arottersman:
I also encountered that bug [here](https://github.com/WikiWatershed/model-my-watershed/pull/1777/files#diff-fa3796f2e3f8773b7a5d1e15f346e1d9R104) in this PR. I think it's because both parents and their children are subscribing to layer changes, so the parents render and destroy their children, who also try to render, but can't because they've been destroyed.

arottersman:
Added some fixups for the lint, changing `change:layer` to `toggle:layer` and the view destroyed bug. Ready for another look

kellyi:
Taking another look at this.

kellyi:
This is working well now!

One more quick question about the behavior:

If the layer selector control's partially scrolled down the list of results for whatever layer type, clicking on a layer in the Analyze tab will cause the layer selector list to scroll back up to the top. If you select a layer in the layer selector control, it doesn't scroll like this.

Is this worth addressing here?

arottersman:
Rats, that's happening because `toggle:layer` triggers render from `LayerPickerView` down, whereas layer change events from within the layerpicker trigger render from just the `LayerPickerGroupView` down.

Unfortunately we have to have some sort of listener on `LayerPickerView` because we need the tab icons to get their proper active color. I will see if there's any easy solution to this when I squash my fixups, and let you know.

arottersman:
Okay, I think I fixed it with b73f798!

kellyi:
Going to check this out again, then squash and merge once it's confirmed to work!

kellyi:
Tested in IE, Firefox, Safari, and Chrome. There's also nothing to squash, so I'm going to merge this one.

Issue #1735, opened 2017-03-21
Sidebar Update: Move draw


mmcfarland:
As per the new wireframes, p11 - move the AoI selection tools from the map into a new sidebar panel, that is linked to from the welcome panel created in #1734.  For the purposes of this card, ignore the "upload file" option.

https://drive.google.com/drive/folders/0B0hs2mX4ApaLN0xLSjY0cVhDckk

Issue #1760, opened 2017-03-23
Implement UI for shapefile/geojson upload


caseycesari:
Move the UI to the sidebar and add a file selector input (per the wireframes, once  #1743 is in).

We'll also need a loading spinner, though most tests I have done are instantaneous.

We should also add an area in the UI where we can list the requirements:
- Must be a shapefile or geojson
- Less than 200 MB
- Only the first feature is used.

caseycesari:
After some feedback from the client, we'll also need to make changes to how this works.

Instead of accepting a `.shp` file, we should accept only a zipfile that contains all of component parts that make up a shapefile. 

For now, just unzip the file and add the shape to the map as before.

We'll still be accepting geojson files directly.

caseycesari:
This may be useful: https://gildas-lormeau.github.io/zip.js/

arottersman:
@caseycesari any reason you suggest `zip.js` over `JSZip` https://www.npmjs.com/package/jszip ? It seems like `zip.js` is a little better managed and more actively developed, but there's no npm package.

caseycesari:
Nope! They both seem like they are worth a try.

Issue #1749, opened 2017-03-23
Tool tip helps with links


mmcfarland:
We've been unable to add links to the help tooltips since they persist only on hover (and likewise difficult to take a screenshot of for this issue).

Update the usage of the library (or provide an alternative) so that links can be clicked within the tooltip.  Keep in mind that the draw tools will be moved in #1735.  The other usage is as legends in the new layer selector.

kellyi:
Shares a point with #1747

arottersman:
- Are there actual links that should go into the app with this card? 
----
- It seems to make a link clickable on these popovers while still making them show on hover we'd have to have the tooltip/popover also count as a hover target, ie the element closes if you leave either the original tooltip icon, OR the tooltip itself. That way you could very carefully move your mouse over to the tooltip to click the link
- Another very easy alternative is just to change the tooltips to show/hide on `click` instead of `hover`, which I assume already happens for mobile

Which would be preferred?

mmcfarland:
I think my preference would be for click, unless "careful" was actually pretty generous while moving the mouse.

We do have actual links that were suggested to us, I'll try to dig them up.

mmcfarland:
Original request:
> Add the following text and link to the bottom of help text in all of the "?" help balloons.
"Info and help for these tools at http://wikiwatershed.org/model/"

arottersman:
Putting this back in the queue because it seems pretty dependent on how https://github.com/WikiWatershed/model-my-watershed/pull/1778 turns out.

Also found https://github.com/WikiWatershed/model-my-watershed/issues/1233 for more links we can add.

ajrobbins:
I asked @aufdenkampe for text for links, but we likely wont have it for another week or so.

PR #1778, opened 2017-03-29
Move draw tools from map to pane

kellyi:
## Overview

This PR moves the draw tools from the map onto a sidebar as displayed in the wireframes linked from #1735 

The styling currently roughly matches what's in the wireframes, and the image below shows the behavior. 

Succinctly: clicking to toggle a draw tool active de-activates and closes any others which are open, and clicking again on an active draw tool deactivates it. The x/close buttons call the reset action.

Connects #1735 

### Demo

mmw-draw-pane

### Notes

The new DrawWindow view mostly supplants the old ToolbarView; the only reason that view has been kept around now is so that we can keep the shapefile upload button there. #1735 suggested not bothering to change it, but if we'd like I can move it -- and thereby remove the rest of the ToolbarView.

Tagging @jfrankl for answers to the following Qs:

* should we keep the tooltips for each item in the lists?
* how should these lists be styled?

## Testing Instructions

 * get this branch, then `./scripts/bundle.sh --vendor` and `./scripts/bundle.sh`
 * visit localhost:8000 and hard refresh
 * when the draw pane loads verify that all the controls still work (note: the Shapefile upload section's currently not active per note in #1735, and I left the old toolbar version on the map).
 * move back and forth between modeling, analyze, and draw, and verify that the tools continue to work
 * verify that the x/ reset button resets the drawing state
 * verify that the rest of the app still works as expected

### DIFF
--- a/src/mmw/js/src/draw/templates/window.html
+++ b/src/mmw/js/src/draw/templates/window.html
@@ -1,2 +1,28 @@
 <h1>Select Area</h1>
-<p>Draw tools go here</p>
+<p id="draw-pane-header-description">
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit.
+    Duis ac pharetra nibh. Nullam gravida massa non egestas
+    pellentesque.
+</p>
+<div id="draw-pane-buttons">
+    {% for drawTool in drawTools %}
+        <div class="btn btn-lg btn-block draw-tool-button" type="button"
+            data-toggle="dropdown" aria-expanded="true"
+            id={{ drawTool.button }}>
+            <p class="draw-button-title pull-left">
+                {{ drawTool.name }}
+            </p>
+            <span class="draw-button-icon caret pull-right"></span>
+            <br />
+            <p class="draw-button-subtitle pull-left">
+                Lorem ipsum dolor sit amet, consectetur
+            </p>
+        </div>
+        {% if activeDrawTool == drawTool.id %}
+            <div class="draw-tool-button-slideout" aria-expanded="true">
+                <i class="fa fa-times pull-right reset-draw-button"></i>
+                <div id={{ drawTool.region }}></div>
+            </div>
+        {% endif %}
+    {% endfor %}
+</div>
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -22,8 +22,8 @@ var $ = require('jquery'),
     windowTmpl = require('./templates/window.html'),
     aoiUploadTmpl = require('./templates/aoiUpload.html'),
     drawTmpl = require('./templates/draw.html'),
-    resetDrawTmpl = require('./templates/reset.html'),
     delineationOptionsTmpl = require('./templates/delineationOptions.html'),
+    uploadFileTmpl = require('./templates/uploadFile.html'),
     settings = require('../core/settings'),
     modalModels = require('../core/modals/models'),
     modalViews = require('../core/modals/views');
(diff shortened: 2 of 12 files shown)


src/mmw/js/src/draw/views.js
```diff
@@ -136,6 +136,85 @@ var DrawWindow = Marionette.ItemView.extend({
...
+            activeDrawTool: this.model.get('selectedDrawTool'),
+        };
+    },
+
+    initialize: function() {
+        this.model.set('selectedDrawTool', null);
+    },
+
+    onShow: function() {
+        this.render();
```
arottersman: Consider dropping this function and having

```javascript
modelEvents: {
    'change': 'render'
},
```
`onShow` is called after `render` and is usually used to show child views.

src/mmw/js/src/draw/views.js
```diff
@@ -131,11 +130,141 @@ function validatePointWithinDataSourceBounds(latlng, dataSource) {
...
+        map.addLayer(ofg);
+        this.rwdTaskModel = new models.RwdTaskModel();
+    },
+
+    onShow: function() {
+        var drawTools = settings.get('draw_tools'),
+            activeDrawTool = this.model.get(this.constants.selectedDrawTool);
+
+        // Update the view to mount the proper region divs
+        this.render();
```
arottersman: I deleted a comment on an earlier commit, but something still seems fishy about calling `render` in `onShow`. I think if you call `render` on the `modelChange` event it will call before `onShow`, and you won't need to call `render` again here.

kellyi: That makes sense -- I'll experiment with this to see if there's a better way to do it. The things that need to happen here are:

- update the active draw tool in the model
- add the region div for the active draw tool to the DOM
- call show on the region div to render the active draw tool

But it's likely this could get restructured to do that without calling `render` in `onShow`.

kellyi: Investigated this and it seems like there's no obligation to use the `onShow` method as the place for doing the steps described above, so I'm going to rename it to something like `updateTools` and keep the call to `this.render()` in it.

kellyi:
Still working on updating the tests & still need to verify the appearance of the draw tool line items but the core functionality's in place.

src/mmw/js/src/draw/templates/window.html
```diff
@@ -1,2 +1,99 @@
 <h1>Select Area</h1>
-<p>Draw tools go here</p>
+<p id="draw-pane-header-description">
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit.
+    Duis ac pharetra nibh. Nullam gravida massa non egestas
+    pellentesque.
+</p>
+<div id="draw-pane-buttons">
```
arottersman: Since these are all pretty similar, consider passing the view the `DrawTools` and generating these based on them:

```javascript
{% for drawTool in drawTools %}
    <div class="btn btn-lg btn-block draw-tool-button" type="button"
         data-toggle="dropdown" aria-expanded="true"
         id={{ id based on drawTool }} >
        <p class="draw-button-title pull-left">
            {{ drawTool }}
        </p>
        {% if activeDrawTool != drawTool %}
            <span class="draw-button-icon caret pull-right"></span>
        {% endif %}
        <br />
        <p class="draw-button-subtitle pull-left">
            Lorem ipsum dolor sit amet, consectetur
        </p>
    </div>
    {% if activeDrawTool == drawTool %}
      <div class="btn btn-lg btn-block draw-tool-button-slideout" type="button"
            data-toggle="dropdown" aria-expanded="true">
             <i class="fa fa-times pull-right reset-draw-button"></i>
             <div id={{region id based on drawTool }}></div>
             </div>
     {% endif %}
{% endfor %}
```

kellyi:
I *think* the X/close button will end up having to reset all the tools, as I don't think more than one tool could be active simultaneously.

src/mmw/js/src/draw/templates/window.html
kellyi: Yeah, that looks much better, so I'll push a fixup with this change!

kellyi: Pushed a new commit to loop over a config object to render these buttons.

kellyi:
I think the links need to be disabled when drawing's happening, so I'll push another commit with that change.

kellyi:
Made some fixes to address the comments above and to bring the style more in tune with the wireframes. Currently looks & works like this:

mmw-draw-tool-fixups

Some notes on things I think we should address elsewhere once the draw tool move is in place:

- I still did not move the actual upload file tool to the pane, but did create a placeholder template for it just so this could loop over all the tools.
- I left the tooltips, but they currently work now. I believe we've already got a card for this one.
- clicking the X to reset now resets the tools, closes the pane, and displays an error. The error happens in the current version of the app, too.
- the boundary tools aren't currently disabled when the draw tools are active -- which also replicates the behavior of the existing app. 

One remaining thing to do here is to disable the boundaries if the map is not zoomed to a valid level.

kellyi:
> clicking on a tool doesn't appear to do anything unless you know how to use it on the map. There could be some sort of active styling on the selected tool to alleviate this

I'm going to spin this off to another issue as it's probably better handled separately from moving the tools.

kellyi:
Rebased on `develop`

kellyi:
Pushed another commit for this which addresses this:

> One remaining thing to do here is to disable the boundaries if the map is not zoomed to a valid level.

The way this works is that the SelectAreaView has an initial check performed when it is first set up to compare each boundary layer's zoom level with the then-current map zoom level. The updated `actOnUI` method then uses jQuery to enable/disable the buttons when the zoom level changes.

This works well enough for now as part of the move, but it would be nice to have the enabled/disabled layers stored as a globally available collection so the layerpicker had access to them, too -- the way the app was just updated to handle the raster and water quality overlays in both places.

If we decide we want to do that, we should make another issue for it as it involves some refactoring.

kellyi:

I'll create the issues mentioned on this card this afternoon and link them here so we can make sure everything's covered.

kellyi:
I made #1784 and #1785 from the additional issues noted here.

#1760 exists to address moving and implementing the upload file option.

#1749 is there for updating the tooltips & tooltip behavior.

kellyi:
Thanks for the help with this! Going to merge.

Issue #1765, opened 2017-03-23
Boundary layer search endpoint


kdeloach:
Create an endpoint which performs a full text search against boundary layer names (ex. HUC 8). The endpoint should accept a single query value and return a limited number of results as JSON.

ajrobbins:
We have several types of boundaries in MMW. For this full text search, we will likely only be enabling search by HUCs (HUC 8s, 10s, and 12s).

PR #1779, opened 2017-03-30
Boundary layer search endpoint

kdeloach:
## Overview

Create a new API endpoint which performs a full text search against HUC boundary layers.

Connects #1765

### Demo

image

### Notes

You need HUC boundary layers loaded locally to test this. This can take a long time. I recommend updating `layer_settings.py` to test searching against school boundary layers instead. These don't take much time to load.

## Testing Instructions

* Download boundary layers
```
vagrant ssh app
/vagrant/scripts/aws/setupdb.sh -b
```
* Update `layer_settings.py` (if you are missing any HUC boundaries)
* Exercise search endpoint:
```
http://localhost:8000/api/modeling/boundary-layers-search/?text=penn
```

### DIFF
--- a/src/mmw/apps/modeling/views.py
+++ b/src/mmw/apps/modeling/views.py
@@ -15,7 +15,7 @@ from django.shortcuts import get_object_or_404
 from django.utils.timezone import now
 from django.conf import settings
 from django.db import connection
-from django.contrib.gis.geos import GEOSGeometry
+from django.contrib.gis.geos import GEOSGeometry, WKBReader
 from django.http import HttpResponse
 from django.core.servers.basehttp import FileWrapper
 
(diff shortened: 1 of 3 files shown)


kdeloach:
Updated so results now contain lat/lng. The preview screenshot is out of date.

scripts/aws/setupdb.sh
```diff
@@ -69,7 +69,10 @@ envdir /etc/mmw.d/env /opt/app/manage.py migrate
 
 function download_and_load {
     for f in "${FILES[@]}"; do
-        curl -s $FILE_HOST/$f | gunzip -q | psql --single-transaction
+        rm -f tmp.sql tmp.sql.gz
```
rajadain: Any reason to prefer storing a file rather than piping?

Also, does this mean that the last call to `download_and_load` will leave a `tmp.sql` file in the directory?

kdeloach: The HUC boundary layers are several hundred MB so piping from curl to gunzip then psql results in "out of memory" exceptions on my machine. And that error only happens after a considerable amount of time. I don't mind reverting this commit.

rajadain: Okay, that makes sense.

Can we make it something like:

```bash
if [ -ne $f.sql ]; then
    curl ... && gunzip ... && psql .. && rm -f $f.sql $f.sql.gz
fi;
```

so that in case there's an error in one of these steps, the file doesn't have to be redownloaded the next time?

kdeloach: I think there's several ways we can enhance this to make it more robust but I'd like to do that in a separate PR. The only reason I included this change is in case somebody needs to populate their workstation with boundary layers and runs into the same problems that I encountered. It probably makes more sense to just revert this since I don't know what impact it will have on deployments.

src/mmw/apps/modeling/views.py
```diff
@@ -526,7 +532,90 @@ def boundary_layer_detail(request, table_code, obj_id):
...
+    Return raw SQL query to perform full text search against
+    all searchable boundary layers.
+    """
+    select_fmt = """
+        SELECT
+            id,
+            '{code}' AS code,
+            name,
+            {priority} AS priority,
+            ST_Centroid(ST_Transform(geom, 4326)) as center
```
arottersman: I'm getting a `Input geometry has unknown (0) SRID` error...it seems like `geom` might not be defined.

src/mmw/mmw/settings/layer_settings.py
```diff
@@ -197,6 +197,8 @@
                         'HUC-8 to analyze.',
             'minZoom': 7,
             'selectable': True,
+            'searchable': True,
+            'search_priority': 100,
```
arottersman: Where did these priorities come from? Would it do just as well to sort these alphabetically on `code`?

src/mmw/apps/modeling/views.py
kdeloach: Curious how you have null geoms in your database. May want to reload your boundary layers. I'll filter out null geoms just to be safe.

src/mmw/mmw/settings/layer_settings.py
kdeloach: I think these should be sorted by relevance. Since we're not sure which boundaries are most relevant yet, I decided to set the sort order explicitly, which seems like the most flexible approach. This initial configuration prioritizes small/local boundaries over large ones.

src/mmw/apps/modeling/views.py
kdeloach: Oh, I misread this comment. I'm going to fix this by removing `ST_Transform` which was a little ceremonious since these boundaries should be imported as 4326 anyway.

kdeloach:
Removed `ST_Transform` from query.

scripts/aws/setupdb.sh
kdeloach: Removed this commit.

arottersman:
Working for me now: 

```http
GET /api/modeling/boundary-layers-search/?text=sch
HTTP 200 OK
Vary: Accept
Allow: OPTIONS, GET

[
    {
        "code": "huc8",
        "text": "Deschutes",
        "label": "HUC-8 Subbasin",
        "priority": 100,
        "y": 46.86230369360085,
        "x": -122.63864858148362,
        "id": 1577
    },
    {
        "code": "huc8",
        "text": "Little Deschutes",
        "label": "HUC-8 Subbasin",
        "priority": 100,
        "y": 43.52048026543526,
        "x": -121.61679787603616,
        "id": 1513
    },
    {
        "code": "huc8",
        "text": "Lower Deschutes",
        "label": "HUC-8 Subbasin",
        "priority": 100,
        "y": 45.01515291504721,
        "x": -121.23695421645023,
        "id": 1517
    },
    {
        "code": "huc8",
        "text": "Schoharie",
        "label": "HUC-8 Subbasin",
        "priority": 100,
        "y": 42.52008640312507,
        "x": -74.3571437195935,
        "id": 1728
    },
    {
        "code": "huc8",
        "text": "Schuylkill",
        "label": "HUC-8 Subbasin",
        "priority": 100,
        "y": 40.39343175506416,
        "x": -75.7706892535944,
        "id": 1748
    }
]
```

Issue #1762, opened 2017-03-23
Display boundary layer search results


kdeloach:
Update the geocoder component to query the boundary layers search endpoints and combine results in the autocomplete dropdown. Clicking the result in the dropdown should display the feature on the map and zoom to the boundary layer extent.

Page 9 of [2017_03_16_mmw-v6.pdf](https://drive.google.com/open?id=0B0hs2mX4ApaLUlhaaWNYTE5CMEk) has a mock up of how the geocoder results dropdown should change as a result of this feature.

![screenshot 2017-03-23 at 17 39 03](https://cloud.githubusercontent.com/assets/43062/24271463/a3cddb2a-0fef-11e7-8c85-b96567bb235f.png)

kdeloach:
Depends on #1765

kdeloach:
For the initial iteration, could we just limit the total number of boundary suggestion to 5 or 10, instead of adding a "view more" link to the search box results?

![image](https://cloud.githubusercontent.com/assets/43062/25093579/d9cc4bc4-2360-11e7-8ed6-ea7a789bf8c6.png)

@jfrankl ^

ajrobbins:
>For the initial iteration, could we just limit the total number of boundary suggestion to 5 or 10, instead of adding a "view more" link to the search box results?

I think it's fine to limit to 5 results for this iteration.

PR #1789, opened 2017-04-05
Display boundary layer search results

kdeloach:
## Overview

Display boundary layer search results in the geocoder autocomplete and the map.

Connects #1762

### Demo

image

## Testing Instructions

 * Configure searchable boundary layers in `layer_settings.py` depending on which boundary layers you have on your workstation
 * Run `./scripts/bundle.sh`
 * Verify that boundary layers and geocoded results show up in the autocomplete
 * Verify that picking geocoded suggestions still work
 * Verify that picking boundary layer suggestions adds the layer to the map

### DIFF
--- a/src/mmw/apps/modeling/views.py
+++ b/src/mmw/apps/modeling/views.py
@@ -572,10 +572,11 @@ def _get_boundary_search_query(search_term):
             'ORDER BY priority DESC, name LIMIT 5').format(subquery)
 
 
-def _boundary_search(search_term):
+def _do_boundary_search(search_term):
     """
     Execute full text search against all searchable boundary layers.
     """
+    result = []
     query = _get_boundary_search_query(search_term)
 
     with connection.cursor() as cursor:
@@ -584,7 +585,6 @@ def _boundary_search(search_term):
 
         wkb_r = WKBReader()
 
-        result = []
         for row in cursor.fetchall():
             id = row[0]
             code = row[1]
@@ -594,7 +594,6 @@ def _boundary_search(search_term):
 
             layer = _get_boundary_layer_by_code(code)
 
-            # Field names should match the ArcGIS API suggest endpoint fields
             result.append({
                 'id': id,
                 'code': code,
@@ -604,7 +603,17 @@ def _boundary_search(search_term):
                 'y': point.y,
                 'x': point.x,
             })
-        return result
+
+    return result
+
+
+def _boundary_search_context(search_term):
+    suggestions = [] if len(search_term) < 3 else \
+        _do_boundary_search(search_term)
+    # Data format should match the ArcGIS API suggest endpoint response
+    return {
+        'suggestions': suggestions,
+    }
 
 
 @decorators.api_view(['GET'])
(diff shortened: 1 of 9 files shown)


kdeloach:
Updated the unit tests. The only remaining thing to do is to group boundary layers together in the autocomplete dropdown to match the mocked up design.

ajrobbins:
Here is an updated wireframes of the grouped boundary layers; when there are more than 2 results in the HUC subsections, a "view more" link should appear that would expand the results to 5 total.
segmented_search_results_1
segmented_search_results_2

This can be spun off into another card, if that makes more sense to you @kdeloach .

kdeloach:
I'll group boundary layers together but would like to do the "view more" link in a separate PR.

kdeloach:
This is ready for review.

arottersman:
@rajadain won't be back until Thursday :'(

kdeloach:
Ah, thanks for the heads up. Reassigned.

arottersman:
Overall the grouping is working, but it's disturbingly slow on my machine. It exposed some weird behavior:

qlon2dxgku

(If you watch the gif long enough you'll see results return) The incremental requests don't cancel, so despite `schuykilll` sitting in the input, you'll first see results for the letter `s`, then they'll disappear to show results for the `sch`, then again `schuykill`.

I wonder if we should:
1. Cancel requests as each new one comes in 
1. Show the results of geocoder and boundary searches independently; if geocoder results come in first, show them, and then add the boundary results to the view as soon as they've returned

kdeloach:
I'll look into cancelling stale requests. The reason that I decided not to show results independently is because it breaks all the geocoder tests (which expect a linear sequence of actions) and I couldn't think of an elegant way to rewrite them. I'll see if there's a way I can hack it without breaking the unit tests.

kdeloach:
I'm also going to add a minimum character limit on the boundary search endpoint since doing a full text search for a single letter is pointless.

arottersman:
Sounds good! Should we make a separate issue for what happens once you get to draw?

kdeloach:
Could you elaborate on that?

kdeloach:
Unit tests are passing on Jenkins.

arottersman:
During RRP we talked about how it's not clear what should happen after selecting a boundary result...currently it displays an AOI you can't intuitively select to take to analyze/model. 

Some possibilities I think were
- go right to analyze/model with the shape
- turn on the corresponding boundary select tool for the search-selected shape

kdeloach:
My recollection is that we decided to keep the existing behavior to zoom to the boundary location. You can use the "Select by boundary" draw tool to analyze it if you want.

arottersman:
My bad  — I think I selectively missed that part of the conversation

kdeloach:
Updated with these changes:

* Cancel pending requests before making new ones
* Display geocode/boundary results independently
  * This may be hard to see since we don't clear the previous results before fetching new ones
* Better "loading" indication

arottersman:
Cool, this is working pretty naturally now!

 One small, buggish thing is the `working` indicator (the spinner) turns off when the first results return:

(I had "dog" in the bar, and it's now working on "schuy" but you can't tell)
screen shot 2017-04-18 at 2 55 00 pm

Consider keeping the spinner until all polling completes.

kdeloach:
Fixed a bug where the loading icon state was not always consistent, and another bug where results would not appear until both requests completed initially.

kdeloach:
Thanks.

Issue #1233, opened 2016-04-08
Consider changing tooltip behavior to accommodate a clickable link


ajrobbins:
As described in #1210 , we were unable to accommodate Stroud's request for a link to be added to the "?" tooltips, because they disappear when the cursor moves away from the "?" icon. We should consider modifying the tooltips to not disappear when the cursor moves onto the tooltip.

Original request:

> Add the following text and link to the bottom of help text in all of the "?" help balloons. "Info and help for these tools at http://wikiwatershed.org/model/"

ajrobbins:
@dtarb suggested that we could incorporate some background on RWD methodology in the "?" tooltips for the RWD options:
![screen shot 2016-12-06 at 9 34 57 am](https://cloud.githubusercontent.com/assets/12868342/20929666/c8f68830-bb98-11e6-9836-7eacdb6e03b6.png)

If these tooltips remained onscreen when the user hovers over them, we could add a "Learn More" link with some additional description of methodology:

*Suggested additional text for DRB RWD option:*
>This moves downhill from the point you click to snap onto the nearest point on the Delaware high resolution stream network and calculates the watershed upstream of this point using a 1/3 arc second (10) m resolution digital elevation model for the Delaware River Basin.  The stream network snapped to was delineated using TauDEM (http://hydrology.usu.edu/taudem).
 
*Suggested additional text for NHD-based RWD option (currently Mid-Atlantic)*
>This moves downhill from the point you click to snap onto the nearest point on the medium resolution National Hydrography Dataset (NHDPlus flowline) and calculates the watershed upstream of this point using the 30 m resolution NHDPlus flow direction grid for the continental US.  If the point you click does not have a NHDPlus flowline downstream of it (e.g. is in an internally draining area) the watershed is calculated from the point you click.  Watersheds calculated may have holes in them due to internally draining areas that do not connect to the outlet in the NHDPlus flow direction dataset.  To learn more about NHDPlus see  http://www.horizon-systems.com/nhdplus/.  The model my watershed watershed delineation uses NHDPlus version 2.1 data model with latest content version accessed 11/22/16.
 
*Rationale from @dtarb*
[...]

PR #1791, opened 2017-04-06
Activate tooltips on click instead of hover

caseycesari:
## Overview

Tooltips are now activated on click instead of hover so that links in the tooltip text are now clickable. Previously, the tooltip closed as the pointer was moved from the tooltip icon to the tooltip text, preventing the user from clicking on any links in the text.

To better unify the UI/UX of the app, the tooltips in the legend were updated to behave in the same way. Additionally, the style of both tooltips was made consistent.

Links were added to the RWD tools. Also, a general link to documentation about the various draw tools was added to the sidebar instead of each draw tooltip (as previously requested). This seemed like a more appropriate solution now that there is a space for explanatory text near the tools.

Connects to #1749
Connects to #1233 

### Demo

mmw7

## Testing Instructions

- View the tooltips for various draw tools, and verify that they open on click and not hover.
- Try to click the links in the RWD tooltips.
- Verify the legend tooltips behave in the same way as the draw tool tooltips.

### DIFF
--- a/src/mmw/js/src/core/layerPicker.js
+++ b/src/mmw/js/src/core/layerPicker.js
@@ -80,7 +80,7 @@ var LayerPickerLayerView = Marionette.ItemView.extend({
 
     onRender: function() {
         this.ui.layerHelpIcon.popover({
-            trigger: 'hover',
+            trigger: 'focus',
             viewport: {
                 'selector': '.map-container',
                 'padding': 10
--- a/src/mmw/js/src/core/templates/layerPickerLayer.html
+++ b/src/mmw/js/src/core/templates/layerPickerLayer.html
@@ -8,10 +8,9 @@
     {% if not isDisabled and legendMapping %}
         <a
             tabindex="-1"
-            role="presentation"
+            role="button"
             data-html="true"
             data-toggle="popover"
-            data-trigger="focus"
             class="layerpicker-info"
             data-layer-help-icon="{{layerDisplay}}"
         >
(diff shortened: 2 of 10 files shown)


rajadain:
Currently the user has to click outside the popup to make it disappear. This can sometimes be tricky, because a user may accidentally click an active element (on the map, a button, or an accordion header) that affects the UI inadvertently.

How difficult would it be to close the popup on re-clicking the help icon (in addition to, rather than instead of, the current behavior)? I suppose there will need to be some way of tracking state, which could get messy, but perhaps not since the each popup parent has a reference to its popup with state.

ajrobbins:
What about putting an X in the popup to close? I think from a user perspective clicking outside of the popup to close it is also expected, but it would be helpful to have an explicit close button too.

rajadain:
Just did a quick read of [bootstrap popover](http://getbootstrap.com/javascript/#popovers), doesn't seem like either case is natively supported. Between static and dismissable popups, I like the dismissable ones (that are currently being used here) better. There's a slightly ugly hack here http://stackoverflow.com/a/13413660/6995854 which I'm not sure will work, but might be worth a shot.

caseycesari:
Adding the close button would be a bit of a hack, but requiring that users click the help icon again to close it is built in to Bootstrap and would be easy to implement. See the example here: http://getbootstrap.com/javascript/#live-demo-1

caseycesari:
One downside of only using the help icon to open and close the tooltip is that it's then possible to open multiple tooltips at once:

screen shot 2017-04-06 at 12 08 30 pm

rajadain:
I'd be okay with keeping it as it is for now, and making another card to enhance popups with a close button.

caseycesari:
Created an issues for follow-up since any of the suggested changes will require some work: https://github.com/WikiWatershed/model-my-watershed/issues/1793

Issue #1748, opened 2017-03-23
Add minimize/collapse button to layer selector


ajrobbins:
Currently, clicking on the active layer type icon collapses the layer selector into its bottom bar. We've gotten reports from Stroud that this is not discoverable enough, so we should consider adding an explicit minimize button to the layer selector.

ajrobbins:
@jfrankl thoughts?

jfrankl:
I think adding a caret to the layer selector would make the collapse functionality more discoverable:

![image](https://cloud.githubusercontent.com/assets/1809908/24406310/0224e200-1396-11e7-8eb9-a2eadbbb9246.png)

In this case, clicking on the caret or "Coverage" would collapse the layer selector. Clicking on the active icon would continue to collapse the layer selector.

ajrobbins:
@jfrankl I think that the caret makes it kind of look like a collapsible list, like we have on TNC. I would expect clicking the caret to collapse the list items into "coverage", not necessarily collapse the whole layer selector. Could we do a caret or other minimize icon at a level above the "coverage" title?

jfrankl:
It's going to make the layer picker take up more space. Worth trying?

ajrobbins:
Yes, I think the tradeoff is worth it if it makes it easy to minimize completely.

jfrankl:
Sounds good. Here are two directions—please send feedback.

### Variation 1
#### Open
![open](https://cloud.githubusercontent.com/assets/1809908/24408326/56f735ee-139b-11e7-832f-f81fb2187297.png)
#### Closed
![closed](https://cloud.githubusercontent.com/assets/1809908/24408327/56f8a442-139b-11e7-9dda-1c95f52706d2.png)

### Variation 2
#### Open
![open](https://cloud.githubusercontent.com/assets/1809908/24408335/5f5f0b3a-139b-11e7-8952-64a905800587.png)
#### Closed
![closed](https://cloud.githubusercontent.com/assets/1809908/24408340/6100749c-139b-11e7-8322-84a53cdfb71e.png)

ajrobbins:
I like var. 2

rajadain:
I like var 2 as well. It seems like clicking the black title bar will re-open the last opened tab?

[...]

Issue #1803, opened 2017-04-10
Uniform Height of Map Controls


rajadain:
Currently the Layer Selector buttons are less tall than map controls. This looks odd when the layer selector is collapsed:

![image](https://cloud.githubusercontent.com/assets/1430060/24883493/1b41df7c-1e13-11e7-9acf-a820334cf2ac.png)

Either make the layer selector tab bar taller, or the map controls shorter / smaller.

ajrobbins:
Now that we have a header bar that's there when the selector is collapsed, I find the height differential less noticeable:
![screen shot 2017-04-12 at 5 08 04 pm](https://cloud.githubusercontent.com/assets/12868342/24979567/af6121d2-1fa2-11e7-94d8-d5d6def8c2e9.png)

rajadain:
Agreed. Closing this.

PR #1805, opened 2017-04-11
Add collapsable header to layerpicker

arottersman:
## Overview

It's not obvious that re-clicking the active tab button on the layerpicker collapses the view. Add a header with an "expand/collapse" icon to collapse the view instead.

**Wireframes**

screen shot 2017-04-11 at 12 12 05 pm

Connects #1748 

### Demo

screen shot 2017-04-11 at 12 52 52 pm

screen shot 2017-04-11 at 12 52 48 pm

## Testing Instructions

 * Play around with the layerpicker:
        - You should be able to collapse/expand it by clicking the header, or the tabs
        - State should be maintained (if you collapse, and then expand the same tab should be open unless you clicked a different one)

### DIFF
--- a/src/mmw/js/src/core/layerPicker.js
+++ b/src/mmw/js/src/core/layerPicker.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var $ = require('jquery'),
+    Backbone = require('../../shim/backbone'),
     Marionette = require('../../shim/backbone.marionette'),
     _ = require('underscore'),
     utils = require('./utils'),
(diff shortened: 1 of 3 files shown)


rajadain:
I wonder if #1803 can be incorporated here easily. Not sure what the right height would be in this case though.

arottersman:
> I wonder if #1803 can be incorporated here easily. Not sure what the right height would be in this case though.

Part of the issue with this is the layerpicker icons have `font-size: 1rem` so they resize on the narrow viewport and the bar gets shorter. I think at the wide viewport they're the same height as the map controls.

Here's the layerpicker icon `font-size` at `20px`...seems closer:
screen shot 2017-04-12 at 8 33 12 am

caseycesari:
You can keep on repeating that process to make all of the tabs active:

image

arottersman:
:man_facepalming: there was some faulty `if-statement` logic. It's fixed now.

src/mmw/js/src/core/layerPicker.js
```diff
@@ -229,37 +230,48 @@ var LayerPickerNavView = Marionette.ItemView.extend({
...
         'click @ui.tabIcons': 'onTabIconClicked',
     },
 
-
     templateHelpers: function() {
+        var doesLayerGroupHaveActiveLayer = function(layerGroup) {
+            var layers = layerGroup.get('layers');
+            if (layers) {
+                return layers.findWhere({ active : true });
+            }
```
caseycesari: Consider returning false at the end of the function just so the result is consistently a boolean.

rajadain:
> Part of the issue with this is the layerpicker icons have font-size: 1rem so they resize on the narrow viewport and the bar gets shorter.

Let's do this separately then. Sorry to conflate these.

arottersman:
Thanks @caseycesari! Added a `return false` and squashed fixups. Going to merge.

Issue #1773, opened 2017-03-24
Map components are visible under sidebar when app is loading


caseycesari:
The map controls and attribution appear on top of the sidebar while the app is loading. I believe this is because the sidebar doesn't have a background color set, and the map is initially sized to the full width of the page.

![mmw4](https://cloud.githubusercontent.com/assets/1042475/24310539/b239b32a-10a6-11e7-8085-31a6c23404ab.gif)

rajadain:
This is likely because of the fact that we use absolute positioning for the map, and pull it back with a value of [`right: 485px`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/sass/base/_map.scss#L43). The `.sidebar` component has full width and `0` height in practice, and the contents are just floating in empty space. We see the same behavior on the Bees application: https://pollinator-modeling-app.azavea.com/

I think a solution to this will be to rearchitect the front-end to use a hierarchy of nested divs with relative positioning, rather than a flat structure with absolute positioning. But such a rearchitecture is probably more than 1 point, and may need some **#design** support.

PR #1806, opened 2017-04-12
Keep map within post-render bounds on load

rajadain:
## Overview

On load, the map would expand to the full page width, and transition to a reduced width after the application finishes loading. This led to a jarring transition, especially when the application loaded slowly.

By adding the classes that would be applied post-load to the `.map-container` before it loads, we ensure it stays within the bounds of actual rendering.

The project view uses a smaller map than the home view, but since it is a transition of the top of the map it isn't as jarring. No buttons move around, and there is no overlapping content.

Connects #1773 [:lipstick:](https://gitmoji.carloscuesta.me/)

### Demo

Loading the home page:

Loading a project page:

## Testing Instructions

 * Check out this branch and rebundle the scripts `./scripts/bundle.sh --debug`
 * Reload the app in various configurations and ensure that you don't see the flash of transition as described in the original issue.

### DIFF
--- a/src/mmw/apps/core/templates/base.html
+++ b/src/mmw/apps/core/templates/base.html
@@ -5,7 +5,8 @@
     {% block header %}
     {% endblock header %}
 
-    <div class="map-container">
+    <div class="map-container map-container-top-sidebar map-container-bottom-sidebar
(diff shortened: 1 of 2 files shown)


caseycesari:
Taking a look.

Issue #1761, opened 2017-03-23
Add more validation to shapefile/geojson uploads


caseycesari:
- Check that coordinates are geographic
- Check that the shape is in the U.S.
- Implement a better file type check (than looking at the extension). 

It may be a good idea to create a Backbone model for uploaded data and move the validation there.

rajadain:
Where are the shapefiles going to be stored?

caseycesari:
Currently, and we should strive to maintain this, we are using the DataTransfer and FileReader browser APIs. This allows us to keep the file local to the user and not have to upload it anywhere.

arottersman:
Should the guard against non-USA shapes apply to all shapes? You can currently draw AoIs anywhere you like

arottersman:
> For shapefiles, check that the .shp and .prj are present, and that the file is otherwise valid (not sure how to do this).

What qualifies as "file is otherwise valid"? It seems like it should be enough to know if we can (1) open and parse the file (2) validate the parsed shape by our usual measures (self intersection, max-aoi, polygons, etc). We know if either of these steps fail, and I think are already  displaying errors accordingly. Is there some additional check that needs to occur?

caseycesari:
> Should the guard against non-USA shapes apply to all shapes? You can currently draw AoIs anywhere you like

Good point. I was just thinking that we know shapes outside the US will fail, so we may want to not bother trying to analyze them, but I'll defer to @ajrobbins.

caseycesari:
> What qualifies as "file is otherwise valid"? It seems like it should be enough to know if we can (1) open and parse the file (2) validate the parsed shape by our usual measures (self intersection, max-aoi, polygons, etc). We know if either of these steps fail, and I think are already  displaying errors accordingly. Is there some additional check that needs to occur?

Not necessarily, I just wanted to make sure we were catching and throwing error messages in all of the likely scenarios (shapefile missing components, zip is unreadable, etc). This may be something we have to iterate on as errors come up.

ajrobbins:
>Should the guard against non-USA shapes apply to all shapes? You can currently draw AoIs anywhere you like
If we make the change here to apply to all shapes, will we use the same error modal to say why we won't analyze (with shapefile upload or regular analyze outside the US)? if the fix needs to be in two places, we should make a separate card for the general guard.

caseycesari:
That's true. Might be worthwhile to work it in to the general workflow.

PR #1816, opened 2017-04-13
Move AOI Upload Into Sidebar and Accept Zipfiles

arottersman:
## Overview

- Move shapefile upload to sidebar
- Wire up "Select file" button
- Instead of accepting `.shp`, take `.zip`
- To unzip the compressed shapefile, I added [`JSZip`](https://github.com/Stuk/jszip)
    - Unzipping files is non-trivial, so a library is probably best
    - The other option in the js package space is [`zip.js`](https://github.com/gildas-lormeau/zip.js) but it isn't available via npm, nor is it versioned

Connects #1760 

### Demo

uploadfilemmw

### Notes

The card mentions a loading spinner, but it's so instantaneous I'm not sure it's necessary to add it with this PR. 

## Testing Instructions

 * If it works`./scripts/npm.sh install`, or `ssh worker` then `cd /vagrant/src/mmw && npm install`
 * `./scripts/bundle.sh --vendor && ./scripts/bundle.sh`
 * Drag and drop some geojson and zipped shp files, and also add them via the "Select file" button
 * Zipfiles without a `.shp` should display an error modal


src/mmw/js/src/draw/templates/aoiUpload.html
```diff
@@ -1 +1,21 @@
...
+        <span class="draw-button-title pull-left"
+              id="dropbox">
+            Drag and drop a file
+            <ul class="draw-pane-description-list">
+                <li>Must be a shapefile or geojson</li>
+                <li>Less than 200 MB</li>
+                <li>Only the first feature is used.</li>
+            </ul>
+        <button class="btn btn-success pull-left"
+                onclick="document.getElementById('draw-tool-file-upload-input').click();">
```
caseycesari: Curious about this inline event binding. Is it not possible to do this in a Backbone view?

arottersman: It is — I'll push a fixup commit

arottersman: Having it in the template was somewhat nice because it is explicit about the button being a dummy for the hidden input. Writing the `onclick` code as a string is not ideal though

caseycesari: Yeah, I think adding a note in the commit message about that might be helpful.

caseycesari:
Re: the loading spinner, I agree. In my tests, the files have been processed almost simultaneously since we don't have to do any actual uploading. Shapefiles are limited to 2gb by design, and since we are only using the first record, I think we should be able to handle any shapefile fine. GeoJSON has no file size limit, but if you upload GeoJSON with a complex shape, the browser hangs as the geometry is parsed. We may want to limit the size of GeoJSON (as a correlate, since we don't know if the geometry is complex until we process it). The browser receives the file metadata before it reads in the file. In any case, I think this could be handled in https://github.com/WikiWatershed/model-my-watershed/issues/1761 or another card.

caseycesari:
Two more things that we talked about, that are for other cards, but that I wanted to make a note of so we didn't forget:

- We may want to add a mask when processing the file, just like geojson.io, to give an indication that the UI is locked while the file is being processed.
- GeoJSON.io seems to parse geojson files way faster than us. We should take a look at what they are doing.

src/mmw/js/src/draw/templates/aoiUpload.html
```diff
@@ -1 +1,22 @@
-<span id="dropbox">Drop a .shp file here.</span>
+<ul class="draw-pane-list">
+  <li class="draw-pane-list-item" role="presentation">
+    <div>
+        <span class="draw-button-title pull-left"
+              id="dropbox">
+            Drag and drop a file
+            <ul class="draw-pane-description-list">
+                <li>Must be a shapefile (.zip) or geojson</li>
+                <li>Less than 200 MB</li>
```
caseycesari: I suggested this, but I think we should remove it until we actually implement a limit.

arottersman:
Thanks for taking a look! 

- Removed the message about the 200mb limit, but not the check (it's usefulness can be evaluated in #1761 )
- Squashed the fixup

Issue #1782, opened 2017-04-03
Zoom control vertically stacked on Firefox


kellyi:
<img width="436" alt="screen shot 2017-04-03 at 11 53 41 am" src="https://cloud.githubusercontent.com/assets/4165523/24618182/45ff251c-1864-11e7-96e2-becf28da0bfa.png">

rajadain:
Seems to be okay on latest develop using latest Firefox:

![image](https://cloud.githubusercontent.com/assets/1430060/24815813/904640e0-1ba4-11e7-97a3-4b7157d93952.png)

Do I have to do something to get it broken?

kellyi:
Probably worth having a third person check this as staging still looks like this for Firefox on my local:

<img width="640" alt="screen shot 2017-04-07 at 4 02 37 pm" src="https://cloud.githubusercontent.com/assets/4165523/24817745/c50b035e-1bab-11e7-96f1-315e857a7cdc.png">

kellyi:
(Running the same FF version, btw.)

rajadain:
I just checked Staging as in your screenshot, and that looks okay too:

![image](https://cloud.githubusercontent.com/assets/1430060/24817891/5637cdbc-1bac-11e7-9d1e-f79171012728.png)

Can another person verify this?

arottersman:
Looks fine on mine, and I'm assuming @caseycesari  (we both tested all major browsers for this change's PR https://github.com/WikiWatershed/model-my-watershed/pull/1769#pullrequestreview-28982451 )
![Uploading Screen Shot 2017-04-07 at 4.14.04 PM.png…]()

caseycesari:
Yeah, still fine for me in Firefox:

![image](https://cloud.githubusercontent.com/assets/1042475/24908553/69ea3680-1e8e-11e7-84f2-349746248bf2.png)

kellyi:
This looks to have been happening because I had `touch_events_enabled` set to true in Firefox's `about:config`:

<img width="769" alt="screen shot 2017-04-11 at 11 06 05 am" src="https://cloud.githubusercontent.com/assets/4165523/24916016/e0e0313c-1ea6-11e7-912d-21c7f9ee87f2.png">

It turns out this triggers Leaflet to apply a `leaflet-touch` class to the entire leaflet-container which then reduces the width of the zoom control box by 2 pixels, narrowing it enough that it dips beneath the 80px width threshold required for the buttons to display horizontally & not stack.

[...]

PR #1819, opened 2017-04-13
Account for fat borders on touch devices

rajadain:
## Overview

Touch devices add a 2px gray border to map controls. Since we use `box-sizing: border-box` for the entire app, these borders eat in to the controls' width rather than expanding out, causing layout and display issues.

By updating the controls to be 4px wider and taller on touch devices, we ensure that the controls are rendered correctly.

:bug: :lipstick:

Connects #1782

### Demo

image

## Testing Instructions

 * Check out this branch and bundle your scripts `./scripts/bundle.sh --debug`
 * Open `about:config` in Firefox and set `dom.w3c_touch_events.enabled` to `1`
 * Open the site. The map controls should render properly.
 * Test on a regular Firefox / Chrome. There should not be the extra gray border on non-touch devices, and the buttons should look the same as before.
 * Test on an iPad, and ensure that the buttons look and behave correctly

### DIFF
--- a/src/mmw/sass/base/_map.scss
+++ b/src/mmw/sass/base/_map.scss
@@ -134,6 +134,9 @@
            font-size: 24px;
            line-height: 38px;
        }
+       .leaflet-touch & {
+           width: 84px;
+       }
     }
 }
 
@@ -156,6 +159,10 @@
         &:hover {
             background-color: #ccc;
         }
+    }
+    .leaflet-touch & {
+        height: 44px;
+        width: 44px;
     }
 }
 


src/mmw/sass/base/_map.scss
```diff
@@ -157,6 +160,10 @@
             background-color: #ccc;
         }
     }
+    .leaflet-touch & {
```
rajadain: Thanks to @jfrankl for teaching me this neat trick! This allows accounting for `.leaflet-touch` coming from the outside scope, since that class is applied to the `.map-container` tag.

arottersman: Neat!

jfrankl: Hooray!

Issue #1793, opened 2017-04-06
Improve tooltip interactivity


caseycesari:
https://github.com/WikiWatershed/model-my-watershed/pull/1791 introduced tooltips that open on click instead of hover, and that are closed by clicking anywhere. This generally works well, but it may not be clear that the tooltip can be closed by clicking anywhere. We could add a close button for this (there appear to be [many solutions](https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=add+close+button+to+bootstrap+popover&*), though none are that clean).

Also, when clicking anywhere to close the tooltip, you may click on another interactive element of the app. It's not clear if this is desirable. To prevent this, we could add a transparent mask over the app when a tooltip is activated that captures clicks and closes the tooltip.

arottersman:
Tooltip doesn't work when is activated, but it probably should. This is especially noticeable when waiting for RWD to finish.

kellyi:
Cataloguing some desired behavior for this:

- [ ] tooltip should close on clicking anywhere outside the tooltip, but those clicks shouldn't propagate through to other elements
- [ ] tooltip should close on clicking the ? icon a second time when it's already open
- [ ] tooltip should have an X to close button somewhere
- [ ] tooltips should be open-able when the draw tools are active

Should the tooltip close on clicking the tooltip text? Or only on the X to close button? If it's the former, we could just add a placebo X to close and have the cursor turn into a pointer on hovering over it, since clicking anywhere on the tooltip would close it already.

kellyi:
> Tooltip doesn't work when a tool is activated, but it probably should. This is especially noticeable when waiting for RWD to finish.

Investigated this and got the circumstances in which this happens narrowed down somewhat. I started replacing the `a` tags we used to trigger these popovers with `button` elements instead. Currently it looks like what happens is that:

- click on a draw tool to activate it disables that tool's button and the other draw tool buttons in that slideout
- the tooltips in that slideout are also non-functional at that moment
- clicking on a draw different draw tool to open its slideout renders it with the draw tool buttons disabled but the popover buttons working
- clicking on the active draw tool's button to open its slideout again renders it as before except the popover buttons now work

I assume there's something happening to the whole slideout on activating a draw tool which causes everything to be non-functional except the close button. This probably also doesn't happen inherently on first rendering the thing, and the slideouts are all re-created anew each time they're opened.

kellyi:
[...]

PR #1825, opened 2017-04-17
Add tooltip close button and enable opening tooltips when a draw tool is active

kellyi:
## Overview

This PR makes two adjustments to the draw window tooltips: it adds a close button in the top right corner, and also enables opening the tooltips when a draw tool is active.

Since the clicking anywhere still closes the popovers, the x to close button works primarily as a highlightable thing to click to close it. It's handled here by adding a string template to the popover including the X to close button which then gets rendered as the markup for the popover div.

Opening the tooltips when a draw tool's active took a big to figure out, but it looks like the tooltips were activated when `onShow` was called for free draw or RWD, but then got deactivated if the `render` method was called again (which happens when the component gets new props from the model). To fix this, I abstracted out an `activatePopovers` function for both free draw and RWD, which is called in `onShow` and then again in `onRender`.

Connects #1793 

### Demo

mmw-tooltip-interactivity

### Notes

Still investigating this part to see if it's common/desirable behavior:

> when clicking anywhere to close the tooltip, you may click on another interactive element of the app. It's not clear if this is desirable.

## Testing Instructions

 * get this branch, then rebundle and view it in the browser
 * click on a draw tool, then open a popover and verify that you see the X button in the top right corner, that the cursor turns into a pointer above the button, and that clicking it closes the popover
 * select free draw, sq km, and the rwd options in series and verify that when each of them is selected the popovers still work as usual in that draw pane slideout

### DIFF
--- a/src/mmw/js/src/draw/templates/delineationOptions.html
+++ b/src/mmw/js/src/draw/templates/delineationOptions.html
@@ -20,7 +20,10 @@
          role="button" data-content="Snaps to the nearest downhill point on the Delaware River Basin
          high resolution stream network and calculates the watershed upstream of this point using a
          1/3 arc second (10m) resolution digital elevation model. The stream network was
-         delineated using <a href='http://hydrology.usu.edu/taudem/taudem5/index.html' target='_blank'>TauDEM</a>.">
+         delineated using <a href='http://hydrology.usu.edu/taudem/taudem5/index.html' target='_blank'>TauDEM</a>."
+         data-template="<div class='popover'><div class='pull-right'
+             id='popover-close-button'><i class='fa fa-times' /></div><div
+             class='popover-content'></div><div class='arrow'></div></div>">
          <i class="fa fa-question-circle"></i>
       </a>
     </li>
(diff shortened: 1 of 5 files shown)


src/mmw/js/src/draw/templates/delineationOptions.html
```diff
@@ -16,33 +16,33 @@
               {{ 'disabled' if toolsDisabled }}>
```
caseycesari: Consider adding a comment to this commit to note why this change was made.

kellyi:
> They do not work in Firefox or Safari

Oof, I think this is because I changed them from `a` elements to `buttons`. Will fix.

kellyi:
Pushed a new set of commits without the `a` -> `button` change, which was what broke things in Safari & Firefox.

kellyi:
> when clicking anywhere to close the tooltip, you may click on another interactive element of the app. It's not clear if this is desirable.

Investigating this a bit and if it makes a difference, the cursor and the interface generally do change when the cursor's placed over an interactive element even when the tooltip is open, which I suppose indicates that the element is still interactive even if the tooltip's open.

The least apparent version of this is the map, though, which also takes up the most space. The cursor changes in a way that indicates it's still interactive, but there's no "element highlighting" like happens with the zoom controls or the layer selector or a link when a tooltip's open.

A single click doesn't have any effect, but a double click when the tooltip's open will close the tooltip and zoom the map, and a click-and-hold will dismiss the tooltip and make the map draggable.

kellyi:
Thanks! I'm going to merge this and I think I'll make another card specifically for deciding whether or not we should disable user interactions when a popover is open.

Issue #1798, opened 2017-04-06
Don't show "turn on related layer" if layer is unavailable


ajrobbins:
DRB-only overlays should not be prompted to be turned on if the AoI is not in the DRB

PR #1829, opened 2017-04-18
Don't show disabled layers in Analyze "Related Layers"

arottersman:
## Overview

Certain layers, eg DRB catchments, are only relevant at certain zoom levels or certain places of the country. We disable such layers in the layerpicker, but we were still showing them in the analyze layer toggle views. 

- disable layers via the model instead of the layerpicker
- don't show disabled related layers

Connects #1798 

### Demo
screen shot 2017-04-18 at 4 17 48 pm

## Testing Instructions

* Pan and zoom the map around to confirm layers still disable as you'd expect
    - check the coverage grid DRB catchments, philly school district, etc
 * Draw an AoI in philly
      - the water quality tab should show 3 related layers
 * Draw an AoI outside the DRB
      - there should be no related layers listed

### DIFF
--- a/src/mmw/js/src/analyze/templates/analyzeLayerToggle.html
+++ b/src/mmw/js/src/analyze/templates/analyzeLayerToggle.html
@@ -1,18 +1,20 @@
-{% if message %}
-    <div class="analyze-layermessage">
-        {{ message }}
-    </div>
-{% else %}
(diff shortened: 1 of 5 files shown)


arottersman:
Thanks — I agree, and added it to https://github.com/WikiWatershed/model-my-watershed/issues/1834

Issue #1826, opened 2017-04-18
Improve boundary layer search performance


kdeloach:
The boundary layer search endpoint is slow on staging. Searching for "schuylkill" takes about 15 seconds.

https://staging.app.wikiwatershed.org/api/modeling/boundary-layers-search/?text=schuylkill

Update query to use PostgreSQL full text search functions instead wildcard match. Consider adding indexes. Verify performance improvements on staging.

Ref: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/views.py#L538

kdeloach:
Took a cursory pass at implementing a few different alternatives to `ILIKE` with no luck.

### ts_vector, ts_query

The builtin text search functions work well, are very fast, and support advanced query syntax. Unfortunately, only exact match search is supported.

### fuzzystrmatch

The range of values produced from `difference` is not granular enough to determine the best match.

```
select difference(soundex(name), soundex('schuy')), count(*)
from boundary_huc10
group by difference(soundex(name), soundex('schuy'));

 difference | count 
------------+-------
            |   214
          0 |    23
          3 | 14148
          4 |  1554
```

### pg_trgm

Boundaries which contain the search term rank lower or equal to boundaries which do not contain the search term in many cases.

```
select name, similarity(name, 'canal')
from boundary_huc10
order by similarity(name, 'canal');

Old Channel Poso Creek-Alpaugh Irrigation District Canal              |  0.109091
Little Quillcene River-Frontal Hood Canal                             |  0.146341
Canadaway Creek-Frontal Lake Erie                                     |  0.147059
```

```
select name, similarity(name, 'schuy')
from boundary_huc10
order by similarity(name, 'schuy');

School Creek                                                          |    0.1875
Middle Schuylkill River                                               |       0.2
Little Schuylkill River                                               |       0.2
Upper Schuylkill River                                                |  0.217391
Lower Schuylkill River                                                |  0.217391
Schulyler Wash                                                        |  0.235294
```

kdeloach:
I improved the query in #1832 so the search performs reasonably well now. Searching for "schuylkill" takes about 3-4 seconds instead of 15. Closing.

PR #1832, opened 2017-04-19
Increase number of results limit on boundary layer search

kdeloach:
## Overview

Instead of displaying 5 total results for all boundary layers, we now
show up to 3 results for each boundary layer.

Also, priority was renamed to rank. Since we now show results for each
boundary layer regardless of priority/rank, the only thing this number
does is determine the sort order for results on the frontend.

The goal of this is to show more results and hopefully improve performance
before we spend any more time researching #1826.

### Demo

See how there are 3 results for each boundary layer. If I had HUC12s on my workstation you might see an additional 3 results.

image

### DIFF
--- a/src/mmw/apps/modeling/views.py
+++ b/src/mmw/apps/modeling/views.py
@@ -541,14 +541,11 @@ def _get_boundary_search_query(search_term):
     all searchable boundary layers.
     """
     select_fmt = """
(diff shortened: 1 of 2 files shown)


kdeloach:
Cool, hopefully it performs better on staging too.

Issue #1792, opened 2017-04-06
Sidebar doesn't scroll if content is too long


caseycesari:
To reproduce, copy and paste the `p` element that is at the top of the sidebar until the contents of the sidebar are moved off-screen. 

<img width="481" alt="screen shot 2017-04-06 at 12 43 05 pm" src="https://cloud.githubusercontent.com/assets/1042475/24765660/9b9913ca-1ac6-11e7-881a-aa00eafafe7f.png">

rajadain:
Similar work was done here: https://github.com/project-icp/bee-pollinator-app/pull/151

PR #1836, opened 2017-04-20
Make sidebar scrollable

rajadain:
## Overview

Makes the sidebar scrollable so that when there is a lot of content, or the screen is really short, the app remains accessible. This change applies to the sidebar in splash and draw views.

:bug: :lipstick:

Connects #1792 

## Testing Instructions

 * Check out this branch
 * Download this patch file: [long-sidebar-text.patch.txt](https://github.com/WikiWatershed/model-my-watershed/files/944709/long-sidebar-text.patch.txt)
 * Apply the patch to add long text to your local

       $ git apply long-sidebar-text.patch.txt

 * Bundle the new stuff `./scripts/bundle.sh --debug`
 * Open the app in your local and ensure that it is workable even though main action buttons are pushed below the fold

### DIFF
--- a/src/mmw/sass/pages/_draw.scss
+++ b/src/mmw/sass/pages/_draw.scss
@@ -6,7 +6,8 @@
     right: 0;
     transition: all 200ms;
     top: 44px;
-    overflow: visible;
+    bottom: 0px;
+    overflow-y: auto;
 }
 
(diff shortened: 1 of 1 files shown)


kellyi:
Setting MMW to take a look at this.

Issue #1840, opened 2017-04-20
Initial UI for data catalog search


mmcfarland:
A new sidebar stage with a single search text box as a placeholder for data catalog search.

* New container and template
* Feature gated to only BigCZ build
* Route (`/search`) for post AoI selection step, instead of analyze

Note, this doesn't need to actually interface with the in-development API endpoint.  That will be handled subsequently.

mmcfarland:
This got more detail on today's call.  @ajrobbins can you follow up with details?

ajrobbins:
Sure:
- There will be an MMW-like analyze stage after AoI selection, so the /analyze route should be preserved after /draw
- A /search route will be available once the user has reached the Analyze step, which would enable the search text box as the original issue stated
@arottersman I realize this conflicts with your implementation. We were operating under the wrong impression regarding the removal of Analyze.

ajrobbins:
I think for now we should change the "Model" button that appears on the Analyze screen to "Search for data", make it not a dropdown, and have that button lead to the new text search screen. @jfrankl @mmcfarland Does that match your understanding from today's call?

jfrankl:
I think that is fine as a placeholder. We're still early in the design stage for MMW, so this might need to change quite a bit in the upcoming sprints.

PR #1842, opened 2017-04-24
BiGCZ: Add Initial `/search` Sidebar View

arottersman:
## Overview

Add the initial data catalog search ui:
- When in BiG CZ mode, ~route to `/search` after draw~ replace the model selection dropdown in analyze with navigation to `/search`
- Add dummy search bar to `/search` sidebar view

*Possibly outdated wireframe I used for inspiration*
screen shot 2017-04-24 at 5 04 09 pm

Connects #1840 

### Demo

screen shot 2017-04-25 at 2 55 40 pm

screen shot 2017-04-24 at 5 06 18 pm

### Notes
 ~This assumes the setting `analyze_enabled=False` is equivalent to "show data catalog"
@kdeloach How did you envision this? Should there be an additional flag `data_catalog_enabled`?~

Replaced `analyze_enabled` with `data_catalog_enabled` as now both apps will have some sort of analyze step

## Testing Instructions
 * `./scripts/mode.sh mmw`: check that analyze and model still work and you can't access the data catalog
 * `./scripts/mode.sh bigcz`: check that there's a button in analyze to "Search data sources", and you are routed to `/search`. You should see filler UI
 * Going direct to `/search` directs you back to the home page

### DIFF
--- a/src/mmw/apps/home/urls.py
+++ b/src/mmw/apps/home/urls.py
@@ -22,6 +22,7 @@ urlpatterns = patterns(
     url(r'^project/compare/$', project, name='project'),
     url(r'^project/(?P<proj_id>[0-9]+)/compare/$', project, name='project'),
     url(r'^analyze$', home_page, name='analyze'),
+    url(r'^search$', home_page, name='search'),
     url(r'^error', home_page, name='error'),
     url(r'^sign-up', home_page, name='sign_up'),
 )
--- a/src/mmw/apps/home/views.py
+++ b/src/mmw/apps/home/views.py
@@ -132,7 +132,7 @@ def get_client_settings(request):
             'vizer_names': settings.VIZER_NAMES,
             'model_packages': get_model_packages(),
             'mapshed_max_area': settings.GWLFE_CONFIG['MaxAoIArea'],
-            'analyze_enabled': settings.ANALYZE_ENABLED,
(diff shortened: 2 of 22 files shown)


arottersman:
Hold the phone...[this is going to have change ](https://github.com/WikiWatershed/model-my-watershed/issues/1840#issuecomment-296827556)

arottersman:
@kdeloach, this should be ready to review.

kdeloach:
I noticed one small issue. If you geolocate while the Data Sources panel is open, this happens:

image

Should we hide the geolocate box while the Data Sources panel is active?

arottersman:
Weird that that happens. I added the geocoder to match the wireframes, but I suppose if the data source search is focused on the AoI, the geocoder doesn't have much use. Going to add a fixup to remove it, unless I'm missing something, @jfrankl?

arottersman:
With @ajrobbins — we think it'd be weird if after drawing an AoI the geocoder disappeared for Analyze, and then the reappeared for Data Source Search. In MMW we currently get rid of the geocoder from Analyze on, and we think we should mimic that behavior in BiGCZ

kdeloach:
Agreed.

arottersman:
Fixup removes geocoder from the data source search stage

arottersman:
Thanks!

Issue #1484, opened 2016-09-21
Analyze & Model data input caching strategy


mmcfarland:
Develop an ADR which defines a strategy for preserving Analysis output and Modeling input for Well Known Areas of Interest (WKAoI) which will have repeated analysis and models run against them.  

Some observations and questions:
- For all WKAoIs, it's unlikely that a majority of all features among all layers will have analysis or models run on them.  I suspect it is not worth the effort to pre-compute all output for all features, but rather capture that data on the fly when an uncached area is processed.
- The data which makes up the Analysis output is anticipated to keep changing, with new sources added.  The data that is gathered for model runs is more static in nature, but also can be refreshed periodically.  An effective method to clear the cache when data changes is crucial.  We currently cache model results _for a single project_ and a hash is generated on the input to the model to determine of the results are still valid.  Is there a similar piece of data that can be used as a key, or can something be synthesized to inform the application to use cached results or recalculate them?
- For this cache to be effective, the WKAoI needs to be identified by more than its shape.  Presently, the app treats a user generated shape the same way it treams a WKAoI - both just post the Geometry.  We'll need to switch this to a request based on Layer Id and Feature Id, with the geometry being sourced from the database.  This will also make the request faster as some AoIs are quite large (see #303).
- Analysis output is often similar to model input - can a parent set of data be generated once that would satisfy both? 
- Is it important to cache the intermediate results of the model data input, or is the end result sufficient?  Probably related to point above.
- Should discreet analysis output be cached/versioned, or should a change in one dataset invalidate the whole "analysis" package? 
- HTTP caching from the client is unlikely to be sufficient since the analysis results are polled based on a `JobId` and modeling input is not returned to the client, only modeling output.  MapShed does return GMS style data, but we may want to cache at a level of detail up from that, since GMS will contain Conservation Practices as well (which should not be cached) and I suspect serializing to GMS is not very expensive.

The goal here is to:
- Increase the speed performance of analysis and modeling requests
- Reduce the overall load on the geoprocessing pipeline for repeatable requests so that ad-hoc AoI requests have more resources
- Potentially be able to expand the size of a WKAoI that the application can handle.  This is at odds with the on-the-fly suggestion earlier (since resource exhaustion would still occur), but with this caching mechanism in place we could do targeted out-of-band result generation.  For example, we could target all HUC-08's within the DRB and generate Analysis and Model data input for them with which we populate the cache using temporary and more robustly provisioned SJS machines.

[...]

PR #1843, opened 2017-04-25
ADR 002 for Caching Geoprocessing Results of Predefined Shapes

rajadain:
## Overview

This ADR provides an overview of a strategy to cache geoprocessing results for predefined shapes, or Well Known Areas of Interest (WKAoIs).

Connects #1484

### DIFF
--- a/doc/arch/adr-002-geoprocessing-caching.md
+++ b/doc/arch/adr-002-geoprocessing-caching.md
@@ -1,0 +1,112 @@
+# 002 - Geoprocessing Caching
+
+## Context
+
+Geoprocessing calls for large shapes can take a long time to complete. While users can draw custom shapes, they can also pick from a list of predefined shapes in the system. Geoprocessing the same shape over and over when multiple users select it is wasteful and unnecessary. By caching the geoprocessing results of predefined shapes, or Well Known Areas of Interest (WKAoIs), we can improve user experience and application performance.
+
+Outside of geoprocessing, there are also [some](https://github.com/WikiWatershed/model-my-watershed/blob/dea12ff9a6fb234e30978e97ed1d0f6266a406c9/src/mmw/apps/modeling/calcs.py#L28) [database](https://github.com/WikiWatershed/model-my-watershed/blob/dea12ff9a6fb234e30978e97ed1d0f6266a406c9/src/mmw/apps/modeling/calcs.py#L51) [calls](https://github.com/WikiWatershed/model-my-watershed/blob/dea12ff9a6fb234e30978e97ed1d0f6266a406c9/src/mmw/apps/modeling/calcs.py#L97), but these are much faster in comparison and are not likely to see significant improvement by caching, especially since the cache could be a table in the database itself.
+
+In this ADR we consider the following questions:
+
+ * What will be cached?
+ * How will it be cached?
+ * Where will it be cached?
+ * When will it be cached?
+ * How will the cache be invalidated?
+
+We also sketch out an implementation plan, and consider consequences and side-effects.
+
+## Decisions
+
+### What will be cached?
+
+The output of the geoprocessing service for WKAoIs will be cached. This is almost always a JSON blob of key-value pairs, where the key is a combination of overlaid cell values in a set of rasters, and the value is a count of such cells. The inputs are a GeoJSON shape and a [set of related arguments](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/base.py#L412) that specify rasters, operation type, CRS, etc.
+
+While we could cache the entire output of a MapShed or TR-55 run, an update to some of the constituent raster or vector data would force recalculation of the whole. By caching only the time-consuming geoprocessing results, we ensure that any updates to constituent rasters would invalidate only that specific cached result, leaving the others still current. And since vector data results would never be cached, they will always be current upon update as well.
+
+In case of MapShed, the modifications do not change any of the geoprocessing queries, thus those requests can be cached easily. For TR-55, the modifications _do_ change the geoprocessing queries. For these cases, we will cache only the Current Conditions (modification-less) run, not other Scenarios, since storing arbitrary shapes can balloon the size of the cache very quickly. This decision may be revisited in the future.
+
+In case we do not foresee updating the rasters very often, it may be beneficial to consider caching the entire JSON response of the API.
+
+### How will it be cached?
+
+The current stack already has [Django Caching](https://github.com/WikiWatershed/model-my-watershed/blob/a49cc115d18d2c67cbd0721cf62faa08e98d13fc/src/mmw/mmw/settings/base.py#L76-L90) setup. This allows us to cache with a single line of code:
+
+```python
+from django.core.cache import cache
+
+cache.set(key, value, None)
+```
+
+where the `key` is a unique identified consisting of WKAoI id and the geoprocessing operation, `value` is the result of the geoprocessing operation, and `None` is the timeout value which ensures the values don't ever expire.
+
+Retrieval is as simple as:
+
+```python
+value = cache.get(key)
+if not value:
+    # Calculate and cache the value
+
+return value
+```
+
+The `key` should be prefixed with `geop` to namespace it from other cache entries in the app, composed of the WKAoI id, which consists of a table name and an integer id, and the [geoprocessing operation name](https://github.com/WikiWatershed/model-my-watershed/blob/a49cc115d18d2c67cbd0721cf62faa08e98d13fc/src/mmw/mmw/settings/base.py#L387). For example, `geop__boundary_huc08__1__nlcd_soils`. In practice, the `key` may actually be something like `:1:geop__boundary_huc08__1__nlcd_soils`, with a configurable app-wide prefix (which defaults to `''`) and a version number are prefixed to our key. However, as long as we always access the cache via `django.core.cache.cache`, we shouldn't need to worry about that.
+
+### Where will it be cached?
+
+Django Caching Framework can be configured to use a number of cache backends, including Redis and Database. It is currently setup to use Redis / [ElastiCache](https://aws.amazon.com/elasticache/redis/) in production.
+
+The advantages of using Redis are:
+
+ * It is already configured
+ * It allows us to take advantage of the Django Caching Framework which is configured to use Redis
+ * Redis is really fast, and a good candidate for storing key/value pairs like we intend to
+ * It is designed to be a cache, and thus comes with mechanisms for timeout, LRU, and cache misses out of the box
+
+The disadvantages of using Redis are:
+
+ * In the case of system failure, the cached values will be lost, and will need to be cached again
+ * If it purges least recently used values, it might not be a good candidate for storing hard-to-process large WKAoIs that are rarely used
+
+### When will it be cached?
+
+When a user selects a WKAoI, and a request is made to `/analyze` or `/modeling`, if the WKAoI results haven't already been cached, we will run and cache them. Over time, we will build up the cache, so that when new users request the same WKAoIs, they will get the cached results. These will be cached in the Celery Tasks where calculation would otherwise happen.
+
+For a select few WKAoIs which are too large to process in the production infrastructure, such as HUC-8s which time-out during MapShed gathering phase, we can run their geoprocessing steps on more powerful infrastructure with longer timeouts in a batch process, cache the results, and then decommission it. This will make them available to regular users of the production app without needing the extra power to render them.
+
+For this purpose, we'll need a pair of new Django management commands. The first should run the geoprocessing steps for given WKAoIs and save the results to a file. This will be run in the super environment. The second should take a file of pre-processed results for given shapes and operations, and add them to the cache. This will be run in the production environment.
+
+### How will the cache be invalidated?
+
+Since every cache entry is tagged with its type, if a certain raster is updated, we can remove all related cache entries. For example, if there are updates to `us-percent-slope-30m-epsg5070` raster, which is used in the [`nlcd_slope`](https://github.com/WikiWatershed/model-my-watershed/blob/a49cc115d18d2c67cbd0721cf62faa08e98d13fc/src/mmw/mmw/settings/base.py#L449) and [`slope`](https://github.com/WikiWatershed/model-my-watershed/blob/a49cc115d18d2c67cbd0721cf62faa08e98d13fc/src/mmw/mmw/settings/base.py#L462) requests, we could simply run
+
+```python
+cache.delete_pattern('geop__*__nlcd_slope')
+cache.delete_pattern('geop__*__slope')
+```
+
+To refresh a specific shape we could do
+
+```python
+cache.delete_pattern('geop__boundary_huc08__1__*')
+```
+
+These could be management commands as well.
+
+## Implementation Plan
+
+1. See if TR-55 geoprocessing, currently done via [`SummaryJob`](https://github.com/WikiWatershed/mmw-geoprocessing/blob/develop/summary/src/main/scala/SummaryJob.scala), can be done via [`MapshedJob`](https://github.com/WikiWatershed/mmw-geoprocessing/blob/develop/summary/src/main/scala/MapshedJob.scala), for consistency. And if so, rewrite the geoprocessing bits to use `MapshedJob` instead.
+2. Split [`nlcdSoilCensus`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/base.py#L388) into three requests: `nlcd`, `soil`, and `nlcd_soil`. The first two can be used for `/analyze` and will be much faster, while the second can be used for `/modeling/tr-55`.
+3. Update the geoprocessing submodule to take GeoJSON shape or WKAoI id, and geoprocessing type, and return the output. Update the Celery Tasks or Django Views which use the geoprocessing submodule to use the new interface.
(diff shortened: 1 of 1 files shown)


doc/arch/adr-002-geoprocessing-caching.md
```diff
@@ -0,0 +1,103 @@
...
+ * How will items be identified, cached, and retrieved?
+ * How will the cache be invalidated?
+
+We also sketch out an implementation plan, and consider consequences and side-effects.
+
+## Decisions
+
+### What will be cached?
+
+The output of the geoprocessing service for WKAoIs will be cached. This is almost always a JSON blob of key-value pairs, where the key is a combination of overlaid cell values in a set of rasters, and the value is a count of such cells. The inputs are a GeoJSON shape and a [set of related arguments](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/base.py#L412) that specify rasters, operation type, CRS, etc.
```
hectcastro: Regarding the cached JSON, I think that it is worth considering how we might handle changes to the structure of the JSON object. For example, we cache something today with v1 of the structure, but then remove a field in v2. Tomorrow, we read an object from the cache with a v1 structure that is unexpected.

Some ideas:

- Include a version and ensure that we read that before parsing
- Make it clear that changes to this schema require a cache purge
- If we use the Django cache framework, it has support for [versioning](https://docs.djangoproject.com/en/1.8/topics/cache/#cache-versioning)

doc/arch/adr-002-geoprocessing-caching.md
```diff
@@ -0,0 +1,103 @@
...
+
+For a select few WKAoIs which are too large to process in the production infrastructure, such as HUC-8s which time-out during MapShed gathering phase, we can run their geoprocessing steps on more powerful infrastructure with longer timeouts in a batch process, cache the results, and then decommission it. This will make them available to regular users of the production app without needing the extra power to render them.
+
+### How will the items be identified, cached, and retrieved?
+
+Each geoprocessing run can be identified by two things: 1) input shape, and 2) JSON package sent to geoprocessing service. Because the input shape is a WKAoI, we can use it's table name and id as the identifier, for example `boundary_huc08__1`. The JSON package can be identified by the key in [`settings.GEOP['json']`](https://github.com/WikiWatershed/model-my-watershed/blob/a49cc115d18d2c67cbd0721cf62faa08e98d13fc/src/mmw/mmw/settings/base.py#L387), such as `nlcd_streams`, `nlcd_soils`, `slope`, etc.
+
+We can have a table structure as follows:
+
+| Column     | Type                     | Modifiers |
```
hectcastro: Instead of crafting a custom schema specifically for this use case, would it be better to use a [database cache backend](https://docs.djangoproject.com/en/1.8/topics/cache/#database-caching) within the existing Django cache framework? This aligns us with the well-defined cache framework API in the event that we want to swap out the backend.

doc/arch/adr-002-geoprocessing-caching.md
```diff
@@ -0,0 +1,103 @@
...
+
+While we could cache the entire output of a MapShed or TR-55 run, in the case of an update to some of the constituent raster or vector data would force recalculation of the whole. By caching only the time-taking geoprocessing results, we ensure that any updates to constituent rasters would invalidate only that specific cached result, leaving the others still current. And since vector data results would never be cached, they will always be current upon update as well.
+
+In case we do not foresee updating the rasters very often, it may be beneficial to consider caching the entire output.
+
+### Where will it be cached?
+
+The current app stack gives us two options to choose as a cache: Redis and PostgreSQL. While reads from Redis can potentially be a bit faster, the in-memory cache will be purged upon restart and deployment. Furthermore, the use case to optimize is a faster read by one user over time, rather than parallel reads by many users concurrently, the speed advantages of Redis are of little use.
```
hectcastro: Just to be clear, the contents of Redis is purged when restarted (in our configuration, but there is an option to enable persistence), but its contents remain unaltered across deployments.

doc/arch/adr-002-geoprocessing-caching.md
rajadain: Didn't know about this, thanks! Will study it a bit and amend the ADR to use that instead.

doc/arch/adr-002-geoprocessing-caching.md
rajadain: Ah that's good to know. Would you still otherwise agree that the database is the better cache for this use case?

doc/arch/adr-002-geoprocessing-caching.md
```diff
@@ -0,0 +1,103 @@
...
+
+```sql
+DELETE FROM geop_cache
+WHERE name = 'boundary_huc08__1';
+```
+
+And finally, to delete the entire cache, we can just do
+
+```sql
+DELETE FROM geop_cache;
```
kdeloach: I'm not sure how important this level of detail is. We could abstract these implementation details away with a utility script or management command no matter how we decide to handle caching.

doc/arch/adr-002-geoprocessing-caching.md
kdeloach: Great idea.

doc/arch/adr-002-geoprocessing-caching.md
kdeloach: May be simpler to just blow away the cache instead of versioning.

doc/arch/adr-002-geoprocessing-caching.md
```diff
@@ -0,0 +1,103 @@
...
+4. Update the geoprocessing submodule to take GeoJSON shape or WKAoI id, and geoprocessing type, and return the output. Update the Celery Tasks or Django Views which use the geoprocessing submodule to use the new interface.
+5. Add caching support to the geoprocessing submodule. All geoprocessing is done in two parts: `_start` and `_finish`. In the case of a cache hit, there should be a signal passed from `_start` to `_finish` that instructs it to fetch the value from the cache instead of `sjs_retrieve`.
+6. Update the UI to send WKAoI id instead of GeoJSON for predefined shapes. RWD and user defined shapes should still be sent as GeoJSON.
+
+## Consequences
+
+This will make the worst case scenario, a cache miss, slightly longer than it currently is, because we'll be checking the cache before doing the actual geoprocessing.
+
+We will have to monitor the characteristics of the cache table to see how it grows. If it gets too big too quickly, we may have to think of ways of routinely emptying the cache, or removing the least recently used entries, or something like that.
```
kdeloach: Using a Redis cache backend with a LRU/LFU eviction policy would negate these concerns.

doc/arch/adr-002-geoprocessing-caching.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Where will it be cached?
+
+The current app stack gives us two options to choose as a cache: Redis and PostgreSQL. While reads from Redis can potentially be a bit faster, the in-memory cache will be purged upon restart and deployment. Furthermore, the use case to optimize is a faster read by one user over time, rather than parallel reads by many users concurrently, the speed advantages of Redis are of little use.
+
+By storing the cached results in the database, we ensure they are valid and available after deployments and new releases, while still being considerably faster than before.
+
+### When will it be cached?
+
+When a user selects a WKAoI, and a request is made to `/analyze` or `/modeling`, if the WKAoI results haven't already been cached, we will run and cache them. Over time, we will build up the cache, so that when new users request the same WKAoIs, they will get the cached results.
```
kdeloach: We should think about the case where corrupt/invalid data is cached which causes an analyze job to fail repeatedly.

doc/arch/adr-002-geoprocessing-caching.md
rajadain: Agreed. This was more a demonstration of how the proposed implementation supported various invalidation requirements, but if we go with the Django cache module this may be handled through that and no longer apply. Implementation details shouldn't be in an ADR. Will update and remove.

lewfish:
I'll check again when it's been updated re: using Django caching, but this generally looks good.

doc/arch/adr-002-geoprocessing-caching.md
```diff
@@ -0,0 +1,103 @@
...
+
+We also sketch out an implementation plan, and consider consequences and side-effects.
+
+## Decisions
+
+### What will be cached?
+
+The output of the geoprocessing service for WKAoIs will be cached. This is almost always a JSON blob of key-value pairs, where the key is a combination of overlaid cell values in a set of rasters, and the value is a count of such cells. The inputs are a GeoJSON shape and a [set of related arguments](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/base.py#L412) that specify rasters, operation type, CRS, etc.
+
+While we could cache the entire output of a MapShed or TR-55 run, in the case of an update to some of the constituent raster or vector data would force recalculation of the whole. By caching only the time-taking geoprocessing results, we ensure that any updates to constituent rasters would invalidate only that specific cached result, leaving the others still current. And since vector data results would never be cached, they will always be current upon update as well.
```
lewfish: This first sentence doesn't seem entirely grammatical. Also, consider 'time-taking' -> 'time-consuming'

doc/arch/adr-002-geoprocessing-caching.md
```diff
@@ -0,0 +1,103 @@
+# 002 - Geoprocessing Caching
+
+## Context
+
+Geoprocessing calls for large shapes can take a long time to complete. While users can draw custom shapes, they can also pick from a list of predefined shapes in the system. Geoprocessing the same shape over and over when multiple users select it is wasteful and unnecessary. By caching the geoprocessing results of predefined shapes, or Well Known Areas of Interest (WKAoIs), we can improve user experience and application performance.
```
lewfish: It's not necessary, but It'd be nice to accompany this discussion with some hard benchmarking data about the time it takes for various operations (geoprocessing vs. database calls vs.  model running).

doc/arch/adr-002-geoprocessing-caching.md
rajadain: Just read into it a bit. The problem with using the database cache backend is that we don't want to use it as a Django middleware. We want to use it between Celery and SJS, independent of the Django layer.

On the other hand, we could choose to cache things between client and Django. This would be a more comprehensive cache, because it would include everything sent to the client, but in case of an update to one of the rasters, we'd have to invalidate _all_ the caches. We could no longer have partial cache invalidation.

doc/arch/adr-002-geoprocessing-caching.md
rajadain: Yeah, I'd rather invalidate the cache, or write a custom transformation by hand and apply to everything, than try and keep track of versions.

doc/arch/adr-002-geoprocessing-caching.md
rajadain: I should have emphasized this more, but the cache module is supposed to have a `force` flag that will bust the cache. We could have some treatment in the client / Django interaction to trigger it.

doc/arch/adr-002-geoprocessing-caching.md
rajadain: I haven't worked with Redis enough to be entirely comfortable with it, but that shouldn't stop us from not choosing it if it is the better option. I am biased towards database simply because I know how to interact with it better (for cache invalidation operations, etc). If there are enough voices in support of Redis, I will switch to that.

doc/arch/adr-002-geoprocessing-caching.md
kdeloach: I was thinking more about in terms of what happens when an unexpected event occurs while a user is interacting with the application. If we cache corrupt/invalid data for a particular boundary, then the analyze job will always fail for every user, for that boundary. There should be a way to autocorrect this without needing manual intervention. But let's wait and see if this happens in practice before thinking about it too much.

rajadain: You're right, we'll need some sort of validation mechanism before we cache the results.

doc/arch/adr-002-geoprocessing-caching.md
kdeloach: I think the advantage to using Redis is that it's already installed and configured so the barrier to entry is low compared to creating our own cache mechanism. We should be able to rapidly prototype caching AOIs using this. The only disadvantage I see is that it's unclear how (or if) using Redis could potentially clash with Celery.

doc/arch/adr-002-geoprocessing-caching.md
rajadain: Hmm, there is a [low level cache API](https://docs.djangoproject.com/en/1.8/topics/cache/#the-low-level-cache-api) which allows direct access. Perhaps this is the best solution, which uses standardized setup and allows switching the cache backend, while still being granular and manual enough for our needs.

lewfish: If you use the low level cache API, I don't see an easy way to invalidate the cache for a specific name or type (using the terminology from your proposed DB schema). You can invalidate an item based on its version, or set an entire key to null, but you can't invalidate based on a particular attribute of the key.

rajadain: Yeah. I'm also not sure how we would do out-of-band caching of very large WKAoIs manually. Let me see if these operations can be supported by the low level cache API, but otherwise we may end up going with a custom table.

kdeloach: Not sure how important manually invalidating cache is, but Redis is able to manage this automatically based on configuration. The Redis CLI has more features depending on how granular we need to be (ex. to manually invalidate objects).

doc/arch/adr-002-geoprocessing-caching.md
hectcastro: They'd both consume resources on the Redis host, but otherwise, the currently configured Redis cache operates on a separate Redis "database" from Celery (and Logstash/Beaver, which also uses Redis to relay log messages).

doc/arch/adr-002-geoprocessing-caching.md
rajadain: That's a good idea. I tried to think of a quick way of doing it, but after dismissing a bash script that hit a submit endpoint, saved the job id, and then counted each poll for it, thought it would be best as a Python script. I think we all have a shared sense of how long things take, and generally agree that database lookups are faster than model runs which are much faster than geoprocessing operations. If I can throw together that Python script quickly I'll include the results.

rajadain:
Just posted an update which switches to using Django Cache Framework and Redis. The one use case I'm worried about is in the **Consequences** section:

> Large WKAoIs that are processed out-of-band may get pushed out of the cache if they are not used and the cache exceeds its maximum size.

Given that this is a large reason for this effort, I'm still half-hearted about going with Redis instead of the Database.

Otherwise this is ready for another look.

rajadain:
> However, we should make sure that the modification (hash or other) is utilized in the cache key generation.

So far we were targeting only WKAoIs for caching. Modifications can be arbitrarily drawn, and thus whatever technique we use to generate a key for them (like the hash) could probably be extended for user-drawn AoIs as well. The argument against caching user-drawn AoIs was to optimally use the limited cache space, and would apply to these modifications as well.

Is the idea that we start with WKAoIs, then add modifications, and then potentially user-drawn shapes if the cache size allows? Or do we assume that the number of modifications will be orders of magnitude fewer than the number of user-drawn AoIs?

mmcfarland:
> So far we were targeting only WKAoIs for caching. Modifications can be arbitrarily drawn, and thus whatever technique we use to generate a key for them (like the hash) could probably be extended for user-drawn AoIs as well.

Sorry, I think that was misinterpreted.  I'm less concerned about _whether_ we cache WKAoI results for TR-55 scenarios with modifications (the space issue it legitimate), but rather that we have to consider them because they use the `nlcdSoilCensus` job, which has a definition stanza in the `geoprocessing.py` module.  I believe the modifications are [applied as the `geometry` attribute](https://github.com/WikiWatershed/model-my-watershed/blob/49a319212abc8a16681ce5e5d1c8b5213d7e5060/src/mmw/apps/modeling/geoprocessing.py#L132).

Consider a TR-55 job on a HUC-08.  The first scenario (current conditions) will have no additional `geometry` value in the census job.  It will be a WKAoI, and the results will be cached.  Then consider the next scenario, after a user has drawn a modification, supplied via `geometry`.  It's still an WKAoI with an Id, and the cache check will have a hit for that id and return the results, without having applied the modifications.  Thus we'll run the model with the wrong input.  I'm pointing out that for TR-55 WKAoIs, we'll have to consider the modifications to either have the cache key _include_ the concept of the modified area (thus making it distinct to other WKAoI runs) or have the task be aware that the job may have geometries which modify the rasters and thus never seek to cache them/read from the cache for them when they do.

I think there is marginal utility in caching the modified scenario TR-55 results, as it will only be applied in situations where a user deletes a second (or third, fourth, etc) modification.  Probably not a use case to prioritize.  My preference is that for `nlcdSoilCensus` jobs with `geometry != None` we always force execution.  

MapShed does not apply modifications at the raster level, so isn't susceptible to this kind of behavior and can be cached with impunity.

doc/arch/adr-002-geoprocessing-caching.md
lewfish: It sounds like we're confident of where the bottleneck is, so I wouldn't worry about it.

rajadain:
Okay. I'll add some of the model-specific design ideas to the ADR as well.

mmcfarland:
Sure - they can be pretty light for this ADR.  I mostly just wanted to point out an important implementation detail so we're aware of it, as a consequence of doing the geop json bit.

rajadain:
Add a simple paragraph capturing our intentions with specific models. If there are no more comments by end of day, I'll merge this. Thanks for the input, everyone!

PR #1844, opened 2017-04-26
Style cleanup

jfrankl:
## Overview

Style cleanup for the recent MMW features.

### Demo

image

image

image

image

### Notes

- I need a developer to make the map smaller on the initial view and the select area view. The sidebar needs to be smaller in those situations.
- I added four variables, `$temp-h1`, `$temp-h2`, `$temp-h3`, `$temp-h4`. I am hoping to standard some of the font sizes across the application. This is the first step in setting that up for the future.

## Testing Instructions

 * Walk through the process from selecting an area to modeling.

### DIFF
--- a/src/mmw/sass/components/_sidebar.scss
+++ b/src/mmw/sass/components/_sidebar.scss
@@ -6,27 +6,63 @@
         background-color: $black-06;
         border-radius: 0;
         border: 0;
-        color: $black-74;
         font-weight: 800;
+        background-color: #389b9b !important;
+        color: #fff;
     }
     .nav-tabs > li {
         width: 50%;
         margin-bottom: -1px;
         float: left;
+        top: 2px;
+        position: relative;
+        & > a {
+            height: 40px;
+            width: 130px;
+            line-height: 40px;
+            color: #fff;
+            font-size: $temp-h3;
+            position: relative;
+            top: 10px;
+            border-radius: 2px 0px 0px 2px;
+        }
+        &:first-child {
+            a {
+                float: right;
+            }
+        }
+        &:last-child {
+            a {
+                float: left;
+            }
+        }
+        &.active {
+            a {
+
+            }
+
+        }
     }
     .nav-tabs > li > a {
         margin: 0;
         border-radius: 0;
         border: 0;
         background-color: $black-12;
-        height: $sidebar-tab-subheader-height;
-        line-height: $sidebar-tab-subheader-height;
         padding-top: 0;
         text-align: center;
         color: $black-39;
+        height: 36px;
+        line-height: 36px;
     }
     .nav-tabs {
         border-bottom: 0;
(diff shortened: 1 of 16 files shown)


rajadain:
In the Draw View:

image

 1. The search icon / spinner should align with text baseline, or be vertically centered
 2. The splash content seems a little small and too cornered with the amount of whitespace around it. Can we horizontally center it?

jfrankl:
@rajadain 
1. Good catch on that. I'll fix it.
2. I was hoping a developer could make the map smaller on those views to reduce the amount of white space. The sidebar doesn't need to be as wide for that intro screen and for the select an area (I tried stretching out that content to be the same width as the sidebar in analyze/model, and it looked off).

rajadain:
Also, can we make the `+` icon a bit more vertically centered in the zoom control?

image

rajadain:
Do you mean to make the map larger and take extra space from the sidebar?

rajadain:
In the Analyze view, the contrast ratio between foreground and background of "Model this area ▾" button is too low:

image

It currently has a [contrast ratio of 2.59:1](http://webaim.org/resources/contrastchecker/?fcolor=828282&bcolor=d4d4d4). Accessibility guidelines recommend a contrast ratio of at least 4.5:1 for 14pt text.

Furthermore, it is also the primary action button at this stage in the application. It almost looks disabled, and may confuse users as to what they must do to proceed. Compare this to current design:

image

The primary action button is bright and draws the eye, and the text is in sharp contrast and easy to read.

rajadain:
This may be out of scope, but can we make the "Download this data" button either more button-y or more link-y? Currently it looks like a label and it isn't obviously clickable until you hover on it.

image

jfrankl:
@rajadain yes exactly, the map should be larger and take up space from the sidebar.

rajadain:
There are remnants of touch boundaries on the map controls that don't match the newer, smaller controls:

image

These can be observed on touch devices. If a touch device is not easily accessible, it can be simulated with Firefox by opening `about:config` and setting `dom.w3c_touch_events.enabled` to `1`.

src/mmw/sass/components/_sidebar.scss
```diff
@@ -6,27 +6,63 @@
...
+            }
+        }
+        &:last-child {
+            a {
+                float: left;
+            }
+        }
+        &.active {
+            a {
+
```
rajadain: Is there something missing here? Or expected to come in the future?

rajadain:
To confirm the sidebar behavior: there are now two widths, narrow for Splash and Draw screens, and as-wide-as-currently for Analyze and Model screens?

Could you add these widths as SASS variables so they can be reused?

rajadain:
Heads up that there are merge conflicts.

arottersman:
This got pulled into #1890, so we can close.

Issue #1837, opened 2017-04-20
Create BiGCZ api endpoint


kdeloach:
Create a Django app which exposes a single DRF endpoint to search BiGCZ data sources.

Stub out an interface to integrate with several external APIs to produce a list of data sources as the result.

Arguments:

* AOI or bounding box
* Search query (text)
* Date range

PR #1846, opened 2017-05-01
BiG CZ Search

kdeloach:
## Overview

Create BiG CZ Django app, initial search endpoint, and design pattern
for adding search catalogs.

The goal is to produce standard response formats for all catalogs by
parsing each response into Resource and ResourceList objects.

Includes initial implementation for HydroShare, CUAHSI, and CINERGI.

Connects #1837

### Demo

http://localhost:8000/api/bigcz/search?catalog=hydroshare&query=diamond&from_date=2000-01-01&to_date=2020-12-31&bbox=-180,-90,180,90

http://localhost:8000/api/bigcz/search?catalog=cinergi&query=diamond&from_date=2000-01-01&to_date=2020-12-31&bbox=-180,-90,180,90

http://localhost:8000/api/bigcz/search?catalog=cuahsi&bbox=9,9,10,10

### Notes

* Does not support pagination
* CUAHSI tends to timeout if the bbox is too large
* ~~All querystring arguments are required even though they are not universally supported by all catalogs~~

## Testing Instructions

* Checkout PR branch
* Provision `app` VM
* Test the `/api/bigcz/search` DRF endpoint (example values provided above)

### DIFF
--- a/src/mmw/apps/bigcz/clients/cinergi.py
+++ b/src/mmw/apps/bigcz/clients/cinergi.py
@@ -1,0 +1,97 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import division
+
+import requests
+import dateutil.parser
+
+from rest_framework.exceptions import ValidationError
+from apps.bigcz.models import Resource, ResourceList, BBox
+
+
+CATALOG_NAME = 'cinergi'
+GEOPORTAL_URL = 'http://cinergi.sdsc.edu/geoportal/rest/find/document'
+
+
+def parse_url(item):
+    """
+    Return "details" URL or first link if not found.
+    """
+    links = item['links']
+    if links:
+        for link in links:
+            if link['type'] == 'details':
+                return link['href']
+        return links[0]['href']
+    return None
+
+
+def parse_date(value):
+    try:
+        # Some dates are 0-based?
+        return dateutil.parser.parse(value)
+    except ValueError:
+        return None
+
+
+def parse_record(item):
+    return Resource(
+        id=item['id'],
+        title=item['title'],
+        description=item['summary'],
+        url=parse_url(item),
+        created_at=None,
+        updated_at=parse_date(item['updated']))
+
+
+def prepare_bbox(value):
+    box = BBox(value)
+    return '{},{},{},{}'.format(box.xmin, box.ymin, box.xmax, box.ymax)
+
+
+def prepare_date(value):
+    return value.strftime('%Y-%m-%d')
+
+
+def search(**kwargs):
+    query = kwargs.get('query')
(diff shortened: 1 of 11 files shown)


kdeloach:
I was able to populate additional info for CUAHSI results (service URL, description), by making an additional request for services, and joining the result.

kdeloach:
Something else to note for CUAHSI is that neither sites nor services have dates. Only "series records" do. 

There is a 1 to many relationship between Service and Sites, and also between Site and SeriesRecords. To enable a date range search we will need to make at least 3 API requests.

kdeloach:
I think this is in a good enough position to wire up to the frontend so be can begin testing this on staging.

src/mmw/apps/bigcz/views.py
```diff
@@ -0,0 +1,55 @@
...
+    search = SEARCH_FUNCTIONS.get(catalog)
+
+    if search:
+        try:
+            return search(**search_kwargs)
+        except ValueError as ex:
+            raise ParseError(ex.message)
+
+    raise ValidationError({
+        'error': 'Catalog must be one of these: cuahsi, cinergi, '
```
mmcfarland: Consider making this dynamic from `SEARCH_FUNCTIONS.keys()`

src/mmw/apps/bigcz/clients/cinergi.py
```diff
@@ -0,0 +1,87 @@
...
+        })
+
+    response = requests.get(GEOPORTAL_URL, params=params)
+
+    data = response.json()
+    if 'records' not in data:
+        raise ValueError(data)
+
+    results = data['records']
+    count = data['totalResults']
```
mmcfarland: For clarity, if `len(data['records']) < data['totalResults']`, it indicates that CINERGI has paged this query result?

src/mmw/apps/bigcz/views.py
kdeloach: Good idea.

src/mmw/apps/bigcz/clients/cinergi.py
kdeloach: Yes, but I expect this to change when we implement a more generalized pagination solution. CINERGI and HydroShare have slightly different pagination mechanisms.

kdeloach:
> No paging, even if the backend returned paged results

Created issue #1850 to address this.

> CHUASI doesn't use the query param

CUAHSI only supports bbox until we figure out a solution for keyword search. Decided to skip this for now since there are other outstanding questions for this data source.

> Resources don't include a lat/lng

CUAHSI is the only data source that provides lat/lng reliably. See #1854.

kdeloach:
Made a few changes to differentiate between required & optional params for each data source. For example, CUAHSI does not currently support date range search, so it doesn't make sense to require dates for this catalog.

PR #1847, opened 2017-05-01
Tweak search box style

kdeloach:
## Overview

Reduce padding and make dropdown content scrollable.

Connects #1838

### Demo

image

With reduced `max-height` to demonstrate overflow:

image

### Notes

I considered using `calc(100vh - 200px)` for `max-height` but I'm concerned about cross-browser compatibility issues. As a result, the search results dropdown can be clipped on small resolutions, even though overflow is enabled.

## Testing Instructions

* Checkout PR branch
* Run:
```
./scripts/bundle.sh
```
* Test in browser

### DIFF
--- a/src/mmw/sass/components/_inputs.scss
+++ b/src/mmw/sass/components/_inputs.scss
@@ -10,7 +10,7 @@
   font-size: 0.7rem;
   width: 300px;
   height: 36px;
-  padding: 0 1rem;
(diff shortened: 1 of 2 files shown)


kdeloach:
@jfrankl 

Let me know if this looks good or if we should tweak it some more. Thanks.

kdeloach:
Thanks.

PR #1853, opened 2017-05-02
Cleanup style

jfrankl:
I made the font size for the search results match the input field. Reformatted the search results category label.

### Demo

image


kdeloach:
Here's what it looks like on my machine:

image

Issue #1866, opened 2017-05-04
On project start, only show current conditions view


arottersman:
When first arriving to the model view, show "Current Conditions" without a scenario. Collapse enhancement controls into "+ Add changes to this area" button"

![screen shot 2017-05-04 at 12 05 18 pm](https://cloud.githubusercontent.com/assets/7633670/25713211/f621f422-30c1-11e7-8d19-dfa76951d4f2.png)

- Clicking the "Add changes to this area" button creates a new scenario

arottersman:
@jfrankl Will the "Add changes to this are" always be present on Current Conditions, or will it only appear when there are no other scenarios?

Issue #1865, opened 2017-05-04
Revamp scenario bar


arottersman:
Collapse scenario tab bar into the same bar as the modifications

- Remove tabs, and add scenarios to dropdown 
- Keep project name in bar — it will house the save button ( #1860) , analyze/model toggle, and change area button
- Move "Compare"

![screen shot 2017-05-04 at 11 59 47 am](https://cloud.githubusercontent.com/assets/7633670/25713094/8f9ea420-30c1-11e7-85c3-9b2981cc73c1.png)

@jfrankl 
Where should "share" and "duplicate" go?
<img width="345" alt="screen shot 2017-05-04 at 12 01 06 pm" src="https://cloud.githubusercontent.com/assets/7633670/25713067/6d3618d2-30c1-11e7-9a3d-c93d432241c5.png">

PR #1881, opened 2017-05-08
Start Model Stage With Only Current Conditions Scenario

arottersman:
## Overview
- Start each project off with Current Conditions as the only scenario
- Add button "+ Add changes to this area" to create new scenario

Eventually on model start we wont show any hint of scenarios at all, but I'll leave that work for #1865 which will have to rework the scenario views anyway.

**Wireframes**
screen shot 2017-05-08 at 1 08 15 pm

Connects #1866 

### Demo

tbjf5qlhli

## Testing Instructions

 * Pull this branch  and `./scripts/bundle.sh`
 * Create new projects with both models
    * Confirm both start with "Current Conditions" and there's a "Add changes to this area" button that on click creates a new scenario
    * Ensure the button does not appear when there are already multiple scenarios

### DIFF
--- a/src/mmw/js/src/modeling/controllers.js
+++ b/src/mmw/js/src/modeling/controllers.js
@@ -332,10 +332,6 @@ function setupNewProjectScenarios(project) {
         new models.ScenarioModel({
             name: 'Current Conditions',
             is_current_conditions: true,
-            aoi_census: aoiCensus
-        }),
-        new models.ScenarioModel({
-            name: 'New Scenario',
             active: true,
             aoi_census: aoiCensus
         })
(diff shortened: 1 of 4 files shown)


rajadain:
Taking a look now.

rajadain:
Tested, works well on both TR-55 and Mapshed projects. On Mapshed I notice that the button doesn't show up until the initial fetch is done, which is good.

I did notice that if I added a New Scenario and then deleted it, the button would not come back. I think it should.

arottersman:
Thanks, @rajadain. Going to merge this.

> I did notice that if I added a New Scenario and then deleted it, the button would not come back. I think it should.

This will be fixed with 978eeb6a7569d7a219d7e3aae2b43e22c644e611

Issue #1876, opened 2017-05-04
Use MapshedJob for TR-55


rajadain:
Currently we have two classes in the [geoprocessing library](https://github.com/WikiWatershed/mmw-geoprocessing): SummaryJob and MapshedJob. SummaryJob was created first, with the aim of counting NLCD and Soil Tiles for Analyze and TR-55 Models. Later, we added MapshedJob which supported a wider variety of operations, configurable via JSON input. The same JSON input can be used to produce the equivalent of SummaryJob results, but we haven't had the need to update Analyze and TR-55 Models so far.

Now, since we aim to [improve performance by caching geoprocessing results](https://github.com/WikiWatershed/model-my-watershed/pull/1843), this is a good opportunity to make the switch.

By having everything use a single geoprocessing class, we ensure that the nature of geoprocessing is the same for all models, that it is configurable enough to support multiple use cases, and that if we ever need to add caching at its level, we'll only need to add it in one place.

This task can be completed in three steps:

 - [x] Ensure that for the right JSON input, the results of MapshedJob are identical to SummaryJob.
 - [x] Switch Analyze and TR-55 Modeling to use MapshedJob
 - [x] Remove SummaryJob from Spark JobServer configuration

After these tasks, SummaryJob may be removed from the geoprocessing service entirely. This can be done separately.

PR #1882, opened 2017-05-08
Use MapshedJob for Analyze and TR-55

rajadain:
## Overview

This PR switches Analyze and TR-55 jobs to use the more configurable [MapshedJob](https://github.com/WikiWatershed/mmw-geoprocessing/blob/develop/summary/src/main/scala/MapshedJob.scala) instead of the older [SummaryJob](https://github.com/WikiWatershed/mmw-geoprocessing/blob/develop/summary/src/main/scala/SummaryJob.scala).

It also fixes a bug in how the results of NLCD / Soil histogram aggregation that would overwrite values instead of summing them.

Connects #1876 

### Notes

A deployment strategy will have to be devised since the bug fix in the first commit changes calculation of saved values in projects.

The final two commits remove and refactor unused code. If we would like to not remove that code at this time, those commits can be dropped.

## Testing Instructions

 * Checkout the first commit of this branch (which includes a fix for aggregating GeoTrellis histogram into a census, but excludes the switch to MapshedJob), and reload the Celery service

 * Open the network tab of developer tools, and run an Analyze job on a WKAoI (so it is easy to reproduce). Find the server response and save the JSON response.
 * Checkout the rest of this branch (which includes the switch to MapshedJob), and reload the Celery service

 * Run an Analyze job on the same shape. Save the JSON response.
 * Compare the two responses and ensure they are identical.
 * Try out Analyze, TR-55, and Mapshed jobs and ensure they all still work.

### DIFF
--- a/src/mmw/apps/modeling/geoprocessing.py
+++ b/src/mmw/apps/modeling/geoprocessing.py
@@ -192,7 +192,9 @@ def data_to_census(data):
             elif soil_str == 'cd':
                 soil_str = 'd'
             key_str = '%s:%s' % (soil_str, nlcd_str)
-            dist[key_str] = {'cell_count': count}
+            dist[key_str] = {'cell_count': (
+                count + (dist[key_str]['cell_count'] if key_str in dist else 0)
+            )}
 
     def after_rule(count, census):
         census['cell_count'] = count
--- a/src/mmw/apps/modeling/geoprocessing.py
+++ b/src/mmw/apps/modeling/geoprocessing.py
@@ -15,11 +15,15 @@ import json
 
 
 @statsd.timer(__name__ + '.sjs_submit')
-def sjs_submit(host, port, args, data, retry=None):
+def sjs_submit(data, retry=None):
     """
     Submits a job to Spark Job Server. Returns its Job ID, which
     can be used with sjs_retrieve to get the final result.
     """
+    host = settings.GEOP['host']
+    port = settings.GEOP['port']
+    args = settings.GEOP['args']
+
     base_url = 'http://{}:{}'.format(host, port)
     jobs_url = '{}/jobs?{}'.format(base_url, args)
 
@@ -68,12 +72,15 @@ def sjs_submit(host, port, args, data, retry=None):
 
 
 @statsd.timer(__name__ + '.sjs_retrieve')
-def sjs_retrieve(host, port, job_id, retry=None):
+def sjs_retrieve(job_id, retry=None):
     """
     Given a job ID, will try to retrieve its value. If the job is
     still running, will call the optional retry function before
     proceeding.
     """
+    host = settings.GEOP['host']
+    port = settings.GEOP['port']
+
     url = 'http://{}:{}/jobs/{}'.format(host, port, job_id)
     try:
         response = requests.get(url)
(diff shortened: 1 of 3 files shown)


rajadain:
It should be noted that the first commit in this branch is 2fec76d, not 78ef1b9 as GitHub shows:

```
* 5f48ebf - (HEAD, origin/tt/caching-1-mapshedjob-everywhere, tt/caching-1-mapshedjob-everywhere) Refactor, simplify (16 hours ago) <Terence Tuhinanshu>
* 62efac7 - Remove SummaryJob (16 hours ago) <Terence Tuhinanshu>
* 78ef1b9 - Switch to MapshedJob for Analyze, TR-55 (16 hours ago) <Terence Tuhinanshu>
* 2fec76d - Fix bug in census counting (16 hours ago) <Terence Tuhinanshu>
```

src/mmw/apps/modeling/geoprocessing.py
```diff
@@ -139,19 +141,25 @@ def histogram_finish(job_id, retry):
...
-            [k1, k2] = map(int, re.sub('[^0-9,]', '', k).split(','))
-            result.append(((k1, k2), v))
-        return result
-
     host = settings.GEOP['host']
     port = settings.GEOP['port']
 
-    data = sjs_retrieve(host, port, job_id, retry)
+    sjs_result = sjs_retrieve(host, port, job_id, retry)
+    data = {make_tuple(key[4:]): val for key, val in sjs_result.items()}
```
mmcfarland: Unclear what format conversion is going here, could use a comment to clear up why it's necessary.

src/mmw/apps/modeling/geoprocessing.py
```diff
@@ -139,19 +141,25 @@ def histogram_finish(job_id, retry):
...
+    # Map NODATA soil cells cells to 3
+    # This was previously done within the SummaryJob: https://github.com/WikiWatershed/mmw-geoprocessing/blob/0d95dee35e729d9fd2f58fb9e73a69dcbe61df61/summary/src/main/scala/SummaryJob.scala#L106  # NOQA
+    # but since MapshedJob doesn't support remapping, we do it here.
+    nlcd_soil_count = {}
+    for (n, s), count in data.iteritems():
+        s2 = s if s != settings.NODATA else 3
+        nlcd_soil_count[(n, s2)] = count + nlcd_soil_count.get((n, s2), 0)
+
+    result = []
+    for (n, s), count in nlcd_soil_count.iteritems():
```
mmcfarland: This is the 3rd iteration over (mostly) the same data set.  I was skeptical at first why they shouldn't just be processed in a single loop, but given the small number of elements, it seems a little more clarity is more important than efficiency.  I'd suggest a small comment to each step describing the transformation process.

rajadain:
Pushed a fixup 460bcd72b8e8797348233f1e63dd500c6f937df4 that combines two of the loops, makes the third loop more idiomatic (one-line), and adds some comments. Thanks for taking a look, will squash and merge in a bit.

rajadain:
I would love to combine all three loops into one, but I don't think we can. What we're trying to do is remap soil values and sum them, which is easier to do in a dictionary, so we do that first. Then the entire dictionary is converted to an array, which is the required output for backwards compatibility reasons.

Issue #1827, opened 2017-04-18
No error or warning should show when draw action is canceled.


caseycesari:
To reproduce:

- Navigate to the select area page.
- Expand the "Draw area" section.
- Select "Free Draw"
- Click the "X" in the "Draw area" section.
- An error modal is displayed.

Clicking the "X" should act as a cancel and no error message should be displayed.

caseycesari:
Related to https://github.com/WikiWatershed/model-my-watershed/issues/1835

arottersman:
Related to #1716

Issue #1716, opened 2017-03-01
Add ability to reject deferred in TaskModel


kellyi:
While investigating #1708 it appears that some/all of the TaskModel objects don't always `.reject` the deferred object on clicking Reset or otherwise cancelling polling for a job. One appearance of this bug is that sometimes RWD can complete after clicking reset; another's that it seems like the Modeling jobs aren't fully canceled on clicking the back button sometimes. 

We fixed the UI issues described in #1708, but we should investigate whether it's feasible to update the TaskModel to give it a direct way to cancel polling immediately by rejecting a the TaskModel's deferred.

PR #1891, opened 2017-05-16
Don't show error modal on canceling draw

kellyi:
## Overview

This PR remedies a bug whereby canceling a draw tool would lead to an error message display. To fix it, I added a guard to the `displayAlert` function in `draw/views` which checks whether there's a legit error message along with the alert.

If so, we display it -- as in the case where RWD starts and fails.
If not, we don't display anything -- as in the case where a draw tool or RWD is cancelled.

Connects #1827 

### Notes

An error message will still display if RWD is canceled after it has been started. This happens because canceling there doesn't actually stop the job from running, so it fails with an actual error message a few seconds later. Going to look into whether we might sort or ignore these messages, but we've got #1716 to handle rejecting the RWD task.

## Testing Instructions

 * get this branch then `bundle`, view it in the browser, and hard refresh
 * click on each draw tool, then click cancel after each and verify that you don't see the error modal
 * click on an RWD option then click cancel and verify that you don't see the warning modal

### DIFF
--- a/src/mmw/js/src/draw/utils.js
+++ b/src/mmw/js/src/draw/utils.js
@@ -8,6 +8,8 @@ var $ = require('jquery'),
     turfBboxPolygon = require('turf-bbox-polygon'),
     coreUtils = require('../core/utils');
 
+var CANCEL_DRAWING = 'CANCEL_DRAWING';
+
 // Keep in sync with src/api/main.py in rapid-watershed-delineation.
 var MAX_AREA = 75000; // About the size of West Virginia (in km^2)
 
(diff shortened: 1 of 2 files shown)


kellyi:
I need to check on whether the shapefile errors generate actual messages, btw.

kellyi:
> Prefer being explicit about rejecting promises purposefully vs rejecting them on actual failure. 

Sounds good to me. I'll push another commit including updates like that.

kellyi:
Pushed a fixup commit including the explicit check described above. When testing it out I did encounter a bug in the draw tools described in #1893, but it happens on staging as well so I created a new card.

kellyi:
I squashed the fixup and checked the commit again on my local. Everything still works as intended, so I'm going to merge. Thanks for your help with this!

Issue #1906, opened 2017-05-19
Add flag to disable geoprocessing caching


rajadain:
It can be useful to disable caching during development. Add a flag that does so.

rajadain:
Shares points with #1884

PR #1907, opened 2017-05-19
Add flag to disable geoprocessing caching

rajadain:
## Overview

Adds flag to disable geoprocessing caching. This is useful when exercising that part of the code during development.

Connects #1906

## Testing Instructions

 * Check out this branch. Reprovision `app` and `worker`
 * Download and apply this patch [caching-test.patch.txt](https://github.com/WikiWatershed/model-my-watershed/files/1014416/caching-test.patch.txt)

       git apply caching-test.patch.txt

 * Debug Celery

       ./scripts/debugcelery.sh

 * In a new terminal window, fire off an Analyze / Land request

       curl 'http://localhost:8000/api/modeling/start/analyze/land/' --data 'analyze_input=%7B%22wkaoi%22%3A%22huc12__55174%22%7D'

 * Observe the Celery output. Ensure you see the `==> geoprocessing caching is enabled` message.
 * In case this shape hadn't been cached before, it should be cached now. Run the same request again to ensure that the geoprocessing service doesn't run the second time.
 * Update `MMW_GEOPROCESSING_CACHE` to be 0, and restart the Django and Celery

       vagrant ssh app -c 'echo "0" | sudo tee /etc/mmw.d/env/MMW_GEOPROCESSING_CACHE'
       vagrant ssh app -c 'sudo restart mmw-app'
       ./scripts/debugcelery.sh

 * Send the above `curl` request again. Ensure that caching is not used this time, and that you see the `==> geoprocessing caching is disabled` message in Celery.
 * Delete the `MMW_GEOPROCESSING_CACHE` file and restart the services

       vagrant ssh app -c 'sudo rm /etc/mmw.d/env/MMW_GEOPROCESSING_CACHE'
       vagrant ssh app -c 'sudo restart mmw-app'
       ./scripts/debugcelery.sh

 * Run the `curl` request again. Ensure that caching is enabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,15 @@ Spark Job Server       | 8090 | [http://localhost:8090](http://localhost:8090)
 
 In order to speed up things up, you may want to consider leveraging the `vagrant-cachier` plugin. If installed, it is automatically used by Vagrant.
 
+To speed up geoprocessing, those requests are cached in Redis. To disable this caching for development purposes, set the value of `MMW_GEOPROCESSING_CACHE` to `0`:
+
+```bash
+$ vagrant ssh worker -c 'echo "0" | sudo tee /etc/mmw.d/env/MMW_GEOPROCESSING_CACHE'
+$ vagrant ssh worker -c 'sudo service celeryd restart'
+```
+
+To enable the geoprocessing cache simply set it to `1` and restart the `celeryd` service.
+
 ### Test Mode
 
 In order to run the app in test mode, which simulates the production static asset bundle, reprovision with `VAGRANT_ENV=TEST vagrant provision`.
--- a/src/mmw/mmw/settings/base.py
+++ b/src/mmw/mmw/settings/base.py
@@ -379,6 +379,7 @@ ITSI = {
 
 # Geoprocessing Settings
 GEOP = {
+    'cache': bool(int(environ.get('MMW_GEOPROCESSING_CACHE', 1))),
     'host': environ.get('MMW_GEOPROCESSING_HOST', 'localhost'),
     'port': environ.get('MMW_GEOPROCESSING_PORT', '8090'),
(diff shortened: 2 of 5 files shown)


arottersman:
Taking a look now!

src/mmw/mmw/settings/base.py
```diff
@@ -379,6 +379,7 @@ def get_env_setting(setting):
 
 # Geoprocessing Settings
 GEOP = {
+    'cache': bool(environ.get('MMW_GEOPROCESSING_CACHE', True)),
```
arottersman: `MMW_GEOPROCESSING_CACHE` comes in as a string and `bool('0') == True`. We'll need to cast to an `int` or something, or base this check on `MMW_GEOPROCESSING_CACHE` having some value.

rajadain:
Good catch. Updated to cast to `int`. Also added some instructions in the README's Caching section.

README.md
```diff
@@ -121,6 +121,16 @@ Spark Job Server       | 8090 | [http://localhost:8090](http://localhost:8090)
 
 In order to speed up things up, you may want to consider leveraging the `vagrant-cachier` plugin. If installed, it is automatically used by Vagrant.
 
+To speed up geoprocessing, those requests are cached in Redis. To disable this caching for development purposes, set the value of `MMW_GEOPROCESSING_CACHE` to `0`:
+
+```bash
+$ vagrant ssh app -c 'echo "0" | sudo tee /etc/mmw.d/env/MMW_GEOPROCESSING_CACHE'
```
arottersman: Thanks for adding this. I think it has to be on `worker` instead of `app` (or perhaps both?)
```bash
vagrant ssh worker -c 'echo "0" | sudo tee /etc/mmw.d/env/MMW_GEOPROCESSING_CACHE'
```

rajadain:
Updated instructions to be more reasonable. Also added a commit that adds a `geop_` prefix to our cache keys, as suggested in the original ADR, to namespace them from other resources using Redis, and make wiping all geoprocessing keys easier.

rajadain:

Issue #1914, opened 2017-05-19
Style TR-55 For Narrow Viewports


arottersman:
With the new wireframes, there are now many buttons in the tr-55's model toolbar. As is, they won't all fit in a line on narrower screens. 

<img width="953" alt="screen shot 2017-05-19 at 12 57 14 pm" src="https://cloud.githubusercontent.com/assets/7633670/26260784/8fe96b68-3c9c-11e7-8eb7-cee6429ee61b.png">

(with `white-space: nowrap` gets cutoff)
<img width="1042" alt="screen shot 2017-05-19 at 1 11 13 pm" src="https://cloud.githubusercontent.com/assets/7633670/26260683/1e319d74-3c9c-11e7-9ea4-8c02e71f07fb.png">

/cc @jfrankl

Issue #1886, opened 2017-05-12
IE11 Analyze and Model Stages Broken On Staging


arottersman:
<img width="1280" alt="screen shot 2017-05-12 at 9 58 35 am" src="https://cloud.githubusercontent.com/assets/7633670/26001398/3e7f3ca0-36fa-11e7-857d-d2df1d0b9122.png">

Issue #1867, opened 2017-05-04
Style modification controls


arottersman:
![screen shot 2017-05-04 at 12 08 44 pm](https://cloud.githubusercontent.com/assets/7633670/25713358/70294aea-30c2-11e7-96fd-73c46dace774.png)

The skinny precipitation slider can leverage styling used by layerpicker's opacity slider.

Issue #1911, opened 2017-05-19
IE11: Can't Add Conservation Practices To Mapshed


arottersman:
Broken on production + staging:
`Object doesn't support property or method 'assign'`

[Here's the offending line](https://github.com/WikiWatershed/model-my-watershed/blob/ad32f519df10b035ddc182a2f03578b9891582a8/src/mmw/js/src/modeling/models.js#L888)

![screen shot 2017-05-19 at 1 55 00 pm](https://cloud.githubusercontent.com/assets/7633670/26260349/d7968d58-3c9a-11e7-9364-34a99ce6fd76.png)

To reproduce:
* On IE11, model an AoI via Mapshed
* Add a conservation practice
* Note that none appears

PR #1915, opened 2017-05-19
Fix IE11 Analyze/Model Views

arottersman:
## Overview

The analyze and model sidebars were blank in IE11, because of [this change to the bootstrap stylesheets](https://github.com/WikiWatershed/model-my-watershed/commit/a100f08183d2b7083455013dd8fd83204305317d#diff-4e8e01e9df294745f58db61a059a2a4fR225). This PR reverts the change, and fixes the model toolbar styling which broke without it. 

Connects #1886 

### Demo

_Analyze no longer blank on ie11_
screen shot 2017-05-19 at 1 41 15 pm

_put this thing on the right so it doesn't overlap with sidebar_
screen shot 2017-05-19 at 1 48 49 pm

_put everything on the right like they are in wireframes_
screen shot 2017-05-19 at 1 41 26 pm

_make sure elements don't wrap; while  #1914 is open_
screen shot 2017-05-19 at 1 11 13 pm

## Notes
-  #1867 will handle actual control styling

## Testing Instructions

 * On all major browsers, test both model stages in their "Current Conditions" view and "Scenario" view
     - Confirm all dropdowns, popovers, and tooltips are styled correctly

### DIFF
--- a/src/mmw/js/src/modeling/controls.js
+++ b/src/mmw/js/src/modeling/controls.js
@@ -24,7 +24,7 @@ var ControlView = Marionette.LayoutView.extend({
     model: models.ModelPackageControlModel,
 
     className: function() {
-        return 'inline ' + this.getControlName();
+        return 'inline control-item ' + this.getControlName();
     },
 
     initialize: function(options) {
--- a/src/mmw/js/src/modeling/templates/controls/modDropdown.html
+++ b/src/mmw/js/src/modeling/templates/controls/modDropdown.html
@@ -2,7 +2,7 @@
     <button class="btn btn-sm btn-primary dropdown-button" type="button" aria-expanded="true">
         <i class="fa fa-plus-circle"></i> {{ controlDisplayName }}
     </button>
-    <div class="dropdown-menu draw-tools menu-left" role="menu"> <!-- Dropdown -->
+    <div class="dropdown-menu menu-right" role="menu"> <!-- Dropdown -->
         <div class="mod-content-region"></div>
     </div> <!-- End Dropdown -->
 </div> <!-- End Conservation Practice Tools -->
--- a/src/mmw/js/src/modeling/templates/controls/precipitation.html
+++ b/src/mmw/js/src/modeling/templates/controls/precipitation.html
@@ -1,4 +1,4 @@
-<div class="btn-sm btn-primary">
+<div class="btn btn-sm btn-primary">
 <label>
     <span class="name">Precipitation</span>
     <input type="range" min="0" max="25" step="0.25" />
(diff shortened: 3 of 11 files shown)


ajrobbins:
Just a note that putting "everything" on the right including layer picker leads to https://github.com/WikiWatershed/model-my-watershed/issues/1916 /cc @jfrankl

kellyi:
About to take a look at this!

kellyi:
Possibly for another issue but the MapShed conservation practices selector doesn't scroll on IE11:

screen shot 2017-05-23 at 12 57 17 pm

kellyi:
Scenario name for MapShed's also not showing up in IE11:

screen shot 2017-05-23 at 1 00 15 pm

kellyi:
Does show up for TR-55 though:
screen shot 2017-05-23 at 1 01 05 pm

kellyi:
Not sure whether this was intended or should be spun off to another card but the TR-55 Modifications tab takes up the full screen width in IE11:
screen shot 2017-05-23 at 1 02 25 pm

kellyi:
For comparison, here's that tab in Chrome:
screen shot 2017-05-23 at 1 03 44 pm

arottersman:
- re: scrolling thumb select; I made #1910 to address...it's a pre-existing condition
- re: modifications taking up whole screen + no scenario name in ie11; I'll push a fixup to address! That was most definitely me.

kellyi:
Some of the TR-55 conservation practice text gets cut off in Firefox:

screen shot 2017-05-23 at 1 10 21 pm

kellyi:
Aside from the quirks mentioned above this seems to be working well! I was able to use the models in Chrome, Safari, Firefox, and IE11.

If this one's easy to fix as part of this, might be worth adding a fixup for it too:

> Scenario name for MapShed's also not showing up in IE11:

arottersman:
re: cutoff text in FF; I snuck that one at some point onto #1897. It's unfortunately on staging

arottersman:
I was able to fix the mod dropdown taking up the whole width, but couldn't reproduce the missing scenario name:

screen shot 2017-05-23 at 4 25 26 pm

kellyi:
Trying this out again; I think I'm able to reliably reproduce the missing scenario name like this in IE11:

1. run MapShed on an AOI
2. when those results return, click to "Add Changes..."
3. select and apply a conservation practice
4. click on the scenarios tab to go back to "Current Conditions"

At that point it looks like the scenario name disappears and the dropdown won't really work. I checked this on Chrome/macOS and it didn't seem to have the same issue btw.

kellyi:
TR-55 modifications tab looks correct now, btw. I think the missing scenario thing in that one situation is the only other breaking IE11 issue I encountered.

If it's not too difficult to address, might be good to handle it as part of this work; if it's super involved to fix, we could spin it off into another issue.

arottersman:
Oh, that's probably related to #1911. There's an `Object.assign` in there that breaks the whole thing after adding a conservation practice.

kellyi:
> There's an Object.assign in there that breaks the whole thing after adding a conservation practice.

Got it. Ok, this is otherwise good to go!

arottersman:
Thanks!

PR #1919, opened 2017-05-24
Collections API ADR

rajadain:
Adds ADR overview of the investigation into Collections API, and proposes an approach to implement it.

Also tagging @echeipesh for any feedback / clarifications.

Connects https://github.com/WikiWatershed/mmw-geoprocessing/issues/43

### DIFF
--- a/doc/arch/adr-003-collections-api.md
+++ b/doc/arch/adr-003-collections-api.md
@@ -1,0 +1,87 @@
+# Switching to Collections API
+
+## Motivation
+
+Currently the geoprocessing service runs in an Apache Spark container, using Spark JobServer as an HTTP interface. Layer reading is done using an RDD, which can help parallelize heavy operations, such as counting the number of cells in a large raster. However, since the geoprocessing service runs on a single machine it does not take advantage of this parallelism, resulting in significant overhead and little payout.
+
+## Benefits
+
+GeoTrellis 1.0 adds a [Collections API for layer reading](https://geotrellis.github.io/scaladocs/latest/#geotrellis.spark.io.s3.S3CollectionLayerReader) which returns a Scala Sequence rather than an RDD. For the MMW use case this is a better fit, since the only geoprocessing operation we do is counting cells and averaging values, both of which are faster on a single machine using Collections API which removes all the Spark overhead. In our use case, this can lead to performance improvements of 300–500%.
+
+Here is a table of performance improvements for NLCD and Soil requests, such as used for TR-55, for a HUC-12, a HUC-10, and a HUC-8, taken on `develop` using current service, on staging using current service, and a prototype service using the Collections API:
+
+| nlcd-soils-huc12 | develop | staging | collections-api |
+|------------------|--------:|--------:|----------------:|
+| 1                |    3.35 |    3.07 |            1.61 |
+| 2                |     2.9 |    2.59 |            1.54 |
+| 3                |    3.24 |    2.21 |            1.52 |
+| 4                |    3.57 |    2.07 |            1.35 |
+| 5                |    3.06 |    2.28 |            1.97 |
+
+| nlcd-soils-huc10 | develop | staging | collections-api |
+|------------------|--------:|--------:|----------------:|
+| 1                |    8.53 |    6.37 |            5.57 |
+| 2                |    7.72 |    4.55 |            4.97 |
+| 3                |    8.59 |    4.12 |            2.93 |
+| 4                |    7.15 |    3.83 |            2.83 |
+| 5                |    6.59 |    3.77 |            2.58 |
+
+| nlcd-soils-huc8 | develop | staging | collections-api |
+|-----------------|--------:|--------:|----------------:|
+| 1               |   31.18 |   16.48 |            8.42 |
+| 2               |   26.28 |   14.98 |            6.02 |
+| 3               |    31.1 |   14.36 |            4.89 |
+| 4               |    27.2 |   15.05 |            4.51 |
+| 5               |   29.27 |   12.48 |            4.09 |
+
+| average          | develop | staging | collections-api |
+|------------------|--------:|--------:|----------------:|
+| nlcd-soils-huc12 |   3.224 |   2.444 |           1.598 |
+| nlcd-soils-huc10 |   7.716 |   4.528 |           3.776 |
+| nlcd-soils-huc8  |  29.006 |   14.67 |           5.586 |
+
+Here's a chart of the table above:
+
+![image](https://cloud.githubusercontent.com/assets/1430060/26412141/c6eb7e26-4075-11e7-9b65-207bb286c171.png)
+
+The difference in `develop` and staging are because of two factors: staging workers are more powerful, and have much faster access to S3 tiles. These advantages should carry over to the Collections API variant as well, thus making it even faster when deployed.
+
+In addition, this will make our code simpler to understand and maintain, as well as make provisioning and deployment simpler by removing Spark and Spark JobServer. Since the geoprocessing service can now be packaged into a single JAR, we can also remove Docker and run it directly using OpenJDK 8.
+
+These improved speeds may also allow us to increase the area of interest threshold on the front-end, allowing users to select even bigger areas than before and get faster results back.
+
+## Challenges
+
+GeoTrellis 1.0 is compiled for Java 8, whereas the current MMW environment is setup with Java 7. We will have to make a number of updates to the MMW environment to support Java 8.
+
+Since Spark JobServer is no longer used, we will have to switch to using Akka HTTP for exposing the Geoprocessing service. This will also require updates to the relevant geoprocessing tasks in Celery. Provisioning scripts will also need to be updated.
+
+Changes to the geoprocessing service may result in differences in how cells are counted, and may lead to different results for the same inputs. This is already seen to some extent in [mmw-geoprocessing#44](https://github.com/WikiWatershed/mmw-geoprocessing/pull/44). It is difficult to say which is _more_ correct, the current or the new implementation. But we will have to explain and justify the changes.
+
+The Collections API has so far never been deployed to production. It comes with risks of "unknown unknowns", and we may have to solve unforeseen technical challenges.
+
+## Drawbacks
+
+Removing Spark JobServer will lead to the interaction between Celery and Geoprocessing becoming synchronous rather than asynchronous. This should be a tolerable difference, since the new synchronous requests should be 3–5 times faster than the old asynchronous ones, and a Celery task waiting for 5-10 seconds should be manageable. The Celery tasks should timeout after a sensible period.
+
+The Collections API supports far fewer MapAlgebra operations than the RDD API. This is okay for now, as we are only counting cells and averaging their values. However, if we were to require MapAlgebra operations in the future, we may have to add back the RDD API.
+
+## Approach
+
+ 1. Make a set of use cases that cover all the functionality of the current geoprocessing service
+ 2. Make a new project within the geoprocessing repo, and add an Akka HTTP service
+ 3. Add endpoints that take in the same JSON input as the current service, and perform the same operations, just using the Collections API instead
+ 4. Fine tune results so they match with the previous ones, or explain and justify why they are different
+ 5. Update geoprocessing service deployment to push JAR to GitHub instead of a Docker image to Quay
+ 6. Update MMW provisioning to remove Spark JobServer and Docker dependencies, add OpenJDK 8, and pull in geoprocessing JAR from GitHub
+ 7. Update geoprocessing Celery tasks to use the new service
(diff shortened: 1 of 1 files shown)


doc/arch/adr-003-collections-api.md
```diff
@@ -0,0 +1,82 @@
...
+
+ 1. Make a set of use cases that cover all the functionality of the current geoprocessing service
+ 2. Make a new project within the geoprocessing repo, and add an Akka HTTP service
+ 3. Add endpoints that take in the same JSON input as the current service, and perform the same operations, just using the Collections API instead
+ 4. Fine tune results so they match with the previous ones, or explain and justify why they are different
+ 5. Deploy the new geoprocessing service to Quay
+ 6. Update provisioning to use the new service
+ 7. Update geoprocessing Celery tasks to use the new service
+ 8. Test the UI to ensure values are still similar and sensible
+ 9. Remove Spark and Spark JobServer from provisioning
```
caseycesari: I may be mistaken, but I thought I remembered hearing that Spark is still required when using the Collections API.

caseycesari:
I know next to nothing about Akka, but on the surface, it seems a little risky to use dependencies which are labeled experimental. https://github.com/WikiWatershed/mmw-geoprocessing/compare/tt+ec/collections-api#diff-74acf5e95f1c2ad5c84ec0b3ffe97957R14

doc/arch/adr-003-collections-api.md
```diff
@@ -0,0 +1,82 @@
+# Switching to Collections API
+
+## Motivation
+
+Currently the geoprocessing service runs in an Apache Spark container, using Spark JobServer as an HTTP interface. Layer reading is done using an RDD, which can help parallelize heavy operations, such as counting the number of cells in a large raster. However, since the geoprocessing service runs on a single machine it does not take advantage of this parallelism, resulting in significant overhead and little payout.
+
+## Benefits
+
+GeoTrellis 1.0 adds a Collections API for layer reading, very similar to the native ones in Scala, which can read layers and count cells without using Spark. For the MMW use case this is a better fit, since the only geoprocessing operation we do is counting cells and averaging values, both of which are faster on a single machine using Collections API which removes all the Spark overhead. In our use case, this can lead to performance improvements of 300–500%.
```
caseycesari: Consider adding a link to the Collections API documentation, if there is any.

doc/arch/adr-003-collections-api.md
kellyi: Not fully sure I understand it yet but I think the RDD class from Spark is still required; not certain about the rest of it.

kellyi:
> I know next to nothing about Akka, but on the surface, it seems a little risky to use dependencies which are labeled experimental.

Could we use a prior version of Akka? Or is there something about the version above that's particularly beneficial?

rajadain:
> dependencies labeled experimental

Good catch. That was used since much of that code was copied from one of the latest GeoTrellis demos. Practically we'll probably use the latest stable version, possibly the same used in Raster Foundry https://github.com/azavea/raster-foundry/blob/fd90e3b4bc1f696d01b5362536c85165b659b715/app-backend/project/Dependencies.scala#L7

mmcfarland:
To get a better sense of the timings you posted, those were for the same job-type; one directly submitting to SJS and the other directly to an akkahttp endpoint?  That is, the `develop` & `staging` numbers don't include celery job overhead and the like?

rajadain:
> directly submitting to SJS and the other directly to an akkahttp endpoint?

Correct. Here are the scripts I used, to post the files and get five results for each:

### develop

```fish
for i in 1 2 3 4 5; time -p http --timeout=90 :8090/jobs sync==true timeout==90 context==geoprocessing appName==geoprocessing-1.2.0 classPath==org.wikiwatershed.mmw.geoprocessing.MapshedJob < nlcd-soils-request-huc12.json 2>&1 > /dev/null | grep real | awk '{print $2}'; end

for i in 1 2 3 4 5; time -p http --timeout=90 :8090/jobs sync==true timeout==90 context==geoprocessing appName==geoprocessing-1.2.0 classPath==org.wikiwatershed.mmw.geoprocessing.MapshedJob < nlcd-soils-request-huc10.json 2>&1 > /dev/null | grep real | awk '{print $2}'; end

for i in 1 2 3 4 5; time -p http --timeout=90 :8090/jobs sync==true timeout==90 context==geoprocessing appName==geoprocessing-1.2.0 classPath==org.wikiwatershed.mmw.geoprocessing.MapshedJob < nlcd-soils-request-huc8.json 2>&1 > /dev/null | grep real | awk '{print $2}'; end
```

### staging

```bash
for ((n = 0; n < 5; n++)); do { time -p curl --silent --data-binary @nlcd-soils-request-huc12.json 'http://localhost:8090/jobs?sync=true&timeout=90&context=geoprocessing&appName=geoprocessing-1.2.0&classPath=org.wikiwatershed.mmw.geoprocessing.MapshedJob' ;} 2>&1 | grep real | awk '{print $2}'; done

for ((n = 0; n < 5; n++)); do { time -p curl --silent --data-binary @nlcd-soils-request-huc10.json 'http://localhost:8090/jobs?sync=true&timeout=90&context=geoprocessing&appName=geoprocessing-1.2.0&classPath=org.wikiwatershed.mmw.geoprocessing.MapshedJob' ;} 2>&1 | grep real | awk '{print $2}'; done

for ((n = 0; n < 5; n++)); do { time -p curl --silent --data-binary @nlcd-soils-request-huc8.json 'http://localhost:8090/jobs?sync=true&timeout=90&context=geoprocessing&appName=geoprocessing-1.2.0&classPath=org.wikiwatershed.mmw.geoprocessing.MapshedJob' ;} 2>&1 | grep real | awk '{print $2}'; done
```

### collections-api

```fish
for i in 1 2 3 4 5; time -p http :8091/submit < nlcd-soils-request-huc12.json 2>&1 > /dev/null | grep real | awk '{print $2}'; end

for i in 1 2 3 4 5; time -p http :8091/submit < nlcd-soils-request-huc10.json 2>&1 > /dev/null | grep real | awk '{print $2}'; end

for i in 1 2 3 4 5; time -p http :8091/submit < nlcd-soils-request-huc8.json 2>&1 > /dev/null | grep real | awk '{print $2}'; end
```

And here are the files themselves:

 * [nlcd-soils-request-huc12.json.txt](https://github.com/WikiWatershed/model-my-watershed/files/1026439/nlcd-soils-request-huc12.json.txt)
 * [nlcd-soils-request-huc10.json.txt](https://github.com/WikiWatershed/model-my-watershed/files/1026437/nlcd-soils-request-huc10.json.txt)
 * [nlcd-soils-request-huc8.json.txt](https://github.com/WikiWatershed/model-my-watershed/files/1026436/nlcd-soils-request-huc8.json.txt)

doc/arch/adr-003-collections-api.md
rajadain: Played around with this a bit, and it seems like I cannot remove the [`com.apache.spark.spark-core`](https://github.com/WikiWatershed/mmw-geoprocessing/blob/a81ac7e65441a3517d11dcd6f70f49565eb300ac/api/build.sbt#L16) compilation dependency. I can remove the [`hadoop-client` and `spark-jobserver`](https://github.com/WikiWatershed/mmw-geoprocessing/blob/a81ac7e65441a3517d11dcd6f70f49565eb300ac/api/build.sbt#L17-L18) ones.

Even though we need Spark as a compilation dependency, I don't think we need it installed in the deployment environment. I certainly don't have Spark running in my local as far as I know. I think everything from Spark that is needed gets packaged into the JAR, which can then simply be run directly.

kellyi: I'm not super confident about this being correct it but it seemed like I needed both of these imports to happen together in whatever file to get `geotrellis.spark` to work -- which entailed having `spark-core` around in `build.sbt` --

```scala
import geotrellis.spark._
import org.apache.spark.rdd.RDD
```

mmcfarland: I'm sure if we go down this route, we'll have to spend some time cleaning, updating and changing the related dependencies.  I think it's enough to know that it would be a sufficiently large change to our provisioning, and we would not be operating a spark local or cluster after the fact, whether or not it remained an application dependency.

doc/arch/adr-003-collections-api.md
```diff
@@ -0,0 +1,82 @@
...
+
+Since Spark JobServer is no longer used, we will have to switch to using Akka HTTP for exposing the Geoprocessing service. This will also require updates to the relevant geoprocessing tasks in Celery.
+
+Changes to the geoprocessing service may result in differences in how cells are counted, and may lead to different results for the same inputs. This is already seen to some extent in [mmw-geoprocessing#44](https://github.com/WikiWatershed/mmw-geoprocessing/pull/44). It is difficult to say which is _more_ correct, the current or the new implementation. But we will have to explain and justify the changes.
+
+The Collections API has so far never been deployed to production. It comes with risks of "unknown unknowns", and we may have to solve unforeseen technical challenges.
+
+## Drawbacks
+
+Removing Spark JobServer will lead to the interaction between Celery and Geoprocessing becoming synchronous rather than asynchronous. This should be a tolerable difference, since the new synchronous requests should be 3–5 times faster than the old asynchronous ones, and a Celery task waiting for 5-10 seconds should be manageable.
```
hectcastro: Agreed with converting these calls to be synchronous. In addition, I think that we should cap the HTTP request timeout to the backend geoprocessing service to some number of seconds and mark the request as a failure if it exhausts the timeout.

doc/arch/adr-003-collections-api.md
```diff
@@ -0,0 +1,82 @@
...
+
+The Collections API supports far fewer MapAlgebra operations than the RDD API. This is okay for now, as we are only counting cells and averaging their values. However, if we were to require MapAlgebra operations in the future, we may have to add back the RDD API.
+
+## Approach
+
+ 1. Make a set of use cases that cover all the functionality of the current geoprocessing service
+ 2. Make a new project within the geoprocessing repo, and add an Akka HTTP service
+ 3. Add endpoints that take in the same JSON input as the current service, and perform the same operations, just using the Collections API instead
+ 4. Fine tune results so they match with the previous ones, or explain and justify why they are different
+ 5. Deploy the new geoprocessing service to Quay
```
hectcastro: If the artifact of the revised geoprocessing service ends up being a single JAR, then I think it might be better to bypass Docker altogether and just install OpenJDK 8 on the machine and wrap the JAR in an Upstart script.

doc/arch/adr-003-collections-api.md
```diff
@@ -0,0 +1,82 @@
...
+
+Removing Spark JobServer will lead to the interaction between Celery and Geoprocessing becoming synchronous rather than asynchronous. This should be a tolerable difference, since the new synchronous requests should be 3–5 times faster than the old asynchronous ones, and a Celery task waiting for 5-10 seconds should be manageable.
+
+The Collections API supports far fewer MapAlgebra operations than the RDD API. This is okay for now, as we are only counting cells and averaging their values. However, if we were to require MapAlgebra operations in the future, we may have to add back the RDD API.
+
+## Approach
+
+ 1. Make a set of use cases that cover all the functionality of the current geoprocessing service
+ 2. Make a new project within the geoprocessing repo, and add an Akka HTTP service
+ 3. Add endpoints that take in the same JSON input as the current service, and perform the same operations, just using the Collections API instead
```
hectcastro: This is mimicking the SJS API or something completely new?

rajadain: The HTTP interface will be completely new, built using Akka HTTP a-la Raster Foundry instead of SJS. It will likely be a single endpoint that takes the same payload as the current SJS `POST /jobs` endpoint, but instead of returning a job ID to poll with it will return the actual values.

rajadain:
What's also interesting about the numbers:

image

Is that while the RDD implementations seem to have linear or super-linear growth (keep in mind that the difference between HUC-12, HUC-10, and HUC-8 is usually an order of magnitude), the Collections API grows sub-linearly and far more slowly.

doc/arch/adr-003-collections-api.md
rajadain: Even better!

rajadain:
I also just confirmed that the output of the experimental collections API branch are identical to that of https://github.com/WikiWatershed/mmw-geoprocessing/pull/44. That is, the only difference is caused by switching from intersecting with a polygon instead of an extent.

rajadain:
Another potential for optimizing performance is adjusting parallelism of S3 reads using configuration here: https://github.com/locationtech/geotrellis/blob/master/s3/src/main/resources/reference.conf

rajadain:
Added a **Descision** section outlining the conditions moving forward. Squashed fixups, merging now. Thanks for all your input!

kdeloach:
Nice write up. 👍

Issue #1889, opened 2017-05-15
Warning modal for shapefile upload with multiple features


ajrobbins:
If there are multiple features in the shapefile: Use the first feature in the shapefile as the AoI, and show a warning to the user that they we’re using the first feature, and if they want to use a different one they need to export it as a single feature shapefile. We should use our standard error/warning modal to display this message.

PR #1927, opened 2017-05-26
Display alert when shapefile has multiple features

mmcfarland:
## Overview

If the shapefile contains additional records after the first is read
in, alert the user that the first feature was used. Their only recourse
is to create a new shapefile with a more strategically selected feature.

Connects #1889 

### Demo

screenshot from 2017-05-26 12 54 35

## Testing Instructions

* Add a shapefile with [multiple features](https://github.com/WikiWatershed/model-my-watershed/files/1032520/cty_animals.zip) and see the new warning modal.
* Add a shapefile with a [single feature](https://github.com/WikiWatershed/model-my-watershed/files/1032521/single.zip) and receive no such information.

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -400,6 +400,13 @@ var AoIUploadView = Marionette.ItemView.extend({
                         if (result.done) { return; }
                         var geom = reproject.toWgs84(result.value, prj);
                         self.addPolygonToMap(geom);
+                        return source.read();
+                    })
+                    .then(function hasMore(result) {
+                        if (!result.done) {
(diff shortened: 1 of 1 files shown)


src/mmw/js/src/draw/views.js
```diff
@@ -400,6 +400,13 @@ var AoIUploadView = Marionette.ItemView.extend({
                         if (result.done) { return; }
                         var geom = reproject.toWgs84(result.value, prj);
                         self.addPolygonToMap(geom);
+                        return source.read()
```
arottersman: Heads up this line is missing a semi and is causing the build to fail

jfrankl:
That language makes sense.

We should get some user feedback on this feature. When someone uploads a shapefile with multiple features, I wonder if they keep moving forward with the first feature as the AoI, or if they'll always want to go back and use a different file.

Issue #1940, opened 2017-06-01
Show BiG CZ search results bounds on map


kdeloach:
We should show all search result bounds on the map. Hovering over individual results should highlight the map feature. Details view should only display bounds for the currently selected result.

Something else I'd like to do as part of this task is to change the API so that it returns GeoJSON instead of 4 floats for the bounding box. This should be more flexible and make it easier to support different kind of boundaries. For example, some services return points as well as boxes, and other services may return multiple points/boxes per result.

![screenshot 2017-06-01 at 12 26 35](https://cloud.githubusercontent.com/assets/43062/26690054/e2e2ad7e-46c5-11e7-9d01-84d5690e5987.png)
![screenshot 2017-06-01 at 12 32 27](https://cloud.githubusercontent.com/assets/43062/26690194/6586140a-46c6-11e7-9162-857aa46dbbcd.png)

PR #1948, opened 2017-06-05
Show data catalog results on map

kdeloach:
## Overview

Displays all search results on the map. Hover over a search result
to highlight the geom. This also improves the API by returning
GeoJSON for each result instead of a bounding box, to support
points, polygons, and multipolygons.

Connects #1940

#### CINERGI

image

image

#### CUAHSI

image

image

### Notes

Spatial fields are not available for HydroShare so there's nothing to draw.

## Testing Instructions

* Run `./bundle.sh`
* Go to [http://localhost:8000/search](http://localhost:8000/search)
* Try different CINERGI and CUAHSI searches
* Verify that shapes appear on the map
* Verify that hovering over results highlights the corresponding shape

### DIFF
--- a/src/mmw/apps/bigcz/clients/cinergi.py
+++ b/src/mmw/apps/bigcz/clients/cinergi.py
@@ -15,7 +15,7 @@ GEOPORTAL_URL = 'http://132.249.238.169:8080/geoportal/opensearch'
 
 
 def parse_date(value):
-    return dateutil.parser.parse(value)
+    return dateutil.parser.parse(value) if value else None
 
 
 def parse_envelope(envelope):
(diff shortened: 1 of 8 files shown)


rajadain:
One interesting behavior, which I don't if it can be fixed or should be at this stage, is that the geometries disappear when dragged off the map, and don't appear until dragging stops when brought back into view:



But it's not a deal breaker, so we probably shouldn't spend too much time thinking about it.

kdeloach:
Interesting. I don't we're doing anything special that would cause this behavior. I'll see if I can reproduce this issue.

mmcfarland:
> disappear when dragged off the map, and don't appear until dragging stops

This is actually the normal Leaflet behavior, I imagine it removes svg elements off screen to save memory.  You can sketch a shape on http://geojson.io and observe the same effect.

kdeloach:
Oh, I didn't realize you were holding the mouse pressed down for so long. Why would you do that? :stuck_out_tongue:

Issue #1951, opened 2017-06-07
Health check is broken for production build


hectcastro:
When AMIs are created for deployment in AWS, the health check reports the following error:

```
ERROR:django.request:Internal Server Error: /health-check/                               
Traceback (most recent call last):          
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 108, in get_response                                                                           
    response = middleware_method(request)   
  File "/opt/app/mmw/middleware.py", line 25, in process_request                         
    callback_kwargs = resolver.resolve(request.path_info)                                
  File "/usr/local/lib/python2.7/dist-packages/django/core/urlresolvers.py", line 365, in resolve                                                                                 
    for pattern in self.url_patterns:       
  File "/usr/local/lib/python2.7/dist-packages/django/core/urlresolvers.py", line 401, in url_patterns                                                                            
    patterns = getattr(self.urlconf_module, "urlpatterns", self.urlconf_module)          
  File "/usr/local/lib/python2.7/dist-packages/django/core/urlresolvers.py", line 395, in urlconf_module                                                                          
    self._urlconf_module = import_module(self.urlconf_name)                              
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module             
    __import__(name)                        
  File "/opt/app/mmw/urls.py", line 12, in <module>                                      
    import apps.bigcz.urls                  
  File "/opt/app/apps/bigcz/urls.py", line 8, in <module>                                
    from apps.bigcz import views            
  File "/opt/app/apps/bigcz/views.py", line 13, in <module>                              
    from apps.bigcz.clients import SEARCH_FUNCTIONS                                      
  File "/opt/app/apps/bigcz/clients/__init__.py", line 6, in <module>                    
    from . import hydroshare, cuahsi, cinergi                                            
ImportError: cannot import name hydroshare  
```

rajadain:
Hmm, it is working locally:

```http
> http :8000/health-check/

HTTP/1.1 200 OK
Connection: keep-alive
Content-Encoding: gzip
Content-Type: application/json
Date: Wed, 07 Jun 2017 12:50:57 GMT
Server: nginx/1.10.1
Transfer-Encoding: chunked
Vary: Accept-Encoding

{
    "caches": [
        {
            "default": {
                "ok": true
            }
        }
    ],
    "databases": [
        {
            "default": {
                "ok": true
            }
        }
    ]
}
```

[...]

PR #1952, opened 2017-06-07
Remove relative import

kdeloach:
## Overview

Attempt to resolve this transient error:

> ImportError: cannot import name hydroshare

Connects #1951


kdeloach:
This is the best solution I could think of without more information. Because I can't reproduce the issue locally and the error fixed itself on staging it's hard to say what impact this will have.

kdeloach:
Agreed.

Issue #1904, opened 2017-05-18
Load BiG CZ settings based on HTTP Host


kdeloach:
Instead of relying on the `MMW_BIGCZ_MODE` environment variable to pick the correct application settings, check the HTTP `Host` header field.

The rationale is that we'd like to host both MMW and BiG CZ using the same infrastructure, but using different URLs.

We may need to keep the environment variable around for local development, but we should be able to remove all of the provisioning logic from the deployment scripts.

hectcastro:
It is possible that we want to use the `Host` header here vs. the `Referer` header, with a default case when neither domain matches (thinking about when we test the dark stack directly with the ELB domain).

mmcfarland:
Good point about the dark stack, hadn't considered that.  We could supplement our test on the appropriate header with an optional query string parameter that could force a particular mode to aid in testing both frontends in a dark stack when we haven't got a reliable `Host`.

kdeloach:
Thanks. We definitely intended to detect this setting based on the HTTP `Host` header field. Updated the issue title and description to reflect this.

PR #1953, opened 2017-06-07
BiG-CZ Middleware

rajadain:
## Overview

Previously we were setting BiG-CZ specific settings based on an environment variable. However, since in live deployment both MMW and BiG-CZ will be running off the same (virtual) servers, we cannot rely on environment variables to distinguish them.

This approach looks for a `?bigcz=true` query parameter. If found, it sets a session variable which turns on all relevant switches. These settings remain until the user’s session expires, or they pass in `?bigcz=false` which turns them off.

In addition, we also inspect the request’s `HTTP_HOST` header, and if it matches the BiG-CZ pattern we turn these settings on. If it matches the MMW pattern we turn these settings off.

Also tagging @mmcfarland and @hectcastro for review.

Connects #1904 

### Notes

Since this is an additional middleware, it will ever-so-slightly slow down _all_ requests.

The query strings are for development support. They will not work on the canonical URLs, i.e. you cannot use `?bigcz=true` on `staging.app.wikiwatershed.org` and have it be BiG-CZ, or vice-versa. The `HTTP_HOST` will trump the query parameter every time.

For the `HTTP_HOST` cases, we're setting the session variable _in every request_. This is redundant, but since each request is completely independent, I couldn't think of a way to not do this.

## Testing Instructions

 * Check out this branch
 * Go to [:8000/?bigcz=true](http://localhost:8000/?bigcz=true) and type in `clientSettings.data_catalog_enabled` in the console. Ensure its value is `true`.
 * Go to [:8000](http://localhost:8000/) without the query parameter, and ensure that `data_catalog_enabled` is still `true`
 * Got to [:8000?bigcz=false](http://localhost:8000/?bigcz=false) and ensure that `data_catalog_enabled` is now `false`
 * Go to [:8000](http://localhost:8000/) without the query parameter, and ensure that `data_catalog_enabled` is still `false`
* Try to spoof the `Host` header to `portal.bigcz.org` or `staging.portal.bigcz.org` and ensure that `daata_catalog_enabled` is `true`


rajadain:
I also just pushed d61189a which is an alternative implementation, a little uglier but potentially more efficient.

mmcfarland:
Instead of attaching to a user session, can it just be added directly to the `request` object?  Or perhaps not through middleware but through a context processor?  Since we really just need to get the flag to the view in order to set the client setting attribute, we may be able to narrow the scope.  The session business is little concerning since it may affect switching between the domains.  A context processor would essentially just parse the actual request and hand of the value to the view.

rajadain:
> context processor

That's a promising idea. Will read up a bit [on them](https://docs.djangoproject.com/en/1.8/ref/templates/api/#django.template.RequestContext).

rajadain:
Closing in favor of #1954.

PR #1954, opened 2017-06-08
Set BiG-CZ Mode by Host or Query Parameter

rajadain:
## Overview

Previously we were setting BiG-CZ specific settings based on an environment variable. However, since in live deployment both MMW and BiG-CZ will be running off the same (virtual) servers, we cannot rely on environment variables to distinguish them.

This approach looks for a `?bigcz=true` query parameter, or the **Host** header of the request coming from `[*.]portal.bigcz.org`. If the request is coming from that host, or that query parameter is passed, we enable BiG-CZ settings. Host inspection will be used in production, and query parameter is to support development.

Also tagging @mmcfarland and @hectcastro for code review.

Connects #1904 
Supersedes #1953 

## Testing Instructions

* Check out this branch
 * Go to [:8000/?bigcz=true](http://localhost:8000/?bigcz=true) and type in `clientSettings.data_catalog_enabled` in the console. Ensure its value is `true`.
 * Go to [:8000](http://localhost:8000/) without the query parameter, and ensure that `data_catalog_enabled` is `false`
* Try to spoof the `Host` header to `portal.bigcz.org` or `staging.portal.bigcz.org` and ensure that `data_catalog_enabled` is `true`


src/mmw/apps/home/views.py
```diff
@@ -116,6 +116,10 @@ def set_url(layer):
 
 
 def get_client_settings(request):
+    # BiG-CZ mode applies when either ?bigcz=true query param present, or
+    # request host contains predefined host. This covers staging sites, etc.
+    bigcz = (request.GET.get('bigcz') == 'true' or
```
hectcastro: I don't really know enough to be confident of how this plays out in the Python interpreter, but I think that it might be better to set the second condition first here. The way it stands now, in staging/production the `?bigcz=true` condition will always be evaluated first (and likely be false the vast majority of the time), which is slightly wasteful. Flipping the conditions should make staging/production have the shortest path to evaluation.

src/mmw/apps/home/views.py
```diff
@@ -116,6 +116,10 @@ def set_url(layer):
 
 
 def get_client_settings(request):
+    # BiG-CZ mode applies when either request host contains predefined host, or
+    # ?bigcz=true query parameter is present. This covers staging sites, etc.
+    bigcz = (settings.BIGCZ_HOST in request.META.get('HTTP_HOST') or
```
kdeloach: There's also [request.get_host](https://docs.djangoproject.com/en/1.11/ref/request-response/#django.http.HttpRequest.get_host) which may be a little more robust.

src/mmw/apps/home/views.py
```diff
@@ -116,6 +116,10 @@ def set_url(layer):
...
+             request.GET.get('bigcz') == 'true')
```
kdeloach: We could decide that the presence of `bigcz` in the query params enables this mode, no matter what the value is. So you could do `?bigcz=1` or `?bigcz=t` etc.

rajadain: At first I had it set to unset the session variable on `?bigcz=false`, but since we're not doing that anymore I'll update this to just check for its presence.

kdeloach:
Changes look good. 

:+1:

rajadain:
Thanks for taking a look! Squashing and merging now.

Issue #1896, opened 2017-05-18
"Change Area" Should Clear AoI


arottersman:
Currently on the analyze page, the change area button navigates back to the draw view, but keeps the AoI on the map. We should clear the AoI, but still position the map to where the AoI was. 

_Button on bottom left_
![image](https://cloud.githubusercontent.com/assets/7633670/26205330/6a1fce92-3baf-11e7-8b60-6ac61fc1db39.png)

The "Change area" button in the project header should have the same behavior. 
<img width="578" alt="screen shot 2017-05-18 at 9 43 54 am" src="https://cloud.githubusercontent.com/assets/7633670/26205352/7ed1368c-3baf-11e7-90b5-46b9b5fd4f89.png">

Issue #1874, opened 2017-05-04
Add "Change Area" button


arottersman:
* May not be the final button text
- Clears the aoi and takes you back to the draw stage to start a new project
![screen shot 2017-05-04 at 12 39 46 pm](https://cloud.githubusercontent.com/assets/7633670/25714690/d037145e-30c6-11e7-8227-4c582b961427.png)

jfrankl:
- Final button text: "New project"
- This should take you back to the draw view with default zoom level

Issue #1860, opened 2017-05-04
Save project workflow; Add save button


arottersman:
The app will no longer save automatically. Add a save button with three states: "Save", "Saving", "Saved"
![screen shot 2017-05-04 at 11 44 24 am](https://cloud.githubusercontent.com/assets/7633670/25712275/3391f378-30bf-11e7-9f59-4b56cde8b897.png)
![screen shot 2017-05-04 at 11 44 34 am](https://cloud.githubusercontent.com/assets/7633670/25712274/3390c26e-30bf-11e7-8c09-67226cc78934.png)
![screen shot 2017-05-04 at 11 44 29 am](https://cloud.githubusercontent.com/assets/7633670/25712276/3392ee4a-30bf-11e7-9bb8-7e9650b8bd0f.png)

Show regular login modal when a guest user tries to save. We'll implement the new prompt in #1861

arottersman:
@jfrankl
Should save *only* ever happen when you click save? Some actions/behaviors in particular I'm wondering about:

- Naming the project
- Having already saved the project, autosaving modifications
- Naming a scenario

jfrankl:
Below is my understanding of the new save functionality. Would like confirmation from @ajrobbins and/or @mmcfarland.

- The initial save only happens when you click save
- After the initial save, autosaving begins on any changes ("Saved" changes to "Saving" to indicate that automatic saving is happening. Then returns to "Saved" when save is complete)
- Attempting to navigate away from your project before the project is saved will trigger an alert ("Do you want to leave this site? [Stay] [Leave]"), see wireframes

I don't think any actions except clicking the "Save" button should trigger a save. As an enhancement, I could imagine giving the user a reminder to save when they name their project, make 10 modifications without saving, etc.

PR #1957, opened 2017-06-09
Add "New Project" Button

arottersman:
## Overview

- Add a button the model stage's project header: "New Project" to clear the map and go back to draw
- update "Change Area" button in analyze to clear the AoI

Connects #1896 
Connects #1874 

### Demo

gu5mv2dvob

## Testing Instructions

 * Pull and bundle
 * In model stage check that the "new project" button takes you back to draw with the AoI cleared
 * In analyze stage check the "change area" button clears the AoI
 * Test that the header in model stage looks okay in IE11, FF, Safari, Chrome

### DIFF
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -294,17 +294,13 @@ var MapView = Marionette.ItemView.extend({
                 attributionControl: options.showLayerAttribution
             });
 
-        // Center the map on the U.S.
-        map.fitBounds([
-            [24.2, -126.4],
-            [49.8, -66.0]
-        ]);
-
         this._leafletMap = map;
         this._areaOfInterestLayer = new L.FeatureGroup();
         this._modificationsLayer = new L.FeatureGroup();
         this._dataCatalogResultsLayer = new L.FeatureGroup();
         this._dataCatalogActiveLayer = new L.FeatureGroup();
+
+        this.fitToDefaultBounds();
 
         if (!options.interactiveMode) {
             this.setMapToNonInteractive();
@@ -344,6 +340,14 @@ var MapView = Marionette.ItemView.extend({
         map.addLayer(this._modificationsLayer);
         map.addLayer(this._dataCatalogResultsLayer);
         map.addLayer(this._dataCatalogActiveLayer);
+    },
+
+    fitToDefaultBounds: function() {
+        // Center the map on the U.S.
+        this._leafletMap.fitBounds([
+            [24.2, -126.4],
+            [49.8, -66.0]
+        ]);
     },
 
     setupGeoLocation: function(maxAge) {
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -306,8 +306,6 @@ var MapView = Marionette.ItemView.extend({
             this.setMapToNonInteractive();
         }
 
-        var maxGeolocationAge = 60000;
-
         if (options.addZoomControl) {
             map.addControl(new L.Control.Zoom({position: 'bottomleft'}));
             // We're overriding css to display the zoom controls horizontally.
@@ -318,7 +316,7 @@ var MapView = Marionette.ItemView.extend({
         }
 
         if (options.addLocateMeButton) {
-            addLocateMeButton(map, maxGeolocationAge);
+            addLocateMeButton(map, coreUtils.MAX_GEOLOCATION_AGE);
         }
 
         if (options.addSidebarToggleControl) {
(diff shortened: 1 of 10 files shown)


kellyi:
> In model stage check that the "new project" button takes you back to draw with the AoI cleared

This works, but/and it seems not to take me back to the geolocated map center & extent. Thought it was worth making sure this was intentional?

kellyi:
> In analyze stage check the "change area" button clears the AoI

Likewise, this one keeps me at the extent & center of the AOI? Wondering if it's worthwhile to have the two back buttons have the same behavior?

kellyi:
Tried this out Chrome, Safari, and FF on macOS & IE11 on Windows and the header appeared properly in each and the functionality seemed to work well. Wondering if we should address the map extent qs above? One way to handle it is to normalize the behavior for each: either each back button stays with the center & zoom from the prior AOI or it refocuses on the continental US?

arottersman:
@jfrankl, @ajrobbins thoughts?

- I think we should fix it so the "New Project" button takes you back to the geolocated extent (if that's enabled)
- I think "Change Area" from analyze should take you back to where the AOI was (without the AOI on the map) for cases where you've found an area you're interested in but haven't gotten the AOI right

jfrankl:
@arottersman that sounds perfect to me.

arottersman:
@kellyi, 4063cfc should fix the geolocating not happening when you click "New Project"

kellyi:
Not sure if it's caused by these changes or just revealed by them, but it seems like using the back button doesn't stop the polling & eventually this error results in the console:

screen shot 2017-06-12 at 5 56 10 pm

kellyi:
(Doesn't seem breaking, btw).

kellyi:
The re-geolocation thing works well, however! Going to check it out in IE11.

kellyi:
Geolocation reset seems like it's not quite working in IE11, but I can't tell why. The correct JS code is there but it just sticks on the continental US map instead of zooming.

kellyi:
Seems to be working well otherwise! The geolocation reset does work in Firefox and Chrome, just not in IE.

I believe the console error shown above is the same as https://github.com/WikiWatershed/model-my-watershed/issues/1716 so we should try to address that soon.

arottersman:
- I think I'll look into that console error for #1860 —it'll touch related pieces of the model
- I haven't been able to get browserstack ie11's geolocate to work for staging or this branch. Trying to boot up the asus to test (can't run my windows VM and MMW at the same time)

arottersman:
Asus is all charged up, and I think the geolocation is working okay there (feel free to take and test). There's a bit of a delay, but it does it eventually.

kellyi:
Taking another look at this.

kellyi:
Works well in Chrome, Firefox, and IE. Safari seems to block geolocation on localhost, so I couldn't test that but the rest of the functionality worked well.

I did notice that the breadcrumb navigation may get slightly out of sync when you click New Project from the modeling phase: the analyze step has been removed from the stack, but the word "Analyze" remains on the breadcrumb:

screen shot 2017-06-13 at 6 13 26 pm

However, looks like we're dropping that altogether in #1958, so probably worth ignoring here.

src/mmw/js/src/core/views.js
```diff
@@ -346,13 +340,29 @@ var MapView = Marionette.ItemView.extend({
...
+        // Center the map on the U.S.
+        this._leafletMap.fitBounds([
+            [24.2, -126.4],
+            [49.8, -66.0]
+        ]);
+    },
+
+    /**
+    Attempt to locate the user and set the map to it
+    @param alwaysSetView -- override the map model setting `geolocationEnabled` (used to prevent
```
kellyi: I like this docstring style, so I'm in favor of keeping it even though the other method comments in this file are more terse.

src/mmw/js/src/modeling/templates/projectMenu.html
```diff
@@ -1,46 +1,53 @@
...
-                <li role="presentation"><a role="menuitem" tabindex="-1" id="itsi-clone">Embed in ITSI</a></li>
-            {% endif %}
-            {% if user and not itsi_embed %}
-                <li role="separator" class="divider"></li>
-                <li role="presentation"><a role="menuitem" tabindex="-1" data-url="/projects/">My Projects</a></li>
+        </button>
+        {% if editable or is_new or itsi %}
+            <ul class="dropdown-menu menu-right" role="menu">
+                {% if not is_new and editable %}
+                    <li role="presentation"><a role="menuitem" tabindex="-1" data-toggle="modal" id="share-project">Share</a></li>
```
kellyi: Consider breaking these element tags across several lines, one tag per line, to make this more legible.

arottersman:
Thanks for all your thorough testing!

arottersman:
@azavea-bot rebuild

Issue #1931, opened 2017-05-31
WDC GetSeriesCatalogForBox3 provides richer response and more search parameters than GetSeriesCatalogForBox2


emiliom:
@kdeloach, I think you (and I) have so far not done much testing on the WDC [`GetSeriesCatalogForBox3`](http://hiscentral.cuahsi.org/webservices/hiscentral.asmx?op=GetSeriesCatalogForBox3) API query. My "series" queries have focused on `GetSeriesCatalogForBox2` (eg, [this notebook](https://gist.github.com/emiliom/4fd9f194e7c295a5577e00441022692a)). But it turns out that `GetSeriesCatalogForBox3` has two important advantages:
- It includes more query parameters (from controlled vocabularies): `sampleMedium`, `dataType` and `valueType`. Out of those, probably only `sampleMedium` is actually valuable for our likely use cases
- More importantly, the response is considerably richer in information! That took me by surprise. It includes organization (data service) names, ID's and descriptions, plus other information. So we can do and show more with the response  -- though it'll also be more verbose.

I ran a test, and `GetSeriesCatalogForBox3` seems to work fine, though probably with the same current bugs and keyword limitations as in `GetSeriesCatalogForBox2`.

cc @aufdenkampe

kdeloach:
Thanks for looking in to this. Do we want to display series or sites? Our current implementation uses `GetSitesInBox2` combined with `GetServicesInBox2` ([Source](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/bigcz/clients/cuahsi.py#L63)). It doesn't look it would be difficult to use `GetSeriesCatalogForBox3` instead though.

emiliom:
Ah, I didn't fully catch that. Got it.

Using `GetSitesInBox2` combined with `GetServicesInBox2` for now seems ok. @aufdenkampe and I should first discuss what should be displayed in the results; depending on what we decide, a series response may make more sense.

BTW, for future consideration: it occurred to me that currently there are only 94 "data services" in WDC (ie, the maximum total that can be returned by `GetServicesInBox2`), and those change rarely. Maybe a future strategy for better search responsiveness could be to get all services in the US lower 48 (based on a `GetServicesInBox2` query) and cache them once a day, then just use that cache together with the actual query results from `GetSitesInBox2`.

kdeloach:
[...]

Issue #1858, opened 2017-05-03
Use GetSeriesCatalogForBox3 for CUAHSI search


kdeloach:
Replace GetSitesInBox2 call with GetSeriesCatalogForBox3. We'll need to collapse multiple variables into a single site result.

kdeloach:
Blocked until all errors are resolved with the `GetSeriesCatalogForBox3` operation. See #1855.

kdeloach:
We should also investigate memoizing the full list of services (depending on how many there are) so we don't need to make multiple requests per search.

PR #1959, opened 2017-06-12
Use GetSeriesCatalogForBox2 for CUAHSI search

kdeloach:
## Overview

Replace `GetSitesInBox2` with `GetSeriesCatalogForBox2` based on
feedback from #1931.

Connects #1858

### Notes

Switching from "sites" to "series catalog" has increased the volume of data points.

### Demo

Before (`GetSitesInBox2`):
image

After (`GetSeriesCatalogForBox2`):
image

### Testing

* Geocode location "Pitt, Washington"
* Select "Klickitat" Huc 8 boundary
* Click "Search data sources"
* Initiate search by typing in a search term and pressing enter
* Click the WDC tab

### DIFF
--- a/src/mmw/apps/bigcz/clients/cuahsi.py
+++ b/src/mmw/apps/bigcz/clients/cuahsi.py
@@ -3,37 +3,69 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import division
 
+from datetime import date
+
 from suds.client import Client
 from rest_framework.exceptions import ValidationError
 from django.contrib.gis.geos import Point
 
 from apps.bigcz.models import Resource, ResourceLink, ResourceList, BBox
+from apps.bigcz.utils import parse_date
 
 
 CATALOG_NAME = 'cuahsi'
 SOAP_URL = 'http://hiscentral.cuahsi.org/webservices/hiscentral.asmx?WSDL'
 
 
+DATE_MIN = date(1900, 1, 1)
+DATE_MAX = date(2100, 1, 1)
+DATE_FORMAT = '%m/%d/%Y'
+
+
 client = Client(SOAP_URL)
 
 
-def parse_geom(site):
-    lat = site['Latitude']
-    lng = site['Longitude']
+def parse_geom(record):
+    lat = record['latitude']
+    lng = record['longitude']
     return Point(lng, lat)
 
 
-def parse_record(site, service):
-    geom = parse_geom(site)
+def parse_details_url(record):
+    """
+    > Site URLs can be similarly constructed from some Services,
+    > such as https://waterdata.usgs.gov/nwis/uv/?site_no=14113000, when
+    > ServCode = "NWISDV" and location = "NWISDV:14113000"
+
+    Ref: https://github.com/WikiWatershed/model-my-watershed/issues/1931
+    """
+    location = record['location']
+    if 'NWISDV' in location:
+        parts = location.split(':')
+        if len(parts) == 2:
+            _, id = parts
+            return 'https://waterdata.usgs.gov/nwis/uv/?site_no={}'.format(id)
+    return None
+
+
+def parse_record(record, service):
+    geom = parse_geom(record)
+
+    links = [
+        ResourceLink('service', service['ServiceDescriptionURL'])
+    ]
+
+    details_url = parse_details_url(record)
+    if details_url:
+        links.append(ResourceLink('details', details_url))
+
     return Resource(
-        id=site['SiteCode'],
-        title=site['SiteName'],
+        id=record['location'],
+        title=record['Sitename'],
         description=service['aabstract'],
         author=None,
-        links=[
-            ResourceLink('details', service['ServiceDescriptionURL'])
-        ],
-        created_at=None,
+        links=links,
+        created_at=record['beginDate'],
         updated_at=None,
         geom=geom)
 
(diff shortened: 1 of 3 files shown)


rajadain:
Taking a look at this now.

src/mmw/apps/bigcz/clients/cuahsi.py
```diff
@@ -45,16 +77,43 @@ def find_service(services, service_code):
...
+    for record in series:
+        service = find_service(services, record['ServCode'])
         if service:
-            record = parse_record(site, service)
+            record = parse_record(record, service)
             result.append(record)
     return result
 
 
+def group_series_by_location(series):
```
rajadain: `series` seems to be an overloaded term here. This takes `series`, and returns a grouping, but the values in the group are a set of key-value pairs, and not a `series`. Or is this common subset of fields across catalogs the true identifier of a series?

src/mmw/js/src/data_catalog/controllers.js
```diff
@@ -14,10 +14,32 @@ var DataCatalogController = {
             'was_data_catalog_visible': true,
         });
 
-        var searchModel = new models.DataCatalogModel();
-        var resources = new models.ResourceCollection();
+        var form = new models.SearchForm();
 
-        var view = new views.DataCatalogWindow(searchModel, resources);
+        var catalogs = new models.Catalogs([
+            new models.Catalog({
```
rajadain: Don't need to specify `new models.Catalog` here, you can just give an array of objects and `new models.Catalogs([...])` will convert it to the right objects automatically.

src/mmw/js/src/data_catalog/models.js
```diff
@@ -5,74 +5,102 @@ var _ = require('underscore'),
...
+});
 
-    getDescription: function() {
-        return this.get('description') || '';
+var Result = Backbone.Model.extend({
+    defaults: {
+        id: '',
+        title: '',
+        description: '',
+        // GeoJSON
```
rajadain: Consider using inline comments for data types:

```js
geom: null,  // GeoJSON
links: null,  // Array
```

src/mmw/js/src/data_catalog/models.js
```diff
@@ -5,74 +5,102 @@ var _ = require('underscore'),
...
+    getSummary: function() {
+        var text = this.get('description') || '';
+        if (text.length <= DESCRIPTION_MAX_LENGTH) {
             return text;
         }
         var truncated = text.substr(0,
-            text.indexOf(' ', DESCRIPTION_LENGTH));
-        return truncated + '... ';
+            text.indexOf(' ', DESCRIPTION_MAX_LENGTH));
+        return truncated + '&hellip;';
```
rajadain: 👏

src/mmw/apps/bigcz/clients/cuahsi.py
kdeloach: I'm open to finding a better term for this. `GetSeriesCatalogForBox2` returns multiple variables per site/location, but we don't want to display individual variables-- just sites. So this groups each series by location and returns fields common to all variables. There are no aggregated values so we just store the first record and discard subsequent records.

This was implemented based on my interpretation of [this comment](https://github.com/WikiWatershed/model-my-watershed/issues/1931#issuecomment-306838234).

src/mmw/js/src/data_catalog/controllers.js
kdeloach: That's an interesting feature. I think I'd prefer to leave this as-is since I can't think of a compelling reason to not specify the type.

src/mmw/js/src/data_catalog/models.js
kdeloach: Updated.

src/mmw/js/src/data_catalog/controllers.js
rajadain: Mostly to reduce repetition in the code.

```js
var catalogs = new Catalogs([new Catalog({}), new Catalog({}), new Catalog({})]);
```

We've used it previously [here](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/analyze/models.js#L71).

But I don't have a hard preference. Type hints are valuable wherever we can get them, and if the formatting favors them then they should be kept.

kdeloach:
> Are the results eventually supposed to be restricted to the area of interest?

For the initial implementation, we planned to restrict results to the AOI bounding box. The next iteration may be able to support masking results using complex shapes, as a future enhancement.

src/mmw/apps/bigcz/clients/cuahsi.py
rajadain: The implementation does seem faithful to the description. Perhaps because of exposure to [`_.groupBy`](https://lodash.com/docs/4.17.4#groupBy), I assumed the function did something like this:

```python
items = [
    {'id': 1, 'color': 'red'},
    {'id': 2, 'color': 'red'},
    {'id': 3, 'color': 'blue'}]

def group_list_by_color(items):
    colors = {}
    for item in items:
        color = item['color']
        if color in colors:
            colors[color].append(item)
        else:
            colors[color] = [item]

    return colors
```

So that the output is grouped by a certain key but the value is the original list item itself:

```ipython
In [3]: group_list_by_color(items)
Out[3]:
{'blue': [{'color': 'blue', 'id': 3}],
 'red': [{'color': 'red', 'id': 1}, {'color': 'red', 'id': 2}]}
```

What it's actually doing is more like extracting distinct locations from the series rather than grouping the series by location. Does that make sense?

kdeloach: Because there are no aggregated fields, it's fair to say that this is more of a distinct operation rather than a grouping. They are very similar operations. The comparable SQL query may look like:

```sql
SELECT ServCode, ServURL, location, ...
FROM series
GROUP BY ServCode, ServURL, location, ...
```

Issue #1863, opened 2017-05-04
Add warning when leaving an unsaved project


arottersman:
Without autosaving, we'll need to ensure the user meant to leave the page if a project is unsaved:
![screen shot 2017-05-04 at 11 44 49 am](https://cloud.githubusercontent.com/assets/7633670/25712747/765d0d4a-30c0-11e7-81ea-20a4a4f7e3ad.png)

arottersman:
Started this here, but am traveling for the rest of sprint:
https://github.com/WikiWatershed/model-my-watershed/tree/arr/warn-user-when-they-try-escape

The leave-the-site browser warning shows up fine, but the backbone navigation is doing something bizarre. When you click the browser back button, it correctly shows a modal before going to analyze, but with the url hash already changed to`/analyze`. Clicking "Stay" results in a state where you stay on the model page, and the back buttons cycle through `/draw` `/analyze` `/` never taking you anywhere or showing the warning modal again.
![thzffquxza](https://user-images.githubusercontent.com/7633670/27499640-c57a6c8a-5832-11e7-9ed9-c4576706a639.gif)

(When you click "Leave" things are okay)

PR #1966, opened 2017-06-13
Add save button

arottersman:
## Overview

- Add save button for the initial project save (displays ["Save", "Saving', "Saved"])
- Auto-save after project has been manually saved by the save button

Connects #1860 

### Demo

(This is with throttling turned on so you can see the "Saving" state)
sc2nxlhsxr

## Notes

- ~Based off of #1957, see  099d5c1 on~

## Testing Instructions

 * Pull, `./scripts/bundle.sh`
 * While not logged in
    - Draw an AoI and try to save
    - Confirm you're prompted to login/signup and after successfully doing so the project saves

* While logged in
   - Create a new project
   - Confirm changes to the scenarios and the project do not cause the project to save
   - Confirm the save button saves the projects and any created scenarios
    - All subsequent changes to the project and scenarios save automatically

### DIFF
--- a/src/mmw/js/src/app.js
+++ b/src/mmw/js/src/app.js
@@ -118,10 +118,10 @@ var App = new Marionette.Application({
         }
     },
 
-    getUserOrShowLogin: function() {
+    getUserOrShowLogin: function(onSuccess) {
         this.user.fetch().always(function() {
             if (App.user.get('guest')) {
-                App.showLoginModal();
+                App.showLoginModal(onSuccess);
             }
         });
     },
--- a/src/mmw/js/src/compare/controllers.js
+++ b/src/mmw/js/src/compare/controllers.js
@@ -157,7 +157,7 @@ function saveAfterLogin(user, guest) {
         });
         // Save the origProject (as opposed to the currentProject)
         // to not include changes to inputs and results.
-        App.origProject.saveProjectAndScenarios();
+        App.origProject.saveExistingProjectAndScenarios();
     }
 }
 
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -147,7 +147,7 @@ var ProjectModel = Backbone.Model.extend({
             this.set('scenarios', new ScenariosCollection());
         }
 
-        this.set('user_id', App.user.get('id'));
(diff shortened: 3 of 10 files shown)


arottersman:
- I added 888aa20 to remove some vestigial code 
- Noticed a bug on staging while testing this. If you log in while the initial model job is still running, the job fails: `Failed to load /api/modeling/jobs/5378f08b-276f-419f-81f0-22ffd57783fb/: the server responded with a status of 404 (NOT FOUND)`. You can refresh the page to fix, but it seems like we're losing the celery job after we create the project

rajadain:
Taking a look at this now.

rajadain:
Having some trouble saving the projects. Demo'd to @arottersman who's looking in to it.

arottersman:
Issue was the snippet I took out with 888aa20 ... I've corrected it, so this should be ready for another look!

rajadain:
If a logged in user creates a new project but doesn't click "☁️ Save", but clicks "New Project", their work is lost without warning. Consider adding a confirmation modal that allows the user to recover from an accident if they mistakenly click "New Project" instead of "☁️ Save", especially likely since the buttons are right next to each other.

rajadain:
When I start logged out, and then sign in with an existing account, the workflow works. But if I sign up with a new account, after signing up and in, the project is not saved. Furthermore, the project menu doesn't populate with anything:

image

Also, clicking "☁️ Save" doesn't do anything, and the project cannot be saved.

arottersman:
> If a logged in user creates a new project but doesn't click "☁️ Save", but clicks "New Project", their work is lost without warning. Consider adding a confirmation modal that allows the user to recover from an accident if they mistakenly click "New Project" instead of "☁️ Save", especially likely since the buttons are right next to each other.

I think this would be better to address in #1863

arottersman:
Applying some changes from #1970 to fix the sign up error here. Thanks for catching that!

arottersman:
Should be fixed.

rajadain:
Taking another look.

rajadain:

arottersman:
Yeah, the callback passing makes me a little uncomfortable, too. It seems like the login/signup modals should be wrapped up as `Deferred` objects. 

Thanks for reviewing!

Issue #1909, opened 2017-05-19
Enhance Scenario Dropdown actions


jfrankl:
Here is an example of the current functionality of the actions in the Scenario Dropdown:

![image](https://cloud.githubusercontent.com/assets/1430060/26114464/592f3e4a-3a2b-11e7-881b-20b13296b981.gif)

This should be modified to match the following screenshots:

![Screen 1](https://cloud.githubusercontent.com/assets/1809908/26254778/bcd2f232-3c85-11e7-8581-35a4be90f1d4.png)
![Screen 2](https://cloud.githubusercontent.com/assets/1809908/26254777/bcd076ce-3c85-11e7-8c09-b580adc16545.png)
![Screen 3](https://cloud.githubusercontent.com/assets/1809908/26254776/bccfedd0-3c85-11e7-9fac-919ab2a97a92.png)

Changes:

- All actions should be hidden behind the ellipses
- Rather than inline, the actions should be displayed as a second menu. This seems possible by making only CSS changes. Please get in touch with me before working on this part.
- Opening the actions should occur on click or hover (with a slight delay)
- Only one of these actions menus should be displayed at once; if a new menu is opened, the previous one should close
- Actions should have labels in addition to icons

See the "New" button on Google Drive for a similar pattern:

![image](http://g.recordit.co/LNa3bAH7Xx.gif)

Issue #1887, opened 2017-05-12
Style Scenario Rename Workflow


arottersman:
See https://github.com/WikiWatershed/model-my-watershed/pull/1883#issuecomment-300795108

cc/ @jfrankl

jfrankl:
Could you please use the modal styling from renaming a project?

![image](https://cloud.githubusercontent.com/assets/1809908/26006823/e5ae2778-370b-11e7-92b7-d745adc69ed3.png)

arottersman:
I brought this in because it made it easier to complete #1984

PR #1984, opened 2017-06-21
Scenario Dropdown Updates

arottersman:
## Overview
- Consolidate all the scenario actions into a sub-dropdown menu to the right of the Scenario dropdown menu (removes hiding extra icons behind the ellipses)
- This initially broke the rename action because the button was no longer part of the same view as the editable text. While we could trigger a backbone event for the parent view, and then do the rename there in the same way, it seemed cleaner/faster to just implement #1887 and show a rename popup like the rest of the actions

(Ignore the typo in the branch name 😨 )

Connects #1909 
Connects #1887 

### Demo

c22evlaoj2

## Testing Instructions
 * For both models, open a project and confirm all the various scenario actions work
 * Confirm renaming a project works as before

### DIFF
--- a/src/mmw/js/src/modeling/templates/scenarioMenuItemOptions.html
+++ b/src/mmw/js/src/modeling/templates/scenarioMenuItemOptions.html
@@ -1,0 +1,31 @@
+{% if not is_new %}
+    <li class="scenario-options-dropdown-menu-item" role="presentation" role="presentation" data-action="share">
+        <i  class="fa fa-share-alt" title="Share"></i> Share
+    </li>
+{% endif %}
+{% if not is_current_conditions %}
+    <li class="scenario-options-dropdown-menu-item" role="presentation" li data-action="duplicate">
+        <i class="fa fa-copy" title="Duplicate"></i> Duplicate
+    </li>
+{% endif %}
+{% if not is_current_conditions %}
+    <li class="scenario-options-dropdown-menu-item" role="presentation" data-action="delete">
+        <i class="fa fa-trash" title="Delete"></i> Delete
+    </li>
+{% endif %}
+{% if is_gwlfe %}
+    <li class="scenario-options-dropdown-menu-item" role="presentation" data-action="export-gms">
+        <form id="export-gms-form" method="post" action="/api/modeling/export/gms/" target="_blank">
+            <input type="hidden" name="csrfmiddlewaretoken" value="{{ csrftoken }}">
+            <input type="hidden" name="filename" class="gms-filename" value="export.gms">
+            <!-- Put `gis_data` in single quotes, to keep any unescaped chars from breaking the template -->
+            <input type="hidden" name="mapshed_data" value='{{ gis_data }}'>
+        </form>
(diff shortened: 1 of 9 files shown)


rajadain:
Some tests are failing

arottersman:
@azavea-bot rebuild

arottersman:
Okay, the build is finally passing!

rajadain:
Taking a look now.

src/mmw/js/src/modeling/templates/scenarioMenuItemOptions.html
```diff
@@ -0,0 +1,30 @@
...
+    <li class="scenario-options-dropdown-menu-item" role="presentation" data-action="delete">
+        <i class="fa fa-trash" title="Delete"></i> Delete
+    </li>
+{% endif %}
+{% if is_gwlfe %}
+    <li class="scenario-options-dropdown-menu-item" role="presentation" data-action="export-gms">
+        <form id="export-gms-form" method="post" action="/api/modeling/export/gms/" target="_blank">
+            <input type="hidden" name="csrfmiddlewaretoken" value="{{ csrftoken }}">
+            <input type="hidden" name="filename" class="gms-filename" value="export.gms">
+            <input type="hidden" name="mapshed_data" value='{{ gis_data }}'>
```
rajadain: I know this carried over from the source from which this was copied, but let's prefer double quotes for all HTML attributes. Specifically, `value` uses single quotes which should be changed to double.

rajadain:
This could use feedback from @jfrankl, but in the implementation the second dropdown looks "disconnected" from the first:

image

As opposed to the mockup where they are more "connected":

image

I'm not sure which is better, but wanted to make sure we be intentional about our choice.

arottersman:
I also forgot the grey square around the ellipsis on click which may help the dropdown feel connected to the row

src/mmw/js/src/modeling/templates/scenarioMenuItemOptions.html
arottersman: Double quotes broke the template. I think`gis_data` has characters that need to be escaped in it. I'll add a comment

arottersman:
Gouda?
w7gpemasst

rajadain:
Cheese 👍

src/mmw/js/src/modeling/templates/scenarioMenuItemOptions.html
rajadain: Ah 🤦‍♂️ serves me right for being too nitpicky. And of course I'm the one who wrote this originally https://github.com/WikiWatershed/model-my-watershed/commit/2834f5ba0093ca968d1fe9fe744d2e0a9687c567

arottersman:
There's a bug on FF (IE11 + Safari are fine though)
screen shot 2017-06-23 at 9 55 52 am

arottersman:
I pushed a small fixup to address. @rajadain mind giving this a quick final glance before I merge?

rajadain:
On it

PR #1995, opened 2017-06-30
Warn Before Leaving Unsaved Project

rajadain:
## Overview

When leaving an unsaved project, shows a warning that the project is not saved. Since projects are not auto-saved anymore, it is possible that a user may accidentally leave a page with unsaved useful work. To prevent them from losing that work, we prompt them for saving it.

Connects #1863 

### Notes

I'm not super happy with this solution. @arottersman made a valiant effort in [this commit](https://github.com/WikiWatershed/model-my-watershed/commit/d6976aadbb87dc4bd626fad8bd1745e1e4e870d2), which I tried to further in [this branch](https://github.com/WikiWatershed/model-my-watershed/compare/arr/warn-user-when-they-try-escape...tt+arr/warn-user-when-they-try-escape), but to no avail.

The fundamental difficulty is the lack of a `beforeunload` event on _the route_. Since we're using a custom patched version of a rather old version of Backbone + Marionette Router, the code is a little difficult to reason about, and any additions therein will take us further from the standard. However, without that implementation, there will be edge cases which are not supported, such as:

 * When the user clicks "back" in the browser, there is no warning
 * If the user is logged in, when the click "My Projects" in the top right corner, there is no warning (and adding one may be difficult, because that button and "New Project" are hierarchically very far apart)

One unfortunate side effect of the `beforeunload` warning for the browser is that it pops up if the user tries to sign in with ITSI:

I'm not sure if we should merge this and make future cards to fix the breakage, or keep chugging on this until everything works.

## Testing Instructions

 * Check out this branch. Run `bundle`. Go to [:8000/](http://localhost:8000/)
 * Try closing the page in `/`, `/draw`, `/analyze`, and `/projects` routes. Ensure that there is no prompt.
 * Pick an AoI and analyze it and choose a model. Get to the `/project` route.
 * Try to close the tab without saving your work. Ensure you see a message like this:

    image

 * Try to reload the tab without saving your work. Ensure you see a message like this:

    image

 * Try and create a new project without saving your work. Ensure you see a message like this:

    image

/cc @jfrankl for advice on language in the last modal. The first two are controlled by the browser and cannot be reworded.

### DIFF
--- a/src/mmw/js/src/modeling/controllers.js
+++ b/src/mmw/js/src/modeling/controllers.js
@@ -105,6 +105,8 @@ var ModelingController = {
             }
             setPageTitle();
         }
+
+        setWarnBeforeLeaving(project);
     },
 
     makeNewProject: function(modelPackage) {
@@ -134,6 +136,8 @@ var ModelingController = {
             updateUrl();
             setPageTitle();
         }
+
+        setWarnBeforeLeaving(project);
     },
 
     projectCleanUp: function() {
@@ -275,6 +279,8 @@ function projectCleanUp() {
     App.getMapView().updateModifications(null);
     App.rootView.subHeaderRegion.empty();
     App.rootView.sidebarRegion.empty();
+
+    unsetWarnBeforeLeaving();
 }
 
 function projectErrorState() {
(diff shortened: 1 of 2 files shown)


ajrobbins:
I spoke to @jfrankl about this and we think the two limitations described in the notes are enough to merit holding off on this entire feature for this release cycle. We should close or hold this PR until the two limitations can be addressed, and should not merge to develop in this iteration.

ajrobbins:
@arottersman Yes - that would require reverting https://github.com/WikiWatershed/model-my-watershed/commit/e0a04eba68b9c254f32ae0da6100e3d87af341e1, correct? Any others?

arottersman:
We'll probably also want to revert 8e9137879b46faec4bbf4ef6629240ab73da41c2 which relied on the save button.

rajadain:
Closing this unmerged as per discussion above.

Issue #1989, opened 2017-06-28
Add timeout to third-party search HTTP requests


hectcastro:
When making HTTP requests against the following services, entire that the `requests` and `suds` contain a reasonable timeout (3-5 seconds?) so that the application web workers aren't tied up waiting for a response:

- https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/bigcz/clients/cinergi.py#L131
- https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/bigcz/clients/cuahsi.py#L25
- https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/bigcz/clients/hydroshare.py#L74

If the services can't respond within that window, then we may need to consider some alternative approaches.

<img width="1175" alt="screen shot 2017-06-28 at 14 49 18" src="https://user-images.githubusercontent.com/43639/27654773-08119d84-5c11-11e7-99b4-311595ffa903.png">

ajrobbins:
5 seconds seems reasonable to start - I'm getting results in under 3 seconds for about 1 square km areas, which is good for testing.

Issue #1851, opened 2017-05-01
Bounding box search argument constraint


kdeloach:
Some data catalogs timeout while searching for results within a very large bounding box. We should add a constraint to the bbox argument to prevent this from happening.

Issue #1960, opened 2017-06-12
Do not re-search when switching tabs


rajadain:
Currently, if returning to a search tab whose results have already been fetched, without any changes to the map, the search is re-run unnecessarily, to return in identical results.

![2017-06-12 16 29 39](https://user-images.githubusercontent.com/1430060/27053772-8aab9380-4f8c-11e7-9c6f-be13957fb342.gif)

Implement a pattern akin to `fetchResultsIfNeeded` that caches results and checks if they need to be refreshed before fetching. This should return a promise that has results when resolved. If usable cached results are present, this promise should resolve immediately.

PR #2003, opened 2017-07-06
Limit bounding box for CUAHSI searches

rajadain:
## Overview

Add bounding box search constraint to CUAHSI results. We don't add this limit to other APIs because they are much faster (CINERGI), and often times don't have results if the bounding box is too narrow (CINERGI), or simply don't take a bounding box parameter (Hydroshare).

I chose a limit of 1,500 km² after some trial and error, but we can change it in the future. This should be in the HUC-10 range, when applied to AoIs. Currently we run the visible map bounds, so if you select a HUC-10 the actual bounds will be much larger.

Connects #1851 

### Notes

At first I did the server side check, but later did the client side. Not sure if we still need the server side, but since a large search can actually bring our server down we should have some sort of check there. On the other hand, this may be accounted for in #1989 

## Testing Instructions

 * Go to [:8000/search?bigcz](http://localhost:8000/search?bigcz)
 * Ensure that you cannot search for WDC / CUAHSI results if the bounds of the map view exceed 1,500 km².
 * Try zooming in and out. Try switching tabs. Ensure behavior is consistent and sensible.

### DIFF
--- a/src/mmw/apps/bigcz/clients/cuahsi.py
+++ b/src/mmw/apps/bigcz/clients/cuahsi.py
@@ -7,12 +7,14 @@ from datetime import date
 
 from suds.client import Client
 from rest_framework.exceptions import ValidationError
-from django.contrib.gis.geos import Point
+from django.contrib.gis.geos import Point, Polygon
 
 from apps.bigcz.models import Resource, ResourceLink, ResourceList, BBox
 from apps.bigcz.utils import parse_date
 
 
+SQKM_PER_SQM = 0.000001
+MAX_AREA_SQKM = 1500
 CATALOG_NAME = 'cuahsi'
 SOAP_URL = 'http://hiscentral.cuahsi.org/webservices/hiscentral.asmx?WSDL'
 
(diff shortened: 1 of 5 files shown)


arottersman:
Taking a look now!

rajadain:
Regarding the irksome behavior described above, this could be addressed by not re-searching every time a tab is clicked, as requested in #1960. Having an explicit button for triggering the search would be better.

arottersman:
I agree, not re-searching and having an explicit search button would help. I think there'd still be some cause to disable/hide the tab — after #2006 the bbox will more static in the sense that if the AoI is too big the CUASHI tab is never worth clicking.

rajadain:
> disable/hide the tab

Ah I see what you're saying. That's great idea, will add to this.

rajadain:
> disable/hide the tab

On second thought, that's a nicety which I think we can defer till later. Adding a "disabled" state to tabs, and a conditional tooltip that informs the user why it is disabled is non-trivial, and best done in its own pointed card.

/cc @ajrobbins

Issue #1912, opened 2017-05-19
Search bar appears on Projects page


arottersman:
If you go to the projects page directly from the splash page, the search bar never gets removed:

<img width="422" alt="screen shot 2017-05-19 at 1 28 54 pm" src="https://cloud.githubusercontent.com/assets/7633670/26260434/37a64f8a-3c9b-11e7-91d2-b46e623450ac.png">

It will stay all the way to the model page
<img width="1200" alt="screen shot 2017-05-19 at 2 00 08 pm" src="https://cloud.githubusercontent.com/assets/7633670/26260514/8244be46-3c9b-11e7-862f-39bf89bebb13.png">

Behavior is correct, however, if you go to the projects page from `/draw`

PR #2005, opened 2017-07-07
Remove search bar when project page is launched

caseycesari:
## Overview

Explicitly remove the search bar when the project page is launched. Some pages remove this element when the page is navigated away from, but this technique does not work for the splash page. I believe this is because it is as the root url and does not have a route name.

Connects #1912

### Demo

mmw1

## Testing Instructions

- Load the app, and navigate directly to the projects page. Verify the search element does not appear.
- Navigate around the app, and verify that the search element is shown and hidden when appropriate.


caseycesari:
Thanks!

Issue #1947, opened 2017-06-05
Filter BiG CZ results based on AOI bounds


kdeloach:
During development, we use Leaflet map bounds to filter BiG CZ search results (for convenience). However, the search view should depend on selecting an AOI first. Search results should be filtered by the bounding box of the currently active AOI.

Ref: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/data_catalog/models.js#L61

Consider reverting https://github.com/WikiWatershed/model-my-watershed/commit/596647342256634582ffad8ddd66f6b42b1af1b3.

rajadain:
Commenting to subscribe to this thread and get email notifications.

ajrobbins:
Also, drop the points outside the AoI but within the bounding box on the front end.

rajadain:
Also, redirect landing on `/search` to `/`.

PR #2006, opened 2017-07-07
Limit Search to Area of Interest

rajadain:
## Overview

Limits "Search" functionality to the area of interest. The results will now be limited and filtered to the area of interest drawn. The Search page is now no longer directly accessible, only via home → draw → analyze → search.

Builds on top of #2003 
Connects #1947 

### Demo

Current Behavior (Search extent of Map View):

image

Step 1 Limit to Bounding Box of Area of Interest:

image

Step 2 Limit to Results Intersecting Area of Interest:

image

## Testing Instructions

 * Go to [:8000/search?bigcz](http://localhost:8000/search?bigcz). Ensure you are redirected to home
 * Select an area of interest. Proceed to Analyze and Search. Search for "water". Ensure that the search results of CINERGI are limited to those that intersect the area of interest.
 * Switch to the Hydroshare tab. Ensure that the results are same as before, since Hydroshare does not accept bounding box as a search parameter and has no geometries in the results.
 * Switch to WDC tab. Ensure that the results are those that intersect the area of interest.
 * Move the map around and play with zoom. Re-run the searches by clicking their tabs. Ensure that the results are the same every time.

### DIFF
--- a/src/mmw/js/src/data_catalog/models.js
+++ b/src/mmw/js/src/data_catalog/models.js
@@ -2,6 +2,8 @@
 
 var _ = require('underscore'),
     Backbone = require('../../shim/backbone'),
+    turfIntersect = require('turf-intersect'),
+    App = require('../app'),
     utils = require('./utils');
 
 var DESCRIPTION_MAX_LENGTH = 100;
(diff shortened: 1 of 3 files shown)


src/mmw/js/src/data_catalog/models.js
```diff
@@ -83,7 +85,13 @@ var Results = Backbone.Collection.extend({
     url: '/api/bigcz/search',
     model: Result,
     parse: function(response) {
-        return response.results;
+        var aoi = App.map.get('areaOfInterest');
+
+        // Filter results to only include those without geometries (Hydroshare)
+        // and those that intersect the area of interest (CINERGI and CUAHSI).
+        return _.filter(response.results, function(r) {
+            return r.geom === null || turfIntersect(aoi, r.geom) !== undefined;
```
rajadain: ES6 would've been nice here

```js
return response.results.filter(({ geom }) => geom === null || turfIntersect(aoi, geom) !== undefined);
```

arottersman:
Now taking a look at this one!

src/mmw/js/src/data_catalog/models.js
arottersman: I think it looks plenty nice how it is :)

Issue #2000, opened 2017-07-06
"Copy link" button on share modal doesn't copy link


ajrobbins:
To replicate:

- Create a project and add scenarios
- Open the scenario dropdown
- Click on the ellipses next to a scenario
- Click the "copy link" button on this modal:
![screen shot 2017-07-06 at 2 45 54 pm](https://user-images.githubusercontent.com/12868342/27927511-019ebd70-625a-11e7-9e3a-3d811fe97e83.png)
- Try to paste it

Observed: the correct link was not pasted.

Also observed: copy link doesn't work when sharing from the project-level dropdown in the model view, or from the My Projects page, either.

rajadain:
2000th issue! 🎉

ajrobbins:
shares a point with #1999

PR #2008, opened 2017-07-07
Fix project/scenario URL copy

caseycesari:
## Overview

ZeroClipboard has become somewhat obsolete since it relies on Flash (which has been turned off by default in Chrome). It has been replaced by Clipboard.js, which operates in a similar way.

Connects #2000 

### Notes

This implementation doesn't work in older version of Safari (as noted [here](https://github.com/zenorocha/clipboard.js#browser-support)). The suggested workaround will require reworking our workflow. In the interest of getting this in for the release, I'm going to create a follow-up issue.

## Testing Instructions

- Load the app, create a project, add a scenario, and save the project. Use the scenario menu to open the share dialog. Make the project public, and press the "Copy Link" button. Verify that the correct URL was added to the clipboard.
- Go through this process in Chrome, Firefox, Safari, and IE.

### DIFF
--- a/src/mmw/js/src/core/modals/views.js
+++ b/src/mmw/js/src/core/modals/views.js
@@ -5,7 +5,7 @@ var _ = require('underscore'),
     coreUtils = require('../utils.js'),
     HighstockChart = require('../../../shim/highstock'),
     Marionette = require('../../../shim/backbone.marionette'),
-    ZeroClipboard = require('zeroclipboard'),
+    Clipboard = require('Clipboard'),
     moment = require('moment'),
     models = require('./models'),
     modalConfirmTmpl = require('./templates/confirmModal.html'),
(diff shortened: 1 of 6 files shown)


rajadain:
@azavea-bot rebuild

src/mmw/js/src/core/modals/views.js
```diff
@@ -5,7 +5,7 @@ var _ = require('underscore'),
     coreUtils = require('../utils.js'),
     HighstockChart = require('../../../shim/highstock'),
     Marionette = require('../../../shim/backbone.marionette'),
-    ZeroClipboard = require('zeroclipboard'),
+    clipboard = require('Clipboard'),
```
rajadain: Since this is a class, later used as `new clipboard()`, consider importing as `Clipboard` to follow JavaScript class naming conventions.

rajadain:
When sharing an inactive scenario, the dialog has the URL of the currently active scenario:

src/mmw/js/src/core/modals/views.js
caseycesari: Will do.

caseycesari:
> When sharing an inactive scenario, the dialog has the URL of the currently active scenario:

Good catch. This looks like an existing issue (it's reproducible on staging). Since this card shares 1 point with another card, and I already spent a good amount of time on it, I'm going to log a new issue to fix at a later time for the issue you found.

rajadain:
Taking another look.

caseycesari:
Yeah, I can't figure it out either. I see the same thing locally as well.

Thanks!

Issue #1979, opened 2017-06-15
Replace "Model My Watershed" in Title


rajadain:
![image](https://user-images.githubusercontent.com/1430060/27197219-6dc8b49a-51db-11e7-99b7-15a4ba859ca6.png)

Need to find out what to change it to.

rajadain:
Shares points with #1978

rajadain:
Any suggestions for this @ajrobbins ?

Issue #1978, opened 2017-06-15
Remove Animals, Point Sources, and Water Quality tabs from Analyze


ajrobbins:
For BiG CZ, we'll have Land, Soil, and Climate tabs. Remove the others that are present in MMW.

rajadain:
Commenting to subscribe to this thread and get email notifications.

PR #2012, opened 2017-07-07
Remove Extra Analyses, Adjust Title

rajadain:
## Overview

BiG-CZ does not require Animals, Point Sources, or Water Quality results in Analyze. We start with a base list of two analyze tasks: land and soil, and only add the rest if the BiG-CZ flag, `data_catalog_enabled`, is false.

We also update all occurrences of `Model My Watershed` in the app with `Critical Zone Data Explorer` if the BiG-CZ flag is set.

Connects #1978 
Connects #1979 

### Demo

MMW

image

BiG-CZ

image

## Testing Instructions

 * Go to [:8000/](http://localhost:8000/). Ensure that all titles read "Model My Watershed".
 * Choose an area of interest and proceed to Analyze. Ensure that all previous analyses are run.
 * Go to [:8000/?bigcz](http://localhost:8000/?bigcz). Ensure that all titles read "Critical Zone Data Explorer"
 * Choose an area of interest and proceed to Analyze. Ensure that only land and soil analyses are run.

### DIFF
--- a/src/mmw/apps/home/views.py
+++ b/src/mmw/apps/home/views.py
@@ -119,6 +119,7 @@ def get_client_settings(request):
     # BiG-CZ mode applies when either request host contains predefined host, or
     # ?bigcz query parameter is present. This covers staging sites, etc.
     bigcz = settings.BIGCZ_HOST in request.get_host() or 'bigcz' in request.GET
+    title = 'Critical Zone Data Explorer' if bigcz else 'Model My Watershed'
     EMBED_FLAG = settings.ITSI['embed_flag']
     client_settings = {
         'client_settings': json.dumps({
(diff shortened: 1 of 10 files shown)


src/mmw/apps/home/views.py
```diff
@@ -137,8 +138,10 @@ def get_client_settings(request):
             'mapshed_max_area': settings.GWLFE_CONFIG['MaxAoIArea'],
             'data_catalog_enabled': bigcz,
             'itsi_enabled': not bigcz,
+            'title': title,
         }),
         'google_maps_api_key': settings.GOOGLE_MAPS_API_KEY,
+        'title': title,
```
rajadain: We need this outer one for Django templates like `templates/head.html`. The `client_settings` one is used in JavaScript.

Issue #1997, opened 2017-07-05
Revert explicit save


ajrobbins:
Due to the issues described in https://github.com/WikiWatershed/model-my-watershed/pull/1995#issue-239884243, revert https://github.com/WikiWatershed/model-my-watershed/commit/e0a04eba68b9c254f32ae0da6100e3d87af341e1 and https://github.com/WikiWatershed/model-my-watershed/commit/8e9137879b46faec4bbf4ef6629240ab73da41c2

PR #2016, opened 2017-07-10
Revert explicit save

arottersman:
## Overview
After our attempt at #1995 we realized warning the user when they try to leave unsaved work is a lot more involved than one would hope. We're reverting the explicit-save feature for this release. 

Connects #1997

## Testing Instructions

 * Pull, and `bundle`
 * While not logged, model an AoI, then log in — confirm a project is created and saved
 * While logged in confirm every time you make it to model a project is created/saved and any subsequent changes to it are also automatically saved


rajadain:
Taking a look now.

rajadain:
Did we need to revert back to showing the login modal on app start?

src/mmw/js/src/app.js
```diff
@@ -112,6 +112,12 @@ var App = new Marionette.Application({
         return this._mapView._leafletMap;
     },
 
+    getUserOrShowLoginIfNotItsiEmbed: function() {
+        if (!settings.get('itsi_embed')) {
+            this.getUserOrShowLogin();
```
rajadain: Did we need to revert back to showing the login modal on app start?

arottersman:
I wondered the same thing. @jfrankl thoughts?

rajadain:
Sorry I deleted my comment, trying to make it part of the code which better keeps track of resolved comments. But here it is for the record:

> Did we need to revert back to showing the login modal on app start?

rajadain:
> I think this is good enough for now.

What I mean is that the recommended changes aren't necessary to be added in this PR, which is good as it is.

arottersman:
Those are good points. I'm going to merge this as is, and we can make issues if @jfrankl would like some intermediate changes in the login/save behavior for this release.

Issue #1897, opened 2017-05-18
Style cleanup


arottersman:
- [x] Upload a file unstyled
![screen shot 2017-05-18 at 10 41 49 am](https://cloud.githubusercontent.com/assets/7633670/26208341/d3e9715e-3bb7-11e7-9a6d-49e9c8117404.png)

- [x] Long named draw tools (e.g. RWD) push icons off line
![screen shot 2017-05-18 at 10 41 40 am](https://cloud.githubusercontent.com/assets/7633670/26208342/d3ecbd6e-3bb7-11e7-8dfe-05a56b87505c.png)

![screen shot 2017-05-18 at 10 51 47 am](https://cloud.githubusercontent.com/assets/7633670/26208416/049b6ab4-3bb8-11e7-807e-5b03abe1eace.png)

- [x] Long WKAoI names don't have enough bottom padding in analyze aoi header
![screen shot 2017-05-18 at 1 07 50 pm](https://cloud.githubusercontent.com/assets/7633670/26214751/889cb8a0-3bcb-11e7-8007-c5bf168af495.png)

- [x] Firefox doesn't wrap thumb select text
<img width="421" alt="screen shot 2017-05-19 at 1 18 40 pm" src="https://cloud.githubusercontent.com/assets/7633670/26260586/bfb3338e-3c9b-11e7-80ed-8ab1ebafb247.png">

kellyi:
> Firefox doesn't wrap thumb select text

Checking into this one and it seems like it may already have been fixed.

kellyi:
Checked the thumb select text in Firefox, Chrome, and IE11 and couldn't reproduce the issue.

PR #2019, opened 2017-07-10
Pad AOI name in analyze tab

kellyi:
## Overview

This PR makes some adjustments to the Analyze tab to enable showing longer AOI names while keeping the scroll behavior intact. To fix it, we drop the 54px fixed height from the header, make some other adjustments to keep it rendering correctly, add some padding to the download csv button at the bottom to ensure it stays visible on scrolling even when the height of the header changes.

Connects #1897

### Demo

screen shot 2017-07-10 at 3 43 36 pm

### Notes

Currently the longest HUC boundary name in the db is 79 char:

```
Little North Fork South Fork Coeur d'Alene River-South Fork Coeur d'Alene River
```

The longest school district is 72 char:

```
Beaufort County School District within Beaufort Marine Corps Air Station
```

Due to some quirks with how the results column's set up the download button will start getting cut off if the title's 160-240+ char long, but since we know the longest current name is just 79, it seems like it makes sense not to adjust the CSS much otherwise

## Testing Instructions

* get branch, then `bundle`, and repeat the following steps in Chrome, FF, Safari, and IE11:
* adjust `aoiHeader.html` to use the 79 char HUC boundary name above as a hard coded name in place of `{{ place }}`, then run analyze and verify that the title no longer overlaps the tab buttons and that the soil and land tabs are fully scrollable & that the download data button's not hidden beneath a div
* model the AOI, then click the analyze tab again and verify that the land and soil tabs are also still fully scrollable

### DIFF
--- a/src/mmw/sass/components/_sidebar.scss
+++ b/src/mmw/sass/components/_sidebar.scss
@@ -1,6 +1,7 @@
 .results-content.downloadcsv-link,
 .downloadcsv-link {
     margin-top: 14px;
+    margin-bottom: 3rem;
     cursor: pointer;
     font-size: $font-size-h5;
     @extend .btn;
--- a/src/mmw/sass/components/_tabs.scss
+++ b/src/mmw/sass/components/_tabs.scss
@@ -36,7 +36,7 @@
 
 .model-stage-results-tab-pane {
     .nav-tabs {
-        background-color: transparent;
+        background-color: $paper;
     }
     .tab-content {
     // Set a specific height to allow y-scroll:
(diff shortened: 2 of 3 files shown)


arottersman:
Working perfectly so far except this:
screen shot 2017-07-10 at 5 44 47 pm

kellyi:
Rebased this on `develop` and pushed ae3cdf6 which fixes the grey block background for "Calculating results..."

kellyi:
Found another place where we need to set the background color explicitly now:

screen shot 2017-07-11 at 4 51 39 pm

d91d840 handles that

rajadain:
@azavea-bot rebuild

kellyi:
Thanks! Going to merge this one.

Issue #1967, opened 2017-06-14
Model fails if you sign in while it's running


arottersman:
To recreate:
- While not logged in, progress to the model stage
- While the model is still running, login
- The model will fail: `Failed to load /api/modeling/jobs/5378f08b-276f-419f-81f0-22ffd57783fb/: the server responded with a status of 404 (NOT FOUND)`

It seems like we're losing the celery job after we create the project.

PR #2020, opened 2017-07-10
Allow sign in while model is running

arottersman:
## Overview

We were querying for jobs with the job's `uuid` and the user, if there was one:
```py
job = get_object_or_404(Job, uuid=job_uuid, user=user)
```

When a user logged-in in the middle of task's polling, we wouldn't be able to find the job anymore because its user didn't match (the job was created without a user).

This PR queries the `Job` model by the `uuid` alone, then, if the found job has a user, confirms the user matches. This way users can't read other users' jobs, but can read anonymous jobs. Before anonymous requests could read any anonymous jobs, so it seems acceptable that users should be able to read them, too. 

Thanks @rajadain for talking this through with me!

Connects #1967 

### Demo

fus0zizq65

## Testing Instructions

 * Pull branch
 * While logged out, model an AoI, and while it's still polling log in. Confirm the task still completes without error and you're able to save the project
 * Try some bad uuid like `curl localhost:8000/api/modeling/jobs/432e2f6b-9bc8-4813-bd58-67fbe8b74b73/` and confirm the response is `Not found`

### DIFF
--- a/src/mmw/apps/core/tasks.py
+++ b/src/mmw/apps/core/tasks.py
@@ -3,9 +3,6 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import absolute_import
 
-from rest_framework.response import Response
-from django.shortcuts import get_object_or_404
-
 from django.utils.timezone import now
 from celery import shared_task
 from apps.core.models import Job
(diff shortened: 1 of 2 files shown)


arottersman:
Thanks!

Issue #2013, opened 2017-07-07
Precipitation slider jitters when toggled in Firefox and Safari


ajrobbins:
Firefox 54.0.1:

![precip-jiggle](https://user-images.githubusercontent.com/12868342/27977294-05fa13ea-6338-11e7-8b0b-411e0ef125ab.gif)

Potentially related to the shift in the Save button area above when it switches from saving to saved?

ajrobbins:
Safari 10.1.1: 
![safari_jitter](https://user-images.githubusercontent.com/12868342/28036266-ab234220-6585-11e7-81d2-d3283678d420.gif)

kellyi:
Seems like this happens because the width of the control including the precipitation number value isn't fixed.

PR #2024, opened 2017-07-10
Fix precipitation slider input width

kellyi:
## Overview

This PR smooths the precipitation slider input on FF & Safari by giving the whole control a fixed width, which prevents it from resizing dynamically when the numeric value changes.

Connects #2013 

## Testing Instructions

 * get this branch, then `bundle`
 * test the precipitation slider in FF, Chrome, Safari, and IE11 to verify that it still works and is doesn't resize on changes

### DIFF
--- a/src/mmw/js/src/core/templates/opacityControl.html
+++ b/src/mmw/js/src/core/templates/opacityControl.html
@@ -3,6 +3,6 @@
    min="0"
    max="99"
    step="3"
-   class="layerpicker-slider slider-leaflet"
+   class="layerpicker-slider slider-leaflet custom-slider"
    value="{{ sliderValue }}"
 >
--- a/src/mmw/js/src/modeling/templates/controls/precipitation.html
+++ b/src/mmw/js/src/modeling/templates/controls/precipitation.html
@@ -1,7 +1,7 @@
 <div>
     <label>
         <span class="name">Precipitation</span>
-        <input type="range" min="0" max="25" step="0.25" />
+        <input type="range" min="0" max="25" step="0.25" class="precipitation-slider" />
         <span class="value"></span>
     </label>
 </div>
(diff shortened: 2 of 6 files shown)


arottersman:
Should we make it match the wireframes while we're here?
screen shot 2017-07-10 at 4 40 16 pm

It could alleviate it looking off centered on FF and bizarre in IE11. The styling already exists as `.layerpicker-slider .slider-leaflet`

kellyi:
> Should we make it match the wireframes while we're here?

Question for @ajrobbins ^

ajrobbins:
Yep, let's do it.

kellyi:
Dropped the prior commit and pushed a new one which adds custom style to the precipitation slider:

screen shot 2017-07-11 at 12 51 37 pm

screen shot 2017-07-11 at 12 51 48 pm

I didn't extract the layer selector slider CSS out into a separate file to DRY it out; the browser-specific rules are repeated, so conceivably they could go into a separate file, but the placement rules seem different enough that I think it makes sense to be able to revise the precipitation slider without also dealing with the layer selector slider.

Testing instructions are mostly the same, but it's worth checking out the compare view too now.

kellyi:
On second thought I am going to try to DRY those slider rules a bit.

kellyi:
Ok, moved those rules in 440e969. The project wasn't using `_slider.scss`, so I just removed all of the existing rules there and put the styles for `custom-slider` in there, then adjusted the import and added the class to the precipitation & layer selector slider.

Testing should now also verify that the layer selector slider appearance & functionality's the same as before.

rajadain:
@azavea-bot rebuild

rajadain:
Taking a look now.

rajadain:
The failing check is because of issues on `develop`. This is good to merge.

kellyi:

Issue #1999, opened 2017-07-06
Delete scenario doesn't work


ajrobbins:
To replicate:

- Create a project and add scenarios.
- Open scenario dropdown
- Click on ellipses
- Select "delete"

Observed: nothing happens

ajrobbins:
Shares a point with #2000

PR #2026, opened 2017-07-11
Fix delete scenario menu item

caseycesari:
## Overview

This function was accidentally dropped during work done in #1984.

Connects #1999 

## Testing Instructions

- Create a new project with multiple scenarios.
- Select a scenario from the scenario menu, and attempt to delete it.
- Click OK and verify the scenario is deleted.


caseycesari:
Thanks!

PR #2033, opened 2017-07-12
Fix JavaScript tests on Jenkins

rajadain:
The mis-casing led to a curiously silent npm failure, which eventually prevented the test bundle from being built, causing no tests to execute on Jenkins. This should fix it.


rajadain:
I think it may have had something to do with Jenkins recreating the VMs every time. Hopefully this fixes it.

Issue #2025, opened 2017-07-11
When sharing an inactive scenario, the dialog has the URL of the currently active scenario


caseycesari:
Via @rajadain 

![2017-07-07 17 16 38](https://user-images.githubusercontent.com/1430060/27977345-439dd1be-6338-11e7-84cd-cfb14052ce6b.gif)

Issue #1982, opened 2017-06-21
Cannot Remove Modifications from MapShed


rajadain:
Create a MapShed scenario, add Veg Buffer Conservation Practice to it. Try to remove it from the sidebar.

It should be removed, but the button doesn't do anything.

![mmw-cannot-remove-modifications-mapshed](https://user-images.githubusercontent.com/1430060/27399378-8f050f10-568a-11e7-8d1a-446f2f59ab7a.gif)

rajadain:
Also, in IE11, long names such as Veg Buffer Strips (Urban) doesn't wrap as it does in Chrome, causing hover issues:

![image](https://user-images.githubusercontent.com/1430060/27399555-2515f406-568b-11e7-8eef-7ac695349469.png)

arottersman:
The delete button is broken because we changed the css class, without updating the marionette view. To fix:
```diff
diff --git a/src/mmw/js/src/modeling/views.js b/src/mmw/js/src/modeling/views.js
index 5b53b75..121b738 100644
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -676,7 +676,7 @@ var GwlfeToolbarView = ScenarioModelToolbarView.extend({

     ui: {
        thumb: '#gwlfe-modifications-bar .thumb',
-       deleteButton: '.delete-button',
+       deleteButton: '.delete-modification',
        closeButton: 'button.close'
     },
```

rajadain:
Shares points with #2025

rajadain:
made #2039 for the IE11 hover issue.

PR #2038, opened 2017-07-13
Fix scenario share links

rajadain:
## Overview

Previously the scenario share modal would show the current URL of the page. This was okay when the scenarios were tabs, and only the active scenario had a context menu. However, since now any scenario can have context menu, we needed to update the share modal to show the right link, even when the selected scenario is not currently active on the page.

Also fixes modification deletion in GWLF-E.

Connects #2025 
Connects #1982

## Testing Instructions

 * Check out this branch, run `bundle`
 * Go to a TR-55 project. Ensure there are multiple scenarios.
 * Try sharing an active scenario. Ensure the URL is correct.
 * Try sharing an inactive scenario. Ensure the URL is correct.
 * Try pasting the URL in a new tab / window. Ensure the page loads correctly.

### DIFF
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -696,7 +696,7 @@ var GwlfeToolbarView = ScenarioModelToolbarView.extend({
 
     ui: {
         thumb: '#gwlfe-modifications-bar .thumb',
-        deleteButton: '.delete-button',
+        deleteButton: '.delete-modification',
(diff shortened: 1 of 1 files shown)


rajadain:
Also added a commit that fixes modification deletion for GWLF-E projects. Trying to see if I can fix the IE11 tooltip issue.

rajadain:
I split the IE11 issue into its own #2039, so this now only targets #2025 and #1982.

Issue #2032, opened 2017-07-12
Reorder columns in csv to match table when downloaded


ajrobbins:
Currently, the csv download has all the correct tabular information but the formatting does not match the table in the UI. Reorder the csv columns to match the column order in the UI tables. This applies to:

- Every tab in Analysis
- [Multiyear] Hydrology
- [Multiyear] Water quality (both tables)

PR #2044, opened 2017-07-14
Adjust CSV downloads

caseycesari:
## Overview

Replace the custom CSV download with the built-in export functionality of bootstrap-table. CSVs will now match their corresponding table in the UI exactly.

A slight hack is required to render tables with pagination. An additional table with the full dataset is rendered beneath the paginated table. It's then used to generated the CSV and then removed
from the DOM.

Connects #2032 

### Demo

 Attached is a zip containing an export of all the tables that can be downloaded as a CSV.
[csv_exports.zip](https://github.com/WikiWatershed/model-my-watershed/files/1149424/csv_exports.zip).

Here is the animal table, as an example.

image

|                      |       | 
|----------------------|-------| 
| Animal               | Count | 
| "Chickens, Broilers" | 138   | 
| "Chickens, Layers"   | 0     | 
| "Cows, Beef"         | 0     | 
| "Cows, Dairy"        | 9     | 
| Horses               | 4     | 
| Pigs/Hogs/Swine      | 13    | 
| Sheep                | 1     | 
| Turkeys              | 26    | 

## Testing Instructions

- Create a new project with an AoI that is large enough to generate multiple pages for Water Quality and Point Source analysis tables. Something larger than Philadelphia should work.
- Select one of the models.
- Go to every table in the UI and download the data. Verify that the CSV matches the table in the UI.
- Repeat the process for the other model.


src/mmw/js/src/analyze/views.js
```diff
@@ -878,47 +880,61 @@ var AnalyzeResultView = Marionette.LayoutView.extend({
...
+            if (dataName === 'pointsource') {
+                census = new coreModels.PointSourceCensusCollection(data);
+                TableView = PointSourceTableView;
+            } else if (dataName === 'catchment_water_quality') {
+                census = new coreModels.CatchmentWaterQualityCensusCollection(data);
+                TableView = CatchmentWaterQualityTableView;
+            }
+
+            // Ensure all the data is rendered
+            census.setPageSize(9999);
```
caseycesari: Thanks. Will do.

src/mmw/bundle.sh
```diff
@@ -115,6 +116,7 @@ JS_DEPS=(backbone
          ./js/shim/shutterbug.js
          ./js/shim/highstock.js
          ./js/shim/nv.d3.js
+         ./js/shim/tableExport.min.js
```
caseycesari: Yeah, let me try to use that.

caseycesari: Turns out `tableExport.min.js` is a fork of that repo that was made into a jQuery plugin. It doesn't have a corresponding npm package.

arottersman: Oh well. At least it has plenty of company with the other 5 shims.

caseycesari:
> One q (can approve depending on answer...): the output CSVs have any number with <4 digits as numbers, and >=4 digits as strings (because of the commas). Apple Numbers can open and read them fine, and I'd imagine excel can, too, so not sure if this is an issue or not?

I'm going to try and drop the comma from the output so that numbers are always treated as numbers.

caseycesari:
> I'm going to try and drop the comma from the output so that numbers are always treated as numbers.

While the bootstrap table export [claims](https://github.com/wenzhixin/bootstrap-table/tree/develop/src/extensions/export#exportoptions) to support this, I can't get it to work after many attempts.

However, I opened the file in both Excel and OpenOffice, and both of those programs correctly treat the quotes surrounding numbers with commas as escape characters, and render the values as numbers.

caseycesari:
Thanks for the review!

src/mmw/bundle.sh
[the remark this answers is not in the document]
rajadain: I need to upgrade this shim for #2666. Wish we had included the original link. Is this it: https://github.com/hhurz/tableExport.jquery.plugin ?

Issue #2031, opened 2017-07-12
On draw error, draw tool disables


arottersman:
- Free Draw a very very big AoI to get a draw error
- Close the modal, and you won't be able to draw without reselecting the Free Draw tool
![9f2pbg8ntl](https://user-images.githubusercontent.com/7633670/28120230-68baa534-66e5-11e7-8112-23fd7afad963.gif)

PR #2046, opened 2017-07-14
On draw error, keep draw tools enabled

arottersman:
## Overview

When canceling a draw session (pressing ESC or drawing an invalid AoI) the map would end up in a bad state where the draw tool view was still open, but the tool was no longer enabled. Now when we cancel the draw, we re-enable the same draw tool, and don't clear the event listeners. 

Connects #2031 

### Demo

e7zqz4t7mf

## Testing Instructions

 * Try to draw a shape the size of the states
    * Confirm you can try to draw again after you cancel the Too Big AoI Modal
 * Start to draw a shape but before finished it, press the escape key
    * Confirm you can go right back to drawing a shape
 * Press the escape key while using the Select square km tool — confirm it does nothing
 * Confirm canceling the draw tool with the "x", drawing valid AoIs, and selecting boundaryies still work

### DIFF
--- a/src/mmw/js/src/draw/models.js
+++ b/src/mmw/js/src/draw/models.js
@@ -13,7 +13,8 @@ var ToolbarModel = Backbone.Model.extend({
         pollError: false,
         activeDrawTool: null,
         activeDrawToolItem: null,
-        openDrawTool: null
+        openDrawTool: null,
+        leafletDrawTool: null,
     },
 
     reset: function() {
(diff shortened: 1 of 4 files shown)


arottersman:
I broke a test.

arottersman:
I fixed the test.

rajadain:
Taking a look at this now.

arottersman:
Thanks for testing!

Issue #1991, opened 2017-06-28
Rank CUAHSI WDC Results By Date of Measurement


rajadain:
Since CUAHSI WDC service returns a large number of results, it would be useful to have the most relevant ones at the top. In this case, relevancy is the recency of the date of measurement.

Investigate if the results can be sorted in the CUAHSI api.

If that is not the case, then fetch all results from CUAHSI, iterating through the pages as necessary, then order the results.

Compare ordering them in Django vs front-end for better user experience.

rajadain:
Depends on #1850

Issue #1990, opened 2017-06-28
Reformat CUAHSI Results


rajadain:
WDC / CUAHSI results on the front-end need to be reformatted to look like this:

```
<location>
<Sitename>[https://waterdata.usgs.gov/nwis/uv/?site_no=14113000]. Observations on <samplemedium 1>[, samplemedium 2, samplemedium 3, …]. 
Variables: <conceptKeyword 1> - <conceptKeyword 2> - …
From [<GSERV:SourceOrg> - <ServCode>](<GSERV:http://hiscentral.cuahsi.org/pub_network.aspx?n=SourceId>) web service.
Date range for site: <series Min beginDate> to <series Max endDate>.
```

see https://github.com/WikiWatershed/model-my-watershed/issues/1931#issuecomment-311170676 for details.

rajadain:
This would involve passing through more fields from our back-end to the front-end.

rajadain:
This is the structure of data returned from `GetSeriesCatalogForBox2`:

```python
{
   ServCode = "ShaleNetworkODM"
   ServURL = "http://hydroportal.cuahsi.org/shalenetwork/webapp/cuahsi_1_1.asmx"
   location = "ShaleNetworkODM:PFBC_Schuylkill_River_3630"
   VarCode = "ShaleNetwork:PFBC_Hard3"
   VarName = "Hardness, total"
   beginDate = "1981-08-19T13:45:00Z"
   endDate = "1981-08-19T13:45:00Z"
   ValueCount = 1
   Sitename = "Schuylkill River 3630"
   latitude = 39.960833
   longitude = -75.180833
   datatype = "Sporadic"
   valuetype = "Unknown"
   samplemedium = "Surface Water"
   timeunits = "day"
   conceptKeyword = "Hardness, total"
   genCategory = "Water Quality"
   TimeSupport = "0"
}
```

rajadain:
Depends on #2036 and #2037

rajadain:
This is the structure of data returned from `GetServicesInBox2`:

```python
{
   servURL = "http://hydroportal.cuahsi.org/mwdis/cuahsi_1_1.asmx?WSDL"
   Title = "Mason Water Data Information System (MWDIS)"
   ServiceDescriptionURL = "http://hiscentral.cuahsi.org/pub_network.aspx?n=5572"
   Email = "cferrei3@gmu.edu"
   phone = None
   organization = "George Mason University"
[...]

Issue #1938, opened 2017-06-01
BiG CZ design support


kdeloach:
Sidebar body contents height should be 100%:

![image](https://user-images.githubusercontent.com/43062/27052175-d9412420-4f86-11e7-90c8-a5e80bcf9260.png)

Sidebar contents should be scrollable; We also need icons and formatting for the date & author fields:

![image](https://user-images.githubusercontent.com/43062/27052218-03909b52-4f87-11e7-9bc5-eb1da4fad094.png)

Need styling for "optional service description" area beneath WDC service tab:

![image](https://user-images.githubusercontent.com/43062/27052321-5a162bcc-4f87-11e7-92d4-e655e115246f.png)

kdeloach:
Updated this issue to include 3 known design issues we may need help with.

rajadain:
Commenting to subscribe to this thread and get email notifications.

PR #2049, opened 2017-07-17
Reformat CUAHSI Results

rajadain:
## Overview

Using the guidelines from https://github.com/WikiWatershed/model-my-watershed/issues/1931#issuecomment-306838234, add aggregate fields to the record grouping, include them in the CUAHSI resource model and serializers, and show them in the front end as requested in https://github.com/WikiWatershed/model-my-watershed/issues/1931#issuecomment-311170676. Also add rudimentary scrolling to ease testing.

Builds on top of #2048 
Connects #1990 
Connects #1991 

### Demo

image

### Notes

In some cases, the Sample Medium has a value `Unknown`. Should these be kept in the UI (as in the screenshot above) or omitted? /cc @aufdenkampe @emiliom 

I also went with semicolons for delimiting a list with commas in the values, which felt more standard to me, but if we'd rather have the hyphens I can easily change to that.

I added some rudimentary scrolling to assist in testing, but actual scrolling should be added as part of #1938.

## Testing Instructions

 * Check out the branch and `bundle`
 * Go to the search page and search for `water`
 * Inspect CUAHSI / WDC results and ensure that they match the formatting as requested in #1990 
 * Ensure that only the `NWISDV` values have links in the title

### DIFF
--- a/src/mmw/apps/bigcz/clients/cuahsi/search.py
+++ b/src/mmw/apps/bigcz/clients/cuahsi/search.py
@@ -76,7 +76,15 @@ def parse_record(record, service):
         created_at=record['begin_date'],
         updated_at=None,
         geom=geom,
-        test_field="hello world")
+        details_url=details_url,
+        sample_mediums=record['sample_mediums'],
+        concept_keywords=record['concept_keywords'],
+        service_org=service['organization'],
+        service_code=record['ServCode'],
+        service_url=service['ServiceDescriptionURL'],
+        begin_date=record['begin_date'],
+        end_date=record['end_date']
+    )
 
 
 def find_service(services, service_code):
--- a/src/mmw/apps/bigcz/clients/cuahsi/search.py
+++ b/src/mmw/apps/bigcz/clients/cuahsi/search.py
@@ -73,7 +73,7 @@ def parse_record(record, service):
         description=service['aabstract'],
         author=None,
         links=links,
-        created_at=record['beginDate'],
+        created_at=record['begin_date'],
         updated_at=None,
         geom=geom,
         test_field="hello world")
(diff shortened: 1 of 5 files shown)


rajadain:
Just added a commit that implements #1991, which was simple since we don't have to deal with pagination for CUAHSI.

rajadain:
Updated PR to point to `develop`

src/mmw/apps/bigcz/clients/cuahsi/search.py
```diff
@@ -102,25 +111,38 @@ def parse_records(series, services):
...
+    records = []
+    for location, group in groups.iteritems():
+        records.append({
+            'ServCode': group[0]['ServCode'],
+            'ServURL': group[0]['ServURL'],
+            'location': group[0]['location'],
+            'Sitename': group[0]['Sitename'],
+            'latitude': group[0]['latitude'],
+            'longitude': group[0]['longitude'],
+            'sample_mediums': ', '.join(sorted(set([r['samplemedium']
```
arottersman: If we're going to normalize the aggregate attribute names into snakey_case, we could consider normalizing all the attribute names

src/mmw/js/src/data_catalog/templates/cuahsiSearchResult.html
```diff
@@ -1,19 +1,26 @@
...
-    </div>
-{% endif %}
-{% if author %}
-    <div class="resource-author">
-        {{ author }}
-    </div>
-{% endif %}
-<div class="resource-test-field">
-    {{ test_field }}
+<div class="resource-concept_keyword">
```
arottersman: Consider removing the class names.

rajadain:
Thanks for taking a look. Squashed fixups, merging as soon as CI is green.

Issue #2039, opened 2017-07-13
Cannot click GWLF-E Modifications in IE11


rajadain:
Long names such as Veg Buffer Strips (Urban) doesn't wrap as it does in Chrome, causing hover issues:

![image](https://user-images.githubusercontent.com/1430060/27399555-2515f406-568b-11e7-8eef-7ac695349469.png)

PR #2057, opened 2017-07-19
Ensure GWLFE modification tooltip text wraps

caseycesari:
## Overview

Setting this property to normal ensures that the text in the
modification tooltip is wrapped in every browser so that it does not
extend outside of the tooltip boundary.

Connects to #2039

### Demo

image
*Screenshot from IE*

## Testing Instructions

- In every major browser (especially IE), open a MapShed project.
- Go to a scenario and add a modification.
- Once the modification has been added, hover over the thumbnail in the modification sidebar, and ensure the text is always wrapped and does not extend outside of the tooltip border.

### DIFF
--- a/src/mmw/sass/base/_header.scss
+++ b/src/mmw/sass/base/_header.scss
@@ -123,7 +123,7 @@ header {
     flex-wrap: nowrap;
 
     .tooltip-inner {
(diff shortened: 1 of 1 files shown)


caseycesari:
Thanks!

Issue #2058, opened 2017-07-19
Draw modifications doesn't select/close on mobile tap


arottersman:
[iPad Chrome and Safari] The thumb select views for modifications doesn't select the modification on tap. It treats the first tap like a hover action, and selects the option on the next tap.

PR #2059, opened 2017-07-19
Fix modeling thumbselect controls on iOS

arottersman:
## Overview

On iOS, tapping on a modification thumb would open its details, but not select the tool:

[state would look like this, but clicking the map wouldn't draw anything]
screen shot 2017-07-19 at 6 05 21 pm

To select the tool you'd have to tap the same thumb again.

Connects #2058 

## Testing Instructions

 * On desktop, ASUS touch+mouse laptop, iPad, and nexus tablet
 -- confirm you can select modifications to TR-55 and gwlf-e with a single click/tap


arottersman:
Thanks for the speedy review!

PR #2063, opened 2017-07-20
Enhance AoI upload UI

caseycesari:
## Overview

Add UI indicators to let users know when a dragged file is within the drop zone and when the file is being processed. Per the design team.

Connects to #1921

### Demo

mmw2

## Testing Instructions

- Try uploading various valid and invalid files, and verify that new UI changes appear and disappear at the appropriate times.

### DIFF
--- a/src/mmw/js/src/draw/templates/aoiUpload.html
+++ b/src/mmw/js/src/draw/templates/aoiUpload.html
@@ -14,7 +14,7 @@
             <i class="fa fa-times pull-right reset-draw-button"></i>
             <span
                   id="dropbox">
-                <div class="draw-button-title">Drag and drop a file</div>
+                <div class="draw-button-title">Drag and drop your file here to upload</div>
                 <ul class="draw-pane-description-list">
                     <li>Must be a shapefile (zip containing shp and prj files) or geojson</li>
                     <li>Only the first feature is used</li>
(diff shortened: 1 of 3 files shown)


jfrankl:
Nice! What are your thoughts on allowing the user to drop the file anywhere on the screen, similar to Github and GeoJSON.io?

caseycesari:
> Nice! What are your thoughts on allowing the user to drop the file anywhere on the screen, similar to Github and GeoJSON.io?

Implementing that was a little tougher. And it didn't seem to fit as well into the existing workflow since it's not clear to the user that they can drag and drop a file until they select the "Upload file" option. But I can take a shot at if you think that's a better option.

jfrankl:
That makes sense. Let's roll with this as is. +1

arottersman:
I'm commandeering this PR :)  (@caseycesari will be out for the rest of the sprint)

Added two commits:
- Selecting a file via the button, resulted in the "processing" mask not getting added during processing, but then getting added when the upload was complete (because of the nature of `toggleClass`. a8c6c64 addresses
- Also added a commit to fix a bug where if you selected a draw tool, it wouldn't disable when you chose to upload a shape file, and you'd end up in analyze still able to draw shapes

kellyi:
About to take a look at this!

kellyi:
This seems to be working well but I noticed one quirk:

After you try to select an invalid file, the app seems to apply a mask behind the "Error" modal. If you dismiss the modal, the mask sticks around:

screen shot 2017-07-25 at 11 24 39 am

kellyi:
(Seems to be working well otherwise, though!)

arottersman:
Pushed a fixup that better leverages the Power of jQuery ⚡️

kellyi:
Cool. Checked this out again and I don't see the same bug! Shapesfiles and geojson work well, and I get the proper error message on trying to give it a pdf.

kellyi:
Taking another look at this.

arottersman:
Thanks, @kellyi! Squashed a commit. Going to merge when the build finishes.

Issue #2066, opened 2017-07-21
Better Manage BiG-CZ Pagination


rajadain:
The current implementation is rough. Consider the following improvements:

 - [ ] Consider adapting [`Backbone.PageableCollection`](https://github.com/WikiWatershed/model-my-watershed/blob/750985e3d4e8e5f7949c04669d14bac344825da3/src/mmw/js/src/core/models.js#L572) for the results
 - [ ] Cache pages on the client side once they have been fetched, perhaps storing `n` pages, so going Next and then immediately Previous does not fire server side calls

caseycesari:
Another thing to consider is that when you progress to the next or previous page, the scroll position of the list is maintained (for example, if you were scrolled all the way to the bottom of a list, you will remain there when the new page is rendered). This isn't immediately clear, and I think it may make more sense to return the user to the top of the list when a new page is rendered.

Issue #1850, opened 2017-05-01
Paginate through data catalog search results


kdeloach:
Each data catalog has a different mechanism for paginating results. Figure out a standard approach and implement pagination for each catalog.

### CUAHSI

Doesn't support pagination.

### CINERGI

The result payload contains these fields:

```
"more": "http://cinergi.sdsc.edu:8080/geoportal/rest/find/document?searchText=diamond&after=2016-05-25&before=2016-06-30&f=pjson&start=11",
"totalResults": 142,
"startIndex": 1,
"itemsPerPage": 10,
```

### HydroShare

The result payload contains these fields:

```
"count": 297,
"next": "https://www.hydroshare.org/hsapi/resource/?format=json&full_text_search=water&page=3",
"previous": "https://www.hydroshare.org/hsapi/resource/?format=json&full_text_search=water",
```

rajadain:
Commenting to subscribe to this thread and get email notifications.

rajadain:
Using various pagination support in different back-ends, get all results for each API at once, either by supplying a greater `itemsPerPage` or by iterating through all pages for Hydroshare.

rajadain:
The new CINERGI API takes `size=n` and `from=n` query arguments instead of `itemsPerPage` and `startIndex` from before. It also takes a `sort={field_name}:{asc|desc}` query argument for sorting. Other behavior can be observed here: http://geoss.esri.com/geoportal2/

rajadain:
Hydroshare does not support sorting. To do so we'd have to page through all results in Django and return a sorted set.

PR #2067, opened 2017-07-21
Add Pagination Support

rajadain:
## Overview

Adds pagination support to CINERGI and Hydroshare results. CUAHSI does not support pagination. In this effort we:

 * Add a `page=` query parameter to the back-end that can take a page number and get a set of at most 100 results corresponding to that page. This 100 page limit is configurable, but comes from Hydroshare's unchangeable page size of 100.
 * Translate the value of `page=` parameter to whatever it needs for CINERGI and Hydroshare under the hood to their pagination scheme. CINERGI uses `size=` and `from=` parameters, whereas Hydroshare just takes a single `page=` parameter, with the page size hard set to 100.
 * Add front-end for exercising this pagination back-end

The implementation was more focused on the back-end than the front-end. There are obvious stylistic issues, as well as some software design concerns. I've made #2066 to follow up.

Connects #1850 

### Demo

```json
{
  "api_url": "http://132.249.238.169:8080/geoportal/opensearch?sort=sys_modified_dt%3Adesc&q=water&from=201&bbox=-75.2585830259156%2C39.876054698521%2C-75.159759422944%2C40.0310812816138&f=json&size=100",
  "catalog": "cinergi",
  "count": 5779,
  "next": "http://localhost:8000/api/bigcz/search?bbox=-75.2585830259156%2C39.876054698521%2C-75.159759422944%2C40.0310812816138&catalog=cinergi&page=4&query=water",
  "page": 3,
  "previous": "http://localhost:8000/api/bigcz/search?bbox=-75.2585830259156%2C39.876054698521%2C-75.159759422944%2C40.0310812816138&catalog=cinergi&page=2&query=water",
  "results": [ ... ]
}
```

### Notes

Known Issues:

 * No UI for jumping to specific page, no display of current page, or how many pages there are total #2066 
 * Switching back to tab does not preserve page because search is retriggered #1960 

## Testing Instructions

 * Go to [:8000/?bigcz](http://localhost:8000/?bigcz), draw a shape and proceed to search.
 * Search for anything, say "water". Wait for CINERGI results. Ensure that you see a large number representing the total count of results in the CINERGI tab header. Ensure that there are 100 results shown in the list.
 * Ensure you cannot click "Previous". Try clicking "Next", and wait for the results to update. Ensure the results are different. Click "Previous" and ensure that the first page of results is shown again.
 * Try the same for Hydroshare.
 * Ensure there are no pagination controls in CUAHSI, and that all results (which may be more than 100, the page size) are shown at once.

### DIFF
--- a/src/mmw/apps/bigcz/serializers.py
+++ b/src/mmw/apps/bigcz/serializers.py
@@ -3,9 +3,14 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import division
 
+from math import ceil
+
 from rest_framework.serializers import \
     Serializer, CharField, DateTimeField, IntegerField, SerializerMethodField
 from rest_framework_gis.serializers import GeometryField
+from rest_framework.utils.urls import remove_query_param, replace_query_param
+
+from django.conf import settings
 
 
 class ResourceLinkSerializer(Serializer):
@@ -29,7 +34,51 @@ class ResourceListSerializer(Serializer):
     api_url = CharField()
     results = SerializerMethodField()
     count = IntegerField()
+    page = SerializerMethodField()
+    previous = SerializerMethodField()
+    next = SerializerMethodField()
 
     def get_results(self, obj):
         serializer = self.context.get('serializer', ResourceSerializer)
         return [serializer(r).data for r in obj.results]
+
+    def get_page(self, obj):
+        if not self.context.get('is_pageable', False):
+            return None
+
+        return self.context.get('page')
+
+    def get_previous(self, obj):
+        request_uri = self.context.get('request_uri')
+        page = self.context.get('page')
+
+        if not request_uri:
+            return None
+
+        if not self.context.get('is_pageable', False):
+            return None
+
+        if page < 2:
+            return None
+
+        if page == 2:
+            return remove_query_param(request_uri, 'page')
+
+        return replace_query_param(request_uri, 'page', page - 1)
+
+    def get_next(self, obj):
+        request_uri = self.context.get('request_uri')
+        page = self.context.get('page')
+
+        if not request_uri:
+            return None
+
+        if not self.context.get('is_pageable', False):
+            return None
+
+        last_page = ceil(obj.count / settings.BIGCZ_CLIENT_PAGE_SIZE)
+
+        if page >= last_page:
+            return None
+
+        return replace_query_param(request_uri, 'page', page + 1)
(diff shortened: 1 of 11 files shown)


arottersman:
Taking a look now

src/mmw/apps/bigcz/serializers.py
```diff
@@ -3,9 +3,18 @@
...
 from rest_framework.serializers import \
     Serializer, CharField, DateTimeField, IntegerField, SerializerMethodField
 from rest_framework_gis.serializers import GeometryField
+from rest_framework.utils.urls import remove_query_param, replace_query_param
+
+from django.conf import settings
+
+# Cannot import CATALOG_NAME as CUAHSI from apps.bigcz.clients.cuahsi.search
+# because that creates circular dependencies
+CUAHSI = 'cuahsi'
```
arottersman: We could consider adding an attribute to each client like`is_pageable`, and adding it to the serializer's context. Then  the checks in `get_page()`, etc. wouldn't have to be specific on CUAHSI, but on `self.context.get('is_pageable')`

src/mmw/js/src/data_catalog/models.js
```diff
@@ -1,45 +1,70 @@
...
-                query: query,
-                bbox: bbox
-            };
+        this.set({
+            query: query,
+            bbox: utils.formatBounds(bounds),
+        });
 
-        return this.startSearch(data)
+        return this.startSearch(1)
```
arottersman: (Just a remark) This threw me off for a sec because it doesn't start with 0, though I suppose that's how pages work.

rajadain: Yeah. When accessing results in the URL, it does make more sense to ask for `page=2` for the second page than `page=1`. In this case, I think a 1-based index is more idiomatic.

src/mmw/apps/bigcz/serializers.py
rajadain: Oooh that's a good idea! Working it in now.

rajadain:
The main hold up with using `Backbone.PageableCollection` is that it was taking me too long to figure out. I had originally thought of this as a back-end only card, but then decided to add some front-end functionality to demo the back-end features, and had already burned enough time on it. I made #2066 for improving the front-end implementation.

rajadain:
Okay posted a bunch of fixups that take care of a few niggling issues. Could probably use one more look before merging.

arottersman:
Sounds good! Will take another look

Issue #1901, opened 2017-05-18
Can't draw modification before initial TR-55 run completes


arottersman:
To reproduce:
* Draw AoI
* Site Storm Model
* While still polling current conditions, draw a land cover mod in a new scenario

![screen shot 2017-05-18 at 1 08 03 pm](https://cloud.githubusercontent.com/assets/7633670/26214670/3c32b550-3bcb-11e7-81a9-3a64690d25b8.png)

From `debugcelery.sh` output
```
[2017-05-18 13:04:31,677: ERROR/MainProcess] Task apps.modeling.tasks.run_tr55[758ef0ea-3f98-4512-845a-286d4250cd67] raised unexpected: Exception(u'Missing censuses for modifications',)
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/opt/app/apps/modeling/tasks.py", line 192, in run_tr55
    raise Exception('Missing censuses for modifications')
Exception: Missing censuses for modifications
[2017-05-18 13:04:31,695: ERROR/Worker-3] Task 758ef0ea-3f98-4512-845a-286d4250cd67 run from job 7627 raised exception: Missing censuses for modifications
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/opt/app/apps/modeling/tasks.py", line 192, in run_tr55
    raise Exception('Missing censuses for modifications')
Exception: Missing censuses for modifications
```

rajadain:
See possibly related: https://rollbar.com/Azavea/MMW/items/194/

mmcfarland:
May be indicated here as well: https://github.com/WikiWatershed/model-my-watershed/issues/1675

PR #2069, opened 2017-07-24
Allow Drawing Modifications During Initial Model Run

arottersman:
## Overview
- If you tried to play with the precipitation slider or draw a modification while TR-55 was still completing its initial job, the new job and all jobs following it would fail
- If the initial job failed, all new jobs with modifications would fail

I've adjusted the `fetchResults` pipeline on the frontend to either:
- wait for the initial run to complete, before running the model with the modifications
- If there's no results and the initial run isn't underway, start `fetchResults` without any modifications, then try

Connects #1901 

Also includes https://github.com/project-icp/bee-pollinator-app/pull/116
### Demo

exqu2lyeok

## Testing Instructions

 * Confirm MapShed still works as expected
 * For TR-55
    * Draw an AOI
    * Before polling completes, move the precipitation slider
    * Create a new scenario, and before polling completes, draw a modification
    * Create a new scenario and while polling, use the Developer Network Tools to make the browser offline (Should see a generic error in the sidebar)
    * Put the browser online again, and draw a modification

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -608,6 +608,7 @@ var ScenarioModel = Backbone.Model.extend({
         job_id: null,
         poll_error: null,
         results: null, // ResultCollection
+        has_fetched_results: false,
         aoi_census: null, // JSON blob
         modification_censuses: null, // JSON blob
         allow_save: true, // Is allowed to save to the server - false in compare mode
@@ -639,10 +640,9 @@ var ScenarioModel = Backbone.Model.extend({
 
         this.on('change:project change:name', this.attemptSave, this);
         this.get('modifications').on('add remove change', this.updateModificationHash, this);
-
-        var debouncedFetchResults = _.debounce(_.bind(this.fetchResults, this), 500);
-        this.get('inputs').on('add', debouncedFetchResults);
-        this.get('modifications').on('add remove', debouncedFetchResults);
+        this.debouncedFetchResults = _.debounce(_.bind(this.fetchResults, this), 500);
+        this.get('inputs').on('add', _.bind(this.fetchResultsAfterInitial, this));
+        this.get('modifications').on('add remove', _.bind(this.fetchResultsAfterInitial, this));
 
         this.set('taskModel', App.currentProject.createTaskModel());
         this.set('results', App.currentProject.createTaskResultCollection());
@@ -808,6 +808,7 @@ var ScenarioModel = Backbone.Model.extend({
 
                 pollSuccess: function() {
                     self.setResults();
+                    self.set('has_fetched_results', true);
                 },
 
                 pollFailure: function(error) {
@@ -835,6 +836,10 @@ var ScenarioModel = Backbone.Model.extend({
             };
 
         return taskModel.start(taskHelper);
+    },
+
+    fetchResultsAfterInitial: function() {
+        this.fetchResultsIfNeeded().done(this.debouncedFetchResults);
     },
 
     updateInputModHash: function() {
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -606,6 +606,7 @@ var ScenarioModel = Backbone.Model.extend({
         modification_hash: null, // MD5 string
         active: false,
         job_id: null,
+        poll_error: null,
         results: null, // ResultCollection
         aoi_census: null, // JSON blob
         modification_censuses: null, // JSON blob
@@ -801,6 +802,7 @@ var ScenarioModel = Backbone.Model.extend({
                 postData: gisData,
 
                 onStart: function() {
+                    self.set('poll_error', null);
                     results.setPolling(true);
                 },
 
@@ -808,8 +810,9 @@ var ScenarioModel = Backbone.Model.extend({
                     self.setResults();
                 },
 
-                pollFailure: function() {
+                pollFailure: function(error) {
                     console.log('Failed to get modeling results.');
+                    self.set('poll_error', error);
                     results.setNullResults();
                 },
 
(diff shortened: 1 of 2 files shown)


rajadain:
Taking a look now.

src/mmw/js/src/modeling/views.js
```diff
@@ -1032,15 +1009,40 @@ var ResultsTabContentView = Marionette.LayoutView.extend({
...
-        }));
+            ResultView = getResultView(modelPackage, resultName),
+            tmvModel = new coreModels.TaskMessageViewModel(),
+            polling = this.model.get('polling'),
+            result = this.model.get('result'),
+            error = self.scenario.get('poll_error');
+
+        // Only show this on the initial polling. On subsequent polling, we
+        // keep showing the current results.
+        if (!result) {
```
rajadain: Hrm the multi-level nesting, while nicely efficient, is pretty hard to read. I tried some variations, but this is the best I could come up with:

```js
if (result) {
    this.resultRegion.show(new ResultView({
        model: this.model,
        areaOfInterest: this.options.areaOfInterest,
        scenario: this.scenario,
    }));
    return;
}

tmvModel.setWorking('Gathering Data');

if (error) {
    if (error.timeout) {
        tmvModel.setTimeoutError();
    } else {
        var message = error === 'NO_LAND_COVER' ?
            'Selected area of interest doesn\'t include any land cover to run the model' :
            'Error';
        tmvModel.setError(message);
    }
}

if (polling) {
    tmvModel.setWorking('Calculating Results');
}

self.resultRegion.show(new coreViews.TaskMessageView({ model: tmvModel }));
```

arottersman: Looks nicer to me! Thanks!

src/mmw/js/src/modeling/models.js
```diff
@@ -833,6 +854,27 @@ var ScenarioModel = Backbone.Model.extend({
         return taskModel.start(taskHelper);
     },
 
+    // If the model has never completed, we wont be able to run
+    // it with modifications
+    // If there's already a `fetchResult` in progress, wait for it
+    // to complete before trying to run the model with the modifications
+    // If there's none in progress, call the initial fetchResultsIfNeeded
+    // Otherwise just fetch results as usual
+    fetchResultsAfterInitial: function() {
```
arottersman: Hm, yeah that makes a lot of sense! I thought there had been an issue involving two `fetchResults` promises chaining to the `fetchResultsPromise` (one here and one from the initial call from the view), and the original `modification_censuses` error returning. I can't recreate it. It was probably wrapped up in some other silly thing I was doing earlier :)

src/mmw/js/src/modeling/views.js
```diff
@@ -1023,6 +995,11 @@ var ResultsTabContentView = Marionette.LayoutView.extend({
         resultRegion: '.result-region'
     },
 
+    modelEvents: {
+        'change:polling': 'onShow',
+        'change:result': 'onShow',
```
rajadain: I'm pretty sure these can be collapsed into one line:

```js
modelEvents: {
    'change:polling change:result': 'onShow',
}
```

src/mmw/js/src/modeling/models.js
```diff
@@ -744,8 +745,20 @@ var ScenarioModel = Backbone.Model.extend({
                 fetchGisDataPromise = App.currentProject.fetchGisDataIfNeeded();
 
             self.fetchResultsPromise = fetchGisDataPromise.then(function() {
-                var promises = fetchResults();
-                return $.when(promises.startPromise, promises.pollingPromise);
+                // Ensure the model completes without modifications before trying to
```
rajadain: What happens if the initial run times out or fails? Will the second one trigger? This is probably a `.done` vs `.then` thing.

arottersman:
Thanks, @rajadain! Pushed 3 fixups to address comments.

src/mmw/js/src/modeling/models.js
arottersman: Good question. I'd expect it'd fail. Will change this to `.done`.

arottersman: By the same logic, is the `gisData` from `fetchGisDataIfNeeded` required for mapshed? Should  `self.fetchResultsPromise = fetchGisDataPromise.then(function() { ...` also be a `.done`?

arottersman:
(Pushed a fourth fixup — `.then` --> `.done`)

src/mmw/js/src/modeling/models.js
```diff
@@ -751,7 +751,7 @@ var ScenarioModel = Backbone.Model.extend({
                     var unmodifiedFetchPromises = fetchResults(true);
                     return $.when(unmodifiedFetchPromises.startPromise,
                                   unmodifiedFetchPromises.pollingPromise)
-                            .then(function() {
+                            .done(function() {
```
rajadain: Uhh sorry I think we should keep the `.then`. `.done` would return the parent promise, whereas `.then` returns a new promise. See https://stackoverflow.com/questions/5436327/jquery-deferreds-and-promises-then-vs-done

arottersman: Oops

arottersman:
I forced push for a clean slate. We decided to drop the changes that covered the case where the initial modeling fails, but you try to add modifications after things are working again (see  #2082)

Now this just covers if you draw a modification while the initial model is running, and everything goes great.

rajadain:
Taking another look.

src/mmw/js/src/modeling/models.js
```diff
@@ -837,6 +838,13 @@ var ScenarioModel = Backbone.Model.extend({
         return taskModel.start(taskHelper);
     },
 
+    fetchResultsAfterInitial: function() {
+        var self = this;
+        this.fetchResultsIfNeeded().done(function() {
+            self.debouncedFetchResults();
```
rajadain: Is there a reason to prefer this indirection? Since this anonymous function does only one thing: execute `debouncedFetchResults()`, we could give it directly to `.done()`, as shown in the comment above.

arottersman:
Thanks for working through this with me! Condensed `fetchResultsAfterInitial` w your suggestion. Will merge after the build completes.

Issue #2080, opened 2017-07-25
Paged search results don't always fit within height of container


mmcfarland:
When paged search results are displayed, they can be pushed lower than the on-screen container, making them invisible.  The problem is made more visible when the query filter box is expanded.

This is very similar to an issue we resolved in the GeositeFramework:
https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/832
https://github.com/CoastalResilienceNetwork/regional-planning/pull/76

![screenshot from 2017-07-25 09 17 14](https://user-images.githubusercontent.com/1014341/28574053-b8459b7e-711a-11e7-8daa-feac08bd1f5a.png)

rajadain:
Thanks for making a specific card for this with additional clues on fixing it. There is a general styling card, #1938, which calls for an overall cleanup. I added rudimentary scrolling in #2049, noting this drawback.

mmcfarland:
Ah, sorry I missed that.

rajadain:
Do with #2091

rajadain:
Closing this since it has been fixed via #2087.

Issue #1933, opened 2017-06-01
Wire up date range filters


kdeloach:
The BiG CZ search API supports `to_date` and `from_date` params ([Source](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/bigcz/views.py#L31)). These need to be wired up from the frontend.

![screenshot 2017-06-01 at 10 56 39](https://cloud.githubusercontent.com/assets/43062/26685768/042c78e6-46b9-11e7-9776-be904d6cc331.png)

![screenshot 2017-06-01 at 10 56 04](https://cloud.githubusercontent.com/assets/43062/26685737/f180e70e-46b8-11e7-8a70-29b1fbd7a251.png)

rajadain:
Commenting to subscribe to this thread and get email notifications.

ajrobbins:
Implement for CUAHSI WDC to start.

rajadain:
The date filter must operate on the date measurements were taken, not of the creation or updated date.

mmcfarland:
@jfrankl It's not completely clear from the wireframe shots above what the trigger should be for executing a search when the date range is in play.  Currently, search is on `enter` from the main search box.  Doing a search on change/blur for the dates doesn't seem ideal, especially if you haven't entered main search text yet.  Entering text, searching and then adding a date filter would require the user focusing the search box again and hitting enter.  

In the absence of a search button, I think keeping it just on search enter is the most straight forward.  Thoughts?

jfrankl:
Sorry I lost track of your question, Matt. I agree, this should happen on "enter". We might also need to have a "submit" type button.

PR #2081, opened 2017-07-25
Add date range filters to the search UI

mmcfarland:
## Overview

Adds UI elements for collecting start and end date filters and passes them along to the existing API endpoint using the pre-existing querystring parameters.  Does some light validation to ensure that
`before < after`, which would otherwise raise exceptions on the backend.

Also adds a small formatting improvement to the warning dialog for areas which are too large for WDC to handle.

Connects #1933  

#### Added filter toggle
screenshot from 2017-07-25 09 03 56

#### Filters exposed with date-picker
screenshot from 2017-07-25 09 04 15

#### Validation response
screenshot from 2017-07-25 09 04 34

#### Warning - before
screenshot from 2017-07-25 09 00 59

#### Warning - after
screenshot from 2017-07-25 08 16 18

### Notes
I uncovered an issue that is visible here, but also on staging (though less pronounced), captured here: #2080.  I also made a good-faith effort to match the styling shown in screenshots, but this could use some touch-ups as part of a final style pass on the larger feature.

I had to derive a bit of UX from the screenshot listed in the issue.  My decisions on the behavior are as followed, and are open to feedback:

* 0, 1 or both dates can be filled in without a validation error.  The backend appears to support this configuration.
* The datepicker library handles making sure a date format is valid, so I only ensure that the date precedence is appropriate.  If after < before in the WDC query, it will return an exception.
* I didn't do validation on blur from the field, since that would be annoying if the user was changing the values and after changing one, had them temporarily in an invalid state.  Validation is only run on search.
* Toggling the filter keeps the date values in place, so they are there when you re-toggle.  This means they are still used, even when hidden.  This could be confusing, but in my opinion, it's not any clearer than having them disappear when you click "Hide Filters".  Or if worse if they are not used when hidden, but are still there when you un-toggle.  I'm open to "Hide Filters" just clearing them from state altogether as an alternative.
* Hitting enter in a datebox will execute the search.  At first, I didn't like that approach, but in practice, having search results displayed and going back and tweaking the date seemed like a natural workflow.  Not having to tab into the query box to enter seems appropriate in that case.  

## Testing Instructions

 * A new JS dependency was added (datepicker), so npm install and rebundle the vendor and client assets.
* Navigate to the search and execute a number of searches with and without dates, or partial dates.
* Attempt to provide bad date values to the search.
* Date range input should be reflected in the results.

### DIFF
--- a/src/mmw/js/src/data_catalog/models.js
+++ b/src/mmw/js/src/data_catalog/models.js
@@ -17,6 +17,8 @@ var Catalog = Backbone.Model.extend({
         id: '',
         name: '',
         description: '',
+        fromDate: null,
+        toDate: null,
         query: '',
         bbox: '',
         loading: false,
@@ -27,10 +29,12 @@ var Catalog = Backbone.Model.extend({
         error: '',
     },
 
-    search: function(query, bounds) {
+    search: function(query, fromDate, toDate, bounds) {
         this.set({
             query: query,
             bbox: utils.formatBounds(bounds),
+            fromDate: fromDate,
+            toDate: toDate,
         });
 
         return this.startSearch(1);
@@ -43,6 +47,8 @@ var Catalog = Backbone.Model.extend({
                 catalog: this.id,
                 query: this.get('query'),
                 bbox: this.get('bbox'),
+                from_date: this.get('fromDate'),
+                to_date: this.get('toDate'),
             };
 
         if (thisPage > 1 && thisPage <= lastPage) {
@@ -166,7 +172,11 @@ var Results = Backbone.Collection.extend({
 
 var SearchForm = Backbone.Model.extend({
     defaults: {
-        query: ''
+        fromDate: null,
+        toDate: null,
+        query: '',
+        showingFilters: false,
+        isValid: true,
     }
 });
 
--- a/src/mmw/js/src/data_catalog/templates/form.html
+++ b/src/mmw/js/src/data_catalog/templates/form.html
@@ -1,6 +1,31 @@
-<i class="data-catalog-search-icon fa fa-search"></i>
-<input
-    class="data-catalog-search-input"
-    type="text"
-    placeholder="Search data sources"
-/>
+<button class="btn btn-link date-filter-toggle small">{{filterText}}</button>
+
+{% if showingFilters %}
+    <div class="data-catalog-search-dates">
+        <h5>Data collected between:</h5>
+        <input
+            class="data-catalog-date-input from-date form-control"
+            value="{{fromDate}}"
+        />
+        <span>and</span>
+        <input
+            class="data-catalog-date-input to-date form-control"
+            value="{{toDate}}"
+        />
+    </div>
+
+    {% if not isValid %}
+    <div class="small data-catalog-validation-msg">From date must be earlier than To date</div>
+    {% endif %}
+
+{% endif %}
+
+<div>
+    <i class="data-catalog-search-icon fa fa-search"></i>
+    <input
+        class="data-catalog-search-input"
+        type="text"
+        placeholder="Search data sources"
+        value="{{query}}"
+    />
+</div>
--- a/src/mmw/js/src/data_catalog/templates/form.html
+++ b/src/mmw/js/src/data_catalog/templates/form.html
@@ -3,15 +3,29 @@
 {% if showingFilters %}
     <div class="data-catalog-search-dates">
         <h5>Data collected between:</h5>
-        <input
-            class="data-catalog-date-input from-date form-control"
-            value="{{fromDate}}"
-        />
+        <div class="data-catalog-clearable-input">
+            <input
+                class="data-catalog-date-input from-date form-control"
+                value="{{fromDate}}"
+            />
+            {% if fromDate %}
+            <a href="javascript:0;" class="data-catalog-clear-icon" title="Clear">
+                <i class=" fa fa-times"></i>
+            </a>
+            {% endif %}
+        </div>
         <span>and</span>
-        <input
-            class="data-catalog-date-input to-date form-control"
-            value="{{toDate}}"
-        />
+        <div class="data-catalog-clearable-input">
+            <input
+                class="data-catalog-date-input to-date form-control"
+                value="{{toDate}}"
+            />
+            {% if toDate %}
+            <a href="javascript:0;" title="Clear">
+                <i class="data-catalog-clear-icon fa fa-times"></i>
+            </a>
+            {% endif %}
+        </div>
     </div>
 
     {% if not isValid %}
--- a/src/mmw/js/src/data_catalog/views.js
+++ b/src/mmw/js/src/data_catalog/views.js
@@ -1,7 +1,9 @@
 "use strict";
 
-var L = require('leaflet'),
+var $ = require('jquery'),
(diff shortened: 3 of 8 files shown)


ajrobbins:
+1 for the UX decisions described in the Notes. 

>Toggling the filter keeps the date values in place, so they are there when you re-toggle. This means they are still used, even when hidden.

It sounds like a UI element like a flag or highlight of some kind would be useful to show that a filter is activated, even when the filter is hidden. /cc @jfrankl

src/mmw/js/src/data_catalog/views.js
```diff
@@ -126,20 +131,84 @@ var FormView = Marionette.ItemView.extend({
...
     events: {
-        'keyup @ui.searchInput': 'onSearchInputChanged'
+        'keyup @ui.searchInput': 'onSearchInputChanged',
+        'click @ui.filterToggle': 'onFilterToggle',
+        'change @ui.dateInput': 'onDateInputChanged',
+        'keyup @ui.dateInput': 'onDateInputKeyup'
+    },
+
+    onRender: function() {
+        $('.data-catalog-date-input').datepicker();
```
rajadain: Need to import `$` before we can use it. That's causing the linting errors:

```
$ kj npm run lint

js/src/data_catalog/views.js: line 151, col 9, '$' is not defined.
js/src/data_catalog/views.js: line 170, col 27, '$' is not defined.

2 errors
```

src/mmw/js/src/data_catalog/views.js
```diff
@@ -90,13 +91,14 @@ var DataCatalogWindow = Marionette.LayoutView.extend({
 
         // CUAHSI should not be fetched beyond a certain size
         if (catalog.get('id') === 'cuahsi' && area > MAX_AREA_SQKM) {
-            var alertView = new modalViews.AlertView({
+            var formattedArea = Math.round(area).toLocaleString(),
+                alertView = new modalViews.AlertView({
                 model: new modalModels.AlertModel({
                     alertMessage: "The bounding box of the current area of " +
-                                  "interest is " + Math.round(area) + " km², " +
+                                  "interest is " + formattedArea + "&nbspkm², " +
```
rajadain: Missing semicolon: `&nbspkm` → `&nbsp;km`

src/mmw/js/src/data_catalog/views.js
```diff
@@ -90,13 +91,14 @@ var DataCatalogWindow = Marionette.LayoutView.extend({
...
                                   "which is larger than the current maximum " +
-                                  "area of " + MAX_AREA_SQKM + " km² supported " +
-                                  "for WDC.",
+                                  "area of " + MAX_AREA_FORMATTED + "&nbspkm² " +
```
rajadain: Here too.

src/mmw/js/src/data_catalog/models.js
```diff
@@ -17,6 +17,8 @@ var Catalog = Backbone.Model.extend({
         id: '',
         name: '',
         description: '',
+        dateStart: null,
```
rajadain: Consider changing to `fromDate` and `toDate` to match the API.

src/mmw/npm-shrinkwrap.json
```diff
@@ -44,6 +44,11 @@
       "from": "https://registry.npmjs.org/bootstrap/-/bootstrap-3.3.4.tgz",
       "resolved": "https://registry.npmjs.org/bootstrap/-/bootstrap-3.3.4.tgz"
     },
+    "bootstrap-datepicker": {
```
rajadain: 👏  I like the neatness of the shrinkwrap changes! There is a typo in the commit message though `javascrip` → `javascript`.

rajadain:
A search with no dates is failing:

image

src/mmw/js/src/data_catalog/views.js
```diff
@@ -126,20 +131,84 @@ var FormView = Marionette.ItemView.extend({
...
+
+    validate: function() {
+        // Only need to validate if there are two dates.  Ensure that
+        // before is earlier than after
+        var dateFormat = "MM/DD/YYYY",
+            before = moment(this.model.get('dateStart'), dateFormat),
+            after = moment(this.model.get('dateEnd'), dateFormat),
+            isValid = false;
+
+        if (!before || !after) {
```
rajadain: Changing this to

```js
if (!before.date() || !after.date())
```

fixes the issue of eager validation:

image

mmcfarland: Looks like the validation message is behaving strangely in your screenshot, I'll revist and make sure I didn't alter something as I was cleaning up my commits.

src/mmw/js/src/data_catalog/templates/form.html
```diff
@@ -1,6 +1,35 @@
...
+<button class="btn btn-link date-filter-toggle small">{{filterText}}</button>
+
+{% if showingFilters %}
+<div class="data-catalog-search-dates">
+    <div class="form-group">
+        <label>Before</label>
+        <input
+            class="data-catalog-date-input form-control"
+            data-date-range="start"
+            value="{{dateStart}}"
```
rajadain: Before should map to `dateEnd` and After should map to `dateStart`

rajadain: The current setup is wrong:

image

since this describes an infinite date range with a gap in between. However, reading "After" come first and "Before" later could also be counterintuitive. Perhaps we should use other labels, like "From" and "To" or "Till".

mmcfarland: Yeah, I found the labels problematic to map to the API attributes.  I'll assume they were determined casually and come up with something that can be better understood.

mmcfarland:
Added a number of fixups which, in addition to some minor cleanup, alters the display of the date ranges so that it reads as "Between X and Y", to prevent confusion about which end of the spectrum a particular date can be on.  This allows a more straightforward mapping from `textbox -> variable -> api`.

You can see the filters being applied in action, here:

screenshot from 2017-07-25 13 46 08
screenshot from 2017-07-25 13 45 47
--
The validation message is updated as well:

screenshot from 2017-07-25 14 07 16

jfrankl:
A few thoughts:

I prefer when the validation message is on its own line below the inputs. In some screen captures, it is displayed to the right of the inputs.

Would be good if there were a clear button to remove the date when one has been entered. The wireframes show a "X Reset" button after a date has been entered. An "X" button inside the inputs, on the right side, would also work.

Other than that, this looks good to me, and once it's in I can do some minor style cleanup.

rajadain:
Tested the new design in a smaller window, the validation message is correctly below, not beside, the input fields:

image

mmcfarland:
@rajadain Thanks for the review.  I squashed my fixups, and added a new commit addressing @jfrankl's comment on the "clear" buttons.  These weren't in the wireframes posted in the issue, but I interpreted it as:

screenshot from 2017-07-25 16 02 10

They only show up when there is a value in the box, and clicking it will clear out the value.  Adding this in created some unwanted visual effects (the box grows and shrinks if they are present/absent, and it's no longer flush with the edge).  I attempted to restore my previous style, but think the effort would be better spent during the design pass.  Could you give it one more spin?

rajadain:
Taking another look.

PR #2087, opened 2017-07-27
Jdf/bigcz style

jfrankl:
## Overview

Style cleanup for BiG CZ application.

Connects #1938 

### Demo

image

image

### Notes

* I removed hyperlinks in the search results text. Users will eventually be clicking on those result boxes, so we should keep them simple. We can expose those links in the detailed view if they're important.
* The pagination should ideally be fixed to the bottom of the page. This is going to take some collaboration with a developer to pull off. 

## Testing Instructions

 * Go through the BiG CZ standard workflow, make sure everything looks okay.

### DIFF
--- a/src/mmw/sass/pages/_data-catalog.scss
+++ b/src/mmw/sass/pages/_data-catalog.scss
@@ -1,25 +1,48 @@
 .data-catalog-window {
     width: $sidebar-width;
     transition: all 200ms;
-    overflow-y: auto;
-    background-color: $ui-light;
+    background-color: #E6E6E6;
+    display: flex;
+    flex-direction: column;
+    position: absolute;
+    bottom: 0;
+    top: 44px;
+
+    .tab-contents-region {
+        background-color: rgba(0, 0, 0, 0.06);
+        flex: 1;
+        overflow: hidden;
+        position: relative;
+    }
+
+    .result-region {
+        padding-top: 6px;
+    }
+
+    .pager-region {
+        padding: 10px 0;
+        background-color: #e6e6e6;
+
+        .row {
+            margin-left: -5px !important;
+            margin-right: -5px !important;
+        }
+    }
 
     .data-catalog-title {
-        margin: 0.5em 0;
-        padding: 0.5em;
+        margin: 10px 0 0;
+        padding: 10px;
     }
 
     .data-catalog-form {
         position: relative;
-        padding: 0.5em;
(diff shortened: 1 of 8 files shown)


rajadain:
Taking a look now.

mmcfarland:
The date range filters don't appear in the screenshots posted - was this based on the most recent UI version?

rajadain:
Also, would it be possible to left align the overflowing lines to the first line? Like this:

image

src/mmw/sass/pages/_data-catalog.scss
```diff
@@ -46,9 +69,9 @@
         }
         .data-catalog-search-icon {
             color: $ui-grey;
-            position: relative;
```
mmcfarland: This was previously set to `relative` to account for the changing height when toggling the date filters.  Using `absolute` keeps it fixed even when the search box itself is pushed down.

jfrankl:
@rajadain I believe I've made all of the fixes you requested. Take a look:

image

src/mmw/sass/pages/_data-catalog.scss
jfrankl: I'm not sure what this means, Matt. I'm going to need some developer support to get the pagination fixed in the correct way. Previously, the pagination didn't move, which was good, but there were other issues with the sidebar layout. I fixed the other issues but I need help to get the pagination working right.

rajadain:
Taking another look.

rajadain:
We seem to have hidden the many sample method fields behind a an ellipsis:

image

Since we have a recommendation of possibly dropping the field entirely (see: https://github.com/WikiWatershed/model-my-watershed/issues/2096#issuecomment-318682012) I think this is fine for now.

src/mmw/sass/pages/_data-catalog.scss
rajadain: We have a card to fix pagination #2091. The date filter opens and closes correctly, pushing things down and up as it should. Don't think this is an issue here.

rajadain:
I'm going to merge these in before they get too far behind `develop` and have conflicts to deal with.

rajadain:
Sorry, had forgotten to assign you @jfrankl. Is this good to merge?

jfrankl:
Yes please go ahead.

rajadain:
Thanks for working on this!

PR #2088, opened 2017-07-27
Sort Pageable Analyze Tables Over Full Collection

arottersman:
## Overview

- Previously, clicking the column headers for the Point Source or Catchment Water Quality tables, would only sort the data within the visible page
- Extract out a `PageableTableBaseView` that handles pagination, and overrides the bootstrap-table's `onSort` to sort the whole collection, and rerender the table

Connects #1594 

### Demo

1rkoyw9t2h

## Testing Instructions

 * Pull + `./scripts/bundle.sh`
 * Analyze a big AoI in the DRB (eg the schuykill huc-08)
 * Confirm the Point Source and Water Quality tables
     - still paginate properly
     - sort across the whole table on column sort for each column
     - still download
  * Site storm model the AoI and confirm the Analyze toggled view works and isn't laggy
  * For a smaller AoI confirm the point source + water quality tables still render and sort okay when not paginated
  * For an AoI outside the DRB, confirm the point source + water quality no-data tables render okay

### DIFF
--- a/src/mmw/js/src/analyze/templates/catchmentWaterQualityTable.html
+++ b/src/mmw/js/src/analyze/templates/catchmentWaterQualityTable.html
@@ -1,59 +1,28 @@
-<div class="pageable-table-container">
-{% if totalPages > 1 %}
-<div class="row paging-ctl-row">
-    {% if hasPreviousPage %}
-        <button
-            class="btn-prev-page btn btn-primary col-sm-1 col-sm-offset-3"
-            data-toggle="tooltip"
-            title="Previous Page"
-            data-placement="right"
-        >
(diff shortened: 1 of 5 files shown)


mmcfarland:
Looks good, though I have one UX suggestion.  When paged into the results, I think doing any sort should reset the page back to 1.  The page you happen to be on won't necessarily contain any of the records it previously did or have much context for the new sort order.  When I was on page 5 and sorted to see the minimum discharge, it sorted correctly but left me on page 5, so I had to scroll back to actually see the minimum charge.

arottersman:
Good point — I'll add a commit to address.

arottersman:
This is ready for another look

Issue #2070, opened 2017-07-24
Compare View Revamp: Import Prototype into Backbone Views and Templates


rajadain:
Import and decompose the static prototype from https://github.com/WikiWatershed/model-my-watershed/commit/51b11731ff080299957355bf33e9d691df421612 into Backbone Views and Templates. These should be static, and not correspond to the number / titles of actual scenarios at this stage. They should look like the prototype does and show up when the use clicks the "Compare" button in the modeling page. The old Compare view will no longer be accessible, but its code and assets can be left for now, to be removed at a later stage.

Given that some pages have items that others do not (Charts or Tables, or Dropdowns that affect behavior), one possible implementation may have a different Marionette View for each type of comparison (TR-55 Runoff Chart v Table, TR-55 Water Quality Chart v Table, GWLF-E Hydrology Chart v Table, GWLF-E Water Quality Chart v Table).

PR #2094, opened 2017-07-27
Compare View Revamp: Import Prototype

rajadain:
## Overview

Imports prototype from [`jdf/compare-prototype`](https://github.com/WikiWatershed/model-my-watershed/tree/jdf/compare-prototype), and decomposes into Marionette views, models, and templates.

This is rough, and meant to unblock other efforts. We can fix issues found herein later.

I couldn't get to the chart templates. Those are preserved in [`compareWindow2.html`](https://github.com/WikiWatershed/model-my-watershed/blob/ee7d3dad760b85bc3bc5cad3920c6336491b2161/src/mmw/js/src/compare/templates/compareWindow2.html#L34-L93) template.

I also did not remove any of the existing Compare stuff, even though it is effectively inaccessible. We can do so once the Compare View Revamp is complete.

Connects #2070 

### Demo

image

## Testing Instructions

 * Check out this branch, and `bundle --debug --vendor`
 * Create / open a TR-55 project with more than once scenario.
 * Hit compare. You should see a view akin to the screenshot above.


src/mmw/js/src/compare/views.js
```diff
@@ -6,17 +6,151 @@ var _ = require('lodash'),
...
+            collection: this.model.get('tabs'),
+        }));
+        this.inputsRegion.show(new InputsView({
+            model: this.model,
+        }));
+        this.scenariosRegion.show(new ScenariosRowView({
+            collection: App.currentProject.get('scenarios'),
+        }));
+
+        this.listenTo(this.model, 'change:mode', this.showSectionsView);
```
arottersman: Does this behave differently than `modelEvents`?

src/mmw/js/src/modeling/views.js
```diff
@@ -324,6 +328,52 @@ var ScenarioButtonsView = Marionette.ItemView.extend({
...
+        var scenarios = this.projectModel.get('scenarios'),
+            runoffTable = [
+                {
+                    name: "Runoff",
+                    unit: "cm",
+                    values: scenarios.map(function(s) {
+                        return s.get('results')
+                                .findWhere({ name: "runoff" })
+                                .get('result')
+                                .runoff.modified.runoff;
```
arottersman: This is pretty brittle, and won't allow us to show the compare view when the model is polling or if one of the scenarios has an error. Want me to open an issue to deal with this, or do you want to handle it now?

src/mmw/js/src/compare/views.js
rajadain: Hmm no there isn't. At one point during development it seemed like `modelEvents` was firing too early, so I had to place `listenTo` in `onShow` which is much further down the startup timeline, but it works just as well now so I'll replace it with that.

src/mmw/js/src/modeling/views.js
rajadain: This is extremely brittle, and is meant to be replaced. I put this stuff in `modeling/views` initially since it seemed like there was enough context to direct the Compare Window, but as the amount of configuration to be supplied has grown, I think we should pass minimal things to the Compare Window and abstract away all this decision making stuff there in. I'll add a note that explains this, but wouldn't want to spend more time on this right now. Not sure which of the cards already made would be a good fit for this task. We may need to make a new one, in which case I will.

rajadain:
> We're not handling the case where there are no model results

Good point. I think we'll need some design guidelines from @jfrankl for how to show loading / absent / error scenarios in the compare view. I'll defer this till we have some feedback, but will make a card to track this.

rajadain:
> The modal positions incorrectly in IE11

I'll take a look at this now and try to fix it here.

rajadain:
I took a brief look at the IE11 issues, and it seems like there are a multitude of problems with Flexbox on IE11. I've made #2098 to take a look at it. We have used `display:flex` elsewhere in the application, so I don't think we need to reconsider the styling completely.

rajadain:
I'm going to take one last pass at changing some architectural decisions for invocation, to try and centralize most of the logic within `compare/`.

jfrankl:
@rajadain does this project use autoprefixer? If not, we'll need to manually run some kind of [autoprefixer script](https://autoprefixer.github.io/) once to handle all of that flexbox. In my experience, doing that, immediately fixes all problems with IE11.

rajadain:
@arottersman I just pushed up a couple of self-explanatory fixups, and a final commit that brings in a lot of logic leaking out of `compare/` back within it. Hopefully this captures the way I see us going forward better. One more look?

arottersman:
Taking another look now!

rajadain:
Okay going to squash and merge. Thanks for reviewing and following along!

Issue #2092, opened 2017-07-27
Handle differing AoI limits across catalogs


ajrobbins:
We implemented AoI thresholds for each catalog, but they are not the same. This leads to the case in which a user searches with a larger AoI and gets CINERGI results (which appear first), then opens the WDC tab to get this error:
![screen shot 2017-07-27 at 10 45 05 am](https://user-images.githubusercontent.com/12868342/28676570-fe637fdc-72b8-11e7-9b43-d005bf1c8774.png)

This is jarrring and requires the user to create a new AoI. We should lower all thresholds to match the lowest, or investigate ways to tune performance to make HUC 8 searches viable for WDC.

Will need consult/recommendations from @jfrankl

arottersman:
Related https://github.com/WikiWatershed/model-my-watershed/issues/2015

Issue #2015, opened 2017-07-10
Improve Handling of Oversized Areas of Interest


rajadain:
Currently, it is unclear whether or not the WDC tab will work or fail with a warning modal:

![zpgsempls4](https://user-images.githubusercontent.com/7633670/28023346-b1b8eae2-655b-11e7-9513-ff1e609865b8.gif)

Make it so that if the bounding box of the area of interest is larger than the current maximum supported area for WDC, the tab is disabled entirely, and hovering on it shows a tooltip explaining the same.

See discussion in https://github.com/WikiWatershed/model-my-watershed/pull/2003#pullrequestreview-48924776 for details.

rajadain:
As of #2092 and #2106 we are not allowing the user to get to this stage if their AoI is bigger than the smallest maximum supported size, so this card doesn't apply any more. Closing.

PR #2106, opened 2017-07-31
BiG-CZ Uniform AoI Limit

rajadain:
## Overview

Previously we allowed much larger AoI limits for CINERGI and Hydroshare APIs, since they are much faster, but had a limit in place for CUAHSI / WDC which is slower. However, it is also the richest in terms of information, and possibly the one most users are interested in. Thus, for them to move through Draw and Analyze and Search, only to be denied at the very end, can be frustrating. Hence, we change the limit on the draw section to 1,500 km².

Connects #2092 
Connects #2015 

## Testing Instructions

 * Go to [:8000/?bigcz](http://localhost:8000/?bigcz) and try to draw, select, delineate, and upload shapes under and over 1,500 km² in size. Those under should succeed, those over should fail
 * Go to [:8000/](http://localhost:8000/), and test that the limit for MMW is still 75,000 km²

### DIFF
--- a/src/mmw/apps/bigcz/clients/cuahsi/search.py
+++ b/src/mmw/apps/bigcz/clients/cuahsi/search.py
@@ -21,7 +21,6 @@ from apps.bigcz.clients.cuahsi.models import CuahsiResource
 
 
 SQKM_PER_SQM = 0.000001
-MAX_AREA_SQKM = 1500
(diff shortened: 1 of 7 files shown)


arottersman:
Taking a look

rajadain:
Thanks for taking a look!

Issue #2090, opened 2017-07-27
Hover effect when bounding box is outside of map extent


jfrankl:
Currently, when a user hovers over a search result with a bounding box outside of the map extent, we show the standard hover effect, which looks like a slightly opaque yellow box covering the map. In practice, we are finding that this is too intense and feels off-putting.

Instead, we want to add a more subtle border around the map when the bounding box is entirely outside of the map extent. This would probably be easier as a CSS border, rather than a Leaflet element.

Here is what this would look like:

![image](https://user-images.githubusercontent.com/1809908/28674140-87817c44-72b2-11e7-8833-7c9a8ff90115.png)

Here is some CSS that could help achieve this effect:

```
#map:after {
  content: "";
  border: 3px solid gold;
  width: 100%;
  height: 100%;
  position: absolute;
}
```

Issue #2089, opened 2017-07-27
Active element on map should only be displayed on hover


jfrankl:
The yellow/active version of bounding boxes or points on the map should only be displayed while the user is hovering over a search result. Moving outside of a search result should reset the map to the state before hovering.

Here is the current functionality:
![bigcz-hover](https://user-images.githubusercontent.com/1809908/28673627-04928ca2-72b1-11e7-9da7-bd7827328e3d.gif)

PR #2109, opened 2017-07-31
BiG-CZ Map Hover Improvements

rajadain:
## Overview

Improves map hover interactions on the search page in two ways:

 * When not hovering on anything, the map resets to the original state, instead of being stuck on the last hovered item
 * When hovering on an item that is larger than the map's bounds, we show a gold border around the map without obscuring the items, making for a less distracting experience

Connects #2089 
Connects #2090 

## Testing Instructions

 * Select an area of interest and search for something
 * Ensure items in the CINERGI tab that are too big to show cause the gold highlight
 * Ensure items fully contained within the map are highlighted as before
 * Ensure items partially contained within the map are highlighted as before
 * Ensure the Hydroshare tab, which has no geometries, works as before
 * Ensure the WDC tab, which has point geometries, supports the new "unhighlight on mouseout" behavior

### DIFF
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -685,14 +685,21 @@ var MapView = Marionette.ItemView.extend({
     },
 
     renderDataCatalogActiveResult: function() {
-        var geom = this.model.get('dataCatalogActiveResult');
+        var geom = this.model.get('dataCatalogActiveResult'),
+            mapBounds = this._leafletMap.getBounds();
 
         this._dataCatalogActiveLayer.clearLayers();
+        this.$el.removeClass('bigcz-highlight-map');
 
         if (geom) {
-            var layer = this.createDataCatalogShape(geom);
-            layer.setStyle(dataCatalogActiveStyle);
-            this._dataCatalogActiveLayer.addLayer(layer);
+            if ((geom.type === 'MultiPolygon' || geom.type === 'Polygon') &&
+                drawUtils.shapeBoundingBox(geom).contains(mapBounds)) {
+                this.$el.addClass('bigcz-highlight-map');
+            } else {
+                var layer = this.createDataCatalogShape(geom);
+                layer.setStyle(dataCatalogActiveStyle);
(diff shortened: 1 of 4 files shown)


ajrobbins:
Is the gold border the same shade as the border of the squares within the viewport? It looks a little brighter. Not sure if that was intentional.

rajadain:
> Is the gold border the same shade ... ?

Yes it is. [Active Element Styling](https://github.com/WikiWatershed/model-my-watershed/blob/3a1193bf6e72f457bbd1b895e27089290d979f86/src/mmw/js/src/core/views.js#L37-L45) vs [Map Element Styling](https://github.com/WikiWatershed/model-my-watershed/blob/03110bc341708c3aa68c843e7437340723ec2c42/src/mmw/sass/base/_map.scss#L33-L39) (notice the color `gold` specified in both).

Unfortunately, since GIF files are limited to 256 colors (8-bit), while our screens run 24- or 32-bit color schemes, the screen capture is lossy and the resulting colors not accurate. When viewing the app in the browser it is easier to see that the border colors are identical.

The fill color of a selected shape is a little darker than the border because it has an opacity of `0.2` or 20%, but the border of both selected shape and map outline are the same colors.

arottersman:
Taking a look now!

jfrankl:
This looks great to me. Nice work @rajadain.

rajadain:
What version of Firefox are you using? I tested the latest v54 on both Mac and Windows and it is working correctly. In fact, the demo GIF is of Firefox on Mac.

rajadain:
My GIF is a little sped up (150%), and my Firefox definitely isn't as slow as yours, but nevertheless I have noticed that the hover operations feel laggy. I'm going to make a card for this. In the interim, let's see if someone else can test on their Firefox.

kellyi:
I took a look at this branch served from @rajadain's machine and saw the outlines properly on my local FF:

screen shot 2017-08-01 at 1 45 54 pm

rajadain:
Thanks. I'm merging this. We're tracking hover performance in #2111

Issue #2014, opened 2017-07-10
IE11 CINERGI results show in all tabs


arottersman:
In IE11 the CINERGI results never clear and appear in the Hydroshare and WDC tabs
![hfijw7osvs](https://user-images.githubusercontent.com/7633670/28025976-c7915c62-6562-11e7-9a35-4201343948ba.gif)

Issue #2054, opened 2017-07-18
BiGCZ shapes don't clear when you leave search


arottersman:
- Draw an aoi that will garner WDC results (eg a philly huc-12)
- search "water" and view the WDC tab
- Click the "Critical Zone...." header link to the splash

![screen shot 2017-07-18 at 1 43 00 pm](https://user-images.githubusercontent.com/7633670/28331581-a1fd0916-6bbf-11e7-9405-cfaecfc07a23.png)

rajadain:
Possibly another symptom of the same underlying problem #2053 has.

PR #2110, opened 2017-08-01
BiGCZ: Only search if query has changed

arottersman:
## Overview

We were searching any time you made a catalog tab active. Now we only run a search if we've updated the query (or the search has never been run).

While looking through data catalog views, noticed we were setting `className` dynamically, which I've had trouble with in the past. Correcting this fixed the bug on IE11 where the tab contents would remain the same for all tabs (ie all tabs showed CINERGI results).

Connects #1960 
Connects #2014  

### Demo

vujimkyvfj

## Testing Instructions

 * Pull and `bundle`
 * `localhost:800?bigcz`
On the major browsers:
  * Draw an AoI and search for something
  * Confirm the search doesn't execute again when you switch between tabs, but does re-execute if you update the query (either date filters or query text)
  * Confirm the pagination still works

### DIFF
--- a/src/mmw/js/src/data_catalog/models.js
+++ b/src/mmw/js/src/data_catalog/models.js
@@ -5,8 +5,7 @@ var $ = require('jquery'),
     Backbone = require('../../shim/backbone'),
     turfIntersect = require('turf-intersect'),
     App = require('../app'),
-    settings = require('../core/settings'),
-    utils = require('./utils');
+    settings = require('../core/settings');
 
 var REQUEST_TIMED_OUT_CODE = 408;
 var DESCRIPTION_MAX_LENGTH = 100;
@@ -29,10 +28,34 @@ var Catalog = Backbone.Model.extend({
         error: '',
     },
 
-    search: function(query, fromDate, toDate, bounds) {
+    searchIfNeeded: function(query, fromDate, toDate, bbox) {
+        var self = this,
+            isSameSearch = query === this.get('query') &&
+                           fromDate === this.get('fromDate') &&
+                           toDate === this.get('toDate') &&
+                           bbox === this.get('bbox');
+
+        if (!isSameSearch) {
+            this.cancelSearch();
+            this.searchPromise = this.search(query, fromDate, toDate, bbox)
+                                     .always(function() {
+                                        delete self.searchPromise;
+                                     });
+        }
+
+        return this.searchPromise || $.when();
+    },
+
+    cancelSearch: function() {
+        if (this.searchPromise) {
+            this.searchPromise.abort();
+        }
+    },
+
+    search: function(query, fromDate, toDate, bbox) {
         this.set({
             query: query,
-            bbox: utils.formatBounds(bounds),
+            bbox: bbox,
             fromDate: fromDate,
             toDate: toDate,
         });
@@ -74,7 +97,12 @@ var Catalog = Backbone.Model.extend({
         });
     },
 
-    failSearch: function(response) {
+    failSearch: function(response, textStatus) {
+        if (textStatus === "abort") {
+            // Do nothing if the search failed because it
+            // was purposefully cancelled
+            return;
+        }
         if (response.status === REQUEST_TIMED_OUT_CODE){
             this.set('error', "Searching took too long. " +
                               "Consider trying a smaller area of interest " +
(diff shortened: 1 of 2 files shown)


rajadain:
> immediately repeated promise

I think you mean "immediately resolved promise"

src/mmw/js/src/data_catalog/models.js
```diff
@@ -29,6 +29,23 @@ var Catalog = Backbone.Model.extend({
         error: '',
     },
 
+    searchIfNeeded: function(query, fromDate, toDate, bounds) {
+        var self = this,
+            isSameSearch = query === this.get('query') &&
+                           fromDate === this.get('fromDate') &&
+                           toDate === this.get('toDate') &&
+                           utils.formatBounds(bounds) === this.get('bbox');
```
rajadain: Consider lifting `utils.formatBounds(bounds)` out of this and `search` below so it is called less frequently.

src/mmw/js/src/data_catalog/models.js
```diff
@@ -29,6 +29,23 @@ var Catalog = Backbone.Model.extend({
...
+
+        if (!isSameSearch) {
```
rajadain: Consider the case when a search is already running. If a user were to change the query and re-search, `!isSameSearch` evaluates to `true` and we return `this.searchPromise`. However, when the first search returns, the `always` block executes and deletes `self.searchPromise`. What happens if this were to be deleted before the client can use its value?

rajadain: Ideally we would cancel `this.searchPromise` before starting a new one. Not sure how that can be done. Another alternative may be to chain the new search onto the old search.

If I'm misreading the code (which I could: promises are hard to follow), please do correct me.

rajadain:
Should this connect #2014 instead of #2054? Seems to be making #2054 worse

src/mmw/js/src/data_catalog/models.js
arottersman: I will work out a way to explicitly cancel the ongoing promise for clarity.

I don't think, however, that `delete self.searchPromise` prevents the new promise from resolving correctly (and we don't use `this.searchPromise` anywhere else). 

This is not what I'd expect the behavior to be, but here's a little test I made: 
```js
var self = this;
var d1 = $.Deferred();

this.myPromise = $.when(d1)
       .done(function (v) {
           console.log(v);
       })
       .always(function() {
	   console.log("deleting in first");
           delete self.myPromise;
       });

var d2 = $.Deferred();

this.myPromise = $.when(d2)
        .done(function (v) {
            console.log(v)
        })
        .always(function() {
	   console.log("deleting in second");
           delete self.myPromise;
        });

d1.resolve( "first done" );
d2.resolve( "second done" );

/** outputs
VM10771:6 first done
VM10771:9 deleting in first
VM10771:17 second done
VM10771:20 deleting in second
**/
```

arottersman:
> Should this connect #2014 instead of #2054? Seems to be making #2054 worse

Yes, that must have been a typo. Thanks! (Although maybe #2054 will be fixed when this done?)

arottersman:
This should be ready for another look! That double fixup corrects a mistake I made while rebasing.

rajadain:
Taking another look

arottersman:
Nice find — I've added a commit.

src/mmw/js/src/data_catalog/models.js
```diff
@@ -29,10 +28,30 @@ var Catalog = Backbone.Model.extend({
...
+    searchIfNeeded: function(query, fromDate, toDate, bbox) {
+        var self = this,
+            isSameSearch = query === this.get('query') &&
+                           fromDate === this.get('fromDate') &&
+                           toDate === this.get('toDate') &&
+                           bbox === this.get('bbox');
+
+        if (!isSameSearch) {
+            if (this.searchPromise) {
+                this.searchPromise.reject({ superceded: true });
```
arottersman: Actually, this isn't going to work. `this.searchPromise` isn't a deferred object. It's a `jqXHR` because that's what backbone returns with `.fetch`. I think I'm going to change this to `abort`. Another option is to wrap the `fetch` in an deferred , but I think that makes it a little more difficult to follow.

arottersman:
Added one more fixup to switch from `.reject()` to `.abort()` when cancelling the promise. Should be ready for another look now.

arottersman:
Thanks for reviewing! Squashed fixups (and fixed typo in that commit message). Will merge once the build completes.

Issue #2091, opened 2017-07-27
Move pagination to bottom of results list


ajrobbins:
Currently, the pagination buttons are above the results list. Jeff recommends moving them to the bottom.

![screen shot 2017-07-27 at 10 45 36 am](https://user-images.githubusercontent.com/12868342/28676522-d16be28a-72b8-11e7-9685-59f91f2a6ff7.png)

PR #2112, opened 2017-08-01
Affix Pagination Controls to the Bottom

rajadain:
## Overview

Previously the pagination controls would be all the way down below the search results, and the user would have to scroll to the very bottom to discover them. Now they are positioned to be fixed.

Connects #2091 

## Testing Instructions

 * Go to [:8000/bigcz](http://localhost:8000/?bigcz)
 * Select an area of interest and proceed to Analyze and then Search
 * Search for "water". Wait for results
 * Ensure the pagination buttons are fixed to the bottom of the screen
 * Ensure you can scroll to see all results and they are never obscured
 * Ensure you can turn on and off the Filters at the top and the search results are still accessible
 * Ensure WDC tab does not have the controls, and CINERGI and Hydroshare both have them, and the styling is correct in each
 * Try out various browsers

### DIFF
--- a/src/mmw/js/src/data_catalog/views.js
+++ b/src/mmw/js/src/data_catalog/views.js
@@ -274,9 +274,7 @@ var TabContentView = Marionette.LayoutView.extend({
             model: this.model,
         }));
 
-        // CUAHSI does not support pagination, but others do
-        if (this.model.id !== 'cuahsi') {
-            this.resultRegion.$el.addClass('paginated');
+        if (this.model.get('is_pageable')) {
             this.pagerRegion.show(new PagerView({
                 model: this.model,
             }));
--- a/src/mmw/js/src/data_catalog/views.js
+++ b/src/mmw/js/src/data_catalog/views.js
@@ -276,6 +276,7 @@ var TabContentView = Marionette.LayoutView.extend({
 
         // CUAHSI does not support pagination, but others do
         if (this.model.id !== 'cuahsi') {
+            this.resultRegion.$el.addClass('paginated');
             this.pagerRegion.show(new PagerView({
                 model: this.model,
             }));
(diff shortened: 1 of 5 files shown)


arottersman:
Taking a look

arottersman:
Was it intentional to have the pagination buttons show when there's only a single page? This isn't the behavior on staging:

screen shot 2017-08-01 at 4 52 52 pm

rajadain:
@caseycesari For the CINERGI tab results persisting in other tabs in IE11, we have #2014, also in this sprint

rajadain:
@arottersman Staging does show the pagination controls at the bottom if you scroll all the way down:

image

Currently, the pagination controls are instantiated by tab type, so CINERGI and Hydroshare always have them, and WDC never has it. If we want to show or hide them based on the page size, we'll have to move them elsewhere. This should probably be offloaded to another card.

caseycesari:
> @caseycesari For the CINERGI tab results persisting in other tabs in IE11, we have #2014, also in this sprint

Got it. So I think the issue with this PR is that the pager buttons are showing on the WDC tab?

src/mmw/js/src/data_catalog/views.js
```diff
@@ -273,6 +273,7 @@ var TabContentView = Marionette.LayoutView.extend({
 
         // CUAHSI does not support pagination, but others do
         if (this.model.id !== 'cuahsi') {
+            this.resultRegion.$el.addClass('paginated');
```
arottersman: Consider adding this class conditionally in the template.

This also might be a good case for having an `is_pageable` attribute here on the frontend's model like we have on the backend.

rajadain:
Pushed an update that moves to using standard model fields rather than one-off checks. I'll wait until #2110 is in so we can test the bug @caseycesari mentioned above.

rajadain:
Rebased on `develop`. This is ready for another look.

rajadain:
Thanks for taking a look! Merging now.

Issue #2108, opened 2017-07-31
Implement Hydroshare Bounding Box Filter


rajadain:
Hydroshare needs `north`, `east`, `west`, `south` parameters, and additionally a `coverage_type=box` query parameter, to work. This currently fails in production, but development can be done against https://playground.hydroshare.org/ (which has a snapshot of production data). The failing bug has been identified and fixed, and will deployed to production in mid-August.

For details, see https://github.com/hydroshare/hydroshare/issues/2257

rajadain:
Production deployment part of release [Hydroshare 1.12.0](https://github.com/hydroshare/hydroshare/milestone/47)

rajadain:
We can probably just add the `coverage_type=box` parameter here: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/bigcz/clients/hydroshare/search.py#L63

Also, since it is a one-line change, consider removing the erroneously added date sort on CINERGI from here: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/bigcz/clients/cinergi/search.py#L121

caseycesari:
Yeah, I was just looking into this. It seems we already send the coordinates.

caseycesari:
> Also, since it is a one-line change, consider removing the erroneously added date sort on CINERGI from here: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/bigcz/clients/cinergi/search.py#L121

This is the change we made in the meeting this morning?

rajadain:
Regarding:

> This currently fails in production, but development can be done against https://playground.hydroshare.org/ (which has a snapshot of production data).

I'm not sure if we should change our local endpoint to `playground` or keep it pointing at production and hope they do deploy v1.12 that has the fix mid-August. I'm leaning towards the latter.

caseycesari:
Yeah, I was going to include testing instructions to update the URL, but leave it as is in the PR.

PR #2116, opened 2017-08-02
BiGCZ: Implement minor search improvements

caseycesari:
## Overview

4cfc5f6 allows Hydroshare searches to use the AoI bounding box for searching.
4db0de7 turns off date sorting in Cinergi, which keeps the results in order of relevancy. 

Connects #2108 

## Testing Instructions

- Apply the following change to the code (the bounding box functionality is not supported in production hydroshare).

```diff
diff --git a/src/mmw/apps/bigcz/clients/hydroshare/search.py b/src/mmw/apps/bigcz/clients/hydroshare/search.py
index 1f7389c..ab37cbc 100644
--- a/src/mmw/apps/bigcz/clients/hydroshare/search.py
+++ b/src/mmw/apps/bigcz/clients/hydroshare/search.py
@@ -14,7 +14,7 @@ from apps.bigcz.utils import RequestTimedOutError


 CATALOG_NAME = 'hydroshare'
-CATALOG_URL = 'https://www.hydroshare.org/hsapi/resource/'
+CATALOG_URL = 'https://playground.hydroshare.org/hsapi/resource/'


 def parse_date(value):
```

- Draw an AoI
- Make the same search locally and on staging.
- For Hydroshare, verify that there are less results locally (since they are restricted to the bounding box). You can also take a look at the data returned from our API and verify that the `coverage_type` parameter is included in the request URL.
- For Cinergi, verify that the results on the first page contain the search term "water" locally, and that they don't on staging.

### DIFF
--- a/src/mmw/apps/bigcz/clients/cinergi/search.py
+++ b/src/mmw/apps/bigcz/clients/cinergi/search.py
@@ -118,7 +118,6 @@ def search(**kwargs):
 
     params = {
         'f': 'json',
-        'sort': 'sys_modified_dt:desc',
         'size': PAGE_SIZE,
     }
 
--- a/src/mmw/apps/bigcz/clients/hydroshare/search.py
+++ b/src/mmw/apps/bigcz/clients/hydroshare/search.py
@@ -74,6 +74,9 @@ def search(**kwargs):
         })
     if bbox:
         params.update(prepare_bbox(bbox))
+        params.update({
+            'coverage_type': 'box'
+        })
     if page:
         params.update({
             'page': page


arottersman:
This is working, but it's difficult to get meaningful results with the current `1500km^2` AoI limit. The search`water` returns the same 8 results for everywhere I tried. I think those 8 results are records have geometries that envelope the whole country, and maybe some have no geometry at all. 

I made the `max_area` really big, and was able to get 14 results for `water` drawing the whole state of Mississippi. Maybe we should be creating huge buffer around the AoI for the hydroshare search?

----

Another thing maybe worth mentioning: I spent some time trying to find one of the top entries on staging: https://www.hydroshare.org/resource/126bb28f05224636a8bc8b3d1bdad6b5/

By its title, it seems like it's associated with the Mississippi region, but the hydroshare pages themselves don't give any indication of a bbox for it. It didn't show up when I searched the entire state of Mississippi...it could be it's bound to a very specific area I haven't found yet, or has no geometry and will never show up while the `bbox` param is present.

caseycesari:
We discussed this and decided to proceed as is, and wait for Hydroshare to be updated before we make more changes. Right @arottersman?

ajrobbins:
Working off casey's branch on Alice's machine, I was able to get an expected result in the Mendon, UT area for the "Water Temperature in the Little Bear River at Mendon Road near Mendon, UT" resource:

BiG CZ portal:
screen shot 2017-08-07 at 1 11 14 pm
Note that the AoI is drawn around Little Bear River, and the Mendon resource is the second result in the list.

Hydroshare UI shows this resource at the little bear river location:
screen shot 2017-08-07 at 1 11 30 pm

Issue #2095, opened 2017-07-27
Allow Navigation between Search and Analyze


rajadain:
Either a back button, or a switcher like MMW Model page has. Any ideas @jfrankl?

jfrankl:
According to the wireframes, this should look like the MMW Model page:

https://marvelapp.com/5a66caa/

caseycesari:
![image](https://user-images.githubusercontent.com/1042475/28921567-712fe6ca-7824-11e7-838e-04ab4d49634d.png)

PR #2119, opened 2017-08-04
Enable toggling between analyze and BigCZ results

caseycesari:
## Overview

Allow the user to switch between the AoI analyze results, and the BigCZ search page. This implementation is based on the analyze and modeling tabs of MMW. Some views, templates, and CSS rules from that implementation were duplicated and adjusted rather than abstracting the MMW
implementation and making it reusable.

Connects #2095

### Demo

bigcz1

### Notes

As noted in the PR description and commit message, it proved to be more difficult to directly reuse the MMW components to implement this. Instead, I copied and pared down what I needed. In the short term, I think this was a easier, and less complicated approached. However, it may have drawbacks if we need to make similar changes to both apps.

Also, there were a number of CSS quirks that came up when I implemented this. I was able to fix all of them except for the height of the tab content. Currently, it is set optimally for the search results and not the analyze tab. The download button on the analyze tab is currently not completely viewable. I'll create a follow-up issue for this.

## Testing Instructions

- Create or select an AoI
- Proceed to the search page.
- Verify that you can toggle between the analyze and data page using the tab buttons.
- Verify that in MMW, the analyze and model pages have not been impacted by this change.

### DIFF
--- a/src/mmw/sass/components/_tabs.scss
+++ b/src/mmw/sass/components/_tabs.scss
@@ -57,6 +57,13 @@
     }
 }
 
+.data-catalog-stage-results-tab-pane {
+    .tab-content {
+      // Set a specific height to allow y-scroll:
+      height: 100% !important;
+    }
+}
+
 //Model Output Tabs
 .output-tabs-wrapper {
   .nav-tabs {
--- a/src/mmw/js/src/data_catalog/controllers.js
+++ b/src/mmw/js/src/data_catalog/controllers.js
@@ -44,11 +44,14 @@ var DataCatalogController = {
             })
         ]);
 
-        var view = new views.DataCatalogWindow({
-            model: form,
-            collection: catalogs
-        });
-        App.rootView.sidebarRegion.show(view);
+        var resultsWindow = new views.ResultsWindow({
+                model: form,
+                collection: catalogs
+            }),
+            header = new views.HeaderView();
+
+        App.rootView.subHeaderRegion.show(header);
+        App.rootView.sidebarRegion.show(resultsWindow);
     },
 
     dataCatalogCleanUp: function() {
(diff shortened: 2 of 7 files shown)


rajadain:
Taking a look now.

rajadain:
Cannot scroll all the way to the bottom of the Analyze / Land tab:

rajadain:
This branch has a dark grey buffer at the bottom of the Search page:

image

which does not exist on `develop`:

image

rajadain:
And opening the filters makes it worse:

image

rajadain:
This may be more of a design task (/cc @jfrankl), but really long area of interest names don't fit in the header:



In the very least, hovering on it should reveal the full name in a tooltip. We should also remove the clickability, since they don't have a dropdown menu as MMW does.

Also, we have now lost the area of interest size. But that problem exists in MMW too.

src/mmw/sass/components/_tabs.scss
```diff
@@ -57,6 +57,13 @@
     }
 }
 
+.data-catalog-stage-results-tab-pane {
+    .tab-content {
+      // Set a specific height to allow y-scroll:
+      height: calc(100% - 53px) !important;
```
rajadain: This is likely causing the grey box issues, since it may double the work here: https://github.com/WikiWatershed/model-my-watershed/blob/88a933ae2126630bf303cd83c5ecf3bd84857852/src/mmw/sass/pages/_data-catalog.scss#L25

arottersman:
Also going to wait for grey box issue resolution to test. Looked over the code — LGTM.

caseycesari:
Added 72cded2 to fix the gray box issue.

jfrankl:
I am going to spend some time on the design aspects of this later today. The removal of the project saving functionality will require some adjustments.

rajadain:
Taking another look.

Issue #2076, opened 2017-07-24
Compare View Revamp: Add TR-55 Water Quality Charts and Tables


rajadain:
Once the basic data is present as of #2071, add charts to the chart view. Each row should be a single chart, and the width of the bars in the chart must correspond to each scenario column.

There should be:

 - [ ] 3 charts in TR-55 Water Quality: Suspended Solids, Nitrogen, Phosphorus
 - [ ] 3 tables in TR-55 Water Quality: Suspended Solids, Nitrogen, Phosphorus

Hovering over any bar in the chart should highlight the value of all of them, as shown in the mockup above.

rajadain:
Depends on #2072

Issue #2074, opened 2017-07-24
Compare View Revamp: Wire Up Precipitation Slider


rajadain:
Wire up the new precipitation slider to trigger a re-run of the model for all scenarios, as shown here: https://marvelapp.com/8hi9jg9/screen/30676297

Issue #2073, opened 2017-07-24
Compare View Revamp: Add GWLF-E Hydrology Charts and Tables


rajadain:
Consult mockups https://marvelapp.com/8hi9jg9/screen/30304099 for details

 - [ ] Create a set of `n` colors to be used for each scenario's line in the charts, and add a little dot next to the Scenario name in this view https://marvelapp.com/8hi9jg9/screen/30304103
 - [ ] 6 charts in GWLF-E Hydrology: Precipitation, ET, Surface Runoff, Subsurface Flow, Point Source Flow, Stream Flow. Precipitation should use a dashed line, the rest should use the assigned color

kellyi:
Superseded by #2909 and #2915

PR #2121, opened 2017-08-04
Compare View Revamp: Wire Up Precipitation Slider

rajadain:
## Overview

Adds precipitation slider to the new Compare View. The slider is initialized to the precipitation value of the active scenario. Moving the slider updates all scenarios in comparison, running the model for the given value.

Also duplicates scenarios before giving them to the Compare View, rather than giving the original set of scenarios. This is done so that we can change the precipitation value in the Compare View and run a number of models, without affecting the underlying original scenarios. When the Compare View is closed and the user returns to the Modeling View, all scenarios retain their original precipitation values.

The synchronization code used to keep the many precipitation sliders in the old Compare View has been removed.

I would recommend code review to be done commit by commit, since they all touch the same files.

Connects #2074 

### Notes

The numbers in the displayed table don't update with the scenarios. These are tasks for #2073 and #2076.

## Testing Instructions

 * Go to either a saved TR-55 project or a new one, make sure it has more than one scenario, and hit "Compare"
 * Ensure that you see the new Compare Window, and it has a precipitation slider
 * Ensure the value of the precipitation slider is the same as that of the active scenario
 * Open the network tab, and then move the precipitation slider. Ensure that whenever there is a new value, a TR-55 job is fired for every scenario in the comparison
 * Close the Compare Window. Ensure that the precipitation values of all scenarios are the same as they were before the comparison.

### DIFF
--- a/src/mmw/js/src/compare/views.js
+++ b/src/mmw/js/src/compare/views.js
@@ -129,19 +129,19 @@ var InputsView = Marionette.LayoutView.extend({
     },
 
     onShow: function() {
-        var setPrecipitationValue = _.bind(this.setPrecipitationValue, this),
+        var addOrReplaceInput = _.bind(this.model.addOrReplaceInput, this.model),
+            controlModel = this.model.get('scenarios')
+                               .findWhere({ active: true })
+                               .get('inputs')
+                               .findWhere({ name: 'precipitation' }),
             precipitationModel = this.model.get('controls')
                                      .findWhere({ name: 'precipitation' });
 
         this.precipitationRegion.show(new PrecipitationView({
             model: precipitationModel,
-            addOrReplaceInput: setPrecipitationValue,
+            controlModel: controlModel,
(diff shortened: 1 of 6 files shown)


src/mmw/js/src/compare/views.js
```diff
@@ -502,15 +520,81 @@ function getGwlfeTabs(scenarios) {
...
+        inputs: new modelingModels.ModificationsCollection(scenario.get('inputs').toJSON()),
+        inputmod_hash: scenario.get('inputmod_hash'),
+        allow_save: false,
+        active: scenario.get('active'),
+    });
+
+    newScenario.get('inputs')
+               .on('add', _.debounce(_.bind(newScenario.fetchResults,
+                                            newScenario),
+                                     500));
```
rajadain: Does 23686ba look better?

rajadain:
Thanks for taking a look!

Issue #2072, opened 2017-07-24
Compare View Revamp: Add TR-55 Runoff Charts and Table


rajadain:
Once the basic data is present as of #2071, add charts to the chart view. Each row should be a single chart, and the width of the bars in the chart must correspond to each scenario column. Consult https://marvelapp.com/8hi9jg9/screen/29125399 for mockups, and talk to @jfrankl for figuring out which element's width should the bar in the chart match.

There should be:

 - [x] 4 charts in TR-55 Runoff: Combined Hydrology, Runoff, Evapotranspiration, Infiltration
 - [x] 3 tables in TR-55 Runoff: Runoff, Evapotranspiration, Infiltration

Hovering over any bar in the chart should highlight the value of all of them, as shown in the mockup above.

rajadain:
In #2094, some demo code for assembling TR-55 Runoff Table data was added to `modeling/views`. This should probably be brought in to a specialized function / view under the `compare/` set.

Issue #1848, opened 2017-05-01
Improve styling for "model running"


jfrankl:
Currently, when a model is running, we display small spinners underneath the topic tabs in the sidebar. I think this change is too subtle for the user to effectively see when the model is finished. On top of that, we don't make any changes to the model output while the model is running, which makes it seem like this content is current, when it does not actually match the map (while the model is running).

Below are a few ideas for improvement, though I think we should explore this further before we decide on an option.

- Change opacity of the model output to make it more clear that the model is running
- Add a spinner over the outdated model output

#### Current
![image](https://cloud.githubusercontent.com/assets/1809908/25595529/042e8c00-2e93-11e7-8dc1-f2ed683a0f1e.png)

jfrankl:
#### Model re-running, refreshing content

In Model mode, when the model is re-run, any data in a chart (bars, lines, etc.) and data in tables should be restyled to a "not ready" state, which will be a decreased opacity and css desaturation filter. That same style will be applied to tabs while those sections of content are processing. When a section of content is ready, that tab should snap back to its regular state and color should be restored to the chart and table content. A single spinner should appear on the tab level, which should be visible until all sections of content are finished processing—not just the one currently being viewed. 

![mmw1](https://cloud.githubusercontent.com/assets/1809908/26800107/06f66aa0-4a06-11e7-8c9a-80305955e99b.gif)

#### First time loading
When the Analyze or Model mode loads for the first time, before a graph or table has been displayed, we should display a special water loader along with the messages we currently display ("Calculating results", etc.). Tabs should be visible but "not ready", as with the model re-running above. The tab-level spinner should spin until all tabs are ready.

![mmw2](https://cloud.githubusercontent.com/assets/1809908/26800103/03ff1e46-4a06-11e7-991e-16adda3b5971.gif)

jfrankl:
Also, please let me know when you start working on this. I have a lot of the CSS written already.

PR #2123, opened 2017-08-07
Compare View: TR-55 Runoff Charts

kellyi:
## Overview

This PR adds the TR-55 Compare View charts and sets up the table & chart values to adjust accordingly. gif demo's below.

Connects #2072 

### Demo

compare-charts

### Notes

Currently there are a few quirks & limitations with this implementation, which we can either handle here or spin off elsewhere. 

1. I did not implement the popovers, which seem to be complicated enough to handle in another card.  NVD3 has tooltips by default, but since the desired behavior's that the hovering over one value on the bar chart opens corresponding popovers for all three, which isn't a built-in NVD3 feature. I think we may end up wanting to use Bootstrap popovers for this.

2. The bars themselves match the width of the scenario title placeholders when there are few enough of them to display in the scenario header without overflow. However, when there are more, the chart div seems not to expand in size beyond what's visible so the bars get squished.

3. There's currently some flicker as new charts are shown when retrieving any of the results, which means it renders once per scenario if the precipitation slider value changes. Not sure what is the best way to handle this -- I think we may want to set some value to check whether any of the scenarios is currently polling and show a loading state if not?

4. I think the maximum y axis value for "Combined Hydrology" should be the value the precipitation slider has been set to, but it seemed like the value these scenarios had was slightly off the number displayed for the slider. Not sure what is the best way to handle this, but we'll need to pass it in to the chart to set the `yDomain`.

Happy to try to address these here or to spin them off elsewhere; figured I'd put up the PR for review rather than spinning wheels on it.

## Testing
- get this branch, then ~`update` and `server`~ `bundle.sh` & hard refresh in the browser
- create a new TR-55 project then create a few scenarios including some which would cause varied values for runoff/infiltration (think: veg basin gswi conservation practice or high density land cover)
- open the compare view and verify that you see the new charts and that the values seem accurate to the underlying data
- move the precipitation slider and verify that when the polling stops, the values in the chart have changed
- click to see the table view, then move the precipitation slider, and verify that those values change too


kellyi:
Realized I left off a change which forces the y axis for the non-Combined-Hydrology charts to go slightly beyond the max value of the bars:

screen shot 2017-08-09 at 5 51 28 pm

Amended the commit & pushed it anew.

rajadain:
> get this branch, then `update` and `server`

This is one of our few projects to which this doesn't apply.

kellyi:
> This is one of our few projects to which this doesn't apply.

Whoops! `bundle.sh` then hard refresh instead!

rajadain:
The tick labels on the Combined Hydrology chart are incorrect:

image

This should be 10 instead of 1. The labels on the other charts are correct.

src/mmw/js/src/compare/views.js
```diff
@@ -46,6 +48,18 @@ var CompareWindow2 = Marionette.LayoutView.extend({
         sectionsRegion: '.compare-sections',
     },
 
+    initialize: function() {
+        var self = this;
+        this.model.get('scenarios').forEach(function(scenario) {
+            scenario.get('results').on('change', function(s) {
+                if (s.attributes.name === 'runoff' && !s.attributes.polling) {
```
arottersman: We should still use the backbone syntax: `s.get('name')` and `s.get('polling')`. We also may want to refer to it as `r`, for "result"

src/mmw/js/src/compare/views.js
```diff
@@ -472,8 +534,74 @@ function getTr55Tabs(scenarios) {
                 })
             },
         ],
-        // TODO Make Runoff charts
-        runoffCharts = [],
+        runoffCharts = [
```
arottersman: @rajadain Was there a reason for all the arrays/objects in this function to not be actual models?
 Why doesn't this return a`TabsCollection`?

kellyi:
> The tick labels on the Combined Hydrology chart are incorrect:

Good catch. The combined hydrology chart needs to get the precipitation slider value as an input so it can use it to set its `yDomain` max value.

rajadain:
This is interesting: when the precipitation value is changed, we run a TR-55 job for all four scenarios (in this example), and update the chart as the values return. As the initial values return, there is great variation in the chart, as we have some new low values coexisting with high old values. Once all the new ones are gathered, the chart rescales itself to the new standard. This is not very obvious to the user (especially because there are no spinners or loading symbols, more on that later), but the effect can be jarring:



Can we make it so that the chart updates when _every_ scenario has finished running instead of individual ones?

rajadain:
In the wireframes: https://marvelapp.com/8hi9jg9/screen/29125398, the background color of the columns alternates. Was any thought given to this here?

rajadain:
It looks like all the edge cases I found have been noted above in the description. Otherwise this works well, and moves us closer. Going to take a look at the code now.

kellyi:
> the background color of the columns alternates. Was any thought given to this here?

Whoops, missed that one. Should be simple enough to do with index mod 2 though, so I can add a commit to fix it.

Working on the combined hydrology ydomain value, too.

src/mmw/js/src/core/chart.js
```diff
@@ -346,6 +346,8 @@ function renderCompareMultibarChart(chartEl, name, label, colors, stacked, data)
         if (maxValue) {
             yDomainMax = maxValue * 1.2;
         }
+    } else {
+        yDomainMax = precipitation || 1;
```
kellyi: The fallback's to ensure that if precipitation's set to 0, the graph will still range from 0 to 1.

kellyi:
8a046c3 departs from the wireframes to show the max value on the y-axis:

screen shot 2017-08-10 at 12 14 08 pm

If we decide we *don't* want to show them, I can either drop the commit or we can set `showMaxMin` to be `false` in the future.

rajadain:
0.1 is shown twice in different locations in the Evapotranspiration chart in the screenshot. That can't be right.

kellyi:
> 0.1 is shown twice in different locations in the Evapotranspiration chart in the screenshot. That can't be right.

Ah, looks like we have to set the tick count differently if there are only two values.

kellyi:
I squashed this down to one commit & rebased it on `develop` so it'll be a single commit while adjusting the model & collection hierarchy here.

jfrankl:
Good catch. Let's model this after the proposed loading designs seen here: https://github.com/WikiWatershed/model-my-watershed/issues/1848

While model is running:

1. Add spinner
2. Change opacity of bars in bar graphs

Details in the screenshot below:

mmw-model-running

rajadain:
Closing in favor of #2148.

Issue #2138, opened 2017-08-10
Implement treatment of HUC search result


ajrobbins:
Currently, searching and selecting a HUC from the MMW geocoder does not do anything (search Schuylkill as an example).

Jeff has created wireframes to show what should happen when a HUC search result is selected: https://marvelapp.com/33efhfb/screen/31228062

- HUC is highlighted on map
- Sidebar is switched to Select Area page
- HUC name replaces placeholder text in geocoder bar
- "Select" button appears in geocoder bar. Clicking this button sends to you analyze with the HUC as the AoI

Also: replace current geocoder bar placeholder text with "Jump to location or HUC", which hints at the HUC-searching ability and is more accurate than "Search"

PR #2147, opened 2017-08-14
Allow analyzing a boundary result from geocoder

arottersman:
## Overview

Implements behavior in [these wireframes](https://marvelapp.com/33efhfb/screen/31228062)

- when you search for and select a location, the selected location's name appears in the geocoder, and you are directed to `/draw`
- if the the selected location is a boundary layer, a `Select` button appears in the geocoder
- clicking the `Select` button analyzes the shape

#### Enhancement
- Pressing `ESC` now clears the selected shape and resets the query

Connects #2138 

### Demo
_Selecting a boundary_
xbgyajudai

_If you select a regular search result, displays the address_
screen shot 2017-08-14 at 3 27 07 pm

## Testing Instructions

**Address search**
* search for an address
* select a result
* confirm it appears in the search bar, that there are no errors in the console, and you are not directed to `/draw`
* press `ESC` and confirm the result clears
* search for something else and press `ENTER`. The first result should get selected

**Boundary search**
* search for and select a HUC
* Confirm you are taken to `/draw`
* Click `Select`
* Confirm you are taken to `/analyze` and the shape name, layer name, and wkaoi properties have come through (the names should be at the top of the sidebar)
* search again and this time press `ESC` --> the selected shape should clear from the map

### DIFF
--- a/src/mmw/js/src/draw/controllers.js
+++ b/src/mmw/js/src/draw/controllers.js
@@ -20,7 +20,10 @@ var DrawController = {
         toolbarModel.set('predefinedShapeTypes',
             _.filter(settings.get('boundary_layers'), { selectable: true }));
 
-        App.rootView.geocodeSearchRegion.show(geocodeSearch);
+        if (!App.rootView.geocodeSearchRegion.hasView()) {
+            App.rootView.geocodeSearchRegion.show(geocodeSearch);
+        }
+
         App.rootView.sidebarRegion.show(new views.DrawWindow({
             model: toolbarModel
         }));
--- a/src/mmw/js/src/core/models.js
+++ b/src/mmw/js/src/core/models.js
@@ -22,7 +22,8 @@ var MapModel = Backbone.Model.extend({
         geolocationEnabled: true,
         previousAreaOfInterest: null,
         dataCatalogResults: null,       // GeoJSON array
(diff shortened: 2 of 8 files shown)


src/mmw/js/src/draw/controllers.js
```diff
@@ -18,9 +18,12 @@ var DrawController = {
             toolbarModel = new models.ToolbarModel();
 
         toolbarModel.set('predefinedShapeTypes',
-            _.filter(settings.get('boundary_layers'), { selectable: true }));
+                         _.filter(settings.get('boundary_layers'), { selectable: true }));
```
arottersman: This auto-indented. I'll fix it.

ajrobbins:
In the wireframes, the search result AoI has a red outline and is not reverse masked, which seems helpful if the user wants to inspect the area around the AoI and maybe change AoIs before selecting.

arottersman:
> In the wireframes, the search result AoI has a red outline and is not reverse masked, which seems helpful if the user wants to inspect the area around the AoI and maybe change AoIs before selecting.

Thanks for catching that. I'll add a commit.

arottersman:
This should be ready for another look.


wiwm21uyrw

arottersman:
Thanks!

Issue #2098, opened 2017-07-28
Compare View Revamp: IE11 Flexbox Issues


rajadain:
Due to a number of Flexbox issues in IE11 (see http://caniuse.com/#feat=flexbox and https://github.com/philipwalton/flexbugs), the new Compare View is unusable in it:

![image](https://user-images.githubusercontent.com/1430060/28732925-cfca32c0-73a7-11e7-884f-8122ce8c6b84.png)

vs

![image](https://user-images.githubusercontent.com/1430060/28732948-e1a91a60-73a7-11e7-97a9-19c945f10602.png)

Note the difference in horizontal alignment, missing sections view, and missing close button.

Fix these issues and make it work as it does in other browsers.

rajadain:
A clue from @jfrankl https://github.com/WikiWatershed/model-my-watershed/pull/2094#issuecomment-318741483

> does this project use autoprefixer? If not, we'll need to manually run some kind of [autoprefixer script](https://autoprefixer.github.io/) once to handle all of that flexbox. In my experience, doing that, immediately fixes all problems with IE11.

rajadain:
As of #2191, the issues still persist:

![image](https://user-images.githubusercontent.com/1430060/29716754-953a5304-897a-11e7-89eb-5635f74242ce.png)

Notably, the modal is not aligned left instead of being centered on the page, and the close button is missing.

rajadain:
Also, the loading spinner from #2226 is not visible.

ajrobbins:
I'm also not seeing the combined hydrology chart graphics in IE compare view:

![screen shot 2017-09-29 at 5 09 50 pm](https://user-images.githubusercontent.com/12868342/31036863-21a0f15e-a53c-11e7-8ec2-51581b5eadb5.png)

caseycesari:
@arottersman My approach to this was to install the autoprefixer CLI and integrate it into our bundling process, and hope that this resolves the IE issues.

PR #2148, opened 2017-08-14
Compare View Revamp: TR-55 Runoff Charts and Tables

rajadain:
## Overview

Updated version of #2123. For more granular albeit noisy commits, see here: https://github.com/WikiWatershed/model-my-watershed/compare/develop...tt+ki/compare-view-tr55-runoff

Connects #2072 

## Testing Instructions

 * Check out this branch
 * Open a saved project, or create a new one, using TR-55 as the model and at least 2 scenarios
 * Open the Compare View
 * Ensure the values are correct in both the chart and the table
 * Move the precipitation slider, ensure the values update correctly
 * Ensure that the values of the original scenarios do not change after closing the Compare View


src/mmw/js/src/core/chart.js
```diff
@@ -314,8 +314,86 @@ function renderLineChart(chartEl, data, options) {
...
+             .reduceXTicks(false)
+             .staggerLabels($svg.width() < widthCutoff)
+             .duration(0)
+             .margin(options.margin)
+             .color(colors)
+             .showXAxis(false)
+             .id(name)
+             .yDomain([0, yDomainMax]);
+
+
```
kellyi: I accidentally added 2 extra whitespace lines here; probably worth dropping them.

src/mmw/js/src/compare/models.js
```diff
@@ -1,11 +1,80 @@
...
+        this.scenarios = options.scenarios;
+
+        this.scenarios.forEach(function(scenario) {
+            scenario.get('results').on('change', update);
+        });
+    }
+});
+
+var Tr55RunoffCharts = ChartRowsCollection.extend({
+    update: function() {
```
arottersman: Noice.

kellyi:
Taking a look.

kellyi:
This is looking good! We can continue to think about this but I kind of like the idea of dropping the top yDomain value, since it's sometimes very close in spatial terms to the existing tick:

screen shot 2017-08-15 at 10 22 06 am

Also I think we may wanna adjust the `tickCount` setting here -- https://github.com/WikiWatershed/model-my-watershed/pull/2148/files#diff-3e33aa79d87dd300905a2a0b6c714e56R351 -- or nearby, as it looks like it's not catching every case in which the values are small:

screen shot 2017-08-15 at 10 24 49 am

I think `if (maxValue <= 0.2)` *should* work there.

kellyi:
Checked this out in Chrome, FF, Safari, and IE11. Works well in Chrome & FF; but the charts & table don't display in either IE11 or Safari. Looks like the DOM elements are there, but they're not rendering visibly:

screen shot 2017-08-15 at 10 39 33 am

I imagine this has something to do with the CSS rules we're applying to the row elements?

rajadain:
What is the desired behavior for the ticks? I'm not sure what changing the value to `0.2` wants to achieve.

kellyi:
> I'm not sure what changing the value to 0.2 wants to achieve.

Looks like the evapotranspiration chart above replicates the quirk noted here whereby a tick value's shown 2x since we're enforcing a set number of tick values, but the range [0.0..0.2] only has 3 to display.

rajadain:
I ended up removing the top label in the charts:

image

Also, Safari issues should now be fixed. Tried to fix IE11 issues, but that was more complicated, so I'm leaving it for #2098.

It should be noted that the hover tips are inaccurate because they are rounded. We should have more accurate values in a future card that should make the tooltips look like the mockup.

rajadain:
@azavea-bot rebuild

kellyi:
Tested this out and generally working well in Chrome, FF, and Safari! I noticed two immediate quirks which I think we should handle elsewhere:

- the stacked bar chart initially shows up as non-stacked on all browsers; subsequently it becomes a stacked chart, and I think this may be worth dealing with when we address the loading state:

screen shot 2017-08-17 at 10 08 29 am

I'm not sure what part of this was obviating this but I think something about the `if (!stacked)... else` check may have guarded against the change.

- in some cases I'm able to hit the back button in the browser and have the compare view show above the Projects list:

screen shot 2017-08-17 at 10 12 33 am

Not sure what causes this but we should make an issue to handle fixing it.

rajadain:
The most recent two commits should take care of the mentioned issues:

rajadain:
Thanks for reviewing! Going to squash and merge shortly.

Issue #1934, opened 2017-06-01
BiG CZ details view


kdeloach:
Clicking on search results should open a details view. The main difference between the list view and the details view is the expanded "description" section and download buttons.

Other details:

* We don't need to expand the description field in the listing view anymore. This functionality can be removed.
* Author and date sections should read "Not Available" if that data is null or empty.
* Context is not always available for service links. Some services point to a details page, some point to an XML file, etc. So we may not be able to display 3 distinct download buttons.
* The list view will show all available boundaries on the map, the details view will only show the boundaries relevant to the selected search result. The actual implementation to show boundaries on the map is not part of this task.

![screenshot 2017-06-01 at 11 03 48](https://cloud.githubusercontent.com/assets/43062/26686127/106c04ae-46ba-11e7-9faa-721e15bf1933.png)

rajadain:
Commenting to subscribe to this thread and get email notifications.

mmcfarland:
This issue is a few months old.  @ajrobbins or @jfrankl can you comment on if the screencap and description are still valid with the latest design and thinking?

jfrankl:
There might need to be other details listed in addition to the author and year. Those details should match what we are now showing on the search results for each catalog. @ajrobbins what do you think?

ajrobbins:
The contents of the detail page will differ across catalogs. For this card, I think we should implement the basic detail page interaction and design (clicking on result slides out a detail page with an X to close), and include the title, author, date, and description from the result as the screencap shows.

As we get finalized detail on other elements to include, we can implement them in subsequent tasks.

PR #2150, opened 2017-08-15
Implement details view for BiG CZ

caseycesari:
## Overview

Clicking on a search result in the BiG CZ results pane will open a pane with the details for that result. Only the fields that were common between all three APIs were added.

Connects to #1934

### Demo

bigcz3

### Notes

I tried various approaches to wiring up the events, and unfortunately, the only one I could get working was propagating the event manually. Triggering a collection event didn't work nor did using `childViewEvents`.

## Testing Instructions

- Visit the BiGCZ site.
- Draw an AoI and make a search.
- For each API, click a few results, and verify the details view is displayed.

### DIFF
--- a/src/mmw/js/src/data_catalog/templates/resultDetails.html
+++ b/src/mmw/js/src/data_catalog/templates/resultDetails.html
@@ -1,0 +1,35 @@
+<div class="result">
+    <div class="result-detail-header">
+        <h2>
+            {{ title }}
+        </h2>
+        <button type="button" class="close" aria-label="Close">
+            <span aria-hidden="true">×</span>
+        </button>
+        <a href="#"><i class="fa fa-search-plus"></i> Zoom to extent</a>
+    </div>
+    <hr>
+    {% if author %}
+        <p>
+            <i class="fa fa-user"></i>{{ author }}
+        </p>
+    {% endif %}
+    {% if activeCatalog != 'cuahsi' %}
+        <p>
+            <i class="fa fa-calendar"></i> {{ created_at|toDateFullYear }}
+        </p>
+    {% else %}
+        {% if begin_date == end_date %}
+            <p>
+                <i class="fa fa-calendar"></i> {{ begin_date|toDateWithoutTime }}
+            </p>
+        {% else %}
+            <p>
+                <i class="fa fa-calendar"></i> {{ begin_date|toDateWithoutTime }} - {{ end_date|toDateWithoutTime }}
+            </p>
+        {% endif %}
+    {% endif %}
+    <p>
+        {{ description }}
+    </p>
+</div>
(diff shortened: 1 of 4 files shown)


src/mmw/js/src/data_catalog/views.js
```diff
@@ -71,18 +72,25 @@ var DataCatalogWindow = Marionette.LayoutView.extend({
 
     ui: {
         introText: '.intro-text',
-        tabs: '.output-tabs-wrapper'
+        tabs: '.output-tabs-wrapper',
+        closeDetails: '.close'
```
caseycesari: Good call. Will do.

src/mmw/js/src/data_catalog/templates/resultDetails.html
```diff
@@ -0,0 +1,20 @@
...
+        <a href="#"><i class="fa fa-search-plus"></i> Zoom to extent</a>
+    </div>
+    <hr>
+    {% if author %}
+        <p>
+            <i class="fa fa-user"></i>{{ author }}
+        </p>
+    {% endif %}
+    <p>
+        {{ description }}
```
caseycesari: I didn't implement this originally because the APIs handle dates in different ways. But after rereading the card, I'm going to take a shot at it.

caseycesari:
Thanks for the thorough review.

> Should this at some point make the corresponding result's shape active on the map?

Yeah, though that will be handled in a follow up card. The original description of this issue says not to do it as part of this issue. I'll make a follow up card.

> When the title wraps, the "x" renders below it. Not sure if now is a good time to address

I'll see if I can fix this.

caseycesari:
@arottersman I made the requested changes. Ready for another look!

src/mmw/js/src/data_catalog/templates/resultDetails.html
arottersman: We could also check for `begin_date` and `end_date` if they're not present, fall back to to `created_at`. That way we wouldn't have to worry about which catalog is active.

arottersman:
I left a review but then deleted its only comment to move it somewhere else. Now the whole review is gone....

src/mmw/js/src/data_catalog/templates/resultDetails.html
```diff
@@ -0,0 +1,35 @@
...
+        </button>
+        <a href="#"><i class="fa fa-search-plus"></i> Zoom to extent</a>
+    </div>
+    <hr>
+    {% if author %}
+        <p>
+            <i class="fa fa-user"></i>{{ author }}
+        </p>
+    {% endif %}
+    {% if activeCatalog != 'cuahsi' %}
```
arottersman: We could also check if there's `begin_date` instead of checking on the active catalog.

caseycesari: I tried something similar, but there is a bit of an overlap in fields between the two catalogs, so I wanted to be more explicit. We may end up having to change this in the future too depending on how much the detail panes diverge between the catalogs.

caseycesari:

Issue #2135, opened 2017-08-10
Create ADR for geoprocessing API


mmcfarland:
The current analyze and RWD generation endpoints are going to be exposed via a proper API, intended for developer consumption.   These are both currently exposed through django-based views, but will need to be reformatted in the context of DRF in order to provide expected API features.  Create an ADR which describes the goals of the API and the details around authentication, usage logging, administration and layout.

PR #2170, opened 2017-08-21
Add Geoprocessing API ADR

mmcfarland:
## Overview

A decision record on how to implement the external geoprocessing API, along with objectives for its implementation. 

 TL;DR: 
>Use the existing Django REST Framework tooling to provide out of the box support for this feature as an MVP, and iterate for more robustness when it warrants.

Connects #2135

### DIFF
--- a/doc/arch/adr-004-geoprocessing-api.md
+++ b/doc/arch/adr-004-geoprocessing-api.md
@@ -1,0 +1,58 @@
+## 004 - Expose Geoprocessing API
+
+### Context
+The MMW/BigCZ platform performs a number of geospatial processing tasks which can be initiated by submitting tasks to an HTTP endpoint.  The web application provides the tools and workflow for assembling and submitting the input, monitoring execution progress, and collecting and visualizing the results of these jobs.  The platform provides value by assembling the required national scale data and the infrastructure necessary to perform the processing. In an effort to extend this capability to alternate workflows that don't involve the web platform, the client wants to formalize an externally consumable HTTP API.
+
+MMW/BigCZ already includes and makes limited use of the Django REST Framework (DRF), which supports convenience methods around creating HTTP APIs.
+
+#### API Objectives
+* Support execution of all current and future "Analyze" sub tasks (Land Use, Soil, etc)
+* Support execution of Rapid Watershed Delineation
+* Model execution functionality will not be exposed
+* Available only for registered users
+* Usage will be recorded
+* Long term goal would be to charge for access
+
+#### Considerations
+* Geoprocessing routines are asynchronous, requiring submission and polling to determine when results are available.
+* Geoprocessing is a resource intensive operation.
+* Geoprocessing endpoints must remain open for anonymous users using the web platform.
+* Azavea maintains the only client of the current API, therefore we have had the freedom to make changes at will.  When an API is released, we'll have to treat releases of the platform carefully and possibly retain backwards compatibility.
+
+Based on the above, an API solution should adhere to the following implementation details:
+##### 1. Have a mechanism for authenticating a user
+Two options seem to be potentially appropriate: an API key associated with a user account, or an expiring token generated based on client supplied credentials.  API keys are easier to use, but inherently less secure since they can be reused by multiple parties and have to be included in the request. They can be revoked, allowing a failsafe against long term unauthorized use. Expiring tokens require the user to securely manage credentials, which can be difficult, especially when the desire is to use the API from frontend browser code.
+##### 2. Have at least rudimentary rate limiting capabilities
+Unrestricted job submission, even if used modestly, could have a considerable negative impact on the web platform performance.  If users anticipate using the API for batch processing, they should be restricted in how often they can submit.
+##### 3. Not result in duplicated code
+The web platform needs to have access to unauthenticated endpoints.  This could theoretically be accomplished through domain white listing within the DRF endpoints or the existing endpoints.  Regardless, the implementation should be generalized in one specific place that the API and platform utilize.
+##### 4. Enforce size and shape validation
+The front end code which submits analysis jobs is responsible for the majority of shape validation.  This will need to be included in the API, as well as for additional potential conflicts like coordinate systems and being contained within the CONUS.  This may require some restriction on upload payload size, which is currently handled through nginx.
+
+#### Options
+##### Reuse existing endpoints
+Our existing endpoints `/start/analyze/*` and `/start/rwd` have the exact functionality that is needed, but are limited by specific implementation details:
+* Do not require authentication
+* Validation is largely handled by front end submission routine
+* Paths and naming were not selected in the context of a public API
+* CORS is not enabled
+
+##### Create new DRF based API endpoints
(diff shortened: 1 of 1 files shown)


doc/arch/adr-004-geoprocessing-api.md
```diff
@@ -0,0 +1,50 @@
+## 004 - Expose Geoprocessing API
+
+### Context
+The MMW/BigCZ platform performs a number of geospatial processing tasks which can be initiated by submitting tasks to an HTTP endpoint.  The web application provides the tools and workflow for assembling and submitting the input, monitoring execution progress, and collecting and visualizing the results of these jobs.  The platform provides value by assembling the required national scale data and the infrastructure necessary to perform the processing. In an effort to extend this capability to alternate workflows that don't involve the web platform, the client wants to formalize an externally consumable HTTP API.
+
+#### API Objectives
+* Support execution of all current and future "Analyze" sub tasks (Land Use, Soil, etc)
+* Support execution of Rapid Watershed Delineation
+* No model execution will be possible
```
arottersman: Is model execution actually not possible, or do you mean that we don't have to make it possible? Consider rephrasing to "Will not support model execution"

doc/arch/adr-004-geoprocessing-api.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Consequences
+Using non-expiring keys is a trade off between security and consumer convenience.  As none of the content is currently protected, it is better to have a low barrier to entry to determine if there is an actual user demand for this feature. If, in the future, a paid tier is added, there are enough workflow updates required to provide an opportunity to revisit that mechanism.  Since we'll be logging to the database, we'll need to provide an interface into that data.  I suggest we keep that simple and create a management task which exports ranges as a CSV so they can be evaluated in Excel.  If this feature ends up gaining traction, we'll want to invest in in a more robust solution.
```
arottersman: > Since we'll be logging to the database, we'll need to provide an interface into that data.  I suggest we keep that simple and create a management task which exports ranges as a CSV so they can be evaluated in Excel.  If this feature ends up gaining traction, we'll want to invest in in a more robust solution.

Consider breaking this off into its own paragraph.

doc/arch/adr-004-geoprocessing-api.md
```diff
@@ -0,0 +1,50 @@
...
+
+
+##### Use AWS API Gateway to proxy API requests
+AGW could pass requests to existing or new endpoints and comes with a very robust system feature set including usage plans, rate limiting, logging and versioning.  The tooling and execution around this service is relatively unknown on the team.  Also, it would need to authorize requests based on authentication to the application, requiring custom code.
+
+
+### Decision
+We will use the existing Django REST Framework to implement new API endpoints, secured by the default Token Authentication provided by the framework.  We will throttle requests to a configurable amount per minute.  We will have to add an account page that lists the users API key and allows it to be invalided/regenerated.  Validation will encoded into the endpoints and the application/environment will have a user associated with it, and will use that user's API key against the same endpoints. Basic logging of requests will be done within the Django model/database with attributes confirmed by the client.
```
arottersman: > Validation will encoded into the endpoints and the application/environment

typo

mmcfarland:
48d0441 added to address feedback

doc/arch/adr-004-geoprocessing-api.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Consequences
+Using non-expiring keys is a trade off between security and consumer convenience.  As none of the content is currently protected, it is better to have a low barrier to entry to determine if there is an actual user demand for this feature. If, in the future, a paid tier is added, there are enough workflow updates required to provide an opportunity to revisit that mechanism.
+
+Since we'll be logging to the database, we'll need to provide an interface into that data.  I suggest we keep that simple and create a management task which exports ranges as a CSV so they can be evaluated in Excel.  If this feature ends up gaining traction, we'll want to invest in in a more robust solution.
```
rajadain: Consider using "exports usage data" instead of "exports ranges" to clarify what will be exported.

doc/arch/adr-004-geoprocessing-api.md
```diff
@@ -0,0 +1,56 @@
...
+##### 2. Have at least rudimentary rate limiting capabilities
+Unrestricted job submission, even if used modestly, could have a considerable negative impact on the web platform performance.  If users anticipate using the API for batch processing, they should be restricted in how often they can submit.
+##### 3. Not result in duplicated code
+The web platform needs to have access to unauthenticated endpoints.  This could theoretically be accomplished through domain white listing within the DRF endpoints or the existing endpoints.  Regardless, the implementation should be generalized in one specific place that the API and platform utilize.
+##### 4. Enforce size and shape validation
+The front end code which submits analysis jobs is responsible for the majority of shape validation.  This will need to be included in the API, as well as for additional potential conflicts like coordinate systems and being contained within the CONUS.
```
caseycesari: This will probably require that the API in some way ingest the geometry that is submitted with the request. We might need to think of a clever way to do this so that ingesting large shapes does hurt the performance of the API.

mmcfarland: Do you mean ingest in a way that the app is currently not doing?  I'm picturing a geojson payload, and we do checks on area size, CONUS and self intersecting.  We may want to guard against extremely large payload sizes, which we could probably do via nginx.  Did you have something more in mind?

caseycesari: > I'm picturing a geojson payload, and we do checks on area size, CONUS and self intersecting.

That's what I was thinking too. I was thinking about this part of the process, and wether we need to store anything, if this part will be asynchronous, if complex shapes will hold things up, etc.

> We may want to guard against extremely large payload sizes, which we could probably do via nginx.

Yeah, I was thinking about this too.

doc/arch/adr-004-geoprocessing-api.md
```diff
@@ -0,0 +1,56 @@
...
+
+
+##### Use AWS API Gateway to proxy API requests
+AGW could pass requests to existing or new endpoints and comes with a very robust system feature set including usage plans, rate limiting, logging and versioning.  The tooling and execution around this service is relatively unknown on the team.  Also, it would need to authorize requests based on authentication to the application, requiring custom code.
+
+
+### Decision
+We will use the existing Django REST Framework to implement new API endpoints, secured by the default Token Authentication provided by the framework.  We will throttle requests to a configurable amount per minute.  We will have to add an account page that lists the users API key and allows it to be invalided/regenerated.  Validation will be encoded into the endpoints and the application/environment will have a user associated with it, and will use that user's API key against the same endpoints. Basic logging of requests will be done within the Django model/database with attributes confirmed by the client.
```
hectcastro: Looking around, there seems to be a [well-used plugin](http://drf-tracking.readthedocs.io/en/latest/) for tracking requests in the database. Could be an easy way to add that feature.

doc/arch/adr-004-geoprocessing-api.md
```diff
@@ -0,0 +1,56 @@
...
+
+#### API Objectives
+* Support execution of all current and future "Analyze" sub tasks (Land Use, Soil, etc)
+* Support execution of Rapid Watershed Delineation
+* Model execution functionality will not be exposed
+* Available only for registered users
+* Usage will be recorded
+* Long term goal would be to charge for access
+
+#### Considerations
```
hectcastro: I think that it would be valuable to consider adding Swagger support from the start. Not just for documentation (which we'll probably need to supply in some capacity), but also for aiding in API design.

mmcfarland:
Thanks for the suggestions.  I've updated the doc to reference a lot of the implementation details presented and it sounds like there is no objection to the broad decisions, so I'm merging in.

Issue #2128, opened 2017-08-08
Regression: no longer displaying RWD-generated AoI when too large to analyze 


ajrobbins:
Currently, RWD works correctly to delineate large watersheds, like this one in the Mississippi Delta. However, if it's too large to analyze, we aren't showing the delineated AoI shape like we did previously. 

(from prod)
![screen shot 2017-08-08 at 3 43 41 pm](https://user-images.githubusercontent.com/12868342/29091317-9468e4a4-7c50-11e7-8e18-8871aacb8d0b.png)

Expected behavior: Delineated AoI is shown on the map, but masked out with the AoI size warning on top.

PR #2171, opened 2017-08-22
Keep invalid RWD AoI on map

mmcfarland:
## Overview

Previous behavior allowed the AoI generated via RWD to remain displayed
on the map, even if it was invalid for analysis and modelling.  In a
recent refactor, this behavior was lost.  This commit restores the
functionality by allowing options to the reset function, based on the
feature invoking it.

Connects #2128 

### Demo

screenshot from 2017-08-22 09 51 24

screenshot from 2017-08-22 09 51 35

### Notes

This was a result of some code consolidation around the various draw tools commited in https://github.com/WikiWatershed/model-my-watershed/commit/b1efa4f80ed9858f6d62c745b41d6f0372ddc9d4

## Testing Instructions

* Ensure RWD data are loaded in your VM
* `./scripts/bundle.sh --debug`
* Attempt to delineate a watershed which will be too large to analyze.  I use a spot along the Illinois River, a few miles upstream from St. Louis.
* Ensure that the shape is preserved after dismissing the warning
* Ensure that selecting another draw mechanism clears the AoI and point markers
* Ensure that other draw methods still function appropriately

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -172,9 +172,12 @@ var DrawWindow = Marionette.LayoutView.extend({
 
     onShow: function() {
         var self = this,
-            resetDrawingState = function() {
-                self.resetDrawingState();
-            };
+            resetRwdDrawingState = function() {
+                self.resetDrawingState({
+                    clearOnFailure: false
+                });
+            },
+            resetDrawingState = _.bind(self.resetDrawingState, self);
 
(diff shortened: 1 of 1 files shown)


kellyi:
This is working well!

screen shot 2017-08-22 at 1 30 51 pm

For future reference it looks like points on this spot just west of King, Illinois can reliably generate huge RWD shapes:

screen shot 2017-08-22 at 1 31 17 pm

Issue #2136, opened 2017-08-10
Clicking on search result map marker should open popup


ajrobbins:
From wireframes, slide 20 https://marvelapp.com/5a66caa/screen/29866796

Currently, clicking on one of the points (WDC) or polygons (CINERGI, HydroShare) that appear on the map after a search does not do anything. Clicking on a map marker should open a popup:

![screen shot 2017-08-10 at 1 16 46 pm](https://user-images.githubusercontent.com/12868342/29182799-32438c96-7dce-11e7-9619-20515a259c7d.png)

Popup should include title, date, and button/link to result detail page.

Issue #1950, opened 2017-06-06
Include properties in BiG CZ GeoJSON results


kdeloach:
Instead of returning a list of results which contain GeoJSON fields, we may to return a  GeoJSON collection with properties. This would simplify interactions between the sidebar and map on the frontend by reducing the amount of data transformations needed.

For example, instead of:

```json
{
   "catalog":"cinergi",
   "count":1,
   "results":[
      {
         "id":"edc761d4fe734000ba0ba228867ae9cc",
         "title":"Tropospheric Emission Spectrometer (TES) Data and Information",
         "description":"...",
         "links":[...],
         "created_at":"2017-03-17T17:14:29.817000Z",
         "updated_at":"2017-01-13T23:01:16Z",
         "geom":{
            "type":"Polygon",
            "coordinates":[...]
         }
      }
   ]
}
```

we could to:

```json
{
   "catalog":"cinergi",
   "count":1,
   "results":[
      {
         "type":"Polygon",
         "coordinates":[...],
         "properties": {
            "id":"edc761d4fe734000ba0ba228867ae9cc",
            "title":"Tropospheric Emission Spectrometer (TES) Data and Information",
            "description":"...",
            "links":[...],
            "created_at":"2017-03-17T17:14:29.817000Z",
            "updated_at":"2017-01-13T23:01:16Z"
         }
      }
   ]
}
```

We just need to think about how to handle results with unavailable geometries (ex. HydroShare).

Ref: https://github.com/WikiWatershed/model-my-watershed/pull/1948/commits/317528a86568925fc85cc993ab28ceff1bf4a00d

rajadain:
Commenting to subscribe to this thread and get email notifications.

PR #2173, opened 2017-08-22
BiG-CZ Map Popups

arottersman:
## Overview
Adds popovers to the BiG-CZ results we display on the map (Currently CINERGI and WDC). 

**Wireframe**
screen shot 2017-08-22 at 11 18 33 am

Connects #2136 

### Demo

screen shot 2017-08-22 at 11 10 47 am
screen shot 2017-08-22 at 11 11 20 am

th8cajmzhy

### Notes
1. The wireframes don't reflect the actual data for the catalogs, so I just reused the `ResultView` for the popups while we determine what they should contain
2. The interaction for opening a CINERGI popup is weird because you have to click the skinny boundary line. Clicking the inside of the boundaries would also be bad though; boundaries overlap.
3. I didn't match the wireframe "detail" button exactly because setting it to `ui-secondary` seemed much to dark on the white background
3. Refactored the enter-detail-mode triggering to allow the popup access

## Testing Instructions

 * Pull and `bundle`. There are style changes, make sure the sass compiles.
 * Get results for the three tabs by searching something like `water` 
    - Confirm you can still open the details from the sidebar from all three catalogs
    - Confirm if you click on a Cinergi map boundary, its corresponding popup opens and you can use its details button
    - Confirm the same for the WDC points


src/mmw/js/src/data_catalog/models.js
```diff
@@ -190,13 +199,53 @@ var Results = Backbone.Collection.extend({
...
+            // and those that intersect the area of interest (CINERGI and CUAHSI).
+            filteredResults = _.filter(data.results, function(r) {
+                return r.geom === null || turfIntersect(aoi, r.geom) !== undefined;
+            });
+
+        // Add the result's id onto the geojson's geom so the map can keep
+        // track of which shape came from where
+        return _.map(filteredResults, function(r) {
+            if (r.geom) {
+                r.geom.properties = {
```
arottersman: We could do this on the backend as well, but if we end up exposing the API, our users might not want the duplicate data clogging up the response.

jfrankl:
When the user clicks on "details" in the popup, the popup should close. Other than that, this is good to go!

FYI, @ajrobbins and I have another concept for the bounding box interaction. We'll write that up in a separate issue.

rajadain:
Waiting for a fixup to come in for the hover issue discussed IRL. Otherwise:

 * I like the popups, and agree that reusing the Result View for this purpose. We should remove the hover interaction of the ResultView in this case.
 * Clicking on the edge of CINERGI results feels a little odd. Good enough for now, but we discussed the possibility of sorting the items on the map to keep the smallest at the top, so you _could_ click on an inside area and it will select the smallest / most locally relevant one. This can be attempted later.
 * When we go into the details view, the map element should be highlighted until the user goes back. I remember there being some discussion of hiding all other map elements, keeping only the selected one, in the details view, but highlighting the selected one seems like an easy thing to do at this stage.

arottersman:
Pushed a commit to modify the highlight behavior.

One quirk is in the details page you can still open additional popups, and they will highlight.

rajadain:
I created #2181 for all the hover issues, even some found herein, so I think this is good to merge. Just going to take a quick peek over the code now.

arottersman:
Thanks, those are good ideas for the hovering. I think it might take some time to get the interactions right, so I'm all for deferring.

src/mmw/js/src/data_catalog/models.js
rajadain: If we ever did #1950 it would assist with this. For now I think this is fine.

arottersman:
Thanks! Going to merge.

Issue #2064, opened 2017-07-20
Analyze tables show too few records in a single page


mmcfarland:
When an analysis result has multiple pages of results, on Point Source/Water Quality, the record count per page is lower than expected, given the amount of space available (13" Chromebook).

![screenshot 2017-07-20 at 1 59 15 pm](https://user-images.githubusercontent.com/1014341/28431824-af0e2762-6d53-11e7-8357-f4a5d51178d2.png)

ajrobbins:
This was a result of an attempt to get the pagination to work on tablets and a variety of desktop sizes, I think in this PR https://github.com/WikiWatershed/model-my-watershed/pull/1849

mmcfarland:
> This was a result of an attempt to get the pagination to work on tablets and a variety of desktop sizes

I recall the issue, but the implementation which calculates the height leaves room for about a 100% increase in rows per page without over-consuming the whitespace (on a 13"), or half the number of pages needed to scroll through.  At other screen resolutions, the table makes use of the available whitespace before creating a smaller page size:

![screenshot from 2017-07-20 14 57 45](https://user-images.githubusercontent.com/1014341/28434064-16c69c56-6d5c-11e7-94b8-523ff6f92984.png)

caseycesari:
This appears to be mostly driven by the "City" column. If the text in that column is long enough to wrap (and there are enough of rows with wrapping text, which seems to happen often), the table takes up about the right amount of space. However, if none of the column values wrap, there is extra space for more rows.

![screen shot 2017-08-18 at 1 13 51 pm](https://user-images.githubusercontent.com/1042475/29469630-26d6acd6-8417-11e7-9f96-f91228bd11be.png)

![screen shot 2017-08-18 at 1 14 03 pm](https://user-images.githubusercontent.com/1042475/29469646-3a21aebc-8417-11e7-8d87-011933b85669.png)

caseycesari:
If we turn off text-wrapping for the "City" column, we could get a more consistent row height and fit more rows per page. The downside of that, is that it pushes the other columns further out of view.

![image](https://user-images.githubusercontent.com/1042475/29471212-252e2b4c-841d-11e7-82dd-705887111b96.png)

Is this acceptable?

rajadain:
Wouldn't this make the City column have different widths in different pages? We exchange dynamic cell height for cell width ...

arottersman:
[...]

PR #2174, opened 2017-08-22
Increase number of rows in pageable tables

caseycesari:
## Overview

Double the number of rows for the point source and water quality tables.

Word wrapping is turned off for these tables so that there is a uniform row
size. Discrepancies in table height were caused by wrapping of the City
column.

Connects to #2064

### Demo

screen shot 2017-08-22 at 2 34 45 pm

screen shot 2017-08-22 at 2 34 52 pm

### Notes

Depending on which browser you are using, there is a varying degree of whitespace in the tab content panes. This makes them scrollable even though there is nothing to scroll to but more whitespace. I made some attempts to reduce this whitespace, but its a little tough to deal with because of how it varies between browsers. If we want to fix this, we should gets some help from a designer.

## Testing Instructions

- Draw an AoI that is at least as large as Philadelphia, to ensure there are lots of results for the point source and water quality tables.
- Advance to analyze, and view the point source and water quality tabs.
- Ensure there are 6 rows of data per page, and there is not excessive whitespace.

### DIFF
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -1091,7 +1091,7 @@ var PointSourceResultView = AnalyzeResultView.extend({
             associatedLayerCodes = ['pointsource'],
             avgRowHeight = 30,  // Most rows are between 2-3 lines, 12px per line
             minScreenHeight = 768 + 45, // height of landscape iPad + extra content below the table
-            pageSize = utils.calculateVisibleRows(minScreenHeight, avgRowHeight, 3),
+            pageSize = utils.calculateVisibleRows(minScreenHeight, avgRowHeight, 6),
             chart = null;
         this.showAnalyzeResults(coreModels.PointSourceCensusCollection,
             PointSourcePageableTableView, chart, title, source, helpText, associatedLayerCodes, pageSize);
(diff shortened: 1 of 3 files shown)


arottersman:
Taking a look! Is it necessary to test on a real iPad or does the chrome simulation suffice?

caseycesari:
> Taking a look! Is it necessary to test on a real iPad or does the chrome simulation suffice?

I used the simulator. It varies by browser, but there always appears to be some amount of vertical scroll due to whitespace in the tab pane. That's to say I don't think it's a big deal if the bottom row is cut off in a particular browser. It looks good in the majority of them.

mmcfarland:
Definitely an improvement, though I still see a good amount of unused whitespace - is it just too variable really dial in the size?

screenshot from 2017-08-23 13 37 25

caseycesari:
> Definitely an improvement, though I still see a good amount of unused whitespace - is it just too variable really dial in the size?

What OS/browser is this screenshot taken from?

mmcfarland:
> What OS/browser is this screenshot taken from?

Chrome on Linux, but with iPad emulation.

caseycesari:
This is how it appears in the MacOS iPad simulator

screen shot 2017-08-23 at 3 58 35 pm

I can continue to tweak the function if this is still subpar, but I think erring on the side of showing too few rows instead of too many may be a better approach for more device compatibility.

caseycesari:
Thanks!

Issue #2127, opened 2017-08-08
Add nlcd ID to csv download of Land table


ajrobbins:
In https://github.com/WikiWatershed/model-my-watershed/pull/2044, we shifted to capturing just what's shown in the Analyze-land table in the csv download. Stroud thinks that knowing the NLCD ID, which we used to include in the csv, would be useful for users who want to reference the NLCD directly with our data. We should add NLCD ID back to the csv.

caseycesari:
With the way things are setup now, the ID would have to be added to the table to show up in the CSV. We could try using `visibility: hidden` on the columns we don't want to show up in the download, but I'm not sure if that would work.

kellyi:
Got this working for the NLCD id.

Would we want to have the same behavior for the soils tab? That one has a letter code rather than a numeric ID:

<img width="446" alt="screen shot 2017-08-22 at 3 00 03 pm" src="https://user-images.githubusercontent.com/4165523/29582417-9fec2f66-874a-11e7-9c3b-1905d79af4a3.png">

kellyi:
Current land tab btw:

<img width="453" alt="screen shot 2017-08-22 at 3 01 41 pm" src="https://user-images.githubusercontent.com/4165523/29582469-d7366d10-874a-11e7-96cd-5bb8ef2ffab4.png">

-> CSV ->

<img width="562" alt="screen shot 2017-08-22 at 3 02 15 pm" src="https://user-images.githubusercontent.com/4165523/29582499-ed471e92-874a-11e7-8601-e60326217c25.png">

ajrobbins:
Just NLCD please!

PR #2176, opened 2017-08-22
Add NLCD code to analyze land results csv

kellyi:
## Overview

This PR adds an "NLCD Code" column to the exported CSV, but not to the visible analyze results table. We accomplish this by adding a hidden column to the table which is registered to be exported along with the CSV.

Since both the land and soils tab use the TableView component, the changes here also include some checks for whether or not to add the hidden column. If we decide we want to show a soil code for that table, we can drop the checks in the template and instead pick the value and header text conditionally. 

Connects #2127

### Demo

screen shot 2017-08-22 at 3 01 41 pm

->

screen shot 2017-08-22 at 3 02 15 pm

## Testing Instructions
- get this branch, then `vagrant up` and `bundle`
- draw an AOI and wait for the analysis results to complete. Once they're back check that the tabs all render as before and that all the csv exports work as before -- aside from the land export, which now includes a "NLCD Code" column in its csv

### DIFF
--- a/src/mmw/js/src/analyze/templates/table.html
+++ b/src/mmw/js/src/analyze/templates/table.html
@@ -2,6 +2,9 @@
     <thead>
         <tr>
             <th data-sortable="true">Type</th>
+            {% if isLandTable %}
+                <th class="hidden-analyze-table-column" data-tableexport-display="always">NLCD Code</th>
+            {% endif %}
             <th class="text-right" data-sortable="true" data-sorter="window.numericSort">Area ({{ headerUnits }})</th>
             <th class="text-right" data-sortable="true" data-sorter="window.numericSort">Coverage (%)</th>
         </tr>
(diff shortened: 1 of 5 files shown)


src/mmw/js/src/analyze/templates/table.html
```diff
@@ -2,6 +2,9 @@
     <thead>
         <tr>
             <th data-sortable="true">Type</th>
+            {% if isLandTable %}
+                <th class="hidden-analyze-table-column" data-tableexport-display="always">NLCD Code</th>
```
caseycesari: Consider dropping this class and using the HTML `hidden` attribute.

kellyi: Tried this out & it seemed like there's some quirk getting the `hidden` attribute to take hold via the Nunjucks template.

Going to leave this as is for now, but we should try it again if we end up adjusting these to include more hidden columns.

Issue #2158, opened 2017-08-17
Compare View Revamp: Chart Tooltips


rajadain:
Add custom tooltips to chart to match the mockups:

![image](https://user-images.githubusercontent.com/1430060/29419897-6e4c13d2-833e-11e7-8b28-940c5629a0cd.png)

https://marvelapp.com/8hi9jg9/screen/29125399

Highlighting one section of a stacked bar chart should highlight the same section across all scenarios.

PR #2177, opened 2017-08-22
Compare View Revamp: TR-55 Water Quality Charts and Table

rajadain:
## Overview

Adds Water Quality charts and table for TR-55. Also enables tab switching, and adds visualization for chart / table switching. Builds on the work done in #2148.

I talked to @jfrankl about how to show the many values of Water Quality in the app:

image

On his recommendation, we're showing only the Loading Rate (kg/ha) values, corresponding to the values shown in the charts. All the values are simple mathematical transformations of the same value, thus the trend seen in the charts would hold for all of them.

Connects #2076 

## Testing Instructions

 * Load a saved project, or make a new one. Ensure it has at least 2 scenarios.
 * Open the Compare View
 * Switch between table and chart modes. Ensure the buttons update correctly.
 * Switch between Runoff and Water Quality tabs. Ensure the contents and the tab buttons update correctly.
 * Ensure the numbers in the Water Quality tab are accurate.
 * Ensure that changing the precipitation slider updates all values in all tabs

### DIFF
--- a/src/mmw/js/src/compare/models.js
+++ b/src/mmw/js/src/compare/models.js
@@ -34,6 +34,7 @@ var ChartRowsCollection = Backbone.Collection.extend({
         var update = _.bind(this.update, this);
 
         this.scenarios = options.scenarios;
+        this.aoiVolumeModel = options.aoiVolumeModel;
 
         this.scenarios.forEach(function(scenario) {
             scenario.get('results').on('change', update);
@@ -70,6 +71,31 @@ var Tr55RunoffCharts = ChartRowsCollection.extend({
             chart.set({
                 precipitation: precipitation,
                 values: values
+            });
+        });
+    }
+});
+
+var Tr55QualityCharts = ChartRowsCollection.extend({
+    update: function() {
+        var aoivm = this.aoiVolumeModel,
+            results = this.scenarios.map(function(scenario) {
+                return scenario.get('results')
+                               .findWhere({ name: 'quality' })
+                               .get('result');
+            });
+
+        this.forEach(function(chart) {
+            var name = chart.get('name'),
+                values = _.map(results, function(result) {
+                    var measures = result.quality.modified,
+                        load = _.find(measures, { measure: name }).load;
+
+                    return aoivm.getLoadingRate(load);
+                });
+
+            chart.set({
+                values: values,
             });
         });
     }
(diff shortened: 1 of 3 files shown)


arottersman:
Finally taking a look!

arottersman:
Taking a look at the code now. Testing went well except I think the rounding is making the loading rates all zero:

screen shot 2017-08-23 at 3 21 57 pm

rajadain:
The tooltips will be redone in #2158, so can be ignored for now.

arottersman:
It's on the y-axis ticks as well.

rajadain:
> It's on the y-axis ticks as well.

Ah, that's a real problem. Looking in to it.

src/mmw/js/src/compare/models.js
```diff
@@ -128,6 +156,35 @@ var Tr55RunoffTable = TableRowsCollection.extend({
...
 
+var Tr55QualityTable = TableRowsCollection.extend({
+    update: function() {
+        var aoivm = this.aoiVolumeModel,
+            results = this.scenarios.map(function(scenario) {
+                return scenario.get('results')
+                               .findWhere({ name: 'quality' })
+                               .get('result');
+            }),
+            get = function(key) {
```
arottersman: Could also name `go`

arottersman:
The code for this looks good! Once the loading rates axis is fixed, we should be good to go.

rajadain:
Okay, so we have 0s in the charts because the values are too small, as small as 0.006 kg / ha:

image

There's two ways to solve this:

 1. Make the axis have three decimal places of precision
 2. Change the units from kg/ha to g/ha and multiply with 1000

I'm leaning towards the latter, but want some feedback /cc @ajrobbins. Suspended Solids will remain in kg/ha, but Nitrogen and Phosphorus may work better as g/ha. Not sure if we'll want to change the values in the table as well, to match the charts, or keep them as kg/ha to better align with the first row.

rajadain:
Just pushed a fix that makes it look like this:

image

image

ajrobbins:
I think for comparison, we have to use the same units for all metrics on the page. Let's do three units of precision for now.

rajadain:
Added a commit that makes tick precision dynamic on the chart:

image

while maintaining the same units. Table is left as before, so same units all the way through. One final look @arottersman ?

rajadain:
Thanks for reviewing! Going to squash, rebase on develop, then merge.

Issue #2122, opened 2017-08-07
Assign color breaks to climate rasters


ajrobbins:
We are ingesting the BiG CZ climate rasters into GeoTrellis to create visual tiles. We need clarification from @emiliom as to the appropriate color ramp and color breaks for the rasters. There is some configuration in the sample JSON in the documentation that references this 1000×1px image: http://data.nanoos.org/files/cz/mapoverlays/ppt_colorbar.png

This could be the desired color ramp - we extracted 243 unique colors from the image to use. However, we need:
- confirmation if this is indeed the color ramp to be using for the climate rasters (both precip and temp); Is it important or meaningful to use the same color breaks as the previous tiles? Or could we use a color scheme more consistent with the new app?
- appropriate raster values to assign to color breaks

@emiliom Can you give us some guidance on the two questions above?

emiliom:
pinging @lsetiawan; remember that he's the real expert here on technical aspects of the JSON files and the climatology overlays. Don, please reply as needed.

> confirmation if this is indeed the color ramp to be using for the climate rasters (both precip and temp)

Based on the "ppt" file prefix, http://data.nanoos.org/files/cz/mapoverlays/ppt_colorbar.png must be the color ramp for precipitation only. The colorbar image for each overlay is referenced in the overlay's JSON file .

> Is it important or meaningful to use the same color breaks as the previous tiles? Or could we use a color scheme more consistent with the new app?

That's not entirely for me to answer. @aufdenkampe? But what I will say is that we spent a lot of time refining the colormaps and value ranges to work well for the variability found spatially and temporally, and based on common approaches for this type of data. And they've had some public exposure already, with generally positive feedback. For the MODIS EVI, I would strongly suggest that we stick with what we have. More broadly speaking, I would recommend using what we developed (color ramps and color breaks) and reassess in the future if needed and based on wider user input. That also would free you up at this time to focus on implementation challenges rather than rendering and styling decisions.

I'm also not quite sure what you mean by tuning color schemes of data to be more consistent with the app. With more overlay layers, in my experience the driving goals are instead distinctiveness across layers, coherence with common rendering practices in a core community, and a balanced representation of variability across the spatial and temporal domains.

> appropriate raster values to assign to color breaks

I'm not sure I understand the question. Those values are defined in the JSON files. I think @lsetiawan explained the file a while ago, but if not, let us know and he can do that.

emiliom:
[...]

Issue #1976, opened 2017-06-15
Tile full-resolution climate pngs


ajrobbins:
PNGs available here: https://github.com/BiG-CZ/BiG-CZ-Portal/blob/master/ClimatologyGriddedData.md#full-resolution-styled-pngs-intended-for-tiling-or-use-as-is

We need to tile these and load them into S3 for visualization in BiG CZ.

rajadain:
Commenting to subscribe to this thread and get email notifications.

mmcfarland:
Note that these are single, full-geography images per month.  I'm not sure if they are georeferenced, but it may be easier to generate tiles from the already ingested tiffs that represent this data.  If so, we'll need to match (or confirm a change) in the color ramps that are in the detailed json of the above link.

rajadain:
This is one of the images:

![image](http://data.nanoos.org/files/cz/mapoverlays/ppt_01.png)

It's pretty low-res (1464 x 647, so less than a non-retina MacBook screen). Are these the kind of images, or are there more sophisticated editions out there?

rajadain:
Found that image from the link above.

mmcfarland:
No, that is representative of the PNGs they've provided, they are explicitly not pyramided tiles.  

We can't evaluate how they were used in the existing application (http://portal.bigcz.org/Explorer) because they don't appear to be available any more.  We either reach out (@ajrobbins) to see if the overlays can be restored for reference or we just generate the tiles ourselves.  Given how coarse the tiffs are (I think ~1km or more cell size), we wouldn't need to create many zoom levels and just stretch out when the numbers get high.

ajrobbins:
@emiliom the pyramided tiles don't appear to be in use on portal.bigcz.org. Would it be possible to restore those for our reference? If it's a high effort, we'll go ahead and generate tiles from the raw tiffs.

emiliom:
@ajrobbins, the overlays were never deployed on http://portal.bigcz.org proper. But they were deployed in other, similar applications, including the Monitor-My-Watershed DRB pilot at http://www.wikiwatershed-vs.org. Look under the "Remote Sensing" button. Also, if it makes any difference (I doubt it), here's another application where they're deployed (under "Remote Sensing" and "Models"): http://viz.criticalzone.org/Christina

@lsetiawan can provide more information about the tiles, if needed.

[...]

PR #2182, opened 2017-08-24
Add scripts to process climate rasters

mmcfarland:
## Overview

Mean precipitation and mean temperature, given in monthly raster format, need to be processed into overlay tiles.  The scripts added here combine code provided from our partners at the APL to handle normalizing and color mapping, with custom code necessary to generate colored sources and tiles.

Connects #2122 
Connects #1976 

### Demo

During PR review, tiles to zoom level 10 can be found on my workstation: 
http://santacruz.internal.azavea.com:9000/

Additionally, sample layers have been loaded into an MMW instance (activate via "Coverages").  This is not the way they will be represented, ultimately, but shows how they look in the app.  This goes to z10, and then stretches to 18.  Since the cells are coarse (~4km), they don't get very pixilated like nlcd would (which goes to 13).  This saves us a lot of bandwidth, storage and processing time and I think still provides an adequate quality at closer zoom levels.
http://santacruz.internal.azavea.com:9000/

screenshot from 2017-08-24 08 46 38

screenshot from 2017-08-24 08 46 20

### Notes

The scripting to generate these tiles is complex enough to warrant a comprehensively codified set of instructions and scripts, but should be run infrequently enough as to not warrant a more automated/sophisticated version.  Therefore, I chose not to extract out options to the command line or fortify the scripts to be run from a single command.  I think it strikes the right balance between repeatability and value.  The zoom levels to generate are reduced in the saved version to allow testing; in practice, we will likely generate to a higher zoom level, but it takes longer to run. 

## Testing Instructions

Follow the instructions in the included README to ensure they generate the tiles as expected.

### DIFF
--- a/scripts/data/climate/colorize.py
+++ b/scripts/data/climate/colorize.py
@@ -1,0 +1,205 @@
+from __future__ import division
+from __future__ import print_function
+
+import os
+import sys
+import rasterio
+import numpy as np
+import matplotlib as mpl
+
+from copy import copy
+
+# These configuration settings were developed by staff at APL to produce
+# visualizations consistent with their domain knowledge
+datasets = {
+    'tmean': {
+        'colors': [(46, 0, 103), (141, 20, 255), (165, 15, 245),
+                   (189, 10, 235), (213, 5, 225), (238, 0, 215),
+                   (178, 8, 225), (119, 16, 235), (59, 25, 245),
+                   (0, 34, 255), (0, 77, 199), (0, 121, 143),
+                   (0, 164, 86), (0, 208, 31), (59, 211, 23),
+                   (117, 214, 15), (176, 217, 7), (236, 221, 0),
+                   (240, 165, 0), (245, 110, 2), (250, 55, 2),
+                   (255, 0, 4), (87, 0, 16)],
+        'position': [0., 0.16666667, 0.2, 0.23333333, 0.26666667,
+                     0.3, 0.33333333, 0.36666667, 0.4, 0.43333333,
+                     0.46666667, 0.5, 0.53333333, 0.56666667, 0.6,
+                     0.63333333, 0.66666667, 0.7, 0.73333333, 0.76666667,
+                     0.8, 0.83333333, 1.],
+    },
+    'ppt': {
+        'colors': [(230, 111, 0),
+                   (246, 155, 58),
+                   (62, 178, 189),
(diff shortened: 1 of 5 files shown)


scripts/data/climate/colorize.py
```diff
@@ -0,0 +1,205 @@
...
+            'min': 0,
+            'max': 500
+        },
+        'evi': {
+            'min': 0,
+            'max': 10000
+        }
+    }
+
+    if var == 'tmean':
```
lsetiawan: I've tweaked this a little bit. Please see https://github.com/WikiWatershed/model-my-watershed/issues/2122#issuecomment-324168504

scripts/data/climate/readme.md
```diff
@@ -0,0 +1,55 @@
...
+
+5. Run the script, it assumes sources are in `./source`
+
+```bash
+python colorize.py
+```
+
+6. Check `./output` and verify all 24 climate rasters have had a color version generated.  You can spot check against:
+* http://data.nanoos.org/files/cz/mapoverlays/ppt_01.png
+* http://data.nanoos.org/files/cz/mapoverlays/tmean.png
```
lsetiawan: This should probably be http://data.nanoos.org/files/cz/mapoverlays/tmean_01.png

lsetiawan:
Apologies for the extra comment. I hope that didn't change things too much. Thanks @mmcfarland.

scripts/data/climate/colorize.py
```diff
@@ -0,0 +1,205 @@
...
+        cdict['green'].append((pos, color[1], color[1]))
+        cdict['blue'].append((pos, color[2], color[2]))
+
+    cmap = mpl.colors.LinearSegmentedColormap('my_colormap', cdict, 256)
+    return cmap
+
+
+def normalize_to_rgba(var, arr, colormap):
+    """
+    ** This code was lightly edited for clarity and correnctness from:
```
rajadain: Typo: `correnctness` → `correctness`

caseycesari:
Not sure how much more similar work there is to do, but consider adding a `data-processing` container, like the one we added on Historic Streets.

mmcfarland:
Thanks @rajadain.  I fixed up the typos via 240314e.  Agree about tweaks, should have the same effect - definitely can revisit if we add more output.

Going to squash and merge.

Issue #2180, opened 2017-08-23
Compare View Revamp: Safari Styling


arottersman:
- [ ] On MacBook 13", the tabs overlap with "Scenarios" on Safari 
<img width="459" alt="screen shot 2017-08-23 at 3 21 38 pm" src="https://user-images.githubusercontent.com/7633670/29634554-3fff4a74-8818-11e7-9a4f-19b38ca1dc64.png">

arottersman:
This only happens on Safari 9.1

Issue #2120, opened 2017-08-04
Compare View Revamp: Make Background Darken


rajadain:
Currently because the Compare Window has no border, the white bleeds in to the background and it looks almost broken:

![image](https://user-images.githubusercontent.com/1430060/28988076-2b43f4bc-793c-11e7-9c22-10ed17b5b279.png)

The background should darken like it does for all other modals, as shown in the wireframe: https://marvelapp.com/8hi9jg9/screen/29166915

jfrankl:
I was thinking we could use the same modal backdrop class as found in the login modal. But I'll leave it up to your judgement. Let me know if I can help.

PR #2191, opened 2017-08-25
Compare View Revamp: Darken Background

rajadain:
## Overview

Extends the Compare View window from ModalBaseView, so that it behaves like a modal:

 * Background darkens
 * Background cannot be interacted with (clicking / scrolling)
 * Modal fades in and fades out
 * Clicking the close button ✖️ or hitting <kbd>Esc</kbd> closes the Compare View

Connects #2120

### Notes

I do not observe #2180 on this branch. Perhaps this incidentally fixes that too?

As a result of the modal styling, the Compare View modal reaches all the way to the bottom even when the content stops earlier, like in the table mode:

image

which I think this is fine.

Issues for IE11 still persist. I will add new screenshots to #2098.

## Testing Instructions

 * Check out this branch. Run `bundle`.
 * Get to the modeling stage with more than one scenario
 * Open the Compare View. Ensure it fades in, and the background darkens.
 * Ensure that you cannot click or scroll or otherwise interact with the background as long as the Compare View is open
 * Ensure clicking the ✖️  button closes the Compare View and it fades out. Ensure that the `#compare` div is emptied out on the page.
 * Open the Compare View again. Ensure that hitting <kbd>Esc</kbd> closes it as well.
 * Test existing modals to ensure they still work correctly. Try renaming a scenario (InputView), making a project public (ConfirmView), sharing a project (ShareView), drawing a very large shape (AlertView), and the observation tab (PlotView).

### DIFF
--- a/src/mmw/js/src/compare/views.js
+++ b/src/mmw/js/src/compare/views.js
@@ -7,6 +7,7 @@ var _ = require('lodash'),
     coreModels = require('../core/models'),
     coreViews = require('../core/views'),
     chart = require('../core/chart.js'),
+    modalViews = require('../core/modals/views'),
     models = require('./models'),
     modelingModels = require('../modeling/models'),
     modelingViews = require('../modeling/views'),
@@ -27,7 +28,7 @@ var _ = require('lodash'),
     compareModificationsPopoverTmpl = require('./templates/compareModificationsPopover.html'),
     compareDescriptionPopoverTmpl = require('./templates/compareDescriptionPopover.html');
 
-var CompareWindow2 = Marionette.LayoutView.extend({
+var CompareWindow2 = modalViews.ModalBaseView.extend({
     template: compareWindow2Tmpl,
 
     id: 'compare-new',
@@ -36,9 +37,9 @@ var CompareWindow2 = Marionette.LayoutView.extend({
         closeButton: '.compare-close > button',
     },
 
-    events: {
-        'click @ui.closeButton': 'closeView',
-    },
+    events: _.defaults({
+        'click @ui.closeButton': 'hide',
+    }, modalViews.ModalBaseView.prototype.events),
 
     modelEvents: {
         'change:mode': 'showSectionsView',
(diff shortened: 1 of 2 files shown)


rajadain:
@jfrankl Any feedback on the modal height / table mode issue above? Do we want to shorten it or leave it tall?

jfrankl:
Would be great to shorten it. Thanks for checking in.

rajadain:
Fixup makes it so that the modal height always fits the content.



Thanks for the feedback! Will squash and merge shortly.

mmcfarland:
I'm also not convinced that having the compare modal height change based on the tab/contents is the right call.  I'd like to revisit that decision if consensus is that it doesn't look good.

ajrobbins:
I agree with @mmcfarland - seeing the gif, it looks a little weird when it changes on selection of another view. I'd prefer if the modal height was fixed and scrolled if the content was too long. @jfrankl ?

jfrankl:
I agree, it's awkward. Let me explore options.

jfrankl:
Let's revert back to the original screen capture where the height of the modal is always full height.

I experimented with other ideas and asked for feedback from the design team. We agreed that the change in height felt awkward, but also had concerns about setting a fixed height for the modal, because we found that made it unclear that the user could scroll when there is content overflow. We think this is the best solution, though not without flaws.

rajadain:
Thanks for looking in to this @jfrankl. I'll drop the change and merge this.

rajadain:
For the record, the diff to shorten the height of the modal is this:

```diff
diff --git a/src/mmw/sass/pages/_compare.scss b/src/mmw/sass/pages/_compare.scss
index 4e07230..48098e5 100644
--- a/src/mmw/sass/pages/_compare.scss
+++ b/src/mmw/sass/pages/_compare.scss
@@ -290,6 +290,7 @@ $compare-chart-table-height: calc(100vh - #{$height-lg} - #{$compare-footer-heig
   background-color: #fff;
   position: fixed;
   top: 0;
+  bottom: auto;
   left: 0;
   right: 0;
   z-index: 8000;
```

Issue #2075, opened 2017-07-24
Compare View Revamp: Wire Up Navigation


rajadain:
Add the back and forward button functionality that slides scenarios in and out of view https://marvelapp.com/8hi9jg9/screen/29125397

rajadain:
Depends on #2072

jtarricone:
I recall there being some uncertainty about this behavior. Should the arrows just slide the oversized element around, or should they push/pop the last/first scenario to shift the lift itself?

ajrobbins:
@jfrankl says that the arrows should just move you over one scenario. The compare-scenario-row-content class should move everything in a uniform way.

jtarricone:
To it's just the *ordering* of scenarios and related stuff that changes? Such that the labels on the chart axis are always visible, for example?

rajadain:
We basically just need to edit the `margin-left` value of the `.compare-scenario-row-content`, as shown here:

![2017-08-23 14 55 31](https://user-images.githubusercontent.com/1430060/29633106-2e96076e-8813-11e7-8891-8d6adc70d644.gif)

rajadain:
The charts may need some more thought, as we'd like to keep the axis visible.

kellyi:
I think @jtarricone mentioned this initially but I wonder whether it'd make sense to shift the list of visible scenario data rather than changing the CSS. The chart could then render a constant number of bars & we wouldn't have to worry about the y axes.

rajadain:
It wouldn't communicate the "movement" though, achieved by something like this:

```css
transition: all 0.3s ease-in-out;
```

rajadain:
The chart axis should always be visible.

jtarricone:
I'm not sure how the chart axis could be visible if the chart if being moved. I was trying it out on a chart and the chart are doesn't seem to respond to just changing the left margin. There's probably some way of doing it but it could be pretty brittle.

jtarricone:
I guess for this pass I'll just use the left margin, which will obscure the y-axis of the charts but otherwise should implement the desired effect.

jtarricone:
One more question: should the directional shift be on a "per-click" basis, or should a continuous click stick?

Issue #2192, opened 2017-08-25
Compare View Revamp: IE11 Chart Issues


rajadain:
Some charts don't render in IE11, even though the data is present:

![2017-08-25 09 52 15](https://user-images.githubusercontent.com/1430060/29717228-62177608-897c-11e7-86a2-61fa98f05d66.gif)

The TR-55 Runoff singular charts work correctly, but the Runoff stacked chart and Water Quality singular charts do not work.

PR #2195, opened 2017-08-25
Implement Compare View Navigation

jtarricone:
## Overview

This PR implements the sliding navigation behavior for the revamped compare view. The navigation only has an effect when there are enough items in the view to be hidden (that is, when the width of the item container is larger than the modal component), otherwise the navigation buttons do nothing. The view may be moved leftwards only until about half the viewable space is empty, and rightwards until the first item is in view.

Connects #2075 

### Demo
No scrolling (not enough items) -->
recording 21

Scrolling -->
recording 22

### Notes

The limit on the leftwards scrolling was a judgment call, but it may be easily revised to some other value.

## Testing Instructions

* rebundle the app and open it
* set an area of interest and proceed to the model view
* check that, for smaller numbers of scenarios, the compare view doesn't change when the navigation buttons are clicked
* check that, for larger numbers of scenarios, the compare view moves as described, with the chart and the header items moving together.

### DIFF
--- a/src/mmw/js/src/compare/models.js
+++ b/src/mmw/js/src/compare/models.js
@@ -6,7 +6,9 @@ var _ = require('lodash'),
     ControlsCollection = require('../modeling/models').ModelPackageControlsCollection;
 
 var CHART = 'chart',
-    TABLE = 'table';
+    TABLE = 'table',
+    CHART_AXIS_WIDTH = 82,
+    COMPARE_COLUMN_WIDTH = 134;
 
 var ChartRowModel = Backbone.Model.extend({
     defaults: {
@@ -225,5 +227,7 @@ module.exports = {
     constants: {
         CHART: CHART,
         TABLE: TABLE,
+        CHART_AXIS_WIDTH: CHART_AXIS_WIDTH,
+        COMPARE_COLUMN_WIDTH: COMPARE_COLUMN_WIDTH,
     },
 };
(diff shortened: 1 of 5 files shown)


ajrobbins:
Tagging @jfrankl for review. I believe the left axis disappearing on scroll will be fixed in a later card, right @jtarricone ?

jfrankl:
This looks good overall.

It seems like a detail on the arrow buttons is missing. The buttons should appear grey when they are inactive and green when active. Active means there is room to scroll in that direction. After the user clicks on the right arrow until they are at the last scenario, that arrow should turn to grey to suggest that it can't be clicked.

rajadain:
Added minor tweaks so that the axis is now always visible and overflow charts are hidden:



There's a bug where scrolling past the end causes a crash. Will fix that and the button  behavior mentioned above.

rajadain:
Had to take it in a slightly different direction (relying more on the Backbone model than dynamic width measurements), but I think I've ironed out all the edge cases now:



It now:

 * Highlights the buttons correctly
 * Maintains chart and scenario sync at all times
 * Preserves chart state when switching between tabs
 * Preserves chart state after re-render or update, such as when receiving new values from server
 * Preserver table state and keeps it in sync
 * Has a configurable variable to set minimum number of scenarios that must always be visible. Currently it is set to 3.

kellyi:
Taking a look at this.

kellyi:
Looks good to me. Everything works well in Safari, FF, and Chrome! Moving through the scenarios works in IE11, too, although the charts aren't rendering properly at the moment (we've got #2192 to address that.)

rajadain:
Thanks for taking a look!

Issue #2133, opened 2017-08-09
Filter Out Some Results from CUAHSI by Default


rajadain:
To show only the most relevant results, we should hide certain elements (results such as NLDAS_FORA, NLDAS_NOAH) unless the user explicitly wants to show them. Because these gridded results are always up to date they show up at the top, but might not be the most relevant for most users.

@aufdenkampe / @emiliom to provide the final set of strings to use for filtering.

A simple checkbox for these items can be added here:

![image](https://user-images.githubusercontent.com/1430060/29145883-d23636c4-7d2c-11e7-8db8-f3a5358f4aff.png)

But there's a larger conversation to have (on another forum) about how we display global and tab-specific filters. /cc @jfrankl @ajrobbins

ajrobbins:
I spoke to @jfrankl about this, and he's working on a design to accomodate a few catalog-specific filters including this one.

emiliom:
I'm adding here services I've identified "manually" so far, to be filtered out:
- [NLDAS_FORA](http://hiscentral.cuahsi.org/pub_network.aspx?n=267)
- [NLDAS_NOAH](http://hiscentral.cuahsi.org/pub_network.aspx?n=274)
- [TRMM_3B42_7](http://hiscentral.cuahsi.org/pub_network.aspx?n=479)

**Note that the simplest, initial filter we should focus on is the "gridded" services.** The 3 services listed are all gridded services.

No action needed at this time. This is a reference for myself (and Anthony, I guess). We'll contact CUAHSI WDC for a definitive list of services or mechanism to filter them out more robustly.

aufdenkampe:
Thanks for follow up on this!

ajrobbins:
To be clear @emiliom, these services would be filtered out by the user in the UI? Or we would filter them out on the backend and never show them to the user? As I recall, we discussed creating a "gridded services" filter in the UI for WDC that would enable the user to filter *for* this type of service. Please correct me if that understanding is incorrect.

emiliom:
> To be clear @emiliom, these services would be filtered out by the user in the UI? Or we would filter them out on the backend and never show them to the user?

[...]

PR #2198, opened 2017-08-28
BiG-CZ Add Search Option Support, Filter CUAHSI Results

rajadain:
## Overview

Filters out "gridded" results, as described in the original issue, by default when fetching CUAHSI results. In cases when the gridded results may be a lot, removing them can potentially speed up the network request, thus we prefer filtering in the middleware rather than the client side. The main use case (as described [here](https://github.com/WikiWatershed/model-my-watershed/issues/2133#issuecomment-325013206)) is that users will rarely want gridded services, so we prioritize the initial request over the switching performance (the former slower and the latter faster in a client-side implementation).

To communicate complex filters such as this to the middleware, I've added an `options` query parameter, which takes a list of string flags. Their presence can enable arbitrary behavior defined on each catalog.

Currently there is only one option: `gridded` supported by the `cuahsi` catalog. If it is absent, we exclude certain services from the back-end request which delay the entire payload. If it is absent, no filtering is done.

On the front-end, we add a `has_filters` flag to the Catalog model, which shows a special filters div if present. We also add an `options` array, to store all enabled flags. For CUAHSI, this div includes an "Include Gridded" checkbox, checking which enables the `gridded` flag and fires a search request.

Connects #2133 

### Notes

The supported flags for each catalog should probably be documented somewhere. I'm not sure if this should be in a separate README, in code comments, or in a special "discovery" API endpoint. Something to think about when we add Swagger support and make the API public down the line.

The front-end implementation of the checkbox is also rather slapdash, because I had burned more time than I'd estimated on this. Since we don't know what kinds of filters we may have in the future (text, date, range, numeric, option buttons, more checkboxes, etc) so instead of attempting a generalization here, I'm deferring until we have more information.

## Testing Instructions

 * Go to [:8000/?bigcz](http://localhost:8000/?bigcz), and perform a search around Philadelphia
 * Go to the CUAHSI tab. Ensure there are no NLDAS_FORA or NLDAS_NOAH results
 * Check the "Include Gridded" box. Ensure there is a new search performed and that those results are now included
 * Ensure the other two tabs work as before

### DIFF
--- a/src/mmw/apps/bigcz/clients/cuahsi/search.py
+++ b/src/mmw/apps/bigcz/clients/cuahsi/search.py
@@ -29,8 +29,33 @@ DATE_MIN = date(1900, 1, 1)
 DATE_MAX = date(2100, 1, 1)
 DATE_FORMAT = '%m/%d/%Y'
 
+GRIDDED = [
+    'NWS-WGRFC_Hourly_MPE',
+    'NWS_WGRFC_Daily_MPE_Recent_Values',
+    'LMRFC_Data',
+    'GLDAS_NOAH',
+    'NLDAS_FORA',
+    'NLDAS_NOAH',
+    'TRMM_3B42_7',
+]
 
 client = Client(CATALOG_URL, timeout=settings.BIGCZ_CLIENT_TIMEOUT)
+
+
+def filter_networkIDs(services, gridded=False):
+    """
+    Transforms list of services to list of ServiceIDs, with respect to
+    given options.
+
+    If gridded=False, then GRIDDED services will not be included.
+
+    If no filters apply, we return an empty list to disable filtering.
+    """
+    if not gridded:
+        return [str(s.ServiceID) for s in services
+                if s.NetworkName not in GRIDDED]
+
+    return []
 
 
 def parse_geom(record):
@@ -173,7 +198,7 @@ def get_services_in_box(box):
         return []
 
 
-def get_series_catalog_in_box(box, from_date, to_date):
+def get_series_catalog_in_box(box, from_date, to_date, networkIDs):
     from_date = from_date or DATE_MIN
     to_date = to_date or DATE_MAX
 
@@ -183,7 +208,7 @@ def get_series_catalog_in_box(box, from_date, to_date):
                           ymin=box.ymin,
                           ymax=box.ymax,
                           conceptKeyword='',
-                          networkIDs='',
+                          networkIDs=','.join(networkIDs),
                           beginDate=from_date.strftime(DATE_FORMAT),
                           endDate=to_date.strftime(DATE_FORMAT))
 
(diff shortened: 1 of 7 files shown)


caseycesari:
I'm not noticing much of a performance difference between including/not including the "gridded" services. The AoI below has 995 WDC results when "gridded" services are not included, and 1005 when they are included. Do you know if this is expected, or should the filter result in a bigger difference?

screen shot 2017-08-29 at 9 16 08 am

src/mmw/apps/bigcz/clients/cuahsi/search.py
```diff
@@ -29,10 +29,29 @@
...
 client = Client(CATALOG_URL, timeout=settings.BIGCZ_CLIENT_TIMEOUT)
 
 
+def get_networkIDs(services, gridded=False):
+    if not gridded:
+        return [str(s.ServiceID) for s in services
+                if s.NetworkName not in GRIDDED]
+
+    # If we don't want to filter, send nothing instead of everything
+    # Sending all networkIDs returns 0 results in GetSeriesCatalogForBox2
```
caseycesari: I've read this a bunch of times, and I can't quite get a handle on it. It seems like sending all networkIDs is filtering, which is why 0 results are returned?

src/mmw/js/src/data_catalog/models.js
```diff
@@ -24,6 +24,8 @@ var Catalog = Backbone.Model.extend({
         active: false,
         results: null, // Results collection
         resultCount: 0,
+        has_filters: false, // If local filters apply
+        options: [],        // Array of String Options
```
caseycesari: Instead of adding an `options` attribute, consider subclassing the Catalog model for WDC and adding a `gridded` attribute. Historically, array attributes have [caused problems](https://stevenschobert.com/2013/01/03/arrays-and-objects-in-backbone-models) with Backbone models. I feel like the `addOption` and the `removeOptions` are somewhat duplicative of built-in Backbone methods. Also, none of the other catalogs have options yet, I'm not sure that this approach is necessary yet.

rajadain:
> I'm not noticing much of a performance difference between including/not including the "gridded" services. I think this is because there isn't much of a difference in the number of results.

This does seem to be the case now, although I remember there being a lot more of them. Perhaps the results are different now, or Stroud has been testing with other areas of interest.

src/mmw/apps/bigcz/clients/cuahsi/search.py
rajadain: This comment should probably be reworded, but the behavior _is_ counterintuitive.

Assume that there are 5 networkIDs total: `[a, b, c, d, e]`. If we make a request with a subset, like `[a, b, c]` we get results with just those networkIDs. If we make a request with an empty list, `[]`, we get results with all networkIDs. If we make a request with the full list `[a, b, c, d, e]`, we get 0 results.

I will reword this to read easier.

rajadain:
Pushed some fixups that should address the comments.

caseycesari:
Also just a reminder to squash fixups.

rajadain:
Thanks for taking a look! Squashed, merging now.

Issue #2018, opened 2017-07-10
Implement climate overlay sliders from wireframes


ajrobbins:
We've ingested monthly mean temperature and precipitation rasters for analysis in BiG CZ, and will be tiling them for visualization in #1976. This card is to wire up a slider that accompanies the new Temperature and Precipitation overlays in the layer selector. Each overlay actually has 12 overlays for display, one for each month. The slider enables the user to toggle between months, changing the overlay displayed.

See: https://marvelapp.com/5a66caa/screen/29342973

(slides 15-16)

rajadain:
Depends on #1976

mmcfarland:
Tiles are available via the azavea data-hub tile endpoint:

**Precipitation**
`https://{s}.tiles.azavea.com/climate/ppt_{month}/{z}/{x}/{y}.png`

**Temperature**
`https://{s}.tiles.azavea.com/climate/tmean_{month}/{z}/{x}/{y}.png`

`{month}` is a two digit, zero padded number 01-12.

PR #2199, opened 2017-08-28
Add time period slider for monthly overlays

mmcfarland:
## Overview

A new layer type which has monthly sub-layers can now be configured and added to the existing layer picker.  This extends the UI so that an additional slider is made available to cycle through the months.
Top-Level layer settings like opacity are preserved across sub-layers when loaded.  Two new layers, Mean Temperature and Mean Precipitation have been added to the config which make use of this layer type.

Connects #2018 

### Demo

screenshot from 2017-08-28 15 17 01

### Notes

There are a few UX calls I made, such as preserving the selected month between time-enabled layers so that one can compare month-to-month, without having to adjust the sliders, between layers.  I also preserve opacity for a layer so that you don't have to re-tune it after loading new month-layers.

A few UX issues, mostly with small screens may have to be revisited in the future.  The current layer picker covers up map controls, and this extension UI makes the problem somewhat worse.  Additionally, I do not hide the time slider if the picker is minimized.

## Testing Instructions

 * After bundling, ensure that the new layers can be activated and months adjusted.  The month should correspond to the azavea datahub tiles loaded.
* Ensure that the month stays selected between layers, so they can be compared month-to-month
* Ensure opacity settings are persisted between months, as applied to the "top level" layer.
* Ensure the tile slider shows and hides under the appropriate conditions.

IE does not fire the `input` event for range inputs, so it has its own `change` event.  Ensure that IE allow updating the month label while dragging.

### DIFF
--- a/src/mmw/js/src/core/layerPicker.js
+++ b/src/mmw/js/src/core/layerPicker.js
@@ -3,6 +3,7 @@
 var $ = require('jquery'),
     Backbone = require('../../shim/backbone'),
     Marionette = require('../../shim/backbone.marionette'),
+    moment = require('moment'),
     _ = require('underscore'),
     utils = require('./utils'),
     layerPickerTmpl = require('./templates/layerPicker.html'),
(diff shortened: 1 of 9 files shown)


src/mmw/js/src/core/layerPicker.js
```diff
@@ -186,6 +276,20 @@ var LayerPickerGroupView = Marionette.LayoutView.extend({
...
+
+        this.timeSliderRegion.show(timeSlider);
+        this.timeSliderRegion.$el.show();
+    },
+
+    hideTimeSliderRegion: function() {
+        if (!this.timeSliderRegion) { return; }
+
+        this.timeSliderRegion.empty();
+        this.timeSliderRegion.$el.hide();
```
arottersman: Consider adding an `:empty` css selector in sass for `#layer-picker-slider-region` instead.

src/mmw/js/src/core/utils.js
```diff
@@ -58,41 +58,89 @@ var utils = {
...
+        selectedLayer.get('leafletLayer').setOpacity(groupOpacity);
+    },
+
+    /**
+     * Remove a layer from the map.
+     *
+     * @param {LayerModel} layer - The selected layer model to remove
+     * @param {Leaflet.Map} map - The map from which to remove
+     */
+    _removeLayer: function(layer, map) {
```
arottersman: Nice

mmcfarland:
Swapped out the explicit `hide/show` in 	bd5b330.  Still works across browsers.

Issue #2184, opened 2017-08-24
Geoprocessing API: Create home for new endpoints


arottersman:
- Determine...
  - where in the project the API should live
  - how it should interface with existing code
                - currently `/apps/modeling` houses all the the code we need to run the analyze and rwd tasks
               - Determine if it should remain there and have the new api depend on it, or if the relevent operations should be extracted into their own module for both apps to use
- Based on decisions above, add basic views for rwd and analyze tasks in the new geoprocessing-api app

arottersman:
@mmcfarland and I met and discussed  a plan of action:

_Where should the API live?_

- in its own geoprocessing-api django project

_How should we reuse code?_

A lot of ways to cut things, we talked about a few options:
- (a)  We could rewrite views in `geoprocessing-api`, import from `modeling` and leave it mostly untouched
- (b) Matt pointed out there should only be one way to call up an analyze or RWD task, so we should move the relevant views into the `geoprocessing-api`. 

We're going with (b). We'll save ourselves work if we ever have to change up the endpoints, and it's easier to keep a mental model of.

_What about the calculations we do for analyze, constructing sql etc? Should `modeling` become a house for all modeling related code, or just a place for "stuff not in the api"?_

- no clear better choice. We'll have to look at specifics for what should stay/go
- want to avoid circular dependencies
- don't want to duplicate task utils (these for the most part seem to already be in `core`)

_Also_
- We're going to build this work off of the collections API feature branch.

Issue #2200, opened 2017-08-28
Geoprocessing API: Responses should be proper json


arottersman:
## Current Behavior

- We send completed job responses like this:

```json
{
    "error": "",
    "finished": "2017-08-28T21:34:44.207Z",
    "job_uuid": "0f90ff96-92f6-44b8-86bb-076394390912",
    "result": "{\"survey\": {\"displayName\": \"Animals\", \"name\": \"animals\", \"categories\": [{\"aeu\": 0, \"typ
e\": \"Sheep\"}, {\"aeu\": 0, \"type\": \"Horses\"}, {\"aeu\": 0, \"type\": \"Turkeys\"}, {\"aeu\": 0, \"type\": \"Ch
ickens, Layers\"}, {\"aeu\": 0, \"type\": \"Cows, Beef\"}, {\"aeu\": 0, \"type\": \"Pigs/Hogs/Swine\"}, {\"aeu\": 0, 
\"type\": \"Cows, Dairy\"}, {\"aeu\": 0, \"type\": \"Chickens, Broilers\"}]}}",                                     
    "started": "2017-08-28T21:34:44.141Z",
    "status": "complete"
}
```

- Errors responses like 404s come back as html

## Desired Behavior

- Result shouldn't be stringified json. It's not what our api users will expect, and it's annoying to work with. 

```json
{
    "error": "",
    "finished": "2017-08-28T21:34:44.207Z",
    "job_uuid": "0f90ff96-92f6-44b8-86bb-076394390912",
    "result": {"survey": {"displayName": "Animals", "name": "animals", "categories": [{"aeu": 0, "typ
e": "...you get the idea" }]}},
    "started": "2017-08-28T21:34:44.141Z",
    "status": "complete"
}
```
- Error responses should come back as json

PR #2201, opened 2017-08-28
Create geoprocessing API django app

arottersman:
## Overview

* Creates `apps.geoprocessing_api`
* `/api/modeling/start/analyze/...` ---> `/api/analyze/...`
* `/api/modeling/start/rwd/` ---> `/api/rwd/`
* Add `/api/jobs/<$job_uuid>/` (imports view from `apps.modeling`)
* Updates client to use new routes
* Modifies analyze and rwd views to correctly parse json encoding

EDIT: *Part II*

- Rename not public `/api` urls to `mmw` ;  `/api/modeling` --> `/mmw/modeling`
- Modify analyze request input dbe7976
- Add location header with `/jobs` url to poll f2362b9

Connects #2184 

### Demo

~[analyze_payload.json.txt](https://github.com/WikiWatershed/model-my-watershed/files/1258521/analyze_payload.json.txt)~
[analyze_geojson.json.txt](https://github.com/WikiWatershed/model-my-watershed/files/1261917/analyze_geojson.json.txt)

```http
http POST 'http://localhost:8000/api/analyze/animals/' < analyze_geojson.json
```

```http
HTTP/1.1 200 OK
Allow: POST, OPTIONS
Connection: keep-alive
Content-Encoding: gzip
Date: Mon, 28 Aug 2017 21:34:44 GMT
Location: http://localhost:8000/api/jobs/e19e6117-4f53-423a-b475-c1bfc6582c1c/
Server: nginx/1.12.0
Transfer-Encoding: chunked
Vary: Accept-Encoding
Vary: Accept, Cookie

{
    "job": "e19e6117-4f53-423a-b475-c1bfc6582c1c",
    "status": "started"
}
```

```http
http GET 'http://localhost:8000/api/jobs/e19e6117-4f53-423a-b475-c1bfc6582c1c/'
```

```http
HTTP/1.1 200 OK
Content-Encoding: gzip
Date: Mon, 28 Aug 2017 21:35:47 GMT
Server: nginx/1.12.0
Transfer-Encoding: chunked
Vary: Accept-Encoding
Vary: Accept, Cookie

{
    "error": "",
    "finished": "2017-08-28T21:34:44.207Z",
    "job_uuid": "e19e6117-4f53-423a-b475-c1bfc6582c1c",
    "result": "{\"survey\": {\"displayName\": \"Animals\", \"name\": \"animals\", \"categories\": [{\"aeu\": 0, \"typ
e\": \"Sheep\"}, {\"aeu\": 0, \"type\": \"Horses\"}, {\"aeu\": 0, \"type\": \"Turkeys\"}, {\"aeu\": 0, \"type\": \"Ch
ickens, Layers\"}, {\"aeu\": 0, \"type\": \"Cows, Beef\"}, {\"aeu\": 0, \"type\": \"Pigs/Hogs/Swine\"}, {\"aeu\": 0, 
\"type\": \"Cows, Dairy\"}, {\"aeu\": 0, \"type\": \"Chickens, Broilers\"}]}}",                                     
    "started": "2017-08-28T21:34:44.141Z",
    "status": "complete"
}
```

## Testing Instructions

**You will need**
- `rwd-docker` running

----
**App still works?**
 * Pull this branch, `bundle`
 * Confirm you can run both rwd endpoints from draw. The request should be going to ~`api/geoprocessing`~ `/api`, and not `api/modeling`
* Analyze an AOI and confirm each of the analyze tasks still works. Their requests should also poll ~`/api/geoprocessing`~ `/api`
* Confirm TR-55 still works

**API Accepts JSON now?**
Using the json in the _Demo_ section or similarly formatted json of your choosing, exercise the API:

`http POST 'http://localhost:8000/api/analyze/animals/' < analyze_geojson.json`


src/mmw/apps/modeling/views.py
```diff
@@ -612,35 +478,33 @@ def drb_point_sources(request):
...
-    created = now()
-    job = Job.objects.create(created_at=created, result='', error='',
-                             traceback='', user=user, status='started',
-                             model_input=job_input)
-
-    success = save_job_result.s(job.id, job_input)
-    error = save_job_error.s(job.id)
+@decorators.api_view(['GET'])
+@decorators.permission_classes((AllowAny, ))
+def get_job(request, job_uuid, format=None):
```
arottersman: This got reordered in the file, so the diff looks off. It should be implemented identically to before.

mmcfarland:
Running through the functionality and things are posting and polling to the right URLs, but I'm getting this on the raster analyze steps (the vectors are coming back).
```
File "/opt/app/apps/geoprocessing_api/tasks.py", line 124, in analyze_soil
    raise Exception('[analyze_soil] {}'.format(result['error']))
Exception: [analyze_soil] Geoprocessing Error.
Details: HTTP method not allowed, supported methods: GET
```

Seems like it may be related to the underlying GT implementation.  I'll try a reprovision and see if that sorts things out for this branch - but just checking that those work correctly on your branch.

arottersman:
> Seems like it may be related to the underlying GT implementation. I'll try a reprovision and see if that sorts things out for this branch - but just checking that those work correctly on your branch.

They do. @rajadain helped me work through the same error. If you haven't destroyed and rebuilt your worker, spark jobserver is likely still running on it. You'll have to `vagrant ssh worker -c 'sudo stop spark-jobserver'`

arottersman:
> URL names (this was discussed internally); I think that api should be reserved for just this app

I'll fix that here. I liked your idea of using `mmw/` for all the other routes.

> Route names /api/geoprocessing/start/analyze/<thing> is pretty verbose. /api/analyze/<thing> or similar would seem sufficient

I'll fix that here, too

> We'll likely want to rethink how we name the input. If all we're sending is geometry, we may want to accept formal geojson and not a json document with an area_of_interest key.

I agree. This seems like it should be its own issue, but I'm going to go for it. Hopefully it will make working on the tasks to come a little cleaner.

> Errors to the api (like 404) return HTML and should return JSON

I think this could be done with #2200. I'll update the issue to that effect.

> Initial responses for submitting an async job should have a mechanism to programmatically call in for the results. We do something similar with a Location header elsewhere.

Seems straightforward enough to add here.

rajadain:
Taking a look at this.

arottersman:
> geoprocessing in the API feels heavy, as does having a geoprocessing_api app. Not sure what captures the entire intent in a shorter form, but geo and geop could be potential candidates.

Could we potentially just call it `api` (eg `/api/analyze`, `/api/rwd`? @mmcfarland are there plans for another API separate from our "geoprocessing" one?

mmcfarland:
> Could we potentially just call it api (eg /api/analyze ....

Yep, I made the same comment in my PR review above.

arottersman:
> Currently the Analyze and RWD tasks are grouped together. Consider splitting in to analyze/tasks and rwd/tasks, since there is no overlap in the code. Also, consider moving calcs to analyze/calcs to better imply the coupling there.

This is a nice idea, but for now I don't mind the flatness. Seems to go nicely with views.

arottersman:
I added three commits to address feedback. See their commit messages for details. I've also tried to keep the PR description at the top up to date.

Quick note on restructuring the expected request input: I left the rwd JSON as is for now. I'll open a card for dealing with it separately. The analyze input seemed weirder/more worthy of an expedited update.

arottersman:
Fixups are to fix failing tests.

rajadain:
Some `flake8` checks are failing:

```
/opt/app/apps/geoprocessing_api/views.py:7:1: F401 'json' imported but unused
/opt/app/apps/geoprocessing_api/views.py:170:1: E302 expected 2 blank lines, found 1
/opt/app/apps/geoprocessing_api/views.py:177:80: E501 line too long (99 > 79 characters)
/opt/app/apps/modeling/views.py:543:1: E302 expected 2 blank lines, found 1
/opt/app/apps/modeling/views.py:546:8: E111 indentation is not a multiple of four
/opt/app/apps/modeling/views.py:585:80: E501 line too long (81 > 79 characters)
/opt/app/apps/modeling/views.py:592:80: E501 line too long (85 > 79 characters)
flake8 check failed
```

src/mmw/apps/modeling/views.py
```diff
@@ -539,6 +540,27 @@ def get_layer_shape(table_code, id):
         else:
             return None
 
+def parse_area_of_interest(area_of_interest, geojson=True):
```
mmcfarland: This function could use some comments about the arguments.  It's not clear why there's a flag for `geoJson` but the function raises an error if there is invalid geojson provided, despite the flag.  If we have two workflows that use this function with different types, it may be worth modifying the calls so they are both passing a single type.  Goes for the related function below.

src/mmw/apps/geoprocessing_api/views.py
```diff
@@ -85,7 +90,9 @@ def start_analyze_soil(request, format=None):
 @decorators.permission_classes((AllowAny, ))
 def start_analyze_animals(request, format=None):
     user = request.user if request.user.is_authenticated() else None
-    area_of_interest, __ = parse_input(request.data['analyze_input'])
+
+    wkaoi_str = request.query_params.get('wkaoi', None)
```
mmcfarland: Seems strange to parse out a `wkaoi_str` if we're going to ignore the results of the lookup done in `parse_input`.  Can this be removed in all the non-raster jobs since `wkaoi_str` has a default value of `None` in the function?

src/mmw/apps/modeling/views.py
arottersman: I'll add comment to these two and the `parse_input` in the `geoprocessing_api`.

> It's not clear why there's a flag for geoJson but the function raises an error if there is invalid geojson provided, despite the flag.

The function always expects `area_of_interest` to be valid geojson. The `geojson` flag controls whether the function returns a geojson string or its `dict` representation. It comes from the original `modeling.views.parse_input` implementation; I'll see if we can take it out.

src/mmw/apps/geoprocessing_api/views.py
arottersman: In the case where no area of interest was sent, we use the `wkaoi_str` to look up the aoi, whether the job is raster or not.

Even though we don't benefit from the wkaoi caching for these jobs, I think we should still accept the wkaoi id as input. This makes our requests on the app more uniform and makes it so the user doesn't have to keep track of which jobs are raster and which aren't.

mmcfarland: I see.  I'm not advocating for changing the request shape, but the method reads like it does not utilize the argument based on it ignoring the wkaoi return.  I think an improvement might be to rename things from `parse_*` to `load_*` or similar, since it's actually fetching data, not just parsing input.  Also, the second return variable which is only occasionally used appears to either be `None` or the id passed in as an argument initially.  Seems like this can be removed (since the caller necessarily has it), so it's clear that this function _only_ returns an area of interest, regardless of what mixture of inputs is received.  It reads like it returns something special only if wkaoi is provided.

arottersman:
- Rebased and retargeted on develop
-  79478b2 Revert change to `/user` endpoints
- 0db1f1b Refactor `parse_input`

mmcfarland:
Changes look good, but looks like `/project` didn't get reverted.  App is still serving it with `mmw`, 

http://localhost:8000/mmw/project/88/scenario/168

src/mmw/mmw/urls.py
```diff
@@ -28,9 +29,10 @@
                                namespace='rest_framework')),
     url(r'^admin/', include(admin.site.urls)),
     url(r'^accounts/', include(registration.backends.default.urls)),
-    url(r'^api/bigcz/', include(apps.bigcz.urls)),
-    url(r'^api/geocode/', include(apps.geocode.urls)),
-    url(r'^api/modeling/', include(apps.modeling.urls)),
+    url(r'^mmw/bigcz/', include(apps.bigcz.urls)),
```
mmcfarland: This will require a corresponding change in `data_catalog/models.js` which is still calling `/api/bigcz/search`.  If you switch to `?bigcz=true`, the search won't work.  

On the topic, perhaps `mmw/bigcz/...` seems a little off.  I'd be open to `mmw, api, bigcz` being the high level non-html rendering routes.  But it's an internal app URL, so not critical.

arottersman:
Thanks for catching those and the rwd input!
- Was maybe a little overzealous with that `/mmw` route prefix. Changed bigcz to just `bigcz`, and fixed `/project`
- rwd now takes pure json, not string encoded. Adjusted the client to send json.

arottersman:
Thanks @mmcfarland (and @rajadain) for the multiple thorough reviews. I've squashed things down and will merge when this has built.

Issue #2152, opened 2017-08-15
Add geometry of selected result to map.


caseycesari:
Follow up to #1934 

When a result is selected, it's geometry should be added to the map.

Example:

![](https://cloud.githubusercontent.com/assets/43062/26686127/106c04ae-46ba-11e7-9faa-721e15bf1933.png)

PR #2213, opened 2017-08-30
Add geometry of selected BigCZ result to map

caseycesari:
## Overview

When a user views the details of a BigCZ result, the geometry of the
selected result is highlighted on the map, and all of the other items
on the map are removed.

Connects #2152

### Demo

bigcz4

### Notes

I initially tried using the same layer group as the highlight layer, but because of the way the events are wired up for that layer, I wasn't able to do it. The `mouseout` event of the result item in the list would fire when the user moves the mouse over detail pane, and the highlight would be cleared. 

Marker styling was based on the wireframes: https://marvelapp.com/5a66caa/screen/29866795

## Testing Instructions

- Start BiGCZ mode, draw an AoI, and make a search.
- Select different results, and verify that when you do, the geometry for that result is the only marker on the map.

### DIFF
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -44,6 +44,16 @@ var dataCatalogActiveStyle = {
     fillOpacity: 0.2
 };
 
+var dataCatalogDetailStyle = {
+    stroke: true,
+    color: 'steelblue',
+    weight: 3,
+    opacity: 1,
+    fill: true,
+    fillColor: 'steelblue',
+    fillOpacity: 0.5
+};
+
 var selectedGeocoderAreaStyle = {
     stroke: true,
     fill: true,
@@ -207,6 +217,7 @@ var MapView = Marionette.ItemView.extend({
         'change:maskLayerApplied': 'toggleMask',
         'change:dataCatalogResults': 'renderDataCatalogResults',
         'change:dataCatalogActiveResult': 'renderDataCatalogActiveResult',
+        'change:dataCatalogDetailResult': 'renderDataCatalogDetailResult',
         'change:selectedGeocoderArea': 'renderSelectedGeocoderArea',
     },
 
@@ -225,6 +236,7 @@ var MapView = Marionette.ItemView.extend({
     // L.FeatureGroup instance
     _dataCatalogResultsLayer: null,
     _dataCatalogActiveLayer: null,
+    _dataCatalogDetailLayer: null,
 
     // The shape for a selected geocoder boundary result
     // L.FeatureGroup instance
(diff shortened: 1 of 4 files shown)


rajadain:
@azavea-bot rebuild

src/mmw/js/src/core/views.js
```diff
@@ -206,7 +216,8 @@ var MapView = Marionette.ItemView.extend({
         'change:size': 'toggleMapSize',
         'change:maskLayerApplied': 'toggleMask',
         'change:dataCatalogResults': 'renderDataCatalogResults',
-        'change:dataCatalogActiveResult': 'renderDataCatalogActiveResult',
+        'change:dataCatalogActiveResult': 'renderDataCatalogResult',
+        'change:dataCatalogDetailResult': 'renderDataCatalogResult',
```
rajadain: Since they both trigger the same function, consider putting them in one line:

```js
'change:dataCatalogActiveResult change:dataCatalogDetailResult': 'renderDataCatalogResult',
```

otherwise it seems like there is a mistake in the first one, where the left and right side don't match.

src/mmw/js/src/data_catalog/views.js
```diff
@@ -129,18 +129,20 @@ var DataCatalogWindow = Marionette.LayoutView.extend({
...
-            App.map.set('dataCatalogActiveResult', null);
+            App.map.set('dataCatalogDetailResult', null);
         } else {
             this.detailsRegion.show(new ResultDetailsView({
                 model: detailResult,
                 activeCatalog: activeCatalog.id
             }));
-            App.map.set('dataCatalogActiveResult', detailResult.get('geom'));
+            App.map.set('dataCatalogResults', null);
+            App.map.set('dataCatalogDetailResult', detailResult.get('geom'));
```
rajadain: Consider combining these into one `set`:

```js
App.map.set({
  dataCatalogResults: null,
  dataCatalogDetailResult: detailResult.get('geom'),
});
```

src/mmw/js/src/core/views.js
caseycesari: Unfortunately, condensing them into one line modifies the shape of the event object. The changed attribute is no longer included in the event payload, which prevents the event handler from detecting which attribute is changing.

> otherwise it seems like there is a mistake in the first one, where the left and right side don't match.

I'm not sure I understand. Can you elaborate?

src/mmw/js/src/data_catalog/views.js
caseycesari: Thanks. Will do.

caseycesari:
Pushed a couple fixup commits to address the feedback. 

Demo:

bigcz1

Issue #1366, opened 2016-06-13
Show AoI name in Model tab/workflow


ajrobbins:
Its not intuitive to click back to the Analyze tab to see the title of the AoI you are modelling, and the Model Tab has no indication of the AoI name. We should consider adding the title to the Model tab or displaying it at somewhere on the map viewer, like we do when a user returns to draw:
![screen shot 2016-06-13 at 1 34 28 pm](https://cloud.githubusercontent.com/assets/12868342/16016965/999d0c92-316b-11e6-81b8-bb0c8d44be19.png)

ajrobbins:
Potentially worth considering in the forthcoming UI update @jfrankl

ajrobbins:
In the current UI, we show AoI name and info in Analyze:

![screen shot 2017-08-22 at 3 59 57 pm](https://user-images.githubusercontent.com/12868342/29584872-21d77e92-8753-11e7-8ff7-c8b7670d1a55.png)

But we lose this info in Model:
![screen shot 2017-08-22 at 4 00 14 pm](https://user-images.githubusercontent.com/12868342/29584890-2a797186-8753-11e7-8056-671959acba2e.png)

This is useful contextual information that we should preserve on the page at the model stage.

jfrankl:
Let's remove the "Current Conditions/Scenario" toolbar when someone switches back to Analyze from the project page and show them a sidebar that looks more similar the initial Analyze view.

In addition to bringing back the AoI and area, this removes the information specific to modeling, which I think will make it extra clear that any modifications are not represented in the Analyze content.

![image](https://user-images.githubusercontent.com/1809908/29636300-190f2ea6-881e-11e7-937a-90020d7e2009.png)

jfrankl:
In addition to the change above (remove the "Current Conditions/Scenario" toolbar when someone switches back to Analyze), we want to add the "Details" button the Project toolbar. The "Details" button would open a dropdown with information about project settings. At first, this would include the AoI and the model.

In the future, this might be how a user would change the AoI or model for their project. Note the `(Edit)` buttons in the screenshot below. However, for now, we can leave those out.

![image](https://user-images.githubusercontent.com/1809908/29676035-e96f8248-88c4-11e7-80a7-fda37a5d59fd.png)

PR #2222, opened 2017-09-01
Show Area of Interest Info In Modeling Stage

arottersman:
## Overview

We show the AoI info in "Analyze" but lose the view in Model. The user may still be interested in the AoI information at the model stage, so this PR....
- Makes the analyze view in Model look identical to the analyze view in Analyze (add aoi header, loses model controls ... keeps project header)
- Adds a "details" button to the project header that displays the aoi name and size, as well as the model name and description. There are "edit" buttons that lead the user to either -- change the AoI (same as new project button) or change the model (routes to analyze page)

[Wireframe](https://user-images.githubusercontent.com/1809908/29676258-92ed355e-88c5-11e7-916a-9393a7cebbb4.png)

Connects #1366 

### Demo

screen shot 2017-09-01 at 1 09 05 pm

screen shot 2017-09-01 at 1 09 14 pm

screen shot 2017-09-01 at 1 32 41 pm

## Notes
- I couldn't find the icon used in the wireframe (file icon with search icon inside). I chose a similar one (file icon with lines of text)

## Testing Instructions

 *  Pull and `bundle`
 * For each model, toggle between analyze and model views, and open the "details" dropdown
 * Choose an AoI with a long name (eg Petty-Island River HUC-12) and confirm the views still scroll all the way to the bottom and the details view fits all the information


kellyi:
This is working well! I tested it out in IE11, Chrome, Safari, and Firefox and it worked as described in each. Going to take a look at the code now.

src/mmw/js/src/modeling/templates/projectMenu.html
```diff
@@ -66,26 +66,28 @@
...
-                   data-toggle="tab" href="#modeling-tab-contents" role="tab"
-                   id="show-model">
-                    <div> Model </div>
-                </a>
-            </li>
-        </ul>
-    </div>
+</div>
+
+
```
kellyi: Extra whitespace here

src/mmw/js/src/core/models.js
```diff
@@ -57,9 +57,11 @@ var MapModel = Backbone.Model.extend({
     },
 
     setNoHeaderSidebarSize: function(fit, sidebarWidth) {
-        var hasSidebar = true,
-            hasProjectHeader = false;
-        this._setSizeOptions(fit, hasProjectHeader, hasSidebar, sidebarWidth);
+        this._setSizeOptions({
+            fit: fit,
```
kellyi: Just to clarify: `fit` & `sidebarWidth` must dynamic based on the screen size? If that's incorrect, we could consider just setting the default values here.

arottersman: This could probably be set up with smarter defaults, but going to defer that for a more proper refactor. `fit` is a bool that determines if the map should fit to the aoi after resize, and `sidebarWidth` is a mode ... `""` or `"wide"`

arottersman:
Thanks! Got rid of that whitespace. Going to merge.

Issue #2181, opened 2017-08-23
BiG-CZ: Map / Sidebar Interaction


rajadain:
- [x] Hovering on an element on the map should highlight it in the sidebar. If the item is outside the current view of the list, then the list should **not** scroll to it.
 - [ ] Clicking an element on the map should highlight it, and keep it highlighted. If other items are hovered over on the list, they should be highlighted as well (perhaps with a different color than the clicked item), but the clicked highlight should not go away
 - [x] Going to the details view of an item should also sticky-highlight the corresponding element on the map
 - [ ] If, at a certain zoom level, the map component is too large to be shown, we use the gold-border trick from #2109. However, as the user zooms out, if the component then becomes contained in the viewport, it should be highlighted normally.

Issue #2214, opened 2017-08-31
Upgrade Leaflet to 1.2.0


caseycesari:
We are currently using Leaflet v0.7. We should consider upgrading to 1.2.0 to get the many new features, bugfixes, and performance improvements made to the latest version.

Some things that come to mind are:

- Using panes to hide all the result markers in https://github.com/WikiWatershed/model-my-watershed/pull/2213 when a single result is selected, instead of having to remove them from the map.
- Implementing fractional zoom levels for the compare view maps to get the most out of the small area that is available.
- Taking advantage of vector rendering improvements: https://github.com/Leaflet/Leaflet/blob/master/CHANGELOG.md#vectors-api-improvements

caseycesari:
Another feature we could take advantage of:

https://github.com/WikiWatershed/model-my-watershed/blob/3d7edbd2450db857de950b28d4b22ef2bf174e00/src/mmw/js/src/core/views.js#L392-L396

caseycesari:
In the context of working on #2181, I think having a more [unified Layer API interface](https://github.com/Leaflet/Leaflet/blob/master/CHANGELOG.md#layers-api-improvements), like the one made available in 1.0.0-beta1 would make this improvements easier to implement.

PR #2225, opened 2017-09-06
BIGCZ Map/Sidebar interaction improvements

caseycesari:
## Overview

Several improvements were made to the interaction between the map and
the results sidebar. They include:

- Hovering over features on the map now highlight the feature on the map
  in the same way that hovering over the result in the list highlights the
  corresponding map feature.
- Hovering over the feature on the map will also now highlight the
  corresponding result in the list.
- For WDC, if a popup is opened for a marker, the marker will remain
  highlighted until that popup is closed. Hovering over other features
  on the map will not remove the highlight of the marker with the popup.

Connects #2181

### Demo

bigcz4

### Notes

This PR does not include the second bullet point from #2181 for CINERGI, nor the last bullet point, which is only relevant to CINERGI.

## Testing Instructions

- Make a search and load the WDC tab. 
- Test each of the features listed above and verify they work as described.

### DIFF
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -712,8 +712,9 @@ var MapView = Marionette.ItemView.extend({
         });
     },
 
-    createDataCatalogShape: function(geom) {
-        var geomId = geom && geom.properties.id,
+    createDataCatalogShape: function(result) {
+        var geom = result.get('geom'),
+            geomId = geom && geom.properties.id,
             style = dataCatalogPolygonStyle,
             pointToLayer = null;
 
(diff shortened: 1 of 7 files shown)


src/mmw/js/src/core/views.js
```diff
@@ -712,8 +712,9 @@ var MapView = Marionette.ItemView.extend({
         });
     },
 
-    createDataCatalogShape: function(geom) {
-        var geomId = geom && geom.properties.id,
+    createDataCatalogShape: function(result) {
+        var geom = result.get('geom'),
+            geomId = geom && geom.properties.id,
```
arottersman: This id is the same as `result.get('id')`. Now that we're passing the whole result around, consider removing the geojson augmentation that happens [here.](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/data_catalog/models.js#L243)

src/mmw/js/src/core/views.js
```diff
@@ -727,64 +728,88 @@ var MapView = Marionette.ItemView.extend({
...
+                layer.on('mouseover', function() {
+                    layer.setStyle(dataCatalogActiveStyle);
+                    result.set('active', true);
+                });
+
+                layer.on('mouseout', function() {
+                    if (geom.type === 'Point') {
+                        // Preserve highlight of marker if popup is open.
+                        // It will get restyled when the popup is closed.
+                        if (!layer._popup._isOpen) {
```
arottersman: The feature works but there's a console error when you hover over the point in the details view now:
screen shot 2017-09-07 at 11 34 05 am

arottersman:
also there are merge conflicts

src/mmw/js/src/core/views.js
caseycesari: Ah good catch. I'll look into this.

rajadain:
I can take a look as soon as the merge conflicts are resolved.

caseycesari:
> Also, curious as to why it was hard to do the new popover highlight treatment on CINERGI...will it not be the same as what's done for WDC?

Unfortunately, it doesn't appear so. The public and private properties are different for the WDC and CINERGI layers. I believe this is because one is constructed as GeoJSON, while the other is constructed from CircleMarkers. In Leaflet 0.7, the only way to detect if a popup is open for a particular layer is to access some of the feature's private properties ( This was resolved for v1.0 https://github.com/Leaflet/Leaflet/issues/2494). However, the WDC implementation doesn't work for CINERGI because the `layer._popup` property isn't consistently available for the CINERGI layer, like it is for the WDC layer, even when the popup is open.

Also, the fourth bullet point from #2181 is blocked until the above is implemented.

caseycesari:
Conflicts fixed. Working on the suggested changes.

src/mmw/js/src/core/views.js
caseycesari: Addressed in e2fb98d.

src/mmw/js/src/core/views.js
caseycesari: Addressed in e2fb98d.

caseycesari:
Ready for another look.

rajadain:
Taking a look now.

rajadain:
The second bullet-point, of the sidebar item staying highlighted when the popup is open, doesn't seem to be working for WDC either:

rajadain:
If the 2 bullet points depend on #2214, I'm open to merging this with what we have and making a follow up card with those two points.

caseycesari:
> The second bullet-point, of the sidebar item staying highlighted when the popup is open, doesn't seem to be working for WDC either:

Fixed and pushed. Thanks for the catch. Accidentally rebased this specific change away earlier.

caseycesari:
> We should make a follow up card for the other work post Leaflet upgrade.

Will do.

caseycesari:
Thanks.

Issue #2134, opened 2017-08-10
Compare View Revamp: Add loading indicator/spinner to compare view


ajrobbins:
When precipitation is changed, the charts have to re-render. There is currently no indication to the user that anything is happening: https://github.com/WikiWatershed/model-my-watershed/pull/2123#issuecomment-321586095

jfrankl:
Good catch. Let's model this after the proposed loading designs seen here: https://github.com/WikiWatershed/model-my-watershed/issues/1848

While model is running:

1. Add spinner
2. Change opacity of bars in bar graphs

Details in the screenshot below:

![mmw-model-running](https://user-images.githubusercontent.com/1809908/29282458-e09e085a-80f0-11e7-8765-6324dc090c34.png)

PR #2226, opened 2017-09-06
Compare View Revamp: Loading States

rajadain:
## Overview

Adds loading states to indicate that the models are being run. All results are faded, and a spinner is added at the top-right corner. Interactivity is not impacted at all. See commit messages for details.

Connects #2134 

## Testing Instructions

 * Open a project with many scenarios. Open the Compare Window.
 * Ensure that the results start out faded, and saturate once all of them stop polling.
 * Try changing the precipitation value. Ensure that the spinner behaves accordingly.

### DIFF
--- a/src/mmw/js/src/compare/models.js
+++ b/src/mmw/js/src/compare/models.js
@@ -208,6 +208,25 @@ var WindowModel = Backbone.Model.extend({
         scenarios: null, // ScenariosCollection
         tabs: null,  // TabsCollection
         visibleScenarioIndex: 0, // Index of the first visible scenario
+        polling: false,  // If any results are polling
+    },
+
+    initialize: function() {
+        var setPolling = _.bind(this.setPolling, this);
+
+        this.get('scenarios').forEach(function(scenario) {
+            scenario.get('results').on('change', setPolling);
+        });
+    },
+
+    setPolling: function() {
+        var getPolling = function(scenario) {
+                return scenario.get('results').pluck('polling');
+            },
+            scenarios = this.get('scenarios'),
+            polling = _(scenarios.map(getPolling)).flatten().some();
+
+        this.set({ polling: polling });
     },
 
     addOrReplaceInput: function(input) {
--- a/src/mmw/js/src/compare/templates/compareWindow2.html
+++ b/src/mmw/js/src/compare/templates/compareWindow2.html
@@ -30,5 +30,5 @@
         </button>
     </div>
 </div>
-<div class="compare-sections">
+<div class="compare-sections polling">
 </div>
--- a/src/mmw/js/src/compare/templates/compareWindow2.html
+++ b/src/mmw/js/src/compare/templates/compareWindow2.html
@@ -3,6 +3,7 @@
     <div class="compare-title">
         <h1>Compare</h1>
     </div>
+    <div class="spinner hidden"></div>
     <div class="compare-close">
         <button>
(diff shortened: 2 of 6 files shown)


kellyi:
Taking a look at this.

kellyi:
Working well in Chrome, FF, and Safari! In IE 11 I don't see the loading spinner -- and the stacked bar chart's not showing up.

I think the latter's a known issue, so depending on whether we want to address the loading spinner icon here or not, this looks good to me.

kellyi:
I did notice once quirk with the compare view slider which probably merits another issue if it's not the intended behavior: viz, it looks like the slider button allows you to move the scenario columns once even if they all fit into the viewport:

screen shot 2017-09-07 at 10 20 18 am

->

screen shot 2017-09-07 at 10 20 23 am

If this isn't intentional, we should make another issue to fix it.

src/mmw/js/src/compare/models.js
```diff
@@ -208,6 +208,25 @@ var WindowModel = Backbone.Model.extend({
...
+            scenario.get('results').on('change', setPolling);
+        });
+    },
+
+    setPolling: function() {
+        var getPolling = function(scenario) {
+                return scenario.get('results').pluck('polling');
+            },
+            scenarios = this.get('scenarios'),
+            polling = _.flatten(scenarios.map(getPolling)).some(_.identity);
```
kellyi: Probably fine to leave but this could also use the lodash/underscore version of `some`?

ajrobbins:
@jfrankl I'm concerned that the loading spinner itself is too subtle - I didn't even notice it for a minute in the GIF on this PR. I think the lightening of the rest of the panel is helpful, but the spinner is very small and subtle. Perhaps we could consider doing something larger, potentially positioned over the greyed-out area?

kellyi:
Going to pull the new commit down and run this again before approving.

rajadain:
Created #2236 for the sliding issue described above.

jfrankl:
Two suggestions that I think will improve the visibility of the spinner.

1. Move the spinner to the center of the header
2. Switch the spinner style to the style recommended in https://github.com/WikiWatershed/model-my-watershed/issues/1848#issuecomment-306285213 (message me for CSS)

kb1eclfcq4

Also, please make the following changes so the application matches the specifications in https://github.com/WikiWatershed/model-my-watershed/issues/2134#issuecomment-322280666:

1. Only the bars in the bar graphs or numbers inside the table should change opacity. Axis labels, headings, etc. should not change opacity.
2. Please remove the transition on the opacity change. It should feel like it's "snapping back" into place.

rajadain:
I think we should use the new spinner, and later add it to the modeling sidebar. Since this is a new component, I'm more confident about including a new visual element here. With the new wave spinner, should it also be centered?

I'll await the wave spinner CSS on Slack, will implement the rest now.

jfrankl:
Yes, I think it should also be centered.

rajadain:
Pushed up some fixups. This is what it looks like now:



I was unable to remove the opacity transition from the bar charts, because we also use it for hiding them as they slide. But they look pretty snappy to me ...

ajrobbins:
Looking good to me.

rajadain:
Great. I'll squash and merge once @jfrankl has had a chance to take a look.

rajadain:
Squashing and merging now.

rajadain:
Thanks for all the feedback and support!

Issue #2185, opened 2017-08-24
Geoprocessing API: Integrate swagger documentation


arottersman:
- We may want to use something like https://github.com/marcgibbons/django-rest-swagger, or a framework on top of swagger like [Sphinx ](https://github.com/azavea/climate-change-api/blob/1955ed487fdd538d458c9e10c6b781e945bc115a/doc/arch/adr-0004-api-documentation.rst) 
- Add documentation for the analyze and rwd endpoints

PR #2227, opened 2017-09-06
Add Swagger Documention To The API

arottersman:
## Overview

- Adds `django-rest-swagger` to the project
- Documents inputs, outputs and descriptions of each of the AOI endpoints (the `/analyze`s and `/rwd`)
- Let's not worry too much about the exact content. The intent is to 
   - make it easy for us to demo, and for the client to check on
   - provide an initial place to stick the docs

Connects #2185 

### Demo

7h3fjjgnmq

### Notes

- The example outputs make the docstrings real lengthy. I'll open an issue to look into moving them out of the implementation files, but for now we can make like @rajadain and set our editors to auto-collapse them.

- The example responses are shown as regular json, but the response currently returns stringified json. We'll fix the response type in #2200

## Testing Instructions
 * `vagrant ssh app -c 'cd /vagrant/src/mmw/requirements && sudo pip install -r development.txt' `
 * `./scripts/manage.sh collectstatic`
 * go to `localhost:8000/api/docs/`
 * Test out each endpoint.

Here's an example `analyze` aoi:

```json
{
  "type": "MultiPolygon",
  "coordinates": [[
    [
     [
      -74.97739791870116,
      40.08164651013995
    ],
    [
      -74.9734228849411,
      40.08164651013995
    ],
    [
      -74.9734228849411,
      40.08403526674243
    ],
    [
      -74.97739791870116,
      40.08403526674243
    ],
    [
      -74.97739791870116,
      40.08164651013995
    ]
    ]
 ]]
}


```

### DIFF
--- a/src/mmw/apps/geoprocessing_api/views.py
+++ b/src/mmw/apps/geoprocessing_api/views.py
@@ -28,8 +28,7 @@ def start_rwd(request, format=None):
     created = now()
     location = request.data['location']
     data_source = request.data.get('dataSource', 'drb')
-
-    snapping = request.data['snappingOn']
+    snapping = request.data.get('snappingOn', False)
 
     job = Job.objects.create(created_at=created, result='', error='',
                              traceback='', user=user, status='started')
--- a/src/mmw/apps/geoprocessing_api/views.py
+++ b/src/mmw/apps/geoprocessing_api/views.py
@@ -23,6 +23,131 @@ from apps.geoprocessing_api import tasks
 def start_rwd(request, format=None):
     """
     Starts a job to run Rapid Watershed Delineation on a point-based location.
+
+    Selects the nearest downhill point on the medium resolution flow lines of
+    either the Delaware River Basin high resolution stream network or the
+    National Hydrography Dataset (NHDplus v2). The watershed area upstream of
+    this point is automatically delineated using the 10m resolution national
+    elevation model or the 30m resolution flow direction grid.
+
+    For more information, see the
+    [technical documentation](https://wikiwatershed.org/
+    documentation/mmw-tech/#delineate-watershed).
+
+    ## Request Body
+
+    **Required**
+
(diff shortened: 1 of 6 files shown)


rajadain:
Taking a look now.

rajadain:
In addition to `sudo pip install -r requirements.txt`, I also had to run `./scripts/manage.sh collectstatic` to get the docs to show up correctly.

rajadain:
Hmm, the markdown isn't rendering correctly for me:

image

arottersman:
Thanks, I've updated the testing instructions to include the `collectstatic`.

Not sure what's up with the markdown. Maybe check if the `markdown` package installed?

rajadain:
I do seem to have it:

```bash
vagrant@app:/vagrant/src/mmw/requirements$ sudo pip list | grep -i markdown
Markdown (2.6.9)
```

rajadain:
A hard refresh fixed it. It may have been a leftover from when I loaded the page with bad JavaScript before running `collectstatic`.

rajadain:
This works well. Not sure where the `api-docs` URL is coming from:

image

src/mmw/apps/geoprocessing_api/views.py
```diff
@@ -52,6 +177,179 @@ def start_rwd(request, format=None):
...
+    </summary>
+
+    <details>
+
+          {
+            "survey": {
+               "displayName": "Land",
+               "name": "land",
+               "categories": [
+                   {
```
rajadain: Can we clean up the spacing here to be more even?

image

Since we use 4-space indentation in our code, we should try and use the same here.

arottersman: I'll normalize these before merging.

arottersman:
Thanks for reviewing! I am also not sure where the `api-docs` url comes from...

Issue #1899, opened 2017-05-18
Add WKAoI field to Project Model


rajadain:
Currently we only use WKAoI for new projects. When resuming an old project, we only hydrate the area of interest shape and name from the database, making it impossible to take advantage of the caching functionalities when resuming saved projects. Add a `wkaoi` field to the Project model and wire it through to the front-end so it is used when resuming saved projects.

mmcfarland:
To be clear, this will effect existing projects only after this change is made, correct?  _Old_ old projects' AoI provenance is lost to history.

rajadain:
Correct. And _all_ projects will still save the area of interest. When running geoprocessing, we attempt to use the WKAoI if available and valid, otherwise fallback to the area of interest. This is purely an enhancement with no loss of current functionality.

PR #2244, opened 2017-09-08
Add WKAoI field to Project Model

rajadain:
## Overview

Adds WKAoI field to project model, so when using a predefined boundary, that boundary's id is now saved to the project. Thus, when the project is reloaded at a later date, the user can still benefit from the cached results by specifying the WKAoI id. Previously, since this id was not available, reloaded used projects would trigger unnecessary fresh bouts of geoprocessing.

Connects #1899 

## Testing Instructions

 * On `develop`, log in to MMW and create and save a TR-55 and a MapShed project with well-known boundaries (HUC-12s)
 * Check out this branch, and run `manage migrate` and `bundle`
 * Create and save a new TR-55 project using a HUC-12. Ensure that the Analyze and TR-55 requests use WKAoI when they start their run.
 * Reload the saved TR-55 project, and ensure that this also uses the saved WKAoI.
 * Do the same for MapShed.
 * Test the TR-55 and MapShed projects created before this branch. Ensure that while they don't use WKAoI, they otherwise work correctly and just as before.
 * Ensure there are no errors in the console.

### DIFF
--- a/src/mmw/apps/modeling/migrations/0022_project_wkaoi.py
+++ b/src/mmw/apps/modeling/migrations/0022_project_wkaoi.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('modeling', '0021_old_scenarios'),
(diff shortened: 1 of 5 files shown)


rajadain:
Thanks for taking a look!

PR #2245, opened 2017-09-11
Show all values in a series on mouse hover

jtarricone:
## Overview

This PR adds a new implementation of series "tooltips" that appear on all series items in a chart when any member of the series is hovered over. The actual elements added are not D3 tooltips, as it ultimately proved prohibitively difficult to implement alternative tooltip rendering. Rather, the elements are SVG objects drawn programmatically in such a way that they fulfill the designed role.

Connects #2158 

### Demo

recording 26

### Notes

There's a bug/undesired rendering style on the charts showing the individual components of the combined hydrology. For bars that extend to the full height of the chart area, the tooltip element is cutoff, as shown here:

download 27

However, the Water Quality charts, which appear to be the same form of chart, do not have this issue:

download 28

After a few hours, I couldn't determine which D3 property or attribute was ultimately responsible for it. It looks like the runoff charts charting area is extended to the full possible area, whereas the water quality are not, and the clipping rules result in the former having clipped tooltip elements while the latter does not. I didn't have much luck beyond that.

## Testing Instructions

* rebundle, then open the app
* set an AoI, do a model run, then add some scenarios and open the compare view
* check that hovering over the charted items results in labels rendering for each item in that series
* check that the hovered item is made more opaque than the others

### DIFF
--- a/src/mmw/js/src/core/chart.js
+++ b/src/mmw/js/src/core/chart.js
@@ -380,8 +380,202 @@ function renderCompareMultibarChart(chartEl, name, label, colors, stacked, yMax,
         }
 
         d3.select(svg)
-          .datum(data)
-          .call(chart);
+            .datum(data)
+            .call(chart);
+
+        // filter definition for drop shadows
+        var filter =
+            d3.select(svg)
+                .selectAll("defs")
+                .append("filter")
+                .attr("id","drop-shadow")
+                .attr("height", "120%");
+
+        filter.append("feGaussianBlur")
+            .attr("in", "SourceAlpha")
+            .attr("stdDeviation", 2)
+            .attr("result", "blur");
+
+        filter.append("feOffset")
+            .attr("in", "blur")
+            .attr("dx", 2)
+            .attr("dy", 2)
+            .attr("result", "offsetBlur");
+
+        var feMerge = filter.append("feMerge");
+        feMerge.append("feMergeNode")
+            .attr("in", "offsetBlur");
+        feMerge.append("feMergeNode")
+            .attr("in", "SourceGraphic");
+
+        var chartContainerId = chartEl.id + "-chart-container";
+
+        // iterate over the series names to add the svg elements to the chart
+        _.forEach(["nv-series-0", "nv-series-1", "nv-series-2"],
+            function(el) {
+                d3.select("#" + chartEl.id).select("svg").selectAll("g." + el).each(function() {
+                    var g = d3.select(this);
+                    var parentG = d3.select(this.parentNode);
+                    parentG.attr("id", chartContainerId);
+
+                    g.selectAll(".nv-bar").each(function(bar) {
+                        var b = d3.select(this);
+                        var barWidth = b.attr("width");
+                        var bgWidth = 40.0;
+                        var bgHeight = 20.0;
+                        var bgVertOffset = -20.0;
+
+                        var dialogTickSize = 5.0;
+
+                        // add drop-shadow
+                        parentG.append("rect")
+                            .each(function () {
+                                d3.select(this).attr({
+                                    "transform": b.attr("transform"),
+                                    "y": (parseFloat(b.attr("y")) + bgVertOffset - 1),
+                                    "x": (parseFloat(b.attr("x")) + (parseFloat(barWidth) / 2) - (bgWidth / 2) - 1),
+                                    "width": bgWidth + 2,
+                                    "height": bgHeight + 2,
+                                    "stroke": 1,
+                                    "class": ("bar-value-text-bg-shadow_" + el)
+                                });
+
+                                d3.select(this).style({
+                                    "fill": "#aaaaaa",
+                                    "opacity": 0,
+                                    "fill-opacity": 0,
+                                    "filter": "url(#drop-shadow)"
+                                });
+                            });
+
+                        // add background shape
+                        parentG.append("rect")
+                            .each(function() {
+                                d3.select(this).attr({
+                                    "transform": b.attr("transform"),
+                                    "y": (parseFloat(b.attr("y")) + bgVertOffset),
+                                    "x": (parseFloat(b.attr("x")) + (parseFloat(barWidth) / 2) - (bgWidth / 2)),
+                                    "width": bgWidth,
+                                    "height": bgHeight,
+                                    "stroke": "#aaaaaa",
+                                    "stroke-width": 0.5,
+                                    "class": ("bar-value-text-bg_" + el)
+                                });
+
+                                d3.select(this).style({
+                                    "fill": "white",
+                                    "opacity": "0",
+                                    "fill-opacity": "0",
+                                    "stroke-opacity": 0.75
+                                });
+                            });
+
+                        // add tick mark under shape
+                        parentG.append("rect")
+                            .each(function() {
+                                var rotatedX = parseFloat(b.attr("x")) + parseFloat(barWidth) / 2;
+                                var rotatedY = parseFloat(b.attr("y")) + bgHeight + bgVertOffset - (dialogTickSize / 2);
+                                d3.select(this).attr({
+                                    "transform": (b.attr("transform") + " rotate(45," + rotatedX +"," + rotatedY + ")"),
+                                    "y": (parseFloat(b.attr("y")) + bgHeight + bgVertOffset - dialogTickSize),
+                                    "x": (parseFloat(b.attr("x")) + (parseFloat(barWidth) / 2)),
+                                    "width": dialogTickSize,
+                                    "height": dialogTickSize,
+                                    "class": ("bar-value-text-bg-tick_" + el)
+                                });
+
+                                d3.select(this).style({
+                                    "fill": "white",
+                                    "fill-opacity": "0",
+                                    "opacity": "0"
+                                });
+                            });
+
+                        // add the text value
+                        parentG.append("text")
+                            .each(function() {
+                                d3.select(this).text(function() {
+                                    if (bar.y === 0) {
+                                        return;
+                                    }
+                                    return parseFloat(bar.y).toFixed(2);
+                                });
+
+                                var width = this.getBBox().width;
+
+                                d3.select(this).attr({
+                                    "transform": b.attr("transform"),
+                                    "y": (parseFloat(b.attr("y")) + 15 + bgVertOffset),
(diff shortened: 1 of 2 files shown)


kellyi:
Taking a look at this.

kellyi:
If one of the charts has a 0 value for the tooltip it looks like the tooltip skips rendering the text:

screen shot 2017-09-12 at 10 47 47 am

Wonder if we might be checking the value for falsiness such that 0 would fail? If so, may be worth checking something like `if (_.isNumber(value))` instead of `if (value)`

src/mmw/js/src/core/chart.js
```diff
@@ -380,8 +380,202 @@ function renderCompareMultibarChart(chartEl, name, label, colors, stacked, yMax,
...
+                                    "fill-opacity": "0",
+                                    "opacity": "0"
+                                });
+                            });
+
+                        // add the text value
+                        parentG.append("text")
+                            .each(function() {
+                                d3.select(this).text(function() {
+                                    if (bar.y === 0) {
```
kellyi: Oh, here's where the zero values seem to get filtered out.

kellyi:
This seems to be working well so far! I tried it out in Chrome, FF, Safari, and IE11 and the tooltips generally rendered correctly aside from the runoff tooltips being cut off at the top.

Tagging @jfrankl here so he can take a look at this and also see if he has any idea why the water quality tooltips may render correctly at the top of the chart while the runoff tooltips don't. I wonder if it might be a z-index issue? I think we're generally using the same Marionette template & method, so most of the work of creating the chart would happen via d3/svg.

jtarricone:
> I wonder if it might be a z-index issue?

That's what I kept looking for, but I think D3 handles z-index in a pretty naive way by just drawing things in the order their procedural order. After staring at the charts and the associated markup I noticed that the water quality charts, as well as the combined hydrology chart, have a margin between the bars and the top of the rendered SVG, whereas the runoff charts don't appear to have one, so the tooltips at the top are clipped. I couldn't figure out *why* though, since the D3 instantiation appears to be the same.

jtarricone:
>Interesting -- not sure if this is still applicable but originally I was trying to make the chart y axis extend vertically further than the tops of the runoff bars by setting the tweaking the ydomain values to be like 120% of the max value from the data. I think we may have dropped that at some point, but maybe we brought something like it back in for the water quality charts?

That would probably result in this problem, but in that case would there be an obvious difference in the heights of the y-axis and of the tallest bar? That doesn't appear to be the case for the water quality charts. I can check the water quality chart instantiation to see if there's some padding added, which I don't recall specifically looking for.

kellyi:
Is this ready for another look?

rajadain:
Any movement on this?

jtarricone:
I was waiting on @jfrankl , but if design changes aren't required (yet) then I think this is ready to go. I didn't have any luck fixing the cropping issue, it's still a bit mysterious actually.

rajadain:
Taking a look at this.

src/mmw/js/src/core/chart.js
rajadain: If these tooltips can't be hidden, I'd rather they say "0.00" than be empty

image

src/mmw/js/src/core/chart.js
```diff
@@ -380,8 +380,202 @@ function renderCompareMultibarChart(chartEl, name, label, colors, stacked, yMax,
...
+                        parentG.append("rect")
+                            .each(function () {
+                                d3.select(this).attr({
+                                    "transform": b.attr("transform"),
+                                    "y": (parseFloat(b.attr("y")) + bgVertOffset - 1),
+                                    "x": (parseFloat(b.attr("x")) + (parseFloat(barWidth) / 2) - (bgWidth / 2) - 1),
+                                    "width": bgWidth + 2,
+                                    "height": bgHeight + 2,
+                                    "stroke": 1,
+                                    "class": ("bar-value-text-bg-shadow_" + el)
```
rajadain: Instead of appending with an underscore, prefer appending with a space so these elements can be targeted via CSS like `.bar-value-test-bg-shadow`

rajadain:
```diff
diff --git a/src/mmw/js/src/core/chart.js b/src/mmw/js/src/core/chart.js
index a462dbe..508e071 100644
--- a/src/mmw/js/src/core/chart.js
+++ b/src/mmw/js/src/core/chart.js
@@ -437,7 +437,7 @@ function renderCompareMultibarChart(chartEl, name, label, colors, stacked, yMax,
                                     "width": bgWidth + 2,
                                     "height": bgHeight + 2,
                                     "stroke": 1,
-                                    "class": ("bar-value-text-bg-shadow_" + el)
+                                    "class": ("bar-value-text-bg-shadow " + el)
                                 });
 
                                 d3.select(this).style({
@@ -459,7 +459,7 @@ function renderCompareMultibarChart(chartEl, name, label, colors, stacked, yMax,
                                     "height": bgHeight,
                                     "stroke": "#aaaaaa",
                                     "stroke-width": 0.5,
-                                    "class": ("bar-value-text-bg_" + el)
+                                    "class": ("bar-value-text-bg " + el)
                                 });
 
                                 d3.select(this).style({
@@ -481,7 +481,7 @@ function renderCompareMultibarChart(chartEl, name, label, colors, stacked, yMax,
                                     "x": (parseFloat(b.attr("x")) + (parseFloat(barWidth) / 2)),
                                     "width": dialogTickSize,
                                     "height": dialogTickSize,
-                                    "class": ("bar-value-text-bg-tick_" + el)
+                                    "class": ("bar-value-text-bg-tick " + el)
                                 });
 
                                 d3.select(this).style({
@@ -510,7 +510,7 @@ function renderCompareMultibarChart(chartEl, name, label, colors, stacked, yMax,
                                     "font-size": "0.8rem",
                                     "font-weight": "normal",
                                     "font-family": "helvetica",
-                                    "class": ("bar-value-text_" + el)
+                                    "class": ("bar-value-text " + el)
                                 });
 
                                 d3.select(this).style({
@@ -531,21 +531,21 @@ function renderCompareMultibarChart(chartEl, name, label, colors, stacked, yMax,
                     .selectAll("rect.nv-bar")
                     .on("mouseover", function () {
                         d3.select("#" + chartContainerId)
-                            .selectAll(".bar-value-text_" + el)
+                            .selectAll(".bar-value-text." + el)
                             .style("opacity", "1");
 
                         d3.select("#" + chartContainerId)
-                            .selectAll(".bar-value-text-bg-shadow_" + el)
+                            .selectAll(".bar-value-text-bg-shadow." + el)
                             .style("opacity", "0.5")
                             .style("fill-opacity", "0.5");
 
                         d3.select("#" + chartContainerId)
-                            .selectAll(".bar-value-text-bg_" + el)
+                            .selectAll(".bar-value-text-bg." + el)
                             .style("opacity", "1")
                             .style("fill-opacity", "1");
 
                         d3.select("#" + chartContainerId)
-                            .selectAll(".bar-value-text-bg-tick_" + el)
+                            .selectAll(".bar-value-text-bg-tick." + el)
                             .style("opacity", "1")
                             .style("fill-opacity", "1");
                     });
@@ -556,21 +556,21 @@ function renderCompareMultibarChart(chartEl, name, label, colors, stacked, yMax,
                     .selectAll("rect.nv-bar")
                     .on("mouseout", function () {
                         d3.select("#" + chartContainerId)
-                            .selectAll(".bar-value-text_" + el)
+                            .selectAll(".bar-value-text." + el)
                             .style("opacity", "0");
 
                         d3.select("#" + chartContainerId)
-                            .selectAll(".bar-value-text-bg-shadow_" + el)
+                            .selectAll(".bar-value-text-bg-shadow." + el)
                             .style("opacity", "0")
                             .style("fill-opacity", "0");
 
                         d3.select("#" + chartContainerId)
-                            .selectAll(".bar-value-text-bg_" + el)
+                            .selectAll(".bar-value-text-bg." + el)
                             .style("opacity", "0")
                             .style("fill-opacity", "0");
 
                         d3.select("#" + chartContainerId)
-                            .selectAll(".bar-value-text-bg-tick_" + el)
+                            .selectAll(".bar-value-text-bg-tick." + el)
                             .style("opacity", "0")
                             .style("fill-opacity", "0");
                     });
```

rajadain: We can then do something like this:

```diff
diff --git a/src/mmw/sass/pages/_compare.scss b/src/mmw/sass/pages/_compare.scss
index 3ebf563..d73574b 100644
--- a/src/mmw/sass/pages/_compare.scss
+++ b/src/mmw/sass/pages/_compare.scss
@@ -463,6 +463,13 @@ $compare-chart-table-height: calc(100vh - #{$height-lg} - #{$compare-footer-heig
     rect {
       transition: opacity 0.3s ease-in-out;
     }
+
+    .bar-value-text-bg-shadow,
+    .bar-value-text-bg,
+    .bar-value-text-bg-tick,
+    .bar-value-text {
+      pointer-events: none;
+    }
   }
 
   .compare-scenario-gradient {
```

Which allows hovering on the stacked chart in areas obscured by hidden tooltips to function correctly.

jtarricone: The reason it's so specific is that those elements have to be manipulated via D3 later on. Could D3 selection be performed (easily) with a lower degree of specificity in the elements?

jtarricone: Did that CSS get tested on the D3 charts? I observed that CSS and D3/SVG elements didn't seem to play very nice together.

rajadain: As shown in the diff, a CSS class of `abcd_wxyz` can be accessed only via `.abcd_wxyz`, but one of `abcd wxyz` can be accessed specifically with `.abcd.wxyz`, but also in other groupings with `.abcd` and `.wxyz`. So the existing selection behavior is preserved by switching to spaces and dots instead of the underscore, and enables this CSS targeting after the fact.

rajadain:
With these suggested fixes, I think this'll be good to go. We can make a follow up card that looks into the tooltips being cutoff for high values, and for removing them for 0 values.

jtarricone:
I pushed fixes but I'm having trouble getting my environment spun up, so I haven't tested yet. I'll loop back once I test it, unless you get a chance before I do.

rajadain:
Made #2278 for follow up improvements.

jtarricone:
Unless @kellyi has any objection I'll merge once the build (successfully) completes.

Issue #2151, opened 2017-08-15
WDC results are not added to map in some scenarios


caseycesari:
To reproduce:

- Draw an AoI roughly around the John Heinz Wildlife Refuge (just northwest of the airport).
![image](https://user-images.githubusercontent.com/1042475/29332982-0850c722-81d0-11e7-8f9d-c8cb2e644ff8.png)

- Search for "hydrogeology".
- Switch to the WDC tab. Note that when the search completes there should be approximately three results in the JSON response. The number of results will also be noted in the WDC tab. However, no results are shown in the list or on the map.

![image](https://user-images.githubusercontent.com/1042475/29333030-3f776a80-81d0-11e7-9220-fc3457b95639.png)

Sample JSON response:

```json
[{
	"catalog": "cuahsi",
	"api_url": null,
	"results": [{
		"id": "NWISGW:395327075164201",
		"title": "DE   510",
		"description": "The USGS National Water Information System (NWIS) provides access to millions of sites measuring streamflow, groundwater levels, and water quality. This web service provides methods for retrieving Ground Water data from NWIS. For more information about NWIS, see the NWIS home page at http://waterdata.usgs.gov/nwis",
		"author": null,
		"links": [{
			"href": "http://hiscentral.cuahsi.org/pub_network.aspx?n=8",
			"type": "service"
		}],
		"created_at": "1946-11-08T00:00:00Z",
		"updated_at": null,
		"geom": {
			"type": "Point",
			"coordinates": [-75.2779627, 39.89066835]
		},
		"details_url": null,
		"sample_mediums": "Groundwater",
		"concept_keywords": "Groundwater level",
		"service_org": "U.S. Geological Survey (USGS)",
		"service_code": "NWISGW",
		"service_url": "http://hiscentral.cuahsi.org/pub_network.aspx?n=8",
		"begin_date": "1946-11-08T00:00:00Z",
		"end_date": "1946-11-08T00:00:00Z"
	}, {
		"id": "NWISGW:395326075164501",
		"title": "DE   511",
		"description": "The USGS National Water Information System (NWIS) provides access to millions of sites measuring streamflow, groundwater levels, and water quality. This web service provides methods for retrieving Ground Water data from NWIS. For more information about NWIS, see the NWIS home page at http://waterdata.usgs.gov/nwis",
		"author": null,
		"links": [{
			"href": "http://hiscentral.cuahsi.org/pub_network.aspx?n=8",
			"type": "service"
		}],
[...]

PR #2246, opened 2017-09-11
Backend Geometry filtering

mmcfarland:
## Overview

Move complete search filtering to API.  The application was responsible for doing post processing on API search results and resulted in a quirk of result count not matching the count displayed in the list or map.  The post processing has been moved to a centralized spot in the API so that results that are returned match the expected query, including boundary filtering.  From the client perspective, the raw geometry is now `POSTed` to the API, instead of the bounding box.  The bbox is still used to perform the catalog search, but it is calculated on the backend so consumers don't need to be aware of the implementation details of the post processing.

Connects #2151 

### Demo

Compared to the screenshot in the issue, expected counts are now shown in the search.
screenshot from 2017-09-11 11 59 43

### Notes
For large, multi-paged results, it is impossible to know the true count, since we're only post-processing the results from a page at a time.  In reality, the count does not reflect the upstream catalog count minus the spatial filtering done locally, since we don't have access to all of the features.  This was the case in the previous implementation and if it becomes problematic/confusing, we may want to generalize the count to something like "1000+" when there are many records.

I also had to infer some desired behavior from the existing code, so please verify the changes make logical sense.

## Testing Instructions

* Test various queries at various level of geographies.  
* For paged results, the approximate total should still be displayed if the count is greater than that of the page size.

### DIFF
--- a/src/mmw/apps/bigcz/views.py
+++ b/src/mmw/apps/bigcz/views.py
@@ -3,19 +3,24 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import division
 
+from django.contrib.gis.geos import GEOSGeometry
 from rest_framework import decorators
 from rest_framework.exceptions import ValidationError, ParseError
 from rest_framework.permissions import AllowAny
 from rest_framework.response import Response
 
 from apps.bigcz.clients import CATALOGS
+from apps.bigcz.models import BBox
 from apps.bigcz.serializers import ResourceListSerializer
 from apps.bigcz.utils import parse_date
 
+import json
+
 
 def _do_search(request):
-    catalog = request.query_params.get('catalog')
-    page = int(request.query_params.get('page', 1))
+    params = json.loads(request.body)
+    catalog = params.get('catalog')
+    page = int(params.get('page', 1))
     request_uri = request.build_absolute_uri()
 
     if not catalog:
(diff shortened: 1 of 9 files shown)


src/mmw/apps/bigcz/views.py
```diff
@@ -3,19 +3,43 @@
...
 from rest_framework.permissions import AllowAny
 from rest_framework.response import Response
 
 from apps.bigcz.clients import CATALOGS
 from apps.bigcz.serializers import ResourceListSerializer
-from apps.bigcz.utils import parse_date
+from apps.bigcz.utils import parse_date, get_bounds, \
+    filter_aoi_intersection
+
+import json
```
rajadain: Consider moving this more generic import to the top

src/mmw/apps/bigcz/views.py
```diff
@@ -3,19 +3,43 @@
...
+from django.conf import settings
 from rest_framework import decorators
 from rest_framework.exceptions import ValidationError, ParseError
 from rest_framework.permissions import AllowAny
 from rest_framework.response import Response
 
 from apps.bigcz.clients import CATALOGS
 from apps.bigcz.serializers import ResourceListSerializer
-from apps.bigcz.utils import parse_date
+from apps.bigcz.utils import parse_date, get_bounds, \
```
rajadain: [Elsewhere](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/mapshed/tasks.py#L12) in [the code](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/geoprocessing_api/tasks.py#L19), multiline imports have been done using parentheses. Consider switching to those for consistency.

mmcfarland: Guess it depends where [you look](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/bigcz/serializers.py#L8).  I was trying to keep it consistent within the django app, and I actually prefer this syntax over the parentheses since it allow you to start the next line indented one tab instead of at the opening parens, which can lead to a long imports section.

I'll update the two imports formatting.

rajadain: One import per line does make it easier to read diffs, but modern diffs highlight changed text in the same line so I think this'll be fine. 👍

Issue #2253, opened 2017-09-13
BiG-CZ: No Climate Data for Small AoIs


rajadain:
Because the size of the raster tiles for climate data (ppt and tmean rasters) are so large, if a small AoI does not intersect with their center, we get 0 values:

![image](https://user-images.githubusercontent.com/1430060/30395791-66565036-9895-11e7-8d56-583778bdae07.png)

To reproduce: draw a 1 sq km AoI anywhere and see that there is no climate data.

One potential way to fix this may be to treat the tiles as polygons rather than pixels. Another way may be to find the nearest tile and just use its values.

Implement such a fix, and describe how it addresses the problem.

mmcfarland:
Rasterio has a flag for how to handle included cells, via centroid or intersection.  If GT exposes something similar, we should expose it as a flag based on what we know about each layer.  In practice, that would mean climate analysis would use intersection and the rest would continue to use centroid.

@echeipesh does GeoTrellis have the ability to specify the mechanism for determining which pixels to use on the fly, or is centroid baked in?

echeipesh:
GeoTrellis has this concept of point vs area and it is already applied in the layer filter logic here: https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/io/LayerFilter.scala#L139

What is the context for this question? If there is a rasterizer that walks over the resulting tile it would have similar options available to it.

PR #2257, opened 2017-09-14
Upgrade geoprocessing to support area pixels

rajadain:
## Overview

Previously, for low resolution rasters with very large pixels, when paired with small areas of interest, we would get 0 values. This was because the pixels were treated as points, and if the centroid of the pixel was outside the area of interest, there would be no intersection.

By adding a `pixelIsArea: true` parameter, we treat those pixels as polygons instead of points, and can calculate the intersection even with small areas of interest. The Precipitation and Temperature rasters are prime use cases for this, and their usage is corrected with this new setting.

Connects #2253

### Demo

For demo and other notes, see https://github.com/WikiWatershed/mmw-geoprocessing/pull/74

## Testing Instructions

 * Check out this branch
 * Reprovision Worker
 * Draw a 1 sq km area of interest. Ensure that the Climate Analyze tab shows real values and not all 0s.

### DIFF
--- a/deployment/ansible/group_vars/all
+++ b/deployment/ansible/group_vars/all
@@ -45,7 +45,7 @@ docker_options: "--storage-driver=aufs"
 geop_host: "localhost"
 geop_port: 8090
 
-geop_version: "3.0.0-beta-2"
+geop_version: "3.0.0-beta-3"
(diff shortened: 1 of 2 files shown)


rajadain:
Thanks for taking a look.

Issue #2232, opened 2017-09-07
BiG-CZ: Implement Filters Sidebar


rajadain:
Implement filters sidebar which should contain date filters (which apply to all catalogs), and any catalog specific filters (such as "Include Gridded" for CUAHSI).

@jfrankl could you please add the mockups to this issue?

jfrankl:
### Screen 1
![image](https://user-images.githubusercontent.com/1809908/30185291-d5a4556a-93ee-11e7-90bc-60b5f5af5ac7.png)

### Screen 2
![image](https://user-images.githubusercontent.com/1809908/30185315-f1a85be4-93ee-11e7-985b-8dbeeab9ffc4.png)

PR #1890, opened 2017-05-15
Revamp Sidebars + Style Touchups

arottersman:
## Overview

This PR began as a humble attempt at matching the analyze stage to the new wireframes (#1871). The existing styling applied to the sidebar, however, was so heavily nested, it was difficult move even simple pieces of text in the view. I decided to wrap #1864 in, and try to refactor the sidebar-related stylesheets. I then realized whatever I merged would possibly redo parts of the work of #1844, and definitely create even more merge conflicts for it. @jfrankl and I paired, and to merge my work, the changes on develop, and #1844.

**Wireframes**
https://marvelapp.com/gji4cg6/screen/27712989
<details>

<summary>
Here's a selection of them
</summary>

screen shot 2017-05-15 at 4 15 37 pm

screen shot 2017-05-15 at 4 15 44 pm

screen shot 2017-05-15 at 4 15 51 pm

screen shot 2017-05-15 at 4 15 57 pm

</details>

Connects #1844 
Connects #1864 
Connects #1871 

### Demo

screen shot 2017-05-15 at 3 53 48 pm

screen shot 2017-05-15 at 3 56 22 pm

screen shot 2017-05-15 at 11 20 15 am

screen shot 2017-05-15 at 11 20 39 am

screen shot 2017-05-15 at 3 57 54 pm

screen shot 2017-05-15 at 3 57 58 pm

screen shot 2017-05-15 at 3 58 09 pm

### Notes

In the wireframes, when the analyze tab is visible in the model stage, we hide the project toolbar, and show the AOI header. I'll be spinning this work off into another card.

screen shot 2017-05-15 at 4 16 06 pm

## Testing Instructions

 * Pull branch, bundle
In MMW mode
 * Move through the app, splash to model, at various viewport sizes, devices, and browsers
 * Ensure everything but the app header, scenario toolbar, project header, and splash is approaching the wireframe styling

In BigCZ mode, check that the data source button looks okay and takes you to the datasource search sidebar

PR #2258, opened 2017-09-14
Replace BiG-CZ Data Catalog Filter Header With Filter Sidebar

arottersman:
## Overview

Consolidate all the data catalog filters into a single sidebar. This will make it easier to add more filters in the future.

This is a big, single commit. I recommend looking at the changes in the data catalog models first; that's where the meat is (see commit message). The view changes should all be either (a) moving operations into the models (b) moving views from the form into the new sidebar views.

**Behavioral differences:**
- Kick off a new search any time there's a change in the date, if the there's a query string and whatever's in the date filter is valid. Before we waited for the explicit <kbd>Enter</kbd>

Connects #2232 

## Notes

This is maybe a little more subclassing than we usually do. My goal with all the refactoring was to make it easy to add new filters, and have them conform to similar interfaces.

### Demo

4stmtwevkv

## Testing Instructions

 * Pull and `./scripts/bundle.sh`. Make sure the sass compiles
 * On all major browsers test combinations of each tab + filter states + query field states

### DIFF
--- a/src/mmw/js/src/core/models.js
+++ b/src/mmw/js/src/core/models.js
@@ -82,7 +82,7 @@ var MapModel = Backbone.Model.extend({
     },
 
     setDataCatalogSize: function(fit) {
-        this.setNoHeaderSidebarSize(fit, utils.sidebarWide);
+        this.setNoHeaderSidebarSize(fit);
     },
 
     setModelSize: function(fit) {
@@ -101,6 +101,14 @@ var MapModel = Backbone.Model.extend({
         this.set('size', updatedSize);
     },
 
+    toggleSecondarySidebar: function() {
+        var sizeCopy = _.clone(this.get('size')),
+            updatedSize =_.merge(sizeCopy, {
+                hasSecondarySidebar: !sizeCopy.hasSecondarySidebar
+            });
+        this.set('size', updatedSize);
+    },
+
 // Set the sizing options for the map. `options` are...
 //      fit                  - bool, true if should fit the map to the AoI
 //      hasProjectHeader     - bool, true if the -projectheader class should
@@ -109,12 +117,9 @@ var MapModel = Backbone.Model.extend({
 //                             be on the map container
 //      hasSidebar           - bool, true if the -sidebar class should
 //                             be on the map container
-//      barWidth             - string, if matches width option (eg 'wide'),
-//                             and `hasSidebar === true`, map will add
-//                             class to container accordingly.
-//                             If option doesn't exist or is falsey, no
-//                             class will be added, and map container will
-//                             use the default sidebar size if `hasSidebar`
+//      hasSecondarySidebar  - bool, true if the -double class
+//                             should be on the map. Will on apply if `hasSidebar`
+
     _setSizeOptions: function(options) {
         this.set('size', options);
     }
--- a/src/mmw/js/src/data_catalog/models.js
+++ b/src/mmw/js/src/data_catalog/models.js
@@ -3,21 +3,110 @@
 var $ = require('jquery'),
     _ = require('lodash'),
     Backbone = require('../../shim/backbone'),
+    moment = require('moment'),
     settings = require('../core/settings');
 
 var REQUEST_TIMED_OUT_CODE = 408;
 var DESCRIPTION_MAX_LENGTH = 100;
 var PAGE_SIZE = settings.get('data_catalog_page_size');
 
-var SearchOption = Backbone.Model.extend({
+
+var FilterModel = Backbone.Model.extend({
     defaults: {
         id: '',
-        active: false,
+        type: '',
+        isValid: true
     },
+
+    reset: function() {
+        // Clear all model attributes and set back any defaults.
+        // Once we've called `.clear()` we lose the ability to detect
+        // actual changed attributes after the `.set()`.
+        // Store the prev state to check if a change event
+        // should fire
+        var prevAttr = _.clone(this.attributes);
+        this.clear({ silent: true }).set(this.defaults,
+                                         { silent: true });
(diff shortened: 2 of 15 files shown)


src/mmw/js/src/core/models.js
```diff
@@ -101,6 +101,14 @@ var MapModel = Backbone.Model.extend({
         this.set('size', updatedSize);
     },
 
+    toggleSecondarySidebar: function() {
+        var sizeCopy = _.clone(this.get('size')),
```
rajadain: Do we need to use `_.clone` here? Could we not have simply done `var size = this.get('size')`?

src/mmw/js/src/data_catalog/models.js
```diff
@@ -3,38 +3,122 @@
...
+
+    isActive: function() {
+        window.console.error("Use of unimplemented function",
+                             "FilterModel.isActive");
+        return false;
+    }
+});
+
+var SearchOption = FilterModel.extend({
+    defaults: function() {
```
rajadain: Since `FilterModel.prototype.defaults` is defined at "compile time" this doesn't need to be a dynamic function. Instead we can simply do:

```js
defaults: _.defaults({
    active: false,
}, FilterModel.prototype.defaults),
```

For example, see https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/core/modals/views.js#L90-L93

src/mmw/js/src/data_catalog/models.js
```diff
@@ -3,38 +3,122 @@
...
+
+    isActive: function() {
+        return this.get('active');
+    }
 });
 
-var SearchOptions = Backbone.Collection.extend({
-    model: SearchOption,
+var GriddedServicesFilter = SearchOption.extend({
+    defaults: function() {
```
rajadain: Same here.

src/mmw/js/src/data_catalog/models.js
```diff
@@ -3,38 +3,122 @@
...
+        return _.extend(SearchOption.prototype.defaults(), {
+            id: 'gridded',
+            type: 'checkbox',
+            label: 'Gridded Services',
+        });
+    },
+});
+
+var DateFilter = FilterModel.extend({
+    defaults: function() {
```
rajadain: Same here.

src/mmw/js/src/data_catalog/models.js
```diff
@@ -3,38 +3,122 @@
...
+
+var FilterCollection = Backbone.Collection.extend({
+    model: FilterModel,
+
+    countActive: function() {
+        var incrementIfActive = function(count, filter) {
+            return filter.isActive() ? ++count : count;
+        };
+
+        return this.reduce(incrementIfActive, 0);
```
rajadain: This can be made clearer and shorter like this:

```js
var isActive = function(x) { return x.isActive(); };

return this.filter(isActive).length;
```

src/mmw/js/src/data_catalog/models.js
```diff
@@ -81,38 +165,59 @@ var Catalog = Backbone.Model.extend({
...
-            fromDate: fromDate,
-            toDate: toDate,
         });
 
         return this.startSearch(1);
     },
 
+    isSearchValid: function() {
+        var query = this.get('query'),
+            dateFilter = this.get('filters').findWhere({ id: 'date'}),
```
rajadain: This can be made more generic, since each filter can have a `validate()` method that returns `true` or `false`:

```js
var query = this.get('query'),
    validate = function(x) { return x.validate(); },
    valid = this.get('filters').map(validate);

return query && _.every(valid);
```

For simpler filters like `GriddedService` which is just a checkbox, we can have its `validate()` always return true.

rajadain:
Clicking "Reset" runs the search again, even if it doesn't need to be run. That is, if clicking Reset multiple times, a new search is run every time, whereas it should be run only the first time.

rajadain:
Checking the GriddedServices checkbox makes the same request as not checking it.

src/mmw/js/src/core/models.js
arottersman: Same as`toggleSidebar` above, [here's our discussion](https://github.com/WikiWatershed/model-my-watershed/pull/1890#discussion_r116815092) on how we need to trigger a change event, and the `merge` oddly doesn't suffice. I'll test this out again to confirm...

rajadain:
Changing the text filter re-runs the search on only the active tab, but changing the date filter re-runs the search on all tabs. While this is fast enough (at least for now), and the tabs are independent enough, it feels inconsistent. Might be worth creating a follow up card for figuring this out. On one hand, we don't want to have stale results. On the other, triggering all the searches may be taxing.

src/mmw/js/src/core/models.js
rajadain: Ah I see. Very well, let's keep it. Good to know I'm hard headed as ever.

rajadain:
The date filter can be selected before the first search. I like that.

src/mmw/js/src/data_catalog/models.js
arottersman: Awesome. Thanks!

arottersman:
e979ba5 makes requested code-style changes. The rest address each of the bugs you found. Thanks for finding those.

arottersman:
> Changing the text filter re-runs the search on only the active tab, but changing the date filter re-runs the search on all tabs. While this is fast enough (at least for now), and the tabs are independent enough, it feels inconsistent. Might be worth creating a follow up card for figuring this out. On one hand, we don't want to have stale results. On the other, triggering all the searches may be taxing.

I agree about the inconsistency, so added a commit for this. The change I made for this is simple enough to remove if we change our minds.

Issue #2243, opened 2017-09-07
BiG-CZ: Investigate WaterML Use for Fetching Sensor Data


rajadain:
Investigate the WaterML interface used for fetching time series data of sensors, as requested in #2238.

Issue #2238, opened 2017-09-07
BiG-CZ: Fetch time series data values from WDC services via WaterML


aufdenkampe:
All services that are discoverable in the CUAHSI Water Data Center (WDC) catalog are registered to provide data values in [WaterML](https://en.wikipedia.org/wiki/WaterML) via Water On Flow (WOF) web services. We need to implement this capability to view time series data for both the BiG-CZ Portal and also for Monitor My Watershed.

There are 3 variants of WaterML (1.0, 1.1, 2.0). We will use 1.1, as it is most uniformly provided.

For Documentation from WDC, see the "Time Series Data Storage and Transmission" section of this page: https://www.cuahsi.org/data-models/for-developers​. These are the most useful PDFs from that help text:
- http://www.cuahsi.org/uploads/pages/img/WaterML1_1.zip

Here are two prior notebooks that @emiliom created demonstrating the use of the [Ulmo Python library](http://ulmo.readthedocs.io/en/latest/#) for WOF/WaterML access of both site metadata and time series data:
- https://github.com/ODM2/WOFpy/blob/master/docs/WOFpy_ulmo_tests.ipynb
- https://github.com/uwescience/Python-for-geosciences/blob/master/20170307/ulmo-Snotel-cuahsiwof.ipynb

cc: @rajadain, @emiliom, @lsetiawan

emiliom:
@rajadain, just a brief comment to help minimize future confusion, regarding this:

> There are 3 variants of WaterML (1.0, 1.1, 2.0). 

1.1 and 1.0 are very similar. But 2.0 is a whole different standard, that borrows only conceptually from WaterML 1.x. Don't bother to even look it up for this phase of things.

> We will use 1.1, as it is most uniformly provided.

Yup.

rajadain:
Hi @emiliom, @aufdenkampe,

I've been making progress on this, and have gotten to the point where I'd like to fetch a small subset of values for a given site and variable. With the `ulmo.cuahsi.wof.get_values` call, I can't specify just the start date and get all values from then until the end. It seems like I need to specify both start and end dates. Unfortunately, it seems like the value of "end date" is different for every CUAHSI endpoint. For a demo of this, please see https://gist.github.com/rajadain/e117e3c0ce16552564e764de65b3de85, especially results `Out[12]`, `Out[15]`, and `Out[17]`.

Essentially, this is the begin and end date as received from the `GetSeriesCatalogInBox2` call:

```
[...]

Issue #2234, opened 2017-09-07
BiG-CZ: WDC Detail View


rajadain:
Need template from @aufdenkampe / @emiliom for what to include.

If the service supports historical data, fetch and display one week's worth of it by default. We can expand on showing more after the user asks explicitly.

rajadain:
The actual charting should be done as part of #2238.

aufdenkampe:
@rajadain, @jfrankl & @emiliom.

I created a mockup in the [Sample_WDC_Site_Detail_BiGCZPortal_SearchResult](https://docs.google.com/document/d/1Kot_Toow6FPpI9FHImsx078W7Jn3UwMME4MOnersX2A/edit) Google Doc.

Here is a screenshot:
![screen shot 2017-09-08 at 4 05 11 pm](https://user-images.githubusercontent.com/5166036/30231238-9eb7f130-94af-11e7-8a05-97203c7ac4b5.png)

Please see the Google Doc for information on how it should be constructed from web service response fields, and comments on optional fields.

aufdenkampe:
Here is the description of how to construct from web service responses, where: 
- we use Markdown syntax to indicate links
- A GetSeries request is implied for all fields, except...
- a GSERV prefix implies that the information comes from a cached GetServices request

---------
![screen shot 2017-09-08 at 4 09 28 pm](https://user-images.githubusercontent.com/5166036/30231368-2f9d06e0-94b0-11e7-99d0-d0102f1717bb.png)
---------

This image does not include comments that re in the Google Doc. Here is my best attempt at copying them
-  After Site, the <ServCode> prefix is removed after <location> in the title
-  After Source: (info) opens a Pop up tool tip, showing text from <GSERV:aabstract>
- for the buttons:
  - Change "Respository" to "Source Data". This is optional, and links or original sources constructed in patterns such as: https://waterdata.usgs.gov/nwis/uv/?site_no=01478120, when ServCode = "NWISDV" and location = "NWISDV:01478120"​
  - "Web services" links to the Service URL. Where:
    - The ServURL in from GetServicesInBox2 does not provide a web-friendly URL (e.g. for ServCode = "NWISDV", ServURL = "http://hydroportal.cuahsi.org/nwisdv/cuahsi_1_1.asmx"). However, a friendly URL can easily be constructed from ServiceID (which has a 1:1 map with ServCode) by following this pattern: http://hiscentral.cuahsi.org/pub_network.aspx?n=1
[...]

PR #2259, opened 2017-09-14
BiG-CZ: WDC / CUAHSI Details View

rajadain:
## Overview

Adds specific items to CUAHSI's details view, based on comments in https://github.com/WikiWatershed/model-my-watershed/issues/2234#issuecomment-328214065 and [this Google doc](https://docs.google.com/document/d/1Kot_Toow6FPpI9FHImsx078W7Jn3UwMME4MOnersX2A). Important to note: this does not fetch the detail information for a given item. That will be done in #2238, which depends on #2243.

Also tagging @aufdenkampe for visual review.

Connects #2234 

### Demo

image

### Notes

Some fields have placeholder values, such as the citation in the above screenshot. These placeholders seem arbitrary, and to convert them all into real values may be a long-tailed effort. Currently they are left as is.

## Testing Instructions

 * Check out this branch. Run `bundle`
 * Search for something and open the WDC tab. Click on any item to get to the Details view.
 * Ensure the values match the description in the Google doc and the accompanying source comment.


caseycesari:
Links in the source tooltips are not clickable because any mouse event on the tooltip closes it. We've dealt with this somewhere before in the app, but I remember it getting a little messy. 

image

src/mmw/js/src/data_catalog/templates/resultDetailsCuahsi.html
```diff
@@ -0,0 +1,79 @@
...
+               data-html="true" data-container="body" role="button"
+               data-content="{{ description }}"
+               data-template="<div class='popover'><div class='pull-right' id='popover-close-button'><i class='fa fa-times' /></div><div class='popover-content'></div><div class='arrow'></div></div>">
+                <i class="fa fa-info-circle black"></i>
+            </a>
+        </p>
+        <p>
+            <!-- TODO Make This Work -->
+            <a class="zoom" href="#">
+                <i class="fa fa-search-plus"></i> Zoom to extent
```
caseycesari: I thought that all WDC results were point-based. If that's the case, it may not make sense to have a "zoom to extent" feature.

rajadain:
> Links in the source tooltips

Also, that description comes in from the API as plain text. We don't know it has links. We'll have to use a regex or something to transform them into links.

caseycesari:
> Also, that description comes in from the API as plain text. We don't know it has links. We'll have to use a regex or something to transform them into links.

Good point. It may make sense to let it be as is for now.

ajrobbins:
>Links in the source tooltips are not clickable
We definitely addressed this elsewhere in the app, such as the boundary selection tooltips:
screen shot 2017-09-14 at 5 56 02 pm

We should replicate that here; all tooltips should be potentially clickable, even if they don't have links at the moment.

mmcfarland:
> We definitely addressed this elsewhere in the app

The difference here is that we're not providing the content for the popup.  They're indicating that we'd have to parse for things that appear to be links and convert them from plain text.

ajrobbins:
>The difference here is that we're not providing the content for the popup. They're indicating that we'd have to parse for things that appear to be links and convert them from plain text.
I don't think we have to convert them, but even to support people copy-pasting a plain text link, they need to be able to click on the popup.

caseycesari:
Tooltip stuff sounds like a good follow-up card.

caseycesari:
> Some fields have placeholder values, such as the citation in the above screenshot. These placeholders seem arbitrary, and to convert them all into real values may be a long-tailed effort. Currently they are left as is.

Just to clarify, the citation values coming back from the server sometimes have placeholder values, like in the PR screenshot?

rajadain:
Yeah, like this:

> Citation: U.S. Geological Survey, [YEAR], National Water Information System data, accessed [DATE ACCESSED] via HIS Central (http://hiscentral.cuahsi.org).

src/mmw/apps/bigcz/clients/cuahsi/serializers.py
```diff
@@ -3,17 +3,22 @@
 from __future__ import unicode_literals
 from __future__ import division
 
-from rest_framework.serializers import CharField, DateTimeField
+from rest_framework.serializers import (CharField,
+                                        DateTimeField,
+                                        ListField
```
caseycesari: Minor, but consider adding a trailing comma here.

rajadain:
Thanks for the review! Squashed all the fixups, added that last trailing comma, will merge now.

aufdenkampe:
This all looks great. Did you all figure out that the text after "Citation:" comes directly from the "citation" field in the response from a GetServices request, which should be cached for all the services.

rajadain:
Yes, those values are coming from the (soon to be cached, see #2260) Services request, as specified in the Google Doc. Thanks for setting that up, it made the implementation considerably more straightforward.

Issue #1932, opened 2017-06-01
Cache CUAHSI GetServicesInBox2 results


kdeloach:
> BTW, for future consideration: it occurred to me that currently there are only 94 "data services" in WDC (ie, the maximum total that can be returned by GetServicesInBox2), and those change rarely. Maybe a future strategy for better search responsiveness could be to get all services in the US lower 48 (based on a GetServicesInBox2 query) and cache them once a day, then just use that cache together with the actual query results from GetSitesInBox2.

Ref: https://github.com/WikiWatershed/model-my-watershed/issues/1931#issuecomment-305317545

rajadain:
Commenting to subscribe to this thread and get email notifications.

rajadain:
This can be done easily along these lines:

```diff
diff --git a/src/mmw/apps/bigcz/clients/cuahsi/search.py b/src/mmw/apps/bigcz/clients/cuahsi/search.py
index 08622b4..955f8cd 100644
--- a/src/mmw/apps/bigcz/clients/cuahsi/search.py
+++ b/src/mmw/apps/bigcz/clients/cuahsi/search.py
@@ -124,8 +124,16 @@ def group_series_by_location(series):
 
 
 def make_request(request, **kwargs):
+    key = 'bigcz_{}_{}'.format(request.__name__,
+                               hash(frozenset(kwargs.items())))
+    cached = cache.get(key)
+    if cached:
+        return cached
+
     try:
-        return request(**kwargs)
+        response = request(**kwargs)
+        cache.set(key, response)
+        return response
     except URLError, e:
         if isinstance(e.reason, timeout):
             raise RequestTimedOutError()
```

Will need to ensure that this expires at appropriate times, because this is not an immutable operation.

PR #2260, opened 2017-09-15
BiG-CZ: Cache CUAHSI Requests

rajadain:
## Overview

CUAHSI searches use two endpoints: one for fetching services, and another for fetching series within them. The services rarely update, so we cache them for a week. The series update more frequently, so we cache them for 5 minutes. Each cache key is composed of `bigcz_{name of request method}_{sorted hashset of arguments}` so new requests should work as expected, and existing requests should be cached for the given TTL.

This work complements the filters PR #2258 which triggers searches on click. This should make enabling and disabling filters a lot quicker.

Connects #1932 

## Testing Instructions

 * Check out this branch, go to [:8000/?bigcz](http://localhost:8000/?bigcz)
 * Run a WDC search. It'll take it's time the first time.
 * Try and run the same search again. It should be noticeably faster.
 * Ensure the results are the same as before.

### DIFF
--- a/src/mmw/apps/bigcz/clients/cuahsi/search.py
+++ b/src/mmw/apps/bigcz/clients/cuahsi/search.py
@@ -9,6 +9,7 @@ from socket import timeout
 from operator import attrgetter
 
 from suds.client import Client
+from suds.sudsobject import asdict
 from rest_framework.exceptions import ValidationError
 from django.contrib.gis.geos import Point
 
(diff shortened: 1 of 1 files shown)


src/mmw/apps/bigcz/clients/cuahsi/search.py
```diff
@@ -228,6 +237,7 @@ def get_series_catalog_in_box(box, from_date, to_date, networkIDs):
     to_date = to_date or DATE_MAX
 
     result = make_request(client.service.GetSeriesCatalogForBox2,
```
mmcfarland: Can you speculate on the size of these responses so we have an idea of the impact to the cache storage we have available?

rajadain:
```
vagrant@services:~$ du -sh *.txt
392K	GetSeriesCatalogForBox2.txt  # Cached for 5 minutes
144K	GetServicesInBox2.txt        # Cached for 1 week
```

rajadain: It should be noted that the original issue only asked to cache `GetServicesInBox2`. I'm also caching the other one to make interacting with it faster.

rajadain:
Thanks for taking a look!

Issue #1754, opened 2017-03-23
Estimate SoilN value


mmcfarland:
Calculate `Soil N` value (first position `[SedNitr]` on Line 37 of “*.gms” file (see MapShed GMS File Key.doc for further detail). This particular value is now set by default as `2000`.

Procedure:
* Use new “soiln” raster to calculate this value ( #1753 )
* For a given AOI, calculate the “weighted average”
* Calculate the new `SedNitr` value as the weighted avg x 4.0, and write this value to Line 37

Depends on #1753

arottersman:
@rajadain moved ingests done in #1751 to better conform to our MMW ingest conventions:

`s3://datahub-catalogs-us-east-1/soiln_5070/soilpn_5070/0/` --> `s3://datahub-catalogs-us-east-1/soiln-epsg5070/0/`

`metadata` json moved to `_attributes` dir at the top level of the bucket.

PR #2261, opened 2017-09-15
Calculate SedNitr value from SoilN raster

rajadain:
## Overview

Calculates the average SoilN value for a given area of interest
and multiplies it by 4.0 to get the SedNitr value. This was
previously staticly set at 2000.

Connects #1754 

### Demo

Here is the GMS file produced on this branch: [MapShed_SoilN.gms.txt](https://github.com/WikiWatershed/model-my-watershed/files/1307427/MapShed_SoilN.gms.txt)

As can be seen on line 37 position 1, the value is not 2000:

```
1004,320,5.576021102150538,0.047421639784946235,0.25,0.25
```

The value of 1004 is for a square kilometer AoI in Center City Philadelphia.

image

### Notes

I imagine that the value of 2000 is typical of agricultural land use heavy areas of interest, so a value of ~1000 for a city sounds reasonable. For an area of interest with most land use in the categories of Cultivated Crops and Pasture / Hay, we are more likely to get a higher value.

## Testing Instructions

 * Check out this branch, and restart Celery `vagrant ssh worker -c 'sudo service celeryd restart'`
 * Go to [:8000/](http://localhost:8000/) and draw a shape and create a MapShed job
 * Export the GMS file and observer line 37, position 1. Ensure that it isn't 2000.

### DIFF
--- a/src/mmw/apps/modeling/mapshed/tasks.py
+++ b/src/mmw/apps/modeling/mapshed/tasks.py
@@ -156,7 +156,7 @@ def collect_data(geop_result, geojson):
 
 
 @shared_task(throws=Exception)
-def nlcd_streams(sjs_result):
+def nlcd_streams(result):
     """
     From a dictionary mapping NLCD codes to the count of stream pixels on
     each, return a dictionary with keys 'ag_stream_pct', 'low_urban_stream_pct'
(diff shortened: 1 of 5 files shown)


src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -319,6 +321,24 @@ def avg_awc(sjs_result):
 
 
 @shared_task(throws=Exception)
+def soiln(sjs_result):
+    """
+    Get `SoilN` from MMW-Geoprocessing endpoint
+
+    Originally a static value of 2000 at Class1.vb@1.3.0:9587
+    """
+    if 'error' in sjs_result:
```
kellyi: Wonder when it would make sense to rename `sjs_result` -> `geoprocessing_result`.

arottersman: Is it just this file that refers to it as such? I'd be happy to include a rename commit while working on the other two cards in this family.

rajadain: I'll add a lint commit to this PR that changes all of them. Good catch.

Issue #2228, opened 2017-09-07
BiG-CZ: Multiple search terms in CINERGI should AND, not OR


rajadain:
Currently, when specifying multiple search terms in CINERGI, they disjunct instead of conjunct, expanding instead of refining the results. Make it so that results contain _all_ search terms, instead of _any_.

PR #2262, opened 2017-09-15
BiG-CZ: Always conjunct CINERGI query items

rajadain:
## Overview

By default, the CINERGI query items would be ORd, thus expanding the search as more terms were specified instead of narrowing it down. By always ANDing the terms, we ensure that the more information a user provides, the narrower the search becomes.

Connects #2228 

## Testing Instructions

 * Check out this branch, and go to [:8000/?bigcz](http://localhost:8000/?bigcz)
 * Search for something on the CINERGI tab
 * Search for more things. Ensure that as the query gets longer, the results get fewer.

### DIFF
--- a/src/mmw/apps/bigcz/clients/cinergi/search.py
+++ b/src/mmw/apps/bigcz/clients/cinergi/search.py
@@ -120,6 +120,25 @@ def prepare_time(from_date, to_date):
     return value
 
 
+def prepare_query(query):
+    """
+    Prepare query to always AND all search terms
+
+    Spaces will be replaced with AND. Existing ANDs and ORs will be preserved.
+    """
+    def intersperse_and(phrase):
+        # For a given phrase, intersperse AND between all words,
+        # except "and" which is removed
+        return ' AND '.join([word for word in phrase.split()
+                             if word != 'and'])
+
+    # If the query has any ORs, split it into phrases to be ANDed
+    phrases = query.split(' or ')
+
+    # AND words in each phrase, and OR all phrases together
+    return ' OR '.join(map(intersperse_and, phrases))
+
+
 def search(**kwargs):
     query = kwargs.get('query')
     to_date = kwargs.get('to_date')
@@ -134,7 +153,7 @@ def search(**kwargs):
 
     if query:
         params.update({
-            'q': query
+            'q': prepare_query(query.lower())
         })
     if from_date:
         params.update({


caseycesari:
In a test area, the searches "water soil", "water and soil", and "water or soil" all returned the same number of results. Looking at the API requests, all of those searches are converted to "water and soil". Do we want to prevent a user from doing an "OR" based search?

rajadain:
Hmm, my impression was that AND was significantly more valuable than OR, but perhaps we should support explicit ORs. I'll transform the query into a sum of products format and that should fix it. Fixup coming soon.

rajadain:
With the latest commit 0135b0c:

```python
In [23]: prepare_query('new york')
Out[23]: 'new AND york'

In [24]: prepare_query('new and york')
Out[24]: 'new AND york'

In [25]: prepare_query('new or york')
Out[25]: 'new OR york'
```

src/mmw/apps/bigcz/clients/cinergi/search.py
```diff
@@ -120,6 +121,21 @@ def prepare_time(from_date, to_date):
...
+
+    Spaces will be replaced with AND. Existing ANDs and ORs will be preserved.
+    """
+    def intersperse_and(phrase):
+        return ' AND '.join([w for w in phrase.split()
+                             if not re.match('and', w, flags=re.IGNORECASE)])
+
+    factors = re.split(' or ', query, flags=re.IGNORECASE)
+
+    return ' OR '.join(map(intersperse_and, factors))
```
rajadain: Earlier I tried writing it as:

```python
return ' OR '.join([' AND '.join(w for w in f.split()
                                 if not re.match('and', w,
                                                 flags=re.IGNORECASE))
                    for f in factors])
```

but thought it less clear. Any suggestions for improving readability welcome.

rajadain: Maybe if I replace `factors` with `phrases` it'll be a little nicer.

caseycesari:
Taking another look.

src/mmw/apps/bigcz/clients/cinergi/search.py
caseycesari: This is a small improvement, but if the API is not case sensitive, you could convert everything to lower/uppercase beforehand which would let you drop `flags=re.IGNORECASE`.

rajadain: Just checked, the API is _not_ case sensitive. Will replace the `re` stuff and that should help make it easier to read. Will add comments if it's still not clear enough. Thanks

rajadain:
Okay, this should be ready for another look now. Functionally the same, the code looks a little neater, and has more comments.

rajadain:
Thanks for the feedback! Going to squash and merge.

Issue #2215, opened 2017-08-31
Hide CINERGI description field in list when it is set to "REQUIRED FIELD"


caseycesari:
Detect if the value of each CINERGI result is set to "REQUIRED FIELD". If it is, do not render it.

This is a special case for a better UX.

PR #2266, opened 2017-09-19
Hide Cinergi "REQUIRED FIELD"

kellyi:
## Overview

This PR hides the "REQUIRED FIELD" text which appeared as the summary/description for some Cinergi results. To accomplish this we check whether the value for the search result summary or the result description is "REQUIRED FIELD" and only show them if not.

Connects #2215 

### Demo

screen shot 2017-09-19 at 10 25 39 am

## Testing Instructions
- get this branch then bundle and visit :8000/?bigcz
- select an AOI, then search data sources for "water"
- in the Cinergi results verify that you do not see any summaries/descriptions with values of "REQUIRED FIELD" but that the other results do show summaries & descriptions
- check that the other catalogs' results display properly


kellyi:
Thanks!

Issue #2249, opened 2017-09-12
Upgrade Celery


rajadain:
There are numerous issues with Celery 3, which are no longer being addressed. The current version of Celery is 4, and can work with Django 1.8, thus we can upgrade to Celery 4 without first upgrading Django #2146.

Upgrade notes can be found here: http://docs.celeryproject.org/en/latest/whatsnew-4.0.html

Upgrade Celery to the latest version, and adjust related code to match.

kellyi:
Started looking at this.

kellyi:
Sounds good. Seems like everything cleanly upgrades to 3.1.25 -- the first suggest step in the upgrade guide above -- so I'll do that then this http://docs.celeryproject.org/en/latest/whatsnew-4.0.html#step-2-update-your-configuration-with-the-new-setting-names and get that working before moving to 4.x

kellyi:
Q: is there any reason we didn't add the `--purge` option to `debugcelery`'s command to start the Celery worker? Looks like it dumps previous messages on restarting.

Going to enable it while working on this on my local, but conceivably we could add it to the script checked in here, too.

kellyi:
Relatedly, looks like we'll have to drop `--autoreload` from that script since...

> The --autoreload feature has been removed.
>
> This was an experimental feature, and not covered by our deprecation timeline guarantee. The flag is removed completely so the worker will crash at startup when present. Luckily this flag isn’t used in production systems.

http://docs.celeryproject.org/en/latest/whatsnew-4.0.html#features-removed-for-lack-of-funding

PR #2267, opened 2017-09-19
WIP Upgrade to Celery 4.1

kellyi:
## Overview

WIP: This is most/much of the way toward upgrading Celery to 4.1. Remaining work seems to be conditionally applying `link_error=errback` or not to the climate-related geoprocessing tasks.
 
Connects #2249


scripts/debugcelery.sh
```diff
@@ -6,6 +6,6 @@ set -e
 
 STOP_SERVICE="(sudo service celeryd stop || /bin/true)"
 CHANGE_DIR="cd /opt/app/"
-RUN_CELERY="envdir /etc/mmw.d/env celery -A 'mmw.celery:app' worker --autoreload -l debug -n debug@%n"
+RUN_CELERY="envdir /etc/mmw.d/env celery -A 'mmw.celery:app' worker --purge -l debug -n debug@%n"
```
kellyi: `--autoreload` no longer exists as an option; if we want to keep `--purge`, it will discard all prior pending messages on restarting the celery workers.

src/mmw/requirements/base.txt
```diff
@@ -18,3 +18,4 @@ rollbar==0.13.8
 retry==0.9.1
 python-dateutil==2.6.0
 suds==0.4
+django_celery_results==1.0.1
```
kellyi: This import is unnecessary.

deployment/ansible/roles/model-my-watershed.celery/defaults/main.yml
```diff
@@ -1,3 +1,3 @@
 ---
-celery_version: 3.1.18
-djcelery_version: 3.1.16
+celery_version: 4.1.0
+djcelery_version: 3.2.1
```
kellyi: Not certain that we have to keep the django-celery upgrade here.

src/mmw/requirements/base.txt
kellyi: Spoke too soon; seems like we may need this after all.

kellyi:
While testing this out I ended up encountering an error related to the geoprocessing service's AWS credenitals -- which I think are not tied to these changes but which made it difficult to test whether the setup here is repeatable on provisioning new app & worker VMs.

I *think* this is most of the way to done but there are two celery-specific tests which are failing & need to be updated.

kellyi:
Failing tests are:

```
======================================================================
ERROR: test_tr55_job_runs_in_chain (app.apps.modeling.tests.TaskRunnerTestCase)
----------------------------------------------------------------------
  File "/usr/local/lib/python2.7/dist-packages/django/test/utils.py", line 196, in inner
    return test_func(*args, **kwargs)
  File "/opt/app/apps/modeling/tests.py", line 221, in test_tr55_job_runs_in_chain
    found_job = Job.objects.get(uuid=task_list.id)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/manager.py", line 127, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 334, in get
    self.model._meta.object_name
DoesNotExist: Job matching query does not exist.

======================================================================
FAIL: test_tr55_job_error_in_chain (app.apps.modeling.tests.TaskRunnerTestCase)
----------------------------------------------------------------------
  File "/usr/local/lib/python2.7/dist-packages/django/test/utils.py", line 196, in inner
    return test_func(*args, **kwargs)
  File "/opt/app/apps/modeling/tests.py", line 257, in test_tr55_job_error_in_chain
    chain(job_chain).apply_async()
AssertionError: Exception not raised

----------------------------------------------------------------------
```

rajadain:
Taking a look at this.

src/mmw/requirements/base.txt
```diff
@@ -18,3 +18,4 @@ rollbar==0.13.8
 retry==0.9.1
 python-dateutil==2.6.0
 suds==0.4
+django-celery-results==1.0.1
```
rajadain: I don't think we need this. Celery task results are stored manually via [`core/job`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/core/models.py), and we don't interact with them via ORM at all.

rajadain:
Also remove [this hack](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/celery.py#L14-L83), and the corresponding [setting variable](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/base.py#L133).

src/mmw/requirements/base.txt
kellyi: I'll try out dropping this again.

kellyi: Ok, I tried out dropping this and the destroying & rebuilding the worker and the upgrade won't work from that state without some additional change because it looks like the worker's version of Celery sticks at 3.1.25.

I think that at the very least `django-celery-results` might be forcing an upgrade to the worker's Celery installation that's not currently captured in the Ansible rules. Can't speak to whether we need the library altogether, but we do need to ensure that there's an explicit rule to upgrade the worker in addition to the app.

mmcfarland: Sounds like the `model-my-watershed.celery` is not applying cleanly to `worker` or that something is downgrading it after the dependency is installed.  Definitely want to make sure we're getting the version installed from `model-my-watershed.celery` and not transitively from another library.

kellyi:
Superseded by #2287, so closing this one. I'll delete the branch after #2287 is in.

Issue #2190, opened 2017-08-24
Geoprocessing API: Log requests


arottersman:
From the ADR:
https://github.com/WikiWatershed/model-my-watershed/blob/develop/doc/arch/adr-004-geoprocessing-api.md
> Since we'll be logging to the database, we'll need to provide an interface into that data. I suggest we keep that simple and create a management task which exports ranges of usage data as a CSV so they can be evaluated in Excel. If this feature ends up gaining traction, we'll want to invest in in a more robust solution. A DRF plugin that may assist with task can be evaluated here http://drf-tracking.readthedocs.io/en/latest/.

arottersman:
The work for this card should be to 
* investigate framework options for logging, and pick one
* Set the api up with

We'll save the mechanism for exporting/viewing the logs for a subsequent card.

arottersman:
`drf-tracking` would cause two disturbances
1. It works via a mixin, which I think would require we convert all the views to be class-based
2. It stores more than we might currently want it to, like the full response and request body

I haven't found a comparable framework — I think we should roll our own
- add a model to store the logging info
-  add a logging decorator we can add to our api views

PR #2271, opened 2017-09-19
Log Geoprocessing API Requests

arottersman:
## Overview

Provide a way to track requests made to the geoprocessing API. Right now this just writes all Geoprocessing API requests to the DB. We'll address getting the logs back out in #2274, and filtering out requests from the frontend in #2275.

- Looked into the [drf-tracking](https://github.com/aschn/drf-tracking) package but it...
    - didn't let us selectively choose what parts of the request we wanted to store. The AoI payloads can be a bit big to store twice (they're already in the `core_job` table)
   - is written as a mixin, and we use function-based views

I decided to write our own `RequestLog` model  and adopt only the elements of drf-tracking model @ajrobbins thought we might need. A custom decorator writes the requests to it.

Connects #2190 

### Demo

```sql
select core_requestlog.*, core_job.uuid, core_job.status
from core_requestlog, core_job
where core_requestlog.job_id = core_job.id;

 id |         requested_at          | response_ms |        path        |                query_params    
            | status_code | method |      host      | remote_addr | job_id | user_id |                 u
uid                 |  status                                                                          
----+-------------------------------+-------------+--------------------+--------------------------------
------------+-------------+--------+----------------+-------------+--------+---------+------------------
--------------------+----------                                                                        
  4 | 2017-09-19 14:10:53.600949+00 |         123 | /api/analyze/land/ |                                
            |         200 | POST   | localhost:8000 | 10.0.2.2    |  17448 |       5 | e98f607f-64c2-4c4
1-a369-9227bb2a0f37 | complete                                                                                                                                           
  6 | 2017-09-19 14:23:18.72764+00  |         113 | /api/analyze/land/ | {u'wkaoi': u'huc12__55174'}    
            |         200 | POST   | localhost:8000 | 10.0.2.2    |  17452 |       5 | 14572a0f-1f13-434
6-bf1b-4a55ba54feaa | complete                                                                         
(2 rows)
```

## Testing Instructions

 * Pull this branch and run migrations to get the new requestlog table `./scripts/manage.sh migrate`
 * Make a variety of requests to the geoprocessing API and confirm the appropriate fields are logged by sshing onto the services VM and inspecting the log table

```
vagrant ssh services
 psql -h localhost -U mmw -p 5432
...
mmw# select core_requestlog.*, core_job.uuid, core_job.status
from core_requestlog, core_job
where core_requestlog.job_id = core_job.id;
```

### DIFF
--- a/src/mmw/apps/core/decorators.py
+++ b/src/mmw/apps/core/decorators.py
@@ -6,8 +6,15 @@ import sys
 import rollbar
 
 from django.utils.timezone import now
+from django.conf import settings
 
 from apps.core.models import RequestLog
+
+
+rollbar_settings = getattr(settings, 'ROLLBAR', {})
+if rollbar_settings:
+    rollbar.init(rollbar_settings.get('access_token'),
+                 rollbar_settings.get('environment'))
 
 
 def log_request(view):
@@ -39,7 +46,8 @@ def log_request(view):
         try:
             log.save()
         except Exception:
-            pass
+            if rollbar_settings:
+                rollbar.report_exc_info(sys.exc_info(), request)
 
         return view_result
 
--- a/src/mmw/apps/core/decorators.py
+++ b/src/mmw/apps/core/decorators.py
@@ -1,0 +1,60 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import sys
+import rollbar
+
+from django.utils.timezone import now
+
+from apps.core.models import RequestLog
+
+
+def log_request(view):
+    """
+    Log the request and its response as a RequestLog model
+    """
+
+    def decorator(request, *args, **kwargs):
+        requested_at = now()
+
+        view_result = view(request, *args, **kwargs)
+
+        user = request.user if request.user.is_authenticated() else None
+
+        response_time = now() - requested_at
+        response_ms = int(response_time.total_seconds() * 1000)
+
+        log = RequestLog.objects.create(
+            user=user,
+            job_uuid=view_result.data.get('job', None),
+            requested_at=requested_at,
+            response_ms=response_ms,
+            status_code=view_result.status_code,
+            path=request.path,
+            query_params=request.query_params.dict(),
+            method=request.method,
+            host=request.get_host(),
+            remote_addr=_get_remote_addr(request))
+        try:
+            log.save()
+        except Exception:
+            pass
+
+        return view_result
+
+    decorator.__name__ = view.__name__
+    decorator.__dict__ = view.__dict__
+    decorator.__doc__ = view.__doc__
+
+    return decorator
+
+
+def _get_remote_addr(request):
+    # get IP
+    ipaddr = request.META.get("HTTP_X_FORWARDED_FOR", None)
+    if ipaddr:
+        # X_FORWARDED_FOR returns client1, proxy1, proxy2,...
+            return [x.strip() for x in ipaddr.split(",")][0]
+    else:
+        return request.META.get("REMOTE_ADDR", "")
--- a/src/mmw/apps/core/migrations/0002_requestlog.py
+++ b/src/mmw/apps/core/migrations/0002_requestlog.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+from django.conf import settings
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ('core', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='RequestLog',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('job_uuid', models.UUIDField(help_text='If async request, the uuid of the submitted job', null=True)),
+                ('requested_at', models.DateTimeField(help_text='When the request was made')),
+                ('response_ms', models.PositiveIntegerField(help_text='How long the response took in ms')),
+                ('path', models.CharField(help_text='The requested URI', max_length=400)),
+                ('query_params', models.TextField(help_text='Stringified requested query params dictionary', null=True)),
+                ('status_code', models.PositiveIntegerField(help_text='HTTP status code sent back')),
+                ('method', models.CharField(help_text='HTTP method', max_length=255)),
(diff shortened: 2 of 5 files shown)


rajadain:
@azavea-bot rebuild

rajadain:
Taking a look now.

caseycesari:
I may be misremembering, but I think there was something similar for OTM2 that ended up slowing things down. Like the audit log table? Might want to have @jwalgran take a look.

jwalgran:
In OTM audit the changes to most of our models, and we have built up quite a large table. Building some pages for general users requires querying this table. Over the years we have made optimizations to avoid reduce or avoid the cost of these queries.

In this logging situation, where the main users of the app will never end up directly querying or joining this log table, I would not expect this to be a drag on performance.

src/mmw/apps/core/decorators.py
```diff
@@ -0,0 +1,66 @@
...
+
+        view_result = view(request, *args, **kwargs)
+
+        user = request.user if request.user.is_authenticated() else None
+
+        job_uuid = view_result.data.get('job', None)
+        job = None
+        if (job_uuid):
+            try:
+                job = Job.objects.get(uuid=job_uuid)
```
rajadain: This is possibly the most expensive call (outside the main `INSERT` below). Ultimately we'll need to associate `RequestLog`s with `Job`s, but this can be done in two places:

 * At the write stage, as done here, where the Job UUID is translated to the Job ID before being inserted
 * At the read stage, when a report is being prepared, translating the stored Job UUID to a Job ID

Since the writes are expected to be orders of magnitude more frequent than reads, we should defer converting the Job UUID to the Job ID until that read stage. We can join the two tables when generating a report.

src/mmw/apps/core/decorators.py
```diff
@@ -0,0 +1,66 @@
...
+            status_code=view_result.status_code,
+            path=request.path,
+            query_params=request.query_params.dict(),
+            method=request.method,
+            host=request.get_host(),
+            remote_addr=_get_remote_addr(request))
+        try:
+            log.save()
+        except Exception:
+            pass
```
rajadain: Rather than silently failing, is it possible for this to alert Rollbar of a failure, but still succeeding the request?

src/mmw/apps/core/migrations/0002_requestlog.py
```diff
@@ -0,0 +1,33 @@
...
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('requested_at', models.DateTimeField(help_text='When the request was made')),
+                ('response_ms', models.PositiveIntegerField(help_text='How long the response took in ms')),
+                ('path', models.CharField(help_text='The requested URI', max_length=400)),
+                ('query_params', models.TextField(help_text='Stringified requested query params dictionary', null=True)),
+                ('status_code', models.PositiveIntegerField(help_text='HTTP status code sent back')),
+                ('method', models.CharField(help_text='HTTP method', max_length=255)),
+                ('host', models.URLField(help_text='The host from which the request was sent')),
+                ('remote_addr', models.GenericIPAddressField(help_text='The IP address from which the request was sent')),
+                ('job', models.ForeignKey(on_delete=django.db.models.deletion.SET_NULL, to='core.Job', null=True)),
```
rajadain: This would turn into a UUID field that is not a Foreign Key

arottersman:
I've pushed some fixups that convert the foreign key job_id  on the `RequestLog` model to `job_uuid`. I've also added a commit  to log exceptions on `log.save()` to Rollbar.

To test again, I believe you'll have to:

```bash
> ./scripts/manage.sh migrate core 0001_initial
> ./scripts/manage.sh migrate
```

src/mmw/apps/core/decorators.py
```diff
@@ -0,0 +1,68 @@
...
+        view_result = view(request, *args, **kwargs)
+
+        user = request.user if request.user.is_authenticated() else None
+
+        response_time = now() - requested_at
+        response_ms = int(response_time.total_seconds() * 1000)
+
+        log = RequestLog.objects.create(
+            user=user,
+            job=view_result.data.get('job', None)
```
rajadain: Should this be `job_uuid` instead of `job`?

arottersman: Yes

rajadain:
There are a _number of_ failing tests on this branch that don't fail on `develop`

arottersman:
Should be fine now (though the build will be marked unstable until this is rebased on develop)

rajadain:
Let's rebase on develop then. I merged with develop and am still getting flake8 errors, so that will likely happen when this gets merged if not rebased.

arottersman:
Rebased and squashed down the _number of_ fixups.

rajadain:
Taking another look.

rajadain:
This seems to be working pretty well. Doesn't add a significant delay to the operations, and logs a lot of useful information. One curiosity I came across is how it doesn't log the `user_id` for `POST` requests, but _does_ log them for the subsequent `GET`s:

```sql
SELECT path, method, host, remote_addr, user_id FROM core_requestlog;

                      path                       | method |      host      | remote_addr | user_id
-------------------------------------------------+--------+----------------+-------------+---------
 /api/analyze/land/                              | POST   | localhost:8000 | 10.0.2.2    |  <NULL>
 /api/analyze/soil/                              | POST   | localhost:8000 | 10.0.2.2    |  <NULL>
 /api/analyze/animals/                           | POST   | localhost:8000 | 10.0.2.2    |  <NULL>
 /api/analyze/pointsource/                       | POST   | localhost:8000 | 10.0.2.2    |  <NULL>
 /api/analyze/catchment-water-quality/           | POST   | localhost:8000 | 10.0.2.2    |  <NULL>
 /api/jobs/2e157683-a3f0-4d35-8418-35ede5870a51/ | GET    | localhost:8000 | 10.0.2.2    |       3
 /api/jobs/c244f3af-298b-40ea-90db-f10d5c3ecf57/ | GET    | localhost:8000 | 10.0.2.2    |       3
 /api/jobs/cb21de54-918e-4f50-8482-dd266468bb66/ | GET    | localhost:8000 | 10.0.2.2    |       3
 /api/jobs/000a1133-6ebe-4f84-8305-72969514cd17/ | GET    | localhost:8000 | 10.0.2.2    |       3
 /api/jobs/dcb2261d-09d1-4ca0-8d1e-438bd2061ee7/ | GET    | localhost:8000 | 10.0.2.2    |       3
(10 rows)
```

arottersman:
Were those requests made via the UI? The user would come from the token, which the UI doesn't send. The `/jobs` endpoint needs to work for both the API and modeling, so it's secured via session as well as token.

PR #2273, opened 2017-09-20
Fix nonsensical Geoprocessing API whitelist check

arottersman:
## Overview

* Remove REMOTE_ADDR check to determine if a request is from our own client
  app...none will ever be on the allowed_host lists unless you're running
  everything on localhost

* Only check if a user is token authenticated if the request comes from the
  swagger docs. This allows us to fix the above problem with a minimal diff, and
  still see that the token auth works

Connects #2248 

## Testing Instructions

 * Pull branch
*  Confirm you can send requests without a token
* Confirm tokens are required when sending a request via swagger

### DIFF
--- a/src/mmw/apps/core/permissions.py
+++ b/src/mmw/apps/core/permissions.py
@@ -8,19 +8,19 @@ from rest_framework.authentication import TokenAuthentication
 from django.conf import settings
 
 
-class IsTokenAuthenticatedOrClientApp(BasePermission):
+class IsTokenAuthenticatedOrNotSwagger(BasePermission):
     """
-    Global permission check for request being
-    (a) from a whitelisted IP
-      ie. is this our own client web app and not the swagger docs?
-    (b) from a user authenticated via auth token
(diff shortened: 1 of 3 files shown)


src/mmw/apps/geoprocessing_api/management/commands/exportrequestcsv.py
```diff
@@ -0,0 +1,28 @@
...
+
+class Command(BaseCommand):
+    """
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument('limit', type=int)
+
+    def handle(self, *args, **options):
+        self.stdout.write('Randomly marking inlets...')
```
rajadain: This feels like it doesn't belong here ...

arottersman: Oh that was me messing around with something on another branch. Sorry.

arottersman:
Was really rushing this — amended and force pushed the commit to get rid of all the stray files.

Issue #2270, opened 2017-09-19
Allow anonymous requests only from client app


arottersman:
Currently the geoprocessing api checks the list of allowed hosts for whether or not it should allow an unauthenticated request. Because all requests come in through the reverse proxy, all requests can be pass this check for being from one of the allowed hosts.

See discussion that starts here:
https://github.com/WikiWatershed/model-my-watershed/pull/2248#pullrequestreview-63227191

PR #2280, opened 2017-09-21
Add API Auth ADR

arottersman:
ADR on how the client app should authenticate its anonymous requests

;tldr/SPOILER ALERT: The client app should have its own user/API token like everyone else. We can't keep it secret, but we can pass it via the client settings in such a way that we can cycle it as frequently as possible. 

Thanks @mmcfarland for coaching me through this, and identifying a lot of the possibilities/trade offs.

@hectcastro — you might want to have a look at this.

Connects #2270

### DIFF
--- a/doc/arch/adr-005-api-client-app-auth.md
+++ b/doc/arch/adr-005-api-client-app-auth.md
@@ -1,0 +1,127 @@
+# Geoprocessing API: Only Allow Anonymous Requests From The Client App
+
+## Context
+
+We're creating a [proper, publicly documented API](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/geoprocessing_api/views.py) around our analyze and 
+RWD endpoints: the [Geoprocessing API.](https://github.com/WikiWatershed/model-my-watershed/blob/develop/doc/arch/adr-004-geoprocessing-api.md) The analyze and RWD endpoints have
+always had allow-any access; we'd now like to secure them via token authentication
+so that we can better track individual users, prevent users from sending off too many
+requests, revoke the access of problematic users, and some day maybe add a 
+paid tier of use.
+
+Enforcing token authentication on these previously open endpoints causes
+a problem for our client app; the client app consumes the same API, and
+want to continue to allow unauthenticated users access to these endpoints
+via the app. We need a way to identify the client app as a special case
+from which the app should allow unrestricted requests.
+
+Because everything we send to and from the client is exposed to the app user,
+any mechanism we give the client app to identify itself will be fairly
+easy to uncover and then fake by anyone trying to get un-checked access to the
+API. With this in mind our solution should prioritize not be overly complicated/difficult
+for us when we know there's no truly secure solution (barring doing all our rendering
+server-side).
+
+Some possibilities are:
+
+1. Give each environment of the client app its own API token
+   - Keeps a single system for API authentication; easy to reason about;
+     as a bonus will allow us to cycle the client app's token if someone
+     decides to start using it
+   
+1. Set a special, is-client-app flag that the API would check to determine if authentication was necessary
+   - To set the flag we'd have to know the request was from the client app in
+     the first place. Preliminary checks for doing this via `http_referer`
+     were unsuccessful, which would leave doing this via custom header (or some
+     other part of the request). Setting a `X-IsClientApp` custom header might
+     be overly naive or hacky, and could result in a lot of conditional logic
+     
+1. Do (1) and also enforce the `HTTP_REFERER` header be from an expected domain.
+[Google Maps ](https://developers.google.com/maps/documentation/javascript/get-api-key#key-restrictions)
+does this. Even though such headers are easily spoofed by a client outside of a browser,
+they're fairly effective for browser-based apps and for naive use from a non-browser client
+
+## Decision
+
+**Give the client app its own API token**
+
+Setting up the client app with its own API token will result a cleaner
+architecture for the API. All users use a token, no exceptions. In addition to
+the simpler mental model the API token method provides, we'll also gain an added
+level of token enforcement; if some API user tries to programmatically use the client
+app's token instead of their own, we'll be able to easily cycle the client app's token 
+to de-incentivize them. 
+
+One easy way to cycle the token:
+```
+./manage.py drf_create_token -r <client_app_username>
+```
+
+We can always implement (3) as an enhancement later on.
+
+##### When should the client app send its token
+
+For all Geoprocessing API requests, whether there's a logged-in user's token available
+to use or not, ie., we will continue to use a user's credentials when needed for the
+application endpoints.
+
+
+##### How the client app should get the token
+
+If we hard-code the token on the frontend (or anywhere), we'll have to deploy any time we need
+to cycle the token. We should instead pass the token to the client app via `clientSettings`. 
+
+##### How the server should get the token
+
+To get the token to put in the client's settings, the server can look up the
+token from the `authtoken_token` table via the client app's `user_id`.
+The `user_id` should remain secret so that there's no programmatic way get
+the token outside of swiping it off the client app.
+
+## Consequences
+
+#### Backdoor to the API
+As discussed in [Context](#Context), allowing the client app to make requests without a user creates a 
+backdoor into the API. This is acceptable because up until now we've allowed unrestricted API access
+(we just haven't advertised the API as something you could use.) There's also no sensitive data available
+via the API. While it's not ideal that someone could use the client app token to make unlimited,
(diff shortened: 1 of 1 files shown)


doc/arch/adr-005-api-client-app-auth.md
```diff
@@ -0,0 +1,108 @@
...
+
+If we hard-code the token on the frontend (or anywhere), we'll have to deploy any time we need
+to cycle the token. We should instead pass the token to the client app via `clientSettings`. 
+
+##### How the server should get the token
+
+To get the token to put in the client's settings, the server can look up the
+token via the client app's `user_id`. The `user_id` will have to remain
+secret so that there's no programmatic way get the token outside of swiping
+it off the client app.
```
caseycesari: I may be missing something, but where is the token stored?

doc/arch/adr-005-api-client-app-auth.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Consequences
+
+#### Backdoor to the API
+As discussed in Context, allowing the client app to make requests without a user creates a
```
caseycesari: Minor but consider linking Context to the context section.

doc/arch/adr-005-api-client-app-auth.md
```diff
@@ -0,0 +1,108 @@
...
+   - Things could get a lot more complex for us, but not that much more complex
+   for any would-be attacker
+
+## Decision
+
+**Give the client app its own API token**
+
+Setting up the app with its own API token will allow us to both
+- cycle the key easily because Django Rest Framework (DRF) handles token
+generation/re-generation, and we already use it for "real" API users
```
caseycesari: How does the cycling happen? Is there an endpoint in DRF to do this?

doc/arch/adr-005-api-client-app-auth.md
arottersman: The token is stored in the `authtoken_token` table created by [`rest_framework.authtoken`](http://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication)

doc/arch/adr-005-api-client-app-auth.md
arottersman: There's a built-in management command `./manage.py drf_create_token -r <username>`. We'll also be exposing an endpoint for use on the new user accounts page.

doc/arch/adr-005-api-client-app-auth.md
```diff
@@ -0,0 +1,108 @@
...
+top as enhancements down the road.
+
+##### How the client app should get the token
+
+If we hard-code the token on the frontend (or anywhere), we'll have to deploy any time we need
+to cycle the token. We should instead pass the token to the client app via `clientSettings`. 
+
+##### How the server should get the token
+
+To get the token to put in the client's settings, the server can look up the
```
hectcastro: Is the idea that each user of the application will use a token that is associated to their user ID, or a token specific to the application that we use on their behalf?

At different points while reading through, I thought it was the latter, but then later felt as though the former was being proposed.

arottersman: I'll make this more clear. Trying to propose the latter; we'd be sending a token specific to the application on their behalf. The `user_id` would be that of a user we'd create for the client app to use for all geoprocessing API requests.

When using the app, an individual user's credentials would only be used for the modeling and user endpoints (they're not part of the public API).

arottersman:
Just spoke with @mmcfarland offline — going to do a quick experiment to see if we can distinguish the app via session. This is the opposite of what the ADR currently proposes, but may save us a lot of the complexity of setting up users for each app.

doc/arch/adr-005-api-client-app-auth.md
```diff
@@ -0,0 +1,108 @@
...
+
+We want to throttle users of the API, and if the client app is just another, regular user
+of the API with a token it would get throttled the same. All guest MMW users in the world
+would share a cumalitive number of requests per minute.
+
+To fix this we'll need to treat the client app's token as a special case for throttling: 
+
+if token is client app:
+
+cache number of requests per IP (like [AnonRateThrottle](http://www.django-rest-framework.org/api-guide/throttling/#anonratethrottle))
```
hectcastro: For this, we just need to make sure the IP it uses is truly the clients vs. the IP of our load balancer, or some other hop internal to our infrastructure. May be hard to test outside of EC2.

doc/arch/adr-005-api-client-app-auth.md
```diff
@@ -0,0 +1,108 @@
...
+
+#### Dev Setup
+
+The `user_id` the app server uses to get the client's token should be fairly
+constant for the life of staging and production. Each developer's machine,
+however, will have different database instances and, therefore, different
+client app `user_id`'s.
+
+We could:
+1. Use some special email or name like `clientapp09212017` for the app server to look up the `user_id` to get the token. This would add a level
```
hectcastro: I think that this is a good place to start. For example, Raster Foundry has a user that Airflow uses to interact with the main API on its behalf called `rf|airflow-user`.

arottersman:
- Per @mmcfarland suggestion, reframed the decision to be more in terms of architectural integrity than security
- After some experiments in detecting what was and wasn't a request from the client app, reworked option 2 to reflect what I found to be possible 
- Added clarification on how the tokens work via DRF
- Added clarification on when the client app would be sending tokens

doc/arch/adr-005-api-client-app-auth.md
```diff
@@ -0,0 +1,121 @@
+# Geoprocessing API: Only Allow Anonymous Requests From The Client App
+
+## Context
+
+We're creating a [proper, publicly documented API](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/geoprocessing_api/views.py) around our analyze and 
+RWD endpoints: the [Geoprocessing API.](https://github.com/WikiWatershed/model-my-watershed/blob/develop/doc/arch/adr-004-geoprocessing-api.md) The analyze and RWD endpoints have
+have allow-any access; we'd now like to secure them via token authentication
```
mmcfarland: "have had", maybe?  Or even, "have always had"?

doc/arch/adr-005-api-client-app-auth.md
```diff
@@ -0,0 +1,121 @@
...
+so that we can better track individual users, prevent users from sending off too many
+requests, revoke the access of problematic users, and some day maybe add a 
+paid tier of use.
+
+Enforcing token authentication on these previously open endpoints causes
+a problem for our client app; the client app consumes the same API, and
+we want to allow it to use these endpoints un-authenticated. We need a way
```
mmcfarland: Consider rephrasing to illustrate that the endpoints may be authenticated, but unauthenticated users of the app should continue to be able to access them.

Something like "..want to continue to allow unauthenticated users access to these endpoints via the app"

doc/arch/adr-005-api-client-app-auth.md
```diff
@@ -0,0 +1,121 @@
...
+
+One easy way to cycle the token:
+```
+./manage.py drf_create_token -r <client_app_username>
+```
+
+##### When should the client app send its token
+
+For all Geoprocessing API requests, whether there's a logged-in user's token available
+to use or not, ie., we will continue to use a user's credentials only for user/model/project
```
mmcfarland: Maybe this is just being pedantic, but technically those requests don't need any credentials to access.  Maybe refer to these as application endpoints, to distinguish them from "api" endpoints?

doc/arch/adr-005-api-client-app-auth.md
arottersman: There are a terrifying number of repeated words...I think this was "have always had" before I mangled it.

doc/arch/adr-005-api-client-app-auth.md
```diff
@@ -0,0 +1,108 @@
...
+1. Give each environment of the client app its own API token with which it
+would make guest user requests
+   - Keeps a single system for API authentication
+   
+1. Set a special flag in the session token that the API would check for in
+unauthenticated requests
+   - Could be the CSRF token or something else. Might involve some custom
+   generation/management
+   
+1. Some combination of (1) and (2)
```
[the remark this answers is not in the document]
arottersman: I removed mention of this, but can add it back in.

arottersman:
Made the final suggested edits for clarity — going to squash and merge.

doc/arch/adr-005-api-client-app-auth.md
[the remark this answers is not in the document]
mmcfarland: Sorry, I thought I deleted this comment with my others after our discussions.  Still an interesting idea, but didn't intend to muddy the waters.

Issue #2279, opened 2017-09-21
OpenJDK provisioning failure on new services VM


jwalgran:
Provisioning a new services VM fails on this step:

TASK [azavea.java : Install OpenJDK] *******************************************
Thursday 21 September 2017  10:28:02 -0700 (0:00:00.308)       0:01:24.539 ****
fatal: [services]: FAILED! => {"cache_update_time": 1506014812, "cache_updated": false, "changed": false, "failed": true, "msg": "'/usr/bin/apt-get -y -o \"Dpkg::Options::=--force-confdef\" -o \"Dpkg::Options::=--force-confold\"     install 'openjdk-8-jdk=7u151-*'' failed: E: Version '7u151-*' for 'openjdk-8-jdk' was not found\n", "rc": 100, "stderr": "E: Version '7u151-*' for 'openjdk-8-jdk' was not found\n", "stderr_lines": ["E: Version '7u151-*' for 'openjdk-8-jdk' was not found"], "stdout": "Reading package lists...\nBuilding dependency tree...\nReading state information...\n", "stdout_lines": ["Reading package lists...", "Building dependency tree...", "Reading state information..."]}

PR #2281, opened 2017-09-21
Specify Java major version and flavor

jwalgran:
## Overview

Provisioning the services VM was failing because the default `java_major_version` set by the `azavea.java` role is 8 and we specify a `java_version` from the 7 series.

While making this change we also specify an explicit `java_flavor` to avoid relying on the shared role default value.

Connects #2279

## Testing Instructions

 * Destroy any existing `services` VM and `vagrant up services`.


rajadain:
Taking a look now.

Issue #2217, opened 2017-08-31
UTFGrids not loaded for Select Boundary when zoomed out


mmcfarland:
## Observed Behavior
When zoomed out past the minimum zoom level for a given boundary layer (eg, z < 10 for HUC 12), the user can still activate the boundary layer from the Draw panel.  No features or errors show.  If you zoom in, the layer does appear, but is not selectable because the grids haven't loaded.

## Expected Behavior
If the user is not within the zoom level criteria, it should not be an option to select that layer.  It should be visibly disabled in the UI with an indication that they must zoom in.

mmcfarland:
As part of this, not necessarily as a resolution, we could probably increase the zoom levels for these layers.  When they were first implemented, the rendered tiles were not cached, and they are now, indefinitely.

PR #2289, opened 2017-09-25
Disable Select Boundary If Below Min Zoom + Lower Boundary Min Zoom

arottersman:
## Overview

Before you were able to use the boundary selection draw tools even if the boundary layer you were selecting couldn't be accessed at the current zoom level. This PR fixes the disabling of the selection tools on zoom, closes out active tools if a zoom event makes them inaccessible, and decreases the min zoom for the boundary layers so they can be accessed at higher altitudes.

Connects #2217 

### Demo

ch94ywf8xn

## Testing Instructions

 * Pull this branch, and `bundle`
* Zoom around the map with select boundary tool open, and confirm layers are disabled when you zoom out passed their min zoom level
* Select a boundary tool and zoom way out — it should deselect when you zoom past its min zoom

### DIFF
--- a/src/mmw/js/src/draw/templates/drawTool.html
+++ b/src/mmw/js/src/draw/templates/drawTool.html
@@ -13,6 +13,17 @@
         <ul class="dropdown-menu draw-pane-list">
             {% for item in toolData.items %}
                 <li class="draw-pane-list-item" role="presentation">
+                    <button class="draw-pane-list-item-button" role="button" tabindex="-1"
+                          data-min-zoom="{{ item.minZoom }}"
+                          href="#" id="{{ item.id }}"
+                          {%- if currentZoomLevel < item.minZoom %}
+                          disabled
+                          {% endif -%}>
+                        {{ item.title }}
+                        {% if currentZoomLevel < item.minZoom %}
+                        &mdash; Zoom in to use
+                        {% endif %}
+                    </button>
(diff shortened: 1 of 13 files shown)


src/mmw/js/src/draw/templates/drawTool.html
```diff
@@ -13,18 +13,24 @@
         <ul class="dropdown-menu draw-pane-list">
             {% for item in toolData.items %}
                 <li class="draw-pane-list-item" role="presentation">
+                    <button class="draw-pane-list-item-button" role="button" tabindex="-1"
```
arottersman: Except for the disable styling the button should look identical as to when it was a `<span>`

kellyi:
Taking a look at this!

kellyi:
Mentioned this but it looks like the button disappears on zooming past the valid level, but when you zoom back in the previously selected grid reappears although the button doesn't:

grid-boundaries

kellyi:
Aside from the quirk noted above this is working well! I tested it out with congressional districts in Chrome, FF, Safari, and IE11.

arottersman:
Thanks for catching that, @kellyi! I've pushed a fixup to correct.

arottersman:
Thanks. Retargeted and squashed. Going to merge when the build completes.

rajadain:
Hmm this seems to have merged a couple of extra things by accident (since it wasn't rebased on top of `release/1.20.0`)

Issue #2149, opened 2017-08-15
Use HydroShare Production URL


rajadain:
Once HydroShare 1.12 is released, containing the fix for https://github.com/hydroshare/hydroshare/issues/2257, revert #2145.

This card should be ready once this URL resolves correctly: https://www.hydroshare.org/hsapi/resource/?full_text_search=water&north=47.599816005559326&west=-122.32020458774991&east=-122.30687060756253&south=47.608806399331975&coverage_type=box

PR #2290, opened 2017-09-25
Use Hydroshare Production URL

rajadain:
## Overview

We switched to using a beta / playground URL in #2145. The production URL is no longer broken, so we switch back to using it instead of the playground URL.

Connects #2149 

### Demo

image

## Testing Instructions

 * Check out this branch, and go to [:8000/?bigcz](http://localhost:8000/?bigcz)
 * Select / draw a shape, then go to the search view
 * Search for any term, such as "water", and go to the Hydroshare tab
 * Ensure you see real values, and there are no errors in the console
 * Ensure that the `search` endpoint response contains the Production URL as the `api_url`


rajadain:
Thanks for taking a look!

Issue #2241, opened 2017-09-07
Geoprocessing API: Expose unsimplified/simplified RWD Params


arottersman:
RWD will allow you to get back either the simplified or unsimplified shape. Bubble this param up to the end API user.

Depends on https://github.com/WikiWatershed/rapid-watershed-delineation/issues/73

PR #2292, opened 2017-09-25
Geoprocessing API: Enable & document `"simplify": false` value for RWD

kellyi:
## Overview

This PR enables and documents passing a `"simplify": false` parameter to RWD via the geoprocessing API. 

Connects #2241 

### Demo

screen shot 2017-09-25 at 6 14 11 pm

### Notes

The example point in the Swagger docs currently returns a really large watershed from the nhd endpoint. While I think that's desirable behavior, the JSON response seems to be so large that it causes the Swagger UI to drag to a halt. It doesn't do the same when I post to :5000/rwd-nhd directly.

Wonder if we should consider changing the demo point?

## Testing Instructions
- get this branch, then set up RWD
- visit `/api/docs` and verify that RWD works with and without the new `simplify` parameter and that `simplify` works as described in the docs.

### DIFF
--- a/src/mmw/apps/geoprocessing_api/tasks.py
+++ b/src/mmw/apps/geoprocessing_api/tasks.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import
 
 import os
 import logging
+import urllib
 
 from ast import literal_eval as make_tuple
 from calendar import month_name
@@ -35,7 +36,7 @@ CM_PER_MM = 0.1
 
 
 @shared_task
-def start_rwd_job(location, snapping, data_source):
+def start_rwd_job(location, snapping, simplify, data_source):
     """
     Calls the Rapid Watershed Delineation endpoint
     that is running in the Docker container, and returns
@@ -47,9 +48,21 @@ def start_rwd_job(location, snapping, data_source):
     rwd_url = 'http://%s:%s/%s/%f/%f' % (RWD_HOST, RWD_PORT, end_point,
                                          lat, lng)
 
+    params = {}
+
     # The Webserver defaults to enable snapping, uses 1 (true) 0 (false)
     if not snapping:
-        rwd_url += '?snapping=0'
+        params['snapping'] = 0
+
+    # RWD also defaults to simplify the shape according to a tolerance.
+    # Passing it `?simplify=0` returns the unsimplified result.
+    if simplify is not False:
+        params['simplify'] = simplify
+
+    query_string = urllib.urlencode(params)
+
+    if query_string:
+        rwd_url += ('?%s' % query_string)
 
     logger.debug('rwd request: %s' % rwd_url)
 
--- a/src/mmw/apps/geoprocessing_api/views.py
+++ b/src/mmw/apps/geoprocessing_api/views.py
@@ -57,11 +57,17 @@ def start_rwd(request, format=None):
 
     `snappingOn` (`boolean`): Snap to the nearest stream? Default is false
 
+    `simplify` (`number`): Simplify tolerance for delineated watershed shape in
+                response. Use `0` to receive an unsimplified shape. When this
+                parameter is not supplied, `simplify` defaults to `0.0001` for
+                "drb" and is a function of the shape's area for "nhd".
+
     **Example**
 
         {
-            "location": [39.97185812402583,-75.16742706298828],
+            "location": [39.67185812402583,-75.76742706298828],
             "snappingOn": true,
+            "simplify": 0,
             "dataSource":"nhd"
         }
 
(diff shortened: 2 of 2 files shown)


src/mmw/apps/geoprocessing_api/tasks.py
```diff
@@ -47,9 +47,18 @@ def start_rwd_job(location, snapping, data_source):
...
     if not snapping:
-        rwd_url += '?snapping=0'
+        query_string += 'snapping=0&'
+
+    # RWD also defaults to simplify the shape according to a tolerance.
+    # Passing it `?simplify=0` returns the unsimplified result.
+    if not simplify:
+        query_string += 'simplify=0'
+
+    rwd_url += '' if query_string == '?' else qs
```
kellyi: Oof, forgot to change `qs` -> `query_string` here.

src/mmw/apps/geoprocessing_api/tasks.py
```diff
@@ -47,9 +47,18 @@ def start_rwd_job(location, snapping, data_source):
     rwd_url = 'http://%s:%s/%s/%f/%f' % (RWD_HOST, RWD_PORT, end_point,
                                          lat, lng)
 
+    query_string = '?'
```
arottersman: Now that there's more than one param, consider using [`urllib.urlencode()`](https://docs.python.org/2/library/urllib.html#urllib.urlencode)

kellyi:
> Looks good, as discussed in person we might want to allow numeric simplify values over the boolean

Sounds good to me. I'll make the adjustment & push an update.

src/mmw/apps/geoprocessing_api/tasks.py
kellyi: Sounds good! Added in e6bc740

kellyi:
e6bc740 adjusts the url params to use urllib.urlencode & also changes the type of the `simplify` param from bool to number.

src/mmw/apps/geoprocessing_api/tasks.py
```diff
@@ -47,18 +48,21 @@ def start_rwd_job(location, snapping, simplify, data_source):
...
     # The Webserver defaults to enable snapping, uses 1 (true) 0 (false)
     if not snapping:
-        query_string += 'snapping=0&'
+        params['snapping'] = 0
 
     # RWD also defaults to simplify the shape according to a tolerance.
     # Passing it `?simplify=0` returns the unsimplified result.
-    if not simplify:
-        query_string += 'simplify=0'
+    if simplify:
```
arottersman: `0` is falsey, so it will never get added as a param. I think we can just add `simplify` to the param list regardless of value and the RWD service will handle the null?

kellyi: Ooof. I pushed a commit to change this to

```py
if not simplify is False:
```

which may merit some parentheses to indicate the grouping.

src/mmw/apps/geoprocessing_api/tasks.py
```diff
@@ -47,9 +48,21 @@ def start_rwd_job(location, snapping, data_source):
...
+    params = {}
+
     # The Webserver defaults to enable snapping, uses 1 (true) 0 (false)
     if not snapping:
-        rwd_url += '?snapping=0'
+        params['snapping'] = 0
+
+    # RWD also defaults to simplify the shape according to a tolerance.
+    # Passing it `?simplify=0` returns the unsimplified result.
+    if not (simplify is False):
```
mmcfarland: For easier reading, I think this can be simply:
`if simplify is not False`

kellyi: Pushed c8242db to make this change.

kellyi:
> We also may need to guard against sending such big responses back

FWIW the large shape loaded totally fine when I got it via Insomnia, so there may be some slowness in the Swagger UI that wouldn't be there for other clients.

mmcfarland:
> some slowness in the Swagger UI

I'm sure that swagger is parsing the JSON response type and loading it to the page, which will be rough.  Is the response getting appropriately gzipped?

kellyi:
Responses are getting gzipped:

```
Server: nginx
Date: Wed, 27 Sep 2017 14:54:44 GMT
Transfer-Encoding: chunked
Connection: keep-alive
Vary: Accept
Allow: OPTIONS, GET
Content-Encoding: gzip
```

Even so the size of the unsimplified response for the demo point is 309k. Going to select a different demo point which generates a smaller watershed before merging.

kellyi:
Changed two numbers in the sample input point and now the unsimplified response is 5.3k. Going to merge this in once the tests pass.

Issue #2298, opened 2017-09-26
MapShed "Add changes to this area button" not shown when user is logged in


kellyi:
The MapShed "Add changes to this area" button currently only shows up when the user is not logged in to the app.

If an aoi is modeled while the user is not logged in, the button will appear.

Once the user logs in, the button disappears at almost the same moment the project url goes from `/project/` to `/project/94/scenario/268` or similar.

caseycesari:
I'm not able to fix this because MapShed will not complete locally for me, but I know what the issue is.

The problem is the [gis_data](https://github.com/WikiWatershed/model-my-watershed/blob/2b17268dc31e6dd07e50c4d8b1cf349ca6fb0a1b/src/mmw/js/src/modeling/templates/gmsDownloadForm.html#L5) variable that contains the output for the GMS export contains quotes and becomes incorrectly escaped:

<img width="471" alt="screen shot 2017-10-03 at 11 25 02 am" src="https://user-images.githubusercontent.com/1042475/31134242-8c851a9a-a82f-11e7-95cf-9f2013ff3088.png">

*It looks like it's incorrectly escaped from the beginning*

<img width="522" alt="screen shot 2017-10-03 at 11 24 38 am" src="https://user-images.githubusercontent.com/1042475/31134241-8c83bdd0-a82f-11e7-8dbb-15da1d068108.png">

*You can see the browser is trying to parse the data into HTML elements*

arottersman:
This was fixed by https://github.com/WikiWatershed/model-my-watershed/pull/2309, but we hadn't run migrations yet. I've just run them. @caseycesari mind taking a quick look on staging and confirming the issue has been resolved?

caseycesari:
Confirmed! It's fixed.

Issue #2296, opened 2017-09-26
Saved MapShed projects won't reload


kellyi:
After saving a MapShed project then trying to reload it by refreshing the page with the same URL, staging reports a 500 server error and shows this screen:

<img width="1216" alt="screen shot 2017-09-26 at 12 45 05 pm" src="https://user-images.githubusercontent.com/4165523/30872617-b47da35e-a2b8-11e7-9076-d9532a0730af.png">

This seems only to happen for MapShed & not TR-55. To replicate:

- log in
- select an arbitrary aoi
- run MapShed
- when the run completes and has saved with the scenario URL, refresh the browser to reload the page

jwalgran:
Error response

```
ValueError at /mmw/modeling/projects/1
Expecting property name enclosed in double quotes: line 1 column 2 (char 1)

Traceback:
File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py" in get_response
  132.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/usr/local/lib/python2.7/dist-packages/django/views/decorators/csrf.py" in wrapped_view
  58.         return view_func(*args, **kwargs)
File "/usr/local/lib/python2.7/dist-packages/django/views/generic/base.py" in view
  71.             return self.dispatch(request, *args, **kwargs)
File "/usr/local/lib/python2.7/dist-packages/rest_framework/views.py" in dispatch
  452.             response = self.handle_exception(exc)
File "/usr/local/lib/python2.7/dist-packages/rest_framework/views.py" in dispatch
  449.             response = handler(request, *args, **kwargs)
File "/usr/local/lib/python2.7/dist-packages/rest_framework/decorators.py" in handler
  50.             return func(*args, **kwargs)
File "/opt/app/apps/modeling/views.py" in project
  78.         return Response(serializer.data)
File "/usr/local/lib/python2.7/dist-packages/rest_framework/serializers.py" in data
  466.         ret = super(Serializer, self).data
File "/usr/local/lib/python2.7/dist-packages/rest_framework/serializers.py" in data
  213.                 self._data = self.to_representation(self.instance)
File "/usr/local/lib/python2.7/dist-packages/rest_framework/serializers.py" in to_representation
  435.                 ret[field.field_name] = field.to_representation(attribute)
File "/opt/app/apps/modeling/serializers.py" in to_representation
  17.         return json.loads(obj)
[...]

Issue #2294, opened 2017-09-26
GMS export not working


kellyi:
Trying to export a GMS on staging currently results in a 500 error; on my local the error is:

![screen shot 2017-09-26 at 11 43 48 am](https://user-images.githubusercontent.com/4165523/30869867-4382155c-a2b0-11e7-97dd-f48e6efb451f.png)

Issue #2268, opened 2017-09-19
User dropdown is hidden by conservation practice button


mmcfarland:
In Multi-Year modeling, the logged-in user dropdown is obscured by other UI elements.

![screenshot from 2017-09-19 17 00 40](https://user-images.githubusercontent.com/1014341/30615205-27d250aa-9d5c-11e7-8d86-ca87ddd26813.png)

PR #2299, opened 2017-09-26
Adjust user dropdown zindex

kellyi:
## Overview

This PR adjusts the z-index for the user dropdown menu to set it higher than MapShed's conservation practice button: the button is 1002, so we set the dropdown to 1010.

Connects #2268 

### Demo

screen shot 2017-09-26 at 12 54 54 pm

### Notes

While working on this I encountered #2294, #2296, and #2298. I started down the path of working on #2298 but I *think* all of these issues may have a similar cause and probably merit tackling together. It might be something related to the new JSON format.

## Testing Instructions
- get this branch, `bundle`, then run a mapshed job when not logged in
- when the job returns, click "Add changed to this area" to reveal the "Conservation Practice" button
- sign in to the app
- toggle the user menu in the top left open and verify that the menu now sits above the "Conservation Practice" button

### DIFF
--- a/src/mmw/sass/base/_header.scss
+++ b/src/mmw/sass/base/_header.scss
@@ -53,6 +53,7 @@ header {
         .dropdown-menu {
           left: auto;
           right: 0px;
+          z-index: 1010;
           a {
             color: $ui-secondary;
 


rajadain:
I also think at some point we should do a massive `z-index` usage review of this project and see if it can be consolidated somehow.

kellyi:

Issue #2242, opened 2017-09-07
Don't show scenario pagination arrows if all scenarios are already shown on page


ajrobbins:
Currently, the compare view pagination arrows are on screen at all times. We should make it so they don't show unless there are more scenarios than the 3 that is our pagination limit.

Issue #2236, opened 2017-09-07
Compare View Revamp: Dynamic Minimum Visible Scenarios


rajadain:
Currently, the minimum number of visible scenarios is [hard-coded to 3](https://github.com/WikiWatershed/model-my-watershed/blob/a22aef004c78d1e9c9849250aa3a707eea77c63a/src/mmw/js/src/compare/models.js#L10). This makes certain interactions awkward, such as this sliding even when all the scenarios fit comfortably in view:

![2017-09-07 13 29 23](https://user-images.githubusercontent.com/1430060/30176559-d82b8c5e-93d0-11e7-92fe-bfd198a20e09.gif)

Make it so the minimum number of visible scenarios is calculated dynamically based on how many can fit in the visible window.

PR #2301, opened 2017-09-26
Show compare pagination buttons only if > 5 scenarios

kellyi:
## Overview

This PR updates the compare view to show the pagination buttons only if there are more than the minimum number of scenarios -- a value which is currently set to 3.

(And was adjusted to 5 for this PR!)

Connects #2242 

### Demo

screen shot 2017-09-26 at 2 28 00 pm

->

screen shot 2017-09-26 at 2 28 13 pm

## Testing Instructions
- get this branch, bundle, then create a TR-55 project
- create one new scenario, then launch compare view and verify that you don't see the pagination buttons
- create another new scenario, then launch compare view and verify that you do see the pagination buttons and that they work as before
- delete one scenario, launch compare again, and verify that you don't see the pagination buttons

### DIFF
--- a/src/mmw/js/src/compare/models.js
+++ b/src/mmw/js/src/compare/models.js
@@ -7,7 +7,7 @@ var _ = require('lodash'),
 
 var CHART = 'chart',
     TABLE = 'table',
-    MIN_VISIBLE_SCENARIOS = 3,
+    MIN_VISIBLE_SCENARIOS = 5,
     CHART_AXIS_WIDTH = 82,
     COMPARE_COLUMN_WIDTH = 134;
 
--- a/src/mmw/js/src/compare/views.js
+++ b/src/mmw/js/src/compare/views.js
@@ -66,16 +66,21 @@ var CompareWindow2 = modalViews.ModalBaseView.extend({
             prevButton = this.ui.prevButton,
             nextButton = this.ui.nextButton;
 
-        if (i < 1) {
-            prevButton.removeClass('active');
+        if (total <= minScenarios) {
+            prevButton.hide();
+            nextButton.hide();
         } else {
-            prevButton.addClass('active');
-        }
+            if (i < 1) {
+                prevButton.removeClass('active');
+            } else {
+                prevButton.addClass('active');
+            }
 
-        if (i + minScenarios >= total) {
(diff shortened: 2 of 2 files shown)


rajadain:
Instead of this, I would vote for the arrow buttons to only be shown if "active". Thus, when at the left-most position, the left arrow should not be seen at all.

Also, I think we should up the minimum number of displayed scenarios to 4 or 5. What's the smallest screen we expect our users to have?

/cc @jfrankl @ajrobbins

rajadain:
> In addition, instead of making the sliding buttons "inactive" when they are no longer clickable, consider just hiding them.

Looks like this bit is already in #2236, so let's not do that here.

Also, the iPad Mini and the iPad have the same resolution, so if 5 works on a regular let's go with that, because it'll work the same on the Mini.

kellyi:
Just to clarify: should I go to 4 or 5 for the MIN_SCENARIOS? Seems like 5 but just want to confirm.

ajrobbins:
5 it is!

rajadain:
Actually, let's do 4. Just tested on iPad landscape:

image

iPad portrait:

image

rajadain:
If we want to support iPad portrait, we can also leave it at 3.

ajrobbins:
No portrait

rajadain:
In that case, we can go with five, since that fits:

image

ajrobbins:
Where did you get this image https://user-images.githubusercontent.com/1430060/30881959-7c031a3c-a2d5-11e7-9897-916986a6dd24.png?

rajadain:
Took a screenshot from my iOS Simulator running iPad 2

caseycesari:
Sorry! I somehow missed this assignment. Will take a look now.

caseycesari:
Not sure I'm following the conversation here correctly, or if the description has been updated, but I'm seeing active buttons with 4 scenarios:

image

caseycesari:
Doesn't have to be addressed here, but we should consider dropping the focus highlight from the buttons:

image

kellyi:
> I'm seeing active buttons with 4 scenarios

Looking into this.

kellyi:
Thanks! Going to merge this one in.

Issue #2293, opened 2017-09-25
Base Flow Index incorrectly calculated for MapShed


mmcfarland:
The bfi48grd GeoTiff has values which range from 0-91.  When calculating Groundwater Recession Coefficient, the average BFI is calculated over the AoI.  This value is then used to calculate the GRC value, thusly:

```
GRC = BFI * -0.0015 + 0.1103
GRC = GRC if GRC >= 0 else 0.01
```

In practice, we are writing out `0.01` for the value in the GMS too frequently, which indicates that GRC is less than 0, which happens only when BFI > ~74.  These are non-typical values, the normal range for GRC is `0.02` - `0.06`.

Upon inspection, it looks like the geoprocessing service is returning average values of > 100, which will result in `0.01` for a GRC, since its original value is negative.  The average of BFI is somehow too large.  Check that either the ingested raster has not had its values shifted, or that the averaging function is not deficient.  The original GeoTiff on the fileshare appears accurate.

```
> gdalinfo -mm bfigrd48
STATISTICS_MAXIMUM=91
STATISTICS_MEAN=42.691618734836
STATISTICS_MINIMUM=0
STATISTICS_STDDEV=19.46358279968
```

### Sample HUC
Name: City of Philadelphia-Schuylkill River, HUC-12 Subwatershed
WKAoI: `huc12__55174`
[RasterGroupedAverage value](https://github.com/WikiWatershed/model-my-watershed/blob/ac2f981f95736bac2edcfeb9f11bfc95dc88c5a6/src/mmw/apps/modeling/mapshed/tasks.py#L350-L353) (BFI): `152.222222222`

This average value is seemingly higher than even the maximum value in the source raster.  A crude check against the tif shows the expected values to be in the mid-40s range.

mmcfarland:
One additional data point, it seems like the Averaging job is correct, but that the raster has inflated values.  A RasterGroupedCount of the same HUC from above counts a single cell of `152`.  A grouped count of the containing HUC 10 shows:

```
{
u'List(163)': 3, 
u'List(171)': 5,
u'List(133)': 1, 
u'List(197)': 1,
u'List(178)': 2,
u'List(152)': 3,
u'List(159)': 2,
u'List(186)': 1,
u'List(148)': 3,
u'List(167)': 6, 
u'List(175)': 5,
u'List(156)': 1, 
u'List(182)': 3, 
u'List(140)': 1
}

```

arottersman:
[...]

PR #2263, opened 2017-09-18
Estimate SedPhos And Recession Coefficient From Rasters

arottersman:
## Overview

- Use the `soilpallland2` raster in the datahub to estimate the mapshed `SedPhos`/`SoilP` value
- Use the `bfi48grd` raster in the datahub to estimate the mapshed `RecessionCoef` value

Connects #1752
Connects #1756  

### Demo
Resulting GMS file

[panther_skin_creek.gms.txt](https://github.com/WikiWatershed/model-my-watershed/files/1311837/panther_skin_creek.gms.txt)

Line 2, pos 2 is not `0.06`
```
1.4.0,0.01,0,10,0,0,0.12504566328764893,13.825475584291853,62287,24254.230851633638,38032.769148366366,1914.431863640184,0,8.460119315139668,0.021151433681949956,28,1963,1990,0.0004415567779477565,7404.395733781011,0,0,0,0.3584968896652178
```

Line 37
```
701.52,816.6336633663367,0.7774450661770334,0.01793083008186061,0.25,0.25
```

## Testing Instructions

- Check out branch
- Restart Celery `vagrant ssh worker -c 'sudo service celeryd restart'`
- Model a WkAoI that isn't cached on the branch and on staging
- Compare line 2, pos 2 and line 37 pos 2 across staging and this branch. They should be different now.

PR #2302, opened 2017-09-26
Update Mapshed bfi48grd and soilpallland2 rasters

arottersman:
## Overview

  The layers underlying the Recession Coefficient and SoilP estimates were made
  from geotiffs incorrectly converted from their original arcgis format -- this
  created layers with bad values. Update the layers to newly ingested,
  corrected layers

Connects #2293

### Demo

|               | Expected    | Approximate     |
|---------------|-------------|-----------------|
| soilpallland2 | 596.992     | 569.380416422   |
| bfi48grd      | 40.71185983 | 42.691618734836 |

Expected values show the output mean of the original rasters. Approximate values are created
by running the grouped average with a very simple CONUS shape. The CONUS shape included a little ocean which decreases the average a bit — you can see in the table though they're close enough we can assume correctness.

You can apply this patch to run the test. [national_test.patch.txt](https://github.com/WikiWatershed/model-my-watershed/files/1334897/national_test.patch.txt)

## Testing Instructions

 * Pull this branch and `vagrant ssh worker -c 'sudo service celeryd restart'`
 * Run some mapshed jobs and check the values of the exported GMS values per the instructions in #2263


rajadain:
Having trouble exporting GMS files (#2294) but I would agree with the assessment of replacing existing rasters with the corrected versions.

PR #2306, opened 2017-09-27
Geoprocessing API: Add Client App API Token

arottersman:
## Overview

Implements what's discussed in #2280. We're creating a user for the client app so that it can use its own token to make requests to the geoprocessing API without having to worry if a UI user is logged in or not.

- Adds migration to create a "dummy" user for the client app
- Adds the dummy user for client app's token to `clientSettings`
    - Added password to the `envdir`
- Adds `Authorization` header to all Geoprocessing API requests
- Secures Geop API endpoints so they accept only token authenticated requests

Connects #2270 

### Demo

screen shot 2017-09-27 at 3 45 37 pm

## Notes

This will require running migrations on staging (once staging is `develop` again). We'll have to remember to do it after the release.

## Testing Instructions

 * Pull this branch
 * `vagrant provision app` — you should see `CLIENT_APP_USER_PASSWORD` print in the logs with the other environment variables. You can confirm it set correctly on your local via `vagrant ssh app -c 'cat /etc/mmw.d/env/MMW_CLIENT_APP_USER_PASSWORD'`
* `./scripts/manage.sh migrate`
* Note the token you get back from `http --json POST 'http://localhost:8000/api/api-token-auth/' username="mmw|client_app_user" password="mmw"`
* `./scripts/bundle.sh`
* Analyze some shapes and delineate some watersheds via the UI 
   - `Authorization` header should be populated with the client app's token regardless if you are logged in or "guest"
* Confirm you can still model shapes and save projects
* Confirm swagger and curl requests require a token and that you're still able to use your own user's token


arottersman:
Thanks! Updated the deployment configs. Going to merge.

PR #2308, opened 2017-09-28
Fix gis_data serialization

jwalgran:
## Overview

We recently changed the `ProjectSerializer` to use a `JsonField` serializer for the `gis_data` field to reflect that fact that this data is now returned from the geoprocessing service as an actual JSON object rather than a string containing JSON data. Because this field serializer change was not also made to the `ProjectUpdateSerializer` we ended up saving the string representation of a Python dict into the database rather than a JSON string. This caused the model to crash when attempting to load a previously saved project.

I took advantage of the fact that `ast.literal_eval` works for both JSON and Python dict text formats to write a simple corrective data migration. Brief description of what this PR does, and why it is needed.

Connects #2296 
Connects #2294

### Notes

I tried to make `ProjectUpdateSerializer` inherit from `ProjectSerializer` but it caused problems with scenario serialization.

## Testing Instructions

 * Before checking out this branch, run a MapShed model on a drawn AOI.
 * Verify that the GMS file fails to download.
 * Refresh the project and verify that it does not load correctly.  
 * Remember the project ID.
 * Check out the branch and run migrations.
 * Reload the previously created project and verify that it loads correctly.
 * Verify that the GMS file download works correctly.
 * Run a new MapShed model, refresh the project, and verify that it also loads correctly and the GMS file download works.

### DIFF
--- a/deployment/ansible/group_vars/all
+++ b/deployment/ansible/group_vars/all
@@ -33,6 +33,8 @@ nodejs_npm_version: 2.1.17
 apache_version: "2.4.7-*"
 
 java_version: "7u151-*"
+java_major_version: "7"
+java_flavor: "openjdk"
 
 graphite_carbon_version: "0.9.13-pre1"
 graphite_whisper_version: "0.9.13-pre1"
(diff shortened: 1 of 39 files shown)


jwalgran:
I will assign this after fixing the unit tests.

jwalgran:
Replaced by #2309

PR #2309, opened 2017-09-28
Fix gis_data serialization

jwalgran:
## Overview

We recently changed the `ProjectSerializer` to use a `JsonField` serializer for the `gis_data` field to reflect that fact that this data is now returned from the geoprocessing service as an actual JSON object rather than a string containing JSON data. Because this field serializer change was not also made to the `ProjectUpdateSerializer` we ended up saving the string representation of a Python dict into the database rather than a JSON string. This caused the model to crash when attempting to load a previously saved project.

Adding a plain `JsonField()` to the `ProjectUpdateSerializer` caused test failures that were resolved by adding attributes signifying that it is a nullable field, which matches the underlying model. I changed both usages of `JsonField` for consistency.

Connects #2296 
Connects #2294

### Notes

I tried to make `ProjectUpdateSerializer` inherit from `ProjectSerializer` but it caused problems with scenario serialization.

## Testing Instructions

 * Before checking out this branch, run a MapShed model on a drawn AOI.
 * Verify that the GMS file fails to download.
 * Refresh the project and verify that it does not load correctly.  
 * Remember the project ID.
 * Check out the branch and run migrations.
 * Reload the previously created project and verify that it loads correctly.
 * Verify that the GMS file download works correctly.
 * Run a new MapShed model, refresh the project, and verify that it also loads correctly and the GMS file download works.

### DIFF
--- a/src/mmw/apps/modeling/migrations/0023_fix_gis_data_serialization.py
+++ b/src/mmw/apps/modeling/migrations/0023_fix_gis_data_serialization.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+import ast
+import json
+
+from django.db import migrations
+
+
+def fix_gis_data_serialization(apps, schema_editor):
+    Project = apps.get_model('modeling', 'Project')
+    for project in Project.objects.filter(gis_data__startswith='{u'):
+        project.gis_data = json.dumps(ast.literal_eval(project.gis_data))
+        project.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('modeling', '0022_project_wkaoi'),
+    ]
+
+    operations = [
+         migrations.RunPython(fix_gis_data_serialization,
+                              migrations.RunPython.noop)
+    ]
--- a/src/mmw/apps/modeling/serializers.py
+++ b/src/mmw/apps/modeling/serializers.py
@@ -40,7 +40,7 @@ class ProjectSerializer(gis_serializers.GeoModelSerializer):
         geo_field = 'area_of_interest'
 
     user = UserSerializer(default=serializers.CurrentUserDefault())
-    gis_data = JsonField()
+    gis_data = JsonField(required=False, allow_null=True)
     scenarios = ScenarioSerializer(many=True, read_only=True)
 
 
@@ -61,3 +61,4 @@ class ProjectUpdateSerializer(gis_serializers.GeoModelSerializer):
 
     user = UserSerializer(default=serializers.CurrentUserDefault(),
                           read_only=True)
+    gis_data = JsonField(required=False, allow_null=True)


src/mmw/apps/modeling/migrations/0023_fix_gis_data_serialization.py
```diff
@@ -0,0 +1,29 @@
...
+
+import ast
+import json
+
+from django.db import migrations
+
+
+def fix_gis_data_serialization(apps, schema_editor):
+    Project = apps.get_model('modeling', 'Project')
+    for project in Project.objects.all():
```
mmcfarland: There are 18,574 projects in prod & 933 in staging. Can you anticipate how the migration will might perform in those environments?  My suspicion is that we don't need to run this migration in prod, since the initiating commit hasn't been deployed, but I'm not sure there is a good mechanism to skip it, other than manually inserting the migration record so it appears to have run.

rajadain: I just ran this query on staging:

```sql
SELECT _left, COUNT(_left)
FROM (
    SELECT LEFT(gis_data, 10) AS _left
    FROM modeling_project
    WHERE gis_data IS NOT NULL) AS x
GROUP BY _left;

   _left    | count
------------+-------
 {"n38": 0, |    80
 {u'n38': 0 |    14
(2 rows)
```

Not sure which one is which, but we can probably use a similar test to significantly limit the operation.

rajadain: I think this is old:

```sql
SELECT LEFT(gis_data, 10) AS _left
FROM modeling_project
WHERE gis_data IS NOT NULL
ORDER BY created_at ASC
LIMIT 5;

   _left
------------
 {"n38": 0,
 {"n38": 0,
 {"n38": 0,
 {"n38": 0,
 {"n38": 0,
(5 rows)
```

and this is new:

```sql
SELECT LEFT(gis_data, 10) AS _left
FROM modeling_project
WHERE gis_data IS NOT NULL
ORDER BY created_at DESC
LIMIT 5;

   _left
------------
 {u'n38': 0
 {u'n38': 0
 {u'n38': 0
 {u'n38': 0
 {u'n38': 0
(5 rows)
```

Note: the sort order is opposite in the above queries.

jwalgran: I pushed a FIXUP commit that alters the migration to be more efficient, and will make it a noop in production, where there are no incorrectly serialized records.

jwalgran:
I pushed a FIXUP commit that alters the migration to be more efficient, and will make it a noop in production, where there are no incorrectly serialized records.

### Testing the Fixup

- Check out this branch and rollback the migration `./scripts/manage.sh migrate modeling 0022`
- Check out the `release/1.20.0` branch
- Run a MapShed model on a drawn AOI and verify that reloading the project fails and GMS fails to download
- Checkout this branch and run migrations
- Reload the previously created project and verify that it loads correctly.
- Verify that the GMS file download works correctly.
- Run a new MapShed model, refresh the project, and verify that it also loads correctly and the GMS file download works.

src/mmw/apps/modeling/migrations/0023_fix_gis_data_serialization.py
```diff
@@ -9,10 +9,8 @@
 
 def fix_gis_data_serialization(apps, schema_editor):
     Project = apps.get_model('modeling', 'Project')
-    for project in Project.objects.all():
-        if project.gis_data:
-            # literal_eval works for both JSON and str(dict)
-            # https://docs.python.org/2.7/library/ast.html#ast.literal_eval
+    for project in Project.objects.exclude(gis_data=''):
+        if "u'" in project.gis_data:
```
rajadain: This `if` can probably be included in the filter like so:

```python
for project in Project.objects.filter(gis_data__startswith='{u'):
    project.gis_data = json.dumps(ast.literal_eval(project.gis_data))
    project.save()
```

See [startswith](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#std:fieldlookup-startswith)

rajadain:
Taking a look at this now

src/mmw/apps/modeling/migrations/0023_fix_gis_data_serialization.py
jwalgran: Thanks. I pushed another commit with this more efficient query.

The count of matching projects on staging

```
>>> Project.objects.filter(gis_data__startswith='{u').count()
14
```

Matches the count from this query on staging

```
SELECT _left, COUNT(_left)
FROM (
    SELECT LEFT(gis_data, 10) AS _left
    FROM modeling_project
    WHERE gis_data IS NOT NULL) AS x
GROUP BY _left;

   _left    | count
------------+-------
 {"n38": 0, |    80
 {u'n38': 0 |    14
(2 rows)
```

Issue #2303, opened 2017-09-27
Map interaction broken after modeling


mmcfarland:
### Problem
All map interactivity, including zooming and panning is broken after running through a model and returning to the main screen.  Observed on staging @ f4ee3ee7f9c8aa5ccb58a4139a0f7c0557662e29.

#### Steps to reproduce
1. Select an area to analyze
1. Select Multi-Year model
1. When the model completes, click "New Project"
1. You are returned to the select AoI screen, and the map cannot be interacted with.

There don't appear to be JS errors in the console.

mmcfarland:
Looks like similar behavior can happen from just going to Analyze, then Change Area.

arottersman:
When going from Analyze -> Change Area -> Sq Kilometer -> Analyze, it freezes, and I see this error in the console:
```
vendor.js:76085 Uncaught TypeError: Cannot read property 'getLatLng' of undefined
    at NewClass._fireCreatedEvent (vendor.js:76085)
    at NewClass._onClick (vendor.js:76085)
    at NewClass.fireEvent (vendor.js:77268)
    at NewClass._fireMouseEvent (vendor.js:79019)
    at NewClass._onMouseClick (vendor.js:78999)
    at HTMLDivElement.handler [as _leaflet_click1] (vendor.js:83198)
_fireCreatedEvent @ vendor.js:76085
_onClick @ vendor.js:76085
fireEvent @ vendor.js:77268
_fireMouseEvent @ vendor.js:79019
_onMouseClick @ vendor.js:78999
handler @ vendor.js:83198
```

I don't notice the same for free draw.

ajrobbins:
I'm observing the freezing behavior when I simply select an AoI via any method, and proceed to Analyze the first time.

arottersman:
Related (?) Base map gets stuck in big cz, but zooming still works for feature group:
![artgalt5r5](https://user-images.githubusercontent.com/7633670/30970869-4a9e90b0-a434-11e7-8b08-88f9dddedc07.gif)

PR #2310, opened 2017-09-28
Fix Map Interaction

rajadain:
## Overview

Apparently by using listenTo on non-Backbone events caused all pre-existing listeners on that event to be removed or corrupted. By using the common "on" pattern we achieve the same behavior while maintaining map interaction.

Connects #2303 

### Notes

I'm not quite sure _why_ this is working. There seem to [be other](https://github.com/WikiWatershed/model-my-watershed/blob/6476f2338921b53217c75f21ddc2d590850c2018/src/mmw/js/src/core/views.js#L399) [use cases](https://github.com/WikiWatershed/model-my-watershed/blob/6476f2338921b53217c75f21ddc2d590850c2018/src/mmw/js/src/core/views.js#L423)  where this works. But since we remove the listeners when the views are destroyed, we effectively achieve the same functionality.

## Testing Instructions

 * Select an area to analyze
 * Ensure you can move and zoom the map in draw/, analyze/, and project/ stages

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -507,7 +507,10 @@ var DrawToolBaseView = Marionette.ItemView.extend({
 
     initialize: function(options) {
         this.resetDrawingState = options.resetDrawingState;
-        var self = this;
+        this.onMapZoom = _.bind(this.onMapZoom, this);
+
+        var self = this,
+            map = App.getLeafletMap();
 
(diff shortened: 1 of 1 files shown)


rajadain:
In the interest of fixing staging quickly, I'm going to take the one approval I have and merge this.

arottersman:
Closed just as I was leaving my plus 1 :(

rajadain:
Very GRRM

Issue #2282, opened 2017-09-21
BiG-CZ: Implement Multiple Overlapping Popup


rajadain:
For when there are multiple overlapping boxes (in CINERGI), or points (in WDC), clicking the geometry on the map should not open a popup for just the top one. Instead, it should open a popup listing all the shapes, in a paginated format.

@jfrankl to provide wireframes here.

jfrankl:
![screen shot 2017-09-25 at 4 07 10 pm](https://user-images.githubusercontent.com/1809908/30828442-b23f7066-a20b-11e7-8a64-c1370ee9458c.png)
![screen shot 2017-09-25 at 4 07 16 pm](https://user-images.githubusercontent.com/1809908/30828441-b23ca908-a20b-11e7-868f-eb64388e817b.png)

arottersman:
@jfrankl should there be a way to navigate back to the list popup from the regular popup?

jfrankl:
Good idea. Could you drop a back button on the top of the popup? 
![image](https://user-images.githubusercontent.com/1809908/30985519-6bc84144-a45e-11e7-9720-2bc064cef76f.png)

PR #2314, opened 2017-09-29
BiG-CZ: List Popup

arottersman:
## Overview

When BiG-CZ shapes overlapped we were only able to display the topmost layer. Add a popup that allows browsing details of all the shapes at the clicked point.

Connects #2282 

### Demo

Multipage

h1d0kybyd2

Single page

screen shot 2017-09-29 at 5 03 44 pm

### Notes

gk1zolrtuu

I left the height dynamic for now. When there's only a single item on a page, the above height "jumping" happens. I'm not sure what's desirable @jfrankl 

## Testing Instructions

 * Pull this branch and `bundle`
 * Spend some time searching and clicking the results of various catalogs
     - If there's a single shape under your click, the details popup should show
     - If multiple, the list popup should show
     - You should still be able to highlight features by hovering over the sidebar items
     - Searching, switching catalogs, and opening the details of a result should close the popup


src/mmw/js/src/core/views.js
```diff
@@ -808,24 +812,94 @@ var MapView = Marionette.ItemView.extend({
...
+
+                        if (shape instanceof L.Circle) {
+                            return shape._point.distanceTo(clickPoint) <= shape._radius;
+                        }
+
+                        return false;
+                    },
+
+                    // Get a list of results intersecting the clicked point
+                    intersectingFeatures = _.reduce(
```
rajadain: Prefer using `_.filter` instead of `_.reduce`:

```javascript
intersectingFeatures = _.filter(e.target._layers, intersectsClickBounds);
```

src/mmw/js/src/core/views.js
```diff
@@ -808,24 +812,94 @@ var MapView = Marionette.ItemView.extend({
...
+                    layer.bindPopup(new SinglePopoverView({
+                        model: result,
+                        catalog: catalogId
+                    }).render().el, { className: 'data-catalog-popover'});
+
+                    layer.openPopup();
+                    layer.setStyle(dataCatalogActiveStyle);
+                    result.set('active', true);
+
+                    layer.on('popupclose', function() {
```
rajadain: Is this event handler taken off at any point? Or do they just pile on with more and more clicks?

rajadain:
Great job on attaching demo gifs for every use case too 👌

ajrobbins:
>In many cases, the title is not useful enough to differentiate in the first few characters that are shown for each item. I wonder if we should have two lines for the titles before (dot dot dot)ing them. This could use the input of @ajrobbins and @jfrankl

I think the existing treatment is sufficient, since we don't know what format the title will take (the examples shown here are only one kind of result)

rajadain:
This is one example of important information being cutoff:

image

the first result is for Maryland, the second for Pennsylvania, but that's hard to make out.

ajrobbins:
That's a good example of it not accommodating important information. However, the same could be true if we have two lines, since there is no standardization on the result titles coming in from the API.

src/mmw/js/src/core/views.js
arottersman: They were probably piling, good catch. I'm changing this to `once`

arottersman:
Thanks, @rajadain. Those were good points. I've pushed some additional fixups + commits to address. Properly linking the popup state and map + sidebar was unexpectedly difficult, but I think it's working now (issue described in commit message).

rajadain:
Taking another look at this.

rajadain:
With changes in the latest commit, there seems to be a regression: if a CINERGI shape is too big for the map, it's gold border (which should convert to regular blue) does not "stick" as it should when the details view is open:

rajadain:
Another one: if a popup is open and its corresponding sidebar element is (correctly) highlighted, and I zoom the map, the popup closes but the highlight gets stuck, and then can't be unhighlighted:

rajadain:
For WDC, if a popup is open for a single element, hovering on others keeps the yellow highlight of the popup-element there. That is, there are two highlighted elements on the map: one for the popup and another for the sidebar element on which the user is hovering.

For the multi-popups, this does not happen. Hovering on any element in the sidebar removes the highlight of the popup element. This should not happen:

rajadain:
I'll take another look once these issues have been addressed.

arottersman:
Will push a fix soon for the 3rd comment about WDC. I'm having trouble recreating the other two issues described.

> With changes in the latest commit, there seems to be a regression: if a CINERGI shape is too big for the map, it's gold border (which should convert to regular blue) does not "stick" as it should when the details view is open:

Here's the blue border I see on the details view:

screen shot 2017-10-05 at 8 55 18 am

> Another one: if a popup is open and its corresponding sidebar element is (correctly) highlighted, and I zoom the map, the popup closes but the highlight gets stuck, and then can't be unhighlighted

Here's what happens when I zoom

vbf3jadxvl

arottersman:
@rajadain and I looked together IRL, and he wasn't making up those bugs! 

I've squashed down the fixups and added a commit that adds a `popupclose` listener for the list popup. I think it's mostly fixed the first two issues. The third issue will require a greater refactor; I'll write up a separate card.

> With changes in the latest commit, there seems to be a regression: if a CINERGI shape is too big for the map, it's gold border (which should convert to regular blue) does not "stick" as it should when the details view is open

This was recreatable by going to the details view from the list, and then hovering over the map (the highlight would disappear). The `popupclose` commit seems to have fixed this, though the detail highlight still doesn't show up on FF until you've panned/zoomed the map. I'll make a card.

 > Another one: if a popup is open and its corresponding sidebar element is (correctly) highlighted, and I zoom the map, the popup closes but the highlight gets stuck, and then can't be unhighlighted

We were also able to recreate this together by going from list popup -> detail popup, then closing the popup. I believe it's fixed now.

> For WDC, if a popup is open for a single element, hovering on others keeps the yellow highlight of the popup-element there. That is, there are two highlighted elements on the map: one for the popup and another for the sidebar element on which the user is hovering.

> For the multi-popups, this does not happen.

This is the one I'll be making an issue for. We currently set the active styling directly on the popup layer, instead of using the map state. We can't do this for the list popup because the popup is on the map, not a specific layer.

arottersman:
@rajadain, is it worthwhile to rebase this on `develop`?

rajadain:
In the interest of not adding things to the `release/1.20.0` branch that don't belong there, I would say yes, we should rebase and retarget to `develop`. Thanks for catching that

arottersman:
Rebased on develop, and updated target.

rajadain:
Waiting on #2351 to fix the build, after which this should pass ...

rajadain:
@azavea-bot rebuild

rajadain:
Taking another look to confirm

arottersman:
Thanks for all your help with this, and uncovering those bugs.

Issue #2313, opened 2017-09-29
Download Runoff and Water Quality output in Compare View


ajrobbins:
Prior to release 1.21.0, wire up the download button at the top right of the compare view modal:
![screen shot 2017-09-29 at 4 38 57 pm](https://user-images.githubusercontent.com/12868342/31035185-1064dec0-a535-11e7-939d-cf8c44687f5e.png)

It should enable a download of the contents of the chart view:
![screen shot 2017-09-29 at 4 39 27 pm](https://user-images.githubusercontent.com/12868342/31035190-19ffff32-a535-11e7-8c07-eee942bd80e4.png)

We should use the same method we're using to download the tables in Analyze, if possible.

mmcfarland:
There is some question about whether to download both measure or just the activate one.  The placement could be interpreted either way, but there is so little information contained in either, and the distinction just a category, that we should include both Runoff and Water Quality in the same CSV when the download button is clicked.

kellyi:
> We should use the same method we're using to download the tables in Analyze, if possible.

Looks like we switched to using an "export" method available on the Analyze tab's table component at some point, so I don't think we'll be able to do that here.

Planning to set this up roughly similarly to how the prior analyze download worked and to assemble both runoff and water quality data in the same csv.

Issue #2312, opened 2017-09-29
Remove download button for Compare View


ajrobbins:
We haven't wired up this download button at the top right of the compare view modal:
![screen shot 2017-09-29 at 4 38 57 pm](https://user-images.githubusercontent.com/12868342/31035132-ddb93c32-a534-11e7-9431-1ef502ca24cc.png)

So let's remove it prior to this release.

PR #2319, opened 2017-10-02
Remove compare view download button

kellyi:
## Overview

This PR removes the download button from the compare view in prep for the next release. Since we're going to be bringing it back in in #2313, I only deleted the element and didn't touch anything in the supporting view file.

Connects #2312 

### Demo

screen shot 2017-10-02 at 4 02 24 pm

## Testing Instructions
- using Chrome, IE, FF, and Safari visit a project, open the compare view, then verify that the download button is no longer there but that the other toggle buttons are & still work.


kellyi:
Thanks!

Issue #2278, opened 2017-09-21
Compare View Revamp: Chart Tooltip Improvements


rajadain:
Following #2158 and #2245, we now have chart tooltips. However, there are two issues:

 * Tooltips for very high values are cropped out:

    ![image](https://user-images.githubusercontent.com/1430060/30705702-31e51744-9ec4-11e7-9a99-18b3038ef569.png)

 * Tooltips are shown for 0 value elements in a stacked chart:

    ![image](https://user-images.githubusercontent.com/1430060/30705739-4dd2241a-9ec4-11e7-8133-6661d1abbc05.png)

Make it so that:

 - [ ] Tooltips are always visible, no matter their position in the chart
 - [ ] 0 value tooltips are shown for single value bar charts, but not for stacked value bar charts

arottersman:
![Screen Shot 2017-10-02 at 1.58.21 PM.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/57b31ed59131fe0e00a15844/125516c66e82c728ace21e0d46db978826878de9bce1b057e40fa0c57e0b76236454f16b7191a379f117234f5d5e5ecc3c4a4356bcbbd671f3e57862804e0ee8833c5771c13811aabc8349ed05be208b4bd1b0edd919d2f42f67b6d5db1063474f4dcbed5727a19041c95df64de84b52a51b326ae1702431.png)

When there are many scenarios and you pan over with the arrows, tooltips still appear for the bars on the left.

Should this issue be addressed with this card or is it lower priority?

rajadain:
Depends on the difficulty. If it can be easily done, it should be included. Else offloaded.

arottersman:
> If it can be easily done
I was going to comment laughing at this, but then the bug turned out to be related to the underlying cutoff tooltip issue, and I was able to "easily" add in a fix for this.

PR #2320, opened 2017-10-02
Compare Chart Tooltip Fixes

arottersman:
## The Case of the Double Bug 

It's been a long day solving this mystery. We needed to make sure the charts' `clipPath`s were encompassing the tooltips, but all the charts with spaces in their names had messed up `clipPaths` so the cutoff-tooltip bug only showed up on Evapotranspiration, etc. 

I've removed the spaces from the chart names, and adjusted the `clipPath` so...
- tooltips don't render outside the chart area when you have many scenarios and slide to the right
- tooltips don't get cutoff at the top on maxed out bars

I also added a guard to prevent rendering tooltips on stacked charts with a value of 0.

Connects #2278 

### Demo

Not cutoff and not floating on the left
screen shot 2017-10-02 at 5 58 19 pm

Not showing for 0.0 valued bar
screen shot 2017-10-02 at 3 58 32 pm

## Testing Instructions

 * Pull, `bundle`
 * Site-Storm model with scenarios that will let you test
      - the slide control (there should be no tooltips outside the chart area)
      - tooltips that go above the chart scale
      - 0 valued tooltips -- they should appear if the chart isn't stacked, and should _not_ appear if it is

### DIFF
--- a/src/mmw/js/src/compare/views.js
+++ b/src/mmw/js/src/compare/views.js
@@ -353,6 +353,7 @@ var ChartRowView = Marionette.ItemView.extend({
             chartDiv = this.model.get('chartDiv'),
             chartEl = document.getElementById(chartDiv),
             name = this.model.get('name'),
+            chartName = name.replace(/\s/g, ''),
             label = this.model.get('unitLabel') +
                     ' (' + this.model.get('unit') + ')',
             colors = this.model.get('seriesColors'),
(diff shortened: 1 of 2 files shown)


kellyi:
0.001 values:

screen shot 2017-10-02 at 6 33 51 pm

->

screen shot 2017-10-02 at 6 34 08 pm

I think we should either use the same number of trailing decimal places for the tooltips as we do in the table view (and keep an adjusted version of the float guard check) OR just let the tooltips show regardless of the value & drop the guard altogether.

cc: @ajrobbins 

Nice job figuring this out!

ajrobbins:
>I think we should either use the same number of trailing decimal places for the tooltips as we do in the table view (and keep an adjusted version of the float guard check)

Let's do this. Looking great @arottersman !!

arottersman:
Added a commit to round to three decimal places like the table does. Okay to merge, @kellyi?
screen shot 2017-10-03 at 7 03 59 pm

kellyi:
Yep! Tested this out again in FF, Chrome, Safari, and IE11 and it works well!

screen shot 2017-10-04 at 9 45 02 am

arottersman:
Thanks for the thorough review, @kellyi!

Issue #2166, opened 2017-08-18
Compare View Revamp: Ensure ITSI Screenshots Still Work


rajadain:
Ensure values shown on screen are faithfully shown in ITSI screenshots, which has been tricky in the past.

kellyi:
Compare view isn't quite working in Chrome:

![screen shot 2017-09-26 at 2 58 13 pm](https://user-images.githubusercontent.com/4165523/30878625-33f67978-a2cb-11e7-887e-5b57c7a0d9ae.png)

kellyi:
Identical problems in Firefox:

![image](https://user-images.githubusercontent.com/4165523/30878908-e2f1bcb2-a2cb-11e7-9ed8-e70fdb9a8fc0.png)

kellyi:
Quick notes on how we got this set up to check locally:

- navigated to the ITSI portal here: https://authoring.concord.org/activities/5644/single_page/dec56c68-6170-4e59-b94a-82fb040e8d88
- used Requestly browser plugin to adjust the HTTP requests to replace app.wikiwatershed.org instance with an ngroked version I'd served from my local

kellyi:
On the off chance that the compare screen was hidden behind the map & sidebar, I manually deleted all those elements from the DOM on my local then took a screenshot -- but then nothing showed up:

![screen shot 2017-09-27 at 11 49 55 am](https://user-images.githubusercontent.com/4165523/30923582-3e2d0be6-a37a-11e7-982a-4afe5c33b7d4.png)

Loading the .html version of that saved screenshot does show the compare view, so I'm assuming that some/all of these elements aren't getting rendered in time for the headless browser to capture them:

![screen shot 2017-09-27 at 11 52 22 am](https://user-images.githubusercontent.com/4165523/30923646-6fb71f94-a37a-11e7-9ffc-eb9e1a5778c1.png)

I'm going to see whether there's a way to force the screenshot to wait til everything's been rendered before making the image. I think I may also add some setup/teardown stuff to make it so the compare view takes up the entirety of the viewport in a screenshot -- as it does on iPads & other narrow screens -- then reset things once the screenshot's done.

kellyi:
The missing compare view is related to something in the `.modal` class on `#compare-new`: with that class, the compare view doesn't appear at all in ITSI screenshots:

[...]

PR #2326, opened 2017-10-04
Enable ITSI screenshots for compare view

kellyi:
## Overview

This PR adds some logic to fix the compare view so that the maps, table, and stacked chart will render during an ITSI screenshot.

To accomplish this, I created some ITSI-specific css rules to apply to the compare view rows in the interval when the screenshot's taking place. PhantomJS doesn't fully support flexbox, so some of the new rules here apply supplements.

Others adjust the compare view to position it and also to hide the controls in the top right corner -- the fonts weren't pulling through for those on my local, although this may be superfluous in production.

Connects #2166 

### Demo

itsi-compare

### Notes

If we want them to work with ITSI screenshots, all the other views using flexbox need to have similar adjustments made OR we need to just add the supplementary rules to the base CSS. We should address this elsewhere.

It's a little unclear how we should handle the scrolling behavior for the compare view screenshot: since PhantomJS renders a new page, it starts scrolled up to the top regardless of what's in the user's viewport. I thought about hiding/showing rows programmatically depending on what was in the viewport, but letting it stick at the top seemed like a decent temporary solution since the stacked chart shows all of the info together.

## Testing Instructions
- `ngrok http 8000`
- use requestly and add a rule to replace `app.wikiwatershed.org` with your ngrok url
- visit the ITSI portal, create a TR-55 activity, then open the compare view
- take screenshots for the tables and charts for runoff and water quality and verify that you see the compare view in them

### DIFF
--- a/src/mmw/js/src/app.js
+++ b/src/mmw/js/src/app.js
@@ -155,6 +155,57 @@ function RestAPI() {
     };
 }
 
+function adjustCompareViewBeforeITSIScreenshot() {
+    toggleCompareViewForITSIScreenshot(true);
+}
+
+function adjustCompareViewAfterITSIScreenshot() {
+    toggleCompareViewForITSIScreenshot(false);
+}
+
+function toggleCompareViewForITSIScreenshot(adjustForScreenshot) {
+    var itsiCompareDialog = 'itsi-compare-dialog',
+        itsiCompareModal = 'itsi-compare-modal',
+        itsiCompareRow = 'itsi-compare-row',
+        compareDialog = '#compare-new-dialog',
+        compareModalContent = '.compare-modal-content',
+        compareCloseButton = '.compare-close',
+        compareChartButton = '#compare-input-button-chart',
+        compareTableButton = '#compare-input-button-table',
+        compareChartRow = '.compare-chart-row',
+        compareTableRow = '.compare-table-row',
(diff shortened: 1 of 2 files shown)


src/mmw/js/src/app.js
```diff
@@ -155,6 +155,57 @@ function RestAPI() {
...
+        compareCloseButton = '.compare-close',
+        compareChartButton = '#compare-input-button-chart',
+        compareTableButton = '#compare-input-button-table',
+        compareChartRow = '.compare-chart-row',
+        compareTableRow = '.compare-table-row',
+        compareScenariosRow = '.compare-scenarios',
+        compareMapsRow = '.compare-scenario-row-content';
+
+    if (adjustForScreenshot) {
+        $(compareDialog).addClass(itsiCompareDialog);
```
rajadain: Consider doing the `$(...)` in the `var` declaration above rather than down here, to avoid repeated DOM scans for the same things.

kellyi:
Thanks for all your help with this! Going to merge.

src/mmw/js/src/app.js
kellyi: Good point. I tried this out a bit & wasn't fully able to get it to work, but I think we should try to make that adjustment when/if we update this to account for the scroll behavior.

PR #2327, opened 2017-10-05
Fix Compare View Modal For IE11

arottersman:
## Overview

On IE11 compare view was showing up stuck to the upper left corner of the viewport.

I initially tried using the [PostCSS autoprefixer](https://github.com/postcss/autoprefixer). After finally getting it to run with the project's version of node, it didn't seem to make a difference to the compare view. Inspecting the DOM, I realized we were using the bootstrap modals in a pretty funky way: omitting the `modal-dialog` and `modal-content` classes. Adding these in fixed the issue. 

Connects #2098 
### Demo

screen shot 2017-10-05 at 8 36 39 am

### Notes

The original issue also describes some chart problems — #2320 seems to have fixed those.

## Testing Instructions

 * Pull and `bundle`
* Open compare view, and also open compare view on staging
* Confirm compare view matches chrome staging on all supported browsers.

### DIFF
--- a/src/mmw/js/src/compare/templates/compareWindow2.html
+++ b/src/mmw/js/src/compare/templates/compareWindow2.html
@@ -1,22 +1,24 @@
-<div class="compare-scenario-gradient"></div>
-<div class="compare-header">
-    <div class="compare-title">
-        <h1>Compare</h1>
-    </div>
-    <div class="spinner hidden"></div>
-    <div class="compare-close">
-        <button>
-            <i class="fa fa-times"></i>
-        </button>
-    </div>
-    <div class="compare-actions">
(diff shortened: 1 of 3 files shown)


kellyi:
Out of curiosity, I checked out the ITSI screenshots from this branch and they changes here enable them to work, roughly, too:

screen shot 2017-10-04 at 6 37 28 pm

I'm going to adjust #2326 to branch off this.

arottersman:
Thanks for catching that! 
- Pushed a fix that delays the map set up until after the modal showing has completed
- Amended the original commit to describe the styling changes (they're to override the effects of the new bootstrap wrappers).

PR #2328, opened 2017-10-05
BiG-CZ: CUAHSI Variable Values using Ulmo and WaterML

rajadain:
## Overview

Fetches values from CUAHSI / WDC using Ulmo for each variable at a site. We're currently only showing the latest value in the table, with charts coming in a follow-up PR.

Connects #2243 
Connects #2238 

### Notes

This PR adds two endpoints: `/details` and `/values`. Currently we are only using `/values` in the UI, but may use `/details` in the future. There is a dummy placeholder for the charts, but no values.

### Known Issues

The popover for the headline for WDC Details View doesn't work.

## Testing Instructions

 * Check out this branch and reprovision `app` and `worker`. Run the `bundle` script.
 * Go to [:8000/?bigcz](http://localhost:8000/?bigcz), pick a shape and search for something
 * Go to the WDC tab. Click on an item to go to the details view.
 * Ensure you see a spinner fetching the values.
 * As values succeed, ensure you see them being filled in the table one by one.
 * Ensure if fetching succeeds for all values, the "Last Updated" value is updated to be more recent. This happens because the API call used for the initial load has older values than the new ones that come via this second API.
 * Ensure that if any fetch fails, there is an error icon where the spinner goes.

### DIFF
--- a/src/mmw/apps/bigcz/views.py
+++ b/src/mmw/apps/bigcz/views.py
@@ -118,6 +118,40 @@ def _get_details(request):
         raise ParseError(ex.message)
 
 
+def _get_values(request):
+    params = request.query_params
+    catalog = params.get('catalog')
+
+    if not catalog:
+        raise ValidationError({
+            'error': 'Required argument: catalog'})
+
+    if catalog not in CATALOGS:
+        raise ValidationError({
+            'error': 'Catalog must be one of: {}'
+                     .format(', '.join(CATALOGS.keys()))})
+
+    values = CATALOGS[catalog]['values']
+
+    if not values:
+        raise NotFound({
+            'error': 'No values endpoint for {}'
+                     .format(catalog)})
+
+    values_kwargs = {
+        'wsdl': params.get('wsdl'),
+        'site': params.get('site'),
+        'variable': params.get('variable'),
+        'from_date': params.get('from_date'),
+        'to_date': params.get('to_date'),
+    }
+
+    try:
+        return values(**values_kwargs)
+    except ValueError as ex:
+        raise ParseError(ex.message)
+
+
 @decorators.api_view(['POST'])
 @decorators.permission_classes((AllowAny,))
 def search(request):
@@ -128,3 +162,9 @@ def search(request):
 @decorators.permission_classes((AllowAny,))
 def details(request):
     return Response(_get_details(request))
+
+
+@decorators.api_view(['GET'])
+@decorators.permission_classes((AllowAny,))
+def values(request):
+    return Response(_get_values(request))
--- a/src/mmw/apps/bigcz/views.py
+++ b/src/mmw/apps/bigcz/views.py
@@ -8,7 +8,7 @@ import json
 from django.contrib.gis.geos import GEOSGeometry
 from django.conf import settings
 from rest_framework import decorators
-from rest_framework.exceptions import ValidationError, ParseError
+from rest_framework.exceptions import ValidationError, ParseError, NotFound
 from rest_framework.permissions import AllowAny
 from rest_framework.response import Response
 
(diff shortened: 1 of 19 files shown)


src/mmw/apps/bigcz/views.py
```diff
@@ -87,7 +87,44 @@ def _do_search(request):
         raise ParseError(ex.message)
 
 
+def _get_details(request):
+    params = request.query_params
+    catalog = params.get('catalog')
+
+    if not catalog:
```
arottersman: I wonder if there should be some sort of request serializer that would handle all the validation here and [in here](https://github.com/WikiWatershed/model-my-watershed/pull/2328/commits/3cbed5c4798a1ee2916b20e32e9585ee1e504a8f#diff-cb0b2d859416dc6a0e55365632b75e0aR14). Validation errors off a serializer [would also be a list](http://www.django-rest-framework.org/api-guide/serializers/#validation), instead of the way it's done here, returning only one missing field error at a time.

This isn't that necessary at the moment, but could be something we add when we want to expose the API.

src/mmw/js/src/data_catalog/models.js
```diff
@@ -390,6 +390,93 @@ var SearchForm = Backbone.Model.extend({
...
+    },
+
+    parse: function(response) {
+        var mrv = null;
+
+        if (response.values) {
+            var values = this.get('values');
+
+            values.reset(response.values);
+            mrv = response.values[response.values.length - 1].value
```
arottersman: Is `response.values` guaranteed to have at least 1 value?

src/mmw/js/src/data_catalog/models.js
```diff
@@ -390,6 +390,93 @@ var SearchForm = Backbone.Model.extend({
...
+
+            delete response.values;
```
arottersman: Why do we delete it?

src/mmw/js/src/data_catalog/views.js
```diff
@@ -467,12 +469,68 @@ var ResultDetailsCuahsiView = ResultDetailsBaseView.extend({
...
+            location = id.substring(id.indexOf(':') + 1),
+            fetching = this.model.get('fetching'),
+            last_date = this.model.get('end_date');
+
+        if (!fetching) {
+            var variables = this.model.get('variables'),
+                last_dates = variables.map(function(v) {
+                        var values = v.get('values');
+
+                        return new Date(values.at(values.length - 1).get('datetime'));
```
arottersman: Consider using `values.last()` if possible.

src/mmw/js/src/data_catalog/views.js
```diff
@@ -527,9 +540,29 @@ var ResultDetailsCuahsiView = ResultDetailsBaseView.extend({
...
-            collection: this.model.get('variables')
-        }));
+        this.showValuesRegion();
+    },
+
+    showValuesRegion: function() {
+        var mode = this.model.get('mode'),
+            variables = this.model.get('variables'),
+            view = mode === 'table' ?
+                   new CuahsiTableView({ collection: variables }) :
```
arottersman: For some of our other toggles, we render all views [`onShow` (or wherever)](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/modeling/views.js#L959) and then do the switching via `display: none;` styling on the non-active view (bootstrap might take care of this?). 

I find the way you implemented it here a lot easier to follow, but I think the other way reduces the amount of rendering needed.

src/mmw/js/src/data_catalog/views.js
```diff
@@ -454,23 +450,164 @@ var ResultDetailsView = Marionette.ItemView.extend({
...
+        this.model.get('variables')
+                  .search({
+                      onEachPromise: showValuesRegion,
+                  })
+                  .done(fetchComplete)
+                  .fail(fetchError);
+    },
+
+    fetchComplete: function() {
+        this.model.set({ fetching: false, error: false });
```
arottersman: Consider moving these (`fetchComplete`, `fetchError`) to the model.

src/mmw/js/src/data_catalog/models.js
arottersman: Also, this line is missing a semi-colon. When the build works again, you may see some bad lint output.

rajadain:
Thanks for taking a look. I'm going to take this feedback into account, and add some more features, and will let you know when this is ready for another look.

rajadain:
I'm going to split this work into back-end and front-end pieces, and open two separate PRs for each. Closing this in the meantime. Thanks for your feedback!

rajadain:
Superseded by #2353 and #2354.

Issue #2276, opened 2017-09-20
IE11: Analyze Tab In Model Shows Model Results


arottersman:
The analyze/model toggle on the model page doesn't actually toggle the view in IE11. No console errors that I saw.
<img width="962" alt="screen shot 2017-09-20 at 4 25 55 pm" src="https://user-images.githubusercontent.com/7633670/30666063-2bc2e772-9e21-11e7-8a03-5ac302db570a.png">

kellyi:
This is happening because toggling the "Analyze" and "Model" tabs in IE11 doesn't properly change the `.active` class on the `#modeling-tab-contents` and `#analyze-tab-contents` divs. I manually changed them in the developer tools and got the desired output:

![screen shot 2017-10-09 at 10 25 27 am](https://user-images.githubusercontent.com/4165523/31343028-7c3b2614-acdc-11e7-88be-62b437869eea.png)

![screen shot 2017-10-09 at 10 25 47 am](https://user-images.githubusercontent.com/4165523/31343033-7f9377bc-acdc-11e7-8746-0994bf79fb5d.png)

Since the modeling tab is the default for IE11, it starts with the `.active` class & then that's never relinquished. Going to check the toggle code to see what adjustment to make.

PR #2352, opened 2017-10-09
Fix IE11 analyze/model tab toggle bug

kellyi:
## Overview

This PR fixes a bug whereby the analyze & modeling tab toggle wouldn't adjust the results view in IE11. This was happening because the `.active` class wasn't being toggled correctly in that browser.

To fix it I added some lines to call `.toggleClass('active')` on them explicitly in the `toggleAoiRegion` method of the modeling `ResultsView`.

Connects #2276 

### Demo

IE11:

ie11-analyze-modeling-tabs

Chrome:

chrome-analyze-modeling-tab

## Testing Instructions
- in Chrome, IE11, FF, and Safari check that you can
    - analyze a new AOI and see the correct tab
    - create new modeling projects for both TR55 and MapShed and then use the toggle to toggle the tabs correctly
    - load saved modeling projects for both TR55 and MapShed and toggle the tabs correctly

### DIFF
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -990,6 +990,9 @@ var ResultsView = Marionette.LayoutView.extend({
         } else {
             this.aoiRegion.empty();
         }
+
+        this.analyzeRegion.$el.toggleClass('active');
+        this.modelingRegion.$el.toggleClass('active');
     },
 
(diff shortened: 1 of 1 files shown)


rajadain:
Taking a look at this now.

kellyi:
Thanks!

PR #2353, opened 2017-10-09
BiG-CZ: Fetch CUAHSI Values using WaterML and Ulmo

rajadain:
## Overview

Adds [Ulmo](https://github.com/ulmo-dev/ulmo/), and uses it in Django endpoints for fetching CUAHSI details and values.

The expected usage is:

 1. Query `/bigcz/details` to fetch general information about all variables in a site
 2. Query `/bigcz/values` for each variable to fetch its values

UI for this is coming in a following PR. The UI may be broken until that next PR is merged.

See commit messages for details.

Connects #2243 
Connects #2238 

### Demo

* **Details**

    ```http
    http --print HhBb :8000/bigcz/details catalog==cuahsi wsdl=="http://hydroportal.cuahsi.org/nwisgw/cuahsi_1_1.asmx" site=="NWISGW:395801075101601"

    GET /bigcz/details?catalog=cuahsi&wsdl=http%3A%2F%2Fhydroportal.cuahsi.org%2Fnwisgw%2Fcuahsi_1_1.asmx&site=NWISGW%3A395801075101601 HTTP/1.1
    Accept: */*
    Accept-Encoding: gzip, deflate
    Connection: keep-alive
    Host: localhost:8000
    User-Agent: HTTPie/0.9.9



    HTTP/1.1 200 OK
    Allow: OPTIONS, GET
    Connection: keep-alive
    Content-Encoding: gzip
    Content-Type: application/json
    Date: Mon, 09 Oct 2017 17:29:25 GMT
    Server: nginx
    Transfer-Encoding: chunked
    Vary: Accept-Encoding
    Vary: Accept, Cookie

    {
        "code": "395801075101601",
        "elevation_m": "90",
        "location": {
            "latitude": "39.96705715",
            "longitude": "-75.17073369",
            "srs": "EPSG:4269"
        },
        "name": "PH   486",
        "network": "NWISGW",
        "series": {
            "NWISGW:72019": {
                "variable": {
                    "code": "72019",
                    "data_type": "unknown",
                    "general_category": "Hydrology",
                    "id": "3017",
                    "name": "Depth to water level, feet below land surface",
                    "no_data_value": "-999999",
                    "sample_medium": "Groundwater",
                    "speciation": "Not Applicable",
                    "time": {
                        "interval": "0",
                        "is_regular": true,
                        "units": {
                            "code": "103",
                            "name": "hour",
                            "type": "Time"
                        }
                    },
                    "units": {
                        "abbreviation": "ft"
                    },
[...]

PR #2354, opened 2017-10-09
BiG-CZ: Fetch and Render CUAHSI Variable Values

rajadain:
## Overview

Fetches CUAHSI variable values and populates the UI. The first few commits are cleanup, the later ones do the real fetching and rendering. See commit messages for details.

Also tagging @jfrankl for visual review.

This PR targets the backend branch to minimize the diff, in reality it will target and merge on to `develop`.

Builds on #2353 
Connects #2243 
Connects #2238 

## Testing Instructions

   - If you haven't reprovisioned for Ulmo in #2353, reprovision now: `vagrant reload app worker --provision`
 * Go to [:8000/?bigcz](http://localhost:8000/?bigcz), select a shape and search for something
 * Switch to the WDC tab. Ensure the search results look correct and no fields are missing
 * Click on any result. Ensure you see a spinner in the top right of the search results
 * Ensure that as the values come in, they are populated in the UI
 * Ensure that if there are any failures, there is an error icon where the spinner was
 * Ensure that if there was a failure, re-opening the Details view re-fetches the results
 * Ensure that if the results were successfully fetched, closing the details view and reopening it does not fetch them again
 * Ensure that if the details view is closed while results are fetching, the completion does not cause any JavaScript errors. Ensure that the details view can be re-opened during or after the fetch, and it behaves correctly in all cases
 * Ensure that popovers behave as they should in all cases
 * Ensure you can switch between Chart and Table views in the Details view (there is currently nothing in the Chart view)

### DIFF
--- a/src/mmw/js/src/data_catalog/models.js
+++ b/src/mmw/js/src/data_catalog/models.js
@@ -11,6 +11,7 @@ var DESCRIPTION_MAX_LENGTH = 100;
 var PAGE_SIZE = settings.get('data_catalog_page_size');
 
 var DATE_FORMAT = 'MM/DD/YYYY';
+var WATERML_VARIABLE_TIME_INTERVAL = '{http://www.cuahsi.org/water_ml/1.1/}variable_time_interval';
 
 
 var FilterModel = Backbone.Model.extend({
@@ -315,7 +316,105 @@ var Result = Backbone.Model.extend({
         created_at: '',
         updated_at: '',
         active: false,
-        show_detail: false // Show this result as the detail view?
+        show_detail: false, // Show this result as the detail view?
+        variables: null,  // CuahsiVariables Collection
+        fetching: false,
+        error: false,
+    },
+
+    initialize: function(attrs) {
+        // For CUAHSI
+        if (attrs.variables) {
+            this.set('variables', new CuahsiVariables(attrs.variables));
+        }
+    },
+
+    parse: function(response) {
+        // For CUAHSI
+        if (response.variables) {
+            var variables = this.get('variables');
+            if (variables instanceof CuahsiVariables) {
+                variables.reset(response.variables);
+                delete response.variables;
+            }
+        }
+
+        return response;
+    },
+
+    fetchCuahsiValues: function(opts) {
+        if (this.fetchPromise && !this.get('error')) {
+            return this.fetchPromise;
+        }
+
+        opts = _.defaults(opts || {}, {
+            onEachSearchDone: _.noop,
+            onEachSearchFail: _.noop,
+            from_date: null,
+            to_date: null,
+        });
+
+        var self = this,
+            variables = self.get('variables'),
+            runSearches = function() {
+                    return variables.map(function(v) {
+                            return v.search(opts.from_date, opts.to_date)
+                                    .done(opts.onEachSearchDone)
+                                    .fail(opts.onEachSearchFail);
+                        });
+                },
+            setSuccess = function() {
+                    self.set('error', false);
+                },
+            setError = function() {
+                    self.set('error', true);
+                },
+            startFetch = function() {
+                    self.set('fetching', true);
+                },
+            endFetch = function() {
+                    self.set('fetching', false);
+                };
+
+        startFetch();
+        this.fetchPromise = $.get('/bigcz/details', {
+                    catalog: 'cuahsi',
+                    wsdl: variables.first().get('wsdl'),
+                    site: self.get('id'),
+                })
+                .then(function(response) {
+                    variables.forEach(function(v) {
+                        var info = response.series[v.get('id')] || null,
+                            interval = info && info[WATERML_VARIABLE_TIME_INTERVAL];
+
+                        if (info) {
+                            v.set({
+                                'units': info.variable.units.abbreviation,
+                                'speciation': info.variable.speciation,
+                                'sample_medium': info.variable.sample_medium,
+                            });
+
+                            if (interval) {
+                                v.set({
+                                    'begin_date': new Date(interval.begin_date_time),
+                                    'end_date': new Date(interval.end_date_time),
+                                });
+                            }
+                        }
+                    });
+                }, function() {
+                    // Handle error in /details/
+                    setError();
+                    endFetch();
+                })
+                .then(function() {
+                    return $.when.apply($, runSearches())
+                                 .done(setSuccess)
+                                 .fail(setError)  // Handle error in /values/
+                                 .always(endFetch);
+                });
+
+        return this.fetchPromise;
     },
 
     getSummary: function() {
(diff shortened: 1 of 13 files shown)


arottersman:
Taking a look now!

src/mmw/js/src/data_catalog/views.js
```diff
@@ -469,23 +464,156 @@ var ResultDetailsView = Marionette.ItemView.extend({
...
+
+        if (!fetching && !error) {
+            var variables = this.model.get('variables'),
+                last_dates = variables.map(function(v) {
+                        var values = v.get('values');
+
+                        if (values.length > 0) {
+                            return new Date(values.last().get('datetime'));
+                        } else {
+                            return new Date('01/01/1900');
```
arottersman: What's the significance of this date? Sometimes I see "last updated 111 years ago", which leads me to believe there might be some that are older than 1900.

src/mmw/js/src/data_catalog/models.js
```diff
@@ -396,6 +474,115 @@ var PopoverControllerModel = Backbone.Model.extend({
...
+        params.to_date = params.to_date.format(DATE_FORMAT);
+
+        this.set('error', null);
+
+        return this.fetch({
+                data: params,
+                processData: true,
+            })
+            .fail(function(_, status) {
+                self.set('error', 'Error ' + status + ' during fetch');
```
arottersman: This ends up displaying a little funny:
screen shot 2017-10-09 at 4 41 37 pm

Maybe a colon would help?

src/mmw/js/src/data_catalog/models.js
```diff
@@ -397,6 +397,95 @@ var PopoverControllerModel = Backbone.Model.extend({
...
+    parse: function(response) {
+        var mrv = null;
+
+        if (response.values && response.values.length > 0) {
+            var values = this.get('values');
+
+            values.reset(response.values);
+            mrv = response.values[response.values.length - 1].value;
+
+            delete response.values;
```
rajadain: I suppose this deletion is no longer required, se we're not returning `response`

src/mmw/js/src/data_catalog/models.js
```diff
@@ -397,6 +397,95 @@ var PopoverControllerModel = Backbone.Model.extend({
...
+        } else {
+            this.set('error', 'No values returned from API');
+        }
+
+        return {
+            name: response.variable.name,
+            units: response.variable.units.abbreviation,
```
rajadain: We should also update `sample_medium` here

src/mmw/js/src/data_catalog/models.js
rajadain: Hmm I thought the `error` was an HTTP code. Will reformat.

src/mmw/js/src/data_catalog/models.js
```diff
@@ -315,7 +316,82 @@ var Result = Backbone.Model.extend({
...
+    },
+
+    initialize: function(attrs) {
+        // For CUAHSI
+        if (attrs.variables) {
+            this.set('variables', new CuahsiVariables(attrs.variables));
+        }
+    },
+
+    fetchCuahsiValues: function(opts) {
```
rajadain: I should add a comment describing the expected shape of `opts` here

src/mmw/js/src/data_catalog/views.js
rajadain: This "default" value was chosen to match the `DATE_MIN` set [here](https://github.com/WikiWatershed/model-my-watershed/blob/fe10d235207e902e7be5cebd8acc3a542e2a0a02/src/mmw/apps/bigcz/clients/cuahsi/search.py#L30), but is otherwise arbitrary. Maybe we should think of a better way to say "Never updated".

src/mmw/js/src/data_catalog/views.js
```diff
@@ -469,23 +464,156 @@ var ResultDetailsView = Marionette.ItemView.extend({
...
+    modelEvents: {
+        'change:fetching': 'render',
+        'change:mode': 'showValuesRegion',
+    },
+
+    initialize: function() {
+        var showValuesRegion = _.bind(this.showValuesRegion, this);
+
+        this.model.fetchCuahsiValues({
+            onEachSearchDone: showValuesRegion,
```
rajadain: Perhaps instead of reattaching a new instance of `CuahsiTableView` on each search done, we should simply `render` the currently attached one instead.

src/mmw/js/src/data_catalog/views.js
```diff
@@ -469,23 +464,156 @@ var ResultDetailsView = Marionette.ItemView.extend({
...
+        });
```
rajadain: And also add that `render` to an `.always` handler here.

rajadain:
Great bug catches. All of them should be addressed by the above fixups. See commit messages for details.

rajadain:

arottersman:
Taking another look!

rajadain:
Sorry about the confusion. Brought in some of @jfrankl's work from the `release/1.20.0` branch, and adjusted some wiring. Now this looks like:

image

I've kept the fixups separate for easier grokking, let me know if you prefer them squashed.

This is now ready for another look. Thanks!

rajadain:
Taking a look at the build failures now.

rajadain:
23c49f4 _should_ fix the build 🤞

arottersman:
Taking a look now!

arottersman:
Changes look great! Some lingering issues:
- setting the date filter to a recent date (like yesterday) causes an error
- the single error icon  for a variable when there are multiple errors thing is still there

rajadain:
For the inconsistent error icon issue — we are using Bootstrap Table for the table view for styling reasons, but I can imagine the client wanting the ability to sort. Because we use Bootstrap Table, all the event wiring is thrown off. The two options are:

 1. Continue using Bootstrap Table, and be prepared for sorting in the future. Switch from using a ColletionView + ItemView combo to just an ItemView, and translating all Backbone events into Bootstrap Table updates
 2. Remove Bootstrap Table, and add CSS to make it look consistent with other places

At first I was trying the former, but after it got a hairy I'm up for trying the latter. Will continue on this tomorrow.

rajadain:
Went with 1. This is ready for another look.

src/mmw/js/src/data_catalog/templates/resultDetailsCuahsiMain.html
```diff
@@ -0,0 +1,73 @@
+<div class="result-detail-header">
```
rajadain: Sorry, this is an extra file, will be deleted.

rajadain:
Thanks for sticking with me through this. I was a little surprised by how well the Bootstrap Table translation worked as well.

Going to squash the fixups, wait for CI to go green, then merge.

Issue #2322, opened 2017-10-03
Site storm results are identical for 100% forest and current conditions


ajrobbins:
Observed:
100% forest results = current conditions results for runoff and water quality:
![screen shot 2017-10-03 at 2 33 41 pm](https://user-images.githubusercontent.com/12868342/31142445-502d3a0a-a848-11e7-808a-13501cf2f302.png)
![screen shot 2017-10-03 at 2 34 40 pm](https://user-images.githubusercontent.com/12868342/31142446-512282a8-a848-11e7-90c8-52b49e2179e7.png)

Expected:
Results should be different, especially in urban areas. This was previously addressed, for water quality at least, in https://github.com/WikiWatershed/model-my-watershed/pull/1802.

arottersman:
We aren't parsing the results for the compare view correctly: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/compare/models.js#L67

We're always accessing the value in the `modified`key, but there are 3 values:
`modified`, `unmodified` and `pc_unmodified`. `pc_unmodified` corresponds to the value of the model run under precolumbian/100% forest conditions.

PR #2355, opened 2017-10-09
Compare View: Fix 100% Forest Cover Scenario

arottersman:
## Overview

Even in urban areas, the compare view Current Condition values were the same as the 100% Forest Cover scenario's. We were unconditionally using the `modified` results, but the 100% Forest Cover results exist under `pc_unmodified`. This PR has the charts use results from the key that corresponds with each scenario. This PR also adds a few tests to confirm the compare charts are using the proper values from the results.

Adding the tests involved updating the mock scenario in `modeling/mocks` which was missing the full result object (it didn't have `pc_unmodified`).

Connects #2322 

### Demo

screen shot 2017-10-09 at 2 43 44 pm

## Testing Instructions

 * Pull this branch and `bundle`
 * Add a scenario to the TR-55 model with a forest modification covering the extent of the AOI
 * Confirm it is similar to the 100% forest scenario results in the compare view for both Runoff and Water Quality

### DIFF
--- a/src/mmw/js/src/compare/models.js
+++ b/src/mmw/js/src/compare/models.js
@@ -43,6 +43,16 @@ var ChartRowsCollection = Backbone.Collection.extend({
         this.scenarios.forEach(function(scenario) {
             scenario.get('results').on('change', update);
         });
+    },
+
+    getScenarioResults: function(typeKey) {
+        return this.scenarios.map(function(scenario) {
+            var resultKey = coreUtils.getTR55ResultKey(scenario),
+                result = scenario.get('results')
+                                 .findWhere({ name: typeKey })
+                                 .get('result');
+            return result[typeKey][resultKey];
+        });
     }
 });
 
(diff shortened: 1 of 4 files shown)


src/mmw/js/src/modeling/tests.js
```diff
@@ -693,14 +693,17 @@ describe('Modeling', function() {
                     });
 
                     model.updateModificationHash();
-                    assert.equal(model.get('modification_hash'), '582b6178186440159bd4ea25f0260892');
+                    assert.equal(model.get('modification_hash'),
+                                 'ca1872c7647c7cb9a7f0a00b683407df' );
```
rajadain: I'm having to revert these three changes to get the tests to pass

src/mmw/js/src/compare/models.js
```diff
@@ -43,6 +43,16 @@ var ChartRowsCollection = Backbone.Collection.extend({
         this.scenarios.forEach(function(scenario) {
             scenario.get('results').on('change', update);
         });
+    },
+
+    getScenarioResults: function(typeKey) {
+        return this.scenarios.map(function(scenario) {
+            var resultKey = coreUtils.getTR55ResultKey(scenario),
+                result = scenario.get('results')
+                .findWhere({ name: typeKey })
```
rajadain: Consider indenting these at least one level further in, so they don't align with the left-hand side declaration of variables.

src/mmw/js/src/modeling/tests.js
arottersman: I had to adjust these to get it to pass on my local, but Jenkins wanted them changed. I've pushed a fixup to revert.

src/mmw/js/src/compare/tests.js
```diff
@@ -0,0 +1,235 @@
...
+                           return !scenario.get('is_current_conditions') &&
+                               !scenario.get('is_pre_columbian');
+                       },
+
+                           modifiedResult = this.scenarios
+                           .find(isUserCreated)
+                           .get('results')
+                           .findWhere({ name: 'runoff' })
+                           .get('result'),
+                           modifiedValue = modifiedResult.runoff['modified'];
```
kellyi: Consider indenting these (and other) `.get` &c calls to distinguish them visually from the variables here.

arottersman: Someday I will correct my editor's auto-indent...will definitely make this look right before merging.

kellyi:
Looks good to me! ->

screen shot 2017-10-09 at 4 00 07 pm

arottersman:
Forced push some changes that fix the indentation. Will merge once the build completes. Thanks for reviewing, guys!

Issue #2291, opened 2017-09-25
BiG-CZ: Add Hydroshare Geometry


rajadain:
As of https://github.com/hydroshare/hydroshare/pull/2265, the Hydroshare API output includes a `coverage` field:

```json
{
  "bag_url": "http://www.hydroshare.org/django_irods/download/bags/197adcdc76b34591bd78a811bf1dfbfe.zip",
  "coverages": [
    {
      "type": "period",
      "value": {
        "end": "2015-10-01",
        "start": "2001-10-01"
      }
    },
    {
      "type": "box",
      "value": {
        "eastlimit": -97.998,
        "northlimit": 50.1769,
        "projection": "WGS 84 EPSG:4326",
        "southlimit": 27.6835,
        "units": "Decimal degrees",
        "westlimit": -126.8262
      }
    }
  ],
  "creator": "John Hammond",
  "date_created": "08-23-2017",
  "date_last_updated": "08-23-2017",
  "discoverable": true,
  "immutable": false,
  "public": true,
  "published": false,
  "resource_id": "197adcdc76b34591bd78a811bf1dfbfe",
  "resource_map_url": "http://www.hydroshare.org/hsapi/resource/197adcdc76b34591bd78a811bf1dfbfe/map/",
  "resource_title": "MODIS MOD10A2 derived snow season for the western U.S.",
  "resource_type": "GenericResource",
  "resource_url": "http://www.hydroshare.org/resource/197adcdc76b34591bd78a811bf1dfbfe/",
  "science_metadata_url": "http://www.hydroshare.org/hsapi/resource/197adcdc76b34591bd78a811bf1dfbfe/scimeta/",
  "shareable": true
}
```

This coverage information is shown on the [Hydroshare website](https://www.hydroshare.org/resource/197adcdc76b34591bd78a811bf1dfbfe/) as:

![image](https://user-images.githubusercontent.com/1430060/30832113-ee16a51c-a217-11e7-90b8-0d0adbe61d5d.png)

Make it so that if a result has a valid `box` coverage, then it is added to the result's `geom` field, which will be picked up and added to the map.

PR #2356, opened 2017-10-09
BiG-CZ: Include geometry for HydroShare results

rajadain:
## Overview

Previously the HydroShare API did not send any geometries in its responses. Now, the "coverages" key is an array of (at least) two types of values: period and box, for time and geography respectively.

We translate the box coverage into a geometry if it is included. It behaves identically to CINERGI in the UI, with respect to multiple overlapping popups and polygons that exceed the map bounds.

Connects #2291 

## Testing Instructions

 * Search for something in HydroShare
 * Ensure results have geometries


rajadain:
Thanks for taking a look!

Issue #2284, opened 2017-09-21
Reduce number of significant figures shown in climate table by 1


ajrobbins:

rajadain:
Shares a point with #2291

PR #2358, opened 2017-10-10
BiG-CZ: Reduce Significant Digits for Precipitation

rajadain:
## Overview

Since it is unlikely that the underlying data is accurate down to 1/10th a millimeter, we reduce the significant digits in the output to 0.1 cm.

Connects #2284 

### Demo

image

## Testing Instructions

 * Pick a shape and Analyze. Switch to the Climate tab. Ensure the table has only one decimal place for all values.
 * Ensure the charts also have only one decimal place for both precipitation and temperature.


rajadain:
Thanks for taking a look!

Issue #2188, opened 2017-08-24
Geoprocessing API: Add Account Page For User To Manage API Tokens


arottersman:
API tokens will be generated for every MMW user. Provide an account page where they manage their token (ie. [view it](http://www.django-rest-framework.org/api-guide/authentication/#by-exposing-an-api-endpoint
), invalidate it)

arottersman:
@jfrankl has designs

jfrankl:
![image](https://user-images.githubusercontent.com/1809908/30435735-7f28e5a8-9938-11e7-850e-29ebfa9f9fd8.png)

![image](https://user-images.githubusercontent.com/1809908/30435747-88fa1f7a-9938-11e7-975b-63203c31611a.png)

![image](https://user-images.githubusercontent.com/1809908/30435756-9085cbea-9938-11e7-8ef3-7135251b932c.png)

ajrobbins:
Need another page or field for API key - talking to @jfrankl about this.

mmcfarland:
> Need another page or field for API key

I think it can go on the Account page with Change Password.  We'd want a text box to copy the code (maybe a copy button like "share", and a button to generate the key.   A warning should be displayed that regenerating the key will invalidate any previous key.

jfrankl:
@arottersman @ajrobbins @mmcfarland how does this look? The reset button would generate a new key and update the date. The user would see some warning after clicking the button to confirm they understand the consequences.

![image](https://user-images.githubusercontent.com/1809908/30438977-99c2f680-9940-11e7-850d-0a13d2260edf.png)

![image](https://user-images.githubusercontent.com/1809908/30438989-a26c660e-9940-11e7-9b3e-3ecb6f3a79c0.png)

arottersman:
- For the confirmation/warning modal, the text should be something like "This will invalidate any previous requests..."

- The throttling message should be on the main page it — refers to the limit on the number of requests the user can make per minute with their API key.

- The "reset key" might be better as "regenerate key" or "get a new key"

ajrobbins:
+1 to Alice's comments. I like "Regenerate API Key"

jfrankl:
How about this?

![image](https://user-images.githubusercontent.com/1809908/30440198-d089f1c0-9943-11e7-889b-3ead0347c04c.png)

ajrobbins:
LGTM

arottersman:
[...]

PR #2359, opened 2017-10-10
Geoprocessing API: /token management endpoint

arottersman:
## Overview
The account page will need to be able to view a users API token and be able to regenerate it. To support that, replace the `/api-auth-token` endpoint we were using stock from DRF with a custom `/api/token` one that...
   - allows sending credentials via session (and also via json object like the stock DRF view allowed)
   - exposes a `regenerate` flag to signal that the returned token should be a new one
   - returns the `created_at` time in the response

Connects #2188 

### Bonus
I was basing my tests off of `user/tests.py`, but found they weren't using a csrf_token in their requests like they claimed -- figuring this out and how to get mine to pass uncovered why the commented out test in `user/tests` fails. Added ad783d4 to address

## Demo

screen shot 2017-10-11 at 3 10 49 pm

```http
> http --json POST 'http://localhost:8000/api/token/' username="dog" password="dog"
HTTP/1.1 200 OK
Allow: POST, OPTIONS
Connection: keep-alive
Content-Encoding: gzip
Date: Wed, 11 Oct 2017 19:01:37 GMT
Server: nginx
Transfer-Encoding: chunked
Vary: Accept-Encoding
Vary: Accept

{
    "created_at": "2017-10-09T21:28:05.805Z",
    "token": "971922817efb5532bfe1ade983aab31a2dcf3fc9"
}

```

## Notes

I removed the request body from the swagger UI because it seemed wrong to have people sending their passwords in it...hopefully the example gives enough info for people to use the endpoint elsewhere.

## Testing
* `localhost:8000/api/docs`
* while logged into MMW, you should be able to make the `api/token` request with no body and get a useable token
* You should be able to get the same token via your credentials inserted here:
```
 > http --json POST 'http://localhost:8000/api/token/' username="dog" password="dog"
```

You should see the created_at and token update if you pass `regenerate=true`
```
> http --json POST 'http://localhost:8000/api/token/' username="dog" password="dog regenerate=true"
```

### DIFF
--- a/src/mmw/apps/geoprocessing_api/permissions.py
+++ b/src/mmw/apps/geoprocessing_api/permissions.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from rest_framework import authentication
+from rest_framework.authtoken.serializers import AuthTokenSerializer
+
+
+class AuthTokenSerializerAuthentication(authentication.BaseAuthentication):
+    def authenticate(self, request):
+
+        # if the user has made no attempt to add
+        # credentials via the request body,
+        # pass through so other authentication
+        # can be tried
+        username = request.data.get('username', None)
+        password = request.data.get('password', None)
+        if not username and not password:
+            return None
+
+        serializer = AuthTokenSerializer(data=request.data,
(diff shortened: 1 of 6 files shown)


arottersman:
Thanks!

Issue #2204, opened 2017-08-29
Develop legend for climate layers


mmcfarland:
Develop legends for Mean Temp and Mean Precip which incorporate the normalization and non-linear ramping done in the creation of the visualizations.  Previous layers have discrete colors, and this is more of a true ramp.  We could either pluck out discrete values and have the scale inferred, or try to regenerate a corrected version of the one supplied from Nanoos.

![image](https://user-images.githubusercontent.com/1014341/29837847-62d0c2a4-8cc7-11e7-8b40-b5e98f0edadb.png)

@jfrankl @ajrobbins

rajadain:
Taking the original color ramp from here: https://github.com/WikiWatershed/model-my-watershed/issues/2122#issuecomment-321374942

```python
colors = {
        "tmean": {
            "colors": [(46, 0, 103),
                       (141, 20, 255),
                       (165, 15, 245),
                       (189, 10, 235),
                       (213, 5, 225),
                       (238, 0, 215),
                       (178, 8, 225),
                       (119, 16, 235),
                       (59, 25, 245),
                       (0, 34, 255),
                       (0, 77, 199),
                       (0, 121, 143),
                       (0, 164, 86),
                       (0, 208, 31),
                       (59, 211, 23),
                       (117, 214, 15),
                       (176, 217, 7),
                       (236, 221, 0),
                       (240, 165, 0),
                       (245, 110, 2),
                       (250, 55, 2),
                       (255, 0, 4),
                       (87, 0, 16)],
            "position": [0., 0.16666667, 0.2, 0.23333333, 0.26666667,
                         0.3, 0.33333333, 0.36666667, 0.4, 0.43333333,
                         0.46666667, 0.5, 0.53333333, 0.56666667, 0.6,
                         0.63333333, 0.66666667, 0.7, 0.73333333, 0.76666667,
                         0.8, 0.83333333, 1.]
        },
        "ppt": {
            "colors": [(230, 111, 0),
                       (246, 155, 58),
                       (62, 178, 189),
                       (97, 212, 231),
                       (3, 116, 116),
                       (0, 49, 96)],
            "position": [0.0, 0.08, 0.16, 0.24, 0.32, 1.0]
        },
        "EVI": {
            "colors": [(229, 229, 229), (182, 165, 134), (160, 138, 91),
                       (138, 111, 49), (140, 127, 43), (142, 143, 37), (144, 159, 31),
[...]

PR #2368, opened 2017-10-12
Create temperature & precipitation legends

kellyi:
## Overview

This PR adds temperature and precipitation legend popovers to the layer selector. We decided to stick with continuous values rather than the discrete steps set in the other legends, and it's implemented here using css gradients. Beneath the gradients I set up tick marks and value labels to echo how the source image looked in #2204, and also added a units label for emphasis.

Connects #2204 

## Demo

climate-legends

## Notes

We decided to use horizontal popovers for these, and doing that made it possible for the user to arrange the popover and month slider such that the values were both visible simultaneously. In turn, I opted not to try to reposition the legends for these -- which means it's possible to have the legend appear above the slider, but also fairly easy to adjust to that it doesn't.

The source values lists for these in `layer_settings.py` include some empty string values -- this was done specifically to force additional tick marks but not to display unit values. This also potentially makes it fairly easy to adjust these to change which increments we display.

Lastly, I'm not totally certain that the tick marks and labels appear at exactly the correct place beneath the gradient legend: they seem to be at least mostly correct, but it seems like there's a slight difference between the legend in #2204 and the versions here.

## Testing

- in FF, Safari, IE11, and Chrome, visit the app and open the precipitation and temperature legends and verify that they appear correctly as pictured in the screenshot above

### DIFF
--- a/src/mmw/mmw/settings/layer_settings.py
+++ b/src/mmw/mmw/settings/layer_settings.py
@@ -130,7 +130,6 @@ LAYER_GROUPS = {
             'code': 'mean_ppt',
             'display': 'Mean Monthly Precipitation',
             'short_display': 'Mean Precip',
-            'css_class_prefix': 'ppt',
             'helptext': 'PRISM monthly mean precipitation.',
             'url': 'https://{s}.tiles.azavea.com/climate/ppt_{month}/{z}/{x}/{y}.png',  # noqa
             'maxNativeZoom': 10,
@@ -138,12 +137,15 @@ LAYER_GROUPS = {
             'opacity': 0.85,
             'has_opacity_slider': True,
             'time_slider_values': MONTH_CODES,
+            'use_color_ramp': True,
+            'color_ramp_id': 'precipitation-legend',
+            'legend_units_label': 'Precipitation (mm/month)',
+            'legend_unit_breaks': [0, '', 100, '', 200, '', 300, '', 400, '', 500]
         },
         {
             'code': 'mean_temp',
             'display': 'Mean Monthly Temperature',
             'short_display': 'Mean Temp',
-            'css_class_prefix': 'ppt',
             'helptext': 'PRISM monthly mean temperature.',
             'url': 'https://{s}.tiles.azavea.com/climate/tmean_{month}/{z}/{x}/{y}.png',  # noqa
             'maxNativeZoom': 10,
@@ -151,6 +153,10 @@ LAYER_GROUPS = {
             'opacity': 0.85,
             'has_opacity_slider': True,
             'time_slider_values': MONTH_CODES,
+            'use_color_ramp': True,
+            'color_ramp_id': 'temperature-legend',
+            'legend_units_label': u'Air Temperature (\xb0C)',
+            'legend_unit_breaks': [-20, '', -8, '', 4, '', 16, '', 28, '', 40]
         },
         {
             'code': 'urban_areas',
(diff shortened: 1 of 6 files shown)


kellyi:
I rebased this on `develop`.

src/mmw/mmw/settings/layer_settings.py
```diff
@@ -130,27 +130,34 @@
...
             'helptext': 'PRISM monthly mean temperature.',
             'url': 'https://{s}.tiles.azavea.com/climate/tmean_{month}/{z}/{x}/{y}.png',  # noqa
             'maxNativeZoom': 10,
             'maxZoom': 18,
             'opacity': 0.85,
             'has_opacity_slider': True,
             'time_slider_values': MONTH_CODES,
+            'use_color_ramp': True,
+            'color_ramp_id': 'temperature-legend',
+            'legend_units_label': u'Temperature (\xb0 Celsius)',
```
rajadain: Prefer `\xb0C` as the unit

rajadain: Or `&deg;C` if HTML escaping is allowed. Basically, there should be no space between `°` and `C`

kellyi:
> the table says mean value should be 8.6 cm/month (or 86 mm/month), but the color of the tile looks more like 40 mm/month according to the legend

Interesting. I think we can tweak how the legend looks by adjusting the percentage values for the gradient a bit. Evenly splitting them rather than compacting most of the splits from 0-32% for the precipitation legend might be enough to have the image above match correctly.

kellyi:
Out of curiosity, I switched the precipitation legend gradient to split at 20, 40, 60, and 80% instead of the original values and it does look like that's slightly more accurate:

screen shot 2017-10-18 at 9 38 27 am

screen shot 2017-10-18 at 9 43 00 am

src/mmw/mmw/settings/layer_settings.py
kellyi: I switched this to `\xb0C`.

mmcfarland:
I spot checked both the values reported by Climate Analyze against the source GeoTiffs and that our renderings matched the MonitorMyWatershed site.  Both are correct. I think there are a few things going on here:

The legend image I linked to in the issue is not correct.  Following the conversation from #2122 you can see the realization:
> I have found the problem. The tickmark values in the legend isn't correct.  https://github.com/WikiWatershed/model-my-watershed/issues/2122#issuecomment-324168504

and the update with the fix:
> I've updated the legend for both precip and tmean. Thanks for catching this error. Apologies for the hassle and the confusions https://github.com/WikiWatershed/model-my-watershed/issues/2122#issuecomment-324389382

I tracked the correct legends down from the MonMW site:
precip

temp

And they seem to match the renderings/values appropriately.  If we stick with the 8,16,24,32 spread but update the label ticks, we end up with a legend that matches very closely, but not exactly the new reference legends:

ppt

I think that the slight offset is due to the staggered spacing of the second and penultimate ticks with respect to the first and last ticks.  The placement and distribution of the colors do look correct.

I didn't evaluate the temperature legend, but it too may need to have its tick labels adjusted to reference the spread show above and have the tick spacing adjusted to be more uniform.

Alternatively, if we really struggle to match the source legend precisely we could...use the source legend.  The images above are links to the source png which we could incorporate into the app.  I think it better to have the capacity to generate on our own so we can adjust styling to our needs, but only at a reasonable cost.

kellyi:
Adjusting the tick values & resetting the percentage splits for the precipitation legend gets them to match more accurately:

screen shot 2017-10-19 at 10 06 12 am

precip

//

screen shot 2017-10-19 at 10 06 17 am

temp

Going to spend a small amount of time seeing if I can get the ticks & values to distribute evenly.

kellyi:
Made some more css adjustments and now it's:

screen shot 2017-10-19 at 10 54 51 am
precip

screen shot 2017-10-19 at 10 54 56 am
temp

kellyi:
Squashed this & rebased on `develop`.

PR #2369, opened 2017-10-13
Add API and Password Management Page 

arottersman:
## Overview

Fill in Account Page -> Account:

- Add section to copy and regenerate your API key
- Add button to reset password
    - Add user modal view to reset password
    - Add new django `change_password` view

See wireframes in original issue. 

See commit messages for more details. 

Connects #2188 
### Demo

et31bcfymy

Error and loading states for API key (throttling turned on to make it obvious)
nco1mlrrub

Error states, reset password....still pretty tragically styled 
screen shot 2017-10-13 at 11 21 59 am
screen shot 2017-10-13 at 11 21 45 am

## Testing Instructions

 * Pull and `bundle`
 * http://localhost:8000/account --> Account
 * Confirm you can get the api key sin your clipboard by clicking the "copy" button
* Confirm the "regenerate key" button updates the view with a new key (and displays a warning
modal)
* Confirm the reset password button opens a modal that allows resetting your password
* Confirm making your browser offline shows an error message when you try to regenerate your api key


rajadain:
Taking a look now

rajadain:
Don't think we've used a visual transition like this one on the "cancel" button before:



Can we keep it consistent with other modals in the app, replacing only the teal of the `ConfirmModal` with red?

rajadain:
We should probably add some more spacing between the two sections to make them more distinct

rajadain:
In case of error in refreshing the API key, this is the message shown:

image

Can it be made less personal? Like:

> Unable to regenerate API key

Or

> Encountered error {code} while regenerating API key

Also, I noticed that the "Copy" button changed to the "Error" icon and is now unclickable. Is this just to demonstrate the fact that the request failed, or does this reflect that the user now has no valid API key? Their old one has been discarded and a new one could not be fetched?

In this case, I made my browser offline to sample the error. In this case, the current API key is still valid, so the user should still be able to copy it.

Perhaps we should add the error icon elsewhere and not interfere with the copying mechanism.

src/mmw/js/src/user/views.js
```diff
@@ -385,10 +387,37 @@ var ItsiSignUpModalView = ModalBaseView.extend({
...
+    onValidationError: function() {
+        this.ui.oldPassword.focus();
+    },
+
+    setFields: function() {
+        this.model.set({
+            old_password: $(this.ui.oldPassword.selector).val(),
+            new_password1: $(this.ui.password1.selector).val(),
+            new_password2: $(this.ui.password2.selector).val(),
+        }, { silent: true });
```
rajadain: Why the `silent`? Is anything listening to this that shouldn't be tripped?

arottersman:
Thanks @rajadain, great suggestions.

re: error icon
There's the case where you click "regenerate key", and while the request is being made, your browser goes offline. The key shown in your browser will likely be stale now, as the server will have generated a new key. You are still able to try the existing key by copying it manually from the field, but the error icon conveys it might be stale/bad. Admittedly this is a pretty rare case with modern connection speeds. I'll probably move the icon to in front of the red error message.

re: style updates
@jfrankl, let me know if you'd like to play with it off my machine. Can send you a link any time. Also happy to grab a room if you want to pair.

src/mmw/js/src/user/views.js
arottersman: Great q. I followed the pattern the other user views were using, and just assumed the `silent` flag was to prevent `validate` firing on change. [Reading the docs now](http://backbonejs.org/#Model-validate) seems that's not the case. Going to remove this, see what happens, and get back to you.

src/mmw/js/src/account/views.js
```diff
@@ -37,6 +54,15 @@ var AccountContainerView = Marionette.LayoutView.extend({
         infoContainer: '.account-page-container'
     },
 
+    initialize: function() {
+        this.tokenModel = new models.ApiTokenModel();
+    },
+
+    onChangeActivePage: function() {
```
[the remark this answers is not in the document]
arottersman: It's not — must be a rebase error.

src/mmw/js/src/user/views.js
arottersman: Not sure why this is on the other views: https://github.com/WikiWatershed/model-my-watershed/commit/ec2dd6434434704051496a33513f34c41d115b5e#diff-bbc329caa6aad1d9e2bf745db301390fR43

Removed it from this new one and it worked fine. Suspect it's not needed on the others, but don't want to mess around right now. Will push a fixup that removes it from the new view, but leaves the others.

rajadain: There's probably a `silent` there because the view [is listening to the model's changes](https://github.com/WikiWatershed/model-my-watershed/commit/ec2dd6434434704051496a33513f34c41d115b5e#diff-bbc329caa6aad1d9e2bf745db301390fR30) and we don't want to re-render the view upon validation.

arottersman: All those views listen as well https://github.com/WikiWatershed/model-my-watershed/blob/5e2385a8f932ff270b006e59af57ae0d0d491610/src/mmw/js/src/user/views.js#L42. Going to keep the `silent` then.

caseycesari: 🤐

arottersman:
Ready for another look. In addition to suggested changes:

- Cancel button on regenerate key modal doesn't show any additional hover styling (no red background, no bold. Requested by @jfrankl)

- Added tooltip to copy button, also requested by Jeff
8racvy4ae5

Updated screenshots:

screen shot 2017-10-17 at 3 43 20 pm

screen shot 2017-10-17 at 4 37 45 pm

rajadain:
Taking another look

arottersman:
Thanks for reviewing! Squashed my fixups down. Will merge when the build goes green.

Issue #2367, opened 2017-10-12
BiG-CZ: Browser back button breaks filter sidebar


arottersman:
- Go to the data catalog, search for something and open the filter panel
- Hit the browser back button
- Go to the data catalog again
- Try to open the filter again, and it will be blank
<img width="960" alt="screen shot 2017-10-12 at 2 37 12 pm" src="https://user-images.githubusercontent.com/7633670/31513558-b4d2c96e-af5c-11e7-8d02-8b2264348826.png">

kellyi:
This element lives in #secondary-sidebar > #secondary-sidebar-content divs as the `.data-catalog-filter-window` div element.

Clicking the back button when this is open doesn't unmount the `.data-catalog-filter-window` component. On clicking "Search data sources" again, the element's still on the DOM.

At that point, clicking "Filters" again removes it from the DOM.

Issue #2366, opened 2017-10-12
BiG-CZ: Browser back button doesn't clear detail geom


arottersman:
If you use the browser back button while a detail view is open in the data catalog, the detail geom will remain on the map:

<img width="959" alt="screen shot 2017-10-12 at 2 31 05 pm" src="https://user-images.githubusercontent.com/7633670/31513520-97d3aed2-af5c-11e7-805d-2bc78deafa6a.png">

kellyi:
Trying to replicate this and it seems like BigCZ has the same behavior as MMW, viz:

Clicking back from the "Analyze" tab will leave the AOI on the map, but with the aoi rendered differently:

<img width="773" alt="screen shot 2017-10-16 at 3 41 25 pm" src="https://user-images.githubusercontent.com/4165523/31631714-8983ebe2-b288-11e7-9514-c06cbdfce57d.png">

->

![screen shot 2017-10-16 at 3 41 31 pm](https://user-images.githubusercontent.com/4165523/31631717-8d68ac2a-b288-11e7-90aa-49ce66986501.png)

The AOI will then clears on trying to select a new AOI.

Assuming I've replicated this correctly, is the desired behavior here that no shape is visible on the map immediately after clicking the back button (and before selecting a new AOI)?

mmcfarland:
The issue references "the detail geom", and there is a small circle in the screenshot.  I think it may be referring to a selected result's geometry, not the AoI shape.  @arottersman ?

kellyi:
> I think it may be referring to a selected result's geometry, not the AoI shape

Ah, that makes sense. I was able to replicate this locally by searching for results in the data catalog for an aoi, then clicking back. Looks like it remains on the map both for the Analyze tab and on returning to the "Select Area" view:

![screen shot 2017-10-16 at 4 01 41 pm](https://user-images.githubusercontent.com/4165523/31632556-5ff87dd0-b28b-11e7-9b05-960d87c087f7.png)

->

![screen shot 2017-10-16 at 4 02 55 pm](https://user-images.githubusercontent.com/4165523/31632593-7fac719a-b28b-11e7-9d56-ae6bef898523.png)

kellyi:
Out of curiosity I checked whether this was happening on the branch from #2371 and it seems like it may not be? Going to investigate a bit more but may have been the same root cause.

PR #2371, opened 2017-10-16
Destroy FilterView from Data Catalog on clicking back button

kellyi:
## Overview

This PR fixes a bug whereby the FilterView wouldn't be re-created if the user hit the back button from a Data Sources page on which it had been shown. I *think* using the `_.bind` function to wrap it for updates in the FormView's initializer may have caused it to lose the usual Marionette life cycle methods. (I wrote up a brief description of how this looks on the DOM in #2367).

Whatever the cause, we fix it by manually emptying the view's region before the FormView unmounts.

Connects #2367 

## Testing Instructions
- visit bigcz, search for an aoi, then open the data catalog
- open the filter view and hit the back button
- open the data catalog and verify that the filter view now opens correctly
- verify that the filter view otherwise works as expected

### DIFF
--- a/src/mmw/js/src/data_catalog/views.js
+++ b/src/mmw/js/src/data_catalog/views.js
@@ -198,6 +198,10 @@ var FormView = Marionette.ItemView.extend({
         'click @ui.filterToggle': 'onFilterToggle',
     },
 
+    onBeforeDestroy: function() {
+        App.rootView.secondarySidebarRegion.empty();
+    },
+
     initialize: function() {
         var updateFilterSidebar = _.bind(function() {
             if (App.rootView.secondarySidebarRegion.hasView()) {


kellyi:
Can you also check whether you're able to re-create the bug described in #2366 on this branch? I was able to replicate the bug on develop but not here which makes me wonder if this fix addressed both issues.

kellyi:
> Can you also check whether you're able to re-create the bug described in #2366 on this branch? I was able to replicate the bug on develop but not here which makes me wonder if this fix addressed both issues.

Nevermind this; I just was able to re-create the bug despite the wishful thinking!

arottersman:
Taking a look now!

kellyi:
Thanks!

PR #2372, opened 2017-10-16
Clear details geom on leaving DataCatalogWindow

kellyi:
## Overview

This PR clears the details geometry on unmounting the DataCatalogView component. To accomplish this we set `"dataCatalogDetailResult"` to `null` in the `onBeforeDestroy` method, which gets called before the component's removed from the stack.

Connects #2366

## Testing Instructions
- visit `localhost:8000?bigcz`, select an AOI including a major river, then search for "water" and open the "WDC" results
- select one of the river gauge results and open its details view. Click back and verify that the details geometry's removed from the map appopriately
- verify that the other geometries work as expected too, and that they don't linger on the map after clicking the back button

### DIFF
--- a/src/mmw/js/src/data_catalog/controllers.js
+++ b/src/mmw/js/src/data_catalog/controllers.js
@@ -64,6 +64,7 @@ var DataCatalogController = {
         App.map.set({
             dataCatalogResults: null,
             dataCatalogActiveResult: null,
+            dataCatalogDetailResult: null,
         });
         App.rootView.sidebarRegion.currentView.collection.forEach(
             function(catalogModel) {


src/mmw/js/src/data_catalog/views.js
```diff
@@ -103,6 +103,10 @@ var DataCatalogWindow = Marionette.LayoutView.extend({
         'change:detail_result': 'onDetailResultChange'
     },
 
+    onBeforeDestroy: function() {
+        App.map.set('dataCatalogDetailResult', null);
```
arottersman: I think it'd be more appropriate to add it [here](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/data_catalog/controllers.js#L64)

kellyi: Sounds good. I pushed a fixup to address this, but it seems like it hasn't shown up on this PR yet.

kellyi:
Thanks! I rebased this on develop, squashed, and tested it again. Everything still worked, so I'll merge this once the tests pass.

PR #2376, opened 2017-10-17
Enable downloading compare data to CSV

kellyi:
## Overview

This PR restores & enables the TR-55 compare view's download button, adding a `downloadCSV` method which formats the data for a CSV, transforms it into a blob, then downloads it.

Connects #2313 

## Demo

For this sq km:

screen shot 2017-10-17 at 3 40 01 pm

Current output data looks like:

```csv
scenario_name	 runoff	 evapotranspiration	 infiltration	 tss_load	 tss_runoff	 tss_loading_rate	 tn_load	 tn_runoff	 tn_loading_rate	 tp_load	 tp_runoff	 tp_loading_rate
Predominantly Forested	0.032306193244325286	0.5257799999999999	1.941913886755675	14.479092616372036	0.032306193244325286	0.14455778895737198	0.33784549438201417	0.032306193244325286	0.0033730150756720125	0.04182848978063032	0.032306193244325286	0.0004176113903212968
Current Conditions	0.04212154065751694	0.524841276465284	1.9330372628771992	24.585768671322167	0.04212154065751694	0.24546181539891856	0.5652460704251255	0.04212154065751694	0.005643359312800157	0.07417454611851056	0.04212154065751694	0.0007405511289902361
New Scenario	0	0.08124225500450855	2.4187578249954917	0	0	0	0	0	0	0	0	0
```

## Notes

Prior to the release, I checked what the old compare view data downloads included and those seemed limited to et, inf, runoff, plus tss, tn, and tp loading rate per scenario. I decided to include all of that data in a single CSV which also includes the names of the scenarios. I left out the precipitation value, but we could also add that as a column if we liked.

I also included tss, tp, and tn values which we don't currently display in the compare view. Happy to drop these if it's unnecessary.

The headers don't currently have units but we can add those, too, if desired.

The download implementation here works in Chrome, FF, IE 11, and Safari, but one quirk with the Safari version is that it doesn't seem to accept the name for the file. I tried to circumvent this in a few ways to no avail.

If we were okay adding a new dependency we could bring in `file-saver` which creates a simpler interface that works across browsers (including Safari) -- basically `fileSaver.saveAs(blob, fileName)` rather than what we've got here. I checked the Bootstrap table downloads and it looks like the export plugin we're using there has `file-saver` as a dependency.

## Testing
- get this branch, `bundle`, then visit localhost:8000 and create a TR-55 project with a few scenarios
- in Chrome, FF, IE11, and Safari click the download button in the compare view and verify that you're able to download the data and that it's accurate wrt to the data shown in the compare view

### DIFF
--- a/src/mmw/js/src/compare/templates/compareInputs.html
+++ b/src/mmw/js/src/compare/templates/compareInputs.html
@@ -3,3 +3,8 @@
     <button id="compare-input-button-chart" class="{{ 'active' if mode == 'chart' }}"><i class="fa fa-bar-chart"></i></button>
     <button id="compare-input-button-table" class="{{ 'active' if mode == 'table' }}"><i class="fa fa-table"></i></button>
 </div>
+<div class="compare-download compare-input">
+    <button id="compare-input-button-download">
+        <i class="fa fa-download"></i>
+    </button>
+</div>
--- a/src/mmw/js/src/compare/views.js
+++ b/src/mmw/js/src/compare/views.js
@@ -2,9 +2,11 @@
 
 var _ = require('lodash'),
     $ = require('jquery'),
+    moment = require('moment'),
     Marionette = require('../../shim/backbone.marionette'),
     App = require('../app'),
     coreModels = require('../core/models'),
+    coreUtils = require('../core/utils'),
     coreViews = require('../core/views'),
     chart = require('../core/chart.js'),
     modalViews = require('../core/modals/views'),
@@ -217,15 +219,25 @@ var InputsView = Marionette.LayoutView.extend({
     ui: {
         chartButton: '#compare-input-button-chart',
         tableButton: '#compare-input-button-table',
+        downloadButton: '#compare-input-button-download'
     },
 
     events: {
         'click @ui.chartButton': 'setChartView',
         'click @ui.tableButton': 'setTableView',
+        'click @ui.downloadButton': 'downloadCSV'
     },
 
     regions: {
         precipitationRegion: '.compare-precipitation',
+    },
+
+    modelEvents: {
+        'change:polling': 'toggleDownloadButtonActive'
+    },
+
+    toggleDownloadButtonActive: function() {
+        this.ui.downloadButton.prop('disabled', this.model.get('polling'));
     },
 
     onShow: function() {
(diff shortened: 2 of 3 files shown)


arottersman:
Testing this out now!

Probably better for @ajrobbins to answer on the content, but in my amateur opinion:

> I left out the precipitation value

I think this is important because it's what all the values depend on

> The headers don't currently have units but we can add those, too, if desired

I think we should add the units

ajrobbins:
+1 for adding precip value and units!

src/mmw/js/src/compare/views.js
```diff
@@ -255,6 +258,88 @@ var InputsView = Marionette.LayoutView.extend({
...
+    downloadCSV: function() {
+        var aoi = App.currentProject.get('area_of_interest'),
+            aoiVolumeModel = new tr55Models.AoiVolumeModel({ areaOfInterest: aoi }),
+            csvHeadings = [['scenario_name', 'runoff', 'evapotranspiration', 'infiltration',
+            'tss_load', 'tss_runoff', 'tss_loading_rate', 'tn_load', 'tn_runoff', 'tn_loading_rate',
+            'tp_load', 'tp_runoff', 'tp_loading_rate']],
+            csvData = this.model.get('scenarios')
+                .map(function(scenario) {
+                    var scenarioName = scenario.get('name'),
+                        result = scenario
```
arottersman: Consider adding a refactoring commit that takes [`getScenarioResults`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/compare/models.js#L48) and moves it somewhere accessible here and also, if you're up for it, [here](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/compare/models.js#L145) (see #2379).

src/mmw/js/src/compare/views.js
```diff
@@ -255,6 +258,88 @@ var InputsView = Marionette.LayoutView.extend({
...
+                            .get('results')
+                            .findWhere({ name: 'runoff' })
+                            .get('result'),
+                        runoff,
+                        quality,
+                        tss,
+                        tn,
+                        tp;
+
+                    switch (scenarioName) {
```
arottersman: If the above comment doesn't happen, prefer not switching on name, and instead using the boolean attributes `is_current_conditions` and `is_precolumbian`.

kellyi: Decided to keep this code but adjust the checks to be on `is_current_conditions` and `is_pre_columbian` rather than checking on the scenario names.

kellyi:
bc961de adds precipitation values, puts units in the column headers, and replaces the switch statement with an if check on `is_pre_columbian` etc.

kellyi:
Current output:

```csv
scenario_name	 precipitation_cm	 runoff_cm	 evapotranspiration_cm	 infiltration_cm	 tss_load_cm	 tss_runoff_cm	 tss_loading_rate_kgha	 tn_load_cm	 tn_runoff	 tn_loading_rate_kgha	 tp_load_cm	 tp_runoff_cm	 tp_loading_rate_kgha
Predominantly Forested	2.5	0.032306193244325286	0.5257799999999999	1.941913886755675	14.479092616372036	0.032306193244325286	0.14455778895737198	0.33784549438201417	0.032306193244325286	0.0033730150756720125	0.04182848978063032	0.032306193244325286	0.0004176113903212968
Current Conditions	2.5	0.04212154065751694	0.524841276465284	1.9330372628771992	24.585768671322167	0.04212154065751694	0.24546181539891856	0.5652460704251255	0.04212154065751694	0.005643359312800157	0.07417454611851056	0.04212154065751694	0.0007405511289902361
New Scenario	2.5	0	0.08124225500450855	2.4187578249954917	0	0	0	0	0	0	0	0	0
```

src/mmw/js/src/compare/templates/compareInputs.html
```diff
@@ -3,3 +3,8 @@
     <button id="compare-input-button-chart" class="{{ 'active' if mode == 'chart' }}"><i class="fa fa-bar-chart"></i></button>
     <button id="compare-input-button-table" class="{{ 'active' if mode == 'table' }}"><i class="fa fa-table"></i></button>
 </div>
+<div class="compare-download compare-input">
+    <button id="compare-input-button-download">
```
arottersman: I think the button should be disabled while polling — this will prevent the case where there's a new precipitation value in the csv, shown with stale results.

kellyi: Good idea -- cc6eb27 handles that and adds a missing unit value to the headers.

src/mmw/js/src/compare/views.js
```diff
@@ -217,17 +219,27 @@ var InputsView = Marionette.LayoutView.extend({
...
     regions: {
         precipitationRegion: '.compare-precipitation',
     },
 
+    modelEvents: {
+        'change:polling': 'toggleDownloadButtonActive'
+    },
+
+    toggleDownloadButtonActive: function() {
+        this.ui.downloadButton.prop('disabled', this.model.get('polling'));
```
arottersman: Fine with this here, but may be worth noting we've generally been doing this in the templates:

https://github.com/WikiWatershed/model-my-watershed/blob/f0c1e0093e25180fc9ab035e973c846b3a7107f1/src/mmw/js/src/data_catalog/templates/pager.html#L9

https://github.com/WikiWatershed/model-my-watershed/blob/1a836bb5f8281b7d7f2b0bedb2a2f7bf9131794a/src/mmw/js/src/modeling/templates/controls/thumbSelect.html#L11

arottersman:
> If we were okay adding a new dependency we could bring in file-saver which creates a simpler interface that works across browsers (including Safari) -- basically fileSaver.saveAs(blob, fileName) rather than what we've got here. I checked the Bootstrap table downloads and it looks like the export plugin we're using there has file-saver as a dependency.

We should reconsider this when we implement the MapShed compare view and all its downloads.

src/mmw/js/src/compare/views.js
kellyi: Ah that makes sense. I'm going to leave this as is for now, but we can change it if we encounter problems with the jquery version.

Issue #2380, opened 2017-10-18
BiG-CZ: Fetch More Data for Charts


rajadain:
Currently for the chart, we pull in about 1 month of data by default as of #XXX:

![image](https://user-images.githubusercontent.com/1430060/31737679-49830002-b416-11e7-83a7-6fd743f06dc3.png)

Add UI and network mechanisms to pull in more data. We may want to progress in stages: fetch 3 months, 6 months, 1 year, 2 years, all (up to 30 years!).

emiliom:
Given that #2402 was closed, should this issue be added to https://github.com/WikiWatershed/model-my-watershed/milestone/60? Otherwise it's easy for some of us to lose sight of it, b/c the source issue is flagged as closed. Thanks.

rajadain:
Yes, good catch.

Issue #2333, opened 2017-10-05
BiG-CZ: Show Time Series Data in Chart


rajadain:
Following up from #2238.

See mockup in https://github.com/WikiWatershed/model-my-watershed/issues/2234#issuecomment-328215676.

PR #2381, opened 2017-10-18
BiG-CZ: WDC Detail Charts

rajadain:
## Overview

Adds a chart showing the most recent 1 month of values for each variable in a result.

Connects #2333 

### Demo

Note that the dropdown isn't captured correctly in this GIF, but it's there.

### Notes

One thing we may tweak going forward is how much data do we pull in the first step for each variable. Previously it was 7 days, but here I've upped it to 1 month. The wait is around ~30s for the bigger results, but often quicker, and I think users can wait that long. If not, we can make it 2 weeks, and wait around ~22s.

Another thing I tried to capture here at first was a mechanism to pull in more results. However, this needs some UI thought, for which I'll need to pair with @jfrankl, so I'm deferring that to #2380.

## Testing Instructions

 * Go to [:8000/?bigcz](http://localhost:8000/?bigcz), select a shape, search for a term, and head to the WDC tab
 * Open a search result near the top of the list (which is ordered by recency). Once the detail values come in, you should see a chart / table switcher. Click the chart button and ensure it renders.
 * Ensure that the chart has about 1 month of data, and shows 2 weeks at a time by default. Ensure tooltips and the y-axis show the right units, and that the timestamps seem correct.
 * Ensure that for results with single observations (such as an NWISGW result), the chart shows a single value
 * Ensure that for results with failing variables (for which the `/values` endpoint fails, such as GHCN), the chart is empty
 * Ensure there are no errors in the console

### DIFF
--- a/src/mmw/js/src/data_catalog/views.js
+++ b/src/mmw/js/src/data_catalog/views.js
@@ -2,7 +2,9 @@
 
 var $ = require('jquery'),
     _ = require('lodash'),
+    Backbone = require('../../shim/backbone'),
     Marionette = require('../../shim/backbone.marionette'),
+    HighstockChart = require('../../shim/highstock'),
     App = require('../app'),
     analyzeViews = require('../analyze/views.js'),
     settings = require('../core/settings'),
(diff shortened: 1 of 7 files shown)


src/mmw/js/src/data_catalog/views.js
```diff
@@ -546,32 +533,81 @@ var ResultDetailsCuahsiView = ResultDetailsBaseView.extend({
...
+        'click @ui.chartButton': 'setChartMode',
+        'click @ui.tableButton': 'setTableMode',
+    },
+
+    modelEvents: {
+        'change:fetching change:mode': 'render',
+    },
+
+    templateHelpers: function() {
+        var id = this.model.get('id'),
```
arottersman: This little guy is causing the build to fail; it's never used.

arottersman:
Just tested this out. Working great on all browsers except for one oddity:

- Select the City Of Philadelphia Schuylkill River HUC-12
- Search water and select the fourth wdc result `NWISDV 01474500`
- Select the first temperature chart. I see this error in the console:
```
vendor.js:56 Error: <rect> attribute width: A negative value is not valid. ("-152708")
```
The chart looks like this:
screen shot 2017-10-19 at 9 23 41 am

The chart is still useable and the bad state can be corrected by selecting `1W` `2W` or `1M`, so this might be okay to put in an enhancement card.

src/mmw/js/src/data_catalog/views.js
rajadain: Ah good catch, will remove.

rajadain:
I noticed that error as well, and chalked it up to bad data at the time. Let me see if I can handle that a little more gracefully.

src/mmw/js/src/data_catalog/views.js
```diff
@@ -649,6 +685,131 @@ var CuahsiTableView = Marionette.ItemView.extend({
...
+        if (!this.chart) {
+            this.chart = this.initializeChart(variable);
+        } else {
+            this.chart.yAxis[0].setTitle({
+                text: variable.get('units'),
+            });
+
+            this.chart.series[0].update({
+                name: variable.get('concept_keyword'),
+                data: variable.get('values').map(function(v) {
```
arottersman: Consider making some function `getData` for here and `initializeChart`.

rajadain:
Regarding the negative rect thing, this feels like a bug in HighStocks itself. See [here](https://github.com/highcharts/highstock-release/blob/v2.1.9/highstock.src.js#L2968). They are trying to not set the stroke width in WebKit and Blink browsers, and check that it is not 0, but they don't check it is not negative. We're using a really old version of HighStocks (current is v6.0.1), and since this doesn't break anything in the UI, I'm willing to leave it in for now.

I'm not exactly sure what it is about this result that is triggering this fault rather than other ones (such as USGS 01474500 which works perfectly), but it is possible that each variable has a different times in its series and makes matching the window across variables impossible.

rajadain:
Put up fixups for the code comments. If you approve of them I'll squash and merge soon. Thanks!

Issue #2379, opened 2017-10-18
Compare view predominantly forested table values are wrong


arottersman:
We're [doing the same thing that we were doing with the charts](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/js/src/compare/models.js#L145), unconditionally using the TR-55 `modified` results, instead of the key corresponding to the precolumbian model run.

Abstract what happened in https://github.com/WikiWatershed/model-my-watershed/pull/2355 so it can be shared with the tables. This might look something like bringing [`getScenarioResult`](https://github.com/WikiWatershed/model-my-watershed/pull/2355/files#diff-963f699e9787e1352bdf69a6e6bc3d71R48) out into a utility function and using it by both chart and table.

Values are different:
![screen shot 2017-10-18 at 10 51 40 am](https://user-images.githubusercontent.com/7633670/31725433-5dfb3d16-b3f2-11e7-957f-5eac626ef25d.png)

![screen shot 2017-10-18 at 10 52 49 am](https://user-images.githubusercontent.com/7633670/31725488-8319543e-b3f2-11e7-884f-d2d848311f81.png)

PR #2384, opened 2017-10-18
Fix Compare View Predominantly Forested Chart Values

arottersman:
## Overview

- Extract helper functions out to handle getting tr-55 results values
- Use the helper functions in the compare charts and now tables, regular tr-55 charts and tables
- Cleanup `tr55/???/views` — still had a a bit of code referencing the old compare view

Connects #2379

### Demo

screen shot 2017-10-18 at 5 43 13 pm
screen shot 2017-10-18 at 5 43 07 pm

### Notes
`utils` probably isn't the greatest forever home for these helper functions. Someday we may wish to refactor the `ResultModel` or maybe the `ScenarioModel` so it has proper TR-55 and Mapshed subclasses, and then place the helpers on the TR-55 subclass.

## Testing Instructions

 * Pull and `bundle`
 * Create a tr-55 project for an easily findable huc. Give it a few scenarios, and cover one in forest landcover
 * Do the same for the same huc on staging
 * Compare the displayed results between staging and this branch and confirm they're the same
 * Open the compare view on this branch and confirm the chart and table values match when rounding has been accounted for

### DIFF
--- a/src/mmw/js/src/core/utils.js
+++ b/src/mmw/js/src/core/utils.js
@@ -407,13 +407,34 @@ var utils = {
     // }
     // Use the scenario attributes to figure out which subresult should be
     // accessed.
-    getTR55ResultKey: function(scenario) {
+    _getTR55ResultKey: function(scenario) {
         if (scenario.get('is_current_conditions')) {
             return 'unmodified';
         } else if (scenario.get('is_pre_columbian')) {
             return 'pc_unmodified';
         }
         return 'modified';
+    },
+
+    /** Access a scenario's result for a given type
+     *   @param {ScenarioModel} scenario
+     *   @param {string} typeKey: the type of result you want
+     *                            back; 'runoff', 'quality'
+     **/
+    _getTR55ScenarioResult: function(scenario, typeKey) {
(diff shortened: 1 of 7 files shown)


kellyi:
Taking a look at this.

src/mmw/js/src/core/utils.js
```diff
@@ -416,6 +416,27 @@ var utils = {
         return 'modified';
     },
 
+    /** Access a scenario's result for a given type
+     *   @param {ScnenarioModel} scenario
```
rajadain: Typo `ScnenarioModel` → `ScenarioModel`

kellyi:
Values now differ. Here's the same AOI (and same modifications) on staging:

screen shot 2017-10-19 at 11 36 29 am

... and on this branch ...

screen shot 2017-10-19 at 11 36 36 am

kellyi:
> If something can be surreptitiously slipped in here, I'd like the order of the rows in the table to be changed to match those of the charts.

In favor of handling this in another card.

rajadain:
Actually the Modeling View makes the same mistake (of value ordering):

image

so that should be its own card. I'll make one.

Issue #2317, opened 2017-09-29
Reset button should only appear when filters are selected


jfrankl:
When there are no filters selected in BiG CZ, we should hide the "Reset" button. The "Reset" button should appear when a filter is selected. That way, the button will never be visible to the user when clicking on it doesn't perform any action.

rajadain:
Should the Reset button clear all filters, or restore all filters to the default state? I'd think the latter. WDC for example has "Exclude Gridded" checked by default:

![image](https://user-images.githubusercontent.com/1430060/31740576-c7302b48-b41f-11e7-9b7a-a51c1b4f18e4.png)

ajrobbins:
I think default

PR #2385, opened 2017-10-18
BiG-CZ: Hide Reset Button for Default Filters

rajadain:
## Overview

In cases when the default filters are currently applied, the Reset button resulted in a no-op. This hides the button when a no-op is expected.

Also fixes an issue where resetting multiple active filters would lead to multiple parallel searches and inconsistent UI state.

Connects #2317

### Demo

## Testing Instructions

 * Go to [:8000/?bigcz](http://localhost:8000/?bigcz) and select an area. Move to the Search stage.
 * Expend the Filters sidebar. Test that setting the filters other than defaults shows the Reset button, and restoring them to the default state (either by manipulating the filter directly or by clicking the Reset button) hides the Reset button.
 * Ensure that reseting multiple filters fires a single subsequent search, after all filters have been reset
 * Ensure there are no errors in the console

### DIFF
--- a/src/mmw/js/src/data_catalog/models.js
+++ b/src/mmw/js/src/data_catalog/models.js
@@ -43,6 +43,12 @@ var FilterModel = Backbone.Model.extend({
         window.console.error("Use of unimplemented function",
                              "FilterModel.isActive");
         return false;
+    },
+
+    isDefault: function() {
+        window.console.error("Use of unimplemented function",
+                             "FilterModel.isDefault");
+        return false;
     }
 });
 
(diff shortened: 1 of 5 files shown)


ajrobbins:
The checked box being the default state is a little odd, but due to previous considerations with it I think this approach is good.

src/mmw/js/src/data_catalog/models.js
```diff
@@ -153,7 +153,17 @@ var Catalog = Backbone.Model.extend({
     },
 
     initialize: function() {
-        var self = this;
+        var self = this,
+            debouncedSearch = _.debounce(function() {
```
arottersman: The commit message is great; a comment might also be helpful on why we're debouncing the search.

rajadain:
Thanks for taking a look! I added a comment, rebased on develop, and merged.

Issue #2344, opened 2017-10-06
Geoprocessing API: Validate RWD Requests Have Required Keys


arottersman:
We currently [assume the required RWD body keys are present](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/geoprocessing_api/views.py#L171). We should be checking that they're present, and when they're not, send a response back informing the user what keys they're missing.

kellyi:
Just to clarify: this is for validating the required *request* body keys and not RWD's response body?

Looks like `location` is the only required request body keys, so we can validate that for presence. Should we be validating the value itself to check that it's the right data type and/or within conus boundaries?

arottersman:
I would say we should validate it's the right data type and that it's present, but leave the conus boundaries for potentially #2340

Issue #2342, opened 2017-10-06
Geoprocessing API: Validate Shape Does Not Self Intersect


arottersman:
We already check for self-intersections on the frontend via `leaflet-draw` and `turf-kink`, but we now have to guard against bad input from API users. Add a check on AOIs submitted to the analyze endpoints.

kellyi:
Django gis GEOSGeometry has a `.valid` method which we can use to validate these, but it will probably catch more than just the self-intersecting. Here's the list of GEOS error messages:

https://github.com/OSGeo/geos/blob/master/src/operation/valid/TopologyValidationError.cpp

```cpp
const char* TopologyValidationError::errMsg[]={
	"Topology Validation Error",
	"Repeated Point",
	"Hole lies outside shell",
	"Holes are nested",
	"Interior is disconnected",
	"Self-intersection",
	"Ring Self-intersection",
	"Nested shells",
	"Duplicate Rings",
	"Too few points in geometry component",
	"Invalid Coordinate",
	"Ring is not closed"
};
```

PR #2386, opened 2017-10-18
Validate RWD API input

kellyi:
## Overview

This PR adds some validation to the RWD API field inputs: each field has a slightly different cardinality, so the checks range from merely checking for presence to validating that the input's in the right set.

I looked into using DRF serializers, but they don't generally seem to support maybe fields without some workaround, so in place of that I wrote the checks manually and sent back `ValidationErrors` with error messages describing what went wrong & how to remedy the error.

Connects #2344

## Demo

screen shot 2017-10-18 at 6 41 31 pm

## Notes

It might be worth abstracting out the validation error message work into a separate method which we can re-use if/when we add validation to the other fields. For now the validation logic lives close to where the RWD job would have been initiated.

## Testing Instructions
- get this branch, then `bundle` and open the swagger docs
- using an auth token, try to give RWD both good and bad input and verify that the invalid inputs generate error messages & that the valid ones succeed

### DIFF
--- a/src/mmw/apps/geoprocessing_api/views.py
+++ b/src/mmw/apps/geoprocessing_api/views.py
@@ -2,6 +2,8 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
+from numbers import Number
+
 from celery import chain, group
 
 from rest_framework.response import Response
@@ -10,6 +12,7 @@ from rest_framework.authentication import (TokenAuthentication,
                                            SessionAuthentication)
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.authtoken.models import Token
+from rest_framework.exceptions import ValidationError
 
 from django.utils.timezone import now
 from django.core.urlresolvers import reverse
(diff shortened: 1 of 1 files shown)


kellyi:
I picked up #2342 also so I'll branch off this & maybe move the validation logic to a separate file

src/mmw/apps/geoprocessing_api/views.py
```diff
@@ -224,11 +227,41 @@ def start_rwd(request, format=None):
...
     snapping = request.data.get('snappingOn', False)
+
+    if type(snapping) is not bool:
+        error = ('Invalid optional `snappingOn` parameter value '
+                 '`{}`. Must be `true` or `false`.'.format(snapping))
+        raise ValidationError(error)
+
     simplify = request.data.get('simplify', False)
 
+    if type(simplify) is not bool and not isinstance(simplify, Number):
```
arottersman: Should this be
```py
if simplify is not None and not isinstance(simplify, Number):
```

src/mmw/apps/geoprocessing_api/views.py
```diff
@@ -224,11 +227,41 @@ def start_rwd(request, format=None):
...
     created = now()
-    location = request.data['location']
+    location = request.data.get('location')
+
+    def validate_location(loc):
+        if loc is None or type(loc) is not list or len(loc) is not 2:
+            return False
+        else:
+            [lat, lng] = loc
+            return type(lat) is float and type(lng) is float
```
arottersman: Should we also allow `int`? It's unlikely someone would send one, but might be difficult for them to debug if they ever do.

kellyi: Good point. Yeah, I think this should be number.

src/mmw/apps/geoprocessing_api/views.py
kellyi: The get-or-else line above assigns it to `False` currently if there's no supplied value, so it should always be either `False` or a number. This also means clients could potentially pass `false` or `true` as the value and have it succeed -- not sure whether this would have adverse effects?

arottersman: Ah, okay. Maybe it should check if `simplify is False` then?

kellyi: TIL:

```py
>>> isinstance(True, Number)
True
```

kellyi: Pushed some changes to address this. I rebased on `develop`, so went ahead and squashed the fixup too.

src/mmw/apps/geoprocessing_api/views.py
kellyi: Replaced these with checks for whether the values were numeric and were not booleans.

kellyi:
It occurred to me that we should also test whether the MMW app client passes these validations.

src/mmw/apps/geoprocessing_api/views.py
```diff
@@ -224,11 +227,43 @@ def start_rwd(request, format=None):
...
     snapping = request.data.get('snappingOn', False)
+
+    if type(snapping) is not bool:
+        error = ('Invalid optional `snappingOn` parameter value '
+                 '`{}`. Must be `true` or `false`.'.format(snapping))
+        raise ValidationError(error)
+
     simplify = request.data.get('simplify', False)
 
+    if isinstance(simplify, Number) and simplify is not False:
```
arottersman: Missing the `not`

kellyi: Oof; fixed in 23e3004

kellyi:
Thanks! Squashed & will merge once CI builds.

Issue #2404, opened 2017-10-19
Hide Temperature Slider On Account Page


arottersman:
<img width="1280" alt="screen shot 2017-10-19 at 2 36 38 pm" src="https://user-images.githubusercontent.com/7633670/31787860-f942190a-b4da-11e7-9c5c-d67e9a67dcb1.png">

- Make sure this also doesn't happen on the project page.

PR #2408, opened 2017-10-19
Hide slider on account & project pages

kellyi:
## Overview

This PR ups the z-index for the `#account-container` and `#projects-container` to ensure these display above the climate slider. Previously their z-indices were low enough that the slider showed up.

Connects #2404

### Notes

I considered also reducing the z-index for the slider's container, but I'm not sure what that number's interleaved with so this seemed to be a more reasonable fix.

## Testing Instructions
- open a climate slider, then visit the accounts page. Verify that the slider's not visible
- repeat with the projects page

### DIFF
--- a/src/mmw/js/src/account/controllers.js
+++ b/src/mmw/js/src/account/controllers.js
@@ -27,7 +27,7 @@ function showAccountView() {
             model: new models.AccountContainerModel()
         })
     );
-    App.rootView.layerPickerRegion.empty();
+    App.destroyLayerPicker();
 
     App.state.set('active_page', coreUtils.accountPageTitle);
 }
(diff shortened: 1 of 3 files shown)


kellyi:
It's worth testing in this in Chrome, FF, Safari, and IE11 -- I checked it in each for both pages and it worked.

kellyi:
> Consider adding an App.emptyLayerPicker to handle emptying the layerpicker + timeslider region in tandem.

Sounds good -- I'll make a fixup to handle it this way instead.

kellyi:
91057e0 adds a method to empty both the layer picker & slider regions on showing either projects or accounts.

kellyi:
Thanks! I squashed & rebased on `develop`. Will merge once the CI builds.

Issue #2388, opened 2017-10-19
Increase AoI limits on BiG CZ Analyze


ajrobbins:
Currently, we limit BiG CZ analyze to 1,500sq km. Increase to 8,000 sq km.

aufdenkampe:
Just discussed on call, that this issue challenging because of WDC response time outs at larger areas. The issue is only for WDC, possibly because of SOAP API, rather than REST, like the other catalogs. Also, WDC doesn't have pagination, unlike other catalogs.

Next steps could be one or both of below:
* improve our efficiency at consuming SOAP
* work with WDC to use newer web requests (some recently developed, some could be done for us).

rajadain:
Yes sorry, this is still not implemented.

Issue #2389, opened 2017-10-19
Implement client-side free text search for WDC (4 pts)


ajrobbins:
The WDC API, unlike the other two APIs we integrate with for the BiG CZ data catalog search, does not support free text search. One of the core ideas of the BiG CZ project is that the user can seamlessly search across multiple catalog and expect the same search mechanisms to be supplied, and the design (large text search box) reinforces that.

We can implement our own free-text search for WDC by:
- making the API call for all results in the AoI
- searching over all of the returned results (every field shown on the result details page + the description we put in the info button popup + the citation) for the inputted search terms

emiliom:
attributes being searched: title, id, site name, description, location; service citation & abstract & title; variables (concept keywords)

emiliom:
Request from @aufdenkampe and I: **exclude** `service citation` and `service abstract` from the free text search.

PR #2409, opened 2017-10-19
Adjust BigCZ max area from 1500 -> 8000

kellyi:
## Overview

This PR increases the max analyzable area for BigCZ from 1500 to 8000.

Connects #2388 

### Demo

screen shot 2017-10-19 at 5 05 58 pm

&

screen shot 2017-10-19 at 5 06 11 pm

## Testing Instructions
- get this branch, bundle, then visit bigcz
- try to analyze an aoi larger than 1500 but below 8000 sq km & verify that it works
- try to analyze an aoi larger than 8000 sq km & verify that you encounter the guard

### DIFF
--- a/src/mmw/mmw/settings/base.py
+++ b/src/mmw/mmw/settings/base.py
@@ -383,7 +383,7 @@ OMGEO_SETTINGS = [[
 MMW_MAX_AREA = 75000  # Max area in km2, about the size of West Virginia
 
 BIGCZ_HOST = 'portal.bigcz.org'  # BiG-CZ Host, for enabling custom behavior
-BIGCZ_MAX_AREA = 1500  # Max area in km2, limited by CUAHSI
+BIGCZ_MAX_AREA = 8000  # Max area in km2, limited by CUAHSI
 BIGCZ_CLIENT_TIMEOUT = 5  # timeout in seconds
 BIGCZ_CLIENT_PAGE_SIZE = 100
 


rajadain:
I'll take a deeper look soon, but the main reason for going with 1500 for the limit was WDC searches timing out. It may be that we need to increase the timeout for those somewhere, or constrain the results otherwise (like limiting to the last five years for example) to make it finish in time, and pair that with this to be viable. I'll know more once I've taken a look.

rajadain:
I tested this, but WDC results choke on even a 2000 sq km area of interest. Rather surprisingly, the bottleneck isn't the actual SOAP call, but this line:

https://github.com/WikiWatershed/model-my-watershed/blob/8afa4c4c245f065a53c9f8425eb443292f9129f4/src/mmw/apps/bigcz/clients/cuahsi/search.py#L212

which serializes the results so they can be cached. This caching code is in a spot that does not know about the response's internal structure, so cannot make sophisticated decisions about it.

This requires a multi-part solution:

 1. Move the caching (and `recursive_asdict`) out of `make_request`, into the level above. Continue caching GetServicesInBox2 for 1 week.
 2. For GetSeriesCatalogForBox2, make the request, then sort the `SeriesRecord`s by descending order of date. Then, pick an adequate cap `n`, and choose the top `n` records, and serialize _those_ and return to the client.

Since the above approach discards results after the first `n`, we may need to add some custom pagination logic. But this should be deferred to a later date, since we may be introducing search in #2389, allowing the user to refine their results that way.

rajadain:
Another alternative is to limit the results to be within the last 5 years if the area of interest is more than 1500 sq km. This may be usable in conjunction with the work above.

rajadain:
Closing in favor of #2418.

Issue #2405, opened 2017-10-19
Show Hydroshare point geometry results


ajrobbins:
We're currently only pulling in bb results for Hydroshare. We should also pull in results with point geometries, such as https://www.hydroshare.org/resource/1de72928f573433290f6c8bb393523df/

Issue #2390, opened 2017-10-19
Implement revised Hydroshare result card


ajrobbins:
Currently looks like:
![screen shot 2017-10-19 at 10 56 16 am](https://user-images.githubusercontent.com/12868342/31777492-2d0ddd6a-b4bc-11e7-9f2b-0da5c5380af9.png)

Waiting for mockup from Anthony + Emilio.

ajrobbins:
Mockup here: https://docs.google.com/document/d/1nsSV6Cl_BwETraKrMnB9bP5fYYRyUtt4zXrliq3Int8/edit

ajrobbins:
Notes are in the doc. Also here:

- Author First, Date Second
- Date should show Coverage Start and End instead of Creation Date
- If not Coverage Dates are available, sort to end of list and show N/A or similar
- If a record does not have spatial information, don’t show it
- If a resource is “private”, it should be filtered out by default, like Gridded Services in WDC (low priority)
-

rajadain:
The date filters wire up to the `from_date` and `to_date` request parameters in the HydroShare API that filters over `created_at` date, which will no longer be shown in the UI:

![image](https://user-images.githubusercontent.com/1430060/31836940-05fb82ce-b5a5-11e7-82bc-9e8d88c924e8.png)

This may be a reason to keep that date in the results. Any feedback on this, @aufdenkampe / @emiliom ?

emiliom:
HydroShare is not my thing, but I *think* I follow what you're asking, so I'll chime in.

The start and end dates that @aufdenkampe specified in the mockup ("coverage dates" something or other) really are the ones that are meaningful in the context of a search, but as we've said, they're also not commonly populated in HydroShare records. It's misleading, and therefore not helpful to users, to enable date filtering on dates other than these dates.

So, the Filters would need to be rewired for HydroShare. And if a user chooses to apply a date filter, all records that have no "coverage dates" will have to drop out, b/c its relevant dates are unknown.

@aufdenkampe, what do you think?

aufdenkampe:
I agree with Emilio.

Sent from my iPhone

On Oct 20, 2017, at 2:35 PM, Emilio Mayorga <notifications@github.com<mailto:notifications@github.com>> wrote:


HydroShare is not my thing, but I think I follow what you're asking, so I'll chime in.

[...]

PR #2411, opened 2017-10-20
BiG-CZ: Handle points, boxes for HydroShare results

rajadain:
## Overview

Previously we were only retrieving the first box of the various results in HydroShare. Now, we are retrieving points, as well as multiple boxes and points if there.

The results are unioned into one result, just as done for CINERGI, before sending to the client.

Connects #2405 

### Demo

image

## Testing Instructions

 * Check out this branch
 * Go to [:8000/?bigcz](http://localhost:8000/?bigcz), and search for and select "Cat Creek-Elwha River, HUC-12 Subwatershed" in Northwestern Washington State, south-west of Port Angeles
 * Proceed to search and query "water". Switch to the HydroShare tab.
 * Ensure results include points as well as boxes, and clicking / popovers / detail views work as expected both for the point and the boxes.

### DIFF
--- a/src/mmw/apps/bigcz/clients/hydroshare/search.py
+++ b/src/mmw/apps/bigcz/clients/hydroshare/search.py
@@ -8,7 +8,7 @@ import dateutil.parser
 from rest_framework.exceptions import ValidationError
 
 from django.conf import settings
-from django.contrib.gis.geos import Polygon
+from django.contrib.gis.geos import Point, Polygon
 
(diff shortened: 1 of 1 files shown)


caseycesari:
For the test location and search term, the first two results (see image below) don't appear to have any geometry. Is that expected?

image

rajadain:
We're going to exclude results without geometries in #2390.

rajadain:
Thanks for taking a look!

Issue #2395, opened 2017-10-19
Remove DRB-specific layers from BiG CZ layer selector


ajrobbins:
@emiliom to provide list of layers to keep.

emiliom:
From the Layers widget/selector (as opposed to AOI selector), **keep**:

### Streams
- Continental US Medium Resolution Stream Network

### Coverage Grid
- National Land Cover
- Hydrologic Soil Groups from gSSURGO
- Mean Monthly Precip
- Mean Monthly Temperature

### Boundary
- USGS * HUC layers
- County Lines

ajrobbins:
Thanks, @emiliom . I just created https://github.com/WikiWatershed/model-my-watershed/issues/2400 for removing non-applicable area selection options for BiG CZ, and added it to our November milestone.

emiliom:
> I just created #2400 for removing non-applicable area selection options for BiG CZ, and added it to our November milestone.

Great, thanks.

PR #2415, opened 2017-10-20
Drop DRB-specific map layers from BigCZ

kellyi:
## Overview

This PR removes the DRB-specific layers from BigCZ. To accomplish this I added a `big_cz` key to each layer in layer settings, then read them into the client & added some if logic to check whether or not to show them contextually.

Connects #2395

### Demo

MMW:

screen shot 2017-10-20 at 11 52 11 am

BigCZ:

screen shot 2017-10-20 at 11 52 17 am

### Notes

I didn't touch the Observations tab, which also has some DRB layers.

## Testing Instructions
- visit BigCZ & verify that only the layers mentioned in #2395 are included in their respective layer group
- visit MMW and verify that all the layers still appear


src/mmw/js/src/core/models.js
```diff
@@ -223,7 +225,11 @@ var LayersCollection = Backbone.Collection.extend({
             _.each(settings.get(options.type), function(layer) {
                 var layerModel = new LayerModel();
                 layerModel.buildLayer(layer, options.type, options.initialActive);
-                self.add(layerModel);
+                if (!settings.get('data_catalog_enabled')) {
```
rajadain: These two `if` conditions can be combined:

```js
if (!settings.get('data_catalog_enabled') || layerModel.get('bigCZ')) {
  self.add(layerModel);
}
```

kellyi: Yeah -- I considered doing this and thought it'd be slightly less straightforward to read, but I may join them now that we've commented on it here.

kellyi: I inlined the checks as presented above, then tested everything out again on my local. Still works as expected, so I pushed an updated commit.

kellyi:
Thanks! I pushed an amended commit & I'll merge this in once the tests pass.

PR #2416, opened 2017-10-20
BiG-CZ: HydroShare List View Updates

rajadain:
## Overview

Updates HydroShare search result list view according to:

 * Show author on the first line, date on the second
 * Instead of creation date, we now show the coverage period range
 * If coverage period range is not available, we say "Unknown Time Period" (which read better than "Time Period Unknown" to me)
 * Results are sorted by descending end of coverage date, so items without coverage dates are at the bottom
 * If a result does not have a geometry (point or box) it is excluded
 * "Private" (`public=false`) results are not shown by default. A "Exclude Private Resources" filter is added which works identically to the "Exclude Gridded" filter
 * HydroShare API requests are cached for fast filter switching

Connects #2390 

### Notes

The date filters now operate on a date that is no longer shown in the UI. I'm waiting for feedback on this here: https://github.com/WikiWatershed/model-my-watershed/issues/2390#issuecomment-338290839, which may incur some revision.

## Testing Instructions

 * Go to [:8000/?bigcz](http://localhost:8000/?bigcz) and select a shape, search for anything in HydroShare
 * Ensure that the results match the specs described in the source card

### DIFF
--- a/src/mmw/apps/bigcz/clients/cuahsi/search.py
+++ b/src/mmw/apps/bigcz/clients/cuahsi/search.py
@@ -202,8 +202,8 @@ def group_series_by_location(series):
 
 
 def make_request(request, expiry, **kwargs):
-    key = 'bigcz_{}_{}'.format(request.method.name,
-                               hash(frozenset(kwargs.items())))
+    key = 'bigcz_cuahsi_{}_{}'.format(request.method.name,
+                                      hash(frozenset(kwargs.items())))
     cached = cache.get(key)
     if cached:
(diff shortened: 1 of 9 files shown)


rajadain:
The date filter has now been resolved, and this is ready to be looked at.

rajadain:
Since code review for this hadn't yet begun, I squashed all the fixups.

arottersman:
Taking a look now!

rajadain:
Thanks for taking a look!

PR #2418, opened 2017-10-23
BiG-CZ: Increase Area of Interest Size

rajadain:
## Overview

Increases size of area of interest to 8000. Limits BiG-CZ searches to the last five years by default whenever area of interest is bigger than 1500 km².

The commit 76e0eca alleviates a CPU and RAM utilization issue where a large number of CUAHSI results would fail to serialize to JSON in order to cache. Now we don't cache the CUAHSI results. However, `suds` itself chokes at a certain point (around 800 or so results). Thus the time limit.

Builds on top of #2409 
Connects #2388 

### Notes

Should we consider adding some sort of message that explains this behavior to the user?

## Testing Instructions

 * Test large areas of interest, ensure CINERGI, HydroShare, and WDC all return values and don't error out

### DIFF
--- a/src/mmw/apps/bigcz/clients/cuahsi/search.py
+++ b/src/mmw/apps/bigcz/clients/cuahsi/search.py
@@ -201,17 +201,9 @@ def group_series_by_location(series):
     return records
 
 
-def make_request(request, expiry, **kwargs):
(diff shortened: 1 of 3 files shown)


rajadain:
We're probably going to reconsider this, so I'm pulling all review requests and closing this PR without merging for now.

Issue #2375, opened 2017-10-17
Increase AoI size limit for MapShed on Staging


ajrobbins:
The project team has stated a goal of enabling modeling and analysis for all HUC 8s nationwide. To test the feasibility of this, increase the Mapshed AoI size limit on staging to 8,000sq km and evaluate performance impacts.

PR #2419, opened 2017-10-23
Increase max size for MapShed AoI processing

mmcfarland:
## Overview

5000 km2 is a minimum threshold client specified target for running
mapshed jobs.  This was verified to run and not time out directly on
staging instances, it's unlikely that a unmodified VM dev environment
will successfully complete regularly.  There are higher thresholds
desired, and potentially better ways to accommodate them by increasing
Worker CPU counts and adjusting Celery concurrency.  To meet a short-term
testing deadline, I'm proposing this modest change as a first step.

Connects #2375 

### Demo

Here's a 3000 km<sup>2</sup> HUC8 in philadelphia that is successfully modeled, even in development:

screenshot from 2017-10-23 10 19 50

## Testing Instructions

* Mapshed areas that previously worked should continue to do so (1000 km<sup>2</sup> and below)
* Areas up to 5000 km<sup>2</sup> will likely only intermittently work in development, but were tested reliably nearing the upper limit of timeout on staging infrastructure (~30s).  You should not be warned about their size.
* Watersheds above 5000 km<sup>2</sup> should still warn and prevent modeling

### DIFF
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -78,14 +78,15 @@ var ModelSelectionDropdownView = Marionette.ItemView.extend({
         if (modelPackageName === 'gwlfe' && settings.get('mapshed_max_area')) {
             var areaInSqKm = utils.changeOfAreaUnits(aoiModel.get('area'),
                                                      aoiModel.get('units'),
-                                                     'km<sup>2</sup>');
+                                                     'km<sup>2</sup>'),
+                mapshedMaxArea = settings.get('mapshed_max_area');
 
-            if (areaInSqKm > settings.get('mapshed_max_area')) {
+            if (areaInSqKm > mapshedMaxArea) {
                 alertView = new modalViews.AlertView({
                     model: new modalModels.AlertModel({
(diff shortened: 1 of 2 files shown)


kellyi:
Testing this out & it's working as described. I was able to model some AOIs > 1000 sq km locally, but things got kind of creaky as the area increased past 2000 sq km.

AOIs over 5000 sq km are properly guarded, but it looks like we hard coded the error message value to 1k, so we should probably change that too:

screen shot 2017-10-23 at 11 02 32 am

https://github.com/WikiWatershed/model-my-watershed/blob/8afa4c4c245f065a53c9f8425eb443292f9129f4/src/mmw/js/src/analyze/views.js#L88

kellyi:
Here's a successful MapShed job for a 2500 sq km aoi:

screen shot 2017-10-23 at 11 08 38 am

mmcfarland:
Thank you for catching that, updated in f22b11f

screenshot from 2017-10-23 11 13 37

kellyi:
screen shot 2017-10-23 at 11 17 57 am

Issue #2189, opened 2017-08-24
Geoprocessing API: Setup throttling


arottersman:
Add the [DRF built-in throttling mechanism](http://www.django-rest-framework.org/api-guide/throttling/
) to limit each users API requests to `~2 req/min`. (We'll likely be tweaking this value). Throttling should only affect only API users, so we'll have to add a custom throttling class that will skip throttling if the user is the client app.

The DRF stores the number of requests already made in the Django cache. We should already have this set up, but additional configuration may be necessary.

arottersman:
Depends on #2187

arottersman:
I'm not sure what led me to write `~2req/min`. That seems much too low. The [climate-change-api uses 5000/day, 20/minute max](https://github.com/azavea/climate-change-api/blob/34a17fa753dd2b1f4d1b20d6b960b347d80733dd/django/climate_change_api/climate_change_api/settings.py#L342), and this seems more reasonable to start.

Issue #2146, opened 2017-08-14
Upgrade Django and related libraries


mmcfarland:
Our version of django is approaching the limits of support and should be upgraded to the latest LTS release.  This should include complimentary library upgrades.

ajrobbins:
In order to give adequate time to test these upgrades before releasing them, the PR from this work should point at `develop` (not the release branch).

rajadain:
We should probably bring this back in at some point. 1.8, the version we're on, ends support in April 2018.

PR #2421, opened 2017-10-23
Throttle Geoprocessing API Requests

arottersman:
## Overview

* Add a custom throttle base class to only throttle true users of the
  API and not the client app

* Throttle users at a burst (per minute rate) and a sustained (per day rate). Settings
  are from the [climate-api](https://github.com/azavea/climate-change-api/blob/34a17fa753dd2b1f4d1b20d6b960b347d80733dd/django/climate_change_api/climate_change_api/settings.py#L342). We may need to finetune them at a later date

* Throttle all start-job geoproccessing api endpoints

Connects #2189 

### Demo

```sh
> curl -H "Content-Type: application/json" \
             -H "Authorization: Token f19b20fe829f92ed176d8f90a1606531b445727c" \
             -X POST -d '{ "location": [39.67185812402583,-75.76742706298828] }' \
             localhost:8000/api/watershed/

{"detail":"Request was throttled. Expected available in 59 seconds."}
```
### Notes

I wrote some (admittedly not too useful) tests for this, but they only pass if you run them independently of the other tests (ie `./scripts/manage.sh test apps.geoprocessing_api.tests`). This is because they rely on django's `override_settings` function to adjust the throttle rates, but when there are certain DRF imports before they're called, the settings never get overriden. This is a [known DRF issue](https://github.com/encode/django-rest-framework/issues/2466) which may be somewhat fixed in later releases. Popping the tests here in case the issue is fixed when we upgrade in #2146....

[throttling_test.patch.txt](https://github.com/WikiWatershed/model-my-watershed/files/1407576/throttling_test.patch.txt)

## Testing Instructions
1. Pull this branch
2. Login into the app, go to `/account` and copy your API key
3. Throw the following script into some file like `hitit.sh` (replacing `<your_key>` with your key...)
```sh
#!/bin/bash
for number in {1..21}
do
        curl -H "Content-Type: application/json" \
             -H "Authorization: Token <your_key>" \
             -X POST -d '{ "location": [39.67185812402583,-75.76742706298828] }' \
             localhost:8000/api/watershed/
done
exit 0
```
4. `./hitit.sh | grep throttling` should show only one throttled request, the last one
5. Update the `DEFAULT_THROTTLE_RATES` in `settings/base.py` so that sustained is `10/day`
6. Wait a minute and then run `hitit.sh` again. The last 10 requests should fail
7. Grab the client app's token off of a browser analyze or rwd request. Replace your token with it in `hitit.sh` and confirm all requests succeed

### DIFF
--- a/src/mmw/apps/geoprocessing_api/throttling.py
+++ b/src/mmw/apps/geoprocessing_api/throttling.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from django.conf import settings
+
+from rest_framework.throttling import UserRateThrottle
+
+
+class GeoprocessingApiUserRateThrottle(UserRateThrottle):
+    """ Only throttle "real" users of the API, not the client app"""
+    def allow_request(self, request, view):
+        if request.user and \
+           request.user.username == settings.CLIENT_APP_USERNAME:
+            # If the user is the client app, pass through and
+            # don't attempt any throttling
+            return True
+
+        return UserRateThrottle.allow_request(self, request, view)
+
+
+class BurstRateThrottle(GeoprocessingApiUserRateThrottle):
+    scope = 'burst'
+
+
+class SustainedRateThrottle(GeoprocessingApiUserRateThrottle):
+    scope = 'sustained'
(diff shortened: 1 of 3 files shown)


arottersman:
Thanks!

Issue #2400, opened 2017-10-19
Remove not applicable AoI selection options for BiG CZ


ajrobbins:
BiG CZ users do not have the same interests/needs as MMW users, so remove the non-applicable select area options from the BiG CZ version (from https://github.com/WikiWatershed/model-my-watershed/issues/2395#issuecomment-337945335), **keeping** only:

### Select boundary
- USGS * HUC layers
- County Lines

### Delineate Watershed
- Continental US Medium Resolution

PR #2422, opened 2017-10-23
Drop some AOI selections for BigCZ

kellyi:
## Overview

This PR removes some AOI selection possibilities from BigCZ, leaving them as is in MMW. Per #2400, we keep the USGS HUCs as well as county lines but drop school districts and congressional districts. We also drop DRB RWD, leaving the NHD version.

Connects #2400 

### Demo

drb-draw-tools

### Notes

It felt nicer to do the known boundary filter using an ES5 filter rather than underscore's, so I dropped the underscore import from there, too.

## Testing Instructions
- get this branch, then bundle
- visit mmw and verify that all the draw tools are still in place
- visit bigcz and verify that the selections now match the changes requested in #2400


kellyi:
Thanks!

Issue #2392, opened 2017-10-19
Implement revised CINERGI result card


ajrobbins:
Currently looks like:
![screen shot 2017-10-19 at 10 58 04 am](https://user-images.githubusercontent.com/12868342/31777595-662714b8-b4bc-11e7-9d03-94da4dead963.png)

Waiting for mockup from Anthony + Emilio.

emiliom:
Final mockup is here: https://docs.google.com/document/d/1NzDzEQ9a0cW3aEDRtrS9tHxazYsq6ScaYsnKOf7S3mw/edit

Anthony and I went over this, so it's good to go. Let me know if you have any questions. I realize the "code" at the bottom may be a bit tricky to understand.

Also referencing this old CINERGI issue, #1929. Please close it once this and the Detail View (#2393) are implemented.

rajadain:
Hi @emiliom, here are some questions regarding this:

 1. A number of results have `1/1/1900` as the `timeperiod_nst` value:

    ![screen shot 2017-10-24 at 3 34 06 pm](https://user-images.githubusercontent.com/7633670/31964199-47e0869c-b8d1-11e7-8d17-ff9da709b404.png)

    Should these be shown as is or should there be any special treatment?

 2. There are a number of results with `ScienceBased_processed` as the source name:

    ![screen shot 2017-10-24 at 3 53 13 pm](https://user-images.githubusercontent.com/7633670/31965072-21972f7e-b8d4-11e7-8c05-9f30bfae451b.png)

    Should these have any special treatment?

 3. When the source organization name is too long, should it be `...`'d like the first line, or wrapped like the categories?

    ![image](https://user-images.githubusercontent.com/1430060/32058006-dccad9a4-ba36-11e7-92fa-1f16d45f8096.png)

emiliom:
1: keep as is
2: no special treatment
3: ....'d


On Oct 26, 2017 7:27 AM, "Terence Tuhinanshu" <notifications@github.com>
wrote:

Hi @emiliom <https://github.com/emiliom>, here are some questions regarding
this:

   1.

   A number of results have 1/1/1900 as the timeperiod_nst value:

   [image: screen shot 2017-10-24 at 3 34 06 pm]
   <https://user-images.githubusercontent.com/7633670/31964199-47e0869c-b8d1-11e7-8d17-ff9da709b404.png>

   Should these be shown as is or should there be any special treatment?
   2.

[...]

PR #2425, opened 2017-10-23
Cinergi Result List View

arottersman:
## Overview
[Template here](https://docs.google.com/document/d/1NzDzEQ9a0cW3aEDRtrS9tHxazYsq6ScaYsnKOf7S3mw/edit)

Adds and displays new CINERGI fields in a CINERGI-specific search result. The displayed date is now the `timeperiod_nst` field which is what the date filters were already searching on.

Connects #2392 
### Demo

screen shot 2017-10-24 at 3 33 32 pm

#### Many dates are garbage: 1/1/1900
We might want to filter these values out.

<details>

<summary>pic </summary>

screen shot 2017-10-24 at 3 34 06 pm

</details>

#### Some dates have same start and end
We might want to display these as a single date instead of range.

<details>

<summary>pic </summary>

screen shot 2017-10-24 at 3 34 54 pm

</details>

#### Some raw descriptions contain html

I somewhat addressed this in 73f398f. Was surprised nunjunks didn't escape the html characters for us.

<details>

<summary>pics</summary>

rendering weird
screen shot 2017-10-23 at 5 13 27 pm

dom
screen shot 2017-10-23 at 5 14 00 pm
</details>

#### A lot of sources are `ScienceBase_processed`, which is ugly

<details>
<summary>pic</summary>

screen shot 2017-10-24 at 3 53 13 pm

</details>

#### Description: 1-line vs 2-lines
It was a little unclear in the template if the description should be shortened to one line. We can drop 
faadb1c if we'd prefer to keep it the length it was.

## Testing Instructions

 * Pull and `bundle`
 * Search somewhere with CINERGI results and visually check the results
 * Confirm you can use the date filters in all sorts of combinations and that the dates in the results conform to the filters


rajadain:
Taking a look now.

rajadain:
I'm going to post some of these questions to the issue and tag Anthony and Emilio to see if we can get their feedback.

src/mmw/js/src/data_catalog/views.js
```diff
@@ -416,6 +416,25 @@ var StaticResultView = Marionette.ItemView.extend({
                                               .join('; '),
             };
         }
+
+        if (this.options.catalog === 'cinergi') {
+            var categories = _.clone(this.model.get('categories'));
+
+            if (!categories) {
+                return null;
```
rajadain: Should we return `null` or `{ 'top_categories': null }` here?

arottersman: It doesn't seem to make a difference to the template, though in the detail view PR I refactored this and it always returns [`{'top_categories': ??}`](https://github.com/WikiWatershed/model-my-watershed/pull/2429/files#diff-068d5676bbc1115f7893be5af4d3e7d0R422)

rajadain: Since it is refactored in the following PR anyway, let's leave it as is now. Interesting to know that Nunjucks treats `null` and `undefined` as equivalent.

rajadain:
Looks like we should keep everything as is: https://github.com/WikiWatershed/model-my-watershed/issues/2392#issuecomment-339702511 (which also makes sense from a turbo pterodactyl perspective). The only update here should be to only show one date when begin and end date are the same.

arottersman:
Updated the template for the dates (force pushed a fixed up 49e5a14)

Will merge when the build completes.

screen shot 2017-10-26 at 12 13 41 pm

PR #2433, opened 2017-10-26
Update, don't overwrite, DRF Production Settings

arottersman:
Staging broke because we're overwriting the DRF settings in `settings/production.py` and losing the default throttle rates. 

* Remove `DEFAULT_PERMISSION_CLASSES` from settings/base.py.
  These have never really seen the light of day because of the
  overwriting, and we use specific permissions for each view

* Use update to modify the REST_FRAMEWORK settings in the
  production settings

Connects #2421 

## Testing Instructions

 * `VAGRANT_ENV=TEST vagrant provision app` and confirm the analyze endpoints still work

### DIFF
--- a/src/mmw/mmw/settings/base.py
+++ b/src/mmw/mmw/settings/base.py
@@ -300,11 +300,6 @@ THIRD_PARTY_APPS = (
 
 # rest_framework
 REST_FRAMEWORK = {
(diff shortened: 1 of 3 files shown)


arottersman:
Thanks! Going to merge this so we don't break staging on our next merge.

PR #2439, opened 2017-10-27
BiG-CZ: WDC Naïve Free Text Search

rajadain:
## Overview

Adds free text search to WDC results. Since the CUAHSI API does not support text filtering, only geometry and date filters, the results must be filtered on the client side.

This is done by add a `serverResults` field to the `Catalog` model, which saves the server results in CUAHSI's case, transferring only those satisfying a `textFilter` to `results`. Since everything else (sidebar entries, map items) are connected to `results`, they all update accordingly.

The search implementation is very simple: for every word in the query (where words are separated by a single space) it searches a set of CUAHSI fields on the result: id, title, citation, service, organization, sample mediums, and concept keywords. For a result to be selected, it must match all the words in the query. There is no support for "OR".

Connects #2389

### Notes

When checking and unchecking the Gridded Services filter, the order of the displayed results changes, even though they remain the same results. I'm going to spin this off into a later card.

## Testing Instructions

 * Select a shape and proceed to search
 * Search for something in WDC. Ensure the results reflect the search term.
 * Try searching for something else. Ensure no server queries are fired.
 * Try switching to another tab and searching for something different there. Ensure that when you return to WDC the new search is performed and the results reflect this.
 * Try the filters in the sidebar and ensure they work correctly with the search terms

### DIFF
--- a/src/mmw/js/src/data_catalog/controllers.js
+++ b/src/mmw/js/src/data_catalog/controllers.js
@@ -46,6 +46,7 @@ var DataCatalogController = {
                 name: 'WDC',
                 is_pageable: false,
                 results: new models.Results(null, { catalog: 'cuahsi' }),
+                serverResults: new models.Results(null, { catalog: 'cuahsi' }),
                 filters: new models.FilterCollection([
                     dateFilter,
                     new models.GriddedServicesFilter()
--- a/src/mmw/js/src/data_catalog/models.js
+++ b/src/mmw/js/src/data_catalog/models.js
@@ -191,8 +191,25 @@ var Catalog = Backbone.Model.extend({
         var self = this,
             error = this.get('error'),
             stale = this.get('stale'),
-            isSameSearch = query === this.get('query') &&
-                           geom === this.get('geom');
+            isCuahsi = this.id === 'cuahsi',
+            isSameQuery = query === this.get('query'),
+            isSameGeom = geom === this.get('geom'),
+            isSameSearch = isSameQuery && isSameGeom;
+
+        if (isCuahsi && isSameGeom && !isSameQuery) {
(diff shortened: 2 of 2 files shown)


arottersman:
Testing this out now. Took a quick look at the code and looks clean. Just out of curiosity, any reason you decided to do this on the frontend?

rajadain:
> any reason you decided to do this on the frontend?

Ah sorry, probably should've addressed this in the overview. The constraints of the problem are:

 1. CUAHSI API does not support text search
 2. Our Python middleware back-end does not save any state
 3. When a user enters a text value in the front-end, we want to filter the results to it

I did this in the front-end rather than the back-end because of (2) above. Given these constraints, the options seemed to be:

 1. Fetch all values for a geometry from CUAHSI once, and filter on the front-end
 2. Fetch all values for geometry from CUAHSI every time, and filter on the back-end
 3. Fetch all values for geometry from CUAHSI once, save state in the back-end, and filter on the back-end

While (2) may be best for API consistency with the other two (for example, we take a polygon but internally convert it to bounding box and filter out non-matching values within the back-end before returning to the front-end) it would have been too slow to rerun the search every time the user adds text. (1) was best for UI interaction (because of the speed of searching). While (3) could potentially do both, it requires further questions answered (how long to save the values? how to identify a session? do we save in redis or database? etc.)  Thus by elimination, optimizing for the current most likely use case and complexity of implementation, targeting the purposes of the November Demo release, I went with (1) above.

rajadain:
The fixup takes care of multiple spaces in the search query.

rajadain:
Latest fixup adds the description / abstract to set of fields being searched, since it is so fast. Going to squash and merge shortly, thanks for taking a look!

Issue #2436, opened 2017-10-27
HydroShare Repeat Search Fail


rajadain:
Search for "water" in HydroShare. Then search for "river". We get the following error:

```
Internal Server Error: /bigcz/search
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/views.py", line 452, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/views.py", line 449, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/decorators.py", line 50, in handler
    return func(*args, **kwargs)
  File "/opt/app/apps/bigcz/views.py", line 158, in search
    return Response(_do_search(request))
  File "/opt/app/apps/bigcz/views.py", line 72, in _do_search
    results = search(**search_kwargs)
  File "/opt/app/apps/bigcz/clients/hydroshare/search.py", line 178, in search
    reverse=True)
TypeError: can't compare offset-naive and offset-aware datetimes
```

PR #2440, opened 2017-10-30
BiG-CZ: Make HydroShare dates timezone aware

rajadain:
## Overview

Some of the dates in the API response were timezone aware, while others were not. Since DATE_MIN was defined without timezone awareness, it could only be compared to similarly naive datetimes.

Using Django's make_aware utility, DATE_MIN is initialized with a timezone. Values coming in from the API are checked for timezone awareness, and are made aware if needed.

This ensures that no information is lost, and comparisons are made between like objects.

Connects #2436 

## Testing Instructions

 * Check out this branch, go to [:8000/?bigcz](http://localhost:8000/?bigcz)
 * Select the Philadelphia HUC-12, search for "water". Ensure results show up and there are no errors.
 * Search for "river". Ensure results show up and there are no errors.
 * Test anything else you can think of.

### DIFF
--- a/src/mmw/apps/bigcz/clients/hydroshare/search.py
+++ b/src/mmw/apps/bigcz/clients/hydroshare/search.py
@@ -11,6 +11,7 @@ from rest_framework.exceptions import ValidationError
 from django.core.cache import cache
 from django.conf import settings
 from django.contrib.gis.geos import Point, Polygon
+from django.utils.timezone import make_aware
 
 from apps.bigcz.models import ResourceLink, ResourceList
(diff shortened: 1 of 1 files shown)


arottersman:
Taking a look now.

arottersman:
Able to complete multiple searches, but now using the date filters causes the same naive date error.

rajadain:
oooh good catch, will fix!

rajadain:
Fixup should take care of the filter problems.

rajadain:
Thanks for taking a look! Squashed and rebased on `develop`, merging now.

Issue #2417, opened 2017-10-23
Bug harvesting data from WDC for services other than NWISUV & NWISDV


emiliom:
At our last call @rajadain mentioned something wrong with the GHCN service available via WDC, such that the data are not plotted in the detailed view. I've done [some digging](https://github.com/BiG-CZ/BiG-CZ-Portal/issues/5) and tests on those other services. I also discussed this with @aufdenkampe on Thursday, but before I did more thorough testing.

I think the bug is on the portal code itself, and I suspect it has to do with trying to query data from services that don't include recent data (the last 1-3 months, the default window used by th BiG CZ portal).

The problem is not just with GHCN. No service other than NWISUV & NWISDV seems to work, even NWISGW. I've taken two sites identified via the portal ([one from GHCN](https://apps.hydroshare.org/apps/timeseries-viewer/?Source=cuahsi&WofUri=cuahsi-wdc-2017-10-21-4707154&QCLID=0&MethodId=0&SourceId=0) and [one from CocoRAHs](https://apps.hydroshare.org/apps/timeseries-viewer/?Source=cuahsi&WofUri=cuahsi-wdc-2017-10-22-23548719&QCLID=0&MethodId=0&SourceId=0) -- note these maybe temporary links), and tried them on the [CUAHSI/WDC HydroClient](http://data.cuahsi.org). In both cases, the sites were fine on the latter.

If confirmed, this is an important bug. WDC data access that's limited to NWISUV & NWISDV is a very constrained implementation.

cc @ajrobbins

rajadain:
Thanks for the investigation and examples @emiliom, this is a great catch. I'll take a look when I get the chance.

rajadain:
Hi @emiliom, I just made a gist demoing the error we get for CoCoRaHs values: https://gist.github.com/rajadain/22b0bc5546fcd0dfdc0dba874bb31698. In this gist, I'm first doing a GetSeriesCatalogForBox2 search, then finding a CocoRaHs series record, then fetching date range for it using the ULMO's `siteinfo`, and finally fetching values using ULMO's `values`, where the failure happens.

Looking at the error output it seems to be related to how the return values are being parsed. Do you have any insight into this?

emiliom:
@rajadain I'm looking into it. I'll try to follow up today; otherwise, definitely by early Monday.

emiliom:
[...]

Issue #2413, opened 2017-10-20
KeyError: 'id'


rajadain:
View details in Rollbar: [https://rollbar.com/Azavea/MMW/items/235/](https://rollbar.com/Azavea/MMW/items/235/)

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/views.py", line 452, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/views.py", line 449, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/decorators.py", line 50, in handler
    return func(*args, **kwargs)
  File "/opt/app/apps/bigcz/views.py", line 170, in values
    return Response(_get_values(request))
  File "/opt/app/apps/bigcz/views.py", line 150, in _get_values
    return values(**values_kwargs)
  File "/opt/app/apps/bigcz/clients/cuahsi/details.py", line 55, in values
    return wof.get_values(wsdl, site, variable, from_date, to_date)
  File "/usr/local/lib/python2.7/dist-packages/ulmo/cuahsi/wof/core.py", line 185, in get_values
    values = waterml.v1_1.parse_site_values(response_buffer)
  File "/usr/local/lib/python2.7/dist-packages/ulmo/waterml/v1_1.py", line 16, in parse_site_values
    methods=methods)
  File "/usr/local/lib/python2.7/dist-packages/ulmo/waterml/common.py", line 52, in parse_site_values
    values_element, metadata_elements, namespace)
  File "/usr/local/lib/python2.7/dist-packages/ulmo/waterml/common.py", line 241, in _parse_metadata
    for item in collection
KeyError: 'id'
```

rajadain:
Don't think we can do much about this error. Probably just need to wrap this in a `try` block and return a managed exception code.

rajadain:
[...]

PR #2446, opened 2017-10-31
BiG-CZ: Fix WDC Fetch Errors

rajadain:
## Overview

Previously our workflow was to get date range from the `/details` endpoint (which uses ulmo's `get_site_info` call), and use that date to query a month's worth of data from `/values` (which uses ulmo's `get_values` call). Unfortunately, for some service providers, this was erroring out.

To fix this, @emiliom has made a fork of the project which addresses that bug here: https://github.com/emiliom/ulmo/commit/90dbfe31f38a72ea4cee9a52e572cfa8f8484adc. Thus, we switch to using his fork instead of the main distribution.

Another issue was that for some services the `/details` endpoint would return incorrect date ranges, and the true data would be outside of them. To get those values, we now fetch a year's worth of data instead of a month's.

A yet another issue was us incorrectly specifying the same start and end dates for services which had data for only one day. This caused the retrieval of only those values recorded at midnight, and not any recorded at other times during the day. We now specify the end date as one day after the start date to capture all values collected at any time during that day.

Finally, some responses had a lot of "no data" values. The no data value is specified in the API response, thus we now filter it out from those shown in the table and charts.

Note that time spent reviewing this should be billed to WPF, not BiG-CZ.

Connects #2417 
Connects #2413 

### Demo

Demo showing retrieval from CoCoRaHs, ShaleNetworkODM, GHCN, and NWISDV:

out-final

### Notes

In cases of missing data or unaligned time series for different variables, there are errors in the console. This has not changed from the previous implementation. This does not impede the app's functionality in any substantial way, so I'm letting it be for now.

Since we are now fetching a year's worth of values, the NWISUV results will take a little longer to load. However, according to the clients this is an acceptable compromise to get all the rest to work.

## Testing Instructions

 * Check out this branch, and reprovision `app`, and `bundle`
 * Select a shape and search for something. Switch to the WDC tab.
 * Ensure all the different kinds of sources work in the details view: they should all have some data in the table, and some data in the chart

### DIFF
--- a/src/mmw/js/src/data_catalog/models.js
+++ b/src/mmw/js/src/data_catalog/models.js
@@ -752,24 +752,37 @@ var CuahsiVariable = Backbone.Model.extend({
     },
 
     parse: function(response) {
-        var mrv = null;
+        var values = this.get('values'),
+            ndv = response.variable.no_data_value,
+            newValues = response.values,
+            variable = {
+                name: response.variable.name,
+                data_type: response.variable.data_type,
+                sample_medium: response.variable.sample_medium,
+                units: response.variable.units.abbreviation,
+                most_recent_value: null,
+            };
 
-        if (response.values && response.values.length > 0) {
-            var values = this.get('values');
-
-            values.reset(response.values);
-            mrv = response.values[response.values.length - 1].value;
-        } else {
+        if (newValues === null || newValues === undefined || _.isEmpty(newValues)) {
             this.set('error', 'No values returned from API');
+            return variable;
         }
 
-        return {
-            name: response.variable.name,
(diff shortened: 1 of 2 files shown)


arottersman:
Had to `sudo pip uninstall ulmo` to get the new ulmo to install.

rajadain:
> Had to `sudo pip uninstall ulmo` to get the new ulmo to install.

Ah I didn't have to do that because I destroyed and recreated `app` rather than just reprovisioning it as I instructed. Fortunately, that is what deployments tend to do, so this shouldn't be a problem there. Thanks for taking a look! Merging now.

Issue #2447, opened 2017-10-31
Cancelled RWD selection doesn't remove clicked point from map


kellyi:
Selecting RWD and clicking a point, then clicking cancel doesn't clear the original point from the map:

![screen shot 2017-10-31 at 12 28 04 pm](https://user-images.githubusercontent.com/4165523/32235979-4df73a38-be37-11e7-8240-2575aa5a6341.png)

To replicate:
1. select an RWD option and click a point to start the operation
2. immediately cancel the operation

The RWD point remains on the map, and it seems to persist if you analyze a different aoi.

In #1771 we cleared the RWD points prior to opening a new project; assuming there's not a regression it may be that we need to generalize that fix or make a similar fix for cancelled RWD ops.

PR #2458, opened 2017-11-01
Remove RWD point if RWD cancel button is clicked

caseycesari:
## Overview

Work in https://github.com/WikiWatershed/model-my-watershed/pull/2171 preserved RWD map layers on failure. A side-effect of this work was that the initial RWD point was also preserved on cancellation.

The point is now explicitly removed when the cancel button is clicked.

Connects to #2447

### Demo

mmw1

### Notes

This is a bit of a quick fix. Initially, I tried to fix this by reworking how `resetDrawingState` worked and making sure that the point was cleared when the job was cancelled (as opposed to when the job failed). However, clicking the cancel button does not always result in a canceled job. I dug into how the `TaskModel` cancels jobs, and decided to not attempt to make any changes there since the `TaskModel` is used extensively throughout the site and this bug is rather small.

## Testing Instructions

- Start the app and activate the RWD tool.
- Before the job completes, attempt to cancel it using the `x` button on the draw UI.
- The clicked point should be removed.
- Select the RWD tool again, and select a point that will generate a very large watershed that can't be analyzed. See https://github.com/WikiWatershed/model-my-watershed/pull/2171#issuecomment-324097881 for an example).
- When the job completes, the AoI and RWD points should remain on the map.

### DIFF
--- a/src/mmw/js/src/draw/views.js
+++ b/src/mmw/js/src/draw/views.js
@@ -173,7 +173,12 @@ var DrawWindow = Marionette.LayoutView.extend({
 
     onShow: function() {
         var self = this,
-            resetRwdDrawingState = function() {
+            resetRwdDrawingState = function(e) {
+                // If the cancel button was clicked
+                if (e && e.type === 'click') {
+                    self.model.clearRwdClickedPoint(App.getLeafletMap());
+                }
+
                 self.resetDrawingState({
                     clearOnFailure: false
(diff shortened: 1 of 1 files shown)


ajrobbins:
Can you retarget this at `release/1.21.0`?

caseycesari:
@ajrobbins Done.

Issue #2360, opened 2017-10-10
Climate analyze tab cut off in IE11 + FF


arottersman:
The analyze climate view is too tall in IE11 and FF; you can't scroll to the bottom

![screen shot 2017-10-10 at 4 46 15 pm](https://user-images.githubusercontent.com/7633670/31409759-f16f1f7c-adda-11e7-8e73-81b77ab74244.png)

aufdenkampe:
Also in cut off in Chrome. Works in Safari.

PR #2459, opened 2017-11-01
Adjust Analyze height in tabbed view

rajadain:
## Overview

When the Analyze window is in a tab, when both analyze and data are available as tabs, its contents weren't scrolling correctly.

By calculating the height and taking into account other items on the page, we ensure that the container's height is set in a way that allows all content to scroll and be visible.

Connects #2360 

## Testing Instructions

 * Select an area and search for something. Go back to the Analyze tab, ensure you can scroll and see all the contents correctly.
 * Ensure search results scroll correctly and their area hasn't changed.
 * Ensure detail views scroll correctly and their area hasn't changed.
 * Test in various browsers.

### DIFF
--- a/src/mmw/sass/components/_tabs.scss
+++ b/src/mmw/sass/components/_tabs.scss
@@ -68,6 +68,12 @@
     .tab-content {
       // Set a specific height to allow y-scroll:
       height: 100% !important;
+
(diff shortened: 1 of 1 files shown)


rajadain:
Thanks for taking a look!

Issue #2332, opened 2017-10-05
ITSI Screenshots: Compare View Scroll State


rajadain:
Currently if you open the Compare View and scroll down, and then take a screenshot, the screenshot is of the state of the compare view being at the top. Make it so the scroll state is preserved in the screenshot.

PR #2463, opened 2017-11-01
Mark Compare View scroll state for ITSI screenshot

rajadain:
## Overview

Previously any scrolling in the compare view would be lost when the screen was scraped for a screenshot. Now we record the current scroll state into the child's margin-top property so that the screenshotting code gets it as part of the HTML.

Connects #2332 

### Demo

While taking a screenshot:

screen shot 2017-11-01 at 5 31 12 pm

The screenshot I took:

fe58a3015c2d00d56d34998ac2eade5b917e4d9e

Page state when the screenshot was taken:

screen shot 2017-11-01 at 5 31 38 pm

### Notes

There's probably other places in the app where we could apply a similar treatment (modeling sidebar), but none so egregious as this. If we need to do it for the others, we can tackle them similarly at a later date.

## Testing Instructions

 * Have your local stack up and running, maybe via ngrok
 * Go to https://authoring.concord.org/activities/5644/single_page/dec56c68-6170-4e59-b94a-82fb040e8d88 and click on "Start Interactive" for MMW. Then, replace the iframe with your local stack.
 * Open a saved project, or create a new TR-55 one and make a few scenarios
 * Open the compare view. Scroll down a bit.
 * Take a screenshot. Ensure that the screenshot generally reflects the scroll state, and after the screenshot is taken the app behaves correctly.

### DIFF
--- a/src/mmw/js/src/app.js
+++ b/src/mmw/js/src/app.js
@@ -230,12 +230,16 @@ function toggleCompareViewForITSIScreenshot(adjustForScreenshot) {
         compareCloseButton = '.compare-close',
         compareChartButton = '#compare-input-button-chart',
         compareTableButton = '#compare-input-button-table',
+        compareSections = '.compare-sections',
         compareChartRow = '.compare-chart-row',
         compareTableRow = '.compare-table-row',
         compareScenariosRow = '.compare-scenarios',
         compareMapsRow = '.compare-scenario-row-content';
 
     if (adjustForScreenshot) {
+        $(compareSections + '> div').css('margin-top',
+            '-' + $(compareSections).scrollTop() + 'px'
+        );
         $(compareDialog).addClass(itsiCompareDialog);
         $(compareModalContent).addClass(itsiCompareModal);
         $(compareCloseButton).hide();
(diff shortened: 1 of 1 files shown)


kellyi:
About to take a look at this!

kellyi:
Snapshot now reflects the scroll position!

screen shot 2017-11-02 at 10 11 06 am

Not sure it's worth spending time adjusting this but I noticed that after the screenshot's been taken, the compare view now scrolls back up to the top even though it had been scrolled down before. Ideally it wouldn't but in it's just a quirky bug that doesn't break anything.

rajadain:
> after the screenshot's been taken, the compare view now scrolls back up to the top even though it had been scrolled down before

Hmm I didn't see that behavior, but I was testing in Firefox. Will try Chrome.

kellyi:
For good measure I verified that this also works on the water quality tab:

screen shot 2017-11-02 at 10 14 56 am

src/mmw/js/src/app.js
```diff
@@ -261,6 +265,7 @@ function toggleCompareViewForITSIScreenshot(adjustForScreenshot) {
         } else if ($(compareTableRow).length) {
             $(compareTableRow).removeClass(itsiCompareRow);
         }
+        $(compareSections + '> div').css('margin-top', '');
```
kellyi: Out of curiosity I checked what happened if we removed this line (on the notion that it might prevent the post-screenshot `scrollTop` behavior). Removing it did leave the scroll position visibly the same BUT it also truncated the overflow altogether such that I couldn't scroll back above the visible lines.

rajadain:
Can you please take another look? I've changed it so that

 1. We are recording the scroll position / margin-top before all the other changes, which can actually lead to a different value of the scroll position
 2. After resetting the margin-top, we also reset the scroll position

Now the screenshot should be accurate, while not perturbing the visual application state.

kellyi:
Taking another look at this.

rajadain:
Thanks for taking a look!

Issue #2460, opened 2017-11-01
TypeError: 'successCallback' is an invalid keyword argument for this function


rajadain:
View details in Rollbar: [https://rollbar.com/Azavea/MMW/items/243/](https://rollbar.com/Azavea/MMW/items/243/)

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/views.py", line 452, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/views.py", line 449, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/decorators.py", line 50, in handler
    return func(*args, **kwargs)
  File "/opt/app/apps/user/views.py", line 112, in profile
    profile = UserProfile(**params)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py", line 480, in __init__
    raise TypeError("'%s' is an invalid keyword argument for this function" % list(kwargs)[0])
TypeError: 'server_errors[]' is an invalid keyword argument for this function
```

arottersman:
We may have to selectively send the profile data. We currently send all model attributes, one of which is a function: 

https://github.com/WikiWatershed/model-my-watershed/blob/5d80d579aadfdefd6265647be1d8a32f4925cf80/src/mmw/js/src/app.js#L153

https://github.com/WikiWatershed/model-my-watershed/blob/5d80d579aadfdefd6265647be1d8a32f4925cf80/src/mmw/js/src/user/views.js#L344

jwalgran:
Do we have a relationship with the person `aufdenkampe+bigcz`? Can we ask them for reproduction steps?

The possible "pollution" of the profile data makes sense, but I am curious about why this did not show up in development.

rajadain:
[...]

PR #2478, opened 2017-11-03
Move callback function to options from attributes

rajadain:
## Overview

Previously a success callback function was added to a model's attributes, and was being sent to the server during save, causing server side errors.

Now we pass in that callback via options, so it is not added to the list of attributes sent to the server.

This callback is used for showing the profile form after logging in for the first time, but is also used for actions that depend on logging in, such as [showing the projects list](https://github.com/WikiWatershed/model-my-watershed/blob/tt/user-accounts-profile-extra-param/src/mmw/js/src/draw/views.js#L253-L255).

Connects #2460 

## Testing Instructions

 * Go to [:8000/](http://localhost:8000/) and make sure you are signed out
 * Click on "Open Projects" and log in. Ensure you are taken to the projects list page
 * Sign up as a new user. Login, and ensure you see the profile form after logging in
 * Ensure saving the profile does not cause any errors

### DIFF
--- a/src/mmw/js/src/app.js
+++ b/src/mmw/js/src/app.js
@@ -145,7 +145,9 @@ var App = new Marionette.Application({
         var loginModalView = new userViews.LoginModalView({
             model: new userModels.LoginFormModel({
                 showItsiButton: settings.get('itsi_enabled'),
-                successCallback:  function(loginResponse) {
+            },
+            {
+                onSuccess:  function(loginResponse) {
                     if (loginResponse.profile_was_skipped || loginResponse.profile_is_complete) {
                         if (onSuccess && _.isFunction(onSuccess)) {
                             onSuccess(loginResponse);
@@ -154,8 +156,8 @@ var App = new Marionette.Application({
                         loginModalView.$el.modal('hide');
                         loginModalView.$el.on('hidden.bs.modal', function() {
                             new userViews.UserProfileModalView({
-                                model: new userModels.UserProfileFormModel({
-                                    successCallback: onSuccess
+                                model: new userModels.UserProfileFormModel({}, {
+                                    onSuccess: onSuccess
                                 }),
                                 app: self
                             }).render();
@@ -169,8 +171,8 @@ var App = new Marionette.Application({
 
     showProfileModal: function (onSuccess) {
         new userViews.UserProfileModalView({
-            model: new userModels.UserProfileFormModel({
-                successCallback: onSuccess
+            model: new userModels.UserProfileFormModel({}, {
+                onSuccess: onSuccess
             }),
             app: this
         }).render();
--- a/src/mmw/js/src/user/models.js
+++ b/src/mmw/js/src/user/models.js
@@ -84,10 +84,17 @@ var LoginFormModel = ModalBaseModel.extend({
     defaults: {
         username: null,
         password: null,
-        successCallback: null
     },
 
     url: '/user/login',
+
+    initialize: function(attrs, opts) {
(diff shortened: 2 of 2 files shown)


jwalgran:
Thanks for working on cleaning up this "ragged edge" of my user profile work. This looks like the beginning of a good change, but I have a few concerns:

- The PR builder reported 7 failing JS unit tests.
- I followed your test instructions on the develop branch and I was not able to reproduce the error that appeared in Rollbar. Moving the callback to attributes is inarguably the right thing to do, but I was not able to prove that it addressed the exception that prompted us to open the connected issue.
- The majority of the exceptions reported in Rollbar failed because of the presence of `server_errors[]` which makes me suspect that we are dealing with a broader state/serialization issue.

screen shot 2017-11-06 at 9 11 08 am

rajadain:
Good points. I'm going to investigate the `server_errors[]` errors more and add a fix for them.

rajadain:
The fixup 8c9e66b is for making the tests pass.

rajadain:
I was able to recreate the `server_errors[]=Server communication error` POST body [as recorded here](https://rollbar.com/Azavea/MMW/items/243/occurrences/33040555316/):

image

Using the following steps:

 1. Run `debugserver`
 2. Register for a new account. Activate this account.
 3. Login to this new account. See the User Profile modal pop up.
 4. Stop `debugserver` and stop the MMW service with `vagrant ssh app -c 'sudo service mmw-app stop'`
 5. Try to submit the User Profile modal. Get a generic "Server communication error" message.
 6. Turn the app back on with `debugserver`
 7. Submit the User Profile modal successfully. See the `server_errors[]: Server communication error` POST entry in the Network tab of Developer Tools

But now, this does not cause a server side error. This is probably because as of 81db33f we use a Serializer instead of feeding the POST variables [directly in to a `UserProfile` constructor](https://github.com/WikiWatershed/model-my-watershed/commit/81db33f160e287282a467c48773fff6bfabd8989#diff-76f6d7d05a4b988f9b9af9034c53a081L112), where the original error was blowing up.

The reason this shows up as `server_errors[]` instead of `server_errors` is because that is how arrays are sent as form data. For example, see this article: https://www.designedbyaturtle.co.uk/2014/submit-an-array-with-a-html-form/.

Ideally we should be sending the data as JSON instead of as a form, but as this is not causing a server side error any more thanks to the recent addition on that end, I'm willing to let it go for now.

jwalgran:
Should this be targeted at the release branch?

rajadain:
The original issue did not have the Production milestone, but I think since it is bug fix for something discovered on staging by the client, we might as well. Thoughts @ajrobbins?

rajadain:
Well, I think because of the serializer introduction we won't see an error for `successCallback=undefined` either. This is probably the reason @arottersman was unable to recreate the Rollbar error even when the front-end was sending extraneous values here: https://github.com/WikiWatershed/model-my-watershed/issues/2460#issuecomment-341428081. So this card is more cleanup than a bug fix (in case that helps decide between `develop` and `release/1.21.0`).

ajrobbins:
Let's target the release branch, as this is cleanup on a new feature that will be introduced in this release. I'll add the milestone to the original issue.

rajadain:
Rebased on and retargeted to `release/1.21.0`

Since this touches some of the new code that you put in recently @jwalgran, would you mind testing this again?

jwalgran:
Sure. I will check it out now.

rajadain:
Thanks for taking a look!

Issue #2452, opened 2017-10-31
Add CORS Support To Geoprocessing API


arottersman:
Consider adding [`django-cors-headers`](https://github.com/ottoyiu/django-cors-headers), which is used by cicero and the climate change api.

PR #2479, opened 2017-11-03
Add CORS Support to Geoprocessing API

rajadain:
## Overview

Add [CORS support plugin](https://github.com/ottoyiu/django-cors-headers) and allow all hosts to access the API.
We may restrict this in the future.

Connects #2452

### Demo

image

## Testing Instructions

 * Check out this branch and reprovision `app`
 * Go to [:8000/](http://localhost:8000/) and open the Network panel of Developer Tools
 * Draw / select a shape and proceed to Analyze. Watch the network requests for analyze submissions.
 * Inspect any such request. The response headers should include

       Access-Control-Allow-Origin: *

### DIFF
--- a/src/mmw/mmw/settings/base.py
+++ b/src/mmw/mmw/settings/base.py
@@ -262,6 +262,7 @@ TEMPLATE_DIRS = (
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#middleware-classes
 MIDDLEWARE_CLASSES = (
     # Default Django middleware.
+    'corsheaders.middleware.CorsMiddleware',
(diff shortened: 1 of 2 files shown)


arottersman:
I think we should restrict this to only apply to the `/api` endpoints.

rajadain:
Added the URL regex in d501570. I don't think we need to add it to the `/mmw/` endpoints because they're not for public consumption anyway?

arottersman:
>  I don't think we need to add it to the /mmw/ endpoints because they're not for public consumption anyway?

That's right

rajadain:
Squashing and merging now. Thanks for taking a look!

Issue #2449, opened 2017-10-31
RWD /watershed endpoint should return valid analyze shape


mmcfarland:
Using the Geoprocessing API, one can generate a `/watershed` which returns as a geoJson FeatureCollection object (within the overall payload).  Copying the geoJson result into an analysis API request returns the error:

```
{
  "detail": "Area of interest must be valid GeoJSON"
}

```

The ideal workflow would be for API consumers to generate watersheds and immediately use that geometry in subsequent calls (such as how the application behaves).

Determine if either the validation on `/analyze` endpoints is too strict, or if `/watershed` should return a different format

arottersman:
How copiable does the result need to be? We currently don't support feature collections because of the way we [parse geometries from the payload](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/geoprocessing.py#L137). We could easily grab the geometry of the collection's first feature, and use that.

RWD, however, gives back a result of the form:
```
{
    "watershed": { <FeatureCollection> },
    "input_pt": { <FeatureCollection> }
}
```
So any api users will have to access (or copy) the contents of `result.watershed`. Would it be worth it to have RWD return a single feature collection, with one feature being the watershed and the other the input_pt, and then parsing for the first feature with type `Polygon`? Or, is it fine to just accept a feature collection and expect the api user to copy the contents of `"watershed"`?

/cc @mmcfarland @ajrobbins

mmcfarland:
Being able to programmatically access an attribute is fine (I was doing that manually in my testing), the issue is more that `result.watershed` is not valid input to any of the other geometry based calls. Is it necessary for the /watershed endpoint to return a feature collection and not a multipolygon?

arottersman:
I don't think we use the FeatureCollection properties for anything, but I suppose someone might?

```
"properties": {
            "Avgslp": 0.044611662626266,
            "BasinLen": 0.265826812744141,
            "RR": 0.012108777649701,
            "Area": 0.014311335,
            "AvgOLF": 0,
            "DrnDen": 0,
            "BR": 3.218837738037109,
            "Strord": 1,
            "StrLen": 0,
            "GRIDCODE": 1
          }
```

arottersman:
[...]

PR #1468, opened 2016-09-15
Fix TypeError in `aoi_resolution` function

kellyi:
This PR tries to fix an `TypeError: unsupported operand type(s) for +: 'int' and 'list'` error reported periodically by Rollbar here: https://rollbar.com/Azavea/MMW/items/50/

The error message made it look like the operation was failing on trying to add a list to the base `0` value of the reduce function, and investigating the arguments passed to each Rollbar error occurrence revealed that one thing the input shapes have in common is that they each include a 5d array. For instance, here are some occurrences where you can search for and find `[[[[[` on the page in the args:

https://rollbar.com/Azavea/MMW/items/50/occurrences/16806247416/
https://rollbar.com/Azavea/MMW/items/50/occurrences/16162795600/
https://rollbar.com/Azavea/MMW/items/50/occurrences/15190217299/
https://rollbar.com/Azavea/MMW/items/50/occurrences/13831642433/
https://rollbar.com/Azavea/MMW/items/50/occurrences/12823943776/
https://rollbar.com/Azavea/MMW/items/50/occurrences/12077389253/

The last one in that list is the earliest occurrence, from March 11, which seems to have been when we released the production version including RWD: 

https://github.com/WikiWatershed/model-my-watershed/releases/tag/1.11.0

I believe what's happening is that some of the shapes RWD creates include the 5d nested array, and that's what triggers this exception. The fix here checks whether `area_of_interest['coordinates'][0][0][0][0]` is a float or a list, and then sets the `pairs` value appropriately to use in calculating the average latitude.

**Testing**
- grab this branch, `vagrant up`, `./scripts/debugcelery.sh` then visit `localhost:8000`
- use the boundary selector, the draw tool, the sq km pointer, and RWD to create some areas of interest, and check that the analysis for each completes as it should and that you can run TR-55 and MapShed jobs on them without seeing the `TypeError: unsupported operand type(s) for +: 'int' and 'list'` error in Celery.

Connects #1225

PR #2487, opened 2017-11-06
Analyze RWD results More Directly

arottersman:
## Overview

The point of this PR is to allow Polygon, MultiPolygon and Feature objects (containing Polygon or MultiPolygon) as input to the API. We currently only accept MultiPolygon, which makes it more difficult than it has to be for API users to pipe RWD output to analyze. 

While working on this, I realized two things:
1. All our AOI parsing + validating would fit in well as a serializer (refactored in e04a882)
2. The mysterious `to_one_ring_multipolygon()` was correcting faulty input that our own frontend was creating (frontend fixed in 8d8200b, removed from backend in e04a882)

Connects #2449 

### Demo

```bash
[13:22:25] arottersman@silver /Users/arottersman/Development/model-my-watershed
> set TOKEN f19b20fe829f92ed176d8f90a1606531b445727c
[13:22:28] arottersman@silver /Users/arottersman/Development/model-my-watershed
> curl -i -H "Content-Type: application/json" \
                 -H "Authorization: Token $TOKEN" \
                 -X POST -d '{ "location": [39.67185812402583,-75.76742706298828] }' \
                 localhost:8000/api/watershed/
HTTP/1.1 200 OK
Server: nginx
Date: Mon, 06 Nov 2017 18:22:39 GMT
Transfer-Encoding: chunked
Connection: keep-alive
Vary: Accept-Encoding
Vary: Accept
Location: http://localhost:8000/api/jobs/329482c2-08a2-4755-bd81-784a00858ef4/
Allow: POST, OPTIONS

{"status":"started","job":"329482c2-08a2-4755-bd81-784a00858ef4"}⏎                                        [13:22:39] arottersman@silver /Users/arottersman/Development/model-my-watershed
> set WATERSHED (curl -H "Content-Type: application/json" \
                 -H "Authorization: Token $TOKEN" \
                 http://localhost:8000/api/jobs/cf1e86fd-a3c5-459e-8409-9f8c245174d4/ | jq ".result.watershed")
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   755    0   755    0     0   9806      0 --:--:-- --:--:-- --:--:--  9934

[13:22:48] arottersman@silver /Users/arottersman/Development/model-my-watershed
> echo $WATERSHED | jq "."
{
  "geometry": {
    "type": "Polygon",
    "coordinates": [
      [
        [
          -75.767778,
          39.672315
        ],
        [
          -75.767685,
          39.67213
        ],
        [
          -75.7675,
          39.67213
        ],
        [
          -75.767407,
          39.671852
        ],
        [
          -75.76787,
          39.672037
        ],
        [
          -75.767778,
          39.672315
        ]
      ]
    ]
  },
  "type": "Feature",
  "properties": {
    "Avgslp": 0.04092837497592,
    "BasinLen": 0.049257610321045,
    "RR": 0.025767648592591,
    "Area": 0.000889405,
    "AvgOLF": 0,
    "DrnDen": -36842.608260578694,
    "BR": 1.269252777099609,
    "Strord": -32768,
    "StrLen": -32.768,
    "GRIDCODE": 1
  }
}

[13:22:50] arottersman@silver /Users/arottersman/Development/model-my-watershed
> curl -i -H "Content-Type: application/json" \
                 -H "Authorization: Token $TOKEN" \
                 -X POST -d "$WATERSHED" \
                 localhost:8000/api/analyze/land/
HTTP/1.1 200 OK
Server: nginx
Date: Mon, 06 Nov 2017 18:23:00 GMT
Transfer-Encoding: chunked
Connection: keep-alive
Vary: Accept-Encoding
Vary: Accept
Location: http://localhost:8000/api/jobs/5c994457-88e2-4e93-9391-5ad62875c371/
Allow: POST, OPTIONS

{"status":"started","job":"5c994457-88e2-4e93-9391-5ad62875c371"}⏎                                        [13:23:00] arottersman@silver /Users/arottersman/Development/model-my-watershed
> curl -H "Content-Type: application/json" \
                 -H "Authorization: Token $TOKEN" \
                 http://localhost:8000/api/jobs/5c994457-88e2-4e93-9391-5ad62875c371/ | jq ".result"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1670    0  1670    0     0  27007      0 --:--:-- --:--:-- --:--:-- 27377
{
  "survey": {
    "displayName": "Land",
    "name": "land",
    "categories": [
      {
        "nlcd": 43,
        "code": "mixed_forest",
        "type": "Mixed Forest",
        "coverage": 0,
        "area": 0
      },
    ...
    ]
  }
}

```

## Testing Instructions

 * Pull and `bundle`
 * Provision the worker to get the new RWD version

#### Testing RWD Polygon and MultiPolygon
- Delineate a shape just about anywhere with the UI. Provided it's not a MultiPolygon, it should succeed, and be analyzable and model-able.

- Use the geocoder to go to first result of the search `Marlborough and Providence`. 
<details>
<summary>You should see a little lake with a tiny island </summary>

screen shot 2017-11-06 at 12 16 45 pm

</details>

DRB delineate a point on the northwestern tip of the island. This should result
in a multipolygon. You should be able to view the shape, but not analyze it.
  
screen shot 2017-11-06 at 12 17 17 pm

#### Testing API
Send the contents of `/api/watershed`'s `result.watershed` into the API analyze endpoints and confirm the jobs are successful.

Confirm garbage shapes are still caught.


mmcfarland:
I'm getting a lot of "intersects with itself" errors for DRB-RWD, which don't occur on production:

screenshot from 2017-11-07 09 57 33

Happens frequently on the Schukyll below Reading, PA.  This spot just above the bridge near Gibraltar is a good spot to see if behave differently:
screenshot from 2017-11-07 09 58 50

src/mmw/js/src/core/utils.js
```diff
@@ -480,14 +480,14 @@ var utils = {
         });
     },
 
-    // Convert polygon to MultiPolyon (mutates original argument).
-    toMultiPolygon: function(polygon) {
-        var geom = polygon.geometry ? polygon.geometry : polygon;
+    // Convert polygon to MultiPolyon
+    toMultiPolygon: function toMultiPolygon(polygon) {
+        var geom = _.clone(polygon.geometry ? polygon.geometry : polygon);
```
mmcfarland: Should this be a `deepClone` instead, since geometry is likely to be nested?

src/mmw/js/src/draw/views.js
```diff
@@ -77,7 +77,7 @@ function actOnLayer(datum) {
 function validateRwdShape(result) {
     var d = new $.Deferred();
     if (result.watershed) {
-        if (result.watershed.features[0].geometry.type === 'MultiPolygon') {
+        if (result.watershed.geometry.type === 'MultiPolygon') {
```
mmcfarland: This is still a little strange to me.  We can analyze MultiPolygons in other situations (including directly to the API, for example `wkaoi=huc12__70859`).  Is this check really a shorthand for some other kind of problem with RWD derived shapes?

src/mmw/js/src/draw/views.js
```diff
@@ -81,7 +81,10 @@ function validateRwdShape(result) {
             d.reject('Unfortunately, the watershed generated at this ' +
                      'location is not available for analysis');
         }
-        validateShape(result.watershed)
+        // Validate the shape as a multipolygon -- this is
+        // how it'll be displayed on the map, and sent to
+        // analyze
+        validateShape(coreUtils.toMultiPolygon(result.watershed))
```
mmcfarland: Is the result of this that RWD output (like the Gibraltar example I gave) can be analyzed via the app, but not the API? That's what I would expect, since there is this conversion which is neccessary.  However, I can take the results from the watershed API and plug that directly into the soils API and get a successful result.  Can you explain what I'm not getting?

```
{"location":[40.28895415740959,-75.86832046508789],"snappingOn":true,"dataSource":"drb"}
```

src/mmw/js/src/draw/views.js
mmcfarland: Another example, here is an RWD result which the app won't analyze, but the API will (when the RWD output is given to `land`):

```json
{"location":[39.978299876162005,-75.46028137207031],"snappingOn":true,"dataSource":"drb"}
```
Which comes from the lake-island in your example above.

arottersman: Good point. Going to remove this check. It was added in https://github.com/WikiWatershed/model-my-watershed/pull/1468 when we were trying to guard against the bug `utils.toMultiPolygon` was creating.

src/mmw/js/src/draw/views.js
arottersman: I can't explain it fully, but my hunch is the holes in the polygon are ignored once it's contained in a multipolygon. The same conversion happens on the backend here (https://github.com/WikiWatershed/model-my-watershed/pull/2487/files#diff-10de29d8b4c29bab0a1f77cc894dc45cR58) before validation.

I'll dig into this a little.

arottersman:
Yes, that would be pretty terrible if we messed up the project serialization again; I'll work through a set of saved project checks. I'm hopeful they're okay though. The loading/saving of projects should be untouched; their `area_of_interest` field goes through the app before it goes to the analyze/model views that use the new AoiSerializer.

src/mmw/js/src/draw/views.js
arottersman: What I think I've found is `turfKinks` and `shapely.is_valid` don't check a certain sort of validity deeply. When we pass a `Polygon` they check if the shape intersects, and if the rings intersect each other. When we pass a `MultiPolygon`, they check if any of the `Polygon`s intersect each other, or intersect themselves , but not if the `Polygon`s have intersecting rings.

So, in one sense, forcing these shapes into `MultiPolygon`s is just muffling a legitimate error. In another sense, the error isn't all that legitimate because the geoproccessing seems like it can handle shapes with these kinds of self-intersection, namely rings that intersect borders:

(The mask in the app hides these rings, but here's one thrown into geojson.io from the same gibraltar shape)
screen shot 2017-11-07 at 4 04 32 pm

It's maybe worth noting that the muffling was happening with RWD shapes before this PR. `result.watershed` was made a `MultiPolygon` before it got to `validateShape`.

If we don't want to keep the muffling, some other ideas I have for handling the ringed shapes are:
- Remove rings either in RWD or on the frontend
- During frontend validation, validate only the outer ring of a polygon
- During frontend validation, validate each ring of a polygon, but not the polygon as a whole 

Any thoughts on what would be cleanest?

arottersman:
We're going to keep this open past the release.

src/mmw/js/src/draw/views.js
mmcfarland: @arottersman Sorry, I didn't catch this message.  Yes, this is kind of what I expected, that the validation is too strong compared to what can actually be run through the processor.  Given your options to move forward, I'd have these recommendations:
* As a priority, try not to alter the shape of the watershed
* Same behavior between app and API
* Don't prevent shapes which can be analyzed from being so, even if that means some validations are ambiguous and we attempt to process polygons which can't be handled well
* In the above case, we catch and report the invalid input error as before, it'll presumably just take a bit longer to get to the user.

Do those guidelines help clarify an approach you'd prefer?

Issue #2482, opened 2017-11-06
Onboarding profile modal should not be scrollable when modal is fully onscreen


ajrobbins:
Observed on my macbook air 13 in: 
![scroll](https://user-images.githubusercontent.com/12868342/32448027-c638eed0-c2db-11e7-8b1b-09d5d028cbf3.gif)

Expected: modal is fixed in the center of the screen (since there appears to be room for it), with no scrolling unless the screen size is so small that scrolling is necessary.

ajrobbins:
Observed on Safari as well

rajadain:
I think it should be scrollable, to allow for when the screen is too short. See discussion on it here: https://github.com/WikiWatershed/model-my-watershed/pull/2462#issuecomment-341427131

jwalgran:
It does need to be scrollable when the modal does not fit onscreen, but there is still a situation where the modal is fully visible but still scrollable. It looks like this is caused by the 72px margin at the top and bottom of the dialog.

<img width="1107" alt="screen shot 2017-11-06 at 10 21 43 am" src="https://user-images.githubusercontent.com/17363/32454500-9f18bd16-c2dc-11e7-8202-c2e01f95d448.png">

I am not sure if there is a way to say "only make this scrollable if the div itself it too tall, not the margin." Any tips @jfrankl?

jfrankl:
If you remove the bottom margin does it address this issue? Ideally, the modal wouldn't change positions, and it wouldn't scroll when it fits on the page.

If that doesn't work, let me know and I'll spin up my VM and take a closer look.

jwalgran:
Of course. That does appear to do exactly what we want. Thanks.

PR #2489, opened 2017-11-06
Remove bottom margin from modals to prevent unneeded scrolling

jwalgran:
## Overview

Removing the bottom margin prevents the page from scrolling when a modal dialog is completely on screen but less than 72px from the bottom of the browser frame.

Connects #2482

### Demo

screen shot 2017-11-06 at 1 02 09 pm

## Testing Instructions

 * Open a modal
 * Adjust the height of the browser so that the entire modal is visible, but there is only a small space below the modal.
 * Confirm that the page does not scroll.
 * Adjust the height of the browser to that the modal is partially outside the viewport.
 * Confirm that the page is now scrollable.

### DIFF
--- a/src/mmw/sass/components/_modals.scss
+++ b/src/mmw/sass/components/_modals.scss
@@ -3,7 +3,7 @@
   background-color: $black-54;
 
   .modal-dialog {
-    margin: 72px auto;
+    margin: 72px auto 0 auto;
   }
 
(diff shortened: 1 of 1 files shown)


rajadain:
Can we merge this? The production deploy is in a few hours, and this is the last non-merged card tagged with this release. @jwalgran @caseycesari

Issue #2495, opened 2017-11-08
Already Fetched WDC Variables Are Blank


arottersman:
1. Search for something with WDC results, like "water" in the philly area
1. Open the details for a result with a date. Mark down its name, and let the variables load
1. Go back to the list and open the date filter. Choose a date that will include the result for which you just opened the details
1. Open the same result's details again and it will have no variables

![screen shot 2017-11-08 at 8 40 45 am](https://user-images.githubusercontent.com/7633670/32552397-cfc68fde-c461-11e7-89aa-1de84eaf45f4.png)

/cc @rajadain

rajadain:
This can be fixed by:

```diff
diff --git a/src/mmw/js/src/data_catalog/models.js b/src/mmw/js/src/data_catalog/models.js
index a065e7d1..3500d680 100644
--- a/src/mmw/js/src/data_catalog/models.js
+++ b/src/mmw/js/src/data_catalog/models.js
@@ -421,6 +421,7 @@ var Result = Backbone.Model.extend({
             if (variables instanceof CuahsiVariables) {
                 variables.reset(response.variables);
                 delete response.variables;
+                delete this.fetchPromise;
             }
         }
```

PR #2498, opened 2017-11-08
BiG-CZ: Refetch Data On Server Refresh

rajadain:
## Overview

When a filter changes, all results for a catalog are refetched. It is possible that a filter may include results that were already included. In this case, there isn't a new instance of the result, but the existing instance is refreshed. When this happens, its CuahsiVariables collection is reset, with no detail values.

Previously, if a result details page had been viewed, it had fetched the detail values for said CuahsiVariables, and cached them, using a fetchPromise that would only be resolved once. After fetching them for the first time, we simply returned the fetchPromise so they would not be fetched again.

When the values are reset, the actual detail values are gone, but the cached fetchPromise remains, thus preventing values from being fetched anew. By deleting the fetchPromise when receiving new values, we ensure that when a detail page is opened for a previously fetched and repeatedly filtered result, the values for its CuahsiVariables are refetched.

Connects #2495 

## Testing Instructions

 * Go to [:8000/?bigcz](http://localhost:8000/?bigcz), select a shape and proceed to search.
 * Search for "water" and switch to the CUAHSI tab. Pick a result and go to the details view. Wait for its values to be fetched.
 * Once the values are in, go back to list view and then go back to the same result's detail view. Ensure that it has cached values and they are not re-fetched.
 * Change the date filter to something that includes the result you picked. Wait for the results to refresh. Then open your result again. Ensure its data is re-fetched.
 * Change the date filter to exclude your result. Change it again to include it. Open its detail view. Ensure its data is refetched.

### DIFF
--- a/src/mmw/js/src/data_catalog/models.js
+++ b/src/mmw/js/src/data_catalog/models.js
@@ -421,6 +421,7 @@ var Result = Backbone.Model.extend({
             if (variables instanceof CuahsiVariables) {
                 variables.reset(response.variables);
                 delete response.variables;
+                delete this.fetchPromise;
             }
         }
 


arottersman:
Taking a look now

rajadain:
Thanks for spotting this and reviewing.

Issue #2330, opened 2017-10-05
Capture DRB user location information


ajrobbins:
Stroud would like to isolate and allow for pulling user data for users in the DRB, since that's a major area of funding and focus of MMW.

In https://github.com/WikiWatershed/model-my-watershed/issues/2255#issue-257734300, we'll be adding more fields to the user onboarding modal, including one for ZIP codes. 
~~We should:~~
~~- Generate list of ZIP codes that intersect the DRB and create a lookup table~~
~~- When a user creates a new account, check the inputed ZIP against said table~~
~~- If the ZIP is a match, stamp that user account record in the DB so it can be queried for later~~

Create a management command that pulls DRB user info out into a csv.

rajadain:
It may be more efficient to simply use a list of zip codes when querying the users to find out which ones are in the DRB, rather than adding this information in the database. If we can get a list of ZIP codes in the DRB, we can construct a query that gets the user details for those users.

Here is a shapefile of all ZIP codes in the United States: https://catalog.data.gov/dataset/tiger-line-shapefile-2015-2010-nation-u-s-2010-census-5-digit-zip-code-tabulation-area-zcta5-na

Here is a GeoJSON of the DRB used in the app: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/data/drb_perimeter.json

If a list of matching ZIPs is posted here, I will post the relevant SQL query.

ajrobbins:
List of matching zips:
[zip2.csv.txt](https://github.com/WikiWatershed/model-my-watershed/files/1455542/zip2.csv.txt)

Produced by intersecting the [TIGER/Line 2017 zip data](https://www.census.gov/cgi-bin/geo/shapefiles/index.php?year=2017&layergroup=ZIP+Code+Tabulation+Areas) with the [DRB shape](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/data/drb_perimeter.json).

Issue #2451, opened 2017-10-31
Adapt API logging mechanism to in-app users


mmcfarland:
Allow application generated requests that go through the API to be logged, tagged to the actual logged in user if available.

* If logged in, track which user generated the request, but also that it originated in MMW app
* If not logged in, track an anonymous MMW app request
* API requests are logged as-is

Question?
Modeling requests are not part of the API and are therefore not tracked.  Do we try to integrate those views to an adapted logging decorator?

PR #2501, opened 2017-11-08
Add command for getting list of DRB users

rajadain:
## Overview

Adds a management command to get a list of all DRB users. For this, we took the [shapefile of ZIP codes from TigerLines](https://catalog.data.gov/dataset/tiger-line-shapefile-2015-2010-nation-u-s-2010-census-5-digit-zip-code-tabulation-area-zcta5-na) and intersected it with [the DRB perimeter](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/mmw/settings/data/drb_perimeter.json) and got a [list of ZIP codes](https://github.com/WikiWatershed/model-my-watershed/files/1455542/zip2.csv.txt). See https://github.com/WikiWatershed/model-my-watershed/issues/2330#issuecomment-342974255

This list is then used to filter users that have listed such a ZIP code in their profile.

Connects #2330 

### Demo

```bash
> kj manage drbusers
+ ARGS=drbusers
+ [[ 1 -eq 1 ]]
+ [[ drbusers == runserver ]]
+ vagrant ssh app -c 'cd /opt/app && envdir /etc/mmw.d/env ./manage.py drbusers'
id,username,organization,user_type,country,postal_code
"1","test1","Azavea","University Professional or Research Staff","US","19103"
"8","test3","Azavea","Commercial/Professional","US","19103"
"12","test7","Azavea","Unspecified","US","19103"
```

### Notes

Currently we have very little validation on user entry (for example we do not validate that ZIP codes are numeric and five digit, although I guess this to support international entry), so the data can be a little noisy.

For production, someone will have to SSH in to the App VM and then run:

```bash
envdir /etc/mmw.d/env python manage.py drbusers > drbusers.csv
```

## Testing Instructions

 * Check out this branch and run `./scripts/manage.sh drbusers`
 * Ensure all the correct users show up, and any users with missing ZIP codes or those outside the DRB are not included


ajrobbins:
So the idea here then is that whenever Stroud wants the list, we would run this command and send them the output? Just trying to nail down the workflow.

mmcfarland:
The issue specifies that this mechanism should "track" user data.  That idea isn't defined at all, but I don't think it's accommodated with this either.  Can @ajrobbins explain the goal there and how we might accommodate it?  My suggestion would be that it tie into the proposed work we describe in https://github.com/WikiWatershed/model-my-watershed/issues/2451

ajrobbins:
Just chatted with @mmcfarland  - this card/PR should just be to pull the user data for users identified by their ZIPs to be in the DRB, so it accomplishes that goal.

rajadain:
> whenever Stroud wants the list, we would run this command and send them the output?

Yes.

I went the manage command route because we don't have an admin reporting interface for MMW. If something like that is developed in the future a similar query can be incorporated therein. In the interim, this should suffice.

The discussion above seems to be in agreement, so I'm going to merge this in.

Thanks for reviewing!

Issue #2165, opened 2017-08-18
Replace TR-55 Water Quality sideways bar chart with bullet chart


ajrobbins:
Our current sideways bar graph is ineffective because TSS is always very large relatively. 
![screen shot 2017-08-18 at 12 04 54 pm](https://user-images.githubusercontent.com/12868342/29467181-813ef5b6-840d-11e7-95d2-2a55425d8ddf.png)

Replace with bullet chart as per wireframes: https://marvelapp.com/8hi9jg9/screen/29823519

caseycesari:
We should consider using a graph type that represents parts-of-a-whole (the whole being "total loads"), illustrated here: http://www.datavizcatalogue.com/search/part_to_a_whole.html.

In the mock-ups, because of their alignment (even though the graphs are separate) and the different scales, a quick glance makes it look like phosphorus exceeds suspended solids. 

![image](https://user-images.githubusercontent.com/1042475/29467673-440b7334-840f-11e7-81eb-6bddca06e0ee.png)

ajrobbins:
@jfrankl thoughts?

jfrankl:
Hey @caseycesari, I see where you're coming from. A "part to the whole" graph does seem to be a natural fit for this. However, if you look at the sideways bar graph in Arianna's original post, one of the issues with this data is that the TSS is always so much higher than the other values that the graph ends up not being useful. Seems like all of the "part to the whole" graphs will run into this.

Multiple scales seems okay with the [bullet chart design specs](http://www.perceptualedge.com/articles/misc/Bullet_Graph_Design_Spec.pdf), so I don't think that part is necessarily a problem. Bullet charts are more difficult to read than some other types of charts, though.

What do you think? If there is a chart that you think would work well here and would address Arianna's original issue, I would love some help on this.

caseycesari:
Looking at it again, I think the existing chart is workable if we add labels to the bars, like so:

![29467181-813ef5b6-840d-11e7-95d2-2a55425d8ddf](https://user-images.githubusercontent.com/1042475/29676556-60b33be6-88c6-11e7-99ac-80478d5100ac.png)

ajrobbins:
@caseycesari I disagree - I think the existing chart is useful for comparing the three metrics to one another but that doesn't appear to be meaningful given the expected large differences among them.

caseycesari:
Talked with Jeff. I'm happy with whatever design we move forward with.

arottersman:
[...]

PR #2503, opened 2017-11-09
Switch TR-55 Water Quality to Bullet Charts

rajadain:
## Overview

Previously the water quality data was represented as a horizontal bar chart, with one bar for each type of observation. Unfortunately, the values of the three observations was quite disparate in scale, making the smallest bar insignificantly small.

To help bring each value to first class citizenship, we switch to using three bullet charts. Each chart is scaled to suit the value thereof, and thus shows real, actionable information.

Connects #2165 

### Demo

image

### Notes

The chart labels in the mock-up are left-aligned, but I couldn't figure out how to left-align the text easily here. Is the right alignment (default design for bullet charts) good enough?

The actual value isn't shown on the chart, but in the tooltip, which looks like this:

image

But enabling that tooltip also enables this one:

image

and I couldn't figure out a way to have one and not the other.

## Testing Instructions

 * Go to [:8000/](http://localhost:8000/) and create or open a TR-55 project
 * Go to the Water Quality tab. Ensure the new chart is there, and the values correspond to the table.

### DIFF
--- a/src/mmw/js/src/core/chart.js
+++ b/src/mmw/js/src/core/chart.js
@@ -3,7 +3,8 @@
 var d3 = require('d3'),
     nv = require('../../shim/nv.d3.js'),
     $ = require('jquery'),
-    _ = require('lodash');
+    _ = require('lodash'),
+    utils = require('./utils');
 
 var widthCutoff = 400;
 
@@ -599,26 +600,14 @@ function renderCompareMultibarChart(chartEl, name, label, colors, stacked, yMax,
 function renderBulletChart(chartEl, title, subtitle, data) {
     var chart = nv.models.bulletChart(),
         svg = makeSvg(chartEl),
+        range = utils.rangeInMagnitude(data),
         datum = {
             title: title,
             subtitle: subtitle,
+            ranges: [range.min, 0, range.max],
             measures: [data],
             color: '#48c5d1',
-        },
-        adjustment = 1;
-
-    // Calculate max range of chart
-    // by rounding up to the next place
-    while (data < 1) {
-        data *= 10;
-        adjustment -= 1;
-    }
-    while (data > 10) {
-        data /= 10;
-        adjustment += 1;
-    }
-
-    datum.ranges = [0, 0, Math.pow(10, adjustment)];
+        };
 
     chart.tooltip.valueFormatter(d3.format('.02f'));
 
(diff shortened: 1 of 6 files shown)


ajrobbins:
Looks good to me visually! How are we getting the max values? It is a little weird to have that tooltip, but doesn't bother me too much.

rajadain:
Adding a GIF of the tooltip to better communicate:

kellyi:
Taking a look at this!

kellyi:
This looks great!

screen shot 2017-11-09 at 5 20 40 pm

kellyi:
> How are we getting the max values?

Just to clarify: we are calculating the max value as the top of the current value's order of magnitude here?

https://github.com/WikiWatershed/model-my-watershed/pull/2503/files#diff-3e33aa79d87dd300905a2a0b6c714e56R611

That seems reasonable to me: it'd ensure the charts always show an appropriate ratio.

rajadain:
> we are calculating the max value as the top of the current value's order of magnitude 

Yes.

kellyi:
I'm in favor of leaving the chart title & subtitle left justified as is.

I was able to move them around manually in Chrome by adding an attribute like `dx="-3em"` to the `nv-title` and `nv-subtitle` classes, but I don't know whether nvd3 has an interface to do this automatically:

screen shot 2017-11-09 at 5 34 47 pm

screen shot 2017-11-09 at 5 35 49 pm

However, it'd be a chore to have to set those manually & verify correctness across browsers & screen sizes.

src/mmw/js/src/core/chart.js
```diff
@@ -596,9 +596,45 @@ function renderCompareMultibarChart(chartEl, name, label, colors, stacked, yMax,
...
+
+    // Calculate max range of chart
+    // by rounding up to the next place
+    while (data < 1) {
+        data *= 10;
+        adjustment -= 1;
+    }
+    while (data > 10) {
+        data /= 10;
+        adjustment += 1;
```
kellyi: `data` looks not to be used again after these adjustments (I checked in the REPL and the `measures: [data]` is passed by value.

Consider making a small utility function which would take a value of whatever magnitude and return its base 10 exponent & adding tests for < 1, > 10, and single digit values.

ajrobbins:
I'm fine with it.

jfrankl:
LGTM!

rajadain:
Rebased on latest `develop`, extracted the range in magnitude logic to a utility function and added some tests (and discovered some failing boundary conditions, so good idea @kellyi!). Will merge when green. Thanks for all the reviews!

PR #2508, opened 2017-11-15
Fix broken log configuration for geoprocessing service

hectcastro:
## Overview

The log routing configuration for the geoprocessing service via Beaver was in a broken state, which was breaking all of the Beaver managed log collection for the application as a whole. This change fixes the geoprocessing configuration, which in turn enables all centralized logging again (Nginx, application, tiler, Celery, Docker, etc.).

### Notes

N/A

## Testing Instructions

First, provision the `app`, `tiler`, and `worker` with the changes in this PR:

```bash
$ vagrant provision app tiler worker
```

Afterwards, ensure that the `services` virtual machine is up-and-running so that you can navigate to http://localhost:5601/. From there, filter messages by `type: geoprocessing` and ensure that there is at least one entry. Also, ensure that logs of other types are also making it into Kibana (e.g., Nginx).

### DIFF
--- a/deployment/ansible/roles/model-my-watershed.beaver/templates/geoprocessing.conf.j2
+++ b/deployment/ansible/roles/model-my-watershed.beaver/templates/geoprocessing.conf.j2
@@ -1,4 +1,3 @@
--[/var/log/upstart/geoprocessing.log]
--multiline_regex_before = ^\s+
--type: geoprocessing
--tags: geoprocessing,beaver
+[/var/log/upstart/geoprocessing.log]
+type: geoprocessing
+tags: geoprocessing,beaver


rajadain:
When I go to http://localhost:5601/, I'm greeted with this message:

image

Should I just go with the defaults or am I missing some configuration elsewhere?

hectcastro:
Ah, I forgot about that step. The options you have set look good. First, click `Create` then when it takes you to the following page, just click `Discover` at the top-left to see the logs.

rajadain:
I checked this out, reprovisioned `app` and `worker`, and recreated `tiler` (because of provisioning issues). Now, I see the following under `type` and `_type`:

image

I ran a TR-55 and MapShed job successfully, but searching for `type: geoprocessing` and `_type: geoprocessing` both result nothing:

image

image

If I search for just geoprocessing, I get these results:

image

Not sure if this is good enough, or if I'm doing something wrong.

hectcastro:
Yeah, this is good enough. The syntax for faceting on a field is something else. Just the presence of the other services on the left signals that it is working.

Issue #2334, opened 2017-10-05
Add streams analysis tab


ajrobbins:
Create a new Analysis tab called "streams", that displays summary statistics, by stream order and total, for the streams (NHD+) in  the AoI.

Here is a mockup of the tab's content: https://docs.google.com/document/d/1dEGFGqQLgvLmvV5NkzVszgqiJxYLrH5m1coEqg57q90/edit

Note that the "mean channel slope (%) statistic" represents length-weighted mean channel slope, and should be calculated as: sum(slope*length)/sum(length), where length is the segment attribute in NHD+.

arottersman:
@ajrobbins the link in the description is taking me to a card about the user modal, which doesn't have the mockup for the stream tab content. Any idea where the actual wireframes are?

ajrobbins:
Sorry about that @arottersman! Updated the issue description with the correct link.

mmcfarland:
The referenced google doc indicates that slope is an attribute of  `nhdflowline`, but I'm not sure if that's captured in our table.  We may need to regenerate that data.  Given that, it seem like the stream order summary and the channel slope should be on separate cards.  I suspect that adding the panel and the new job just for streams is sufficient for this pointed work and the slope work might be shifted to a new card.

kellyi:
I've been able to fit the tab by slightly adjusting the padding on the other tabs, but for the next tab we'll have to consider some way to page or toggle them:

<img width="517" alt="screen shot 2017-11-20 at 11 59 31 am" src="https://user-images.githubusercontent.com/4165523/33030864-78d74db6-cdea-11e7-8933-b29a0778b1b9.png">

kellyi:
The range of stream order values in the `nhdflowline` table is currently 0-10, plus some which have `null` values. (Total row count for that table is 2993513; 325622 have `null` stream_order values, which is about 11%.)

How should I handle the `null` stream order rows? Discard them? Sum them together and send them back like the other values?

mmcfarland:
Let's sum them and group them under "Other" while we determine what the null values actually indicate.

kellyi:
Got the tab set up and set up the backend to return (currently-not-properly-formatted) data:

![screen shot 2017-11-20 at 6 08 43 pm](https://user-images.githubusercontent.com/4165523/33046213-08fd3c88-ce1e-11e7-9029-9a6116c824ac.png)

TODO:
- [ ] get stream length data formatted properly for client
[...]

PR #2524, opened 2017-11-20
Add NHD Streams analysis task & tab

kellyi:
## Overview

This PR adds a Celery task & client-side tab for generating & displaying NHD Streams analysis results.

8dab1e3 sets up the Django-side Streams analysis Celery task and adds Swagger documentation to indicate the job's purpose & expected output.

6fb628d adds the Streams analysis tab to MMW, matching the outline linked in #2334.

Currently we're displaying the "Total Length (km)" values in the table -- and using 0 values for "Mean Channel Slope", "Length in Ag", and "Length in non-Ag". See "Notes" below for more on this.

Connects #2334

## Screenshot

screen shot 2017-11-22 at 11 33 34 am

## Notes

Slope, length in ag, and length in non-ag aren't available in `nhdflowline`, so we'll need to to update the table for slope and possibly calculate the ag-length values.

I used 0 values as placeholders for all these for now on the Celery side, then mostly configured the client to display whatever's sent back properly. We should handle this stuff in another issue.

I also adjusted the spacing between analyze tab headers to fit the new streams tab. The next tab will require making changes to how it works.

## Testing
- get this branch, bundle, then visit MMW and select and analyze an area of interest
- verify that you see the Streams tab and that it eventually returns results
- download the Streams data and verify that the CSV matches the table
- visit bigcz and verify that you don't see the streams tab

### DIFF
--- a/src/mmw/apps/geoprocessing_api/calcs.py
+++ b/src/mmw/apps/geoprocessing_api/calcs.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 from __future__ import absolute_import
 
 import json
+from operator import itemgetter
 
 from django.contrib.gis.geos import GEOSGeometry
 
@@ -47,6 +48,62 @@ def animal_population(geojson):
         'displayName': 'Animals',
         'name': 'animals',
         'categories': aeu_return_values
+    }
+
+
+def stream_data(geojson):
+    """
+    Given a GeoJSON shape, retreive stream data from the `nhdflowline` table
+    to display in the Analyze tab
+
+    Returns a dictionary to append to outgoing JSON for analysis results.
+    """
+
+    sql = '''
+        SELECT sum(lengthkm) as lengthkm, stream_order
+        FROM nhdflowline
+        WHERE ST_Intersects(geom, ST_SetSRID(ST_GeomFromGeoJSON(%s), 4326))
+        GROUP BY stream_order;
+        '''
+
+    with connection.cursor() as cursor:
+        cursor.execute(sql, [geojson])
+
+        if cursor.rowcount:
+            columns = [col[0] for col in cursor.description]
+            streams = [
+                dict(zip(columns,
+                         [
+                             float(row[0]) if row[0] else 0,
+                             int(row[1]) if row[1] and row[1] != 0 else 999,
+                         ]))
+                for row in cursor.fetchall()
+            ]
+        else:
+            streams = []
+
+    stream_data = {
+        str(s['stream_order']): {
+            "order": s['stream_order'],
+            "lengthkm": s['lengthkm'],
+            # TODO: implement `slopepct`
+            "slopepct": 0,
+        } for s in streams
+    }
+
+    # Add stream orders missing from query result
+    for x in list(range(1, 11)) + [999]:
+        stream_data.setdefault(str(x), {
+            "order": x,
+            "lengthkm": 0,
+            "slopepct": 0,
+        })
+
+    return {
+        'displayName': 'Streams',
+        'name': 'streams',
+        'categories': sorted(list(stream_data.values()),
+                             key=itemgetter('order')),
     }
 
 
--- a/src/mmw/apps/geoprocessing_api/views.py
+++ b/src/mmw/apps/geoprocessing_api/views.py
@@ -581,6 +581,141 @@ def start_analyze_soil(request, format=None):
 @decorators.permission_classes((IsAuthenticated, ))
 @decorators.throttle_classes([BurstRateThrottle, SustainedRateThrottle])
 @log_request
+def start_analyze_streams(request, format=None):
+    """
+    Starts a job to display streams & stream order within a given area of
+    interest.
+
+    For more information, see
+    the [technical documentation](https://wikiwatershedorg/documentation/
+    mmw-tech/#additional-data-layers)
+
+    ## Response
+
+    You can use the URL provided in the repsonse's `Location` header
+    to poll for the job's results.
+
+    <summary>
+      **Example of a completed job's `result`**
+    </summary>
+
+    <details>
+
+        {
+            "survey": {
+                "displayName": "Streams",
+                "name": "streams",
+                "categories": [
+                    {
+                        "order": 1,
+                        "lengthkm": 30.72,
+                        "slopepct": 0,
+                    },
+                    {
+                        "order": 2,
+                        "lengthkm": 61.2,
+                        "slopepct": 0,
+                    },
+                    {
+                        "order": 3,
+                        "lengthkm": 21.51,
+                        "slopepct": 0,
+                    },
+                    {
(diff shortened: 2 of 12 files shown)


src/mmw/apps/geoprocessing_api/calcs.py
```diff
@@ -50,6 +50,70 @@ def animal_population(geojson):
...
+            "stream_order": x,
+            "lengthkm": 0,
+            "slopepct": 0,
+            "ag_length": 0,
+            "non_ag_length": 0,
+        } for x in list(range(1, 11)) + [999]
+    }
+
+    for s in streams:
+        stream_data[str(s['stream_order'])]['lengthkm'] += s['lengthkm']
```
rajadain: Instead of doing this aggregation in Python, it'll be much faster to do it in the query itself:

```sql
SELECT SUM(lengthkm) AS total_lengthkm, stream_order
FROM nhdflowline
WHERE ST_Intersects(geom, ST_SetSRID(ST_GeomFromGeoJSON(%s), 4326))
GROUP BY stream_order;
```

src/mmw/apps/geoprocessing_api/calcs.py
```diff
@@ -50,6 +50,70 @@ def animal_population(geojson):
...
+            "slopepct": stream_data[k]['slopepct'],
+            "ag_length": stream_data[k]['ag_length'],
+            "non_ag_length": stream_data[k]['non_ag_length'],
+        } for k in stream_data
+    ]
+
+    return {
+        'displayName': 'Streams',
+        'name': 'streams',
+        'categories': sorted(stream_data, key=lambda e: e['order']),
```
rajadain: This is optional and a matter of taste, but I find `itemgetter` easier to read here:

```python
'categories': sorted(stream_data, key=itemgetter('order')),
```

See [here](https://github.com/WikiWatershed/model-my-watershed/blob/20ad7af7a407085b3848d5b474a4562cc86d5d3e/src/mmw/apps/bigcz/clients/cuahsi/search.py#L207) for an example.

src/mmw/js/src/analyze/views.js
```diff
@@ -638,6 +640,60 @@ var ClimateTableView = Marionette.CompositeView.extend({
...
+    childView: StreamTableRowView,
+    childViewContainer: 'tbody',
+    template: streamTableTmpl,
+
+    onAttach: function() {
+        $('[data-toggle="table"]').bootstrapTable();
+    },
+
+    templateHelpers: function() {
+        var data = this.collection.toJSON(),
```
rajadain: Since `lodash(data)` is used numerous times below, consider doing that once here so it is reused.

src/mmw/apps/geoprocessing_api/views.py
```diff
@@ -576,6 +576,163 @@ def start_analyze_soil(request, format=None):
...
+        - application/json
+    produces:
+        - application/json
+
+    """
+    user = request.user if request.user.is_authenticated() else None
+    area_of_interest, wkaoi = _parse_input(request)
+
+    return start_celery_job([
+        tasks.analyze_streams.s(area_of_interest)
```
rajadain: I think we can easily add [`nlcd_streams`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/mapshed/tasks.py#L161) here to calculate agricultural length of streams. I'll create an example.

rajadain: Oh wait, we need agricultural and non-agricultural length _per stream order_. Hmm. Okay let's leave that for a later card.

rajadain:
The tab headers on BiG-CZ, which still only sports three analyses, now look too close together:

image

Perhaps we should add a top-level `bigcz` class to the entire body when in that mode to be able to write special rules.

src/mmw/apps/geoprocessing_api/calcs.py
kellyi: Sounds good -- I'll make this update.

src/mmw/apps/geoprocessing_api/views.py
kellyi: My impression from the original google doc wireframe version was that it was total and not per stream order, but I'll make another card for hashing out those details.

kellyi:
Pushed fixups to update the trivial changes; going to push another to move the `groupby` into sql.

Not quite sure how to handle the fallout on the BigCZ tab spacing issue -- happy to add a separate BigCZ-specific class for now, but if we're likely to add more tabs in the future for MMW (and not BigCZ), it may make sense to make a MMW-specific one.

rajadain:
I think having a top-level class (maybe even on the `html` tag) of `mmw` or `bigcz` can be useful down the line. If you end up adding one, might as well add the other.

rajadain:
Alternatively, I wonder if we can use some Flexbox styling to distribute the items evenly given the container width.

src/mmw/apps/geoprocessing_api/calcs.py
kellyi: 7f07e18 adds this change; since the data returned from the sql query doesn't always include a full set of all stream orders, also added a loop to use `setdefault` to merge in 0 values for missing stream orders.

src/mmw/apps/geoprocessing_api/views.py
ajrobbins: My interpretation of the wireframe is the same as kelly's - ag and non-ag lengths are total, not per stream order.

ajrobbins:
Let's defer on tab spacing styling, as we'll need to handle that shortly with Jeff as soon as we add another tab and run out of tab room.

kellyi:
> Let's defer on tab spacing styling, as we'll need to handle that shortly with Jeff as soon as we add another tab and run out of tab room.

Sounds good. This is ready for another look; I'm going to make cards for the ensuing work.

rajadain:
Taking another look.

kellyi:
Occurred to me that I should drop the "length_in_ag" and "length_in_non_ag" keys from the API response since that sets them up to return per stream order, so I'll push a fixup to do that.

rajadain:
Okay I'll wait for that.

kellyi:
3d625cc drops length in ag/non-ag from the API
c206ab2 adjusts the panel to pass 0 values for length in ag/non-ag to the template

#2528 is to retrieve & display length in ag/non-ag data

rajadain:
Okay taking another look now.

src/mmw/apps/geoprocessing_api/views.py
rajadain: Okay. In that case I have an idea on how to do it, but will tackle it in #2528

rajadain:

kellyi:
Thanks for your help with this! I squashed this down to two commits then tested it out again locally. Everything still works, so I'll merge this once the tests pass.

Issue #2538, opened 2017-11-30
Scenario Share Dialog


rajadain:
In #2511 we changed the Project Share modal to be more comprehensive and allow integration with third parties such as HydroShare. This makes sense at the project level, but not so much at the scenario level. Currently Scenarios can also be shared, and are still using the old share dialog:

![image](https://user-images.githubusercontent.com/1430060/33438539-27f5a114-d5b9-11e7-8968-d98a56f13e3d.png)

Is this inconsistent? Should this be changed? How should this be done?

/cc @ajrobbins @jfrankl @mmcfarland

jfrankl:
It is inconsistent, but I think it's okay to leave it for now. By my estimation, it will take a significant amount of design time to resolve this.

Issue #2514, opened 2017-11-16
Implement Front-End for Connecting to HydroShare


rajadain:
Following #2513, implement the HydroShare connection front-end like this:

![image](https://user-images.githubusercontent.com/1430060/32909505-ad61a888-cad4-11e7-8224-f85f9e4dc041.png)

![image](https://user-images.githubusercontent.com/1430060/32909515-b17cf864-cad4-11e7-8ef3-a717a7ec8537.png)

![image](https://user-images.githubusercontent.com/1430060/32909521-b5ca425a-cad4-11e7-8bc5-f40aba0671fc.png)

This card is just for connecting a user's HydroShare account to their MMW account, not for the actual export.

- [x] When a user enabled the "Publish on HydroShare" switch, check if there is already a HydroShare account for this user
- [x] If there is already a HydroShare account for the user, do nothing, simply let them slide the switch on and off
- [x] If there is not an account for this user, show the informational modal similar to the one in the wireframe above
- [x] If the user clicks "Connect" in the information modal, handle the HydroShare login OAuth flow, keeping in mind that the user may create a new HydroShare account
- [x] Once the OAuth flow is complete, save this association to the database and return to the Share modal, now with "Publish on HydroShare" active

PR #2539, opened 2017-11-30
HydroShare Export Share Modal

rajadain:
## Overview

Changes the share modal of a project to match the new mockups. They HydroShare sharing part will come in a forthcoming card.

Connects #2511 

### Notes

The Scenario sharing uses the same modal as before. Should this also be updated? I'm creating a card for follow up discussion: #2538.

The HydroShare parts of the mockup haven't been implemented yet, they will come in a follow up card: #2514.

## Testing Instructions

* Go to [:8000/](http://localhost:8000/) and sign in.
* Create or open a TR-55 project. In the project menu, select "Share".
* Turn on link sharing for that project. Copy the link and open in a private browser window (where you're not signed in). Ensure the project is viewable.
* Turn off link sharing. Refresh the private browser window. Ensure you get a 404.
* Try turning link sharing on and off a bunch of times. Ensure the copy button works in all cases.

### DIFF
--- a/src/mmw/js/src/core/modals/templates/multiShareModal.html
+++ b/src/mmw/js/src/core/modals/templates/multiShareModal.html
@@ -12,6 +12,9 @@
                         <div class="toggle-switch"></div>
                     </label>
                 </div>
+                {% if not is_private %}
+                    <p>Anyone with this link will be able to view this project</p>
+                {% endif %}
                 <div class="input-group {{ 'hidden' if is_private }}">
                     <input class="form-control" id="share-url" type="text" value="{{  url }}">
                     <span class="input-group-btn">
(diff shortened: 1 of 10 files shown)


rajadain:
Also tagging @jfrankl for visual review.

jfrankl:
Looks good. One thing: I think the "Make private" button should go away. That's being replaced with this idea of a sharable link. If a shareable link is enabled, the project is effectively public. If it's not, it's effectively private.

rajadain:
There is some language in the "Make Private" / "Make Public" modals:

image

image

Should we replicate any of it in the new MultiShare modal?

caseycesari:
We may be able to condense it to something like "Anyone with the link will be able to view this project", and show that near the link sharing button.

jfrankl:
Something like this?

image

I like Casey's language above better than what's in this mockup. @ajrobbins curious what you think.

ajrobbins:
I like the new language better too!

rajadain:
Removed the Make Private / Make Public button, added language, added the new MultiShare modal to the Project Listing page as well.

image

rajadain:
Going to rebase and merge. Thanks for the reviews everyone!

Issue #2518, opened 2017-11-17
MapShed Fails on AoIs with No Streams


rajadain:
View details in Rollbar: [https://rollbar.com/Azavea/MMW/items/251/](https://rollbar.com/Azavea/MMW/items/251/)

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 374, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 629, in __protected_call__
    return self.run(*args, **kwargs)
  File "apps/modeling/mapshed/tasks.py", line 195, in nlcd_streams
    raise Exception('[nlcd_streams] {}'.format(result['error']))
Exception: [nlcd_streams] Geoprocessing Error.
Details: The request content was malformed:
Expected String as JsString, but got null
```

rajadain:
For example, run [`mmw-cli.py`](https://gist.github.com/rajadain/57d80a1fc5e1c03f87802967dcd57588) with:

```python
MAPSHED_INPUT = '{"area_of_interest":{"type":"MultiPolygon","coordinates":[[[[-95.89450638357287,39.60119298143448],[-95.88283736642711,39.60119298143448],[-95.88283736642711,39.61018337520713],[-95.89450638357287,39.61018337520713],[-95.89450638357287,39.60119298143448]]]]}}'
```

and see the following output:

```
> python mmw-cli.py
{"status":"started","job":"2652f02f-8006-4a9c-8c81-063fba6321d3"}
==> Submitted in 0.251928091049s
{"status":"failed","started":"2017-11-17T22:14:19.356Z","finished":"2017-11-17T22:14:20.903Z","result":"","error":"[nlcd_streams] Geoprocessing Error.\nDetails: The request content was malformed:\nExpected String as JsString, but got null","job_uuid":"2652f02f-8006-4a9c-8c81-063fba6321d3"}
==> Evaluated in 1.10281085968s
==> Total time 1.35473895073s
>
```

This is what the shape looks like:

![screen shot 2017-11-17 at 5 13 15 pm](https://user-images.githubusercontent.com/1430060/32971457-4acf8a18-cbbb-11e7-862d-43d08f20ec84.png)

PR #2546, opened 2017-12-04
Fix running MapShed on AoIs with no streams

rajadain:
## Overview

Previously, if no streams intersected the area of interest, instead of an array of stringified GeoJSON MultiLines, a singleton array would be sent with the only value being null. This caused parse issues in the geoprocessing service.

Now this case is caught on the Celery task itself, and if so we return an empty object immediately, and do not invoke the geoprocessing service at all. This short circuit makes the feedback faster and removes unnecessary communication.

Connects #2518

## Testing Instructions

* Check out this branch and restart Celery

* Go to [:8000/](http://localhost:8000/) and turn on the streams overlay
* Draw a shape deliberately avoiding streams
* Model with MapShed. Ensure that it works successfully.
* Download and run this file: [mmw-cli.py.txt](https://github.com/WikiWatershed/model-my-watershed/files/1528038/mmw-cli.py.txt), which has the test shape from the original issue.

### DIFF
--- a/src/mmw/apps/modeling/geoprocessing.py
+++ b/src/mmw/apps/modeling/geoprocessing.py
@@ -68,6 +68,13 @@ def run(self, opname, input_data, wkaoi=None, cache_key=''):
     data = settings.GEOP['json'][opname].copy()
     data['input'].update(input_data)
 
+    # If no vector data is supplied for vector operation, shortcut to empty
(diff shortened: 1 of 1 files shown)


arottersman:
Taking a look now.

rajadain:
Thanks for taking a look!

Issue #2528, opened 2017-11-27
Add ag-length & non-ag-length to Streams analysis


kellyi:
#2524 adds a Streams analysis task along with an analyze tab:

![screen shot 2017-11-27 at 11 52 45 am](https://user-images.githubusercontent.com/4165523/33278436-89ebb2b6-d369-11e7-85c1-14c15ba53918.png)

Currently the task returns data for "Total Length" per stream order which is then summed on the client side.

It doesn't however, return data for "Length in Ag" or "Length in non-Ag".

For this card: update the Geoprocessing API to return data for "Length in Ag" and "Length in non-Ag" along with the stream analysis results, and update the streams analyze tab to display that data.

PR #2547, opened 2017-12-04
Agricultural Streams Length in Analyze

rajadain:
## Overview

Adds length of streams in agricultural and non-agricultural areas in an area of interest to the Analyze tab. This reuses the `nlcd_streams` MapShed geoprocessing task, and also caches it, so if the user runs MapShed after this, the streams bit will be readily available. This does increase the time it takes for the streams tab to load though.

Connects #2528 

### Demo

image

## Testing Instructions

* Check out this branch and `bundle --debug` and restart Celery

* Go to [:8000/](http://localhost:8000/) and choose an urban area of interest.
* Go to the Analyze tab and ensure that the streams length distribution between agricultural and non-agricultural makes sense
* Try out a rural area of interest and ensure the values still seem appropriate

### DIFF
--- a/src/mmw/apps/geoprocessing_api/calcs.py
+++ b/src/mmw/apps/geoprocessing_api/calcs.py
@@ -51,7 +51,7 @@ def animal_population(geojson):
     }
 
 
-def stream_data(geojson):
+def stream_data(results, geojson):
     """
     Given a GeoJSON shape, retreive stream data from the `nhdflowline` table
     to display in the Analyze tab
@@ -86,6 +86,7 @@ def stream_data(geojson):
         str(s['stream_order']): {
             "order": s['stream_order'],
             "lengthkm": s['lengthkm'],
+            "ag_stream_pct": results['ag_stream_pct'],
             # TODO: implement `slopepct`
             "slopepct": 0,
         } for s in streams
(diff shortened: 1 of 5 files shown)


arottersman:
Taking a look

rajadain:
#2546 was just merged. Could you take another look at this when you get the chance?

src/mmw/apps/geoprocessing_api/calcs.py
```diff
@@ -86,6 +86,7 @@ def stream_data(geojson):
         str(s['stream_order']): {
             "order": s['stream_order'],
             "lengthkm": s['lengthkm'],
+            "ag_stream_pct": results['ag_stream_pct'],
```
kellyi: We should also add `ag_stream_pct` to the example completed job JSON in the Swagger docs: https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/geoprocessing_api/views.py#L605

arottersman: Yes we should! You are 💯

rajadain: Good catch, added in 6fdf4eb. Will merge shortly. Thanks for reviewing!

Issue #2532, opened 2017-11-28
Add slope pct data to Streams Analysis


kellyi:
Per https://github.com/WikiWatershed/model-my-watershed/issues/2334#issuecomment-345734852 slope is not currently available as an attribute on the `nhdflowline` table.

We implemented the streams analysis tab with 0 values for stream order slope percentages. The tasks for this card are to:

- [ ] get slope data into nhdflowline (or figure out how to calculate it somehow)
- [ ] adjust the streams analysis job to return slope percentage data per stream order
- [ ] display the slope percentage data per stream order in the client-side app

kellyi:
https://github.com/WikiWatershed/model-my-watershed/pull/2524 added the streams task & tab

arottersman:
It seems like we need another NHD table `ElevSlope` which has the `ComId` of the stream and `Slope`. 

@mmcfarland, have we ever received this table, and if not would stroud be able to send us a dump of it (if it is what we need)?

mmcfarland:
That looks like the right dataset to me, using the `slope` attribute.

![screenshot from 2017-12-01 12 38 18](https://user-images.githubusercontent.com/1014341/33495447-85641804-d694-11e7-83bd-a15832dd3667.png)

via [User Docs](https://s3.amazonaws.com/nhdplus/NHDPlusV21/Documentation/NHDPlusV2_User_Guide.pdf)

We have previously used the `NHDPlusAttributes` data sets to extract `stream_order` via `comid` and I still have all the attributes dbfs for all NHD Regions on an external drive, so we should be able to script the extraction.

@ajrobbins can you confirm with @aufdenkampe that this attribute is appropriate for the new stream analysis slope tab?  Note that the NHD dataset is slope in *meters* and the referenced issue indicates that a mean *percentage* is desired.  We'll need to see if a conversion is necessary (or possible) and if so if we should pre-process that per reach or simply load up meters and do adjustments based on AoI.

ajrobbins:
Just emailed Anthony the Q.

aufdenkampe:
[...]

PR #2551, opened 2017-12-06
Stream Slope

arottersman:
## Overview
- Modify the `nhdflowline` table to include a `slope` column
- Add `avgslope` to the stream info query, excluding null values
- Update UI to display slope as a percent, and No Data where not available

Connects #2532 

### Demo

screen shot 2017-12-06 at 2 39 12 pm

Somewhere Near Veil, CO
screen shot 2017-12-06 at 2 40 24 pm

## Testing Instructions

 * `vagrant ssh app -c 'cd /vagrant && ./scripts/aws/setupdb.sh -f nhdflowline_with_slope.sql.gz'` 
 * `vagrant ssh worker -c 'sudo service celeryd restart'`
 * `./scripts/bundle.sh`
 * http://localhost:8000/
 * Think of some places  of which you feel like you have a good sense of the elevation level changes, and analyze their streams. Eg, flat philly and the rocky mountains in Veil, CO

### DIFF
--- a/src/mmw/apps/geoprocessing_api/calcs.py
+++ b/src/mmw/apps/geoprocessing_api/calcs.py
@@ -59,12 +59,16 @@ def stream_data(results, geojson):
     Returns a dictionary to append to outgoing JSON for analysis results.
     """
 
+    NULL_SLOPE = -9998.0
+
     sql = '''
-        SELECT sum(lengthkm) as lengthkm, stream_order
+        SELECT sum(lengthkm) as lengthkm,
+               stream_order,
+               sum(lengthkm * NULLIF(slope, {NULL_SLOPE})) as slopesum
         FROM nhdflowline
         WHERE ST_Intersects(geom, ST_SetSRID(ST_GeomFromGeoJSON(%s), 4326))
         GROUP BY stream_order;
-        '''
+        '''.format(NULL_SLOPE=NULL_SLOPE)
 
     with connection.cursor() as cursor:
         cursor.execute(sql, [geojson])
@@ -76,19 +80,25 @@ def stream_data(results, geojson):
                          [
                              float(row[0]) if row[0] else 0,
                              int(row[1]) if row[1] and row[1] != 0 else 999,
+                             float(row[2]) if row[2] else None,
                          ]))
                 for row in cursor.fetchall()
             ]
         else:
             streams = []
 
+    def calculate_avg_slope(slope, length):
+        if slope and length:
+            return slope / length
+        return None
+
     stream_data = {
         str(s['stream_order']): {
             "order": s['stream_order'],
             "lengthkm": s['lengthkm'],
             "ag_stream_pct": results['ag_stream_pct'],
-            # TODO: implement `slopepct`
-            "slopepct": 0,
+            "total_weighted_slope": s['slopesum'],
+            "avgslope": calculate_avg_slope(s['slopesum'], s['lengthkm']),
         } for s in streams
     }
 
@@ -98,7 +108,8 @@ def stream_data(results, geojson):
             "order": x,
             "lengthkm": 0,
             "ag_stream_pct": results['ag_stream_pct'],
-            "slopepct": 0,
+            "avgslope": None,
+            "total_weighted_slope": None,
         })
 
     return {
--- a/src/mmw/js/src/analyze/templates/streamTable.html
+++ b/src/mmw/js/src/analyze/templates/streamTable.html
@@ -7,7 +7,7 @@
             <th class="text-right" data-field="lengthkm" data-sortable="true" data-sorter="window.numericSort">
                 Total Length (km)
             </th>
-            <th class="text-right" data-field="slopepct" data-sortable="true" data-sorter="window.numericSort">
+            <th class="text-right" data-field="avgslope" data-sortable="true" data-sorter="window.noDataSort" data-formatter="window.percentFormatter">
                 Mean Channel Slope (%)
             </th>
         </tr>
@@ -23,7 +23,7 @@
                 {{ totalLength|round(2)|toLocaleString(1) }}
             </th>
             <th class="text-right">
-                {{ avgChannelSlope|round(2)|toLocaleString(1) }}
+                {{ (avgChannelSlope * 100)|round(2)|toLocaleString(1) }}%
             </th>
         </tr>
     </tfoot>
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -664,7 +664,8 @@ var StreamTableRowView = Marionette.ItemView.extend({
         return {
             displayOrder: displayOrder,
             lengthkm: this.model.get('lengthkm'),
-            slopepct: this.model.get('slopepct')
+            avgslope: this.model.get('avgslope'),
+            noData: utils.noData,
         };
     }
 });
@@ -681,7 +682,7 @@ var StreamTableView = Marionette.CompositeView.extend({
     templateHelpers: function() {
         var data = lodash(this.collection.toJSON()),
             totalLength = data.pluck('lengthkm').sum(),
-            avgChannelSlope = data.pluck('slopepct').sum(),
+            avgChannelSlope = data.pluck('total_weighted_slope').sum() / totalLength,
             // Currently we only calculate agricultral percent for the entire
             // set, not per stream order, so we use the first value for total.
             agPercent = data.first().ag_stream_pct,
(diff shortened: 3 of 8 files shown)


src/mmw/js/src/analyze/views.js
```diff
@@ -681,7 +682,7 @@ var StreamTableView = Marionette.CompositeView.extend({
     templateHelpers: function() {
         var data = lodash(this.collection.toJSON()),
             totalLength = data.pluck('lengthkm').sum(),
-            avgChannelSlope = data.pluck('slopepct').sum(),
+            avgChannelSlope = data.pluck('avgslope').sum(),
```
kellyi: Is simple `sum` the right operation here? I was thinking it may make sense to show an average of the average slopes weighted by what percentage of total stream length a stream order represents.

In other words: if stream order 1 represents 90% of the total stream length within the AOI, its slope percentage would be weighted more heavily than another order representing < 10%. On the other hand, this does conflict a bit with the "Total" label on the summation row, since the lengths would be a total but the slope percentage would be an average.

arottersman: I was wondering about this, too. I left it because "Total" means "sum" to me. I think the length-weighted average sounds more meaningful though... @ajrobbins or @jfrankl any insight?

arottersman:
Will rebase this in a sec.

arottersman:
Rebased.

src/mmw/js/src/analyze/templates/streamTable.html
```diff
@@ -7,7 +7,7 @@
             <th class="text-right" data-field="lengthkm" data-sortable="true" data-sorter="window.numericSort">
                 Total Length (km)
             </th>
-            <th class="text-right" data-field="slopepct" data-sortable="true" data-sorter="window.numericSort">
+            <th class="text-right" data-field="avgslope" data-sortable="true" data-sorter="window.numericSort">
```
kellyi: Since there are potentially some `noData` values here, wonder if we should switch to use `window.noDataSort` for this column.

kellyi:
Got this set up and working:

screen shot 2017-12-06 at 6 19 56 pm

Code looks good. (Note the q above about the `noDataSort`.) I'll spend some more time testing this out tomorrow while we wait on a verdict on sum vs average.

src/mmw/js/src/analyze/templates/streamTable.html
arottersman: Good idea!

mmcfarland:
Super minor nitpick, but I don't prefer to have text in columns that otherwise have numbers.  What do you think about replacing `No Data` with `--`?  It is much less explicit, but I think the context of the length column being 0 would be enough context to infer the meaning that there is no slope data.  I think it makes it easier to scan, visually..

arottersman:
> What do you think about replacing No Data with --?

I'm with you on this preference, but we are using `No Data` in our other numerical analyze tables. I think if we change this instance of it, we should change all of them.

Pointsource table:
screen shot 2017-12-07 at 9 49 39 am

ajrobbins:
Mean channel slope should be sum(slope*length)/sum(length)
as per https://docs.google.com/document/d/1dEGFGqQLgvLmvV5NkzVszgqiJxYLrH5m1coEqg57q90/edit#

I think the total for that column should be a sum of the (%) values. @aufdenkampe is that correct?

kellyi:
If it's simple to do I'd also be in favor of globally replacing `noData` with `--`. However, we could handle that in another card.

I tested this out and it seems to be working well. I was wrong about directly using the `noData` sort, btw -- since there are percentages displayed in these columns, we'll need to add a new sort or adjust the existing sort to drop the `%` before converting to a number. Currently the sort's not deterministic & not ordered:

screen shot 2017-12-07 at 10 16 27 am

src/mmw/apps/geoprocessing_api/calcs.py
```diff
@@ -60,7 +60,9 @@ def stream_data(results, geojson):
     """
 
     sql = '''
-        SELECT sum(lengthkm) as lengthkm, stream_order
+        SELECT sum(lengthkm) as lengthkm,
+               stream_order,
+               avg(nullif(slope, -9998.0)) as avgslope
```
kellyi: Tagging @rajadain here to verify that this query matches the formula described here https://github.com/WikiWatershed/model-my-watershed/pull/2551#issuecomment-349994504

rajadain: Where does the `-9998.0` number come from? Can we make it a constant that is easier to read and conveys more context?

arottersman: It doesn't match. I'll be adding a fixup.

`-9998.0` comes from the original NHD table. I can set it to a constant so it's clear.

screen shot 2017-12-07 at 10 55 43 am

aufdenkampe:
Regarding the question from @ajrobbins:
> I think the total for that column should be a sum of the (%) values. @aufdenkampe is that correct?

No, that is not correct. It is more akin to the weighted average that is calculated just as it is for each row, except `sum(slope*length)/sum(length)` is evaluated for all the stream reaches in the AoI, not just the reaches of a certain stream order.

arottersman:
@kellyi, this should be ready for another look.

src/mmw/apps/geoprocessing_api/calcs.py
```diff
@@ -87,8 +97,8 @@ def stream_data(results, geojson):
             "order": s['stream_order'],
             "lengthkm": s['lengthkm'],
             "ag_stream_pct": results['ag_stream_pct'],
-            # TODO: implement `slopepct`
-            "slopepct": 0,
+            "total_weighted_slope": s['slopesum'],
+            "avgslope": s['slopesum'] / s['lengthkm'] if s['slopesum'] else None  # noqa,
```
kellyi: Could we push the else onto a second line directly beneath the `if` rather than disabling the linter?

arottersman: It's silly but I just could not get the linter to be happy and have it look good. When the `if` goes to the next line the linter seems to want it lined up with the `"avgslope"`. Maybe we should make this a legit for-loop instead of in-lining it.

kellyi: What if we moved this out into a function and then called `calculateAvgSlope(s)` here?

kellyi:
Sort is working well!

screen shot 2017-12-07 at 5 21 32 pm

The updated computations appear to be correct, but I'll ask a question about them in a note on the code.

src/mmw/apps/geoprocessing_api/calcs.py
```diff
@@ -59,12 +59,21 @@ def stream_data(results, geojson):
     Returns a dictionary to append to outgoing JSON for analysis results.
     """
 
+    NULL_SLOPE = -9998.0
+
     sql = '''
-        SELECT sum(lengthkm) as lengthkm, stream_order
+        SELECT sum(lengthkm) as lengthkm,
+               stream_order,
+               sum(case when NULLIF(slope, {NULL_SLOPE}) is NULL
```
rajadain: This could be simplified by doing something like:

```sql
SUM(lengthkm * NULLIF(slope, {NULL_SLOPE})) AS slopesum
```

Since arithmetic with NULLs results in NULLs.

src/mmw/apps/geoprocessing_api/calcs.py
```diff
@@ -59,12 +59,21 @@ def stream_data(results, geojson):
...
+                       then
+                           NULL
+                       else
+                           slope * lengthkm
```
kellyi: `total_weighted_slope` is now `slope * lengthkm` here. Then on https://github.com/WikiWatershed/model-my-watershed/pull/2551/files#diff-15b3e3581e8f864be02ed1013a02b466R685 we do:

```
data.pluck('total_weighted_slope').sum() / totalLength
```

That seems right to me but cc @rajadain for another double check.

Q: is `data.pluck` tossing out any null values it sees in before the `sum` op?

arottersman: - The grand total is `sum(slope_i * lengthkm_i) for i=0; i<all streams`. This is equivalent to breaking out the sum for each stream order, and then adding them together: 
```
  sum(slope_i * lengthkm_i) for i=0; i < order 1 streams +
  sum(slope_i * lengthkm_i) for i=0; i< order 2 streams +
  ...
  sum(slope_i * lengthkm_i) for i=0; i< order N streams +
```

- `data.pluck` doesn't toss out null values.
```js
> _.pluck([{ a: 0 }, { a: null }, { a: 123 }], 'a');
(3) [0, null, 123]
```

arottersman:
Thanks for the thorough review and the reminder! I'll run `setupdb` when this hits staging.

PR #2556, opened 2017-12-08
HydroShare Export Login Frontend

rajadain:
## Overview

Adds frontend to handle and facilitate connecting an MMW user to a HydroShare account. This does not actually export the project to HydroShare, but takes care of all the setup necessary for it.

The workflow for this setup is:

 * If a user has already connected to HydroShare, they can turn the export on and off at will
 * If they haven't, they'll be prompted to log in to HydroShare and allow MMW access
 * If the don't have a HydroShare account, they can make one, although the workflow there is confusing as after account creation they aren't led to the OAuth authorization page. They will have to close the popup and try again.
 * Currently the state of the sharing button isn't saved anywhere. That will be recorded in a forthcoming PR.

Connects #2514 

### Notes

All this UI interaction is new, and will probably need some tweaking before it is smooth. For example:

* Currently in case of error, the iframe is shown plainly, without any dismiss button. One has to click outside to dismiss it. Is this too confusing?

    image

## Testing Instructions

* Go to [:8000/](http://localhost:8000/) and create or open a TR-55 project.
* Click share and select HydroShare. Login with `HydroShare OAuth MMW Staging` credentials from LastPass. Authorize MMW. Ensure you see an error page.
* Reprovision `app` by initializing `MMW_HYDROSHARE_SECRET_KEY` environment variable using the client secret available in `HydroShare OAuth MMW Staging` on LastPass.

      (bash) $ MMW_HYDROSHARE_SECRET_KEY=*** vagrant reload app --provision
      (fish) > env MMW_HYDROSHARE_SECRET_KEY=*** vagrant reload app --provision

* Attempt to share to HydroShare again. Ensure it works this time.
* Try sharing a new project, or the same project again. You should no longer be prompted to login, but allowed directly.

### DIFF
--- a/src/mmw/apps/user/templates/user/hydroshare-auth.html
+++ b/src/mmw/apps/user/templates/user/hydroshare-auth.html
@@ -1,0 +1,47 @@
+{% include 'head.html' %}
+{% load staticfiles %}
+
+<body>
+    <div style="display:flex; flex-direction: column; height: 100%;
+                justify-content: center; align-items: center;">
+        {% if error %}
+            <h1>
+                <i class="fa fa-times-circle"
+                   style="font-size: 48px; margin-bottom: 24px; color: maroon"></i>
+            </h1>
+            <h1>Error</h1>
+            <p>{{ error }}</p>
+            <p style="margin: 20px;">
+            {% if error = "access_denied" %}
+                Model My Watershed needs to be authorized to post to your
+                HydroShare account in order to export projects.
+            {% else %}
+                There was an error while authorizing Model My Watershed to
+                access your HydroShare account.
+            {% endif %}
+            </p>
+            <a class="btn btn-md btn-block btn-active" href="/user/hydroshare/login/" target="_self">Try Again</a>
+            <button type="button" class="btn btn-md btn-block" id="hydroshare-cancel">Cancel</button>
+        {% else %}
+            <h1>
+                <i class="fa fa-check-circle"
+                   style="font-size: 48px; margin-bottom: 24px; color: teal"></i>
+            </h1>
+            <h1>Success!</h1>
+        {% endif %}
+    </div>
+    <script>
+        document.addEventListener('DOMContentLoaded', function() {
+            {% if error %}
+            var message = 'mmw-hydroshare-failure';
+            {% else %}
+            var message = 'mmw-hydroshare-success';
+            {% endif %}
+            parent.postMessage(message, window.location.origin);
+        });
+
+        document.getElementById('hydroshare-cancel').addEventListener('click', function() {
+            parent.postMessage('mmw-hydroshare-cancel', window.location.origin);
+        });
+    </script>
+</body>
--- a/src/mmw/apps/user/views.py
+++ b/src/mmw/apps/user/views.py
@@ -6,7 +6,7 @@ from django.contrib.auth import (authenticate,
                                  login as auth_login)
 from django.contrib.auth.models import User
 from django.core.urlresolvers import reverse
-from django.shortcuts import redirect
+from django.shortcuts import redirect, render_to_response
 from django.contrib.auth.forms import (PasswordResetForm,
                                        PasswordChangeForm)
 from django.core.exceptions import ObjectDoesNotExist
@@ -22,12 +22,11 @@ from rest_framework.response import Response
 from rest_framework.permissions import (AllowAny,
                                         IsAuthenticated)
 
+from apps.home.views import get_context
 from apps.user.models import ItsiUser, UserProfile, HydroShareToken
 from apps.user.itsi import ItsiService
 from apps.user.hydroshare import HydroShareService
-from apps.user.serializers import (UserProfileSerializer,
-                                   HydroShareTokenSerializer,
-                                   )
+from apps.user.serializers import UserProfileSerializer
 
 EMBED_FLAG = settings.ITSI['embed_flag']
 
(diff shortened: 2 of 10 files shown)


rajadain:
Also tagging @jfrankl, @ajrobbins, @mmcfarland for visual review.

jwalgran:
>Currently in case of error, the iframe is shown plainly, without any dismiss button. One has to click outside to dismiss it. Is this too confusing?

I think so and I expect some users will hit the back button or reload the page after they click the dialog and nothing happens. Users also end up in this state if they click "Cancel" on the authorization page.

If adding a button to this content is an involved process, I think it is justifiable to create a separate issue.

src/mmw/js/src/core/modals/views.js
```diff
@@ -252,6 +254,58 @@ var MultiShareView = ModalBaseView.extend({
...
+                        signalFailure: 'mmw-hydroshare-failure',
+                    })
+                });
+
+            checkbox.prop('checked', false);
+            confirm.render();
+
+            confirm.on('confirmation', iframe.render);
+            iframe.on('success', function() {
+                // Fetch user again to save new HydroShare Access state
```
jwalgran: The word "save" here in this comment is a little confusing. Does calling `fetch()` persist something?

src/mmw/js/src/core/modals/views.js
```diff
@@ -252,6 +254,58 @@ var MultiShareView = ModalBaseView.extend({
...
+                checkbox = self.ui.hydroShareEnabled,
+                confirm = new ConfirmLargeView({
+                        model: new models.ConfirmModel({
+                            titleText: 'Publish on HydroShare',
+                            question: [
+                                'You will need to connect your HydroShare ' +
+                                'account to your Model My Watershed account, ' +
+                                'or create a new HydroShare account.',
+
+                                'Lorem ipsum dolor sit amet, consectetur adipiscing elit. ' +
```
jwalgran: Do we know that we will be adding more text to this dialog and the lipsum is acting as a "TODO?"

rajadain:
> I think so and I expect some users will hit the back button or reload the page after they click the dialog and nothing happens. Users also end up in this state if they click "Cancel" on the authorization page.

Ah that's a good point. Very well, I'll add a cancel, and maybe a retry button as well.

src/mmw/apps/user/views.py
```diff
@@ -44,6 +44,8 @@ def make_successful_response_data(user):
             'result': 'success',
             'username': user.username,
             'itsi': ItsiUser.objects.filter(user_id=user.id).exists(),
+            'hydroshare': HydroShareToken.objects
```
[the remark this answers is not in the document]
rajadain: True, although `itsi` just above works the same way (is `true` or `false` for a given user). I'll see if I can think of something terse and descriptive.

src/mmw/js/src/core/modals/views.js
rajadain: When the OAuth dance succeeds, there is a new entry in the `user_hydrosharetoken` table for this user. When we call `fetch` again, the value of `user.hydroshare` is now `true`. This is the value that will be checked the next time the slider is slid.

I'm going to reword this and remove `save`.

src/mmw/js/src/core/modals/views.js
rajadain: It _is_ a bit of a TODO, and matches the wireframes, but also if we don't have a second paragraph in this modal, the one underneath peeks through, and I thinking hiding it completely is cleaner:

image

Hopefully we can come up with some copy that does that. /cc @jfrankl

src/mmw/apps/user/templates/user/hydroshare-auth.html
```diff
@@ -0,0 +1,32 @@
...
+                <i class="fa fa-check-circle"
+                   style="font-size: 48px; margin-bottom: 24px; color: teal"></i>
+            </h1></div>
+            <div><h1>Success!</h1></div>
+        {% endif %}
+    </div>
+    <script>
+        document.addEventListener('DOMContentLoaded', function() {
+            {% if error %}
+            var message = 'mmw-hydroshare-error';
```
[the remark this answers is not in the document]
rajadain: Good catch, this should be `mmw-hydroshare-failure` not `-error`. They are used here: https://github.com/WikiWatershed/model-my-watershed/blob/tt/hydroshare-export-login-frontend/src/mmw/js/src/core/modals/views.js#L292-L293. I should probably put them in constants to guard against such a mismatch.

The reason I went with strings instead of a promise is because there are three states: success, failure (during authorization), and dismissal (user canceling before authorization can complete). I hadn't thought of the case when the user cancels the authorization. I'll see if this can be refactored in to a promise ...

rajadain:
> Is it expected that the state of the "HydroShare Export" switch does not persist if you close and reopen the share dialog?

Yes. Likely in #2557. I already have code that does that, it's just not part of this PR because we're still working out the exact interaction sequence.

rajadain:
Now when they hit cancel in the HydroShare Authorize stage, they see this message:

image

"Try Again" takes them back to the Authorize stage, and cancel dismisses the modal.

If there's some other error (like misconfigured secret key, etc) they see this:

image

rajadain:
There are no buttons in the "Success" case, as the modal closes itself in half a second.

src/mmw/js/src/core/modals/views.js
jfrankl: For this Publish on Hydroshare modal, can we treat this as a change in the previous modal as opposed to a modal that sits on top of the previous modal? Visually, I think that will make more sense.

jwalgran:
Is this ready for one more review?

rajadain:
Let me address the last couple of items. Will be ready shortly.

src/mmw/apps/user/views.py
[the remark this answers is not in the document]
rajadain: Given that `itsi` above works the same way as this, I would prefer to keep this as `hydroshare` rather than something longer.

rajadain:
Removed the confirmation modal, now the workflow is:

1. Before authorizing:

    image

2. Switch takes directly to HydroShare instead of the confirmation modal:

    image

3. After the authorization is complete, the message goes away:

    image

rajadain:
Fixups are rather spread out for logistical purposes. Their functional grouping is:

* 1e6edd0a is fixing the `mmw-hydroshare-failure` label
* fa911405, b7453e25, 5a2d7d03, b5d8a23d, a7dedbd9 handle the "user canceled authorization" use case
* ec0538a8 improves a comment
* 05ac510e, a9b63a47 remove the confirmation modal and incorporate the messaging in to the multi-share modal

rajadain:
This is now ready for another review. Thanks for your patience!

jwalgran:
Looking this over now

rajadain:
Thanks for reviewing! Going to squash and merge now.

Issue #2560, opened 2017-12-14
Investigate slow streams tab


kellyi:
The streams tab is currently slow; we should investigate why & make a fix.

arottersman:
Query on `nhdflowline` was taking almost 9 sec:
![screen shot 2017-12-19 at 2 23 19 pm](https://user-images.githubusercontent.com/7633670/34174803-39551f5c-e4c8-11e7-963f-40805b7ca27d.png)

Add a spatial query on the geom, and now takes 242.3ms

![screen shot 2017-12-19 at 2 23 06 pm](https://user-images.githubusercontent.com/7633670/34174804-396226e8-e4c8-11e7-8b44-865b4dc468c3.png)


```sql
CREATE INDEX  nhdflowline_gix ON nhdflowline USING GIST (geom);
```

Going to upload table with index to s3.

PR #2562, opened 2017-12-19
Add spatial index to nhdflowline

arottersman:
## Overview

The stream analyze query took about 9sec because
we were scanning all the rows to find the intersecting flowlines.

Add a spatial index to the nhdflowline table so
we don't sequentially scan. Now takes ~230ms

- Streams table already had to be reloaded for production
- We'll need to load the newly dumped table on staging for this to take effect

Connects #2560 

### Demo

screen shot 2017-12-19 at 2 23 19 pm

## Testing Instructions

`vagrant ssh app -c 'cd /vagrant && ./scripts/aws/setupdb.sh -s`
Run some analyze jobs and confirm the streams tab completes within a few seconds

### DIFF
--- a/scripts/aws/setupdb.sh
+++ b/scripts/aws/setupdb.sh
@@ -114,7 +114,7 @@ fi
 
 if [ "$load_stream" = "true" ] ; then
     # Fetch stream network layer sql files
-    FILES=("nhdflowline_with_slope.sql.gz")
+    FILES=("nhdflowline_1513899196.sql.gz")
     PATHS=("nhd_streams_v2")
 
     download_and_load $FILES


mmcfarland:
I was surprised to see this because we did a fair bit of optimizing on the table when we did the stream tile rendering earlier in the contract, which included ensuring there was a spatial index on geom.  On my dev, if does exist (`nhdflowline_wkb_geometry_geom_idx`):

```
mmw=# \d nhdflowline
              Table "public.nhdflowline"
    Column    |            Type            | Modifiers 
--------------+----------------------------+-----------
 ogc_fid      | integer                    | not null
 geom         | geometry(LineStringZ,4326) | 
 comid        | numeric(9,0)               | 
 fdate        | date                       | 
 gnis_id      | character varying(10)      | 
 gnis_name    | character varying(65)      | 
 lengthkm     | numeric(19,11)             | 
 reachcode    | character varying(14)      | 
 flowdir      | character varying(15)      | 
 ftype        | character varying(24)      | 
 fcode        | numeric(9,0)               | 
 shape_leng   | numeric(19,11)             | 
 gnis_nbr     | numeric(9,0)               | 
 stream_order | numeric(2,0)               | 
Indexes:
    "nhdflowline_pkey" PRIMARY KEY, btree (ogc_fid)
    "nhdflowline_stream_order_idx" btree (stream_order)
    "nhdflowline_wkb_geometry_geom_idx" gist (geom)

```

I don't doubt that a subsequent dump/load could have inadvertently removed it, but could you confirm that the index is not present on stg/prd database?

arottersman:
> I don't doubt that a subsequent dump/load could have inadvertently removed it, but could you confirm that the index is not present on stg/prd database?

Yes, I'll check this.

kellyi:
Setting this up to review.

kellyi:
Encountering this error when trying to run `./scripts/setupdb.sh -s`:

```
+ curl -s https://s3.amazonaws.com/data.mmw.azavea.com/nhdflowline12192017.sql.gz
gzip: stdin: not in gzip format
```

I dropped the old `nhdflowline` table locally and it appears that encountering the gz error prevents it from loading the new one.

arottersman:
Thanks for the good recall and catching this, @mmcfarland. Staging is missing the indices, but production is not. I am going to:
1. import the backed up `nhdflowline.sql.gz`
2. add the slope column onto it
3. `pg_dump` after confirming all indices still exist on the table
1. delete the bad dumps from s3 and upload the new one

Motivated by my services VM running out space yesterday, I'm also going to take another look at the `stream_geom` column on the `nhdpluscatchment` table. We kept it because I determined it was too slow to do the catchment-flowline intersection calculation on the fly, but I must have been timing it without any indices. I'm going to time it again and assess whether the duplication of geometries is necessary.

arottersman:
1. This is ready for another look

2. re: `nhdpluscatchment` <> `nhdflowline`

With the index, joins are now fast enough that we could find a catchment's streams on the fly:
```sql
select st_union(geom), nhdpluscatchment.comid from nhdflowline, nhdpluscatchment where ST_Contains(geom_catch, geom) and nhdpluscatchment.comid in (select comid from nhdpluscatchment limit 100) group by nhdpluscatchment.comid;
# Time: 166.029 ms
```

Even though it's now feasible, I'm not going to drop the column because
- It's not clear what the `geom_stream` is (it's not a straight `ST_Contains` or `ST_Intersects`). I created #2565 to look into this. 
- It's not worth it. The duplicate streams take up only 7% of the table's total space, about 1GB. 
```sql
select
sum(pg_column_size(geom_stream)) as total_size,
sum(pg_column_size(geom_stream)) * 100.0 / pg_relation_size('nhdpluscatchment') as percentage
from nhdpluscatchment;

total_size |     average_size     |     percentage
------------+----------------------+--------------------
 1006660286 | 476.4561124342581058 | 6.9750485137192310
(1 row)
```

rajadain:
Taking a look now.

rajadain:
For the above, I measured the time between selecting a shape and for all the analyze tabs to finish.

arottersman:
Thanks for benchmarking those. Squashed the fixup and will merge when the build completes. I'll run this on staging tomorrow AM.

Issue #2561, opened 2017-12-19
Aggregate Precipitation Data


rajadain:
Currently, when querying temperature data for multiple weather stations, the results are averaged:

https://github.com/WikiWatershed/model-my-watershed/blob/586efe5c8454af386a5ba151b7ba876598d2bebd/src/mmw/apps/modeling/mapshed/calcs.py#L392-L411

However, the same is not done for precipitation, leading in non-averaged, non-aggregated data:

https://github.com/WikiWatershed/model-my-watershed/blob/586efe5c8454af386a5ba151b7ba876598d2bebd/src/mmw/apps/modeling/mapshed/calcs.py#L412-L423

It is unclear why this difference exists, as both were added in the same commit https://github.com/WikiWatershed/model-my-watershed/commit/fc1a41c3171072110bd0fbdada13a1ab1cb277c3 in the same PR #1241. But on the surface it looks wrong.

Add the same averaging aggregation to precipitation as exists for temperature.

rajadain:
Note: the previous version of `ms_weather_station.sql.gz` (now available at [`ms_weather_station_backup.sql.gz`](https://s3.amazonaws.com/data.mmw.azavea.com/ms_weather_station_backup.sql.gz)) had two rows for each weather station. Since we match for the [top 2](https://github.com/WikiWatershed/model-my-watershed/blob/bf593766a4300610a14cdbe65c746e919d4d3c98/src/mmw/mmw/settings/gwlfe_settings.py#L463) [closest stations](https://github.com/WikiWatershed/model-my-watershed/blob/586efe5c8454af386a5ba151b7ba876598d2bebd/src/mmw/apps/modeling/mapshed/calcs.py#L56-L61). Thus, for precipitation, the two resulting rows would have had identical values, thus discarding one had no real effect. However, as of #2549, we now fetch two actually different weather stations, and do need to average the values, else lose data.

rajadain:
Blocked until @ajrobbins confirms with subject matter experts.

ajrobbins:
Addressed in #2737

Issue #2549, opened 2017-12-05
Update values for ET variable in Mapshed


ajrobbins:
From BME:
> we are currently using an attribute field (Etadj) in our national weather station layer to determine the value for the "PctET(I)" variable that gets written to the last position of lines 8-19 in the gms file (this is the monthly ET adjustment factor for each station). After doing some initial model calibration checks, I have seen that the values we have been using in this field are not quite right for stations in the watersheds that I have been checking around the country. Consequently, I have revised these values, and have attached an updated version of the dbf file associated with our weather station shapefile. Would it be possible to replace the older values with these newer ones sometime in the near future?

[remove .txt] 
[weathstaalb.dbf.txt](https://github.com/WikiWatershed/model-my-watershed/files/1532035/weathstaalb.dbf.txt)

ajrobbins:
Additional task:

>When you replace the new ET values as described below, would it also be possible to make a small change in the equation used to calculate the value for the "SedNitr" factor in the model input (gms) file that is based on the "soil nitrogen" map used in MMW? Specifically, could we change the multiplication factor from "4.0" to "7.0" as shown in the attached document that has been updated from the previous version?

https://drive.google.com/open?id=0B0hs2mX4ApaLYnNkdThNZXppWUN2dE1USUp2X1hqZG9sekM0

ajrobbins:
shares points with #2559

rajadain:
The DBF file above contains five columns for ET Adjustment: `ETADJ`, `ET_ADJ2`, `ET_ADJ3`, `ETADJ_ORIG`, and `ETADJ_NEW`. Of these, `ET_ADJ3` mirrors the current values, as shown here:

<table>
  <tr>
    <th rowspan="2">STATION</th>
    <th rowspan="2">LOCATION</th>
    <th>Current</th>
    <th colspan="5">New</th>
  </tr>
  <tr>
    <th>etadj</th>
    <th>ETADJ</th>
    <th>ET_ADJ2</th>
    <th>ET_ADJ3</th>
    <th>ETADJ_ORIG</th>
    <th>ETADJ_NEW</th>
  </tr>
  <tr>
    <td>3103</td>
    <td>Flagstaff</td>
    <td>1.3</td>
    <td>1.3</td>
    <td>1.48</td>
    <td>1.3</td>
    <td>1.85</td>
    <td>1.3</td>
  </tr>
  <tr>
    <td>3812</td>
    <td>Asheville</td>
[...]

PR #2563, opened 2017-12-19
MapShed: Update ETAdj, SedNitr Calculations

rajadain:
## Overview

A revision from BME, we now multiply with 7.0 instead of 4.0 for more accurate results. More context can be found here: https://github.com/WikiWatershed/model-my-watershed/issues/2549#issuecomment-350090295

Additionally, update the `ms_weather_station.sql.gz` file on S3 with new values for etadj. The old table is still available at [`ms_weather_station_backup.sql.gz`](https://s3.amazonaws.com/data.mmw.azavea.com/ms_weather_station_backup.sql.gz).

Connects #2549 

### Demo

Old vs New GMS file differences for Little Neshaminy Creek:

image

Sample new GMS file: [LittleNeshaminy_Local.gms.txt](https://github.com/WikiWatershed/model-my-watershed/files/1573433/LittleNeshaminy_Local.gms.txt)

### Notes

Previously we mistakenly had 2 of each entry in the `ms_weather_station` table. When we would select the top 2 matching weather stations for any area of interest, we would always get 2 identical stations. I removed the extraneous duplication in the updated script, and now we select 2 actually different stations. This causes new values for weather data.

These values are more accurate than before, but we'll need to do #2561 to make them truly accurate.

## Testing Instructions

* Check out this branch
* Fetch new weather stations table:

      $ vagrant ssh app -c 'cd /vagrant && ./scripts/aws/setupdb.sh -f ms_weather_station.sql.gz'

* Clear your geoprocessing cache:

      $ vagrant ssh services -c 'redis-cli -n 1 --raw KEYS ":1:geop*" | xargs redis-cli -n 1 DEL'

* Run MapShed on a familiar shape and generate a GMS file
* Run MapShed on the same shape on Staging or Production and generate a GMS file
* Compare the two files. You should notice differences in the PrcntET(i) values in lines 8-19 position 8, and SedNitr in line 37 position 1. There will also be a large number of incidental weather data changes, as explained in **Notes** above.

### DIFF
--- a/src/mmw/apps/modeling/mapshed/tasks.py
+++ b/src/mmw/apps/modeling/mapshed/tasks.py
@@ -374,7 +374,7 @@ def soiln(result):
 
     result = parse(result)
 
-    soiln = result.values()[0] * 4.0
+    soiln = result.values()[0] * 7.0
 
     return {
         'soiln': soiln


arottersman:
Taking a look now.

rajadain:
Thanks for taking a look!

Issue #2533, opened 2017-11-28
Unable to select boundary for analysis if HUC Search is used previously


mmcfarland:
### Overview
When using the search box to select a HUC boundary, it adds the selected vector shape to the map for display.  If the user does not utilize the "Select" button in the search box, and instead adds a boundary layer for selection, the "search" shape prevents clicking on the UTF8 grid layer for the boundary.

### To reproduce
1. Search for `Brandywine` in the search box
2. Select the `Brandywine-Christina` result under the HUC8 header
3. Observe the HUC has been added to the map
4. Under the `Select Boundary` option on `/draw`, add the HUC8 layer
5. The shapes should overlap the Brandywine HUC selected previously
6. Note you cannot click on the Brandywine HUC to analyze, presumably the click events are being absorbed by the reference shape.

caseycesari:
In the test case described above, is it preferable that the initial HUC added to the map from searching is removed when the HUC8 layer is added?

ajrobbins:
I believe the desired behavior should be:
- User selects HUC from search box
- HUC displays on map
- User selects HUC8 option from boundary selector from the sidebar tools
- Displayed HUC is removed from the map, and clickable boundary viz/UTF8 grid appears. Clicking on a boundary selects it for analysis

PR #2564, opened 2017-12-20
Allow boundary selection after HUC search

caseycesari:
## Overview

When an alternate AoI generation method is selected, and a geocoder HUC result is rendered on the map, clear it and reset the state of the geocoder control. The geocoder result was preventing users from making a selection from the underlying boundary layer.

Connects to #2533

### Demo

mmw

## Testing Instructions

- Search for `Brandywine` in the search box
- Select the `Brandywine-Christina` result under the HUC8 header
- Observe that the HUC has been added to the map
- Under the `Select Boundary` option, select the HUC8 layer
- Verify that the geocoder HUC layer was removed from the map
- Try and select the Brandywine HUC from the boundary layer

### DIFF
--- a/src/mmw/js/src/geocode/views.js
+++ b/src/mmw/js/src/geocode/views.js
@@ -92,6 +92,10 @@ var SearchBoxView = Marionette.LayoutView.extend({
         'click @ui.selectButton': 'validateShapeAndGoToAnalyze'
     },
 
+    initialize: function() {
+        this.listenTo(App.map, 'change:selectedGeocoderArea', this.resetIfSelectedAreaIsCleared);
+    },
+
     modelEvents: {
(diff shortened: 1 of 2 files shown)


src/mmw/js/src/geocode/views.js
```diff
@@ -263,6 +267,12 @@ var SearchBoxView = Marionette.LayoutView.extend({
         App.map.set('selectedGeocoderArea', null);
     },
 
+    resetIfSelectedAreaIsCleared: function(model, selectedGeocoderArea, options) {
```
rajadain: `options` is defined but never used, causing build errors.

caseycesari: Thanks.

caseycesari:
Fixed up!

Issue #2577, opened 2017-12-28
MapShed Projects should export GMS to HydroShare


rajadain:
Currently, the HydroShare export only receives Analyze CSV files from the front-end. In case of MapShed projects, we should also export a GMS file for every scenario. The front-end has a dictionary of MapShed inputs, but generates the GMS file by POSTing that dictionary to an endpoint. Thus, it may be more prudent for the front-end to POST the dictionary to the back-end, which then generates GMS files, rather the front-end fetching them and sending back to a different endpoint.

Issue #2584, opened 2018-01-02
HydroShare Backend Refactor


rajadain:
Refactor backend to group all HydroShare related entities under the `export/hydroshare` namespace.

arottersman:
Migrations have already been applied to staging, and will have to be rolled back to move the models under the `export` namespace.

rajadain:
Refactor `HydroShareClient.add_files` to have a binary option, where `contents` would  be a file-like object instead of a string.

PR #2585, opened 2018-01-02
Export GMS File In HydroShare Resource

arottersman:
For mapshed projects include a GMS file
for each scenario

* Named like `scenario_this-is-a-scenario-name.gms`
* List of mapshed data is passed from client; gms
  file is created on the server before adding to
  hydroshare

Connects #2577 

### Demo

[Project with exported GMS files](https://beta.hydroshare.org/resource/41f8d77644654e7195f28d5604ff8c68/)

## Testing Instructions

 * Create a mapshed project and confirm you can share and re-share it on hydroshare with both 1 and multiple scenarios
* Confirm you can open the GMS files for each scenario
* Create a site storm project and confirm you can still share it on hydroshare

### DIFF
--- a/src/mmw/apps/user/hydroshare.py
+++ b/src/mmw/apps/user/hydroshare.py
@@ -3,6 +3,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import division
 
+import json
 import StringIO
 
 from rauth import OAuth2Service
@@ -12,6 +13,7 @@ from hs_restclient import HydroShare, HydroShareAuthOAuth2, HydroShareNotFound
 from django.conf import settings
 
 from apps.user.models import HydroShareToken
+from apps.modeling.tasks import to_gms_file
 
 
 CLIENT_ID = settings.HYDROSHARE['client_id']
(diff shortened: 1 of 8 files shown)


rajadain:
I merged #2580 and haven't started taking a look at this yet, so feel free to squash and rebase on `develop` and I'll take a look then.

src/mmw/apps/user/hydroshare.py
```diff
@@ -12,6 +13,7 @@
 from django.conf import settings
 
 from apps.user.models import HydroShareToken
+from apps.modeling.tasks import to_gms_file
```
rajadain: Importing this here feels like it couples this too much with modeling. Some of this will be refactored in #2584, but I think it would be a better idea to use `to_gms_file` in `views.py`, and use it to construct a `{name: 'scenario_xyz.gms', contents: '...'}` object that is added to the `files` array given to `hs.add_files`. Does this make sense?

arottersman: Yes, but it won't be as clean as described.`to_gms_file` returns a file, and shouldn't be passed as `contents` to `hs.add_files` which will try to rewrite it via `StringIO`.

rajadain: Ah right. Very well, we'll tackle that later.

src/mmw/js/src/core/modals/views.js
```diff
@@ -19,6 +19,7 @@ var _ = require('lodash'),
     modalIframeTmpl = require('./templates/iframeModal.html'),
     vizerUrls = require('../settings').get('vizer_urls'),
 
+    GWLFE = 'gwlfe',
```
rajadain: Can we import this from `modeling/models`? I'd prefer not defining the same constant in two places.

arottersman: This is in `core` which is usually agnostic to the other directories (which import it). I think instead it's time to move the model constants out into `core/utils`. We reference them in analyze, modeling, and now here.

arottersman:
Added a commit to move the model constants into `core/utils. I think this is ready to merge once the fixup is squashed. Let me know if you want to take another look.

rajadain:
Taking another look.

arottersman:
Thanks for the review!

Issue #2346, opened 2017-10-06
Add Terrain tab to Analysis


ajrobbins:
Add an additional tab to Analysis for Terrain summary statistics, mocked up in this google doc: https://docs.google.com/document/d/1-iTJadY2V5uE46Jh4XFQKyiNKOcqVPqJVCM_TkEBkBI/edit#

The tab will have two related layers, slope and elevation. We currently use a % slope raster for Mapshed, so that will need to be tiled for visualization. We'll ingest the elevation raster in #2347

This card is to implement the geoprocessing to derive the following summary statistics for those two rasters: 
- average, minimum, and maxiumum % slope in AoI
- average, minimum, and maximum elevation (m) in AoI

PR #2586, opened 2018-01-02
Analyze Terrain

rajadain:
## Overview

Adds a Terrain tab to the Analyze stage. This uses the new `RasterSummary` operation in the geoprocessing service to get summary statistics for Elevation and Slope rasters.

Connects #2346

### Demo

image

Also tagging @ajrobbins and @jfrankl for visual review.

### Notes

* There are no related visual layers corresponding to this tab currently
* Can someone confirm the data source name? I took it from the mockup but want to make sure, and add it to the API doc.
* At first I made the transposed table, which felt more natural in my mind:

    image

    Later I corrected it according to the mockup. The correction code is in the last two fixups, so can be easily reverted if we like this version better.

## Testing Instructions

* Check out this branch and reprovision worker to install the new geoprocessing service

      vagrant reload worker --provision

* Go to [:8000/](http://localhost:8000/) and select an area of interest. Proceed to Analyze.
* Ensure you see a Terrain tab under Analyze. Ensure it has reasonable values. Ensure its design matches [this mockup](https://docs.google.com/document/d/1-iTJadY2V5uE46Jh4XFQKyiNKOcqVPqJVCM_TkEBkBI/edit#)

### DIFF
--- a/src/mmw/apps/geoprocessing_api/tasks.py
+++ b/src/mmw/apps/geoprocessing_api/tasks.py
@@ -34,6 +34,7 @@ RWD_PORT = os.environ.get('RWD_PORT', '5000')
 
 ACRES_PER_SQM = 0.000247105
 CM_PER_MM = 0.1
+M_PER_CM = 0.01
 
 
 @shared_task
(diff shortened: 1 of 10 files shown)


ajrobbins:
I see the logic of the transposed table, but all of our other Analyze tables have the variable names & units on the top, so this should match. Let's keep it to matching the mockup.

The data source name should be: "NHDPlus V2 NEDSnapshot DEM"

kellyi:
Setting this up to review.

kellyi:
Are we adjusting the size of the panel as part of this work or will we do that subsequently? As pictured above the Terrain tab currently goes onto a second row and when "Point Sources" or "Water Quality" is highlighted, both "Terrain" and "Streams" are sent to the second row.

rajadain:
> Are we adjusting the size of the panel as part of this work or will we do that subsequently?

I think subsequently. @jfrankl is tagged for visual review here, and if he has any immediate suggestions I'll implement, but I've also heard talk that we may be dropping some analyses, which will make room.

kellyi:
Working well! Large AOI timed out on my local but smaller AOI worked fine:

screen shot 2018-01-03 at 2 53 09 pm

screen shot 2018-01-03 at 2 53 20 pm

src/mmw/apps/geoprocessing_api/tasks.py
```diff
@@ -229,3 +230,112 @@ def collect_climate(results):
...
     }
+
+
+@shared_task
+def analyze_terrain(result):
+    """
+    Starts a job to produce summary statistics for elevation and slope in a
+    given area.
+
+    TODO Where does the NED and Slope data come from?
```
kellyi: cc @arottersman -- do you know the answer to this?

src/mmw/apps/geoprocessing_api/tasks.py
```diff
@@ -229,3 +230,112 @@ def collect_climate(results):
...
+    produces:
+        - application/json
+    """
+    if 'error' in result:
+        raise Exception('[analyze_terrain] {}'.format(result['error']))
+
+    categories = [
+        dict(type='average',
+             elevation=(result[0]['avg'] * M_PER_CM),
+             slope=result[1]['avg']),
```
kellyi: This is probably fine since we're just unpacking the results but if we wanted to assign these beforehand I think we could do:

```python
[elevation, slope] = result
categories = [
...
```

and then use `elevation` and `slope` in place of `result[0]` & `result[1]`. Arguable whether that's an improvement, though.

src/mmw/apps/geoprocessing_api/views.py
```diff
@@ -1168,6 +1168,26 @@ def start_analyze_climate(request, format=None):
...
 
 
+@decorators.api_view(['POST'])
+@decorators.authentication_classes((TokenAuthentication, ))
+@decorators.permission_classes((IsAuthenticated, ))
+@decorators.throttle_classes([BurstRateThrottle, SustainedRateThrottle])
+@log_request
+def start_analyze_terrain(request, format=None):
+    """
+
```
kellyi: Should this have a docstring?

rajadain: Good catch meant to put that in.

src/mmw/apps/geoprocessing_api/tasks.py
rajadain: I like it! Will fix.

src/mmw/apps/geoprocessing_api/tasks.py
ajrobbins: NED data comes from NHDPlus V2 NEDSnapshot, and Slope data is from NED (National Elevation Dataset)

rajadain:
Fixed source name, docstrings, and code style. Going to squash and merge shortly.

rajadain:
Screenshot with updated source:

image

Merging this now. Thanks for reviewing!

PR #2587, opened 2018-01-03
HydroShare Export Backend Refactor

rajadain:
## Overview

Refactors HydroShare backend, grouping the models, serializers, and services under `export/hydroshare`. Also updates migrations. Also adds to `HydroShareClient.add_files` the ability to handle file-like objects.

Connects #2584 

## Testing Instructions

* Check out this branch
* Run migrations

      $ ./scripts/manage.sh migrate

* Export a project to HydroShare and ensure it works correctly


rajadain:
Will rebase over #2585 once that is in.

rajadain:
> on staging we'll run all migrations which will delete the hydroshare resources already created in modeling. Related projects are not deleted because the project is a foreign key on the resource and not the other way around.

Correct. The projects won't be deleted, but their link to HydroShare will be broken. Any existing projects will have to be re-exported to new resources, and any existing HydroShare resources will be orphaned.

> On production running migrations will first create the modeling hydroshare resource, destroy it, and then finally create the export version.

Correct. We'll be creating and then immediately destroying a table, but it will have no lasting effects.

rajadain:

src/mmw/apps/export/views.py
```diff
@@ -75,12 +76,16 @@ def hydroshare(request):
...
     if hsresource and request.method == 'POST':
         # Update files
-        hs.add_files(hsresource.resource,
-                     params.get('files', []),
-                     overwrite=True)
-        hs.add_gms_files(hsresource.resource,
-                         params.get('mapshed_data', []),
-                         overwrite=True)
+        files = params.get('files', [])
+        for md in params.get('mapshed_data', []):
```
arottersman: Seeing an error for TR-55. I think we needed that guard for when there's no mapshed data:

```
TypeError at /export/hydroshare
'NoneType' object is not iterable
...
 126.     for md in params.get('mapshed_data', []):
```

rajadain: Or we could just send an empty array when not MapShed. Good catch, will fix.

rajadain:
bb0c26f should fix it. Please confirm when you get a chance.

rajadain:
Thanks for taking a look! Squashed the fixup, merging this shortly.

Issue #2579, opened 2017-12-28
HydroShare Export Design Issues


rajadain:
### Language Review

* Is "unsynchronize" the best term? Should we use "desynchronize" or "disconnect" instead?
* ???

### Interaction Review

* Add error message if trying to export project without filling in Title and Abstract
* ~~If modal is closed after export is started but not finished, we get a JavaScript error:~~

    ![image](https://user-images.githubusercontent.com/1430060/34530388-07257ee4-f07d-11e7-997a-d3532a02814e.png)

* Add a "Share" button to the secondary nav as shown here: https://share.goabstract.com/1bd9e502-5ef5-40d8-aba0-868134986fb5 and remove it from the project dropdown.
  - While HydroShare is autosyncing, add a special class to that button, which will be used to style it later.

jfrankl:
>Is "unsynchronize" the best term? Should we use "desynchronize" or "disconnect" instead?

I agree, it's clunky. When someone accepts, we remove their project from HydroShare, correct? What if we use language that involves "removing"?

![image](https://user-images.githubusercontent.com/1809908/34539341-18ef3efa-f09e-11e7-9781-047f4aefdaa6.png)

>Add error message if trying to export project without filling in Title and Abstract

Yes, please refer to the design pattern used in the login modal for the validation. https://share.goabstract.com/01ead2a3-57f8-493d-bd74-0df6d445ec5d

Issue #2574, opened 2017-12-28
HydroShare Autosync


rajadain:
Currently a project is only exported to HydroShare once when the switch is turned on. Instead, we want to be able to do multiple exports. As seen in [this mockup](https://app.goabstract.com/projects/76ed1390-dea8-11e7-a27c-0f868efbc56f/branches/master/commits/9e8969dfe9b49ab710e304f769841f4eaad70b0a/files/7F4A25A6-1E0D-48A0-B999-8629B42E4E23/layers/C07A47A4-4529-41DF-BBF4-D1F096C00279):

![ce2c2eaf-ff6e-475c-93e7-438359702a02](https://user-images.githubusercontent.com/1430060/34419905-6e96a9d4-ebd4-11e7-9213-dcc60bdeea33.png)

If this box is checked, the project to export to HydroShare every time there are any changes to the project or scenarios. Since the Area of Interest is not changing, we do not need to re-export that or any of the Analyze tabs.

Furthermore, the use may uncheck the "autosync" checkbox, as seen in [this mockup](https://app.goabstract.com/projects/76ed1390-dea8-11e7-a27c-0f868efbc56f/branches/master/commits/9e8969dfe9b49ab710e304f769841f4eaad70b0a/files/7F4A25A6-1E0D-48A0-B999-8629B42E4E23/layers/0B77D142-5D4D-4D26-82B2-FC6DD28ACC4A)

![822b407b-5a34-459b-a14b-7e35e69693e3](https://user-images.githubusercontent.com/1430060/34419957-bfbf47da-ebd4-11e7-8f93-ff38bede8ec7.png)

In this case, whenever the user clicks "Publish Now", the project should be exported to HydroShare.

Depends on #2573

PR #2590, opened 2018-01-04
HydroShare Export Autosync

rajadain:
## Overview

Refactors the network interaction for HydroShare Export into the Project model from the MultiShareModal view. Reuses that functionality in the Scenario save code, triggering on every save if autosync is enabled. Adds backend and frontend support to turn autosync on or off. Updates UI to match the mockups.

Connects #2574 

### Demo

Exporting to HydroShare manually:

Exporting to HydroShare automatically (on scenario change):

### Notes

* Exporting to HydroShare takes about ~15-25 seconds. Running a model takes between 1 and 30 seconds. When a change is made to a scenario, we first save the inputs / modifications, then run the model. Once the model results return, we save them to a scenario again. To avoid triggering two exports to HydroShare in quick succession of input / modification change and result save, I added a debounce of 5 seconds to the export. This means that for every save, the export will wait 5 seconds to see if there is another, and will trigger an export only after that time passes.

  I think we can adjust this number going forward, but if you can think of a better initial setting please comment here.

* I'm currently using default styling for the checkbox:

    image

  But can switch to something else if @jfrankl recommends to match the mockup:

  ce2c2eaf-ff6e-475c-93e7-438359702a02

  Or we can tackle this in the design review card #2579 

* When a scenario is renamed, a GMS file with the old name continues to live in HydroShare, and a new one with the new name is added. This is because we only delete the files we are uploading to overwrite them. An old named scenario file must be deleted by hand on HydroShare.

## Testing Instructions

* Check out this branch and bundle
* Try exporting a MapShed project to HydroShare. Ensure it works.
* Make changes to the project (add a scenario, rename a scenario, add a modification). Ensure they all trigger a HydroShare export.
* Disable autosync. Ensure changing scenarios does not trigger exports anymore.
* Export manually. Ensure it works.

### DIFF
--- a/src/mmw/apps/export/views.py
+++ b/src/mmw/apps/export/views.py
@@ -29,7 +29,7 @@ HYDROSHARE_BASE_URL = settings.HYDROSHARE['base_url']
 SHAPEFILE_EXTENSIONS = ['cpg', 'dbf', 'prj', 'shp', 'shx']
 
 
-@decorators.api_view(['GET', 'POST', 'DELETE'])
+@decorators.api_view(['GET', 'POST', 'PATCH', 'DELETE'])
 @decorators.permission_classes((IsAuthenticated, ))
 def hydroshare(request):
     # Get HydroShare client with user's credentials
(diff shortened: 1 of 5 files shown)


ajrobbins:
Visual review: appears to work as advertised! @jfrankl ?
The 15-25 time to sync to HS is a bit unfortunate. Do you have a sense of what's causing the bulk of that time? Did you get a chance to look at any other apps that push to HS, and see if they take the same amount of time?

rajadain:
> Do you have a sense of what's causing the bulk of that time?

The file uploads. We're uploading files one at a time, and we push the Analyze results each time. Currently that's 8 files, each takes a couple seconds to go through.

> Did you get a chance to look at any other apps that push to HS, and see if they take the same amount of time?

I haven't, but that's a good idea.

One thing we can look in to off the top of my head is bulk uploads, which will at least get rid of the overhead for each file by batching them all together. I'll make an "investigate performance" card for follow up.

ajrobbins:
Sounds good - would be an enhancement.

arottersman:
Taking a look!

arottersman:
Wi-Fi is down at the office, so will approve this tomorrow after I've finished testing IE11 and FF. Was working perfectly on chrome, and the refactor looks great.

Is there a rate limit on the hydroshare api?

rajadain:
> Is there a rate limit on the hydroshare api?

Not as far as I can tell. I think the only limit is the REST timeout, which may prevent uploading very large files, but none of our files approach that limit.

rajadain:
Thanks for reviewing!

PR #2595, opened 2018-01-08
Add MMW Users to API Logging

rajadain:
## Overview

Adds MMW Users to API request logging by adding `SessionAuthentication` as an option to all API fields, and adding an `api` field to the `RequestLog` table that is `True` when using `TokenAuthentication` and `False` otherwise.

To see all legitimate API requests, run:

```sql
SELECT l.*
FROM core_requestlog l INNER JOIN auth_user u ON l.user_id = u.id
WHERE l.path LIKE '/api%'
  AND u.username <> 'mmw|client_app_user'
  AND l.api = 'true';
```

To see logged-in UI requests for API paths, run:

```sql
SELECT l.*
FROM core_requestlog l INNER JOIN auth_user u ON l.user_id = u.id
WHERE l.path LIKE '/api%'
  AND l.api = 'false';
```

To see anonymous UI requests for API paths, run:

```sql
SELECT l.*
FROM core_requestlog l INNER JOIN auth_user u ON l.user_id = u.id
WHERE l.path LIKE '/api%'
  AND u.username = 'mmw|client_app_user'
  AND l.api = 'true';
```

To see logged-in requests for non-API paths, run:

```sql
SELECT l.*
FROM core_requestlog l
WHERE l.path NOT LIKE '/api%'
  AND l.user_id IS NOT NULL;
```

To see anonymous requests for non-API paths, run:

```sql
SELECT l.*
FROM core_requestlog l
WHERE l.path NOT LIKE '/api%'
  AND l.user_id IS NULL;
```

Connects #2451 

### Notes

The `core_requestlog.api` field is meaningless when the `path` does not begin with `/api`. Going forward, all those requests will have `api=False`, but for all existing rows they will have `api=True`. However, this doesn't matter, as they cannot be accessed via API anyway.

## Testing Instructions

* Check out this branch and run migrations `./scripts/manage.sh migrate`
* Make some requests via the front-end when logged in.
* Make some anonymously.
* Make some via the API.
* Ensure they are all logged according to the description above.

### DIFF
--- a/src/mmw/apps/core/decorators.py
+++ b/src/mmw/apps/core/decorators.py
@@ -42,7 +42,9 @@ def log_request(view):
             query_params=request.query_params.dict(),
             method=request.method,
             host=request.get_host(),
-            remote_addr=_get_remote_addr(request))
+            remote_addr=_get_remote_addr(request),
+            referrer=request.META.get('HTTP_REFERER'),
+            api=_is_api_call(request))
         try:
             log.save()
         except Exception:
@@ -66,3 +68,9 @@ def _get_remote_addr(request):
             return [x.strip() for x in ipaddr.split(",")][0]
     else:
         return request.META.get("REMOTE_ADDR", "")
+
+
+def _is_api_call(request):
+    # request.auth is None for SessionAuthentication,
+    # but a Token instance for TokenAuthentication
+    return request.auth is not None
--- a/src/mmw/apps/core/migrations/0003_requestlog_api_referrer.py
+++ b/src/mmw/apps/core/migrations/0003_requestlog_api_referrer.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0002_requestlog'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='requestlog',
+            name='api',
+            field=models.BooleanField(default=True, help_text='Whether this was an API call (true) or MMW UI call (false)'),
+        ),
+        migrations.AddField(
+            model_name='requestlog',
+            name='referrer',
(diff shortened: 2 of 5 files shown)


src/mmw/apps/core/decorators.py
```diff
@@ -66,3 +67,9 @@ def _get_remote_addr(request):
             return [x.strip() for x in ipaddr.split(",")][0]
     else:
         return request.META.get("REMOTE_ADDR", "")
+
+
+def _is_api_call(request):
+    # request.auth is None for SessionAuthentication,
+    # but a Token instance for TokenAuthentication
+    return request.auth is not None
```
arottersman: This is perhaps a low-priority edge case, but I think at least worth bringing up: Making a swagger API call with an MMW session will cause a false negative (maybe because `SessionAuthentication` is listed first in the `authentication_classes` it takes precedence). I think this will be the case for all users making their API requests from the browser (eg from a notebook).

We might want to consider another method of filtering, like checking if the `TokenAuthentication` user is the `MMW_CLIENT_APP_USER`.

arottersman: Terence just pointed out to me it doesn't make any sense at all for the session info to be sent from a non-MMW app instance, so we don't have to worry about python notebooks or anything like that.

If we want to keep track of the swagger calls separately or exclude them from the UI counts, we can keep track of the referrer which will end in `/api/docs`.

rajadain:
I just pushed a set of updates that adds a `referrer` field to the `RequestLog` model so we can exclude some of them during reporting.

To re-test this, please remove the `api` column from the `core_requestlog` table and the outdated migration:

```sql
ALTER TABLE core_requestlog DROP COLUMN api;

DELETE FROM django_migrations WHERE id = (SELECT MAX(id) FROM django_migrations);
```

Then check out this branch from scratch and re-run the migrations:

```bash
$ ./scripts/manage.sh migrate
```

Then test again.

rajadain:
Thanks for reviewing. Merging this now.

Issue #2571, opened 2017-12-27
Can't highlight "jump to location/HUC" text in geocoder


ajrobbins:
Observed:
- Entered a HUC search term in the geocoder, "Brandywine"
- Selected one of the HUC results
- HUC was selected appropriately
- Double clicking on the search term moves the cursor to the beginning of the term, and doesn't highlight it. Typing replaces the term.
![screen shot 2017-12-27 at 5 15 39 pm](https://user-images.githubusercontent.com/12868342/34394832-ffb0f16e-eb29-11e7-999b-9cf5cf434053.png)

Expected:
- Double clicking highlights the search term
- Single click places the cursor appropriately where the click occurs

PR #2606, opened 2018-01-12
Make Geocoder Boundary Result Highlightable

arottersman:
## Overview

We were displaying the name of the selected
boundary result as the geocoder input's
placeholder. Display it instead as the input's
value so that it can be edit/selected, copied,
etc.

Connects #2571 

### Demo
ajarcfdeqf

## Testing Instructions

 * Search for a HUC (eg `Brandywine`) and select a result. 
 * Confirm you can edit the result and select it by double clicking
 * Confirm the geocoder searches again if you change the text and that you can select a new result


arottersman:
Thanks!

Issue #2597, opened 2018-01-10
Manage HydroShare connectivity in user profile


ajrobbins:
Implement designs to add a "Linked Accounts" page in the user profile, which currently shows one account type: HydroShare. Unlinked view offer button to link account, which triggers sign in/registration iframe workflow. If account is already linked, there is a "remove" button which triggers a warning modal. 

https://app.goabstract.com/projects/76ed1390-dea8-11e7-a27c-0f868efbc56f/branches/master/commits/bc0a76294913cb01f9f5f9e9414e10ba488f47f7/files/7F4A25A6-1E0D-48A0-B999-8629B42E4E23/layers/1780B15C-07E8-43A8-94C5-AD93273D54AD

https://app.goabstract.com/projects/76ed1390-dea8-11e7-a27c-0f868efbc56f/branches/master/commits/4dfccc5aeb3a8f9092ded94ffed725f9bac30168/files/7F4A25A6-1E0D-48A0-B999-8629B42E4E23/layers/927B3021-5B59-441D-AA77-338E617B76AE

https://app.goabstract.com/projects/76ed1390-dea8-11e7-a27c-0f868efbc56f/branches/master/commits/bc0a76294913cb01f9f5f9e9414e10ba488f47f7/files/7F4A25A6-1E0D-48A0-B999-8629B42E4E23/layers/A73E4F1C-6E56-4419-B962-FDF5B3009075

https://app.goabstract.com/projects/76ed1390-dea8-11e7-a27c-0f868efbc56f/branches/master/commits/14aab341b7aae7bc15b6832790f07abdaa2fc605/files/7F4A25A6-1E0D-48A0-B999-8629B42E4E23/layers/2B17CA8E-C59E-4921-BE34-D72F563A63B4

PR #2607, opened 2018-01-15
Manage Linked Accounts

rajadain:
## Overview

Adds a "Linked Accounts" section to the "My Account" page, where users can manage various third party applications that they have connected to Model My Watershed. Currently there is only one: HydroShare. From this page they can connect or disconnect their account from HydroShare. If they disconnect their account, any linked projects will be disconnected as well, orphaning the corresponding resources on HydroShare. If they reconnect their account, those projects will have to be setup to export again, and they will export to new resources, not the ones they were previously linked to.

Connects #2597 

### Demo

Text of the remove link modal:

image

## Testing Instructions

* Go to [:8000/](http://localhost:8000/) and log in
* Go the My Accounts page and see the new Linked Accounts section. Ensure sidebar navigation works correctly.
* Try connecting to HydroShare. Ensure you are able to, and the flow makes sense.
* Make a project or two and export it to HydroShare.
* Try disconnecting from HydroShare. Ensure you are able to, and the flow makes sense.
* Ensure the projects created above are no longer connected to HydroShare, but the resources in HydroShare remain and are not deleted.

### DIFF
--- a/src/mmw/apps/user/views.py
+++ b/src/mmw/apps/user/views.py
@@ -14,6 +14,7 @@ from django.contrib.auth.forms import (PasswordResetForm,
 from django.core.exceptions import ObjectDoesNotExist
 from django.contrib.sites.shortcuts import get_current_site
 from django.conf import settings
+from django.db import transaction
 
 from registration.models import RegistrationProfile
 from registration.forms import RegistrationFormUniqueEmail
@@ -25,6 +26,7 @@ from rest_framework.permissions import (AllowAny,
                                         IsAuthenticated)
 
 from apps.export.hydroshare import HydroShareService
+from apps.export.models import HydroShareResource
 from apps.home.views import get_context
 from apps.user.models import ItsiUser, UserProfile, HydroShareToken
 from apps.user.itsi import ItsiService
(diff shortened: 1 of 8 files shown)


src/mmw/apps/user/views.py
```diff
@@ -381,3 +382,21 @@ def hydroshare_auth(request):
                                'possibly misconfigured', 'warning')
 
     return render_to_response('user/hydroshare-auth.html', context)
+
+
+@decorators.api_view(['POST'])
+@decorators.permission_classes((IsAuthenticated, ))
+def hydroshare_logout(request):
```
jwalgran: We are deleting from two different tables so we should decorate this with `@transaction.atomic` so that the deletes succeed or fail together.

src/mmw/js/src/account/views.js
```diff
@@ -48,6 +50,34 @@ var LinkedAccountsView = Marionette.ItemView.extend({
...
+
+                        'Continue?'
+                    ],
+                    confirmLabel: 'Remove link',
+                })
+            });
+
+        confirm.render();
+        confirm.on('confirmation', function() {
+            $.post('/user/hydroshare/logout/').done(function() {
```
jwalgran: Do we need to worry about reporting failure here? I used "offline" mode to test this and it fails relatively gracefully by just not doing anything. We may want to consider showing some failure message as a way of confirming that we did try to unlink the account.

src/mmw/apps/user/views.py
rajadain: Ah great idea, I didn't know about that!

jfrankl:
LGTM

rajadain:
Thanks for reviewing!

Issue #2602, opened 2018-01-11
Streams tab small edits


ajrobbins:
- [x] For bottom row of table in streams tab: Since "totals" could mean "sum" or "average" in this table, change to "combined".

- [x] Use two decimal places for all values in streams table

Issue #2601, opened 2018-01-11
Typo in terrain tab


ajrobbins:
"statistics" is misspelled

rajadain:
Shares points with #2602

PR #2608, opened 2018-01-15
Analyze Tabs Minor UI Fixes

rajadain:
## Overview

* Fixes typo in Terrain tab
* Changes "Total" to "Combined" in Streams tab since in one column we sum and in the other we average
* Updates Stream tab to always use 2 decimal places for all values

Connects #2601 
Connects #2602

### Demo

image

image

## Testing Instructions

* Pick a shape and analyze. Inspect the Terrain and Streams tabs for the above fixes.

### DIFF
--- a/src/mmw/js/src/analyze/templates/streamTable.html
+++ b/src/mmw/js/src/analyze/templates/streamTable.html
@@ -17,22 +17,22 @@
     <tfoot>
         <tr>
             <th>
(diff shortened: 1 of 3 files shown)


kellyi:
Setting this up to take a look!

rajadain:
Thanks for reviewing!

Issue #2578, opened 2017-12-28
HydroShare Export Details


rajadain:
Make a project "public" and "shareable", and mark it as "published" when exporting to HydroShare.

Also include the bounds of the area of interest in the "coverage" science metadata so that it may be shown on the map in HydroShare and can be searched for.

rajadain:
When trying to add the geographic coverage to a resource, to enable a map in HydroShare like this:

![image](https://user-images.githubusercontent.com/1430060/34620796-bb2f96ac-f214-11e7-9aaa-a30c10181cf7.png)

I get the following API error:

```json
{
    "resource": "Resource metadata update failed: <class 'django.core.exceptions.ValidationError'>, Coverage metadata can't be updated for CompositeResource resource"
}
```

Apparently a `CompositeResource` cannot have coverage specified? Should we switch to some other resource type or talk to the HydroShare developers? /cc @ajrobbins

rajadain:
Asking them in https://github.com/hydroshare/hydroshare/issues/2543. Also it seems like using `GenericResource` doesn't work all the time. Will post here once I have answers.

rajadain:
From this comment: https://github.com/hydroshare/hydroshare/issues/2543#issuecomment-356639662

> > While trying to add a coverage to a CompositeResource, I get this error:
>
> This is by design for now. Composite resource coverage is a superset of all the coverages of all the file types. Resource level coverage is auto updated when the file type coverage changes. However, after phase-2 of composite resource implementation, it should be possible to edit resource level coverage.

So do we not add the geographic information to HydroShare or do we switch to using `GenericResource`?

rajadain:
Unblocked thanks to https://github.com/hydroshare/hydroshare/issues/2543#issuecomment-358096010

> Here is an example of using hs_restclient to set/create a NetCDF file type:
> 
>     >>> from hs_restclient import HydroShare, HydroShareAuthBasic
>     >>> auth = HydroShareAuthBasic(username='myusername', password='mypassword')
>     >>> hs = HydroShare(auth=auth)
>     >>> options = {
>                      "file_path": "file.nc",
>                      "hs_file_type": "NetCDF"
>                   }
>     >>> result = hs.resource('ID OF RESOURCE GOES HERE').functions.set_file_type(options)
> 
> For creating a GeoFeature file type using a shp file set the file_path to path of the shp file and hs_file_type to 'GeoFeature'
> 
> Note: The file specified for file_path must exist for the specified resource.

PR #2609, opened 2018-01-16
HydroShare Export Details

rajadain:
## Overview

Makes the resource in HydroShare public and shareable, so it can be seen and discovered by other users with simple search. Also declares the shapefile as a "GeoFeature" which allows HydroShare to display a bounding box around the area of interest in its UI. This may also be used for geospatial filtering of results.

Connects #2578 

### Demo

screen shot 2018-01-16 at 3 59 03 pm

## Testing Instructions

* Check out this branch
* Create a new project and export to HydroShare. Open the resource in HydroShare.
* Ensure that the resource is public and shareable, and shows the bounding box of the area of interest.

### DIFF
--- a/src/mmw/apps/export/views.py
+++ b/src/mmw/apps/export/views.py
@@ -164,6 +164,11 @@ def hydroshare(request):
                                'area-of-interest.{}'.format(ext))
         os.remove(filename)
 
+    # Make resource public and shareable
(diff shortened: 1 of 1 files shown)


arottersman:
Taking a look now!

rajadain:
Thanks for reviewing!

Issue #2611, opened 2018-01-18
Projects Page: share button doesn't show hydroshare connection


arottersman:
On the project list page (`/projects`), clicking the share button correctly shows the MultiShareModal. The modal, however, doesn't show the hydroshare state correctly, always showing the connection as "Off". If you switch hydroshare "On" for a project that already has a connected resource, you'll be prompted to create a new one.

PR #2614, opened 2018-01-22
Fix Share modal behavior in project listing

rajadain:
## Overview

The Share modal relies on the `project.hydroshare` field, which was missing from the `ProjectListing` serializer. By adding it there, we ensure that the share modal works correctly.

Connects #2611 

## Testing Instructions

* Ensure you have at least one project exported to HydroShare
* Go to the My Projects listing, and click the Share button for any project. Ensure it reflects the correct state of sharing for that project.
* Try sharing a project from the listing page. Ensure it is exported to HydroShare correctly.

### DIFF
--- a/src/mmw/apps/modeling/serializers.py
+++ b/src/mmw/apps/modeling/serializers.py
@@ -93,7 +93,10 @@ class ProjectListingSerializer(gis_serializers.GeoModelSerializer):
     class Meta:
         model = Project
         fields = ('id', 'name', 'area_of_interest_name', 'is_private',
-                  'model_package', 'created_at', 'modified_at', 'user')
+                  'model_package', 'created_at', 'modified_at', 'user',
+                  'hydroshare')
+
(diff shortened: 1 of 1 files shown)


rajadain:
Hmm. That is a quirk, but not very destructive, since even if the user doesn't see the right state in the detail page, they can see it if they refresh the page. Even if they share again without refreshing the page, we re-export to the same resource, so it doesn't perform any destructive operations.

The way to fix it would be to either:

1. Persist the export progress in the database before it completes, so subsequent fetches know it's happening, which is a non-trivial change
2. Transfer that state to `App.currentProject` somehow in addition to loading everything from the URL, which would be very hacky

I think it's small enough to warrant ignoring. /cc @ajrobbins 

I'm going to merge this in for now. If we deem that to be necessary, we can make a card for it later.

ajrobbins:
I agree that we do not need to accommodate that edge case at this time.

Issue #2615, opened 2018-01-23
HydroShare Timeout Causes Front-end Crash


rajadain:
![image](https://user-images.githubusercontent.com/1430060/35285834-5bff3374-002c-11e8-8f6b-7ef5cebceef5.png)

This should be handled gracefully.

PR #2616, opened 2018-01-23
HydroShare Export Handle Timeout

rajadain:
## Overview

Prevents front-end crash from HydroShare export timeouts. Also shortens re-export time by skipping pre-exported analyses.

Connects #2615 

## Testing Instructions

* On `develop`, export a MapShed project to HydroShare. Then add a scenario / make a change. Observe the network tab and the console, and see the timeout and JavaScript error.
* Check out this branch, but only the first commit 532c6605. Update your project and see the timeout now being handled in the front-end.
* Now checkout the second commit 0b54dc18. Update your project and see no timeout.

### DIFF
--- a/src/mmw/apps/export/hydroshare.py
+++ b/src/mmw/apps/export/hydroshare.py
@@ -121,3 +121,9 @@ class HydroShareClient(HydroShare):
             return True
         except HydroShareNotFound:
             return False
+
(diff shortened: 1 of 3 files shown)


rajadain:
Thanks for the quick review! Merging now.

Issue #2600, opened 2018-01-11
Add explanatory tooltips to streams tab


ajrobbins:
- Next to "other" - add a tooltip that reads: "Contains canals, artificial paths, coastlines, and other flowlines without a flow direction"
- Add tooltip next to the length in X areas lines that reads: "Derived using a uniform one-pixel (30m) riparian width for flowlines"

rajadain:
Is this also blocked awaiting further instruction?

> need to ascertain what features
> (cell size?)
> riparian buffer

ajrobbins:
I need to clarify the riparian buffer comment, but the "other" one needs to be done by us - As implemented in the streams tab, determine which water feature/stream types are included in the "other" row in the tables.

rajadain:
The code for generating that table is here:



So we group all 0 or empty stream orders into the "Other" field. If we check to see how many of each type there are in the table:

```sql
SELECT COUNT(*) FROM nhdflowline; -- 2993515

SELECT stream_order, COUNT(*) AS count, ROUND(100.0 * COUNT(*) / 2993515, 5) AS percent
FROM nhdflowline
GROUP BY stream_order
ORDER BY stream_order;
```

we get the following results:

| stream_order | count   | percent     |
|--------------|---------|-------------|
| **0**       | **101**     | **0.00337396**  |
| 1            | 1352677 | 45.18691238 |
| 2            | 602309  | 20.1204604  |
| 3            | 323143  | 10.79476802 |
| 4            | 182916  | 6.110408667 |
| 5            | 103278  | 3.450057875 |
| 6            | 59909   | 2.001292795 |
| 7            | 29097   | 0.972001142 |
| 8            | 10403   | 0.347517884 |
| 9            | 3417    | 0.114146747 |
| 10           | 603     | 0.020143544 |
| **NULL**     | **325662**  | **10.87891659** |

From the above, **0** and **NULL** values are grouped under "Other". I'm not sure what that signifies. Maybe @arottersman has a better idea, having touched that table last.

kellyi:
I used Other as a catch all for the 0 and Null values; Null are there because the data was unavailable. 0 initially seemed like bad or missing data, too, but it may signal small streams which aren't connected to any others: https://en.wikipedia.org/wiki/Strahler_number

If 0 is meaningful enough we can add a row for it to the table.

ajrobbins:
[...]

PR #2618, opened 2018-01-24
Add help text tooltips to stream tab

mmcfarland:
## Overview

Adds static text for stream length popovers and dynamic text for popovers on each stream order row.  Only "Other" has been provided with  text at present.

Connects #2600 

### Demo

Optional. Screenshots, `curl` examples, etc.
image

## Testing Instructions

 * Analyze any AoI and verify that 3 new tooltips are added for the items listed in the connected issue.


mmcfarland:
Thanks.  I do agree that the :heavy_multiplication_x: is superfluous, but I kept the markup consistent with the existing popup on that tab, which has the same behavior.  Since it's not technically wrong (the :heavy_multiplication_x: _does_ close it), I'm inclined to leave it for efficiency's sake.

PR #2625, opened 2018-01-25
Download Area of Interest

rajadain:
## Overview

Adds a button to the Analyze tab which allows downloading the area of interest as a Shapefile or GeoJSON.

Connects #2603 

### Notes

Any recommendations on how to validate the shape?

## Testing Instructions

* Go to [:8000/](http://localhost:8000/) and select / draw an area of interest. Notice there is a download button on the top right of the Analyze sidebar.
* Click the button to reveal two options: Shapefile and GeoJSON. Click both and inspect the files in Qgis or similar to verify it is the right shape.
* Ensure this works with different styles of area of interest.
* Ensure that this button is available in the Analyze tab when in the Modeling stage.
* Go to [:8000/?bigcz](http://localhost:8000/?bigcz) and select an area of interest. Search for anything, like "Water". Click the download button in the top right and ensure you can still download a JSON of results.

### DIFF
--- a/src/mmw/apps/export/urls.py
+++ b/src/mmw/apps/export/urls.py
@@ -5,9 +5,10 @@ from __future__ import division
 
 from django.conf.urls import patterns, url
 
-from apps.export.views import hydroshare
+from apps.export.views import hydroshare, shapefile
 
 urlpatterns = patterns(
     '',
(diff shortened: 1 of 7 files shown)


rajadain:
Thanks. Added a commit for validation, merging now.

Issue #2622, opened 2018-01-24
Add BiG-CZ Search to Monitor Tab


rajadain:
Mockups:

![image](https://user-images.githubusercontent.com/1430060/35342397-f3e62520-00f5-11e8-8ca6-9f3982658bb1.png)

![image](https://user-images.githubusercontent.com/1430060/35342408-fae2b398-00f5-11e8-86ed-a2d5b05cbeab.png)

![image](https://user-images.githubusercontent.com/1430060/35344208-88dd96d2-00fa-11e8-83b2-adfa9213b179.png)

Current BiG-CZ Screenhots:

![image](https://user-images.githubusercontent.com/1430060/35342938-611120d6-00f7-11e8-8b0f-067b982f3b32.png)

![image](https://user-images.githubusercontent.com/1430060/35342961-6f4b36e6-00f7-11e8-9ccf-f2c61a8ddd0a.png)

Add this functionality to the Monitor tab introduced in #2620.

- [x] Searching for a keyword should behave the same as in BiG-CZ and use the same endpoints
- [x] The map should be updated with results, same as in BiG-CZ
- [ ] When user switches to the Analyze tab, the map search results should be cleared from the map
- [x] When user switches back to Monitor tab, the tab state should be preserved (scroll, sub-tab selection, detail view)
- [ ] The "Download Results" button should now be in the tricolon button menu ⁝, since the download icon is now used for downloading the area of interest shape.
- [x] Do not worry about pagination in this issue.

Issue #2620, opened 2018-01-24
Add Monitor Tab to Analyze Sidebar


rajadain:
Mockups:

![image](https://user-images.githubusercontent.com/1430060/35339849-a83d4352-00ef-11e8-91b3-924601635b91.png)

![image](https://user-images.githubusercontent.com/1430060/35339906-c91178dc-00ef-11e8-8928-a12b104440dc.png)

- [x] Add a top-level tab bar to the Analyze Sidebar, which should say "Analyze" and "Monitor"
- [x] The Monitor tab should have the area of interest name and size, and the download button (from #2603), but nothing else at this time.
- [x] Switching between the tabs should not affect the map
- [x] Switching between the tabs should preserve scroll-state / sub-tab selection
- [x] ~~Switching to Monitor should set the URL to be `/monitor`, and Analyze should set it to `/analyze`.~~
- [x] Both tabs should show the "Change area" and "Select a model" buttons below. Alternatively, the buttons should be outside the tabs so they are not re-rendered on tab switching.
- [x] When coming from Draw, the first tab shown on this page should be Analyze
- [x] Update styling so that the area of interest name and area is now on a white background, and the new tab bar is on the grey background

rajadain:
I have a very rudimentary setup with both working tabs ready:

![2018-01-29 17 57 22](https://user-images.githubusercontent.com/1430060/35539110-460deb4a-051e-11e8-9371-f2bed4716273.gif)

Still need to work out the styling and map interaction. It is possible that the desired styling for this tab is different enough from what currently exists in BiG-CZ that we may want to make a new template for MMW use rather than change BiG-CZ's.

Issue #2621, opened 2018-01-24
Add Monitor Tab to Modeling Header


rajadain:
Following #2620, add a "Monitor" tab to the modeling header, as shown in these mockups:

![image](https://user-images.githubusercontent.com/1430060/35341211-1d52d406-00f3-11e8-9dec-390b053910fb.png)

![image](https://user-images.githubusercontent.com/1430060/35341218-23ee6550-00f3-11e8-9a1f-083e50c1fed0.png)

![image](https://user-images.githubusercontent.com/1430060/35341229-29d61d64-00f3-11e8-8658-b73b6e6bb185.png)

PR #2632, opened 2018-01-31
Add Monitor Tab

rajadain:
## Overview

Adds Monitor tab to the Analyze stage. Updates styling to match the mockup. The Monitor tab currently has no behavior, which will be added in #2622, mostly by reusing the models and copying a lot of the views from BiG-CZ.

Connects #2620 

### Notes

As part of an earlier attempt I had the full search available in the monitor tab (see https://github.com/WikiWatershed/model-my-watershed/issues/2620#issuecomment-361416979), I ultimately decided against using the BiG-CZ views as is for this. This decision is motivated by two constraints:

1. Existing BiG-CZ behavior should not change at all
2. Monitor tab should fit in well with MMW

There are things we'll need to do for 2 which would contradict 1. For example, in BiG-CZ when switching between the Data and Analyze tabs the map state does not change. Whereas in MMW it will. Another issue is that much of BiG-CZ's search UI is absolutely positioned, which does not readily fit in with the altered sizes in MMW. We should create a UI using flexbox and other more modern methods so that the contents fit and work nicely.

Thus, I created a `/monitor` package in the JavaScript source folder, to be populated with views and templates used in the monitor tab in MMW, and will leave BiG-CZ untouched.

## Testing Instructions

* Go to [:8000/](http://localhost:8000/) and draw / select a shape
* Proceed to Analyze. Ensure you see two tabs above, "Analyze" and "Monitor"
* Switch between the tabs and ensure that scroll state is preserved
* Ensure the styling matches the mockups in the original issue
* Ensure that all the specifications in the original issue have been satisfied

### DIFF
--- a/src/mmw/sass/pages/_monitor.scss
+++ b/src/mmw/sass/pages/_monitor.scss
@@ -1,0 +1,34 @@
+.monitor-window {
+  padding: 1rem;
+
+  .monitor-search {
+    display: flex;
+
+    button {
+      border-radius: 2px;
+    }
+
+    .search-group {
+      display: flex;
+      flex: 3;
+      padding-right: 5px;
+
+      input[type="text"] {
+        border-radius: 0;
+        box-shadow: none;
+        flex: 1;
+        height: auto;
+      }
+
+      .btn-search {
+        border-bottom-left-radius: 0;
+        border-top-left-radius: 0;
+        margin-left: -1px;
+      }
+    }
+  }
+
+  .intro-text {
+    margin-top: 15px;
+  }
+}
--- a/src/mmw/js/src/analyze/templates/resultsWindow.html
+++ b/src/mmw/js/src/analyze/templates/resultsWindow.html
@@ -1,8 +1,31 @@
+<div class="sidebar-tab-header output-tabs-wrapper" role="tabpanel">
+    <ul class="nav nav-tabs" role="tablist">
+        <li class="active" role="presentation">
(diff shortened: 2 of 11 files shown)


rajadain:
Talked to @mmcfarland about this a little bit. The conclusion from that conversation was:

* We are not under strict orders to "not change" BiG-CZ
* We should aim to be efficient and not introduce duplications that will be hard to maintain

Given this, I think it will be reasonable to reuse as many of the data_catalog views as possible (such as search card, search details, etc), and just style them differently. The parent view will have to be different to accommodate the new tab header and such, but all the inner views should be reused.

arottersman:
Taking a look now

src/mmw/sass/pages/_monitor.scss
```diff
@@ -0,0 +1,34 @@
+.monitor-window {
+  padding: 15px;
```
arottersman: Consider setting this to `1rem` so it lines up with the AOI title
screen shot 2018-01-31 at 4 12 24 pm

rajadain:
> Should this view also appear in Model > Analyze tab mode?

Yes, to be addressed in #2621.

jfrankl:
Looks good!

rajadain:
Thank you for taking a look! Amended the padding. Will merge when green.

rajadain:
Jenkins is failing all builds for unrelated reasons (that are being worked on separately). Going to merge this in so that downstream work is unblocked.

Issue #2541, opened 2017-11-30
Improve subbasin modeling performance


arottersman:
@caseycesari, @rajadain and I chatted briefly on how to improve the performance of subbasin modeling (see #2537)

Some ideas that came up:

- One of the slowest pieces in our geoprocessing pipeline is the network cost of pulling down tiles. Some of this effort is wasted in the case of subbasin modeling because some huc12s will pull down the same tile
  - We could add a new geoprocessing job that takes a list of shapes (the huc12s) so we could pull down the tiles once for all of them
  - We could try pyramiding the tiles to decrease the number we have to pull
- Currently the cache happens at the geoprocessing job level, of which mapshed is composed of several. If there's a lot of overhead in submitting the many tasks to get the cached jobs, it may help to create a higher level cache, ie store the full `/mapshed` results for a huc12 under a single key. (The same could be done for climate analysis)
 - Related to the above we could store all `/mapshed` output in the database. There are ~86000 huc12s

arottersman:
See @rajadain's comments https://github.com/WikiWatershed/model-my-watershed/pull/2537#pullrequestreview-81623057 and https://github.com/WikiWatershed/model-my-watershed/pull/2537#issuecomment-349404631 for some initial investigation into speeding up the query that splits the shape into its component huc12s. While this isn't the bulk of the time spent on the model, it can still be several secs of work.

One possibility is to add a `huc12_huc` sort of table to lookup the larger hucs a `huc12` is part of (or these could be columns on `boundary_huc12`)

arottersman:
Met with GT team and have decided to speed things up by sending the geop pipeline a set of shapes and a set of operations instead our current pipeline which creates one job per operation+aoi combo. This will decrease the number of duplicate tiles fetched (tile-fetching is the biggest geop bottleneck). We briefly discussed pre-computating HUC-12s instead, but determined that updating our geoprocessing code will improve performance for more than just sub-basin mapshed jobs, and only involve minimal changes.

@rajadain will follow up with a formal ADR on the decision and the scope of the changes involved.

PR #2633, opened 2018-02-01
Add ADR for Multi-Operation Geoprocessing

rajadain:
## Overview

Captures discussion with GeoTrellis team on approach to optimize subbasin performance. Tagging @echeipesh and @moradology for review.

Connects #2541 

### Notes

This doesn't follow the classical ADR structure of considering multiple options. Instead this focuses on the chosen option and outlines its proposed architecture.

## Testing Instructions

* Read the ADR and ensure it makes sense, and doesn't miss out on any part of the discussion

### DIFF
--- a/doc/arch/adr-006-geoprocessing-multioperation.md
+++ b/doc/arch/adr-006-geoprocessing-multioperation.md
@@ -1,0 +1,140 @@
+# Multi-Operation Geoprocessing Operation
+
+## Context
+
+### Current Operations
+
+Currently the geoprocessing service provides this set of operations:
+
+  * [`rasterGroupedCount`][github 1]
+  * [`rasterGroupedAverage`][github 2]
+  * [`rasterAverage`][github 3]
+  * [`rasterLinesJoin`][github 4]
+  * [`rasterSummary`][github 5]
+
+Each operation takes a multipolygon and a list of rasters, queries those
+rasters cropped to the multipolygon, and performs the respective calculation,
+and returns the result grouped by the given rasters.
+
+These operations are then [configured][github 6] and [called][github 7] in
+Model My Watershed (MMW) for different sets of rasters. The results are then
+aggregated in MMW using Python. In some cases, such as MapShed, **each project
+invokes 10 separate calls to the geoprocessing service**, the results of which
+are then aggregated in Python before being sent back to the front-end.
+
+The performance of each of these operations for a shape around the size of a
+HUC-12 (which is slightly smaller than a 512 tile) is about 1s. For a larger
+shape like a HUC-8 which can be 15-50 times larger, the average time is about
+6s. Since the front-end invokes a number of these jobs, and spends time polling
+with Celery, the average wait time for a HUC-8 MapShed job is around 30s. This
+has so far been acceptable.
+
+When a geoprocessing service returns successfully, we cache the results. So if
+another user selects the same shape and the same rasters are queried for it,
+the cached results are used and the overall wait is much shorter.
+
+### Subbasin Modeling
+
+Subbasin Modeling divides a larger HUC-8 into its component HUC-12s to improve
+accuracy of results. Under this new scheme, instead of running a single HUC-8,
+we run MapShed for all component HUC-12s and then aggregate those results. This
+severely affects the runtime, to the point of not being usable. For more
+discussion on this methodology and affect on runtime, please see
+[#2537][github 8].
+
+Much of this additional runtime comes from processing the same tiles for
+multiple HUC-12s. For example, consider [the following case][ocks]:
+
+![image][githubusercontent]
+
+The red shape intersects five tiles, and the purple shape intersects only one
+tile, one of red's five. When querying the red shape we fetch all five tiles,
+and when querying the purple shape we fetch that tile again. This kind of
+duplication happens all across a HUC-8, since it is querying a number of
+adjacent HUC-12s.
+
+Furthermore, it is our understanding that fetching tiles from S3 is the largest
+portion of a single request's execution time. Thus, removing its duplication is
+the greatest speedup we could achieve in a single step.
+
+## Proposed Solution
+
+To optimize and reuse tile fetches, the geoprocessing service needs to be aware
+that they will be reused. Thus, rather than performing an operation for a
+single shape, the geoprocessing service should take a set of shapes. In this
+set, one should be the parent shape, and all the rest should be the component
+shapes. The parent shape will be used to fetch tiles from S3 and store them in
+memory. Then, the operation will be performed on every component shape,
+choosing the intersecting subset of tiles already in memory. Instead of the
+output being a single set of results, it will be a list of sets of results, in
+the same order as the incoming component shapes.
+
+The above can be done either to the existing operations, and adapting their
+output in MMW to account for the new list return format. Previous operations
+can still be supported with just a singleton list as the output. Or the above
+can be implemented as new, "multishape" versions of the same operations.
+
+Furthermore, since a number of the geoprocessing calls invoked by MapShed query
+the same rasters, e.g. [n][github 9][l][github 10][c][github 11][d][github 12],
+including them in them in the same geoprocessing call would be a speed up as
+well. This would require not only sending a set of shapes, but a set of
+operations as well, each with an `operationType` and a list of `rasters`. The
+output should be a list of lists of sets of results, where the first list
+matches the list of component shapes, and the second the list of operations.
+
+Since both the input and output of this operation can be quite different than
+the current implementation, it should be added as a new operation type rather
+than a modification to an existing one. MapShed should be updated to use this
+new format. The new operation should also support submitting a single shape
+instead of a list of shapes, i.e. the use case for a single HUC-12. In this
+case the output should be a singleton list of list of results.
+
+## Decision
+
+We will add a new operation to the geoprocessing service that takes a list of
+shapes and a list of operations. The first shape in the list is assumed to be
+the parent shape that contains all the other child shapes. Using the parent
+shape we pull tiles corresponding to every raster in the list of operations,
+and collect them in a [`CompositeTile`][github 13] per raster.
+
+Then, for each child shape, we go through the list of operations and calculate
+the results using the respective `CompositeTile`s from above. All the results
+are then collected in order and returned as a list.
+
+Once this ADR is approved, the next steps will be:
+
+  1. Define input and outputs for new geoprocessing operation
+  2. Implement new geoprocessing operation
+  3. Benchmark the new operation against current ones
+  4. Make a new release of the geoprocessing service and publish it
+  5. Update MMW to use the new geoprocessing service
+  6. Update MapShed to use the new operation
+  7. Update Subbasin code to use the new operation
+
+## Consequences
+
+This change is not limited to the geoprocessing service, so MMW will have to be
+updated accordingly.
+
+The underlying wisdom seems to be that in order to gain efficiency, we must
+move more of the execution logic from Python to Scala. This efficiency could
+come at a loss of flexibility that we have had so far, but can be avoided with
+the right design of input format. Furthermore, the existing operations will
+continue to function, and will be available for other use cases, e.g. other
+Analyze tasks.
+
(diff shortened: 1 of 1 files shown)


doc/arch/adr-006-geoprocessing-multioperation.md
```diff
@@ -0,0 +1,65 @@
...
+
+Currently the geoprocessing service provides this set of operations:
+
+  * [`rasterGroupedCount`](https://github.com/WikiWatershed/mmw-geoprocessing/blob/c212f29e0b090dd9624b8f637e59fdf6cbbf4f4c/api/src/main/scala/Geoprocessing.scala#L240)
+  * [`rasterGroupedAverage`](https://github.com/WikiWatershed/mmw-geoprocessing/blob/c212f29e0b090dd9624b8f637e59fdf6cbbf4f4c/api/src/main/scala/Geoprocessing.scala#L192)
+  * [`rasterAverage`](https://github.com/WikiWatershed/mmw-geoprocessing/blob/c212f29e0b090dd9624b8f637e59fdf6cbbf4f4c/api/src/main/scala/Geoprocessing.scala#L149)
+  * [`rasterLinesJoin`](https://github.com/WikiWatershed/mmw-geoprocessing/blob/c212f29e0b090dd9624b8f637e59fdf6cbbf4f4c/api/src/main/scala/Geoprocessing.scala#L114)
+  * [`rasterSummary`](https://github.com/WikiWatershed/mmw-geoprocessing/blob/c212f29e0b090dd9624b8f637e59fdf6cbbf4f4c/api/src/main/scala/Geoprocessing.scala#L281)
+
+Each operation takes a multipolygon and a list of rasters, queries those rasters cropped to the multipolygon, and performs the respective calculation, and returns the result grouped by the given rasters.
```
kellyi: This may be a personal preference but consider adding line breaks at column 80 for each line.

doc/arch/adr-006-geoprocessing-multioperation.md
```diff
@@ -0,0 +1,65 @@
...
+
+Since both the input and output of this operation can be quite different than the current implementation, it should be added as a new operation type rather than a modification to an existing one. MapShed should be updated to use this new format. The new operation should also support submitting a single shape instead of a list of shapes, i.e. the use case for a single HUC-12. In this case the output should be a singleton list of list of results.
+
+## Decision
+
+We will make add a new operation to the geoprocessing service that takes a list of shapes and a list of operations. The first shape in the list is assumed to be the parent shape that contains all the other child shapes. Using the parent shape we pull tiles corresponding to every raster in the list of operations, and collect them in a [`CompositeTile`](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/CompositeTile.scala) per raster.
```
kellyi: Typo: `make add` -> `add`

doc/arch/adr-006-geoprocessing-multioperation.md
```diff
@@ -0,0 +1,65 @@
...
+Then, for each child shape, we go through the list of operations and calculate the results using the respective `CompositeTile`s from above. All the results are then collected in order and returned as a list.
+
+Once this ADR is approved, the next steps will be:
+
+  1. Define input and outputs for new geoprocessing operation
+  2. Implement new geoprocessing operation
+  3. Benchmark the new operation against current ones
+  4. Make a new release of the geoprocessing service and publish it
+  5. Update MMW to use the new geoprocessing service
+  6. Update MapShed to use the new operation
```
mmcfarland: It's unclear what the proposed scope of this task is aiming for.  You mention both that _general_ geoprocessing tasks could be modified and combined to reuse tiles from within the same job in addition to making this change to support the disaggregation needed for the new subbasin routines.  

If this step is meant to include a wider overhaul of the way Mapshed generally requests & receives tasks (not just related to times when subbasin is enabled) I think it'd be good to make that an explicit decision so it can be discussed apart from the approach recommended for subbasin.

rajadain:
Reformatted to make the source easier to read. Fixed typo.

doc/arch/adr-006-geoprocessing-multioperation.md
```diff
@@ -0,0 +1,65 @@
...
+  7. Update Subbasin code to use the new operation
+
+## Consequences
```
mmcfarland: It seems like an additional consequence would be our ability to create and use cached results at the huc12 level when subbasin is on.  Could you add a description of the impact on our use of caching?  I think we could either make the geop service aware of the cache instance, or check and populate cache results from celery before submitting and after getting the results back from the service.

rajadain: Hmm good point. I think we should keep the caching at the Celery level. Here's an outline of the steps I'm imagining:

1. `[django]` Receive request, start Celery job
2. `[celery]` Enumerate all the shapes and operations needed for a subbasin MapShed run
3. `[celery]` Check the cache for their keys, and make a list of all that are not available. Invoke geoprocessing.
4. `[geoprocessing]` Execute the new multi-operation geoprocessing for the ones not available
5. `[celery]` Once the results come back, cache them with the right key
6. `[celery]` Get all results out of the cache and return

I'll add this to the Consequences section.

doc/arch/adr-006-geoprocessing-multioperation.md
rajadain: Yes this should be clearer. My understanding of subbasin is as follows:

1. `[front-end]` Sends special Subbasin modeling request (based on either app flag or user setting)
2. `[django]` Receives parent WKAoI. Finds children WKAoIs.
3. `[celery | geoprocessing]` Runs MapShed on each child WKAoI.
4. `[celery]` Compiles results together. Returns to front-end.

Essentially, Subbasin is a wrapper around MapShed, not a completely different thing. In order to support multiple operation optimizations, we'll have to amend how MapShed itself behaves. Since this is simply an organizational difference in how and when the operations are specified and executed, it will not change their final computed values.

The existing geoprocessing operations can be used for Analyze tasks.

doc/arch/adr-006-geoprocessing-multioperation.md
rajadain: Added in 3088461e.

mmcfarland:
> GWLF-E: 27s and 36s

This makes sense, especially if there's a single celery worker thread on your dev box.  The gwlfe jobs are effectively being executed sequentially.  Aside from improving the execution speed of the model code (which Scott took a look at but so no easy wins), adding additional executors would be potential positive step.  Staging should at least have 4 (2 per machine), so that might be a meaningful difference to your dev.

arottersman:
> especially if there's a single celery worker thread on your dev box. The gwlfe jobs are effectively being executed sequentially.

Still working on the staging benchmark, but I think those local timings were taken with a worker with 2 CPUS + 2 celery threads. It seemed like both threads were working at the time.

arottersman:
Okay, I guess GWLF-E doesn't really need to be addressed here if optimizations have already been looked into and adding executors is always available as an code-change-free improvement. Here's what we're dealing with in terms of the performance hit to it:

```
--> subbasin
gwlfe started:  2018-02-02T15:04:01.593Z
gwlfe finished:  2018-02-02T15:04:16.164Z
--------
~14.5 sec

--> not-subbasin
gwlfe started:  2018-02-02T15:11:41.967Z
gwlfe finished:  2018-02-02T15:11:45.252Z
--------
~3.3 sec
```

So on staging, subbasin takes a little more than 4x as long as the current setup.

<details>
<summary>
script
</summary>

```sh
URL=https://staging.app.wikiwatershed.org
# URL=http://localhost:8000
SUBBASIN=true
# SUBBASIN=false

JOB_ID=$(curl -X POST \
  -F 'mapshed_input={"wkaoi": "huc10__10017"}' \
  "$URL/mmw/modeling/mapshed/?subbasin=$SUBBASIN" | jq -r .job)

STATUS="started"
while [ "$STATUS" == "started" ]
do
    echo polling...
    JOB=$(curl -H "Content-Type: application/json" \
                 "$URL/mmw/modeling/jobs/$JOB_ID/")
    STATUS=$(echo $JOB | jq -r .status)
    echo job status $STATUS
done
echo "mapshed error: " $(echo $JOB | jq -r .error)
echo "mapshed started: " $(echo $JOB | jq -r .started)
echo "mapshed finished: " $(echo $JOB | jq -r .finished)

MAPSHED_DATA=$(echo $JOB | jq -r .result)

echo "model_input=$MAPSHED_DATA" > "./mapshed_data_$JOB_ID"

JOB_ID=$(curl -X POST \
        -d "@mapshed_data_$JOB_ID" \
  -d inputmod_hash=d751713988987e9331980363e24189ced751713988987e9331980363e24189ce \
  "$URL/mmw/modeling/gwlfe/?subbasin=$SUBBASIN" | jq -r .job)

echo $JOB_ID

STATUS="started"
while [ "$STATUS" == "started" ]
do
    echo polling...
    JOB=$(curl -H "Content-Type: application/json" \
                 "$URL/mmw/modeling/jobs/$JOB_ID/")
    STATUS=$(echo $JOB | jq -r .status)
    echo job status $STATUS
done

echo "gwlfe error: " $(echo $JOB | jq -r .error)
echo "gwlfe started: " $(echo $JOB | jq -r .started)
echo "gwlfe finished: " $(echo $JOB | jq -r .finished)
```

</details>

rajadain:
> So subbasin takes a little more than 4x as long as the current setup.

How many HUC-12s are in this HUC-8? If it's more than 4, then we're scaling sub-linearly which is good.

arottersman:
> How many HUC-12s are in this HUC-8? If it's more than 4, then we're scaling sub-linearly which is good.

This is a HUC-10 with 8 HUC-12s (Lower Deep River), so yes, sub-linearly.

rajadain:
Okay, so it is not sub-linear, but linear growth. 4x the time for 8x the shapes because we're processing 2 shapes at once. Adding more workers can help alleviate that, but the order of growth is linear.

doc/arch/adr-006-geoprocessing-multioperation.md
rajadain: Clarification added in f9e14757b38d2020de1b1627e7f0accfda2c0866.

arottersman:
Oh, I think it's worse then. There should be 4 shapes processing at once on staging (if all executors are available).

doc/arch/adr-006-geoprocessing-multioperation.md
```diff
@@ -0,0 +1,167 @@
...
+allow regular MapShed to use the same operation, rather than have a competing
+implementation for Subbasin modeling.
+
+## Decision
+
+We will add a new operation to the geoprocessing service that takes a list of
+shapes and a list of operations. The first shape in the list is assumed to be
+the parent shape that contains all the other child shapes. Using the parent
+shape we pull tiles corresponding to every raster in the list of operations,
+and collect them in a [`CompositeTile`][github 13] per raster.
```
echeipesh: Let me call out our discussion here:

`CompositeTile` essentially provides `Seq[(SpatialKey, Tile)] => Tile`. This would make it easier to express operations over the area because you could treat the query result as a single raster without worrying about tile boundaries and the associated code of iterating over tiles.

However as @rajadain pointed out the current code is already implemented as [operation over `Seq(SpatialKey, Tile)](https://github.com/WikiWatershed/mmw-geoprocessing/blob/c212f29e0b090dd9624b8f637e59fdf6cbbf4f4c/api/src/main/scala/Geoprocessing.scala#L253-L262) so its unlikely that this refactor would be a benefit in this task.

Additionally the task [parallelizes](https://github.com/WikiWatershed/mmw-geoprocessing/blob/c212f29e0b090dd9624b8f637e59fdf6cbbf4f4c/api/src/main/scala/Geoprocessing.scala#L252) over each individual tile. So converting to `CompositeTile` would remove that level of parallelism. This is something to consider but it likely won't have noticeable effect on performance because the proposed changes could/should parallelize over operations per query geometry.

TLDR: Introducing `CompositeTile` usage would simplify the code and likely offset the complexity introduced by handling multiple areas and operations per request but isn't strictly necessary.

rajadain: Alright I'll add a note that makes `ComposeTile` refactor optional.

rajadain:
Alright ready for one more look!

Issue #2596, opened 2018-01-09
Center UI


jfrankl:
The Stroud logo and Analyze/Model buttons should be centered on the page. Currently, neither is centered.

![image](https://user-images.githubusercontent.com/1809908/34744852-96088622-f55c-11e7-96e5-fdcd6a39e071.png)

PR #2634, opened 2018-02-01
Center header

jfrankl:
## Overview

This PR centers the primary navigation on the project view. Previously, the primary navigation was sometimes centered, but would become uncentered when other elements on the page changed. Now, it should not be dependent on other elements.

The Stroud logo is not centered. That will be removed in a later PR, so I'm ignoring that for now.

Connects #2596

### Demo

image

## Testing Instructions

 * Rebuild CSS
 * Go to project view
 * Make sure the Analyze/Model buttons are centered. Try to change the name of the project. The centered buttons should not move as a result.


rajadain:
Taking a look now.

Issue #2638, opened 2018-02-02
Make Project Public on HydroShare Export


rajadain:
When a user exports a project to HydroShare, it should also be made public to the general population, so that when someone clicks a link from HydroShare to the project they can view it.

Optionally, consider disabling the "link sharing" switch and locking it in the "on" position with some explanatory text to ensure someone doesn't set a HydroShared project to Private.

PR #2658, opened 2018-02-09
Mandatorily Make Projects Public After HydroShare Export

rajadain:
## Overview

When a project is exported to HydroShare, there may be a link from HydroShare back to the project in MMW. To allow this, we need to make the project public in MMW so that any incoming user can see it.

After a successful export to HydroShare, we update the project's privacy and make it unchangeable from the MultiShareView. We also add a message explaining that any project exported to HydroShare will always have link sharing enabled.

Connects #2638 

### Demo

Once a project has been exported to HydroShare, the modal looks like this:

image

For a project not exported to HydroShare, the link sharing can be toggled independently:

image

Turning HydroShare off leaves Link Sharing on, but can be toggled independently by the user.

Tagging @ajrobbins and @jfrankl for visual review.

## Testing Instructions

* Ensure your VM is setup to work with HydroShare. This essentially involves `vagrant ssh app -c 'sudo vim /etc/mmw.d/env/MMW_HYDROSHARE_SECRET_KEY'` and entering the secret key for HydroShare Beta from LastPass. Restart `mmw-app` after changing that file.
* Go to [:8000/](http://localhost:8000/) and create a project
* Export it to HydroShare. Ensure that if the export fails or is canceled, link sharing remains off.
* Ensure that a successful export turns link sharing off, and that you are unable to turn it off.
* Ensure that removing a project from HydroShare leaves link sharing on, but can be turned off by the user.

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -464,8 +464,12 @@ var ProjectModel = Backbone.Model.extend({
                 mapshed_data: mapshedData,
             }, payload))
         }).done(function(result) {
-            self.set('hydroshare', result);
-            self.set('hydroshare_errors', []);
+            self.set({
+                hydroshare: result,
+                hydroshare_errors: [],
+                is_private: false,
+            });
+            self.saveProjectAndScenarios();
         }).fail(function(result) {
             if (result.responseJSON && result.responseJSON.errors) {
                 self.set('hydroshare_errors', result.responseJSON.errors);
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -467,9 +467,12 @@ var ProjectModel = Backbone.Model.extend({
             self.set({
                 hydroshare: result,
                 hydroshare_errors: [],
+                // Exporting to HydroShare make projects public
+                // in apps.export.views.hydroshare. We manually
+                // make the switch here rather than fetching it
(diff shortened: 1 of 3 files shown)


ajrobbins:
LGTM visually!

arottersman:
Taking a look!

src/mmw/js/src/modeling/models.js
```diff
@@ -464,8 +464,11 @@ var ProjectModel = Backbone.Model.extend({
                 mapshed_data: mapshedData,
             }, payload))
         }).done(function(result) {
-            self.set('hydroshare', result);
-            self.set('hydroshare_errors', []);
+            self.set({
+                hydroshare: result,
+                hydroshare_errors: [],
+                is_private: false,
```
arottersman: This is a little unconventional in its assumption that the server is also updating this field, but I understand that receiving the updated project from the server to kick off this model change might require some refactoring that it's maybe not worth investing in now.

rajadain: Agreed. While a refactoring would be expensive, I will add a comment for clarity.

rajadain:
Added a comment to the effect, explaining the behavior pointed out in https://github.com/WikiWatershed/model-my-watershed/pull/2658#discussion_r168302959. Will merge once green.

Thanks for taking a look! ⚖️

jfrankl:
Looks good

Issue #2648, opened 2018-02-08
Flip order of MapShed Water Quality Total Loads and Load Tallies Tables


arottersman:
Put the Totals table on top instead of under the load tally table.

![screen shot 2018-02-08 at 8 47 21 am](https://user-images.githubusercontent.com/7633670/35978649-5082924a-0cb4-11e8-8c41-dcdc5e7d3088.png)

arottersman:
Shares a point with #2643

Issue #2643, opened 2018-02-06
Reorder Analysis tabs & shorten titles to address overflow


aufdenkampe:
@ajrobbins & @jfrankl,

I think we want to re-order our Analysis tabs, which are presently on staging as:
![screen shot 2018-02-06 at 3 26 41 pm](https://user-images.githubusercontent.com/5166036/35885025-3142c5ca-0b52-11e8-8280-ab90f1a2b759.png)

As discussed in our [2017-11-28 Bi-Weekly MMW2 Tech Meeting](https://docs.google.com/document/d/1mG6m6wxxJjoSxK4SyYrNRaUTHZujSHYz5B7aXmCJWsQ/edit#bookmark=kix.iaknd5xpowhi), we would like to change the order of the tabs. I'm suggesting the following:

Streams, Land, Soil, Terrain, Climate, Point Sources, Animals, Water Quality

To address the tab overflow issue (first raised with #1525), I'm wondering if we might shorten some of these to:

Streams, Land, Soil, Terrain, Climate, Pt Sources, Animals, Water Qual

Water Qual will go away this spring, but will be public for at least one release cycle.

However, we eventually may be adding one more tab: Area (or something like it). This tab would show stats on the boundaries that overlap with your Area of Interest, with ability to show related layers and interactivity on the map:
* HUC8, HUC10, HUC12
* County
* School District
* Congressional District
We first discussed this in our [2017-09-19 Bi-Weekly MMW2 Tech Meeting](https://docs.google.com/document/d/1mG6m6wxxJjoSxK4SyYrNRaUTHZujSHYz5B7aXmCJWsQ/edit#bookmark=kix.uw054qp1stwd), with the idea that it might be useful for the API, but that we would need to hold off for now on prioritizing this effort. 

@SRGDamia1, can you forward to Arscott and Kerlin for their input?

PR #2659, opened 2018-02-12
Analysis & Model Reorders

mmcfarland:
## Overview

Reorder and renames the Analysis tab labels to better fit on a single line.  Also swaps the positions of the Multi-Year WQ output, based on client preferences. 

A small UI defect is introduced here, where the Water Qual tab is pushed down slightly by the "download AoI" button.  A brief attempt to correct didn't yield results, so I'm deferring in the hopes of better efficiency through a UI cleanup card.

Connects #2648 
Connects #2643 

### Demo

**New Tabs**
image

**New Positions**
image

### Notes

#2643 references a proposed new tab that was not considered here.  This will come about as a new issue when and if it is prioritized.

## Testing Instructions

 * Build and Bundle
 * In MMW mode, ensure the names and order of the tabs match those specified in the issue
 * In BigCZ mode, ensure that the analyze output is still restricted to Land, Soil, Climate - in that order
 * Create a new Mapshed project and confirm that summary table is on top, and that the download csv buttons still correspond to the table they are underneath

### DIFF
--- a/src/mmw/js/src/analyze/models.js
+++ b/src/mmw/js/src/analyze/models.js
@@ -141,7 +141,7 @@ function createAnalyzeTaskCollection(aoi, wkaoi) {
         },
         {
             name: "pointsource",
-            displayName: "Point Sources",
+            displayName: "Pt Sources",
             area_of_interest: aoi,
             wkaoi: wkaoi,
             taskName: "analyze/pointsource"
@@ -155,7 +155,7 @@ function createAnalyzeTaskCollection(aoi, wkaoi) {
         },
         {
             name: "catchment_water_quality",
-            displayName: "Water Quality",
+            displayName: "Water Qual",
             area_of_interest: aoi,
             wkaoi: wkaoi,
             taskName: "analyze/catchment-water-quality"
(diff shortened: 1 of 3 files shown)


rajadain:
Unfortunately the rename seems to have not been enough to get the tabs on one row (tested in Chrome and Firefox on macOS):

image

Maybe we need to reduce the gaps in between tab headers as well.

mmcfarland:
Interesting, Chrome & FF & IE on Linux & Windows all fit on the line.  I'll check browserstack for mac and see if it duplicates.

mmcfarland:
Sure enough, Mac/Safari renders it differently and still overflows.  I'll add the quirk to #2660.

rajadain:
I think we can probably do it here: 

```diff
diff --git a/src/mmw/sass/components/_tabs.scss b/src/mmw/sass/components/_tabs.scss
index aa97e853..de07efed 100644
--- a/src/mmw/sass/components/_tabs.scss
+++ b/src/mmw/sass/components/_tabs.scss
@@ -93,6 +93,7 @@
     }
 
     li > a {
+      margin-right: 0px;
       color: $ui-secondary;
       padding: 0 0.4rem;
       line-height: 42px;
```

This seems to take care of it.

mmcfarland:
Added 2c0ec03 for the mac fix as suggested by @rajadain

Issue #2558, opened 2017-12-14
Add Ability to Open MMW Project From HydroShare


rajadain:
HydroShare has some "blessed" apps whose names and logos are shown in a dropdown in the top right corner of a resource's detail page. These apps differ by resource.

_TODO_ Insert screenshot when HydroShare is back up

Investigate how MMW can be added to that menu (this may involve collaborating with HydroShare folks) and ensure that clicking it opens the correct project link.

rajadain:
Blocked awaiting documentation for this from HydroShare folks /cc @ajrobbins

ajrobbins:
Just emailed requesting documentation.

ajrobbins:
HydroShare folks devised a proposed integration plan here: https://docs.google.com/document/d/1CjctGQRIdAC0OxY2S6MhpN7ETaEWSmbESMfNdVzujfA/edit?ts=5a87092b#

PR #2664, opened 2018-02-16
HydroShare Export: Allow Open With Behavior

rajadain:
## Overview

To indicate to HydroShare that a given resource can be opened in MMW, add extra metadata to that effect. HydroShare currently does not have this functionality implemented, but we add the metadata all the same in the interim so that when HydroShare is ready it opens it in MMW.

Connects #2558 

### Demo

image

### Notes

HydroShare may ask us to use a different key than `webapp`. I'll update the PR in that case.

## Testing Instructions

* Check out this branch, and go to [:8000/](http://localhost:8000/)
* Export a project to HydroShare
* Go to the resource in HydroShare, scroll down, inspect the Extra Metadata and ensure it has an entry with `webapp` as the key and `model-my-watershed` as the value


rajadain:
Thanks for taking a look! ⚖️

Issue #2623, opened 2018-01-24
Splash Page and Header Updates


rajadain:
Mockup: 

![image](https://user-images.githubusercontent.com/1430060/35345051-9ba98aee-00fc-11e8-8915-de224871e782.png)

- [x] Update Step 2 to read "Analyze & Monitor"
- [x] Remove "Stroud" Logo from the center
- [x] Replace "Model My Watershed" text with "WikiWatershed" logo. Clicking it should behave identically to now.
- [x] Add "Projects" button to the top right
  - [x] Clicking this should take the user to their "My Projects" page
  - [x] If the user is not logged in, it should open the login modal
  - [x] Remove the "My Projects" option from the user dropdown in the top right

rajadain:
Maybe we should consider making the WikiWatershed logo in the top left not-clickable. I remember it not being clickable originally, and us adding it on purpose, but given this new design:

![image](https://user-images.githubusercontent.com/1430060/36263471-4f7a057a-1238-11e8-86b2-8d0129fbeeff.png)

users can accidentally click the title and go home when in fact they mean to click Analyze or Monitor, losing their work. I've done that at least once.

Thoughts @ajrobbins @jfrankl?

ajrobbins:
I think it's a common pattern to have a home/webpage name link in the top left, and users are used to that. Though I could see now a mistake could be make, I think the new logo there (as opposed to more unstyled text_ will hopefully make that less likely. So, I'm in favor of keeping it as a link.

PR #2665, opened 2018-02-19
Updates to Splash and Header text

mmcfarland:
## Overview

Makes small changes to the text and links in the splash an header bar in MMW mode.

Connects #2623 

### Notes

The main implementation detail noted in conversation is the unusual import used in 2ac1f9f, which allows a cheap escape hatch from the initialization call stack which would otherwise prevent an `App` instance to be available.  I tried various refactors, but I think a comprehensive solution is outside this scope and perhaps outside the remaining project budget.

Additionally:
* The word "Monitor" was added to Step 2, but the description was not updated as per the issue.  The language as it exists may not be sufficient to explain the Monitor concept @ajrobbins 
* Previously, in BigCZ mode, one could get to My Projects which would then awkwardly convert the app to MMW mode without updating any title elements.  I decided this was a bug and it is no longer possible to get to the Project listing in BigCZ mode from the user or header links.

## Testing Instructions

A new image has been added, use `./scripts/bundle.sh --vendor` in addition to your normal bundling workflow.

MMW:
* Ensure the text & images appear as described on the splash screen and header
* While logged out, use the Projects header link to be prompted then redirected to the Project listing
* While logged in, do the same and do not get prompted
* There is no longer a Project link in the user drop down

BigCZ:
* Ensure the splash & header text remains unchanged
* The user drop down nor the header include a link to My Projects


rajadain:
@azavea-bot rebuild

rajadain:
> I decided this was a bug and it is no longer possible to get to the Project listing in BigCZ mode from the user or header links.

👌

rajadain:
Also, the CI reported a front-end test failing. The tests are running correctly on my local, so I'm retriggering the build in case it's a fluke on CI.

mmcfarland:
Good catch.  I added 16695e9 to make the logo re-route to home.  I also included 0224ac2 with comments around the `require` that didn't make it into the PR because of an oversight.

Tests pass locally, I'll keep an eye on CI.

rajadain:
Interestingly, we import `app` inline in that file just a few lines above too:

https://github.com/WikiWatershed/model-my-watershed/blob/a415b619d3daaeb9ebe150aff085e5a0a84c825e/src/mmw/js/src/core/views.js#L149-L153

But I think it's the right thing to do for now.

Issue #2670, opened 2018-02-20
Non-project owner opening project via HS should not be able to edit in MMW


ajrobbins:
Scenario:
- MMW user creates project and exports to HS
- Another HS user (user 2) opens the resource in MMW (not the project owner)
- Project is successfully opened in MMW

Currently, user 2 is able to edit the project as if they were the project owner. Desired behavior: that project is locked for editing. If the user wants to edit, they need to make a copy of the project to their own account and edit that.

For this card, we should just lock the project to editing, as if they had opened another user's shared MMW project when they are not the owner.

arottersman:
@jwalgran dug into this a bit for #2671: the underlying bug is any public project with only a Current Conditions scenario still has the "Add changes to this area" button. If you click this button a new scenario is created that you're able to edit. The edits are not persisted anywhere, so it's not so bad, but it is pretty misleading.

PR #2674, opened 2018-02-22
Don't show "Add changes" if project was not created by current user

jwalgran:
## Overview

Users can't save changes to projects they do not own so it is confusing to allow editing.

Connects #2670 

### Demo

screen shot 2018-02-22 at 8 48 28 am

## Testing Instructions

 * Log in, create a project, run TR-55, and share the project to Hydroshare.
 * Refresh the project page and verify that the "Add changes" button is visible.
 * Copy the project URL.
 * Log in as a different user and browse the project page.
 * Verify that the "Add changes" button is not visible.

### DIFF
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -800,14 +800,15 @@ var GwlfeToolbarView = ScenarioModelToolbarView.extend({
 
     templateHelpers: function() {
         var activeMod = this.getActiveMod(),
-            modifications = this.model.get('modifications').toJSON();
+            modifications = this.model.get('modifications').toJSON(),
+            editable = isEditable(this.model);
 
         activeMod = activeMod ? activeMod.toJSON() : null;
-
         return {
             modifications: modifications,
             activeMod: activeMod,
-            displayNames: gwlfeConfig.displayNames
+            displayNames: gwlfeConfig.displayNames,
+            editable: editable
         };
     }
 });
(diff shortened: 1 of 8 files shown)


src/mmw/js/src/modeling/views.js
```diff
@@ -874,13 +874,15 @@ var ScenarioToolbarView = Marionette.CompositeView.extend({
         var gisData = this.currentConditions.getGisData().model_input,
             isGwlfe = this.modelPackage === coreUtils.GWLFE && !_.isEmpty(gisData),
             isOnlyCurrentConditions = this.collection.length === 1 &&
-                this.collection.first().get('is_current_conditions');
+                this.collection.first().get('is_current_conditions'),
+            createdByCurrentUser = this.model.get('user').id === App.user.get('id');
```
rajadain: Prefer reusing the [`isEditable`](https://github.com/WikiWatershed/model-my-watershed/blob/986f51fa7d46eb2ece2a53a8776c5cfb75ca4f77/src/mmw/js/src/modeling/views.js#L1144-L1146) method here, as it is elsewhere in this file.

arottersman:
I noticed the "Remove Modification" buttons remain for both MapShed and TR-55. They are a lot more subtle than the "Add changes to area" button, but I think their removal is in line with this work.

screen shot 2018-02-22 at 11 26 12 am

screen shot 2018-02-22 at 11 40 42 am

jwalgran:
>Prefer reusing the isEditable method

Thanks for the pointer. I added a fixup.

> I noticed the "Remove Modification" buttons remain for both MapShed and TR-55.

Thanks for pointing those out. I'll add a commit to hide them as well.

rajadain:
@azavea-bot rebuild

arottersman:
@jwalgran, is this ready for another look?

jwalgran:
Yes. This is now ready.

Issue #2684, opened 2018-02-26
Modification Dropdown Does Not Re-Open


rajadain:
Clicking "Land Cover" or "Conversation Practice" works the first time, but not subsequently:

![2018-02-26 17 12 18](https://user-images.githubusercontent.com/1430060/36698690-809b0b5c-1b18-11e8-8a12-490ef260240e.gif)

It should work every time.

ajrobbins:
I'm not able to replicate this - was this opened via HydroShare or are there any other special steps to replicate?

rajadain:
Confirmed on staging. Added to sprint.

Issue #2676, opened 2018-02-22
Hide BMPs for Analyze and Monitor tabs


rajadain:
Currently, land cover and conservation practice modifications drawn in the Model tab are visible (and editable) in the Analyze and Monitor tabs:

![2018-02-22 12 36 48](https://user-images.githubusercontent.com/1430060/36554456-53a827d6-17cd-11e8-9480-827f11c1d0ae.gif)

This should not be the case. Make it so that:

- [x] When switching to Monitor or Analyze, any user drawn modifications are hidden and not clickable
- [x] When switching back to Model, any user drawn modifications should be made visible again

One potential idea would be to refactor [`drawPolygon`](https://github.com/WikiWatershed/model-my-watershed/blob/aaede34ee39f158dbdc4c73e8bc7cf0331190f4d/src/mmw/js/src/draw/utils.js#L23) to take an optional layer name, which can be specified in [`modeling/controls.js`](https://github.com/WikiWatershed/model-my-watershed/blob/aaede34ee39f158dbdc4c73e8bc7cf0331190f4d/src/mmw/js/src/modeling/controls.js#L103). Later, that layer name can be used to easily hide and show items. Or remove and re-add them.

PR #2685, opened 2018-02-26
Show drawn modifications only in Modeling tab

rajadain:
## Overview

Hides user-drawn modifications from the Monitor and Analyze tabs, and restores them in the Model tab.

Discovered #2684 during development.

Connects #2676 

## Testing Instructions

 * Check out this branch, and `bundle`
 * Go to [:8000/](http://localhost:8000/) and create a new TR-55 project
 * Once in the modeling stage, add a new scenario and draw some modifications on it
 * Switch to Analyze. Ensure the modifications go away.
 * Switch back to Modeling. Ensure they come back, in the correct order.
 * Switch to Monitor. Ensure they go away. Search for something, wait for results to show up.
 * Switch to Modeling. Ensure search results are hidden, and modifications are shown.
 * Try switching when a map popup is open. Ensure the popup is removed when hiding shapes.
 * Try switching while in Current Conditions. Ensure no shapes are shown when returning to Model tab.
 * Try saving the project to database and refreshing the page. Ensure switching still works correctly.
 * Try a MapShed project to ensure things look good there too.

### DIFF
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -973,6 +973,7 @@ var ResultsView = Marionette.LayoutView.extend({
                 this.monitorRegion.$el.removeClass('active');
                 this.monitorRegion.currentView.setVisibility(false);
                 this.modelingRegion.$el.removeClass('active');
+                App.getMapView().updateModifications(null);
                 break;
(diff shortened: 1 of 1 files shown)


rajadain:
Thanks for taking a look!

Issue #2677, opened 2018-02-22
Axis labels overlapping on soil bar chart


ajrobbins:
Observed on a 13" Macbook Air in firefox:

![screen shot 2018-02-22 at 1 49 39 pm](https://user-images.githubusercontent.com/12868342/36558181-37d1d5b0-17d8-11e8-95ae-365fbadaaf34.png)

PR #2686, opened 2018-02-27
Adjust bar chart ticks on tab switch

rajadain:
## Overview

Previously, the updateChart function which adjusts ticks on bar charts to ensure they don't overlap only fired when the browser window was resized. Although there is a listener for a `bar-chart:refresh` event, designed to trigger during a tab switch, this would never fire:

https://github.com/WikiWatershed/model-my-watershed/blob/5b295d7d25b361e4cf452cb7c902f488c9a41bbc/src/mmw/js/src/core/chart.js#L162-L167

By firing that event purposefully when switching tabs, it is ensured that the ticks are always updated accordingly.

Connects #2677 

### Demo

image

image

### Notes

I looked in to firing `updateChart` via the parent view, or the grandparent view, but the nesting was too deep to try and pass that reference around. Triggering this event via a jQuery selector does have some "action at a distance" magic feeling, but was the quickest approach outside a larger refactoring.

## Testing Instructions

* Go to [:8000/](http://localhost:8000/) and draw a shape. If you have boundaries installed, select the Petty Island Delaware River HUC-12 (as shown in the screenshots).
* In the Analyze stage, ensure that there are no overlapping labels in the bar charts in the Land and Soil tabs.
* Start over, and select a shape. Quickly switch to Land tab before the Analyze populate, and ensure there are no overlaps even if the chart loads when the tab is active.
* Test in various browsers.

### DIFF
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -300,8 +300,23 @@ var TabPanelView = Marionette.ItemView.extend({
         role: 'presentation'
     },
 
+    ui: {
+        'tabPanelButton': 'a[data-toggle="tab"]',
+    },
+
+    events: {
+        'shown.bs.tab @ui.tabPanelButton': 'onTabShown',
+    },
+
     initialize: function() {
         this.listenTo(this.model, 'change:polling', this.render);
+    },
+
+    onTabShown: function() {
+        var targetId = this.ui.tabPanelButton.attr('href'),
+            chartEl = targetId + ' .bar-chart';
+
+        $(chartEl).trigger('bar-chart:refresh');
     }
 });
 


src/mmw/js/src/analyze/views.js
```diff
@@ -300,8 +300,23 @@ var TabPanelView = Marionette.ItemView.extend({
         role: 'presentation'
     },
 
+    ui: {
+        'a': 'a',
```
arottersman: Will `[data-toggle=tab]` select the same element? Might be a little more descriptive. Or we could name the `ui` key something other than `a`.

rajadain: So rename either one? You don't like the same key being used in both cases, or would like both to be more descriptive?

arottersman: I wasn't sure what element the `'a'` referred to without looking at the template, so thought it might help to either change what we're selecting on, or rename what we've selected, or both. Also, happy to have it left as is.

rajadain:
Updated to be more readable. Will merge when green.

Issue #2667, opened 2018-02-19
Export TR-55 Model Output as CSVs to HydroShare


rajadain:
Just like the Analyze CSVs are exported.

ajrobbins:
Filename should be: model_sitestorm__<scenarioname>_storm-runoff|storm-waterquality
Should output as a csv exactly matching the UI

rajadain:
Shares points with #2666

Issue #2666, opened 2018-02-19
Add Mapshed model output to HS export


ajrobbins:
As reported by @aufdenkampe , the HydroShare export currently does not include MapShed model output (hydrology table and water quality tables). Add csv's for those tables to the [HS export](https://github.com/WikiWatershed/model-my-watershed/commit/7b2b1125d158f7287c81ac6e497c5141f7aa1052).

rajadain:
There are more items in the results view than just the tables, such as summary statistics:

![image](https://user-images.githubusercontent.com/1430060/36681829-ad4f6c46-1ae7-11e8-9d2a-29380785f3ef.png)

Should these also be included, or just the tables?

ajrobbins:
@aufdenkampe thoughts?

ajrobbins:
Filename should be: model_multiyear_<scenarioname>_hydrology | waterquality
Should output as a csv exactly matching the UI

aufdenkampe:
@rajadain, regarding additional summary statistics, can you append those to the CSVs, as extra columns, but with blanks for all the rows except the Total / Entire Watershed row?

rajadain:
Like this:

| Sources                            |     Sediment | Total Nitrogen | Total Phosphorus | Mean Flow (m3/year) | Mean Flow (m3/s) |
|------------------------------------|-------------:|---------------:|-----------------:|---------------------:|------------------:|
| Total Loads (kg)                   | 10,842,816.7 |      299,694.8 |         45,412.5 |          115,777,006 |              3.67 |
| Loading Rates (kg/ha)              |     1,320.94 |          36.51 |             5.53 |                      |                   |
| Mean Annual Concentration (mg/l)   |        93.65 |           2.59 |             0.39 |                      |                   |
| Mean Low-Flow Concentration (mg/l) |        99.41 |           2.05 |             0.48 |                      |                   |

?

Issue #2592, opened 2018-01-04
Investigate HydroShare Export Performance Improvements


rajadain:
Currently exporting to HydroShare takes between 15-25s, which is quite long. Investigate:

* What is causing the delay? Is it the file generation? The API?
* How do other apps that export to HydroShare perform?
* What can we do to make it faster?

Some possible solutions may include:

* Batching file uploads if allowed by the API
* Turning the Export into a Celery job instead of a Request / Response to spare `app` connections

Issue #2673, opened 2018-02-21
Unify CSV Generation


rajadain:
The creation of the UI and HydroShare Export CSVs should be refactored out into a single function and used for both implementations. It's important for users to have field names that include units, as we do with the UI-generated CSVs, but also to have full significant digits for use in other applications, as we do with the HydroShare Export CSVs. However, we want to remain in sync with the API field names so as not to confuse users. We need to:

- [ ] Determine what should the API field names be called so that they include units.
  - Keeping in mind that as JSON keys they cannot have spaces or parentheses
- [ ] Refactor the CSV generation code to be a single utility function
- [ ] Reuse the above function in the UI CSV generation
- [ ] Reuse the above function in the HydroShare CSV export

PR #2688, opened 2018-02-27
HydroShare Export: Include Model Result Tables as CSV

rajadain:
## Overview

Exports model results for MapShed and TR-55 projects as CSVs identical to their tables to HydroShare. This is done by cycling through all scenarios in the UI, rendering the tables, exporting them, and collecting the results. The scenario switching is done in the background, and the previously user-selected scenario is reselected in the end.

The additional data gathering step, and the additional files being exported to HydroShare, delay the export process. I can see it becoming quite time taking with a handful of active scenarios. Optimizing this process is something we can look in to later as an enhancement (see #2592).

Switching through all the scenarios can potentially affect the scroll state of the sidebar.

Connects #2666 
Connects #2667 

### Notes

One of the interim commits 1272ac8 exports models as JSON, directly in the format returned from the API. Instead of omitting this work, I decided to leave it in for history.

I also had to upgrade the [`tableExport`](https://github.com/hhurz/tableExport.jquery.plugin) to the latest `1.9.9` to be able to export hidden tables. Existing functionality for downloads seems to be unaffected.

## Testing Instructions

* Check out this branch and `bundle --debug --vendor`
* Go to [:8000/](http://localhost:8000/) create a TR-55 project.
* Add a scenario. Make some modifications.
* Export the project to HydroShare. Inspect the export to ensure that all the model results are identical to the tables, and to the exports generated from the UI.
* Do the same for a MapShed project. Ensure that its values correspond as well.
* Ensure the export names match up with the specifications in the original issues.
* Ensure CSV export still functions as before.

This is a potentially complex code review. I will be available to pair on this if necessary.


arottersman:
Code is looking good! Taking this for a spin now

arottersman:
For TR-55 and Mapshed the secondary tabs (Water Quality) don't render their CSVs for hydroshare (the files are blank). I can also see the model tab getting changed to the whatever the hydroshare export is working on which is a bit undesirable. Might be easiest to set up some time tomorrow to show.

rajadain:
> For TR-55 and Mapshed the secondary tabs (Water Quality) don't render their CSVs for hydroshare (the files are blank).

Ah sorry, this is fixed by the `tableExport` update. You'll need to `bundle --debug --vendor` to reload that dependency, and then it should be fixed.

> I can also see the model tab getting changed to the whatever the hydroshare export is working on which is a bit undesirable.

Yes this is less than ideal. However, since we're using `tableExport` which exports a rendered table, we need to render the table for each scenario before exporting it. Transforming the data model into this format would require more refactoring or duplication of logic currently used in rendering the tables, both of which sound worse than this current implementation. Hopefully these issues can be addressed in #2673, if we ever get around to it.

arottersman:
Oh, my bad. I'll test this out with the updated `tableExport`! Everything else was working, so I bet we're good.

rajadain:
> Oh, my bad.

No, I hadn't included in the test instructions. Just updated them now.

arottersman:
* If you're on the analyze tab while the share export is happening, the AOI header will disappear when the export releases the tables
screen shot 2018-03-01 at 11 30 30 am
_poof!_
screen shot 2018-03-01 at 11 30 17 am

* Saw this a few times — still working on concrete steps to reproduce
```
HydroShareHTTPException at /export/hydroshare
Received status 500 Internal Server Error when accessing https://beta.hydroshare.org/hsapi/resource/d0abe6b8fd104b5d871be6cfb6b1bb9a/files/ with method GET and params None.
```

* Now that the export takes a while, we might want to have some sort of loading indicator on the share button — sometimes I'll click "New Project" or navigate away from the page, and the project won't have finished its export

Moving to other browsers now

rajadain:
> * Now that the export takes a while, we might want to have some sort of loading indicator on the share button — sometimes I'll click "New Project" or navigate away from the page, and the project won't have finished its export

Hmm. We are adding a special class to the "Share" button when an export is active, but never got around to styling it. I'll just change the text to "Exporting ..." while sharing for now, so there's some visual feedback.

---

It occurs to me that cycling through scenarios can be really disruptive for when the export is triggered automatically. Changing the UI for a background process seems like a very fragile and error-prone idea. Ideally we would be using the model itself and nothing else. Let me see if I can think of a decent alternative.

rajadain:
Thank you for the detailed review. After much thought I decided that doing this via the UI was just not sustainable, and we have to use the model. We probably should have been doing that all along. Anyway, I finally put up a PR #2700, which does exactly that, and am closing this in its favor.

PR #2694, opened 2018-03-02
Update libpq glob to reflect main PostgreSQL repository

arottersman:
## Overview

Same as just about every commit [here](https://github.com/WikiWatershed/model-my-watershed/search?q=libpq&type=Commits&utf8=%E2%9C%93)

@tnation14 — something feels wrong that we have to keep doing this. Is there any other repo for this we can use that doesn't require we bump the version every month?

## Testing Instructions

 * Pull branch
 * `vagrant destroy worker` then `vagrant up worker`


tnation14:
Unfortunately, I don't think we have a better option. The PGDG repo has the most up-to-date package versions, but they're really "good" about pruning old versions from their repository.

Issue #2692, opened 2018-03-01
Hydroshare Link In Safari Downloads Html


arottersman:
Clicking the link to a hydroshare resource in safari downloads the page instead of opening it
<img width="399" alt="screen shot 2018-03-01 at 11 54 26 am" src="https://user-images.githubusercontent.com/7633670/36857797-73d65d78-1d47-11e8-85c6-f7c3744527e6.png">

PR #2698, opened 2018-03-02
Fix hydroshare resource link in safari

arottersman:
Hydroshare link was downloading in safari instead of opening in a new tab. We had `target="_blank"` but no `rel="noreferrer noopener"`

screen shot 2018-03-02 at 4 51 18 pm
screen shot 2018-03-02 at 4 51 37 pm

Connects #2692

## Testing Instructions

 * Create a hydroshare-shared project and open the share modal in safari. Confirm clicking the link to the resource opens it in hydroshare.


arottersman:
Thanks! Going to merge

PR #2700, opened 2018-03-05
HydroShare Export: Properly Include Model Results as CSV

rajadain:
## Overview

Converts each result model into the appropriate CSV, identical to the CSV generated in the UI, to export to HydroShare. As opposed to #2688, this is done from the model rather than the UI, thus there is no user penalty. It is also faster, so we can forego much of the promise nesting done there.

Since this export is now being generated independently of the tables, we can make changes here without making changes to the UI (for example, by adding more precision, or changing the column names / order). On the other hand, if we want to keep this export and the UI in sync, we'll have to remember to update this whenever updating the UI.

Currently the UI export still uses the `tableExport` plugin, although it could possibly be switched to use this (since the results are identical or superior, since we include the Totals row in MapShed Hydrology, and the Mean Flow columns in MapShed Water Quality Summary) to satisfy #2673. If deemed worthy, I can include that here, or we can do it at a later date.

Supersedes #2688 
Connects #2666
Connects #2667

### Demo

* Exporting Model Data:

    

* `Share` → `Exporting ...` → `Share`

    

Tagging @ajrobbins and @jfrankl for visual review.

## Testing Instructions

* Check out this branch
* Export a TR-55 project. Ensure the CSVs in HydroShare are named correctly and are identical to those generated from the UI
* Export a MapShed project. Ensure the CSVs in HydroShare are named correctly and are identical to those generated from the UI

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -10,7 +10,9 @@ var $ = require('jquery'),
     coreModels = require('../core/models'),
     turfArea = require('turf-area'),
     turfErase = require('turf-erase'),
-    turfIntersect = require('turf-intersect');
+    turfIntersect = require('turf-intersect'),
+    AoiVolumeModel = require('./tr55/models').AoiVolumeModel,
+    round = utils.toRoundedLocaleString;
 
 var ModelPackageControlModel = Backbone.Model.extend({
     defaults: {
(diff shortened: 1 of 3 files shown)


src/mmw/js/src/modeling/models.js
```diff
@@ -557,23 +557,63 @@ var ProjectModel = Backbone.Model.extend({
                         contents: at.getResultCSV(),
                     };
                 }),
+            scenarios = self.get('scenarios'),
+            lowerAndHyphenate = function(name) {
+                    return name.toLowerCase().replace(/\s/g, '-');
+                },
+            isTR55 = self.get('model_package') === utils.TR55_PACKAGE,
```
arottersman: Consider instead adding helper functions like `createTR55ModelFileCSVs()` and `createMapshedModelFileCSVs()`, so that we don't have everything `isTR55 ?`ed. It doesn't seem like the duplication would be too bad.

rajadain: Where were you envisioning these helper methods would go? Within `exportToHydroShare`, or within `modelFiles`, or within `ResultModel` or `ResultCollection`?

arottersman: I was picturing within `exportToHydroShare`, though the `ResultCollection` might be a good idea too...it only needs to know the scenarios and the model package?

src/mmw/js/src/modeling/models.js
```diff
@@ -80,7 +82,127 @@ var ResultModel = Backbone.Model.extend({
...
+        rows = rows.concat(results.map(function(result, i) {
+            var cols = [monthNames[i]].concat(runoffVars.map(function(runoffVar) {
+                    return round(result[runoffVar], 2);
+                })).join('","');
+
+            return '"' + cols + '"';
+        }));
+
+        // Total row
+        var totalCols = ['"Total"'].concat(runoffVars.map(function(runoffVar) {
```
arottersman: This is creating an extra quotation mark somewhere:
`""Total"","103.77","50.97","21.27","33.14","81.99","136.39"`
vs 
`"Dec","8.81","1.27","2.61","3.61","6.96","13.19"`

rajadain:
> is it possible to write a test that makes sure they match

It is possible, but expensive. There are currently no tests for any ResultViews in the Modeling tests, and we'll need to capture a number of mocks (scenario, area of interest, results) to make it work. Furthermore, we won't be able to match the `tableExport` results directly with `.toTr55RunoffCSV()` etc because `tableExport` only quotes values that need them (strings with spaces, numbers with commas), whereas we quote every value. We will likely have to render the view in a sandbox region and then reconstruct the CSV within the test using the same approach used above, just with table values instead of model values, and then match it with `.toTr55RunoffCSV()` etc.

I'll make a follow up card for this.

src/mmw/js/src/modeling/models.js
```diff
@@ -80,7 +82,127 @@ var ResultModel = Backbone.Model.extend({
...
+    toTR55WaterQualityCSV: function(isCurrentConditions, aoiVolumeModel) {
+        var rows = ['"Quality Measure","Load (kg)","Loading Rate (kg/ha)","Average Concentration (mg/L)"'],
+            resultKey = isCurrentConditions ? 'unmodified' : 'modified',
+            results = this.get('result')['quality'][resultKey];
+
+        return rows.concat(results.map(function(result) {
+            var load = result.load,
+                loadingRate = aoiVolumeModel.getLoadingRate(load),
+                adjustedRunoff = aoiVolumeModel.adjust(result.runoff),
+                concentration = adjustedRunoff ? load / adjustedRunoff : 0,
```
arottersman: There's a discrepancy between the downloaded and table concentration. I believe we sneakily convert this value from grams to milligrams, and need to do so here.
https://github.com/WikiWatershed/model-my-watershed/blob/2afb33243b187e9b64bef42d2d8db442a041c008/src/mmw/js/src/modeling/tr55/quality/views.js#L98

screen shot 2018-03-05 at 5 38 27 pm
screen shot 2018-03-05 at 5 38 23 pm

src/mmw/js/src/modeling/models.js
rajadain: Added in 21e453e.

rajadain:
Split the TR-55 / MapShed export into more digestible chunks, added some error resilience so even if the results are not ready yet the export will not fail (the non-ready results will be ignored, the rest exported), and addressed the comments above.

Ready for another look.

rajadain:
Thanks for reviewing! I'll merge the fixups and wait for green before merging. Will also create a card for adding CSV generation tests.

rajadain:
Made #2706 for ensuring the CSVs match the tables.

Issue #2699, opened 2018-03-02
Analyze Results Don't Clear When Selecting Shape from Search


rajadain:
## Steps to Reproduce

* Select a shape and proceed to Analyze. Wait for the analyses to finish.
* Click "Change Area" and type in something in the search bar.
* Select a suggested HUC. Click "Select" in the search bar to proceed to Analyze.
* See analyze results from previous shape.

### Observed Behavior

Analyze results from prior shape are shown.

### Expected Behavior

Analyze results for new shape are kicked off.

## Demo

![2018-03-02 17 37 36](https://user-images.githubusercontent.com/1430060/36925483-23535b94-1e41-11e8-8edb-379f3f4ac46c.gif)

ajrobbins:
Good catch! I was able to replicate this just now.

PR #2705, opened 2018-03-05
Clear analysis results on selecting geocoder shape

kellyi:
## Overview

This PR clears the analyze results on selecting a shape from the geocoder search bar. We also clear the data catalog, since that seems always to be paired with other similar `clearAnalyzeCollection` method calls.

Connects #2699

## Testing
- get and serve this branch and try to replicate the bug described in #2699 and verify that: you can't


kellyi:

Issue #2543, opened 2017-11-30
Handle large subbasin gwlf-e payloads


arottersman:
When submitting the GMS output for all huc12s in a large huc8 the payload sometimes exceeds the `5MB` limit on the nginx server.

- Send the job id instead + any BMPs
- The BMP application that currently happens on the frontend will now have to happen on the server after we've looked up the result via the job id
- The implementation of the point above will be dependent on #2542

arottersman:
@rajadain, we chatted at RRP about using the project's `gis_data` attribute instead of the job table's results. Anonymous users, however, don't create stored projects.

In practice we've never deleted the jobs table, and even though it feels impermanent it's not the cache or anything. Any strong concerns if we use it for the mapshed results?

rajadain:
re: Anonymous users, good point, that is true. I suppose the front-end does have the job id, and we don't make a habit of clearing out the jobs table, and even in this case we're using its data in a relatively temporary fashion (will not persist between user sessions if they log out / back in). Okay I approve this methodology ✅

PR #2722, opened 2018-03-13
Allow sending mapshed job UUID to GWLF-E

arottersman:
## Overview

In subbasin mode, the mapshed GMS payloads can grow to over
5MB (our nginx limit) when an AOI has more than about 20 child
HUC-12s.

Allow sending the MapShed Job UUID to the GWLF-E endpoint so
we don't worry about exceeding the payload size.

Now that the un-modified GMS can come from the job table, burning modifications into the GMS
needs to happen on the backend.

Connects #2543 

### Demo

screen shot 2018-03-13 at 1 48 04 pm

### Notes

Using the job UUID to look up MapShed results puts pressure
on maintaining the jobs table which is not necessarily meant for long
term storage. We must now be careful if we ever need to empty
it, and we should consider other places to store the MapShed
results in the future.

## Testing Instructions
[Before pulling this branch make sure you have an old mapshed project.]

Pull, run migrations, and `bundle` the app
```sh
./scripts/manage.sh migrate
```

 **Anonymous user**
* Open your developer tools' network tab
* Model the `City of Philadelphia-Schuylkill River, HUC-12 Subwatershed` with MapShed
* Confirm the water quality results match the screen shot taken from develop below
screen shot 2018-03-13 at 2 43 32 pm
* Confirm only the `mapshed_job_uuid` is sent, not the `model_input`

* Add a new scenario and add two mods: 3ha of Nutrient Management and 4000ha of Surface Water Retention
* Confirm the new water quality results match the develop screenshot below 
screen shot 2018-03-13 at 2 43 53 pm

**Logged in user**

* Load a mapshed project you created before this branch. Confirm MapShed runs again to populate the UUID field on the model
* Confirm you can model and reload a model while logged in

**Sub-basin** 
Download [mapshed_model.sh.txt](https://github.com/WikiWatershed/model-my-watershed/files/1808297/mapshed_model.sh.txt) and remove the `.txt` ending. Run it with `SUBBASIN=true`. 

**`model_input`**
Follow the instructions started at `L58` of the above script, and set `SUBBASIN=false`. Call the script again and confirm the gwlf-e endpoint can still work like it did.

### DIFF
--- a/src/mmw/apps/modeling/models.py
+++ b/src/mmw/apps/modeling/models.py
@@ -6,6 +6,8 @@ from __future__ import division
 from django.contrib.gis.db import models
 from django.contrib.auth.models import User
 
+from apps.core.models import Job
+
 
 class Project(models.Model):
     TR55 = 'tr-55'
@@ -43,6 +45,20 @@ class Project(models.Model):
         null=True,
         help_text='Serialized JSON representation of additional'
                   ' data gathering steps, such as MapShed.')
+    mapshed_job_uuid = models.ForeignKey(
+        Job,
+        to_field='uuid',
+        related_name='mapshed_job',
+        null=True,
+        help_text='The job used to calculate the MapShed results.'
+                  ' Used for getting the results of that job.')
+    subbasin_mapshed_job_uuid = models.ForeignKey(
+        Job,
+        to_field='uuid',
+        related_name='subbasin_mapshed_job',
+        null=True,
+        help_text='The job used to calculate the MapShed results'
+                  ' for each HUC-12 sub-basin of the shape.')
     wkaoi = models.CharField(
         null=True,
         max_length=255,
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -248,12 +248,14 @@ var ProjectModel = Backbone.Model.extend({
         area_of_interest: null,            // GeoJSON
         area_of_interest_name: null,       // Human readable string for AOI.
         wkaoi: null,                       // Well Known Area of Interest ID "{code}__{id}"
-        subbasin_modeling: false,          // Use subbasin modeling for MapShed/GWLF-E
+        subbasin_modeling: false,          // (GWLFE) Use subbasin modeling
         model_package: utils.TR55_PACKAGE, // Package name
         scenarios: null,                   // ScenariosCollection
         user_id: 0,                        // User that created the project
         is_activity: false,                // Project that persists across routes
         gis_data: null,                    // Additionally gathered data, such as MapShed for GWLF-E
+        mapshed_job_uuid: null,            // (GWLF-E) The id of a successful MapShed job
+        subbasin_mapshed_job_uuid: null,   // (GWLF-E) The id of a successful sub-basin MapShed job
         needs_reset: false,                // Should we overwrite project data on next save?
         allow_save: true,                  // Is allowed to save to the server - false in compare mode
         sidebar_mode: utils.MODEL,         // The current mode of the sidebar. ANALYZE, MONITOR, or MODEL.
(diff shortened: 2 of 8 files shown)


src/mmw/js/src/modeling/models.js
```diff
@@ -1257,25 +1260,15 @@ var ScenarioModel = Backbone.Model.extend({
...
-                    if (subbasinModeling) {
-                        // TODO This currenlty results in incorrect results
-                        // because we're not properly taking the subbasin areas
-                        // into account
-                        _.forEach(mergedGisData, updateGisData);
-                    } else {
-                        updateGisData(mergedGisData);
-                    }
-                });
+                var modificationsCollection = self.get('modifications'),
```
rajadain: These lines can probably be collapsed in to:

```js
var modifications = self.get('modifications').pluck('output');
```

rajadain:
When I ran MapShed (without Subbasin) on the Philadephia HUC-12 on this branch, I get different results than your screenshot:

image

I get identical results on staging. Is your screenshot (quoted below) taken with Subbasin on?

> screen shot 2018-03-13 at 2 43 32 pm

rajadain:
What's important is that the numbers I see are consistent with staging even when using `mapshed_job_uuid` rather than `model_input`, which is true, both for current conditions and a new scenario with modifications. Proceeding to test the other use cases now.

rajadain:
After running the script I get this as the final output:

```json
{
  "message": "Forbidden"
}
```

Is that expected?

arottersman:
That's a response from the Drexel SRAT API, so it's a good sign if it got that far. Do you have `MMW_SRAT_CATCHMENT_API_KEY` set?

rajadain:
That was it. Set the key in both `app` and `worker` and now I get this:

```json
{
  "message": "Internal server error"
}
```

which I suppose is the SRAT response, but means that our code is working correctly. Right?

arottersman:
Can confirm the error is from the Drexel API (it chokes when it gets fed the `Stinking Lake` HUC-12 that's part of the test payload in the `mapshed_model.sh` script.) If you uncomment the `WKAOI=huc8__492` line, that huc8 should return complete results after a while.

src/mmw/apps/modeling/models.py
```diff
@@ -43,6 +43,14 @@ class Project(models.Model):
         null=True,
         help_text='Serialized JSON representation of additional'
                   ' data gathering steps, such as MapShed.')
+    mapshed_job_uuid = models.UUIDField(
```
rajadain: Since this really a `ForeignKey` field, consider replacing with:

```python
mapshed_job = models.ForeignKey(
  Job,
  to_field='uuid',
  null=True,
  help_text='The job used to calculate the MapShed results. '
            'Used for getting the results of that job.'
),
```

src/mmw/js/src/modeling/models.js
```diff
@@ -521,12 +523,13 @@ var ProjectModel = Backbone.Model.extend({
         var self = this,
             saveProjectAndScenarios = _.bind(self.saveProjectAndScenarios, self);
 
-        if (self.get('gis_data') === null && self.fetchGisDataPromise === undefined) {
+        if (self.get('mapshed_job_uuid') === null && self.fetchGisDataPromise === undefined) {
```
rajadain: Nice. This ensures that when existing projects are reloaded, they are converted to the new format.

src/mmw/apps/modeling/models.py
rajadain: Same for `subbasin_job` below.

arottersman:
Thanks, those were great suggestions. The UUID fields are now legitimate foreign keys. I also found a bug with our the gis_data download button that is fixed by the last fixup above. Let me know if this looks good to merge. You'll have to unapply the old migration to apply the new one.

rajadain:
Taking another look shortly.

rajadain:
Friendly reminder to squash fixups.

arottersman:
Thanks for the multiple, thorough reviews. Will run migration when this hits staging.

Issue #2721, opened 2018-03-12
HydroShare Login Doesn't Work in IE


rajadain:
In Internet Explorer, disconnect your account from HydroShare, then reconnect it. Enter credentials to log in, expect to proceed to the authorization stage. Instead, get this screen:

![image](https://user-images.githubusercontent.com/1430060/37309460-62cda602-2617-11e8-8c90-e9cd6c6c0bae.png)

This is because IE doesn't allow setting third party cookies, as described here: https://chase-seibert.github.io/blog/2011/06/17/django-csrf-verification-failed-in-ie-iframe.html

This can potentially be alleviated by opening the login and authorization link in a new window for IE users.

PR #2725, opened 2018-03-13
Fix HydroShare Login in Internet Explorer

rajadain:
## Overview

Internet Explorer stringently disallows third party cookies, which make logging in to HydroShare via an iframe impossible, since we cannot send the CSRFToken necessary for Django (which HydroShare is written in) to work. Thus, we detect when the browser is running IE, and redirect instead of iframe in that case. Other browsers will continue to work as is.

Connects #2721 

### Notes

The workflow would be nicer if the "Linked Accounts" section had a URL that we could redirect to, instead of just the "Accounts" home, but that seemed out of scope for this card.

I also tried opening HydroShare in a new tab, which works for the most part, but we need to refresh / refetch the user model from the server to get the new permissions, which we cannot do automatically like we can with the iframe (since `postMessage` does not work outside of iframes). We could have had a message like "please go back to your original tab and refresh it" but this seemed more intuitive.

## Testing Instructions

* Go to "My Account" → "Linked Accounts" in IE. Ensure that when connecting to HydroShare you are redirected there and back success and failure.
* Go to "My Account" → "Linked Accounts" in any other browser. Ensure that you see the iframe and you are not redirected.
* Go to any project and share to HydroShare when your account is not connected in IE. Ensure you are shown a confirmation popup and are redirected to "My Accounts" on clicking it.
* Go to any project and share to HydroShare when your account is not connected in any other browser. Ensure you are shown the iframe and subsequently the HydroShare export modal.

### DIFF
--- a/src/mmw/apps/user/templates/user/hydroshare-auth.html
+++ b/src/mmw/apps/user/templates/user/hydroshare-auth.html
@@ -38,11 +38,27 @@
             var message = 'mmw-hydroshare-success';
             {% endif %}
             parent.postMessage(message, window.location.origin);
+
+            {% if not error %}
+            // Redirect to home if not in iframe and already closed
+            // since iframes are closed after 500ms in core.modals.views.IframeView.
+            setTimeout(function() {
+                window.location.href = window.location.origin + '/account/';
+            }, 750);
+            {% endif %}
         });
 
         {% if error %}
         document.getElementById('hydroshare-cancel').addEventListener('click', function() {
             parent.postMessage('mmw-hydroshare-cancel', window.location.origin);
+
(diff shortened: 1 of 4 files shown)


arottersman:
Taking a look!

rajadain:
I can't recreate that behavior on my VM:



where I see the proper "wrong password" message. Maybe its a browserstack quirk, maybe a Windows 10 / IE 11 combination issue. I think we're close enough for now, and if we find this bug in the wild we can fix it then.

Does that sound good @ajrobbins?

ajrobbins:
@rajadain yep!

rajadain:
Thanks for taking a look!

Issue #2738, opened 2018-03-22
Investigate Celery Timeout Failure


arottersman:
Investigate celery's failure when geoprocessing times out. See comments in
https://github.com/WikiWatershed/model-my-watershed/issues/2695

PR #2751, opened 2018-04-02
Add Geoprocessing Request Timeout

arottersman:
## Overview

We've witnessed a state twice on production where the geoprocessing service times out on every response and must be stopped and started to get back into a good state. Because there was no  timeout on the geoprocessing task requests, the tasks would take longer than the `celery` hard task timeout limit (300s).  Exceeding this timeout would put the celery service in its own bad state, and it, too would need a restart (perhaps related celery/celery#2451).

While we are determining the underlying cause of the geoprocessing failures, add a timeout on the geoprocessing request that is less than the celery hard task timeout (request timeouts are probably best practice anyway. There is even a warning specifically about not putting a timeout on I/O ops in a [task in the celery docs](http://docs.celeryproject.org/en/latest/userguide/tasks.html)). This way a rollbar error will notify us of a bad state instead of an email from our client saying all analyze and model runs timeout.

Connects #2738 

### Demo

screen shot 2018-04-02 at 12 40 54 pm
screen shot 2018-04-02 at 12 40 34 pm

## Testing Instructions

 * Pull this branch, and `./scripts/debugcelery.sh`
 *  I've add a commit that drops the timeout down to a second. I'll drop this commit before merging so the timeout is 200 seconds. Draw a big shape so the operation isn't cached and it takes the geoprocessing service longer than a second to chew through an analyze task. Confirm you see an error caught in the celery logs

### DIFF
--- a/src/mmw/apps/modeling/geoprocessing.py
+++ b/src/mmw/apps/modeling/geoprocessing.py
@@ -11,7 +11,7 @@ from ast import literal_eval as make_tuple
 from celery import shared_task
 from celery.exceptions import Retry
 
-from requests.exceptions import ConnectionError
+from requests.exceptions import ConnectionError, Timeout
 
 from django_statsd.clients import statsd
 
@@ -232,10 +232,13 @@ def geoprocess(endpoint, data, retry=None):
     try:
         response = requests.post(geop_url,
                                  data=json.dumps(data),
-                                 headers={'Content-Type': 'application/json'})
+                                 headers={'Content-Type': 'application/json'},
+                                 timeout=settings.TASK_REQUEST_TIMEOUT)
     except ConnectionError as exc:
         if retry is not None:
             retry(exc=exc)
+    except Timeout as exc:
+        raise Exception('Geoprocessing service timed out.')
 
     if response.ok:
         result = response.json()
--- a/src/mmw/mmw/settings/base.py
+++ b/src/mmw/mmw/settings/base.py
@@ -138,6 +138,12 @@ CELERY_TASK_QUEUES = {
 }
 CELERY_TASK_DEFAULT_EXCHANGE = 'tasks'
 CELERY_TASK_DEFAULT_ROUTING_KEY = "task.%s" % STACK_COLOR
+
+# The longest running tasks contain geoprocessing requests
+# Keep the task and request time limit above, but in the ballpark of
+# https://github.com/WikiWatershed/mmw-geoprocessing/blob/develop/api/src/main/resources/application.conf#L9
+CELERY_TASK_TIME_LIMIT = 90
+TASK_REQUEST_TIMEOUT = CELERY_TASK_TIME_LIMIT - 10
 # END CELERY CONFIGURATION
 
 
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -8,7 +8,7 @@ import requests
 import json
 
 from ast import literal_eval as make_tuple
-
+from requests.exceptions import ConnectionError, Timeout
 from StringIO import StringIO
 
 from celery import shared_task
@@ -322,9 +322,12 @@ def run_srat(watersheds):
     try:
         r = requests.post(settings.SRAT_CATCHMENT_API['url'],
                           headers=headers,
-                          data=json.dumps(data))
-    except Exception as e:
-        raise Exception('Request to SRAT Catchment API failed: %s' % e)
+                          data=json.dumps(data),
+                          timeout=settings.TASK_REQUEST_TIMEOUT)
+    except Timeout as e:
+        raise Exception('Request to SRAT Catchment API timed out')
+    except ConnectionError:
+        raise Exception('Failed to connect to SRAT Catchment API')
 
     if (r.status_code != 200):
         raise Exception('SRAT Catchment API request failed: %s %s' %


src/mmw/apps/modeling/geoprocessing.py
```diff
@@ -232,10 +232,13 @@ def geoprocess(endpoint, data, retry=None):
     try:
         response = requests.post(geop_url,
                                  data=json.dumps(data),
-                                 headers={'Content-Type': 'application/json'})
+                                 headers={'Content-Type': 'application/json'},
+                                 timeout=200)
```
mmcfarland: I tried to find where we set the Celery hard timeout to 300, but couldn't find it at a quick glance.  Is it possible to set this value as a function of that setting since the relationship is important?

src/mmw/apps/modeling/geoprocessing.py
```diff
@@ -232,10 +232,13 @@ def geoprocess(endpoint, data, retry=None):
...
     except ConnectionError as exc:
         if retry is not None:
             retry(exc=exc)
+    except Timeout as exc:
+        raise Exception('Geoprocessing service timed out: {}'.format(exc))
```
mmcfarland: Is there a way to report the actual exception contents in the job table but restrict what gets sent to the client?  The timeout exception reports internal hosts and ports which while not crucial to protect does feel like an unnecessary leak.  It's likely that the qualitative description you provide is more meaningful than the `Timeout` exception contents in the case where we have to choose between leaking and not capturing that exception detail.

src/mmw/apps/modeling/geoprocessing.py
arottersman: Yes, we don't set it currently; it has a default of `300`. I will set it explicitly, and set a request timeout setting based on it.

src/mmw/apps/modeling/geoprocessing.py
arottersman: I'll look into splitting our job error and what gets sent back to the client, though I'm not sure we're set up for that. We could also write details we don't want to send to the client to the celery log instead. I agree they're not even that helpful, so I may just remove the printing of the exception entirely.

src/mmw/mmw/settings/base.py
```diff
@@ -138,6 +138,9 @@ def get_env_setting(setting):
 }
 CELERY_TASK_DEFAULT_EXCHANGE = 'tasks'
 CELERY_TASK_DEFAULT_ROUTING_KEY = "task.%s" % STACK_COLOR
+
+CELERY_TASK_TIME_LIMIT = 300
```
mmcfarland: It looks like in v4, which we're on now, has no default set for time limit:
http://docs.celeryproject.org/en/latest/userguide/configuration.html#task-time-limit

So setting this alone may have an impact, although it looks like the prefix should be CELERY**D** according to 
http://docs.celeryproject.org/en/latest/userguide/configuration.html#new-lowercase-settings  

The setting you have appears to have an effect when tweaked so it's likely backwards-compatible.

arottersman: I'll add the `D` at the end to be safe.

That's weird there's no set default. The error we were seeing on staging spoke of a HardTimeLimit of 300sec being exceeded.

arottersman: It doesn't appear to work for me if I add the `D`...were you able to set it that way? If not, I'm going to open an issue in celery. Could be a mistake in the docs...

mmcfarland: Nope, but definitely works the way you have it set now, so seems good to leave.

arottersman:
re: `save_job_error` error. When we exceed the hard time limit, we are trying to save a Job that doesn't have a traceback, a column for which there is a NOT NULL constraint. This throws an `IntegrityError`, which doesn't have an `errno` attribute, which then crashes our `catch` block.

I don't think, however, the hanging always has to do with a failure in `save_job_error`. Sometimes it seems the job never even makes it to `save_job_error` at all. After this message my worker entered its horribly stuck state:

```
[2018-04-03 11:40:25,852: INFO/MainProcess] Received task: apps.geoprocessing_api.tasks.analyze_pointsource[c818276b-08a1-4c8f-8db1-9390c9f7a01d]
[2018-04-03 11:40:26,095: ERROR/MainProcess] Task fbaeedf3-ec08-4f4d-9036-25dbfb4d2730 run from job 4875 raised exception: TimeLimitExceeded(1,)
None
[2018-04-03 11:40:26,184: ERROR/MainProcess] Task handler raised error: TimeLimitExceeded(1,)
  File "/usr/local/lib/python2.7/dist-packages/billiard/pool.py", line 658, in on_hard_timeout
    raise TimeLimitExceeded(job._timeout)
TimeLimitExceeded: TimeLimitExceeded(1,)
[2018-04-03 11:40:26,186: ERROR/MainProcess] Hard time limit (1s) exceeded for apps.modeling.geoprocessing.run[fbaeedf3-ec08-4f4d-9036-25dbfb4d2730]
[2018-04-03 11:40:26,221: ERROR/MainProcess] Task fbaeedf3-ec08-4f4d-9036-25dbfb4d2730 run from job 4875 raised exception: TimeLimitExceeded(1,)
None
[2018-04-03 11:40:27,366: ERROR/MainProcess] Process 'ForkPoolWorker-1' pid:10117 exited with 'signal 9 (SIGKILL)'
[2018-04-03 11:40:27,488: DEBUG/MainProcess] TaskPool: Apply <function _fast_trace_task at 0x7f3277201c80> (args:('apps.modeling.geoprocessing.run', '6a4ddfe8-55bf-4415-9682-7c34240a4ce1', {'origin': 'gen3201@app', 'lang': 'py', 'task': 'apps.modeling.geoprocessing.run', 'group': None, 'root_id': '6a4ddfe8-55bf-4415-9682-7c34240a4ce1', u'delivery_info': {u'priority': 0, u'redelivered': None, u'routing_key': 'Black', u'exchange': u''}, 'expires': None, u'correlation_id': '6a4ddfe8-55bf-4415-9682-7c34240a4ce1', 'retries': 0, 'timelimit': [None, None], 'argsrepr': '(\'soil\', {\'polygon\': [\'{ "type": "MultiPolygon", "coordinates": [ [ [ [ -82.488346990747843, 38.472211785968319 ], [ -82.476862970189657, 38.472211785968319 ], [ -82.476862970189657, 38.481202179740976 ], [ -82.488346990747843, 38.481202179740976 ], [ -82.488346990747843, 38.472211785968319 ] ] ] ] }\']}, None)', 'eta': None, 'parent_id': None, u'reply_to': '47c326ee-66c8-3795-b218-6012e9a1097f', 'id': '6a4ddfe8-55bf-4415-9682-7c34240a4ce1', 'kwargsrepr': '{}'}, '[["soil", {"polygon": ["{ \\"type\\": \\"MultiPolygon\\", \\"coordinates\\": [ [ [ [... kwargs:{})

```

If we're interested in addressing the `save_job_error` problems, I was thinking of something like the diff below. I don't think it is the underlying problem though; I think we need to avoid hitting the hard timeout in the first place. I am inclined to fix up this PR to have lower request and hard time outs, and call it a day.


```diff
diff --git a/src/mmw/apps/core/tasks.py b/src/mmw/apps/core/tasks.py
index 09c90e9..282692a 100644
--- a/src/mmw/apps/core/tasks.py
+++ b/src/mmw/apps/core/tasks.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 from __future__ import absolute_import

 from django.utils.timezone import now
+from django.db import IntegrityError
 from celery import shared_task
 from apps.core.models import Job

@@ -31,10 +32,14 @@ def save_job_error(request, exc, traceback, job_id):
     try:
         job = Job.objects.get(id=job_id)
         job.error = exc
-        job.traceback = traceback
+        job.traceback = traceback or ''
         job.delivered_at = now()
         job.status = 'failed'
         job.save()
+    except IntegrityError as e:
+        logger.error('Failed to save job error status. Job will appear hung. \
+                     Job Id: {0}'.format(job.id))
+        logger.error('IntegrityError: {}'.format(e))
     except Exception as e:
         logger.error('Failed to save job error status. Job will appear hung. \
                      Job Id: {0}'.format(job.id))

```


What the integrity error ends up looking like:

```
[2018-04-03 11:00:02,366: ERROR/MainProcess] Task ca7b5415-6914-4f21-97c5-3a390e5339cf run from job 4858 raised exception: TimeLimitExceeded(1,)
None
[2018-04-03 11:00:02,516: ERROR/MainProcess] Failed to save job error status. Job will appear hung.                      Job Id: 4858
[2018-04-03 11:00:02,518: ERROR/MainProcess] IntegrityError: null value in column "traceback" violates not-null constraint
DETAIL:  Failing row contains (4858, 88111856-0df4-4397-a790-6a82cc4bc157, { "type": "MultiPolygon", "coordinates": [ [ [ [ -82.26219177246..., 2018-04-03 15:00:00.679+00, , 2018-04-03 15:00:02.507433+00, TimeLimitExceeded(1,), null, failed, 1).
```

arottersman:
Thanks for all your help with this. Going to merge.

Issue #2716, opened 2018-03-08
Selecting AoI from geocoder breaks collapsed sidebar


ajrobbins:
![screen shot 2018-03-08 at 3 03 06 pm](https://user-images.githubusercontent.com/12868342/37173576-fbee0ae0-22e1-11e8-94e5-f08b046e9d00.png)
To replicate:
- Open MMW landing page
- Collapse sidebar using button at bottom right of map
- Enter a search term like "Schuylkill" in geocoder, select a result from the list
- Observe that sidebar re-opens and is blank

Expected: When an AoI is selected, sidebar re-opens and is not blank

PR #2757, opened 2018-04-03
Fix Geocoder Behaviour When Sidebar Is Collapsed

arottersman:
## Overview

The sidebar toggle control was setting a hidden
class on the sidebar. When a suggestion is selected
from the geocoder, the sidebar was activated again
but the hidden class was never removed. Modify the
toggle control so that it always represents the
map model state (size.hasSidebar) and modify the
`toggleSidebar` control on the map model to handle
hiding the hidden class on the sidebar.

Connects #2716 

### Demo

fdp0by3ijo

## Testing Instructions

 * Pull and `bundle`
 * Search for stuff with the sidebar open and closed. Be sure to test selecting both locations and boundaries

### DIFF
--- a/src/mmw/js/src/core/sidebarToggleControl.js
+++ b/src/mmw/js/src/core/sidebarToggleControl.js
@@ -1,7 +1,6 @@
 'use strict';
 
 var L = require('leaflet'),
-    $ = require('jquery'),
     Marionette = require('../../shim/backbone.marionette'),
     sidebarToggleControlButtonTmpl = require('./templates/sidebarToggleControlButton.html');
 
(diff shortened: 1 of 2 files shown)


kellyi:
Taking a look at this.

kellyi:
Is the desired behavior for this that the sidebar is re-expanded when a geocoder selection is made? That's what I'm seeing. Doesn't seem like anything's still broken though!

arottersman:
> Is the desired behavior for this that the sidebar is re-expanded when a geocoder selection is made? That's what I'm seeing. Doesn't seem like anything's still broken though!

I was wondering this myself. It seems like it was made to take you direct to draw mode on geocoder selection. It doesn't make much sense to me to send someone to draw with the sidebar closed, so I left the behavior that opens it.

Issue #2758, opened 2018-04-03
Subbasin: Fix 504 On Large Watersheds


arottersman:
Trying to subbasin model the Schuylkill HUC-8 on staging results in a 504 Gateway Timeout response at the GWLF-E stage. The MapShed modeling portion generates the 62 GMS files correctly, but trying to submit the GWLF-E job with the mapshed UUID fails. The same request works fine during local development. 
The failing request can be made via [schuykill_subbasin_staging.sh.txt](https://github.com/WikiWatershed/model-my-watershed/files/1873356/schuykill_subbasin_staging.sh.txt)

The GWLF-E step works fine for smaller watersheds with fewer subbasins. For Schuylkill, the resulting set of GMS files generated by the MapShed step appear to be about 16MB. Investigate if this is too heavy to pass into the job initialization step, or if we need to adjust or nginx configuration; make the necessary adjustments.

arottersman:
We are now able to submit the Schuykill HUC-8. It runs the GWLF-E portion of the model in ~84sec on staging..

PR #2770, opened 2018-04-10
Chunk Subbasin Gwlfe Tasks

arottersman:
## Overview

Previously we were submitting
a group where each task in the
group was `run_gwlfe` on a subbasin.
Celery waits until all these tasks
submit before allowing access to
the whole chain's job id, which is
needed for initial job submission
response. When a subbasin request
was submitted, it was taking longer
the 60s for all subbasin gwlfe tasks
to submit if there were 60+ subbasins
in a shape. To alleviate this problem,
we need to cut down the number of
tasks in the group. We do this by
chunking the subbasin into smaller
groups and then running gwlfe
synchronously inside single task
on that group.

Notes:
* Celery+redis wasn't handling
  the large GMS payloads well, and
  would hang when called with this
  chunked approach. Instead of passing
  the GMSs, read the mapshed result
  for each subbasin inside the task
  itself.

* We tried using the builtin celery
  "chunk" task, but it resulted
  in mysterious hanging not fixed
  by limiting the payload size.

* The chunk size is set to 8 based
  on local trial and error. This value
  needs to be big enough to break up
  large (60-100) sets of subbasins
  into manageble task sets, but not
  so big that the task takes longer
  than the celery hard time limit.

Connects #2758 

## Testing
Pull the branch, and restart celery. 
Get setup to call this script:
[test_subbasin.sh.txt](https://github.com/WikiWatershed/model-my-watershed/files/1896203/test_subbasin.sh.txt)

It may be helpful to set up 2 celery threads to see how the chunk tasks distribute.

* Calling it as is will test the big Schuylkill huc-8 (60+ subbasins) that motivated the changes
* Changing the WKAOI at the top of the script to El Vado Reservoir (huc10__13907) will test the changes with a shape with fewer subbasins than the chunk size
* Changing SUBBASIN to false at the top of the script will test a regular mapshed job submission

### DIFF
--- a/src/mmw/apps/core/tasks.py
+++ b/src/mmw/apps/core/tasks.py
@@ -3,6 +3,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import absolute_import
 
+from django_statsd.clients import statsd
 from django.utils.timezone import now
 from celery import shared_task
 from apps.core.models import Job
@@ -43,6 +44,7 @@ def save_job_error(request, exc, traceback, job_id):
 
 
 @shared_task(bind=True)
+@statsd.timer(__name__ + '.save_job_result')
 def save_job_result(self, result, id, model_input):
     """
     Updates a job row in the database with final results.
--- a/src/mmw/apps/modeling/calcs.py
+++ b/src/mmw/apps/modeling/calcs.py
@@ -31,6 +31,14 @@ def split_into_huc12s(code, id):
         return cursor.fetchall()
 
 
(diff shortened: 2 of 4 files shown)


arottersman:
@azavea-bot rebuild

arottersman:
Added a commit that passes `save_job_result` the mapshed job uuid instead of the model_input dictionary of GMS files. This shaves 20sec off the job. Observe the second purple bar (ie the final `save_job_result) has essentially disappeared.

screen shot 2018-04-13 at 10 57 17 am

Issue #2750, opened 2018-04-02
Delete Stored Mapshed Results


arottersman:
MapShed results will change with the updates to the precipitation data in #2737. Write a migration to cleared saved project results.

arottersman:
Yes, good catch.

Issue #2650, opened 2018-02-08
Sub-basin: Add Sub-basin HUC-12 Level Container


arottersman:
When first entering subbasin mode display:
- a back button to exit subbasin mode
- the same AOI aggregate info (name + model description)
- two sub-container tabs "detail" and "huc-12s"

For the purpose of this card, do _not_ add:
- tables
- "related layers"

<img width="219" alt="screen shot 2018-03-13 at 3 50 10 pm" src="https://user-images.githubusercontent.com/7633670/37368048-71cd732a-26db-11e8-8d7a-c29b3b118659.png">

kellyi:
Blocked by #2647

PR #2772, opened 2018-04-11
Add Subbasin Container; Refactor Result Models

arottersman:
## Overview

This PR gets the subbasin view ready to display results in the tables. 

The refactoring and changes are described in the commit messages, but most notably this PR

* moves the gwlfe error off of a scenario and on to result models so that subbasin errors don't muddy the regular model views
* adds mapshed error attributes to the scenario so that we can catch when mapshed fails. Previously mapshed failure would make Gathering Data spin forever. Now there will be a proper error.
* Adds the subbasin model to the GWLFE model collection
* Refactor the subbasin container view to reuse the existing polling state/result view set up
* Add AOI info and blank tabs to the subbasin view 

Connects #2650 

### Demo

xxvv1nc49d

## Notes
* Existing mapshed projects won't be subbasin modelable until #2750 

## Testing Instructions

 * Pull and bundle
* Select a HUC-10 and progress it to subbasin modeling. Confirm you end up on a view with the aoi details and empty tabs once the polling completes.

* Put the browser offline or otherwise cause a failure during the mapshed/Gathering data step. Confirm you see an error

* Briefly test TR-55 and confirm there are no regressions

### DIFF
--- a/scripts/toggle_feature.sh
+++ b/scripts/toggle_feature.sh
@@ -32,8 +32,9 @@ enableFeature() {
 RESTART_APP=0
 for n; do
     case $n in
-        subbasin)
-            enableFeature subbasin
+        # Replace 'example' with the feature you to be togglable
+        example)
+            enableFeature example
             shift
             ;;
         clear)
--- a/src/mmw/js/src/modeling/gwlfe/subbasin/models.js
+++ b/src/mmw/js/src/modeling/gwlfe/subbasin/models.js
@@ -1,0 +1,19 @@
+"use strict";
+
+var Backbone = require('../../../../shim/backbone');
+
+var SubbasinTabModel = Backbone.Model.extend({
+    defaults: {
+        displayName: '',
+        name: '',
+    },
+});
+
+var SubbasinTabCollection = Backbone.Collection.extend({
+    model: SubbasinTabModel,
+});
+
+module.exports = {
+    SubbasinTabModel: SubbasinTabModel,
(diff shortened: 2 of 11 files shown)


arottersman:
I broke a unit test. Will fix momentarily!

arottersman:
This is ready for review now.

kellyi:
Setting this up to review!

kellyi:
Aside from that everything is working. We chatted briefly about this but: the perpetual "gathering data" on `MAPSHED data` failures is how the prior modeling tabs work too.

Seems like it's doable to fix just this subbasin tab to show an error instead of "gathering data" when the `SUBBASIN MAPSHED data` step fails. I can take another look at this when you'd like -- and we should probably make another enhancement card for dealing with the non-subbasin mapshed failures.

arottersman:
Thanks for catching that. The errors weren't working because they populate on the project, not the scenario. The fixup should address both the regular mapshed and subbasin mapshed error cases!

kellyi:
Great -- taking another look at this!

kellyi:
This is working very well! I did notice one case in which the errors are not fully propagated through to the UI, viz.:

If you turn off the network before clicking "View attenuated subbasin results" the network request will fail but the spinner will keep going. Same thing happens with Mapshed. We should handle this in another card though, as it's outside the scope of this one.

arottersman:
Thanks for all your testing, Kelly! Just squashed the fixups and will merge after this builds.

Issue #2726, opened 2018-03-13
Sub-basin: Add List table for huc-12 results


arottersman:
Under the "HUC-12" tab in the sub-basin view, show the total loads per huc-12:

<img width="219" alt="screen shot 2018-03-13 at 3 50 10 pm" src="https://user-images.githubusercontent.com/7633670/37370364-582ad08c-26e2-11e8-9694-4792eb053da2.png">

No need to add map or row interaction for this card.
https://app.goabstract.com/projects/76ed1390-dea8-11e7-a27c-0f868efbc56f/branches/master/commits/a49748bccbdf373c6c9c88eae15b2ee1a73d5e71/files/7F4A25A6-1E0D-48A0-B999-8629B42E4E23/layers/DB560561-C085-4BB7-9830-6B0A1B80173E

aufdenkampe:
For the "HUC-12 List" tab (formerly "HUC-12") when viewing **_AoI-level model results_** for Water Quality, please mimic the following table, which has extra columns to the one shown above. The live mockup can be found in the [AoI or HUC12 Source Details Table](https://docs.google.com/spreadsheets/d/1OzuN-3SV3_CHrVDoHSRd7ukGMCvVrWpUnnMVik_ZpT8/edit#gid=0) Google Sheet.

NOTE: 
* this table is _nearly identical to the NHD+ Catchment list table_ referenced in #2730, except for the first few columns.
* this table mimics the AoI-level and HUC-12-level Sources tables_ referenced in #2724 & #2729 respectively.

<img width="911" alt="screen shot 2018-04-04 at 11 37 19 am" src="https://user-images.githubusercontent.com/5166036/38321240-954cbbe4-37fc-11e8-957d-4e9016fd77b9.png">

Issue #2651, opened 2018-02-08
Sub-basin: Show HUC-12s on map


arottersman:
In hotspot mode, show the AOI's contained HUC-12's on the map. On huc-12 row table hover, highlight the corresponding shape. On shape hover, highlight the corresponding the row.

![screen shot 2018-02-08 at 9 44 09 am](https://user-images.githubusercontent.com/7633670/35980298-5cee4430-0cb8-11e8-9581-d0d817d21f71.png)
https://app.goabstract.com/projects/76ed1390-dea8-11e7-a27c-0f868efbc56f/branches/master/commits/a49748bccbdf373c6c9c88eae15b2ee1a73d5e71/files/7F4A25A6-1E0D-48A0-B999-8629B42E4E23/layers/9B372828-3221-410C-833F-A590A5B69E63

PR #2780, opened 2018-04-17
Add Subbasin HUC-12 Totals Table

arottersman:
## Overview

Add the HUC-12 totals table to the subbasin view. Because this table includes the HUC-12's common name, it was necessary to add a new endpoint to fetch that information (it's not part of the result until that point). I added the HUC-12 shape to the new endpoint as well because we will need it for #2651. 

Connects #2726 

### Demo

screen shot 2018-04-17 at 10 19 40 am
screen shot 2018-04-17 at 10 19 52 am
screen shot 2018-04-17 at 2 18 07 pm

## Notes
I originally implemented this by making the `get_huc12s` db query in the original gwlfe task to avoid the additional complexity of more requests. I realized, however, that anything we add to the result, like the shape, will then be stored on the job table as the output and on the scenario in the results. Having all the huc-12 geometries duplicated in these places seemed bad considering some huc8s can have upwards of 100 of them.

## Testing Instructions

 * Pull, restart celery, and bundle
 * Select a huc-10 in the app and proceed to subbasin modeling with it (Analyze -> Multi year -> Water Quality -> Attenuated Results)
 * View the HUC-12 tab and confirm the new table has all its values

### DIFF
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -182,7 +182,7 @@ def format_subbasin(huc12_gwlfe_results, srat_catchment_results, gmss):
             'SummaryLoads': summary_loads,
             'Catchments': {comid: format_catchment(result)
                            for comid, result
-                           in srat_huc12['catchments'].iteritems()}
+                           in srat_huc12['catchments'].iteritems()},
         }
 
     aggregate = {
--- a/src/mmw/apps/modeling/tasks.py
+++ b/src/mmw/apps/modeling/tasks.py
@@ -195,9 +195,9 @@ def format_subbasin(huc12_gwlfe_results, srat_catchment_results, gmss):
                                             .next()['inputmod_hash'],
     }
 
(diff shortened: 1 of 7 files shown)


src/mmw/apps/modeling/tasks.py
```diff
@@ -195,7 +195,7 @@ def add_huc12(srat_huc12, aggregate):
                                             .next()['inputmod_hash'],
     }
 
-    aggregate['HUC-12s'] = {huc12_id: add_huc12(result, aggregate)
+    aggregate['HUC_12s'] = {huc12_id: add_huc12(result, aggregate)
```
rajadain: Elsewhere (database, cache keys, layer configuration) we just use `huc12` (without hyphen or underscore). Can we do that for simplicity? Like `HUC12s`?

arottersman: Yes, great idea. I'll remove the underscore before merging.

Issue #2760, opened 2018-04-03
Replace Python suds dependency with suds-jurko


emiliom:
The app seems to be using the very old and deprecated `suds` package:  
https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/requirements/base.txt#L23
suds is used for SOAP implementation, for CUAHSI HIS interactions. As mentioned in #2756, this particular package stopped being developed many years ago, and is known to have memory problems. Its use should be replaced with [`suds-jurko`](https://bitbucket.org/jurko/suds), a fork that's maintained more actively. 

It's a simple, drop-in replacement, with no changes to code needed.

Note that `ulmo` already lists `suds-jurko` in its [requirements.txt](https://github.com/emiliom/ulmo/blob/wml_values_md/requirements.txt) file, and ulmo is already listed in the app `requirements/base.txt` file, so suds-jurko might already be installed but not used when doing `import suds`. It may be that removing the `suds==0.4` line is all that's needed.

cc @aufdenkampe @ajrobbins @rajadain

aufdenkampe:
@ajrobbins & @rajadain, this seems to be an easy update that has a lot of potential to increase performance with minimal effort. Please implement.

aufdenkampe:
@rajadain just reported during today's meeting that using `suds-jurko` does indeed substantially improve performance, and that 5000 to 8000 km2 could be reasonable. Yay!

Issue #2756, opened 2018-04-03
Increasing CUAHSI WDC-related AOI search limit beyond 1500 km2


emiliom:
### I've investigated CUAHSI WDC search API performance issues, following up on our last Monitor MW call. Briefly, I:
- Reviewed existing, relevant catalog API's.
- Performed a series of API performance tests with 1&deg; x 1&deg; AOI boxes (8,500 - 10,700 km2) across the country, and compiled the results.
- Contacted CUAHSI (Tony Castronova and Martin Seul) to ask about their use of SOLR and what catalog API's were new (relative to our development efforts last Summer) and recommended.
- Reviewed previous Azavea work and findings on this topic last year, during BiG-CZ portal development efforts

### Summary of my findings and recommendations
- The only new relevant API is [`GetSeriesMetadataCountOrData`](http://hiscentral.cuahsi.org/webservices/hiscentral.asmx?op=GetSeriesMetadataCountOrData). Its response is consistently slower than the one we currently use, `GetSeriesCatalogForBox2`. The only near-term potential use I can foresee is its capability to return only a count of the number of series records it will return; that response is extremely fast, and may be used to guide further client actions (including self pagination).
- All catalog API's are leveraging SOLR
- No existing catalog API is paginated.
- **We should stick with using the current catalog API, `GetSeriesCatalogForBox2`**
- Azavea tests in Oct 2017 concluded that an 8,000 km2 was unworkable. The various tests involved issues of time outs, problems with the Python SOAP library suds, and an internal application code involving caching where a problem was occurring (it looks like that caching is no longer done)
- I did not encounter any actual failures either on the CUAHSI WDC server end or my client end.
- **I see no current reason why the AOI could not be safely enlarged to at least 3,000 km2, possibly 5,000 km2. Results will be slower, and we'll need to decide what's acceptable. I believe 3,000 km2 will not cause any unacceptable slowdowns to users.**
- CUAHSI is open to providing more performant API's, including direct access to SOLR requests. This will undoubtedly increase server and client performance. But it's unclear how long it'll take CUAHSI to do this, and whether Azavea has time/funds to make corresponding changes.
- There are strategies we can explore on the client side, and suggestions to CUAHSI for providing resources that could enable smarter client searching. But all would require development time.

### Detailed information and discussion of API tests and related previous Azavea assessments
[...]

PR #2784, opened 2018-04-18
Monitor: Increase Area of Interest Limit with suds-jurko

rajadain:
## Overview

Switches to the `suds-jurko` fork which allows slightly more efficient SOAP calls. We are tentatively increasing the CUAHSI search limit from 1500 sq km to 8000 sq km, which allows small HUC-8s and almost all HUC-10s.

Connects #2756 
Connects #2760 

### Demo

image

### Notes

Performance is still hit or miss. It sometimes works for larger shapes, then fails for the same shapes, then works again. The CPU usage of the `app` VM still goes up to 100% when interpreting the results, and can still fail for large payloads.

For future optimization, we can also give [`zeep`](https://github.com/mvantellingen/python-zeep) a try. 

## Testing Instructions

* Check out this branch and reprovision `app`, and `bundle`
* Open [:8000/](http://localhost:8000/) and select a large-ish shape (more than 1500 sq km)
* Go to the Monitor tab and search for something in CUAHSI
* Ensure you get some results

### DIFF
--- a/deployment/ansible/roles/model-my-watershed.app/tasks/dependencies.yml
+++ b/deployment/ansible/roles/model-my-watershed.app/tasks/dependencies.yml
@@ -13,3 +13,13 @@
   when: "['packer'] | is_in(group_names)"
   notify:
     - Restart mmw-app
+
+- name: Remove suds
+  pip:
+    name: suds
+    state: absent
+
+- name: Reinstall suds-jurko
+  pip:
+    name: https://bitbucket.org/jurko/suds/get/94664ddd46a6.tar.gz#egg=suds-jurko
+    state: forcereinstall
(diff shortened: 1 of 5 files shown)


caseycesari:
Code looks good, but I haven't been able to get a shape around 5,000 km2 to succeed. Also, the shape in the screenshot works for me only intermittently.

Can we consider a more modest bump? Like 2,500 km2?

rajadain:
After playing around with this a bit more, I think a limit of 5,000 km² works consistently. Can you check that on your end?

caseycesari:
I'm not seeing the same performance. For example, I can't get this shape to succeed. It is consistently timing out.

image

However, given that staging and production have more resources, the changes here might perform better once this change is deployed.

caseycesari:
This is an existing issue, but it will become more prevalent with larger shapes. If a lot of results are returned, like more than a couple hundred, the DOM starts to perform very slowly. I think this is because of the large number of result elements in the CUASHI list. It's probably not feasible that someone would scroll through hundreds or thousands of results, so we may want to paginate or infinite-scroll the results list.

rajadain:
Rebased on `develop`, squashed fixups. Merging shortly.

Issue #2718, opened 2018-03-10
EnviroDIY (from CUAHSI WDC): Add Source Data link in Details tab


emiliom:
As in NWIS services (eg, NWISUV), a web page exists for each each EnviroDIY site, and the web site URL can be easily constructed from the site code returned by the CUAHSI WDC catalog for the EnviroDIY site. For example, for site `WCC019` (`EnviroDIY:WCC019`), the corresponding "Source Data" link would be: http://data.envirodiy.org/sites/WCC019/

Adding this link following existing practices (the use of a Source Data button, when feasible) would add a lot of value to users, in sending them to the EnviroDIY data web site where they can explore and download all data for the site.

cc'ing @aufdenkampe so he can weigh in about this suggestion, and @horsburgh in case I'm missing something about the URL pattern at http://data.envirodiy.org

emiliom:
Forgot to add example from NWISUV. See site 01478120, whose Source Data link is https://waterdata.usgs.gov/nwis/uv/?site_no=01478120

horsburgh:
@emiliom - URL pattern looks good.

aufdenkampe:
@ajrobbins, this is an easy, high priority feature that we want to implement.

PR #2785, opened 2018-04-18
Add EnviroDIY source data link

rajadain:
## Overview

Add Source Data link to EnviroDIY results in CUAHSI

Connects #2718 

## Testing Instructions

* Check out this branch
* Select a shape and search for "EnviroDIY" in Monitor → CUAHSI.
* Ensure every EnviroDIY result has a Source Data link. Ensure it goes to a real functioning page.


rajadain:
Thanks for taking a look!

Issue #2767, opened 2018-04-05
Missing Required streamLines


arottersman:
https://rollbar.com/Azavea/MMW/items/302/

We assume all shapes will have streams
https://github.com/WikiWatershed/model-my-watershed/blob/46b52479e211e3b6ac03ecff271439c5a757feb8/src/mmw/apps/modeling/geoprocessing.py#L172

To recreate, draw a small shape that doesn't intersect any streams. Try to model it with mapshed.

ajrobbins:
Another failing project: https://staging.app.wikiwatershed.org/project/1251/, not sure if related

PR #2790, opened 2018-04-19
Allow MapShed to run for AoIs without streams

caseycesari:
## Overview

If a MapShed run is started for an AoI that does not contain streams, the stream related geoprocessing operation is removed from the list of jobs. Variables and functions that use the output of that job are modified in to account for the output not existing. This mostly involves setting certain values to zero.

Connects to #2767

### Demo

AoI w/o streams
image

AoI w/ streams
image

## Testing Instructions

- Select an AoI that does not contain any streams. A 1 km2 AoI somewhere in Center City works well.
- Proceed to the modeling stage and select mapshed.
- Verify the model succeeds and that the results are valid.
- Select another AoI that does contain streams. Select mapshed again and verify that the model succeeds and that the results are valid.
- Make sure to double-check that the results are valid in both scenarios.

### DIFF
--- a/src/mmw/apps/modeling/mapshed/tasks.py
+++ b/src/mmw/apps/modeling/mapshed/tasks.py
@@ -113,12 +113,23 @@ def collect_data(geop_results, geojson, watershed_id=None, weather=None):
     z['Prec'] = prcps
 
     # Begin processing geop_result
-    z['AgLength'] = geop_result['ag_stream_pct'] * z['StreamLength']
-    z['UrbLength'] = z['StreamLength'] - z['AgLength']
-    z['n42'] = round(z['AgLength'] / 1000, 1)
-    z['n46e'] = (geop_result['med_high_urban_stream_pct'] *
-                 z['StreamLength'] / 1000)
-    z['n46f'] = geop_result['low_urban_stream_pct'] * z['StreamLength'] / 1000
+
+    # Set stream related variables to zero if AoI does not contain
+    # any streams.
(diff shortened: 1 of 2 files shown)


src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -160,8 +171,14 @@ def collect_data(geop_results, geojson, watershed_id=None, weather=None):
...
                                    z['CN'], z['AEU'], z['AvKF'], z['AvSlope'])
 
-    z['LS'] = ls_factors(geop_result['lu_stream_pct'], z['StreamLength'],
-                         z['Area'], z['AvSlope'], ag_lscp)
+    # Use zeroed out stream variables if there are no streams in the AoI
+    if 'lu_stream_pct' in geop_result:
+        z['LS'] = ls_factors(geop_result['lu_stream_pct'], z['StreamLength'],
+                             z['Area'], z['AvSlope'], ag_lscp)
+    else:
+        zeroed_lu_stream_pct = [0 for i in range(0, 16)]
```
rajadain: Prefer:

```python
[0.0] * 16
```

for declaring a 16-index zeroed array, as done elsewhere:

https://github.com/WikiWatershed/model-my-watershed/blob/3902fb97e9cc114c0c035a4ec2b5ec2307f86a8e/src/mmw/apps/modeling/mapshed/calcs.py#L273

caseycesari: Thanks

Issue #2791, opened 2018-04-20
Subbasin Hover Only Works First Time


rajadain:
The hover interaction works very well the first time the subbasin mode is loaded. But if the user goes back to MapShed and then back to Subbasin, it doesn't work:

![2018-04-20 14 08 55](https://user-images.githubusercontent.com/1430060/39066896-b81a58ca-44a4-11e8-90ca-3cee1dd6f163.gif)

The failure happens in `bringToBack` and `bringToFront` invocations here:

https://github.com/WikiWatershed/model-my-watershed/blob/02661b382408ac9d3498f4231c8de312db1d625b/src/mmw/js/src/core/views.js#L1048-L1060

Also see related https://github.com/Leaflet/Leaflet/issues/2439 and https://github.com/Leaflet/Leaflet/issues/2200.

Make it so the interaction always works and doesn't error out.

arottersman:
Shares a point #2730

rajadain:
Not done yet. Still happening in some cases.

rajadain:
Shares points with #2655.

Issue #2789, opened 2018-04-19
Streamline Changing Scenarios while in Sub-basin mode


rajadain:
Currently when changing scenarios while sub-basin mode is active breaks the map, by leaving sub-basin elements on the map (such as HUC-12 boundaries). Ensure that when this happens, the next scenario is loaded in regular mode, and has no sub-basin elements on the map.

PR #2792, opened 2018-04-20
Close sub-basin view when changing scenarios

rajadain:
## Overview

Closes the sub-basin view when switching scenarios to alleviate any map issues resulting from remnant sub-basin artifacts.

Discovered #2791 while working on this, but ended up cutting that into a separate issue.

Connects #2789

### Notes

Switching between scenarios seems to take a pause, not sure if that's coming from `.hasView()` or something else, but I think it's tolerable right now. 🚀 🐦 

## Testing Instructions

* Open subbasin mode on a project
* Change scenarios / add a scenario from within subbasin mode. Ensure that in the next scenario you see MapShed, not subbasin.
* Ensure everything else still works.

### DIFF
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -959,6 +959,11 @@ var ResultsView = Marionette.LayoutView.extend({
         ));
 
         if (scenario) {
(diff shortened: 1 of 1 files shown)


rajadain:
Thanks for taking a look!

PR #2781, opened 2018-04-17
Add HUC-12 Level Subbasin View

arottersman:
## Overview

* Make the huc-12 table rows clickable.
* Add a Subbasin HUC-12 view over the subbasin AOI-level view.
* Make the SourcesTableView, subbasin ResultView and SubbasinTabContentView extendable and use them to fill out the HUC-12 level subbasin region

Connects #2728 
Connects #2729 

### Demo

qlabzqvwgc

screen shot 2018-04-17 at 7 03 12 pm

## Testing Instructions

 * Pull and bundle
 * Select a HUC-10 and progress to Subbasin View. Select the HUC-12 tab in subbasin view
 * Click a row in the HUC-12 table
 * Confirm you can see the huc-12s source results and can navigate back and forth through all the levels of views (Unattenuated model results <-> Subbasin AOI-Level <-> Subbasin HUC-12)

Issue #2730, opened 2018-03-13
Sub-basin: Add List Table for NHD+ Catchment results


arottersman:
Under the "Catchments" tab in the sub-basin HUC-12  view, add a table displaying the total loads, loading rates, and mean annual concentrations for each catchment of the HUC-12.

<img width="1008" alt="38377615-f1465ba2-38c0-11e8-94ec-7bb85a7be16a" src="https://user-images.githubusercontent.com/7633670/38559155-3f2b00c4-3ca0-11e8-9440-b3d2ca9f34cd.png">

Total Loads should correspond to the `TotalLoadingRates` for each catchment, Loading Rates should be calculated as the `TotalLoadingRates` over the area of the catchment (we'll have to look this up from the `nhdcatchments` table via the catchment's comid). The Mean Annual Concentrations are in `LoadingRateConcentrations`.

[Abstract wireframes here, use for style but not table content](https://app.goabstract.com/projects/76ed1390-dea8-11e7-a27c-0f868efbc56f/branches/master/commits/a49748bccbdf373c6c9c88eae15b2ee1a73d5e71/files/7F4A25A6-1E0D-48A0-B999-8629B42E4E23/layers/10801C0C-5E75-4492-A3EC-DAB0ECA3C56F)

aufdenkampe:
For the "Catchments" List tab when viewing **_HUC-12-level model results_** for Water Quality, please mimic the following table, which has extra columns to the one shown above. The live mockup can be found in the [Catchments List Table](https://docs.google.com/spreadsheets/d/1OzuN-3SV3_CHrVDoHSRd7ukGMCvVrWpUnnMVik_ZpT8/edit#gid=1066341079) Google Sheet.

NOTE: 
* this table is _nearly identical to the HUC-12 list table_ referenced in #2726, except for the first few columns.
* this table mimics the AoI-level and HUC-12-level Sources tables_ referenced in #2724 & #2729 respectively.

<img width="812" alt="screen shot 2018-04-04 at 11 40 38 am" src="https://user-images.githubusercontent.com/5166036/38321412-3ca7afb6-37fd-11e8-9ead-e3555511a806.png">

aufdenkampe:
[...]

PR #2794, opened 2018-04-20
Add Subbasin NHD+ Table

arottersman:
## Overview

Similar to #2781 where we got HUC-12 details — get catchment details and geometries and display them in the new catchment table. Also clean up some map-timing bugs.

Connects #2730
Connects #2791

### Demo

screen shot 2018-04-20 at 5 34 31 pm
screen shot 2018-04-20 at 5 34 21 pm

## Testing Instructions

 * Pull and `bundle`
 * In subbasin view, select a huc-12 and confirm the catchment table is populated
 * Confirm if you select the same huc-12 the catchment details request doesn't execute again
 * Toggle between the various views in various states (eg Analyze toggle while viewing catchments, exit attenuated and then view again). Confirm there are no errors in the console and nothing breaks on the second viewing.

### DIFF
--- a/src/mmw/js/src/modeling/gwlfe/subbasin/views.js
+++ b/src/mmw/js/src/modeling/gwlfe/subbasin/views.js
@@ -2,12 +2,14 @@
 
 var Marionette = require('../../../../shim/backbone.marionette'),
     $ = require('jquery'),
+    _ = require('lodash'),
     App = require('../../../app'),
     models = require('./models'),
     resultTmpl = require('./templates/result.html'),
     tableTabContentTmpl = require('./templates/tableTabContent.html'),
     tableTabPanelTmpl = require('./templates/tableTabPanel.html'),
     huc12TotalsTableTmpl = require('./templates/huc12TotalsTable.html'),
+    catchmentTableTmpl = require('./templates/catchmentTable.html'),
     sourcesTableTmpl = require('./templates/sourcesTable.html');
 
 var ResultView = Marionette.LayoutView.extend({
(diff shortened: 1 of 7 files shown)


rajadain:
Looks like the HUC / Catchments detail table isn't being initialized correctly anymore:

image

It is initialized correctly on `develop` / staging:

image

There may be a missing `.bootstrapTable()` call somewhere. Will see if I can find it.

arottersman:
Hm, I noticed that intermittently before I moved that bootstrap init call to `onRender`. I'd wager we still need it `onAttach` and need to be a bit more clever about where we call it again on re-render.

src/mmw/js/src/modeling/gwlfe/subbasin/views.js
```diff
@@ -224,6 +226,85 @@ var Huc12TotalsTableView = Marionette.ItemView.extend({
...
+    onRender: function() {
+        $('[data-toggle="table"]').bootstrapTable();
+    },
+    initialize: function() {
+        var self = this;
+        self.catchmentDetails = App.currentProject.get('subbasins').getActive().get('catchments');
+        if (this.catchmentDetails.isEmpty()) {
+            self.listenToOnce(self.catchmentDetails, 'add', this.setupCatchmentDetails, this);
+        } else {
+            this.setupCatchmentDetails();
```
rajadain: Calling `render` within `initialize` feels like a mistake ... but I'm not seeing any errors in the console. It is possible that this is the cause of the broken bootstrap table. Consider moving it to a custom function and calling that directly here, and also from within `onRender`.

arottersman:
Great suggestions re: promises. I've pushed one fixup to address the fetching, and another for the table renders (I believe the issue is fixed now though it doesn't look all that pretty).

> Sending the list of catchments as a URL encoded JSON array is an interesting choice, but it works so I'm not going to challenge it right now

Definitely with you on this — I wanted it to stay a GET request like its HUC-12 sibling, and for some reason didn't want to consider how the list should come in. Encoding a JSON array happens easily and predictably so it seems okay for now! 🚀 🐦

src/mmw/js/src/modeling/models.js
```diff
@@ -825,14 +825,17 @@ var SubbasinDetailModel = Backbone.Model.extend({
 
     fetchCatchmentsIfNeeded: function(comids) {
         var catchments = this.get('catchments');
-        if (!catchments.isEmpty() || this.fetchCatchmentsPromise) { return; }
+        if (!catchments.isEmpty() || this.fetchCatchmentsPromise) {
+            return $.when();
```
rajadain: Sorry, I missed this earlier, but if `this.fetchCatchmentsPromise` exists, we should probably return that instead of `$.when()`:

```js
if (!catchments.isEmpty() || this.fetchCatchmentsPromise) {
  return this.fetchCatchmentsPromise || $.when();
}
```

arottersman: Oh, that's my bad. I'll fix before merging.

arottersman:
@azavea-bot rebuild

PR #2795, opened 2018-04-21
Upgrade Omgeo to correctly resolve suds-jurko

rajadain:
## Overview

Previously, python-omgeo would bring in suds as a dependency requiring we remove it and reinstall it. This led to various build issues.

By upgrading python-omgeo, we make it so that suds is never brought in, and doesn't need to be uninstalled for suds-jurko to work correctly. This fixes the build issues.

Connects #2760

## Testing Instructions

* Checkout this branch, destroy and recreate `app`
* Run `vagrant ssh app -c 'sudo -H pip list | grep suds'` and ensure you only see `suds-jurko` and no `suds` dependency
* Open the app and ensure the geocoder works correctly
* Go to the Monitor tab after selecting a shape and search for something CUAHSI. Ensure you see results.

### DIFF
--- a/deployment/ansible/roles/model-my-watershed.app/tasks/dependencies.yml
+++ b/deployment/ansible/roles/model-my-watershed.app/tasks/dependencies.yml
@@ -13,13 +13,3 @@
   when: "['packer'] | is_in(group_names)"
   notify:
     - Restart mmw-app
-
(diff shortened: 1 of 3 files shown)


rajadain:
Thanks for taking a look!

Issue #2774, opened 2018-04-13
Calibrate Subbasin Timeouts


arottersman:
Large sub-basin shapes will take longer than the default 45sec frontend time out for each part of the model. Calibrate a special sub-basin timeout based on staging performance.

arottersman:
If we want to use the Schuylkill HUC-8 as a target:
```
Staging MapShed: ~60sec uncached, 25sec cached | Local Mapshed: ~80sec uncached, 40sec cached
Staging GWLF-E:   ~85sec                       | Local GWLF-E:  ~150sec
```
Maybe 90 or 100sec would be appropriate for each portion of the model? This would prevent local UI execution of the Schuylkill, but give a good margin for staging.

/cc @ajrobbins

Issue #2764, opened 2018-04-04
Add new loading indicator messaging for subbasin


ajrobbins:
Due to the large number of subbasins in some HUCs, total execution times for subbasin modeling may range from 30-90 seconds. Since this is farily long loading time for a web app, we should implement some more helpful messaging for the loading indicator, including:
- fraction of steps complete
- name of step in process that's currently running
- up-front warning about likely time range

mockup (note that the loading indicator styling itself should not change):
![image 15](https://user-images.githubusercontent.com/12868342/38316588-11d6924c-37f9-11e8-9332-b7efdc3aae3c.png)

arottersman:
If we need to get the in-flight progress for a celery task, we can use something like this: http://docs.celeryproject.org/en/latest/userguide/calling.html#on-message

rajadain:
Another alternative may be to simply report the polling stage, allowing the user to see how far along the process is. In some cases the process may finish before polling times out, and when it doesn't we'll be more transparent regarding the wait.

@jfrankl Do you have any suggestions for a progress bar library or something that could support ~150 ticks? I suppose it could be done with simple CSS too.

arottersman:
Here's the bootstrap 3 progress bars if they're of interest https://getbootstrap.com/docs/3.3/components/#progress

PR #2796, opened 2018-04-23
Subbasin: Increase front-end timeout to 160s

rajadain:
## Overview

Previously the timeout was set to 45s. This was problematic because the front-end timeout didn't help anyone. It doesn't help the user, because it is giving up for them. If the user wants to give up, they can do that. If they want to wait, they should be able to do that too.

Furthermore, it doesn't help the server, since the timeout doesn't actually stop the Celery job from running and using up server resources. Even when the front end has given up, we keep the job running.

Increasing the front end timeout from 45s to 160s for all use cases is a simple, harmless adjustment. It will not increase our server loads at all, and allow patient users to see results (subbasin and otherwise) that they didn't before.

Connects #2774 

### Demo

screen shot 2018-04-21 at 5 21 46 pm

### Notes

160s is really an unreasonable time to expect users to wait on the internet, especially if they don't appreciate the processing happening on the servers. We need to improve the UI for waiting as a result of this.

#2764 is planned to add a progress bar to the wait message, which should help the wait a bit.

## Testing Instructions

* Run Analyze, Mapshed, and Subbasin on a large area of interest. Ensure it doesn't timeout.

### DIFF
--- a/src/mmw/js/src/core/models.js
+++ b/src/mmw/js/src/core/models.js
@@ -469,13 +469,13 @@ var LayerTabCollection = Backbone.Collection.extend({
 var TaskModel = Backbone.Model.extend({
     defaults: {
         pollInterval: 1000,
-        /* The timeout is set to 45 seconds here, while in the
-           src/mmw/apps/modeling/tasks.py file it is set to 42
-           seconds.  That was done because the countdown starts in the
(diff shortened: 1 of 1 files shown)


rajadain:
A going away present.

Issue #2761, opened 2018-04-03
Layers > Observations: Is Historical Data time series plotting disabled?


emiliom:
At the 3/20 Monitor MW call we discussed the problems with displaying historical time series plots available through Layers > Observations. 

It appears that the functionality is currently disabled on the App -- as opposed to the problem being solely on the UW server end, where requests for very long time series are not working.

@rajadain and/or @arottersman can you confirm that's the case? If so, please reenable it on the test app so we can run some tests.

I've confirmed that the UW web service is functioning normally for a request for a relatively short time series, such as this one (about 65 days): http://www.wikiwatershed-vs.org/services/get_asset_info.php?opt=data&asset_type=siso&asset_id=USGS_01408168&var_id=H1_RiverStage&units_mode=v2

Once our tests confirm that we can make this work with shorter time series, I'll make changes on the UW end so that all time series are short enough.

cc @aufdenkampe @ajrobbins

aufdenkampe:
Decided on today's meeting that @mmcfarland & @rajadain will investigate

aufdenkampe:
From our discussion today, it sounds like Azavea has done what they need to do on their end, and now it's back to Emilio to set up auto trimming of datasets on his end to contain < 1 year of data.

emiliom:
> and now it's back to Emilio to set up auto trimming of datasets on his end to contain < 1 year of data.

I've trimmed the data on my end and did a bunch of spot testing on the staging app. With 1 year, the USGS data was fine, but the other two data sources (NOAA NOS and DEOS) were not. So I trimmed everything to 6 months, and now all sites seem to be able to generate complete "historical" time series plots.

aufdenkampe:
@emiliom, thanks!

ajrobbins:
I just spot checked several sites in the staging app, and it's looking good! Closing this issue.

emiliom:
Final note: I've set up our database cache to automatically trim data that's a little over 6 months (190 days). This happens daily.

PR #2797, opened 2018-04-23
Fix Observations Layer tab data loading

rajadain:
## Overview

This bug was introduced in 02b5a7fb. Allows the Observations Layer tab to load historical data again.

Connects #2761

### Notes

The historical data loading still often fails because of server issues.

## Testing Instructions

* Open the app and the javascript console
* Go to the observations tab in the layer picker and enable a layer
* Click an observation point and click through to Historical Data
* Ensure there are no javascript errors in the console (disregard any server side errors)

### DIFF
--- a/src/mmw/js/src/core/modals/views.js
+++ b/src/mmw/js/src/core/modals/views.js
@@ -634,7 +634,7 @@ var PlotView = ModalBaseView.extend({
     plotMeasurement: function(varId) {
         var self = this,
             series = self.model.get('seriesMap')[varId],
-            dataUrl = settings.vizerUrls.variable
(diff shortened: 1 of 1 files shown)


kellyi:
Setting this up!

kellyi:
Was able to get River Gauge historical data to load:

screen shot 2018-04-25 at 11 11 16 am

The other observation layers seemed to send back 500 errors from the server -- not sure whether those originate from MMW or from the observations service?

rajadain:
Correct: we are not touching the server side at all here, just the client side. We may move away from the backend service used for this in the future, so I'm not going to make a card for the intermittent failures right now. It is a known issue.

Thanks for taking a look!

PR #2786, opened 2018-04-18
Show Subbasin Huc-12s On Map

arottersman:
## Overview
Add the huc-12 subbasin shapes to the map. They have two-way interactivity: you can active their hover state by hovering over them on the map or on the huc-12 totals table. You can select one via clicking the map or clicking their row in the huc12 totals table. 

Connects #2651 

### Demo

[ ignore the error on the analyze page in this gif. It isn't related. ]
3bn3ac1ejw

## Testing Instructions

 *  Pull this branch and `./scripts/bundle.sh`
 * Proceed to subbasin modeling and confirm you see shapes appear on the map

Some things to try:
 * hover over the shapes
 * hover over the huc12 total table rows
 * click the shapes
 * click the huc12 total table rows
 * toggle the analyze/monitor/modeling view
 * navigate away from the model view while the shapes are on the map
 * exit subbasin mode and go back into it

Issue #2655, opened 2018-02-08
Sub-basin: Add "Related Layers" To HUC-12 View 


arottersman:
At the HUC-12 level of hotspot mode, display a "related layer" toggle like we do on analyze. Selecting a water quality metric should render a visualization of that metric on the map via colored catchment polygons and colored streams. Use the same color breaks we use for the DRB water quality streams and coverage layers.

NB: Rendering so many polygons might not be performant, and might fail. Hopefully only doing this at the HUC-12 level will keep the number down, but we should point this card accordingly.

Mimic the water quality related layer:
![screen shot 2018-02-07 at 4 01 10 pm](https://user-images.githubusercontent.com/7633670/35980639-3ab155be-0cb9-11e8-92f4-035e22de43a2.png)

![screen shot 2018-02-07 at 4 01 13 pm](https://user-images.githubusercontent.com/7633670/35980619-30aeedd8-0cb9-11e8-94f4-54e7133e161a.png)

[Wireframes](https://app.goabstract.com/projects/76ed1390-dea8-11e7-a27c-0f868efbc56f/branches/master/commits/c97f50abf47968d7ccb1679a20b6af9b2b0e7ff3/files/7F4A25A6-1E0D-48A0-B999-8629B42E4E23/layers/F356B552-990A-446E-B648-D891E255F8C8)

rajadain:
Comparing the breaks used in DRB Water Quality rasters vs the values in subbasin:

![image](https://user-images.githubusercontent.com/1430060/39775783-e673554c-52cc-11e8-95de-413113b6bb42.png)

TotalP is always going to be more than 1.20 kg/y, so it'll all be the same black. Is that expected / desired? Should we be using other breaks? /cc @ajrobbins

rajadain:
Also, any suggestions for where the legend will go once we have the colors in?

Issue #2654, opened 2018-02-08
Sub-basin: Show catchments and stream geoms on map


arottersman:
At the HUC-12 level, show the HUC-12 and its contained NHD Plus catchments and catchment stream geometries. On catchment row hover, highlight that shape. On shape hover, highlight that catchment row.

![screen shot 2018-02-08 at 9 44 15 am](https://user-images.githubusercontent.com/7633670/35980376-955ea5bc-0cb8-11e8-8b8a-f97f4a52e036.png)

arottersman:
Blocked by #2730

PR #2798, opened 2018-04-24
Subbasin Catchments On Map

arottersman:
## Overview

Similar to #2786, add the catchment geometries and streams to the map. They are two-way interactive with the catchment table's hover state.

Styling is temporary — the streams and catchment fill will be determined by the catchment's loads in #2655.

Connects #2654 

### Demo

5vjsb7oyvs

## Testing Instructions

 * Pull and `bundle`
 * For a subbasin project in subbasin view, test that at the HUC-12 level the catchments render with their streams
   * hover on the map highlights the map and table
   * hover on the catchment table highlights the map and table
* Confirm you can flip between Analyze, Monitor and Model without the shapes carrying over or not reappearing when re-entering the model view.
* Confirm you can add scenarios
* Confirm you can leave subbasin view and enter it again without console errors

### DIFF
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -80,6 +80,42 @@ var subbasinHuc12ActiveStyle = {
     color: '#dbba29',
     weight: 5,
     opacity: 1,
+    fill: false,
+};
+
+var subbasinCatchmentStyle = {
+    stroke: true,
+    color: 'grey',
+    weight: 2,
+    opacity: 1,
+    fill: true,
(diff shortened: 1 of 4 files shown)


rajadain:
Saw a similar bug as in the HUC-12 PR, where cached catchments aren't rendered correctly on the map:

arottersman:
I've added fixup to clear the catchment map state every time we try to show a new set of catchments. Thanks for that catch!

rajadain:
@azavea-bot rebuild

Issue #2749, opened 2018-03-30
Beef Up Development VMs


arottersman:
This is a placeholder card to gather improvements for our development environment. Working on the Sub-basin feature has exposed the need for at least some additional memory — probably for our services and worker VMs.

We could also consider syncing the VM clocks for easier celery task timing, and adding second CPU to the worker to better match staging:
[beefy_vm.patch.txt](https://github.com/WikiWatershed/model-my-watershed/files/1864175/beefy_vm.patch.txt)

PR #2803, opened 2018-04-25
Give VMs More Resources

rajadain:
## Overview

Increase VM capacities to keep up with expanding MMW capabilities.

Expands development VM sizes to match what's on staging so local testing is more accurate. Ups staging app VM size to match production so that large monitor queries are parsed without crashing. Sync VM times so that any timing script running across VMs (app, worker) is consistent.

Tagging @hectcastro to verify that this (in addition to changes in the deployment configuration file) will upgrade staging to t2.small with 2GB RAM, as discussed with @mmcfarland and @ajrobbins yesterday.

Connects #2749 

### Notes

I will edit the deployment configuration files once this is merged.

## Testing Instructions

* Reload `app`, `worker`, and `services` (provisioning should not be necessary)
* Run Monitor on the Brandywine-Christina HUC-08 near Wilmington, DE and ensure it completes
* Run Subbasin on a HUC-10 and ensure it completes

### DIFF
--- a/deployment/cfn/application.py
+++ b/deployment/cfn/application.py
@@ -83,7 +83,7 @@ class Application(StackNode):
         'StackType': 'Staging',
         'StackColor': 'Green',
         'KeyName': 'mmw-stg',
-        'AppServerInstanceType': 't2.micro',
+        'AppServerInstanceType': 't2.small',
         'AppServerInstanceProfile': 'AppServerInstanceProfile',
         'AppServerAutoScalingDesired': '1',
         'AppServerAutoScalingMin': '1',
@@ -128,7 +128,7 @@ class Application(StackNode):
         ), 'RDSPassword')
 
         self.app_server_instance_type = self.add_parameter(Parameter(
-            'AppServerInstanceType', Type='String', Default='t2.micro',
+            'AppServerInstanceType', Type='String', Default='t2.small',
             Description='Application server EC2 instance type',
             AllowedValues=EC2_INSTANCE_TYPES,
             ConstraintDescription='must be a valid EC2 instance type.'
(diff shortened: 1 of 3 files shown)


deployment/cfn/application.py
```diff
@@ -83,7 +83,7 @@ class Application(StackNode):
         'StackType': 'Staging',
         'StackColor': 'Green',
         'KeyName': 'mmw-stg',
-        'AppServerInstanceType': 't2.micro',
+        'AppServerInstanceType': 't2.small',
```
hectcastro: These are OK to change to keep things in sync, but they are effectively treated as defaults in the event that `AppServerInstanceType` isn't defined anywhere up the chain. I think a corresponding change in `default.yml.example` would complete the syncing of possible defaults.

That said, the persistence of this is driven by a local copy of `default.yml` on Jenkins (for staging) and the file share (for production). Changes would need to be made to those for future deployments to use a `t2.small`. (We can do this for staging now and test an out-of-band deployment via Jenkins to confirm it works.)

Lastly, the change from a `t2.micro` -> `t2.small` makes us come up short based on our existing EC2 reservations. Each `t2.micro` equates to 2 `t2.nano` and each `t2.small` equates to 4 `t2.nano`, so I think that we'd need to purchase 4 `t2.nano` units to have full coverage again.

We'd probably want to get signoff on that before doing it since it ends up being a  1 year commitment.

See: https://aws.amazon.com/blogs/aws/new-instance-size-flexibility-for-ec2-reserved-instances/

hectcastro: `t2.small` application server is up on staging now.

hectcastro: Sounds good. I'll hold off on purchasing until we get a confirmation from Stroud.

Issue #2731, opened 2018-03-13
Sub-basin: Handle Stale BMPs


arottersman:
When a BMP is added while the user is in sub-basin mode, apply the BMP to the sub-basin model. When the user navigates out of sub-basin mode, the unattenuated model should re-run with the new BMPs. The same should happen if a user already has sub-basin results, and adds BMPs in the unattenuated mode.

arottersman:
This may be handled by the `inputmod_hash` when moving from Subbasin -> unattenuated.

Issue #2542, opened 2017-11-30
Apply Area-Weighted BMPs To Subbasins


arottersman:
EDIT

We will need to weight the following modifications on each HUC-12s stream `Length` value:

    'buffer_strips',
    'streambank_fencing',
    'streambank_stabilization'

The rest are already represented as percents.

--------------

While there are future plans to allow drawing BMPs over specific areas for Mapshed, for now we'll handle BMPs under subbasin modeling as follows:

- If BMP is given as percentage of AEU, overwrite the value for each subbasin GMS with the new percentage
- If BMP value is a specific area/length, calculate what percentage the new given area is of the shape's total area. Set each subbasin's corresponding value to that percentage multiplied by the subbasin's original value.

  For example if the shape's cover crop area is 10ha, and the user inputs 1ha as the new desired cover crop area, multiply each of the subbasins' cover crop areas by 0.10.

PR #2804, opened 2018-04-26
Subbasin Weighted BMPs + Misc. Fixes

arottersman:
## Overview

* Make modifications work as expected in subbasin mode
  * Weight absolute stream length modification types
  * Don't rerun model for subbasin or unattenuated until that view is active

* Fix GMS download
    * Mods need to be applied before download and we need to read the base GMS from the active project

* Sum catchment table
   * We can't re-use the HUC-12 sums because catchments seem to have different results, and their areas extend past the HUC-12

Connects #2542 
Connects #2731 

### Demo

vn7dgnu5bx

#### Modifications

 * Pull, bundle
* Apply this patch so you can see what mods are being applied [print_modifications.patch.txt](https://github.com/WikiWatershed/model-my-watershed/files/1952101/print_modifications.patch.txt)
* Restart and debug celery
 * Start a subbasin project and confirm you can add modifications while in regular mode and while in subbasin mode. Adding in regular mode should kick off a subbasin re-run when you next enter subbasin mode and vice versa.
* There should be a new spinner indicating when subbasin is polling
* In the celery logs you should see the weighted BMPs printed

#### Downloading GMS
* Confirm you can download the GMS file when the only scenario is Current conditions and when there are multiple scenarios and you download via dropdown.

#### Catchment Summation
* The "entire area" row of the catchment table should still have values. They will likely be larger than what they used to be.

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -343,15 +343,15 @@ var ProjectModel = Backbone.Model.extend({
         return createTaskResultCollection(this.get('model_package'));
     },
 
-    fetchResultsIfNeeded: function(isSubbasinMode) {
+    fetchResultsIfNeeded: function() {
         var promises = [];
 
         this.get('scenarios').forEach(function(scenario) {
-            promises.push(scenario.fetchResultsIfNeeded(isSubbasinMode || false));
+            promises.push(scenario.fetchResultsIfNeeded());
         });
 
         this.get('scenarios').on('add', function(scenario) {
-            scenario.fetchResultsIfNeeded(isSubbasinMode || false);
+            scenario.fetchResultsIfNeeded();
         });
 
         return $.when.apply($, promises);
@@ -1072,6 +1072,7 @@ var ScenarioModel = Backbone.Model.extend({
         modifications: null, // ModificationsCollection
         modification_hash: null, // MD5 string
         active: false,
+        is_subbasin_active: false,
         job_id: null,
         poll_error: null,
         results: null, // ResultCollection
@@ -1111,6 +1112,7 @@ var ScenarioModel = Backbone.Model.extend({
         this.updateInputModHash();
 
         this.on('change:project change:name', this.attemptSave, this);
+        this.on('change:is_subbasin_active', this.queueFetchResultsIfNeeded, this);
         this.get('modifications').on('add remove change', this.updateModificationHash, this);
         this.debouncedFetchResults = _.debounce(_.bind(this.fetchResults, this), 500);
         this.get('inputs').on('add', _.bind(this.fetchResultsAfterInitial, this));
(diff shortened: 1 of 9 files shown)


rajadain:
If I add 59% Nutrient Management to the Lower Schuylkill HUC-10:

image

But this reports no difference between original and subbasin'd modifications:

```
original modifications
[{'n28b': 59.376456752067476}]
weighted modifications
[{'n28b': 59.376456752067476}]
```

Is this expected?

arottersman:
Yes, modifications that are given as percent values, aren't weighted and will be identical to the original results.

The only ones you'll see a difference for are

screen shot 2018-04-26 at 3 06 27 pm
screen shot 2018-04-26 at 3 06 22 pm

rajadain:
This is hard to reproduce, but every once in a while I'll run into this issue of catchments not loading on the map correctly:

image

src/mmw/js/src/modeling/views.js
```diff
@@ -880,7 +884,10 @@ var ScenarioToolbarView = Marionette.CompositeView.extend({
     },
 
     templateHelpers: function() {
-        var gisData = this.currentConditions.getGisData().model_input,
+        // We don't need to apply modifications to gis_data here
+        // because it's are only downloadable from this view if
```
rajadain: > because it's are only downloadable from this view if

Is there a word missing?

src/mmw/js/src/modeling/models.js
```diff
@@ -1345,6 +1350,13 @@ var ScenarioModel = Backbone.Model.extend({
         return taskModel.start(taskHelper);
     },
 
+    queueFetchResultsIfNeeded: function() {
+        if (this.fetchResultsPromise) {
+            return this.fetchResultsPromise.always(this.fetchResultsIfNeeded);
```
rajadain: It seems like the intent of this line is to queue up `fetchResults` requests, but currently, since `this.fetchResultsPromise` is not updated, if `queueFetchResultsIfNeeded` is called multiple times before the first promise resolves, all subsequent fetches will be queued up after the first one, and be fired simultaneously rather than in order.

To fire them in order, prefer doing something like this:

```js
if (this.fetchResultsPromise) {
  this.fetchResultsPromise = this.fetchResultsPromise.then(this.fetchResultsIfNeeded, this.fetchResultsIfNeeded);
  return this.fetchResultsPromise;
}
```

Also note that `.always()` returns the parent promise, while `.then()` returns a new one, which is what we want since the next time we'll queue it after the latest one, not the first one. https://stackoverflow.com/a/32882576/6995854

src/mmw/js/src/modeling/views.js
arottersman: An extra word. Two for the price of one. "because it's ~are~ only downloadable from this view if"

arottersman:
The promise and typo have been fixed. This is ready for another look.

rajadain:
Taking another look now.

arottersman:
> +1 this works great! Nice work. Friendly reminder to squash changes before merging.

I shall cherish this final comment from you. It has all of our favorite catch phrases. 👍 🎉

Issue #2811, opened 2018-04-30
Subbasin Timing Out on Staging


rajadain:
Doesn't time out locally. Upon inspecting the Celery logs, I see the last entry in the log is:

```
[2018-04-30 16:45:19,173: INFO/MainProcess] Received task: apps.modeling.tasks.run_subbasin_gwlfe_chunks[9c3ab6c8-1da0-42b7-917e-f964c82cf1a7]
```

After which there is silence. Restarting Celery did not help. `gwlf-e` 0.6.3 is installed on the machine.

Find out what's wrong and make it work again.

rajadain:
Okay here’s the issue (as far as I can tell today):

1. `stream_lengths` here: https://github.com/WikiWatershed/model-my-watershed/blob/88b838f3c8ffd3a6135f8ef54811a34d8036a944/src/mmw/apps/modeling/views.py#L256 is a typed class that looks like this:
```
StreamLengthSummary(ag=21255.097623364967, urban=416954.90237663506)
```
2. When it is used here: https://github.com/WikiWatershed/model-my-watershed/blob/88b838f3c8ffd3a6135f8ef54811a34d8036a944/src/mmw/apps/modeling/calcs.py#L95, on my local, it appears as a dict:
```
{'urban': 416954.90237663506, 'ag': 21255.097623364967}
```
3. When it is used in the same spot on staging, it appears as a list:
```
[21255.097623364967, 416954.90237663506]
```
which causes the named lookup to fail.

Now I’m investigating at what point this mistranslation happens, so we can fix it. Notably, the class is composed in the `app` VM, and parsed in the `worker` VM. Celery has had trouble deserializing classes in the past, but I don’t know why we’re seeing different behaviors in the two environments.

PR #2817, opened 2018-05-06
Simplify Stream Totaling for Staging Compatibility

rajadain:
## Overview

For reasons currently not understood, the namedtuple result of `sum_subbasin_stream_lengths` is interpreted as a dict by Celery when run locally, but as a list when run on staging.

By outputting a dict always, we ensure that it is interpreted uniformly across platforms.

I've currently manually applied this patch on staging to get it to work.

Connects #2811

### Demo

Staging woking with the patch:

image

### Notes

I was unable to find out exactly why this was happening, but I did find some discussion on conflicting behavior of converting namedtuples to dicts here: https://stackoverflow.com/questions/26180528/python-named-tuple-to-dictionary. In the interest of expediency, I settled for a solution sans explanation. 🚀🐦

## Testing Instructions

* On `develop`, run subbasin HUC-10 locally and save the table data as a CSV
* Clear your local Redis cache by running: `vagrant ssh services -c 'redis-cli -n 1 --raw KEYS ":1:geop_*" | xargs redis-cli -n 1 DEL'`
* Run subbasin HUC-10 localy and export the table as a CSV again
* Compare the two CSVs and ensure they are identical


rajadain:
Updated commit removes unused import, thus pacifying flake8.

Issue #2827, opened 2018-05-14
Subbasin: Continuously Colorize Catchments


rajadain:
From the "MMW Load Rate Concentration Color Breaks" titled email thread.

> I just realized that we had previously done these as grey-scale to distinguish them from average stream concentrations. We still want to do view both at the same time, but I think we can achieve that goal with the following. First we can tint (mix in white) to the colors for the loading rates, and we can shade (mix in black) to the colors for the stream concentrations. Second, the stream concentrations rarely reflect the local loading rates for the NHD+ catchment that they are passing through. For these two reasons, I think it will work. As a third measure, I think we should always make it easy for the user to select blue lines.

## Catchments:

TN (kg/ha/y):

* ​0: dark green
* 2: light green
* 5: pale yellow
* 10: orange
* 20: red (perhaps set this to 15)
* max: black (or very dark red)

TP (kg/ha): 

* ​0.0: dark green
* 0.2: light green
* 0.5: pale yellow
* 1.0: orange
* 2.0: red  (​perhaps set this to 1.5)
* max: black (or very dark red)​

TSS (kg/ha): 0-500, 500-1000, 1000-1500, 1500-2000, >2000

* ​0: dark green
* 200: light green
* 500: pale yellow
* 1000: orange
* 2000: red (perhaps set this to 1500)
* max: black (or very dark red)​

---

## Streams:

TN (mg/l): 
* ​0: dark green​
* 1: light green (most criteria range from 0.3 to 0.9)
* 3: pale yellow (corn belt region VI is 2.18)
* 6: orange
* 12: red
* max: black (or very dark red)​

TP (mg/l): 
* ​0.00: dark green
* 0.08: light green (most criteria range from 0.010 to 0.076)
* 0.2: pale yellow (Mississippi valley region X is 0.128)
* 0.5: orange
* 1.0: red
* max: black (or very dark red)​

TSS (mg/l): 
* 0: dark green​
* 8: light green (most criteria range from 1.5 to 4.5 mg/L or 1.3 to 7.8 NTU)
* 20: pale yellow (Mississippi valley region X is 8 mg/L or 17.5 NTU)
* 50: orange
* 150: red
* max: black (or very dark red)​

Issue #2821, opened 2018-05-10
Adjust Subbasin map styling


ajrobbins:
Several styling elements on the subbasin mode map implementation were placeholder, and should be reviewed/updated by a designer:

![image 16](https://user-images.githubusercontent.com/12868342/39883911-2aa87b68-5456-11e8-9c8e-5d1b8b6ba755.png)

- yellow HUC-12 outline
- grey subbasin outlines
- black outline + extra opacity (0.4 instead of 0.3) that indicates hover

rajadain:
Also, any idea where we should put the legend?

rajadain:
For the legend, we can use this pattern:

![image](https://user-images.githubusercontent.com/1430060/40126224-d293b592-58fa-11e8-8a64-b43849459121.png)

In the area same as the slider here:

![image](https://user-images.githubusercontent.com/1430060/40126250-e0ecb2c4-58fa-11e8-8f94-77af14fe295e.png)

rajadain:
Also see if these comments can be addressed: https://github.com/WikiWatershed/model-my-watershed/pull/2829#issuecomment-389629926

PR #2829, opened 2018-05-14
Subbasin: Use Continuous Color Ramp for Streams, Catchment Load Rates

rajadain:
## Overview

Previously we used discrete color breaks for loading rates and concentration rates for catchments and their streams. Now we switch to a continuous color ramp where each value is interpolated within the gradient.

Each gradient is expected to have 5 stops. Beyond this, we add an addition sixth stop, immediately after the fifth, which goes from a darker red eventually to black.

This color ramp uses the same colors used for stream coloring in _variables.scss.

Connects #2827 

## Testing Instructions

* Go to subbasin mode and enter a HUC-12
* Turn on a related layer and observe the colors. Ensure they match with the ramp specifications in #2827.

### DIFF
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -909,7 +909,7 @@ var SubbasinCatchmentDetailModel = Backbone.Model.extend({
                 stroke: true,
                 color: '#49B8EA', // $water in _variables.scss
                 weight: 2,
-                opacity: 0.8,
+                opacity: 1,
                 fill: false
             },
             ramps = {
@@ -932,7 +932,7 @@ var SubbasinCatchmentDetailModel = Backbone.Model.extend({
 
     getStreamHighlightStyle: function() {
         return _.defaults({
-            opacity: 1
+            weight: 3,
         }, this.getStreamStyle());
     }
 });
--- a/src/mmw/js/src/modeling/models.js
+++ b/src/mmw/js/src/modeling/models.js
@@ -12,6 +12,7 @@ var $ = require('jquery'),
     turfErase = require('turf-erase'),
     turfIntersect = require('turf-intersect'),
     AoiVolumeModel = require('./tr55/models').AoiVolumeModel,
+    makeColorRamp = require('./gwlfe/subbasin/models').makeColorRamp,
     round = utils.toRoundedLocaleString;
 
 var ModelPackageControlModel = Backbone.Model.extend({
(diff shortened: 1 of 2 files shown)


src/mmw/js/src/modeling/models.js
```diff
@@ -877,42 +878,18 @@ var SubbasinCatchmentDetailModel = Backbone.Model.extend({
...
-                    [ 250, '#A0A0A0'],
-                    [ 500, '#888888'],
-                    [ 750, '#707070'],
-                    [1000, '#484848'],
-                    [Number.MAX_VALUE, '#202020']
-                ]
+            ramps = {
+                TotalN: makeColorRamp([0, 2, 5, 10, 20]),
+                TotalP: makeColorRamp([0, 0.2, 0.5, 1.0, 2.0]),
+                Sediment: makeColorRamp([0, 200, 500, 1000, 2000]),
```
caseycesari: Double-checking that we don't want to use `Number.MAX_VALUE` anymore.

src/mmw/js/src/modeling/models.js
```diff
@@ -935,29 +912,10 @@ var SubbasinCatchmentDetailModel = Backbone.Model.extend({
...
-                    [ 50, '#1A9641'],
-                    [100, '#A6D96A'],
-                    [150, '#FFFFBF'],
-                    [200, '#FDAE61'],
-                    [Number.MAX_VALUE, '#D7191C']
-                ]
+            ramps = {
+                TotalN: makeColorRamp([0, 1, 3, 6, 12]),
+                TotalP: makeColorRamp([0, 0.08, 0.2, 0.5, 1.0]),
+                Sediment: makeColorRamp([0, 8, 20, 50, 150]),
```
caseycesari: These numbers are different than the values they are replacing.

rajadain:
Re: Legend, I've made a note of it in #2821. We'll need input from Design for where to put it.

rajadain:
@aufdenkampe this work is now available on staging: https://staging.app.wikiwatershed.org/

aufdenkampe:
@rajadain, thanks for letting me know. It's fun to play with these!
Here are a few suggestions:
* The border lines for the NHDplus catchements are too thick, interfering with the visualization of the stream segments. Please make the border lines to always be less thick than the streams (or at least no thicker than thinest stream). If there is a concern that these lines might barely be visible, that is OK with me, as the interactivity with the map should be enough to distinguish catchments.
* The darkest grey color isn't really working for me. It should be clearly Red in hue, in a ramp that shades from light red to dark red as loading rates increase. Also, there appear to be two different blocks of grey in the color ramp, one that is at the highest levels and another at moderate levels below orange.
* The implemented color ramp doesn't really make sense visually, and it doesn't seem like what I had suggested. Is there a bug? When I sort the table by loading rates, what I can discern is this ramp from low to high:
  * light green
  * salmon pink
  * green yellow? (hard to distinguish from light green)
  * beige
  * dark grey
  * grey orange with a hint of violet
  * dark grey (I can't distinguish from above)
* It will be good to have the transparency slider wired up, so that it is easier to see the colors over the base map, if desired.

Issue #2715, opened 2018-03-08
Check entire project for Safari target="_blank"


ajrobbins:
New Safari updates mean that links are downloading as html instead of opening in a new tab if they have `target="_blank`" but no `rel="noreferrer noopener"`. Check the entire project for these cases to fix them for Safari, as was done in rel="noreferrer noopener"

ajrobbins:
An example where this is happening: clicking on "source data" or "web services" buttons in Monitor results downloads the page as html

PR #2838, opened 2018-05-21
Fix external links for Safari

rajadain:
## Overview

Recent versions of Safari download links as files that have target _blank set, unless also accompanied by the noreferrer noopener rel attributes. By adding them we ensure that all links work correctly in Safari as well as other browsers.

Also fix a lingering reference to the Micro App which is now called Runoff Simulation.

Connects #2715

## Testing Instructions

* Ensure there are no `a` tags with `target="_blank"` that _don't_ also have `rel="noreferrer noopener"` set
* Try clicking on new links in Safari. Ensure they open instead of downloading.


rajadain:
Thanks for taking a look!

Issue #2427, opened 2017-10-24
Add Lat-Lon input to "Jump to location or HUC" search


aufdenkampe:
According to Tara Muenz, Assistant Director of Education at the Stroud Water Research Center, many users know the latitude and longitude of their field site much better than a street address (i.e. a stream location or potential BMP site in a field on someone's farm). The users would like to be able to type in a Lat & Lon into the Model My Watershed search field to zoom to that location, in order to do then follow their workflows.  Presently they need to first use Google Maps, then try to match the location on Model My Watershed.

PR #2845, opened 2018-05-25
Jump to Coordinates

rajadain:
## Overview

Adds support for users to paste in a lng,lat or lat,lng coordinate in the search bar. If it is within the continental United States, we show a "Go to {{ coordinates }}" button which they can click to jump the map there. The zoom is hard set to 14, using the same defaults used for other kinds of searches.

Connects #2427 

### Demo

Tagging @ajrobbins and @jfrankl for brief visual review.

## Testing Instructions

* Try searching by coordinates. Ensure Lat, Lon and Lon, Lat both work.
* Ensure you cannot jump to invalid coordinates, or those outside the US.
* Ensure geocoding and HUC searches still work as expected.

### DIFF
--- a/src/mmw/js/src/core/tests.js
+++ b/src/mmw/js/src/core/tests.js
@@ -779,6 +779,31 @@ describe('Core', function() {
             assert.equal(output.version, 'Local');
             assert.equal(output.releaseNotesUrl, latestUrl);
         });
+
+        it('parses coordinates correctly', function() {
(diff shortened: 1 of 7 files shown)


ajrobbins:
Looking good! Can you change the default placeholder text in that element to read "Jump to location, HUC, or coordinates"? This will help hint about the feature availability to users.

rajadain:
Thanks for taking a look! Added a commit to update the placeholder text:

image

Merging now.

Issue #2851, opened 2018-05-29
Fix links in Contribute Data modal


ajrobbins:
A couple of the links in the Contribute Data modal have https when the sites don't support it. Revise to use http and ensure that all links in the modal open pages properly.

rajadain:
Shares points with #2815

rajadain:
Some more background about the upcoming changes in Google Chrome that will mark HTTP pages that have form fields as "Not Secure" https://security.googleblog.com/2018/02/a-secure-web-is-here-to-stay.html /cc @aufdenkampe

PR #2853, opened 2018-05-29
Update Contribute modal links

rajadain:
## Overview

Some of these sites do not have HTTPS support, so we point to their HTTP versions instead.

Connects #2851 

## Testing Instructions

* Checkout this branch and `bundle --debug`
* Go to [:8000/](http://localhost:8000/), pick a shape and go to the Monitor tab
* Open the Contribute Data modal and ensure all the links correctly work


rajadain:
Build failure is because of Firefox version upgrade addressed in #2852.

Issue #2859, opened 2018-05-30
Monitor: Limit /values/ to 5s


rajadain:
When searching for some EnviroDIY values (such as _SHPK3S Pickering Creek Eastern Trib at Bryn Coed Lane_), the EnviroDIY server responds with 504s:

![image](https://user-images.githubusercontent.com/1430060/40746842-7fb9e032-6429-11e8-99ee-349928ded5aa.png)

This is tolerated and managed locally. However, on staging, it causes all the `gunicorn` processes to restart. Additionally, we start getting 503s from the server, and it is eventually cycled out.

Find out why this is happening and make it so that handling this is just as graceful on staging as it is on local.

rajadain:
Paired on this with @hectcastro and @tnation14. This is what we observed:

* Staging has 3 worker processes, evaluated based on how many CPUs are on a platform (see #1724). Since Staging is a `t2.small` which has 1 CPU, the formula `2n + 1` yields 3, where `n` is the number of CPUs.
* When the client fires off 4 EnviroDIY requests, 3 of those are given to the 3 worker processes. The 4th is enqueued.
* These requests take ~ 60s, at which point the EnviroDIY server returns a 504 timeout.
* Because these requests each take 60s, and all workers are busy, other requests are enqueued and not responded to.
* These other requests include health check requests by the ELB, which are fired every 30s.
* If the health check fails twice, the app VM is recycled.
* Since the workers are busy for >60s, the health check fails twice, and the app VM is recycled.

Proposed ideas for solving this:

* Add a timeout to the `/values/` Ulmo calls of 3-10s (the lower the better. This seems to be enough for the other data providers). If Ulmo does not support timeouts natively, wrap it in something that will kill it after the timeout.
* Increase workers on Staging. (Dev has 2 CPUs so 5 workers. Dev also doesn't have the ELB kill switch for 60s of inactivity.) Since each worker process has the full app in memory, we may run in to RAM issues.
* Switch to `gevent` and aynschronous workers. This allows us to have far more worker threads, instead of the processes that we currently have. This _should not_ have ill consequences, since most of our requests are IO bound rather than CPU, with the exception of the `/details/` for CUAHSI which require a lot of CPU for parsing the data.

Other notes:

* Update local dev timeout to be the same as staging here:

[...]

PR #2866, opened 2018-06-04
Monitor: Limit CUAHSI /values/ to 5s

rajadain:
## Overview

Adds a 5s timeout to CUAHSI /values/ endpoint, which was tying up workers and causing app VM cycles on staging.

Note that if the front-end times out, the user can always go back to the result list and click the same result again, and we'll attempt to re-fetch it. This is useful for cases like NWISUV, which can sometimes have occasional failures, while still protecting from EnviroDIY, which fails consistently.

Also reduce gunicorn timeout on development builds to match production.

Connects #2859 

### Demo

Rare partial failure with NWISUV:

Common total failure with EnviroDIY:

### Notes

I initially found [this Stack Overflow post](https://stackoverflow.com/a/2282656/6995854) for how to implement it, but ultimately went with [a library](https://github.com/pnpnpn/timeout-decorator) that does the same thing with a clean API.

## Testing Instructions

* Check out this PR and `bundle --debug`
* Go to [:8000/](http://localhost:8000/) and select the Lower Schuylkill HUC-10
* Go to Monitor tab, search for "water" and select the CUAHSI WDC tab
* Click through to any result. Ensure none of the values have requests active for more than 5 seconds.
  - For EnviroDIY, this should be all values
  - For every other data source, the data should come through

### DIFF
--- a/src/mmw/apps/bigcz/clients/cuahsi/details.py
+++ b/src/mmw/apps/bigcz/clients/cuahsi/details.py
@@ -10,6 +10,8 @@ from rest_framework.exceptions import ValidationError
 
 from django.conf import settings
 
+from apps.bigcz.utils import ValuesTimedOutError
+
 DATE_FORMAT = '%m/%d/%Y'
 
 
@@ -29,7 +31,7 @@ def details(wsdl, site):
     return wof.get_site_info(wsdl, site, None)
 
 
-@timeout(settings.BIGCZ_CLIENT_TIMEOUT)
+@timeout(settings.BIGCZ_CLIENT_TIMEOUT, timeout_exception=ValuesTimedOutError)
 # NOTE: This @timeout decorator will have to be modified for a multi-threaded
 # environment, with the use_signals=false attribute. Unfortunately, that does
 # not support functions that return values that cannot be pickled, such as this
--- a/src/mmw/apps/bigcz/clients/cuahsi/details.py
+++ b/src/mmw/apps/bigcz/clients/cuahsi/details.py
@@ -4,8 +4,11 @@ from __future__ import unicode_literals
 from __future__ import division
 
 from datetime import date, timedelta
+from timeout_decorator import timeout
 
 from rest_framework.exceptions import ValidationError
+
+from django.conf import settings
 
 DATE_FORMAT = '%m/%d/%Y'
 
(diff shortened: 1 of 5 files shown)


src/mmw/requirements/base.txt
```diff
@@ -28,3 +28,4 @@ numpy==1.13.0
 hs_restclient==1.2.10
 six==1.11.0
 fiona==1.7.11
+timeout-decorator==0.4.0
```
hectcastro: We talked about the risks of switching from Gunicorn sync workers vs. async workers. It looks like this library is an example of one that would be difficult to use with async workers due to its use of signals.

See: https://github.com/pnpnpn/timeout-decorator#multithreading

It looks like we're in the clear, but definitely worth exercising the timeout process multiple times in one session, with the service launched via Upstart vs. the `debugserver` script.

hectcastro:
For reference, I took a peek at `ulmo` and it looks like it creates the `suds` client here:

https://github.com/ulmo-dev/ulmo/blob/master/ulmo/cuahsi/wof/core.py#L283

The HTTP and HTTPS transport classes for `suds` have `timeout` options, but it doesn't look like the `ulmo` API surfaces a way to set it (it does have a way to set a [cache](https://github.com/ulmo-dev/ulmo/blob/master/ulmo/cuahsi/wof/core.py#L284-L292) though). If there were a way, that approach would likely be better than this timeout wrapper (I don't think there is, so not pushing for a change in direction with this comment).

- https://bitbucket.org/jurko/suds/src/94664ddd46a61d06862fa8fb6ba7b9e054214f57/suds/transport/http.py?at=default&fileviewer=file-view-default#http.py-49
- https://bitbucket.org/jurko/suds/src/94664ddd46a61d06862fa8fb6ba7b9e054214f57/suds/transport/https.py?at=default&fileviewer=file-view-default#https.py-46

src/mmw/apps/bigcz/clients/cuahsi/details.py
```diff
@@ -4,10 +4,19 @@
...
+from rest_framework.exceptions import APIException, ValidationError
 
 DATE_FORMAT = '%m/%d/%Y'
+WOFPY_TIMEOUT = 5
+
+
+class WOFpyTimeout(APIException):
+    status_code = status.HTTP_504_GATEWAY_TIMEOUT
+    default_detail = \
+        'WOFpy request did not finish in {} seconds'.format(WOFPY_TIMEOUT)
```
caseycesari: Is it useful to say `WOFpy` in the error message for the typical user? It could potentially stay as is without `WOFpy`.

rajadain: Admittedly this was a little defensive, in the sense we want to differentiate between an MMW timeout (our fault) and a WOFpy timeout (not our fault). But if it feels too unnecessarily detailed we can simplify it.

rajadain:
Good call on surfacing the suds timeout option in Ulmo. I'll ask the devs there to consider it for the next version.

src/mmw/apps/bigcz/clients/cuahsi/details.py
caseycesari: I'm just not sure that anyone will know what it is. Maybe something like `Details request did not finish in {} seconds`? Not a dealbreaker though.

rajadain:
Created https://github.com/ulmo-dev/ulmo/issues/155 for adding timeout support to Ulmo.

rajadain:
Added fixups to reuse the BIGCZ_CLIENT_TIMEOUT variable which is already set to 5s for uniformity.

rajadain:
Demo of the invalid date catch:



This could use another look before merging.

rajadain:
I'm merging this today so we can get this on staging.

PR #2868, opened 2018-06-04
Subbasin: Layer Legend and Opacity

rajadain:
## Overview

Adds a opacity slider and legend box next to the layer picker for a related subbasin layer when selected.

Connects #2821 

### Demo

There is some lag above because of GIF recording using up the CPU, that is not visible otherwise. Note that the opacity slider adjusts opacity of the catchment fill, not the stream stroke. None of our stroke layers support opacity, so I didn't add them here either. It would be easy to add if we want it though.

Tagging @ajrobbins, @jfrankl for visual review.

## Testing Instructions

* Go to [:8000/](http://localhost:8000/) and select a HUC-10 and go into subbasin mode
* Ensure the HUC-12 boundaries now use Red instead of the previous Blue for their outline
* Select a HUC-12. Ensure it uses Red instead of the previous Yellow for its outline
* Select a related layer. Ensure its legend and opacity control are shown next to the layer picker.
* Unselect a related layer. Ensure the legend / opacity control go away.
* Select a different related layer. Ensure the legend is properly updated, and that the opacity value persists.
* Turn on a Climate layer, such as Mean Precipitation or Temperature. Ensure the time slider (with the months) is not visible (or rather, hidden behind) the legend / opacity control of a subbasin layer.

### DIFF
--- a/src/mmw/apps/home/templates/home/home.html
+++ b/src/mmw/apps/home/templates/home/home.html
@@ -12,6 +12,7 @@
 <div id="map"></div>
 <div class="layerpicker" id="layer-picker-region"></div>
 <div id="layer-picker-slider-region"></div>
+<div id="subbasin-slider-region"></div>
 <div id="boundary-label"></div>
 {% endblock map %}
 
--- a/src/mmw/js/src/core/models.js
+++ b/src/mmw/js/src/core/models.js
@@ -26,6 +26,7 @@ var MapModel = Backbone.Model.extend({
         dataCatalogActiveResult: null,  // Model
         dataCatalogDetailResult: null,  // Model
         selectedGeocoderArea: null,     // GeoJSON
+        subbasinOpacity: 0.85,
     },
 
     revertMaskLayer: function() {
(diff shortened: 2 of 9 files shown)


ajrobbins:
LGTM from gif - did you get the chance to test in the other browsers?

jfrankl:
Looks good overall. I'd be interested in what it would look like to get the labels and units of measurement on the same line instead of separate lines. Seems like that might reduce the amount of space used by that component.

#### Before:
Mean Annual Concentration 
(mg/L)

#### After:
Mean Annual Concentration (mg/L)

rajadain:
Making the box wider to fit things on one line:

image

If we go with this, I'll further adjust it a couple pixels to match the height of the layer picker sans titlebar.

rajadain:
Thanks for prompting that @ajrobbins.

Chrome:

image

Safari:

image

IE11 is having an issue where adjusting the opacity slider doesn't update the catchments until you hover on them:

image

Looking into that now.

rajadain:
Okay, the IE11 problem is that sometimes the mouse can move too fast and the `mouseout` event doesn't fire, leaving multiple catchments highlighted. When a highlighted catchments tries to bring its layer to the front, the given layer does not have a `._map` reference, as opposed to when hovering on the map when it does. Looking further into it.

rajadain:
IE11 is fixed:

rajadain:
I will merge this in an hour or so if no objections are raised.

rajadain:
Thanks for the feedback, everyone!

Issue #2867, opened 2018-06-04
CINERGI searches failing


ajrobbins:
I'm observing 400 errors when attempting CINERGI searches on AoIs of various sizes:

![screen shot 2018-06-04 at 2 58 33 pm](https://user-images.githubusercontent.com/12868342/40936010-de708946-6807-11e8-84e3-ebd561a9e7d6.png)

rajadain:
Looks like CINERGI changed their API and didn't tell us 😞 

<details>
<summary>
Old Response
</summary>

```json
{
  "took": 141,
  "timed_out": false,
  "_shards": {
    "total": 5,
    "successful": 5,
    "failed": 0
  },
  "hits": {
    "total": 5779,
    "max_score": 4.464109,
    "hits": [
      {
        "_index": "metadata_v5",
        "_type": "item",
        "_id": "4146e0ebc9a441988246314918a0869e",
        "_score": 4.464109,
        "_source": {
          "src_source_type_s": "WAF",
          "src_source_uri_s": "WAF:http://132.249.238.151:8080/cinergi/waf/U.S._Geoscience_Information_Network/",
          "src_source_name_s": "US GIN",
          "src_uri_s": "http://132.249.238.151:8080/cinergi/waf/U.S._Geoscience_Information_Network/850932de-b2b0-40a2-aa87-7645b23b5743.xml",
          "src_lastupdate_dt": "2017-01-14T15:23:24Z",
          "sys_created_dt": "2017-03-14T16:57:59.957Z",
          "sys_modified_dt": "2017-03-15T21:12:23.776Z",
          "sys_xmlmodified_dt": "2017-03-14T16:57:59.957Z",
          "sys_owner_s": "gptadmin",
          "sys_metadatatype_s": "iso19115-2",
          "fileid": "850932de-b2b0-40a2-aa87-7645b23b5743",
          "title": "US Water Bodies",
          "description": "U.S. Map Data Water Boundaries represents water feature areas within United States. Water boundaries include the following: basic hydrography, naturally flowing water features, man-made channels to transport water, inland bodies of water, man-made bodies of water, seaward bodies of water, bodies of water in a man-made excavation, and special water features.",
          "keywords_s": [
            "Hydrography",
            "polygon",
            "water",
            "hydrography",
            "naturally flowing",
            "channel",
            "inland body",
            "seaward body",
            "man-made",
            "inlandWaters",
            "oceans",
            "transportation",
            "United States",
            "Channel",
[...]

PR #2870, opened 2018-06-06
Monitor: Fix CINERGI

rajadain:
## Overview

The CINERGI API response was changed recently, although it is unclear when this happened. Thankfully, most of the schema we expect is still preserved in the `_source` key of the response, so we can keep parsing it with minimal modifications.

Connects #2867 

### Demo

image

## Testing Instructions

* Go to [:8000/](http://localhost:8000/), select a shape and go to the Monitor tab
* Search for "water". Switch to the CINERGI tab. Ensure you see results.
  - A lot of the results will have "Unknown Time Period" or no contact people, or people listed in the contact organizations field. This is just the data we're given, and same as before.


rajadain:
Thanks for taking a look! Just refired the CI, will merge when green.

Issue #2849, opened 2018-05-29
Conservation practice popup offscreen in Safari


ajrobbins:
Observed in Safari 11.1 (and by Anthony) on Macbook Air

To replicate:
- Run the multiyear model on an AoI
- Add a conservation practice
- CLick on the applied conservation practice in the sidebar

![screen shot 2018-05-29 at 2 02 51 pm](https://user-images.githubusercontent.com/12868342/40676716-93955cca-6349-11e8-9c1e-51afd45662c8.png)

PR #2871, opened 2018-06-06
Fix MapShed modification tooltip in Safari

rajadain:
## Overview

`$().offset()` doesn't work in Safari, returning `{top: 0, left: 0}` in all cases. We switch to using the Event Target `.getBoundingClientRect()` which works for all browsers. There are no other instances of `$().offset()` in the code.

Connects #2849 

### Demo

Safari:

image

Firefox:

image

Chrome:

image

IE11:

image

## Testing Instructions

* Go to [:8000/](http://localhost:8000/) in Safari and log in. Either create a new MapShed project or open an existing one
* Add a new scenario and a couple of modifications
* Click a modification. Ensure its tooltip is shown next to it.
* Open various browsers and go to the same scenario and check the tooltips therein.

### DIFF
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -795,7 +795,7 @@ var GwlfeToolbarView = ScenarioModelToolbarView.extend({
 
     onThumbClick: function(e) {
         var modKey = $(e.currentTarget).data('value'),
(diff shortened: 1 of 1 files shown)


rajadain:
Thanks for taking a look!

Issue #2824, opened 2018-05-12
Update SedNitr and SedAAdjust multipliers in Mapshed


ajrobbins:
**SedNitr change**
From BME:

>Back in March of 2017, I sent a document to you and Matt describing how some new GIS layers should be used to calculate various model parameters (see attached). I was wondering - would it be possible to make a slight change in the value used to multiply the derived "SedNitr" value by (i.e., 6.0 instead of 4.0)? Based on my recent calibration work, I'm finding that the nitrogen loads are being calculated slightly lower than they should be, and this slight change would help correct that problem. Thanks.

See routines descriptions in https://drive.google.com/open?id=0B0hs2mX4ApaLYnNkdThNZXppWUN2dE1USUp2X1hqZG9sekM0

**SedAAdjust Change**
From BME:
>Last summer, we made a change in a default value that was being written to the gms file. Specifically, the default value for the Sediment adjustment factor (SedAAdjust, the 7th value on line 36 of the gms file) was changed from "1" to "1.4". Based on my calibration work, I think this value should now be changed to ~"2.1"~ 1.5. [edited 6.20]

rajadain:
Since `SedNitr` had already been set to `7.0` in #2563, I asked BME what we should and this is what he said:

> I guess I forgot that we had changed it to 7.0. So I guess the easiest way to answer this is to say, if it's currently 7.0, change it to 8.0. However, if it is still 4.0 (which I know would be easy enough to check), change it to 6.0. Thanks for your help.

So we should change it to `8.0`.

rajadain:
Update from BME:

> Sorry, I changed my mind. If the multiplication factor for SedNitr is already at 7.0, please change it to 9.0 instead to make the resultant change in load significant enough. I promise this will be my last request for this factor!

We need to set it to `9.0`.

PR #2876, opened 2018-06-22
Adjust SedAAdjust to 1.5

kellyi:
## Overview

Per the request in #2824, this PR adjusts the SedAAdjust constant for GWLFE to be 1.5 instead of 1.4.

Connects #2824

### Notes

~We may have already changed the SedNitr value in this PR -- https://github.com/WikiWatershed/model-my-watershed/pull/2563 -- which changed it from 4.0 to 7.0.~

~The document linked from #2824 requests changing it from 4.0 -> 7.0; however the quote in the issue indicated 6.0. Not certain which value to keep.~

@mmcfarland confirmed that the new value should be 6.0, so that change was made in 8f8e847

## Testing Instructions

- get, build, and serve this branch, then run a mapshed job and verify that it still works

### DIFF
--- a/src/mmw/apps/modeling/mapshed/tasks.py
+++ b/src/mmw/apps/modeling/mapshed/tasks.py
@@ -404,7 +404,7 @@ def soiln(result):
 
     result = parse(result)
 
-    soiln = result.values()[0] * 7.0
+    soiln = result.values()[0] * 8.0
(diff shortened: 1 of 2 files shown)


kellyi:
Thanks! Will merge once the tests pass.

PR #2878, opened 2018-06-25
Change soiln adjustment factor to 9.0

kellyi:
## Overview

This PR changes the adjustment factor for soiln from 8.0 to 9.0

Connects #2824 

## Testing

- get build and serve this branch, then run a mapshed job and confirm that it works


kellyi:
Thanks!

Issue #2769, opened 2018-04-09
Incorporate PA municipalities boundary with PA urbanized areas overlay


ajrobbins:
PA DEP has a need to view the PA urbanized areas overlay and the PA municipalities boundary overlay at the same time in the same "coverage grid" layer, so they can turn on the HUC 12 overlay as well for a digitization workflow. We can move the muncipalities overlay to the urban areas layer, and change the red color to something that will contrast with the red HUC12 lines when they are viewed together.

ajrobbins:
To be clear, the desired new behavior is that when the user turns on the PA urbanized areas layer in the "coverage grid" section of the layer selector, the PA municipalities overlay also appears (not activated in boundary menu, but is actually merged into this one layer). The municipality lines should be a different color than red. The urbanized areas layer would then be renamed "PA Municipalities and Urbanized Areas", and I believe that the existing Municipalities layer can be removed from the boundaries section of the layer selector.

rajadain:
I have the paired layers working now:

![2018-06-26 13 28 02](https://user-images.githubusercontent.com/1430060/41929005-f689a10e-7944-11e8-9dd0-5a7d622a7d24.gif)

Any recommendations on what color these boundary lines should be so they don't conflict with actual boundary layers (such as HUC-12) when they are turned on too:

![image](https://user-images.githubusercontent.com/1430060/41929056-1518d54a-7945-11e8-9f4f-0bc8f4829267.png)

/cc @ajrobbins @jfrankl

rajadain:
Current color scheme is defined here:

https://github.com/WikiWatershed/model-my-watershed/blob/7e7fd74a0b24df8589f00cbc721666c388e32244/src/tiler/styles.mss#L5-L26

ajrobbins:
Chatted with @jfrankl about this - he recommends using the black we use in subbasin hover mode instead of the red for the municipalities line color.

![screen shot 2018-06-26 at 1 58 03 pm](https://user-images.githubusercontent.com/12868342/41930460-1e2a3800-7949-11e8-9c28-341a145655af.png)

It looks like there's a thin white stroke around the red now, which we should keep with the black.

rajadain:
The color used there is a very dark green #1d3331 ![1d3331](https://via.placeholder.com/16/1d3331?text=%20). Here's what that looks like:

![image](https://user-images.githubusercontent.com/1430060/41931171-437420ec-794b-11e8-9f96-105a0742ee7c.png)

[...]

PR #2879, opened 2018-06-26
Combine PA Municipalities with Urbanization Layer

rajadain:
## Overview

Users would like to be able to view the PA Urbanization Layer and Municipalities in combination with hydrological boundaries such as HUC-12s. This is currently not possible since only one boundary may be active at a time.

To allow this, we combine the PA Municipalities layer with the Urbanization Layer into a new PA Urbanization with Municipalities coverage layer, which can be combined with any boundary layer.

We also develop a framework where any number of layers can be combined for any coverage layer going forward.

Connects #2769 

### Notes

I tried adding an endpoint in Windshaft that did this, but could not figure out the SQL syntax for it. Also considered making a database view that had combined results, but thought that could be very expensive in terms of processing. Ended up going with a Leaflet based solution.

Upon deployment, we will need to clear the tile cache for this layer.

## Testing Instructions

* Restart the tiler:

    ```bash
    $ vagrant ssh tiler -c 'sudo service mmw-tiler restart'
    ```
* Go to [:8000/](http://localhost:8000/) and switch to the "Coverage" tab in the layer selector
* Select the new "PA Urbanization with Municipalities" layer. Ensure you see a gray overlay in urban areas and a dark green outline demarcating municipal boundaries.
  - You may have to open the network tab in dev tools and select "Disable Cache" to get the new tiles in case the old ones are cached in the browser
* Ensure changing the opacity works for both the grey and the dark green.
* Ensure zooming and panning loads tiles for both layers (albeit with a little delay, but always in the correct z-order).
* Enable a boundary layer, such as HUC-12. Ensure it renders atop the coverage layer and its styling is distinct from the municipal boundary.
* Enable a stream layer. Ensure it renders atop the coverage layer.


kellyi:
Got this up and running:

screen shot 2018-06-26 at 3 20 10 pm

rajadain:
Added some legend entries:

image

Thanks for taking a look! Will merge when green.

Issue #2884, opened 2018-07-10
Move WQP to Monitor


rajadain:
Currently WQP is only available on BiG-CZ. Make it available on Monitor as well.

Issue #2875, opened 2018-06-19
Add IP anonymization flag to Google Analytics


ajrobbins:
Request from Stroud, spurred by GDPR:

>I'd like our Google Analytics tracking codes to be consistent within a property, so since I've updated the copy in wikiwatershed.org to anonymize the last four digits of site visitors' IP addresses, could you see that it's updated where it's embedded in the app? 

>The last line before the closing tag will need to be updated to this:

> gtag('config', 'UA-47047573-7', { 'anonymize_ip': true });

rajadain:
Shares a point with #2884

PR #2480, opened 2017-11-05
Add Google Analytic Events

mmcfarland:
## Overview

Adds Event tracking for high profile activity within the app, for the purposes of describing app user behavior to WPF.

* Aoi Creation: Captures the type of aoi creation selected: boundary type, freedraw, squarekm, drb-rwd, nhd-drb, shapefile or geojson
* Analyze: An analyze request for an AoI, one for each type: animals, land, soil, etc.
* Modeling: Captures creating a project, the model type, adding a scenario and adding a modification.  Scenario and Modification events capture the model package as well 

Connects #2450 

### Demo
Events showing up:
image

Event requests being sent (Network -> JS)
image

### Notes

There are a large number of potential events to capture, however I chose the top actions that I anticipated WPF may wish to know.  This could be expanded to determine which features are commonly used, but since we don't have explicit funding for that type of analysis, I'm keeping it as simple as possible for now.

## Testing Instructions

 * Application behavior should not have changed at all
 * GA Events take up to 24 hours to register in the dashboard, but Real-Time events can be seen almost immediately.  I can help set up access for you to test.

PR #2887, opened 2018-07-13
Monitor: Add WQP, Anonymize IP Addresses

rajadain:
## Overview

Adds WQP to Monitor since it has not exhibited any issues in its BiG-CZ deployment (restricted to BiG-CZ in #2856). It is now available as a fourth tab in Monitor.

Also anonymizes IP addresses of users in Google Analytics. The issue #2875 gives instructions for `gtag.js`, but we currently use `analytics.js`, so I found and used instructions for that. There is a [migration guide](https://developers.google.com/analytics/devguides/collection/gtagjs/migration#track_events) for upgrading to `gtag.js`, but we [most likely don't need to](https://reflectivedata.com/gtag-js-migrate/):

> ##### Using `analytics.js`
> This scenario is probably applicable to most websites on the web. As this solution is working fine and will continue to be most popular snippet for years (probably), there is no rush for migrating. You have the widest range of tutorials and third-party support.

Furthermore, changing all the tracking code added in #2480 would have been outside the scope of this card. We can revisit this in the future if we need to.

Connects #2884 
Connects #2875 

### Demo

Monitor:

image

BiG-CZ:

image

## Testing Instructions

* Go to [:8000/](http://localhost:8000/) and select a shape. Go to Monitor and search for Water. Click the final tab of WQP and ensure the search works.
* Go to [:8000/?bigcz](http://localhost:8000/?bigcz) and do the same. Ensure the search works.

### DIFF
--- a/src/mmw/apps/core/templates/head.html
+++ b/src/mmw/apps/core/templates/head.html
@@ -20,6 +20,7 @@
         })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
 
         ga('create', '{{ GOOGLE_ANALYTICS_ACCOUNT }}', 'auto');
+        ga('set', 'anonymizeIp', true);
         ga('send', 'pageview');
     </script>
 </head>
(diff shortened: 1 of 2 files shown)


rajadain:
Thanks for taking a look!

Issue #2891, opened 2018-07-24
Drop a pin as a result of address or lat/lon search


ajrobbins:
Request from Stroud:

>Add pin or other icon for specific location when searching by GPS location? This location search by GPS is being used regularly now by him and DRWI watershed associations for locating in MMW their sensor stations and then modeling areas around them. 

They are searching for specific locations, and then want to delineate AoIs based on them, so they want to see the result on the map for either address or lat/lon.

The pin should be visible in the Draw stage, and when you proceed to Analyze it should be removed from the map.

PR #2896, opened 2018-07-30
Add Marker for Coordinate, Address Search

rajadain:
## Overview

Adds a maker to the map when search for coordinates or an address. Does not do it when selecting a HUC. Removes it when moving to Analyze.

🚀 🐦 

Connects #2891 

### Notes

I tried overloading `selectedGeocoderArea` for this, since theoretically it is only used for highlighting the HUC in the exactly exclusive case (of HUC search, _not_ coordinate or address), but the resulting logic was too complicated and hard to follow, so I ended up adding `searchResult` as a field on the `map` object.

## Testing Instructions

* Check out this branch, `bundle`, go to [:8000/](http://localhost:8000/)
* Search for coordinates. Ensure you see a marker.
* Search for an address. Ensure you see a marker.
* Ensure there is at most one marker on the map at any time.
* Search for a HUC. Ensure you don't see a marker.
* Search for something so that you see a marker. Then move to Analyze. Ensure you don't see the marker anymore.

### DIFF
--- a/src/mmw/js/src/core/models.js
+++ b/src/mmw/js/src/core/models.js
@@ -27,6 +27,7 @@ var MapModel = Backbone.Model.extend({
         dataCatalogDetailResult: null,  // Model
         selectedGeocoderArea: null,     // GeoJSON
         subbasinOpacity: 0.85,
+        searchResult: null,             // [lat, lng]
     },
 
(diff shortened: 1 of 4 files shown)


rajadain:
Thanks for reviewing!

Issue #2885, opened 2018-07-10
Resolve HydroShare Export timeout


aufdenkampe:
@darscott and I have both experienced issues exporting MMW projects to HydroShare.

For example, when I export this project, https://app.wikiwatershed.org/project/24728/, I get this response after waiting quite a while for the ball to spin:

![screen shot 2018-07-10 at 3 49 19 pm](https://user-images.githubusercontent.com/5166036/42536858-d9eb841c-8458-11e8-949e-e3024ef47287.png)

I followed the suggestion of @rajadain to try multiple times, but I just got multiple new resources on HydroShare:

![screen shot 2018-07-10 at 3 50 29 pm](https://user-images.githubusercontent.com/5166036/42536942-1af8264a-8459-11e8-99a8-eda639c9f4ec.png)

Each resource has .csv files for Analyze and Model outputs, plus the AoI geoJSON. However, I also don't see a GMS file.

@dtarb, @rayi113 and @horsburgh, Could there have been a change on HydroShare?

sblack-usu:
Assuming the first project uploaded to hydroshare is missing the GSM file and timed out like the others, I don't believe we can blame this on the release downtime (The first one happened on July 6th, the others on July 10th).

If someone can give me the call to hydroshare that is made by model my watershed, then I can help troubleshoot.  It'd also be helpful to get a copy of the file set to be uploaded to a resource.

aufdenkampe:
Is this on staging yet?
@darscott and I were presenting MMW today for the DRWI and this issue cropped up again when we tested it on staging. It seems that Beta.HydroShare.org is up, so that's not the issue.

rajadain:
Hi @aufdenkampe,

After coming back, beta.hydroshare.org may have returned as a new clone of production. In which case, any existing OAuth permissions saved with MMW have expired, leading to the issues above. They can be solved by going to your account, unlinking and relinking to HydroShare, and then exporting the projects. This will not happen on production, because the production HydroShare database isn't refreshed like the beta is.

Thanks.

aufdenkampe:
Update: it just worked for me on staging / beta. I linked and unlinked the accounts, but it still took 2 tries to complete.

aufdenkampe:
@rajadain, thanks for explaining that a bit more clearly. It is now definitely working better now.

[...]

PR #2897, opened 2018-08-01
Celerify HydroShare Export

rajadain:
## Overview

Since exporting multiple files to HydroShare can take a long time, and projects with multiple scenarios can balloon the number of files really quickly, the exports would time out in some cases.

To mitigate that, we wrap the creation and updating methods for HydroShare resources in Celery Tasks, and return a Job ID instead. This frees the export from the constrains of a request-response cycle, while also not tying up a `gunicorn` worker process.

Celery Tasks have their own timeouts however, and these are set to timeout after 5 minutes. The front-end which polls for these results has the same timeout.

Since these exports take so long, as users add changes in the front-end that can keep being queued for export. But if each export takes ~4 minutes, and each single change triggers an export, we're looking at potentially hours of exports queued up. If the browser is closed, or the user navigates away during this time, the queued exports would be dropped silently. Furthermore, there is no visibility of this queue, or any sort of progress of the exports. Thus, the autosync behavior is too obfuscated with potential data loss for users, which is why the final commit disables it. The autosync options are removed from the UI, all new projects default to it disabled, and a migration turns it off for all currently-syncing HydroShare resources.

~To assist or prompt users to export to HydroShare, we can think about adding a "dirty" or "stale" flag to a HydroShare resource, that can be set when any change to a project is made, and unset when a HydroShare export finishes. This is left to a future task.~

The above has been incorporated here by comparing the last HydroShare export date with the last scenario modification date. If a scenario has been modified more recently than the last export, we add a message to the HydroShare Export modal saying so.

Connects #2885 

### Demo

image

https://beta.hydroshare.org/resource/d1f4f20cd6344b7a809552164cc512a6

image

### Notes

I also tried cutting every piece of the export into a Celery Task. That work is vieweable in this diff [hydroshare-celerification-2.patch](https://github.com/WikiWatershed/model-my-watershed/files/2250696/hydroshare-celerification-2.patch.txt). This was suboptimal because since Celery doesn't allow sharing complex objects between tasks, the HydroShare Rest Client could not be shared, and each task would have to create its own which involved making OAuth requests to HydroShare, further slowing down the process. And while splitting up into smaller tasks allowed us to have an unspecified timeout at the Celery level (since each individual task could finish under the default 90s timeout), we would still have a timeout in the front-end, so there is no getting rid of timeouts completely.

### HydroShare Recommendations

There are potential improvements that can be made to the HydroShare API which would help us significantly:

 1. **Support Bulk File Uploads**. Since the most time is spent uploading a file at a time, even though all the files are ready at once, the ability to support bulk file uploads would speed up the process significantly. Probably by an order of magnitude or more. https://github.com/hydroshare/hydroshare/issues/2859 https://github.com/hydroshare/hs_restclient/issues/86

 2. **Allow Overwriting Files**. Currently, if a file already exists in HydroShare, you cannot add it again. We have to first delete that file, then add it. This leads to updates taking twice as long as creation, because for each file we must first delete it then add it. If adding files supported overwriting, this could be done in half the time. https://github.com/hydroshare/hydroshare/issues/2860 https://github.com/hydroshare/hs_restclient/issues/87
 
 3. **Speed Up API Responses**. Currently each API request takes ~1s to turn around. If this was faster the export would be too.

## Testing Instructions

* Populate the `MMW_HYDROSHARE_SECRET_KEY` in the `app` and `worker` VMs using the "HydroShare Beta" item in LastPass
* Go to https://beta.hydroshare.org/ and setup an account if you don't already have one
* Check out this branch, `bundle`, `vagrant ssh app -c 'sudo restart mmw-app'` and `vagrant ssh worker -c 'sudo service celeryd restart'`
* Go to [:8000/](http://localhost:8000/) and log in to an account
* Go to "My Account" → "Linked Accounts" and link to HydroShare (it should automatically connect to the beta site)
* Make a new project. Add a bunch of scenarios to it (I did 11).
* Export to HydroShare. Ensure the export completes successfully.
* Once exported, re-export to HydroShare. Ensure the export completes successfully.
* Make a change to any scenario. Open the HydroShare export dialog. Ensure it has text indicating that there are new changes.
* Test both TR-55 and MapShed projects.

### DIFF
--- a/src/mmw/apps/export/hydroshare.py
+++ b/src/mmw/apps/export/hydroshare.py
@@ -47,9 +47,9 @@ class HydroShareService(OAuth2Service):
 
         return token
 
-    def renew_access_token(self, user):
+    def renew_access_token(self, user_id):
         # TODO Add checks for when user not found, refresh_token is null
-        token = HydroShareToken.objects.get(user=user)
+        token = HydroShareToken.objects.get(user_id=user_id)
         data = {'refresh_token': token.refresh_token,
                 'grant_type': 'refresh_token'}
 
@@ -63,11 +63,11 @@ class HydroShareService(OAuth2Service):
 
         return token
 
-    def get_client(self, user):
+    def get_client(self, user_id):
         # TODO Add check for when user not found
-        token = HydroShareToken.objects.get(user=user)
+        token = HydroShareToken.objects.get(user_id=user_id)
         if token.is_expired:
-            token = self.renew_access_token(user)
+            token = self.renew_access_token(user_id)
 
         auth = HydroShareAuthOAuth2(CLIENT_ID, CLIENT_SECRET,
                                     token=token.get_oauth_dict())
--- a/src/mmw/apps/export/migrations/0002_hydroshare_disable_autosync.py
+++ b/src/mmw/apps/export/migrations/0002_hydroshare_disable_autosync.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+def disable_hydroshare_autosync(apps, schema_editor):
+    HydroShareResource = apps.get_model('export', 'HydroShareResource')
+    HydroShareResource.objects.all().update(autosync=False)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('export', '0001_hydroshareresource'),
+    ]
+
+    operations = [
+        migrations.RunPython(disable_hydroshare_autosync)
+    ]
--- a/src/mmw/apps/export/tasks.py
+++ b/src/mmw/apps/export/tasks.py
@@ -146,7 +146,7 @@ def create_resource(user_id, project_id, params):
         project=project,
         resource=resource,
         title=params.get('title', project.name),
-        autosync=params.get('autosync', True),
+        autosync=params.get('autosync', False),
         exported_at=now()
     )
     hsresource.save()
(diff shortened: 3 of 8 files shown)


rajadain:
Hmm, just noticed that the above demo project doesn't have GMS files. Will look into that.

rajadain:
One strategy that could work for notifying users that they should re-export to HydroShare is to compare the last HydroShare export date with the most recent scenario update date. If a scenario was updated after the export date, then we know there are new changes. Might need to confirm that the scenario update date is available on the front end.

rajadain:
Added a commit that indicates if a project has changed since the last export.

caseycesari:
Taking a look.

rajadain:
Think you may have found a genuine bug. Looking into it now.

rajadain:
That was a great catch. I had only been testing with projects that had previously been exported to HydroShare. Doing so with a new project tripped on not having `hydroshare` defined on the project model the first time. Now we test for that before proceeding. Also renamed some variables to read better.

Ready for another look.

rajadain:
Added a final commit that only re-exports scenarios that have changed, or have not been exported before. If you could test this one last thing, I think we'll be good to go for the near future.

Recommended tests:

* Update branch. `bundle`
* Open existing project that has been exported already. Open the dev tools and network tab.
* Re-export the project. Inspect the request. Ensure it only has Analyze files (which will be skipped on the server side), and no scenario files.
* Add a new scenario. Re-export the project. Inspect the request. Ensure it has the new scenario files.
* Edit an existing scenario (by adding a modification or something). Re-export the project. Inspect the request. Ensure it has the modified scenario files.
* Make a new project and export to HydroShare. Inspect the request. Ensure it has all scenario files.

rajadain:
Since re-exporting existing projects takes longer than exporting them the first time (because all files have to be checked if they exist in HydroShare, and if so, deleted, then re-exported), limiting the number of files we send on the re-export significantly speeds up the process.

rajadain:
Thanks for the multiple thorough reviews! Going to merge this now.

rajadain:
Will squash fixups before merging.

Issue #2895, opened 2018-07-25
Modify geocoder placeholder text to clarify coordinate search


ajrobbins:
We got feedback from Stroud that people are confused about what format of coordinates to enter in the geocoder if they want to search by lat/lon. For example, Google Maps both displays and searches by coordinates like this: 39°57'44.9"N 75°09'16.2"W, while our geocoder only accepts x,y or y,x.

Determine the best approach to surface this requirement with @jfrankl - potentially adding some placeholder text, which may necessitate resizing the geocoder element:

![screen shot 2018-07-25 at 5 04 10 pm](https://user-images.githubusercontent.com/12868342/43227446-c5da6b7a-902c-11e8-864b-e80bdcd25e7c.png)

rajadain:
Follow up from Slack discussion @jfrankl, the approach here will be to have a tooltip that shows up after a small delay (~500ms), recommending the format, like:

> Enter coordinates as **Lat,Lng** (e.g. **32.95,-75.16**)

PR #2900, opened 2018-08-02
Add hint for entering coordinates

rajadain:
## Overview

Since we only accept coordinates in a specific format, add a hint that reveals that format to users. The hint is positioned to the left because there are suggestions below the text box, and no room to the right.

This review should be billed to **NSF**.

Connects #2895 

### Demo

Also tagging @ajrobbins and @jfrankl for visual review.

Note: The GIF is a little sped up. The current delay is set to 500ms, but I can make it longer if needed.

### Notes

It feels like if they are searching for an address we shouldn't show this tool tip, but we need to show it when they are _not_ searching for coordinates since we can't parse it as a coordinate. Thus, it always shows after the delay when the search is focused.

## Testing Instructions

* Try searching for something. Ensure you see the hint pop up.
* Ensure the pop up goes away once the search box is not focused.

### DIFF
--- a/src/mmw/js/src/geocode/templates/search.html
+++ b/src/mmw/js/src/geocode/templates/search.html
@@ -1,6 +1,11 @@
-<input class="search-input" id="geocoder-search" type="text"
-       placeholder="Jump to location, HUC, or coordinates"
->
+<a data-toggle="popover" tabindex="0"
+   data-html="true" data-container="body" role="button"
+   data-placement="left" data-trigger="focus" data-delay="500"
+   data-content="Enter coordinates as <strong>Lat,Lng</strong><br />(e.g. <strong>32.9526,-75.1652</strong>)">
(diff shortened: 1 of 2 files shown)


ajrobbins:
LGTM, but I emailed Steve Kerlin and @aufdenkampe to check, since they hadn't seen the visual design.

caseycesari:
Unrelated to this PR, but related to the feature: If I enter coordinates and then hit enter (before or after the "go to" element appears), a marker gets added and the map is centered on the center of the U.S. in Kansas.

ajrobbins:
+1 acquired from Steve Kerlin @ stroud

aufdenkampe:
This looks good. thanks!

rajadain:
@caseycesari I can't reproduce that. If you can, could you open an issue for it?

Since it isn't blocking the work done here, I'm going to merge this. Thanks for the review everyone!

Issue #2920, opened 2018-08-13
MapShed Compare: Show GWLFE Modifications in Scenario Item popover


kellyi:
Currently the Scenario Item Popovers don't display the modifications correctly, since they expect them to be structured like the TR-55 modifications. We should adjust to show GWLFE data.

Issue #2913, opened 2018-08-10
MapShed Compare: Add Headers


rajadain:
Add the compare view header, which includes tabs for "Hydrology" and "Water Quality", buttons for chart and table view, and close button, same as TR55:

![image](https://user-images.githubusercontent.com/1430060/43979338-664e6416-9cb8-11e8-9cb4-32ad91254a6d.png)

Also add Scenario header row, similar to the above, except:

- There should be no map for a scenario
- There should be a color for each scenario, as shown here:

<img width="568" alt="screen shot 2018-08-10 at 4 12 37 pm" src="https://user-images.githubusercontent.com/4165523/43979285-3b4fa5cc-9cb8-11e8-8179-a63b7743fd19.png">

Use colors as determined in #2912.

Issue #2912, opened 2018-08-10
MapShed Compare: Pick Scenario Colors


rajadain:
Pick a set of colors to be used for scenarios in the MapShed Compare View. This can be a static list, of at least 15 colors. These will be used to differentiate scenarios in the Compare View:

<img width="568" alt="screen shot 2018-08-10 at 4 12 37 pm" src="https://user-images.githubusercontent.com/4165523/43979285-3b4fa5cc-9cb8-11e8-8179-a63b7743fd19.png">

Work with @jfrankl on this.

rajadain:
Relatedly, found [this great blog post](https://blog.graphiq.com/finding-the-right-color-palettes-for-data-visualizations-fcd4e707a283) for picking colors for visualizations.

If we need to go with a different scheme, I recommend picking some from the [Material Color tool](https://material.io/tools/color/).

I also took a look at [ColorBrewer2](http://colorbrewer2.org/), but that only supported schemes up to 9 categories, and we needed more than that.

jfrankl:
@rajadain those colors look good to me. Interested in how they look in the actual application.

rajadain:
Screenshots from Staging:

![image](https://user-images.githubusercontent.com/1430060/44588120-dec77200-a782-11e8-91fd-9f96bbe58e11.png)

![image](https://user-images.githubusercontent.com/1430060/44588127-e6871680-a782-11e8-8586-3e913e849c5e.png)

You can take a look here: https://staging.app.wikiwatershed.org/project/1366/

rajadain:
If you find any design issues that need fixing, please add them here: https://github.com/WikiWatershed/model-my-watershed/issues/2929

Issue #2908, opened 2018-08-10
MapShed Compare: Enable displaying compare view modal & hydrating component with data


kellyi:
Per the wireframes we need to set up the basic compare view modal for MapShed.

1. Click a button, show empty Compare Modal
2. Give Compare Modal copy / reference to Scenario Collection

rajadain:
Shares points with #2913.

PR #2921, opened 2018-08-13
MapShed Compare: Window and Scenario Header Row

rajadain:
## Overview

Enables Compare View for GWLF-E projects. Adds colors for 32 scenarios to be used in the charts. Adds a header row for GWLF-E scenarios that don't have a map, but do have a color badge. The popover works the same as in TR-55, but the modifications aren't listed correctly. That is left for #2920.

Connects #2908 
Connects #2912 
Connects #2913 

### Demo

image

## Testing Instructions

* Make a GWLF-E project. Add a scenario. Try the Compare View. Ensure you see a colored badge against each Scenario's name in the header, and that there is no map.
* Make a TR-55 project. Add a scenario. Try the Compare View. Ensure works same as before.


kellyi:
I have this branch up now and it's working well. TR-55 Compare View still works as before and the MapShed compare view works as depicted above!

rajadain:
Thanks for pairing!

Issue #2909, opened 2018-08-10
MapShed Compare View: Add hydrology chart


kellyi:
Per the wireframes...

<img width="568" alt="screen shot 2018-08-10 at 4 12 37 pm" src="https://user-images.githubusercontent.com/4165523/43979285-3b4fa5cc-9cb8-11e8-8179-a63b7743fd19.png">

... add the hydrology chart.

It should:
    - have one line chart per MapShed attribute (Precip, ET, etc), each for all scenarios
    - have a line for each scenario, colored appropriately
    - span width of entire Compare Modal
    - reuse (with necessary revisions) chart shown in Hydrology tab in Modeling

PR #2923, opened 2018-08-14
Create GWLFE Compare View Hydrology line charts

kellyi:
## Overview

This PR implements the Hydrology line charts for the GWLFE compare view. Was able to reuse the existing line chart function with one small modification; otherwise the main work here is transforming the GWLFE  Hydrology Data into a format usable in the line charts.

Connects #2909 

### Demo

screen shot 2018-08-15 at 11 33 51 am

## Notes

Encountered some quirks vis a vis Marionette lifecycle methods viz the charts wouldn't always be sized correctly unless the class had code both in `onAttach` and `onRender` to create the charts. Practically speaking I think the duplicate call should be a no-op after the second call since there's not really input, but I'm open to having this work differently.

## Testing Instructions

- serve this branch then visit a MapShed project and open the compare view
- verify that you see the hydrology line charts, that they match the wireframes, and that they are created and destroyed properly when toggling through the different GWLFE compare views

### DIFF
--- a/src/mmw/js/src/compare/views.js
+++ b/src/mmw/js/src/compare/views.js
@@ -184,9 +184,11 @@ var CompareWindow2 = modalViews.ModalBaseView.extend({
                 }));
             } else {
                 window.console.warn('TODO: Implement GWLFE Water Quality Chart');
+                this.sectionsRegion.empty();
             }
         } else {
             window.console.warn('TODO: Implement GWLFE Table');
+            this.sectionsRegion.empty();
         }
     },
 
@@ -660,14 +662,68 @@ var LineChartRowView = Marionette.ItemView.extend({
     models: models.LineChartRowModel,
     className: 'compare-chart-row -line',
     template: compareLineChartRowTmpl,
+
+    ui: function() {
+        return {
+            chartElement: '#' + this.model.get('chartDiv'),
+        };
+    },
+
+    onAttach: function() {
+        this.addChart();
+    },
+
+    onBeforeDestroy: function() {
+        this.destroyChart();
+    },
+
+    onRender: function() {
+        this.addChart();
+    },
+
+    addChart: function() {
+        this.destroyChart();
+
+        var scenarioNames = this.model.get('scenarioNames'),
+            chartData = this.model.get('data')
+                .map(function(scenarioData, index) {
+                    return {
+                        key: index,
+                        values: scenarioData.map(function(val, x) {
+                            return {
+                                x: x,
+                                y: Number(val),
+                            };
+                        }),
+                        color: SCENARIO_COLORS[index % 32],
+                    };
+                })
+                .slice()
+                .reverse(),
+            chartOptions = {
+                yAxisLabel: 'Water Depth (cm)',
+                yAxisUnit: 'cm',
+                xAxisLabel: function(xValue) {
+                    return monthNames[xValue];
+                },
+                xTickValues: _.range(12),
+            },
+            tooltipKeyFormatFn = function(d) {
+                return scenarioNames[d];
+            };
+
+        if (chartData.length && this.ui.chartElement) {
+            chart.renderLineChart(this.ui.chartElement, chartData, chartOptions, tooltipKeyFormatFn);
+        }
+    },
+
+    destroyChart: function() {
+        this.ui.chartElement.empty();
+    },
 });
 
 var GWLFEHydrologyChartView = Marionette.CollectionView.extend({
     childView: LineChartRowView,
-
-    onShow: function() {
-        window.console.log('TODO Implement GWLFE Hydrology chart');
-    }
 });
 
 var TR55ChartView = Marionette.CollectionView.extend({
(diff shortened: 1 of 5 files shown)


src/mmw/js/src/compare/views.js
```diff
@@ -1008,10 +1090,85 @@ function getTr55Tabs(scenarios) {
...
+                .get('results')
+                .models
+                .reduce(function(accumulator, next) {
+                    if (next.get('displayName') !== models.constants.HYDROLOGY) {
+                        return accumulator;
+                    }
+
+                    return accumulator.concat(next
+                        .get('result')
+                        .monthly
```
kellyi: It seemed like there would be a `.get('monthly')` way to get this attribute but for whatever reason that didn't work.

src/mmw/js/src/compare/views.js
```diff
@@ -1008,10 +1090,85 @@ function getTr55Tabs(scenarios) {
...
+var getMemoizedHydrologyChartData = _.memoize(mapScenariosToHydrologyChartData);
+
+function mapScenariosToNames(scenarios) {
+    return scenarios
+        .map(function(scenario) {
+            return scenario.get('name');
+        });
+}
+
+var getMemoizedScenarioNames = _.memoize(mapScenariosToNames);
```
kellyi: After the first time this is called per compare view, it'll just be a constant.

rajadain:
Seeing this error:

image

src/mmw/js/src/compare/views.js
```diff
@@ -1026,39 +1036,83 @@ function getTr55Tabs(scenarios) {
...
+                        .get('result')
+                        .monthly
+                        .map(function(result) {
+                            return result[key];
+                        }));
+                }, []);
+        });
+}
+
+var getMemoizedHydrologyChartData = _.memoize(mapScenariosToHydrologyChartData);
```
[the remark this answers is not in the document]
kellyi: Good point. I'll drop the `memoize` stuff.

src/mmw/js/src/compare/views.js
```diff
@@ -558,6 +574,72 @@ var BarChartRowView = Marionette.ItemView.extend({
...
+
+    onRender: function() {
+        this.destroyChart();
+        this.addChart();
+    },
+
+    addChart: function() {
+        var scenarioNames = this.model.get('scenarioNames'),
+            chartData = this.model.get('data')
+                .reduce(function(accumulator, scenarioData, index) {
```
rajadain: `map` is clearer and more compact than `reduce`, especially when the output is expected to be an array of the same size as the input:

```js
var chartData = this.model.get('data')
        .map(function(values, index) {
            return {
                key: index,
                values: values.map(function(value, index2) {
                    return {x: index2, y: Number(value)};
                }),
                color: SCENARIO_COLORS[index % 32],
            };
        }).reverse();
```

kellyi:
> That's because the output of .get('result') is a plain JavaScript object, not a Backbone model.

Makes sense. Because this is only used inline it seems better to keep `.monthly` rather than trying to wrap it in a model just to use `.get('monthly')`.

src/mmw/js/src/compare/views.js
```diff
@@ -1008,10 +1090,85 @@ function getTr55Tabs(scenarios) {
...
+        .map(function(scenario) {
+            return scenario
+                .get('results')
+                .models
+                .reduce(function(accumulator, next) {
+                    if (next.get('displayName') !== models.constants.HYDROLOGY) {
+                        return accumulator;
+                    }
+
+                    return accumulator.concat(next
```
rajadain: Similarly to the above, can we use `map(...).reverse()` here instead of `.reduce()` with `accumulator.concat`?

src/mmw/js/src/compare/views.js
kellyi: Sounds good to me.

rajadain:
> Because this is only used inline it seems better to keep `.monthly` rather than trying to wrap it in a model just to use `.get('monthly')`.

Yes, and also because the contents of the `result` vary between model packages, and for flexibility were never formalized on the front-end. It has been enough to keep them as plain objects so far.

src/mmw/js/src/compare/views.js
kellyi: This `reduce` combines a `filter` and a `map` operation so changing it will involve another pass. In this case the output is not the same length as the input.

rajadain: Ah I see. Let's keep it then.

src/mmw/js/src/compare/views.js
```diff
@@ -1026,39 +1036,83 @@ function getTr55Tabs(scenarios) {
...
+                            return result[key];
+                        }));
+                }, []);
+        });
+}
+
+var getMemoizedHydrologyChartData = _.memoize(mapScenariosToHydrologyChartData);
+
+function mapScenariosToNames(scenarios) {
+    return scenarios
```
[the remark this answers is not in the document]
kellyi: Nice -- that's definitely more succinct. Adjusted this in e319ac8

src/mmw/js/src/compare/views.js
```diff
@@ -576,14 +578,66 @@ var LineChartRowView = Marionette.ItemView.extend({
...
+        this.addChart();
+    },
+
+    addChart: function() {
+        var scenarioNames = this.model.get('scenarioNames'),
+            chartData = this.model.get('data')
+                .reduce(function(accumulator, scenarioData, index) {
+                    return [
+                        {
+                        key: index,
```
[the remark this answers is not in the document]
kellyi: Good catch. I incidentally fixed this in 6d8e7ee when changing this operation from `reduce` to `map`.

kellyi:
I did see that `cannot read property 'x'` error once while toggling the compare view open and closed, but I'm having trouble reliably creating it.

I also noticed another bug which is that sometimes the first chart gets rendered as very tiny:

screen shot 2018-08-16 at 11 09 06 am

I'm guessing that both of these might be timing issues?

rajadain:
I also saw that tiny chart issue. How do you mean "timing issues"? Is the order of events not deterministic here?

kellyi:
re: timing: I was wondering whether in some cases D3 tries to create the chart before the parent DOM element fully exists & has a size to reference.

kellyi:
https://github.com/WikiWatershed/model-my-watershed/commit/0df45e3 changes the lifecycle methods which create when the component's created and remove it when the component's destroyed.

It also adds a 50 ms delay to the function call which actually creates the line chart to try to circumvent some bugs whereby the chart element seemed to get created before the parent element was set up.

rajadain:
Taking another look.

rajadain:
Still seeing the issue where the charts are tiny after closing and opening the compare view:

image

kellyi:
I think the tiny chart issue happened because NVD3 prefers having fixed values for div size rather than proportional, so I changed the width/height of the div to 800px x 350 px. Going to toggle for a bit to see whether I can still create the tiny charts.

Toggling a lot did seem to show that x is not defined error periodically. Not sure what to do about it since it emanates from the vendor bundle and doesn't seem to affect the chart functionality.

kellyi:
The `x is not defined` error seems to be related into a bug/quirk of NVD3 which applies to versions of D3 > 3.3 specifically for scatter and line charts where some of the series share points:

https://github.com/novus/nvd3/issues/330

I'm not sure what is the preferred way to fix this; it appears that we're using a forked version of NV alongside D3 version 3.5.5. I tried out the `chart.useVoronoi(false)` suggestion from that GitHub issues thread but it didn't prevent the error.

rajadain:
Alright let's make a follow up issue for that. I'll take another look ignoring that soon.

kellyi:
Sounds good. I'm still ironing out the tiny chart issue. I thought setting exact element widths would prevent it from happening but I can still periodically re-create it.

kellyi:
OK! I have been clicking back and forth to open and close the charts in both Firefox and Chrome and I haven't been able to get the tiny chart bug to happen again. Downside is that the charts now have a fixed width of 700px

kellyi:
Oof, noticed that we're now missing the scenario names, so will fix.

kellyi:
Made https://github.com/WikiWatershed/model-my-watershed/issues/2926 for the tiny chart bug. Otherwise this is ready for review.

rajadain:
Made #2928 for the NVD3 bug.

kellyi:

Issue #2915, opened 2018-08-10
MapShed Compare: Add Hydrology Table


rajadain:
Table view should have a dropdown to select which MapShed attribute to show (Precip, ET, etc). For that attribute, there should be a row for every month, and a value for every scenario, as mocked up here:

![image](https://user-images.githubusercontent.com/1430060/43979769-caea5e06-9cb9-11e8-82c5-d515c4898ffb.png)

PR #2924, opened 2018-08-15
Create hydrology tables

kellyi:
## Overview

This PR implements the hydrology tables as depicted below. Was able to reuse the the table view from TR55, with the data adjusted to reformat it for GWLFE.

The bulk of the change here is a reduce function to transform the GWLFE data into a shape for display in the tables.

Connects #2915 

### Demo

hydrologytable

## Testing Instructions

- serve this branch, then visit a MapShed project and open the compare view
- toggle the hydrology table and then verify that you see data and can toggle through the attributes using the selection box

### DIFF
--- a/src/mmw/js/src/compare/views.js
+++ b/src/mmw/js/src/compare/views.js
@@ -174,20 +174,24 @@ var CompareWindow2 = modalViews.ModalBaseView.extend({
     },
 
     showGWLFESectionsView: function() {
-        if (this.model.get('mode') === models.constants.CHART) {
-            var activeCollection = this.model.get('tabs').findWhere({ active: true });
+        var activeTab = this.model.get('tabs').findWhere({ active: true }),
+            activeName = activeTab.get('name'),
+            activeMode = this.model.get('mode');
 
-            if (activeCollection.get('name') === models.constants.HYDROLOGY) {
-                this.sectionsRegion.show(new GWLFEHydrologyChartView({
-                    model: this.model,
-                    collection: activeCollection.get('charts'),
-                }));
-            } else {
-                window.console.warn('TODO: Implement GWLFE Water Quality Chart');
-                this.sectionsRegion.empty();
-            }
+        if (activeMode === models.constants.CHART &&
+            activeName === models.constants.HYDROLOGY) {
+            this.sectionsRegion.show(new GWLFEHydrologyChartView({
+                model: this.model,
+                collection: activeTab.get('charts'),
+            }));
+        } else if (activeMode === models.constants.TABLE &&
+                   activeName === models.constants.HYDROLOGY) {
+            this.sectionsRegion.show(new GWLFEHydrologyTableView({
+                model: this.model,
+                collection: activeTab.get('table'),
+            }));
         } else {
-            window.console.warn('TODO: Implement GWLFE Table');
+            window.console.log('TODO Implement GWLFE Water Quality table & chart');
             this.sectionsRegion.empty();
         }
     },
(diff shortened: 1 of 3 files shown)


src/mmw/js/src/compare/views.js
```diff
@@ -174,20 +174,24 @@ var CompareWindow2 = modalViews.ModalBaseView.extend({
...
+            }));
+        } else if (activeMode === models.constants.TABLE &&
+                   activeName === models.constants.HYDROLOGY) {
+            this.sectionsRegion.show(new GWLFEHydrologyTableView({
+                model: this.model,
+                collection: activeTab.get('table'),
+            }));
         } else {
-            window.console.warn('TODO: Implement GWLFE Table');
+            window.console.log('TODO Implement GWLFE Water Quality table & chart');
```
kellyi: Committed this by mistake but I guess it'll be removed semi-immediately anyway.

src/mmw/js/src/compare/views.js
```diff
@@ -201,10 +205,21 @@ var CompareWindow2 = modalViews.ModalBaseView.extend({
             var selected =
                 activeTab.get('selections').findWhere({ active: true });
 
+
```
kellyi: Also committed by mistake, but this function's going to get removed when the water quality work is in.

rajadain:
This looks really good! I noticed that the pagination wasn't working out of the box as I expected, so I looked into the code, and found that extending the existing TableRowView and TableView works just as well, makes pagination work, and is more compact:

```diff
diff --git a/src/mmw/js/src/compare/views.js b/src/mmw/js/src/compare/views.js
index 780a7634..79195ba4 100644
--- a/src/mmw/js/src/compare/views.js
+++ b/src/mmw/js/src/compare/views.js
@@ -747,32 +747,6 @@ var GWLFEHydrologyChartView = Marionette.CollectionView.extend({
     childView: LineChartRowView,
 });
 
-var GWLFEHydrologyTableRowView = Marionette.ItemView.extend({
-    models: models.MonthlyTableRowModel,
-    className: 'compare-table-row -hydrology',
-    template: compareTableRowTmpl,
-
-    modelEvents: {
-        'change': 'render',
-    },
-
-    templateHelpers: function() {
-        var selectedAttribute = this.model.get('selectedAttribute');
-
-        return {
-            values: this.model
-                .get('values')
-                .map(function(v) {
-                    return v[selectedAttribute];
-                }),
-        };
-    },
-});
-
-var GWLFEHydrologyTableView = Marionette.CollectionView.extend({
-    childView: GWLFEHydrologyTableRowView,
-});
-
 var TR55ChartView = Marionette.CollectionView.extend({
     childView: BarChartRowView,
 
@@ -856,6 +830,22 @@ var TableView = Marionette.CollectionView.extend({
     }
 });
 
+var GWLFEHydrologyTableRowView = TableRowView.extend({
+    className: 'compare-table-row -hydrology',
+
+    templateHelpers: function() {
+        var selectedAttribute = this.model.get('selectedAttribute');
+
+        return {
+            values: this.model.get('values').pluck(selectedAttribute)
+        };
+    },
+});
+
+var GWLFEHydrologyTableView = TableView.extend({
+    childView: GWLFEHydrologyTableRowView,
+});
+
 var CompareWindow = Marionette.LayoutView.extend({
     //model: modelingModels.ProjectModel,
 
```

I would recommend adding that change and this will be good to merge! Will take care of the other demo stuff in this PR in upcoming Water Quality PRs.

kellyi:
Sounds good to me. I'm making that change & will amend the existing commit to include it.

For whatever reason I've not been able to get `.pluck` to work, so I'll keep the `.map` op instead.

kellyi:
Here's the table view with the column toggles:

screen shot 2018-08-17 at 5 02 55 pm

kellyi:
Great! Thanks for your help with this!

Issue #2910, opened 2018-08-10
MapShed Compare View: Add Water Quality chart


kellyi:
Per the wireframes...

<img width="578" alt="screen shot 2018-08-10 at 4 14 35 pm" src="https://user-images.githubusercontent.com/4165523/43979367-835e0f52-9cb8-11e8-9b06-8fdecefa664e.png">

... add the Water Quality Chart.

It should:
    - have two dropdowns for selecting which chart to show
    - allow using the first dropdown to select between summary and land cover tables
    - allow using the second dropdown to select row of selected table
    - have one bar chart per column (Sediment, Total Nitrogen, Total Phosphorus), which includes all scenarios color appropriately

Note:
    * Need to talk to Matt / Client to confirm dropdowns

rajadain:
May also need to talk to someone to figure out how subbasin fits into this.

PR #2925, opened 2018-08-16
MapShed Compare: Add Dropdown for Attributes

rajadain:
## Overview

Since the MapShed Compare view involved comparing multiple attributes for every scenario, we cannot show all the information in a single table (or in the case of Water Quality, a single chart). Thus, we need to be able to pick which attribute to compare.

This PR adds a dropdown menu above the comparison sections that is populated with the tab's options, and a demo listener that shows how other items (such as table or chart values) may listen for changes here.

Connects #2910 

### Notes

We need the dropdown to be visible for all Water Quality tabs, but only for Hydrology table. That will be handled in subsequent PRs.

## Testing Instructions

* Open a MapShed project and open Compare view
* Ensure you see the dropdown for Water Quality but not for Hydrology
* Ensure selecting items in the dropdown logs the group and item name to the console

### DIFF
--- a/src/mmw/js/src/compare/views.js
+++ b/src/mmw/js/src/compare/views.js
@@ -181,10 +181,30 @@ var CompareWindow2 = modalViews.ModalBaseView.extend({
                 this.model.get('mode') === models.constants.CHART &&
                 activeTab.get('name') === models.constants.HYDROLOGY;
 
+        var demoFunction = function() {
+            var selected =
+                activeTab.get('selections').findWhere({ active: true });
+
+            console.log(
+                selected.get('group') + ' ' + selected.get('name'));
+        };
+
+        // TODO Remove demo listener
+        this.model.get('tabs').forEach(function(tab) {
+            var selections = tab.get('selections');
+
+            if (selections) {
+                selections.off();
+            }
+        });
+
         if (activeTab.get('selections') && !isHydrologyChart) {
             this.selectionRegion.show(new SelectionView({
                 model: activeTab,
             }));
+
+            // TODO Demo Listening for Changes
+            activeTab.get('selections').on('change', demoFunction);
         } else {
             this.selectionRegion.empty();
         }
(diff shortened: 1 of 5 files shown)


kellyi:
This is working well! Can we give it some provisional styling, too, to make it match the wireframes?

This change...

```
index 7e811b61..7d9fd57e 100644
@@ -318,6 +318,13 @@ $compare-chart-table-height: calc(100vh - #{$height-lg} - #{$compare-footer-heig
       flex-direction: column;
       height: calc(100% - 20px);
       -webkit-font-smoothing: antialiased;
+
+      > #compare-selection-region {
+          > select {
+              width: 225px;
+              margin: 1.5rem;
+          }
+      }
   }
 
   .compare-actions {
```

Adjusts it to look like this:

screen shot 2018-08-16 at 12 01 01 pm

rajadain:
Great suggestion! I went with `250px` to accommodate IE11 style dropdowns with the longest name:

image

src/mmw/js/src/compare/views.js
```diff
@@ -1030,6 +1131,7 @@ function getGwlfeTabs(scenarios) {
             name: 'Water Quality',
             table: qualityTable,
             charts: qualityCharts,
+            selections: qualitySelections,
```
kellyi: The hydrology tab will only show the selection dropdown for the table and not for the line chart. Given that, I think it makes sense to split this into two variables -- `chartSelections` and `tableSelections` -- and redo the check on whether to display it to check on the one for visualization type.

kellyi: This is potentially not blocking, btw -- if we agree to adjust it on the fly we can leave it as is -- or we can make the update quickly here and then adjust the `if (selections)` check above to be `if (chartSelections || tableSelections)` and then determine how to handle each specific case later.

rajadain: I did think about that briefly, although we may have to figure out a way to keep `chartSelections` and `tableSelections` in sync for Water Quality. I'll think more about this.

kellyi: Would it work to have water quality chart and table selections share a single collection rather than making a new one for each?

kellyi: Another way to handle it, btw, is to treat hydrology line charts as a special case and then pass a bool somewhere to suppress the dropdown only for that view.

rajadain:
Added a special exception for Hydrology Charts in bf4c067.

kellyi:
Cool, taking another look.

kellyi:
Working well!

rajadain:
Alright, the Dropdown Menu is in. Feel free to use this for the Hydrology Table.

Issue #2907, opened 2018-08-10
Fix TR-55 compare view CSV download


kellyi:
Fix CSV export on TR55
    - Currently download file has unescaped "\n" instead of actual new line and is quoted

PR #2936, opened 2018-08-22
Compare View Revamp: Fix TR-55 Downloads

rajadain:
## Overview

Previously, `downloadAsFile` always stringified the input data, assuming it to be a JSON object. This was not the case for the TR-55 Compare View download, which was a CSV string. Stringifying that led to all the formatting elements, such as quotes and new-lines, being escaped in the final download, thus corrupting it.

We replace `downloadAsFile` by two derived methods, `downloadJson` and `downloadText`, each of which calls the now renamed `_downloadAsFile` with a parameter that tells it to stringify or not.

All existing use cases have been replaced by the appropriate alternative. This allows JSON downloads to continue to work as before, while also allowing CSV downloads to work correctly.

Connects #2907 

### Demo

image

## Testing Instructions

* Open a TR-55 project with multiple scenarios
* Open Compare View. Click the download button in the top right. Ensure the downloaded file is legit.
* Ensure you can still download the shape GeoJSON from Analyze, and it is valid JSON.

### DIFF
--- a/src/mmw/js/src/analyze/views.js
+++ b/src/mmw/js/src/analyze/views.js
@@ -440,8 +440,8 @@ var AoiView = Marionette.ItemView.extend({
     },
 
     downloadGeojson: function() {
-        utils.downloadAsFile(this.model.get('shape'),
-                             this.model.get('place') + '.geojson');
+        utils.downloadJson(this.model.get('shape'),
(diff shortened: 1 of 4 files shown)


rajadain:
Thanks for all the reviews!

Issue #2124, opened 2017-08-08
Compare View Revamp: Remove Old Unused Code


rajadain:
Once the [Compare View Revamp](https://github.com/WikiWatershed/model-my-watershed/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20Compare%20View%20Revamp) is complete, remove old and unused code of the previous compare view.

rajadain:
Consider standardizing naming from of `TR55` and `GWLFE` prefixed items to `Tr55` and `Gwlfe` as done elsewhere in the code.

PR #2940, opened 2018-08-27
Clean up compare view code

kellyi:
## Overview

This PR undertakes some cleanup of the compare view code. I broke the commits up so that they should be straightforward to follow, but the changes here include:

- removing lots of unused code from the old compare view, including templates, routes, views, and the entire controller.js file
- renaming some TR55 and GWLFE prefixed variables to Tr55 and Gwlfe, respectively, to match how the typical casing for those names throughout the app
- moving constants from models & views into a separate constants.js file
- creating a separate utils file for two functions which are mostly about data-formatting and not precisely the views. Conceivably I guess these things could go in a controller, but they mostly just reduce one list to another list immutably so they seem more like utils to me

Connects #2124 

## Testing Instructions

- serve this branch then load a TR55 compare view and click through it to verify that it still functions and appears exactly as it did before
- repeat with a GWLFE compare view, verifying that it also appears and functions the same as before


kellyi:
Thanks! Going to merge.

Issue #2937, opened 2018-08-23
MapShed BMPs: Reclassify Developed Land Categories


rajadain:
* Separate all four NLCD developed land categories for MapShed; currently these four categories map to three Mapshed codes.
  - [x] Instead, map NLCD 21 to Mapshed 17 (low density open space)
  - [x] Rename the label in GMS to “Ld_Open_Space” instead of “Ld_Residential”
  - [x] Mapshed Water Quality results table should be modified to add a Low-Density Open row of results
* Update % Impervious surface values for all four land categories. These values are currently hard coded, so this task will just require a 1:1 change in values to figures that are more accurate for the NLCD
  - [x] Ld_Mixed: from 15% to 35% impervious
  - [x] Md_Mixed: from 52% to 65%
  - [x] Hd_Mixed: from 87% to 90%
  - [x] Ld_Residential: from 15% to Ld_Open_Space 10%
* [x] Remove the Other Upland Areas row in the Water Quality results table, as it is not currently used

Table Mapping Old Results to New:

| NLCD2011 name               | NLCD ID | MapShed name                   | Mapshed ID | %Imp old | %Imp new |
|-----------------------------|---------|--------------------------------|------------|----------|----------|
| Developed, Low Intensity    | 22      | Ld_Mixed                       | 2          | 15%      | 35%      |
| Developed, Medium Intensity | 23      | Md_Mixed                       | 20         | 52%      | 65%      |
| Developed, High Intensity   | 24      | Hd_Mixed                       | 3          | 87%      | 90%      |
| Developed, Open Space       | 21      | Ld_Residential / Ld_Open_Space | 17         | 15%      | 10%      |

**Note**: Existing projects do not need to be updated.

PR #2944, opened 2018-08-28
MapShed BMPs: Reclassify Developed Land Categories, and other adjustments

rajadain:
## Overview

This reassigns NLCD 21 Open Space to go to Low Density Residential instead of Low Density Mixed, and updates the impervious percentage values for all urban land use types. Also upgrades GWLF-E, which in https://github.com/WikiWatershed/gwlf-e/pull/84 renames Low Density Residential to Low Density Open Space and adds it to the output. It also removes Other Upland Areas from the output.

Connects #2937 

### Demo

image

### Notes

Once this is approved, I will merge the GWLF-E PR and make a new release, and update the final commit here to point to that new release instead of the PR branch.

It was easier to remove "Other Upland Areas" from gwlf-e than to remove it by hand in all the various places in MMW (model results, compare view, CSV exports). This does change the output of gwlf-e, and we'll have to bump up the version number accordingly.

This new value won't show up in Subbasin because SRAT doesn't have a key for this. If they ever add support, we'll have to add that mapping here:

https://github.com/WikiWatershed/model-my-watershed/blob/5260b2e1fd7ce6dbc77e10dc4834b246e03513ff/src/mmw/mmw/settings/gwlfe_settings.py#L623-L638

https://github.com/WikiWatershed/model-my-watershed/blob/5260b2e1fd7ce6dbc77e10dc4834b246e03513ff/src/mmw/mmw/settings/gwlfe_settings.py#L640-L656

## Testing Instructions

* Check out this branch
* Install the latest version of `gwlf-e` in both `app` and `worker` by doing the following:

      $ vagrant ssh app -c 'sudo -H pip uninstall gwlf-e && cd /opt/app && sudo -H pip install -r requirements/development.txt && sudo restart mmw-app'
      $ vagrant ssh worker -c 'sudo -H pip uninstall gwlf-e && cd /opt/app && sudo -H pip install -r requirements/development.txt && sudo service celeryd restart'

* Go to [:8000](http://localhost:8000) and make a new MapShed project.
* Go to Water Quality results. Ensure you see Low-Density Open Space results. Ensure you don't see Other Upland Area results.
* Generate a GMS file. Ensure you see Ld_Open_Space in the file and not Ld_Residential.
* Open Compare View. Ensure the land use categories there match what you saw in model results.


kellyi:
Everything here is working well! Mixed-projects (e.g. those with scenarios created before and after the GWLFE changes here) also work well.

I did notice that "Other Upland Areas" is hard-coded in the GWLFE Water Quality Compare view dropdown. On `develop` that lives in `views.js` but I moved it to a `constants.js` file as part of #2940: 

https://github.com/WikiWatershed/model-my-watershed/pull/2940/commits/56936e71d067709f5d49db698b82c06d0312599b

rajadain:
Rebased on `develop`. Added a commit that limits the sources in Compare View to those that are on every scenario, so we won't see Low-Density Open Space in the dropdown if all scenarios don't have it. Other Upland Areas is removed completely, but it was always 0, so its removal has no ill effect.

rajadain:
Also added a note on how this doesn't change subbasin results because SRAT doesn't support Low-Density Open Space.

kellyi:
Set this up to take another look. For projects with mixed scenarios, the "Low Density Open Space" option's gone -- as expected -- from the compare view dropdown list. For entirely new projects, "Low Density Open Space" is there:

screen shot 2018-08-28 at 6 07 14 pm

src/mmw/js/src/compare/utils.js
```diff
@@ -57,8 +58,33 @@ function mapScenariosToHydrologyTableData(scenarios) {
...
+function getQualitySelections(scenarios) {
+    return _.filter(constants.gwlfeQualitySelectionOptionConfig,
+                    function(selection) {
+        return scenarios.every(function(scenario) {
+            var result = scenario
+                         .get('results')
+                         .findWhere({ name: 'quality' })
+                         .get('result');
+
+            return _.some(result[selection.group], function(load) {
```
kellyi: Totally fine to leave as is, but since `.every` is used above, this implementation could potentially also replace `_.some` and `_.filter` with the vanilla JS versions.

rajadain: `scenarios` is a collection, not an array, so that `every` is actually using `underscore`'s equivalent. Thus, they are all using libraries.

rajadain:
Thanks for the double review!

Issue #2955, opened 2018-09-04
MapShed Compare: Missing Scrollbar in Chrome on Windows / Linux


rajadain:
Client reports that the Compare View does not have a visible / draggable scrollbar on Windows / Linux, counter to OS conventions, which makes it difficult to scroll without a scroll wheel or trackpad.

PR #2965, opened 2018-09-12
Unhide compare view scrollbar

kellyi:
## Overview

The compare view's scrollbar was obscured a bit behind another div, so
we change the scrollable section's z-index to 1000 to make it apparent.

Connects #2955

### Demo

screen shot 2018-09-12 at 5 22 31 pm

## Testing Instructions

- serve this branch then visit the site in Chrome for Windows 7 or 10
- create or load a MapShed project then open the compare view to verify that the compare view's vertical scrollbar is visible


kellyi:
Thanks!

Issue #2959, opened 2018-09-05
MapShed BMPs: Add Waste Water Tab


rajadain:
Using this as a general mockup:

![image](https://user-images.githubusercontent.com/1430060/45119060-46d56b00-b128-11e8-9af5-f8f62a8c96cb.png)

Make a tab for Waste Water with the following fields:

![image](https://user-images.githubusercontent.com/1430060/45118403-6b304800-b126-11e8-8147-7710f8de400f.png)

Notes to keep in mind:

- The mockup does not show headings. Adapt them to our current styling.
- Each row should have one corresponding text box. This box should have a placeholder value, coming from `scenario.get('gis_data')`.
- If a user overwrites a value, they should see a blue revert button 🔁 next to the text box
- If a user hovers on that button, they should see a popup that says "Revert to Auto-estimated" and the original value
- If a user clicks that button, the value of the text box should be cleared so that the placeholder returns
- When the user clicks "Save", the values they override should be added as a modification named `waste_water` to the scenario's modifications collection

The mapping between the values and the GMS file are as follows:

| Dialog Field                   | MapShed Variable | Notes                                                           |
|--------------------------------|------------------|-----------------------------------------------------------------|
| Annual TN Load (kg/yr)         | PtSrcNitro(m)    | Need to divide user value by 12 and distribute equally in array |
| Annual TP Load (kg/yr)         | PtSrcPhos(m)     | Need to divide user value by 12 and distribute equally in array |
| Daily Effluent Discharge (MGD) | PtSrcFlow(m)     | Same value for all 12 slots in array                            |
| Normally Functioning Systems   | NormalSys        |                                                                 |
| Surface Failures               | ???              |                                                                 |
| Subsurface Failures            | ???              |                                                                 |
| Direct Discharges              | ???              |                                                                 |

Requires #2958

Issue #2962, opened 2018-09-06
MapShed Manual Entry: Add BMP Efficiencies Tab


rajadain:
Similar to #2959, add the ability to edit the efficiencies of these BMPs:

| BMP                         | Nitrogen Efficiency | Phosphorus Efficiency | Sediment Efficiency | Notes                                                 |
|-----------------------------|---------------------|-----------------------|---------------------|-------------------------------------------------------|
| Cover Crops                 | `n63`               | `n71`                 | `n79`               |                                                       |
| No Till Agriculture         | `n65`               | `n73`                 | `n81`               | ~~Different values, same target as Conservation Tillage~~ Read Only |
| Conservation Tillage        | `n65`               | `n73`                 | `n81`               | Read Only                                                      |
| Reduced Tillage             | `n65`               | `n73`                 | `n81`               | ~~Different values, same target as Conservation Tillage~~ Read Only |
| Nutrient Management         | `n70`               | `n78`                 | N/A                 |                                                       |
| Livestock Waste Management  | `n85h`              | `n85i`                | N/A                 |                                                       |
| Poultry Waste Management    | `n85j`              | `n85k`                | N/A                 |                                                       |
| Vegetation Buffer Strips    | `n64`               | `n72`                 | `n80`               | Same for Urban and Rural                              |
| Streambank Fencing          | `n69`               | `n77`                 | `n85`               |                                                       |
| Streambank Stabilization    | `n69c`              | `n77c`                | `n85d`              | Same for Urban and Rural                              |
| Surface Water Retention     | `n70b`              | `n78b`                | `n85c`              |                                                       |
| Infiltration / Bioretention | `n71b`              | `n79b`                | `n79c`              |                                                       |

~~When the value is updated here, these values should be used when area-weighting Crop Tillage modifications (No Till Ag, Conservation Tillage, Reduced Tillage). Otherwise it should just override the respective spot in the GMS data model.~~

Requires #2958

rajadain:
From BME:

> Sorry, I thought I had responded to this. My opinion would be to show the 3 different tillage efficiencies, but not allow the user to edit them since I agree that devising a more complex approach to do this is outside your scope of work.

Issue #2961, opened 2018-09-05
MapShed BMPs: Add Other Model Data


rajadain:
Very similar to #2959, with the following data:

![image](https://user-images.githubusercontent.com/1430060/45121539-72a81f00-b12f-11e8-8e0a-481b0d454b21.png)

With the following mappings:

| Dialog Field                                  | MapShed Variable | Notes               |
|-----------------------------------------------|------------------|---------------------|
| Streambank Erosion Adjustment Factor (0-2)    | SedAAdjust       | ??? Need to Confirm |
| Sediment Delivery Ratio (0-1)                 | SedDelivRatio    |                     |
| Fraction of Cropland Tile Drained (0-1)       | TileDrainDensity | ??? Need to Confirm |
| Avg. Tile Drain N Concentration (mg/l)        | TileNconc        |                     |
| Avg. Tile Drain P Concentration (mg/l)        | TilePConc        |                     |
| Avg. Tile Drain Sediment Concentration (mg/l) | TileSedConc      |                     |
| Groundwater N Concentration (mg/l)            | GrNitrConc       |                     |
| Groundwater P Concentration (mg/l)            | GrPhosConc       |                     |
| N Wetland / Water Filtration Fraction (0-1)   | ???              | ??? Need to Confirm |
| P Wetland / Water Filtration Fraction (0-1)   | ???              | ??? Need to Confirm |
| TSS Wetland / Water Filtration Fraction (0-1) | ???              | ??? Need to Confirm |
| N Wetland / Water Retention Fraction (0-1)    | RetentNLake      |                     |
| P Wetland / Water Retention Fraction (0-1)    | RetentPLake      |                     |
| TSS Wetland / Water Retention Fraction (0-1)  | RetentSedLake    |                     |

Requires #2958

Issue #2960, opened 2018-09-05
MapShed BMPs: Add Animals Tab


rajadain:
Very similar to #2959, with the following values:

![image](https://user-images.githubusercontent.com/1430060/45120922-910d1b00-b12d-11e8-8a30-b68aa78a2f9d.png)

and the following mappings:

| Dialog Field                               | MapShed Variable | Notes             |
|--------------------------------------------|------------------|-------------------|
| Dairy Cows                                 | NumAnimals[0]    |                   |
| Are Dairy Cows Allowed to Graze?           | GrazingAnimal[0] | `<Yes>` or `<No>` |
| Beef Cows                                  | NumAnimals[1]    |                   |
| Broilers                                   | NumAnimals[2]    |                   |
| Layers                                     | NumAnimals[3]    |                   |
| Hogs / Swine                               | NumAnimals[4]    |                   |
| Are Hogs / Swine Allowed to Graze?         | GrazingAnimal[4] | `<Yes>` or `<No>` |
| Sheep / Goats                              | NumAnimals[5]    |                   |
| Are Sheep / Goats Allowed to Graze?        | GrazingAnimal[5] | `<Yes>` or `<No>` |
| Horses                                     | NumAnimals[6]    |                   |
| Are Horses Allows to Graze?                | GrazingAnimal[6] | `<Yes>` or `<No>` |
| Turkeys                                    | NumAnimals[7]    |                   |
| Fraction of Livestock Served by AWMS (0-1) | AWMSGrPct        |                   |
| Fraction of Poultry Served by AWMS (0-1)   | AWMSNgPct        |                   |

Requires #2958

PR #2966, opened 2018-09-12
MapShed Manual Entry: Add Waste Water Tab

rajadain:
## Overview

This PR is a double-header: it sets up the plumbing for configuring sections and fields for all tabs, and adds the specific sections and fields for Waste Water. Other tabs are temporarily hidden, to be re-added with the their fields in later cards.

Since the user-facing field configuration doesn't always map one-to-one with the underlying GMS representation, each field is configured with a "calculator" helper which provides methods for converting to and from the GMS format. Some calculator helpers have already been added, with more expected to be added when required in the future.

The manual entries are added as special modifications to the scenario. By adding them as modifications, we take advantage of existing code which overrides the GMS variables with these changes, both in the front- and back-ends. This also scopes them to a scenario, so different scenarios can have different manual entries. They are also saved to the scenario, so when a project is reloaded, all the manual entries are retrieved and reused.

The special modifications are prefixed with `entry_` so they can be ignored by certain parts of the UI, such as the tiles in the top-right added for each active BMP.

Connects #2959 

### Demo

image

## Testing Instructions

* Create or open a MapShed project. Create or open a scenario other than "Current Conditions".
* Open the settings modal. Ensure there are no errors in the console.
* Add a value to one of the fields. Ensure you see a revert / undo button once the field loses focus.
* Hover over the undo button. Ensure it shows the previous / auto-estimated value.
* Click the undo button. Ensure it removes the user value.
* Add a user value and click "Save". Ensure the results are re-run.
* Open the settings modal again. Ensure the user value is still there.
* Go to a different scenario, and open the settings modal. Ensure the user values from the previous scenario did not get carried over, and this has a fresh state.
* Go back to the previous scenario. Ensure its settings still persist.
* Reload the page. Open the settings modal. Ensure the settings were persisted to the database and load correctly.
* Export a GMS file from Current Conditions. Export a GMS file from a scenario with manual entries. Compare the two files, and ensure the diff corresponds to the manual entries. Refer to #2959 to see where the values for each field should reflect.

### DIFF
--- a/src/mmw/js/src/modeling/gwlfe/entry/calcs.js
+++ b/src/mmw/js/src/modeling/gwlfe/entry/calcs.js
@@ -1,0 +1,74 @@
+"use strict";
+
+var _ = require('lodash'),
+    turfArea = require('turf-area'),
+    App = require('../../../app');
+
+/**
+ * This module exports a set of objects for calculating outputs of, and auto-
+ * estimatates for, a manual entry field for GWLF-E.
+ *
+ * Each calculator object has two methods: `toOutput` which given a user
+ * value converts it to the expected GMS format, and `getAutoValue` that
+ * gets a user-friendly value from the GMS dataModel given a field name.
+ *
+ * Each manual entry field will be configured with one of these calculators.
+ */
+
+module.exports = {
+    // Takes the userValue and uses it directly in the output
+    Direct: {
+        toOutput: function(userValue) {
+            return userValue;
+        },
+        getAutoValue: function(fieldName, dataModel) {
+            return dataModel[fieldName];
+        },
+    },
+
+    // Takes the userValue and makes an array with the value repeated 12 times
+    Array12: {
+        toOutput: function(userValue) {
+            return _.fill(new Array(12), userValue);
+        },
+        getAutoValue: function(fieldName, dataModel) {
+            return dataModel[fieldName][0];
+        },
+    },
+
+    // Takes the userValue and divides into an array of 12 values evenly
+    EqualMonths: {
+        toOutput: function(userValue) {
+            return _.fill(new Array(12), userValue / 12);
+        },
+        getAutoValue: function(fieldName, dataModel) {
+            return _.sum(dataModel[fieldName]);
+        },
+    },
+
+    // Takes the user value in million gallons per day and converts it to
+    // centimeters per month, and applies it to twelve months. Equivalent to
+    // point_source_discharge method in mapshed/calcs.py
+    PointSourceDischarge: {
+        toOutput: function(userValue) {
+            var MONTHDAYS = [31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31],
+                AREA_SQM = turfArea(App.currentProject.get('area_of_interest')),
+                M3_PER_MGAL = 3785.41178,
+                CM_PER_M = 100.0;
+
+            return MONTHDAYS.map(function(DAYS_PER_MONTH) {
+                return userValue *
+                    (DAYS_PER_MONTH * M3_PER_MGAL * CM_PER_M / AREA_SQM);
+            });
+        },
+        getAutoValue: function(fieldName, dataModel) {
+            var DAYS_PER_YEAR = 365,
+                AREA_SQM = turfArea(App.currentProject.get('area_of_interest')),
+                M3_PER_MGAL = 3785.41178,
+                CM_PER_M = 100.0;
+
+            return _.sum(dataModel[fieldName]) /
+                (DAYS_PER_YEAR * M3_PER_MGAL * CM_PER_M / AREA_SQM);
+        },
+    }
+};
--- a/src/mmw/js/src/modeling/gwlfe/entry/models.js
+++ b/src/mmw/js/src/modeling/gwlfe/entry/models.js
@@ -1,6 +1,7 @@
 "use strict";
 
-var Backbone = require('../../../../shim/backbone');
+var Backbone = require('../../../../shim/backbone'),
+    GwlfeModificationModel = require('../../models').GwlfeModificationModel;
 
 var EntryFieldModel = Backbone.Model.extend({
     defaults: {
@@ -44,6 +45,31 @@ var EntryTabModel = Backbone.Model.extend({
         name: '',
         sections: null,  // EntrySectionCollection
     },
+
+    getOutput: function() {
+        var output = {},
+            userInput = {};
+
+        this.get('sections').forEach(function(section) {
+            section.get('fields').forEach(function(field) {
+                var name = field.get('name'),
+                    userValue = field.get('userValue');
+
+                if (userValue !== null &&
+                    userValue !== undefined &&
+                    userValue !== '') {
+                    output[name] = field.toOutput(userValue);
+                    userInput[name] = userValue;
+                }
+            });
+        });
+
+        return new GwlfeModificationModel({
+            modKey: 'entry_' + this.get('name'),
+            output: output,
+            userInput: userInput,
+        });
+    }
 });
 
 var EntryTabCollection = Backbone.Collection.extend({
--- a/src/mmw/js/src/modeling/gwlfe/entry/models.js
+++ b/src/mmw/js/src/modeling/gwlfe/entry/models.js
@@ -59,6 +59,30 @@ var WindowModel = Backbone.Model.extend({
     },
 });
 
+/**
+ * Returns a FieldCollection for a section
+ * @param tabName  Name of the tab
+ * @param dataModel  Project's gis_data, cleaned
+ * @param modifications  Scenario's current modification collection
+ * @param fields  Object with {name, label, calculator} for each field, with
+ *                minValue and maxValue optionally specified
+ * @returns A FieldCollection with specified fields
+ */
+function makeFieldCollection(tabName, dataModel, modifications, fields) {
+    var mods = modifications.findWhere({ modKey: 'entry_' + tabName }),
+        userInput = mods ? mods.get('userInput') : {};
+
+    return new EntryFieldCollection(fields.map(function(fieldInfo) {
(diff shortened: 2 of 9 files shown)


src/mmw/js/src/modeling/gwlfe/entry/calcs.js
```diff
@@ -0,0 +1,81 @@
...
+            return dataModel[fieldName][0];
+        },
+    },
+
+    // Takes the userValue and divides into an array of 12 values evenly
+    EqualMonths: {
+        toOutput: function(userValue) {
+            var monthValue = userValue / 12;
+            return [monthValue, monthValue, monthValue,
+                    monthValue, monthValue, monthValue,
```
kellyi: I believe this could be `_.fill(12, (userValue / 12))` (and the `Array12` implementation could be `_.fill(12, userValue)`. 

`_.fill` is a little more succinct, but it may be that the less succinct version's easier to comprehend at a glance.

src/mmw/js/src/modeling/gwlfe/entry/calcs.js
```diff
@@ -0,0 +1,81 @@
...
+    PointSourceDischarge: {
+        toOutput: function(userValue) {
+            var MONTHDAYS = [31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31],
+                AREA_SQM = turfArea(App.currentProject.get('area_of_interest')),
+                M3_PER_MGAL = 3785.41178,
+                CM_PER_M = 100.0;
+
+            return MONTHDAYS.map(function(DAYS_PER_MONTH) {
+                return userValue *
+                    (DAYS_PER_MONTH * M3_PER_MGAL * CM_PER_M / AREA_SQM);
```
kellyi: What does this represent? May be worth making it a named var before multiplying it with `userValue`.

src/mmw/js/src/modeling/gwlfe/entry/calcs.js
```diff
@@ -0,0 +1,81 @@
...
+            });
+        },
+        getAutoValue: function(fieldName, dataModel) {
+            var DAYS_PER_YEAR = 365,
+                AREA_SQM = turfArea(App.currentProject.get('area_of_interest')),
+                M3_PER_MGAL = 3785.41178,
+                CM_PER_M = 100.0;
+
+            return _.sum(dataModel[fieldName]) /
+                (DAYS_PER_YEAR * M3_PER_MGAL * CM_PER_M / AREA_SQM);
```
kellyi: This could also be a named var -- and assigned in the vertical sequence above, since it doesn't seem to depend on anything in the sum op.

src/mmw/js/src/modeling/gwlfe/entry/models.js
```diff
@@ -1,12 +1,73 @@
...
+    },
+
+    initialize: function(attrs, options) {
+        if (options && options.toOutput) {
+            this.toOutput = options.toOutput;
+        }
+    },
+
+    toOutput: function() {
+        throw "Not implemented";
```
kellyi: Is this going to be implemented in the future? I think throwing here is probably fine, but this could also be a `console.warn` if we'd prefer not to have it cause an error.

src/mmw/js/src/modeling/gwlfe/entry/calcs.js
rajadain: We're converting between MGD (million gallons per day) to cubic centimeters per month. I'll think of naming this cleaner, although the current naming is the same as:

https://github.com/WikiWatershed/model-my-watershed/blob/1e5376e5b06160b45cca3ac9aeaadc3bac68ee7e/src/mmw/apps/modeling/mapshed/calcs.py#L389-L390

src/mmw/js/src/modeling/gwlfe/entry/views.js
```diff
@@ -76,6 +112,110 @@ var TabContentsView = Marionette.CollectionView.extend({
...
+    },
+
+    updateUserValue: function() {
+        var value = this.ui.input.val();
+
+        if (value) {
+            this.model.set('userValue', value);
+        } else {
+            this.model.set('userValue', null);
+        }
```
kellyi: I believe this entire if block could also be:

```js
this.model.set('userValue', value || null);
```

src/mmw/js/src/modeling/gwlfe/entry/models.js
rajadain: No, this will _always_ be overridden by the calculator's `toOutput`. If this is firing it means that it was not properly overridden. Not having this method here would've also raised an error at runtime if that were to happen, but I thought this made it clearer that this should over ridden. Maybe I can change the error output to say that explicitly.

src/mmw/js/src/modeling/gwlfe/entry/views.js
```diff
@@ -76,6 +112,110 @@ var TabContentsView = Marionette.CollectionView.extend({
...
+
+var SectionsView = Marionette.CollectionView.extend({
+    className: 'section',
+
+    childView: SectionView,
+});
+
+function showSettingsModal(title, dataModel, modifications, addModification) {
+    var tabs = new models.EntryTabCollection([
+            // { name: 'efficiencies', displayName: 'Efficiencies' },
```
kellyi: Commenting here to make sure this was intentionally left in albeit commented-out.

kellyi: (Same q about the "animals" and "other" sections below -- I assume the answer's the same too).

rajadain: That is correct. These tabs will have their sections and fields filled in #2960, #2961, and #2962.

src/mmw/js/src/modeling/gwlfe/entry/models.js
kellyi: :+1: In that case I think `throw` is fine here since presumably if this ever got called but the calculation kept working the values would be incorrect anyway.

kellyi:
Tried entering some junk values in the inputs and the validation's mostly working with the exception of allowing the letter "e" as an input for whatever reason:

screen shot 2018-09-12 at 5 53 11 pm

kellyi:
Generated GMS files for two scenarios in this AOI:

screen shot 2018-09-13 at 10 00 08 am

555.gms has the following custom input:

screen shot 2018-09-13 at 10 00 44 am

777.gms has this:

screen shot 2018-09-13 at 10 01 20 am

diff is:

```diff
diff --git a/555.gms b/777.gms
index 46a55d6..ae29b76 100644
--- a/555.gms
+++ b/777.gms
@@ -70,18 +70,18 @@ Sediment
 5,1.5,0,0
 2.44,0.38
 2.44,0.38
-46.25,31056.54,1.5225059779215633
-46.25,31056.54,1.3751666897356059
-46.25,31056.54,1.5225059779215633
-46.25,31056.54,1.4733928818595778
-46.25,31056.54,1.5225059779215633
-46.25,31056.54,1.4733928818595778
-46.25,31056.54,1.5225059779215633
-46.25,31056.54,1.5225059779215633
-46.25,31056.54,1.4733928818595778
-46.25,31056.54,1.5225059779215633
-46.25,31056.54,1.4733928818595778
-46.25,31056.54,1.5225059779215633
+64.75,31056.54,1.5225059779215633
+64.75,31056.54,1.3751666897356059
+64.75,31056.54,1.5225059779215633
+64.75,31056.54,1.4733928818595778
+64.75,31056.54,1.5225059779215633
+64.75,31056.54,1.4733928818595778
+64.75,31056.54,1.5225059779215633
+64.75,31056.54,1.5225059779215633
+64.75,31056.54,1.4733928818595778
+64.75,31056.54,1.5225059779215633
+64.75,31056.54,1.4733928818595778
+64.75,31056.54,1.5225059779215633
 1
 8963,0,0,0,0
 8963,0,0,0,0
```

kellyi:
This is working well so far! I made some comments inline include some which are stylistic suggestions.

Do you think the input validation is worth addressing? For what it's worth, in addition to `e` I was also able to enter negative values like -1000 into each of the inputs, which wouldn't make any sense for the models.

src/mmw/js/src/modeling/gwlfe/entry/models.js
```diff
@@ -22,8 +83,42 @@ var WindowModel = Backbone.Model.extend({
     },
 });
 
+/**
+ * Returns a FieldCollection for a section
+ * @param tabName  Name of the tab
+ * @param dataModel  Project's gis_data, cleaned
+ * @param modifications  Scenario's current modification collection
+ * @param fields  Array of [name, label, calculator] for each field
+ */
```
kellyi: Consider specifying `@returns` value in the docstring too.

src/mmw/js/src/modeling/gwlfe/entry/templates/field.html
```diff
@@ -0,0 +1,27 @@
+<div class="col-sm-8">
+    <div class="manual-entry-label">
+        {{ label }}
+    </div>
+</div>
+<div class="col-sm-4">
+    <input
+            type="number"
```
kellyi: This must be what's doing the validation currently. If we aren't going to do additional validation stuff here, it's probably worth making a new enhancement card for custom validations.

src/mmw/js/src/modeling/gwlfe/entry/calcs.js
rajadain: I adjusted the comments above this to be a bit more explicit, but decided to leave the code as is.

kellyi: Sounds good!

rajadain:
Added some validation highlighting:

image

We won't stop you from saving the invalid values right now. That can be done in a later card.

rajadain:
Thanks for the thorough test and all the great suggestions!

Issue #2967, opened 2018-09-13
Update Staging SRAT URL


rajadain:
Change it to https://***.us-east-1.amazonaws.com/dev/SratRunModel_DEV

PR #2971, opened 2018-09-13
Use staging SRAT url for dev, test, staging

rajadain:
## Overview

Updates the Staging SRAT URL so the service providers can experiment with changes without affecting our production instance.

In addition to these places, I've also updated `civicci01` and the fileshare. The production config still points to the production URL.

Connects #2967 

## Testing Instructions

* Check out this branch and `ssh` in to your `app` and `worker` instances and update the `/etc/mmw.d/env/MMW_SRAT_CATCHMENT_API_URL` file with the new URL. Then restart `mmw-app` and `celeryd`.
* Run subbasin on a HUC-10. Ensure it finishes correctly.


kellyi:
screen shot 2018-09-18 at 4 07 09 pm

rajadain:
Thanks for taking a look!

Issue #2972, opened 2018-09-17
Failing CUAHSI Service Brings Down Production


rajadain:
See outage here: https://my.panopta.com/outage/Outage?outage_id=1694183136

From @mmcfarland's investigation [here](https://azavea.slack.com/archives/C044CG212/p1537184869000100):

> * App Servers are failing health checks and cycling since 2:30am, haven't isolated why.  RDS & Redis seem to be ok.
> * I think we might have a check for the cuahsi catalog in the health check for the app
> * Yeah, the http://hiscentral.cuahsi.org/webservices/hiscentral.asmx?WSDL service is 503ing, and our health checks are failing because of that
> * Search happens just by invoking the client, which happens through a series of `imports`.  It's not tested specifically as part of the health check
https://github.com/WikiWatershed/model-my-watershed/blob/de7bdcb77d21fa08a6a9967b62e7009ec2a9d907/src/mmw/apps/bigcz/clients/cuahsi/search.py#L44
> * I'm hotfixing the code on the live servers to get them to pass the checks
> * Ok, think I got them all.  We'll need to figure out a more resilient way to initialize the Client without trying to resolve the URL.  My hack was to just remove the cuashi from the `clients` import.

Make it so that the `client` instance is created inside the view handlers rather than as a common variable. This would still fail the searches themselves, but would not bring down the app.

rajadain:
Also, in case @aufdenkampe @emiliom didn't know already, the CUAHSI service is down.

![image](https://user-images.githubusercontent.com/1430060/45628661-8bcf9a80-ba62-11e8-8437-9f887f5c65e2.png)

http://hiscentral.cuahsi.org/webservices/hiscentral.asmx?WSDL

emiliom:
@rajadain I wasn't aware of this. Thanks for posting here. It looks like they're having a major problem at CUAHSI, as their main user-oriented web application is also down (http://data.cuahsi.org/). I'll see what we can learn.

emiliom:
The CUAHSI services were restored a couple of hours ago.

PR #2973, opened 2018-09-17
Monitor: Tolerate CUAHSI Service Being Unavailable

rajadain:
## Overview

Previously, if the CUAHSI service was unavailable, it would stop MMW from loading at all. This was because the SUDS client that talked to the service was initialized at a top level at app load. Now it is moved to be initialized within a view handler, and is wrapped in a `try` block, and the front-end updated to reflect the error. This insulated MMW from CUAHSI service disruptions.

Connects #2972 

### Notes

Tagging @tnation14 for deployment notes. I've made a `hotfix/1.23.1` branch at the `1.23.0` tag, and am planning to merge this PR to that branch. Once it has been merged, we should tag that with `1.23.1`, deploy it to production, and merge to `master`.

We shouldn't need to take a database backup as part of this deployment, as it is error handling related change only.

## Testing Instructions

    * If you see some `parser` errors when loading the app, that is because `master` and `develop` have different version of `gwlf-e`. Running `vagrant reload app --provision` should fix those.
* Select a shape and click "Monitor". Search for something, like "water". Ensure HydroShare and CINERGI return results. Ensure CUAHSI says "service not available".

### DIFF
--- a/src/mmw/apps/bigcz/clients/cuahsi/search.py
+++ b/src/mmw/apps/bigcz/clients/cuahsi/search.py
@@ -10,6 +10,7 @@ from operator import attrgetter, itemgetter
 
 from suds.client import Client
 from suds.sudsobject import asdict
+from suds.transport.http import TransportError
 from rest_framework.exceptions import ValidationError
 from django.contrib.gis.geos import Point
 
@@ -17,7 +18,11 @@ from django.core.cache import cache
 from django.conf import settings
 
 from apps.bigcz.models import ResourceLink, ResourceList, BBox
-from apps.bigcz.utils import parse_date, RequestTimedOutError
+from apps.bigcz.utils import (
+    parse_date,
+    RequestTimedOutError,
+    ServiceNotAvailableError,
+)
 
 from apps.bigcz.clients.cuahsi.models import CuahsiResource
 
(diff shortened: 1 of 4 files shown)


src/mmw/apps/bigcz/clients/cuahsi/search.py
```diff
@@ -235,6 +238,12 @@ def make_request(request, expiry, **kwargs):
 
 
 def get_services_in_box(box):
+    try:
+        client = Client(CATALOG_URL, timeout=settings.BIGCZ_CLIENT_TIMEOUT)
```
mmcfarland: My recollection is that the `Client` initialization is a somewhat expensive process (which makes sense now that we see it is making HTTP requests).  Would it make sense to keep the variable scoped to the module, but initialize it conditionally inside the view function?  I'm not sure how much of an impact this has on a search request response time.

rajadain:
Good point. I pushed a fixup that initializes a global instance if it hasn't already been initialized, sort of like a singleton factory pattern #enterprise. Note that it is harder to test now that the service is back up, but mistyping the catalog URL works:

https://github.com/WikiWatershed/model-my-watershed/blob/88fa6e1e737e7ad55ae97298b8756272bb085d7b/src/mmw/apps/bigcz/clients/cuahsi/search.py#L27

rajadain:
Thanks for reviewing! Squashed the fixup, merging now.

Issue #2953, opened 2018-09-04
Protected Lands: Add to Layer Picker


rajadain:
Once the tiles have been ingested and uploaded to S3 in #2952, add an entry and corresponding legend to the Layer Picker in the Coverages tab.

Issue #2952, opened 2018-09-04
Protected Lands: Paint Layer


rajadain:
Once the layer has been ingested in #2951, using the colors picked in #2950 generated painted tiles for the layer, matching the data type / resolution / etc used in other layers such as NLCD. Verify the painted tiles, and upload to S3.

Issue #2951, opened 2018-09-04
Protected Lands: Reclassify and Ingest


rajadain:
Once we have the colors picked in #2950, ingest the Protected Lands data into RDDs, using the following table of values for each classification type:

| Class                             | Color          | raster value |
|-----------------------------------|----------------|--------------|
| Park or Recreation Area - Federal | Green/Yellow 1 | 1            |
| Park or Recreation Area - State   | Green/Yellow 2 | 2            |
| Park or Recreation Area - Local   | Green/Yellow 3 | 3            |
| Park or Recreation Area - Private | Green/Yellow 4 | 4            |
| Park or Recreation Area - Unknown | Green/Yellow 5 | 5            |
| Natural Resource Area - Federal   | Green Dark 1   | 6            |
| Natural Resource Area - State     | Green Dark 2   | 7            |
| Natural Resource Area - Local     | Green Dark 3   | 8            |
| Natural Resource Area - Private   | Green Dark 4   | 9            |
| Natural Resource Area - Unknown   | Green Dark 5   | 10           |
| Conservation Easement             | Brown/Green    | 11           |
| Agricultural Easement             | Brown 1        | 12           |

and using the `reclass2` column in [this spreadsheet](https://docs.google.com/spreadsheets/d/1dSQnohrYreug8eaBwsrtYoC4U_cV9YVtL1gg31WTjCg/edit#gid=0) to convert input values to reclassified values. Ingest the raster into RDDs and upload to S3.

rajadain:
The input raster has already been reclassified, so we don't need to do that.

<details>
<summary>Here is the `gdalinfo` output:

```
$ gdalinfo -mm -hist PADUS1.tif
```
</summary>

```
Driver: GTiff/GeoTIFF
Files: PADUS1.tif
       PADUS1.tif.ovr
       PADUS1.tif.aux.xml
Size is 161190, 104424
Coordinate System is:
PROJCS["Albers_Conical_Equal_Area",
    GEOGCS["NAD83",
        DATUM["North_American_Datum_1983",
            SPHEROID["GRS 1980",6378137,298.2572221010042,
                AUTHORITY["EPSG","7019"]],
            AUTHORITY["EPSG","6269"]],
        PRIMEM["Greenwich",0],
        UNIT["degree",0.0174532925199433],
        AUTHORITY["EPSG","4269"]],
    PROJECTION["Albers_Conic_Equal_Area"],
[...]

PR #2975, opened 2018-09-18
Add protected lands overlay to layer selector

kellyi:
## Overview

This PR adds the protected-lands-30m overlay option to the layer selector along with a legend.

Connects #2951 
Connects #2952 
Connects #2953 

### Demo

screen shot 2018-09-18 at 11 10 04 am

## Testing Instructions

- get this branch, `bundle.sh --vendor`, then visit localhost:8000
- select the "Protected Lands" overlay from the layer selector and verify that
    - the overlay is visible
    - the opacity slider works
    - the legend matches the color map here https://github.com/WikiWatershed/model-my-watershed/issues/2951#issuecomment-421442719


src/mmw/mmw/settings/layer_settings.py
```diff
@@ -198,6 +198,32 @@
             'legend_unit_breaks': [-20, '', -8, '', 4, '', 16, '', 28, '', 40],
             'big_cz': True,
         },
+        {
+            'code': 'protected-lands-30m',
+            'display': 'Protected Lands',
+            'short_display': 'Protected Lands',
+            'css_class_prefix': 'protected-lands-30m',
+            'help_text': '',
```
kellyi: Do we know what should be the help text for this overlay?

rajadain: We don't. The help text shows up for "related layers" in the Analyze tabs. Since this isn't currently a related layer to any tab, we don't have to worry about this field.

kellyi:
> Since this isn't currently a related layer to any tab, we don't have to worry about this field.

Makes sense -- I'm going to remove the `helpText` key from the layer then.

kellyi:

Issue #2882, opened 2018-07-09
Make the Download GMS Button Always Visible


ajrobbins:
Currently, once the user proceeds to add modifications to an area with the Multi Year model, their ability to download a gms file for current conditions goes away:

![screen shot 2018-07-09 at 12 17 36 pm](https://user-images.githubusercontent.com/12868342/42463026-f911678a-8372-11e8-8012-633c88899fed.png)

Add a similar "..." control next to Current Conditions in the scenario dropdown with only the "export GMS" option in it, and enable GMS download.

rajadain:
Instead of the above, make it so the "Download GMS" button is always available in the top bar, as it is for Current Conditions before new scenarios are added. Will have to see how that changes the positioning of the "Add New Scenario" button. Discuss with @jfrankl.

PR #2980, opened 2018-09-21
Add GMS Download button to Scenario Buttons View

rajadain:
## Overview

This view classically does not correspond to a single scenario, but this is where they want the GMS download button to be, so we add it, always querying for the active scenario and then populating its `gis_data` in the download form.

Much of the code has been copied from `ScenarioDropDownMenuOptionsView`, which still has an Export GMS button. A GMS file can be exported from both places.

A third place is in `ScenarioToolbarView`, which is more specific to the scenario, but that comes after the Scenario Buttons of Add New Scenario and Compare, and we want this to be seen before, so it matches to when there's only Current Conditions.

Connects #2882 

### Demo

Current Conditions with no other scenarios (same as before):

image

Current Conditions with other scenarios (now with Export GMS button added):

image

Other scenario (now with Export GMS button added):

image

Tagging @jfrankl 	for visual review.

### Notes

Instead of adding the GMS form to a third place, I tried changing `ScenarioToolbarView` which already has it when Current Conditions is the only scenario. I didn't end up sticking to that because:

1. It used flexbox, and as such the button styling was quite different, and would have had to be brought in line. It was simpler to reuse existing styling to match the other buttons.
2. It would place the "Export GMS" button _after_ the "New Scenario" and "Compare" buttons. While that may make sense from one perspective (the buttons get more scenario-specific as you go to the right), the movement of the button from its place when only current conditions exist to when other scenarios are available felt confusing.

Also note that the hover state for the only current conditions case is like a link:

image

Whereas for other cases it is a teal button:

image

Should this be made consistent?

## Testing Instructions

* Make a new MapShed project. Ensure you can export GMS of current conditions when there are no more scenarios.
* Add a scenario. Ensure you can export its GMS. Ensure it is identical to Current Conditions (since you have not made any changes to it).
* Make some changes. Export GMS again. Ensure it is no longer identical to Current Conditions.
* Export GMS from the Scenario List dropdown and the header button. Ensure that the generated files are identical.

### DIFF
--- a/src/mmw/js/src/modeling/templates/scenariosBarButtons.html
+++ b/src/mmw/js/src/modeling/templates/scenariosBarButtons.html
@@ -1,6 +1,10 @@
 {% if editable and not isOnlyCurrentConditions %}
     <button class="btn btn-sm btn-default inverted btn-scenario" id="add-scenario"><i class="fa fa-plus"></i>New scenario</button>
 {% endif %}
+{% if isGwlfe and activeScenario and not isOnlyCurrentConditions %}
+    <button class="btn btn-sm btn-default inverted btn-scenario" id="export-gms"><i class="fa fa-download"></i>Export GMS</button>
+    {% include './gmsDownloadForm.html' %}
+{% endif %}
 {% if showCompare %}
     <button class="btn btn-sm btn-default inverted btn-scenario" id="show-compare"><i class="fa fa-th"></i> Compare</button>
 {% endif %}
--- a/src/mmw/js/src/modeling/views.js
+++ b/src/mmw/js/src/modeling/views.js
@@ -320,11 +320,14 @@ var ScenarioButtonsView = Marionette.ItemView.extend({
     },
 
     ui: {
+        exportGms: '#export-gms',
+        exportGmsForm: '#export-gms-form',
         addScenario: '#add-scenario',
         showCompare: '#show-compare',
     },
 
     events: {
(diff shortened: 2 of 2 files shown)


kellyi:
> Also note that the hover state for the only current conditions case is like a link: ... Should this be made consistent?

If it's not too big an undertaking I'm in favor of giving these a consistent style.

kellyi:
I'm not sure what's causing this but the Export GMS button from the main header bar does not seem to export the same data as the Export GMS button located in the scenario tab. To replicate this:

1. Create a new scenario and adjust the animals numbers in that tab
2. export a gms file for that scenario using the header bar
3. export a gms file for that scenario using the scenarios tab option
4. git diff to see the difference

Not sure whether this applies to all modifications or just animals, but it's probably worth checking the others too.

rajadain:
Hmm, taking a look at that now.

rajadain:
Great catch. Since we were populating the hidden `gis_data` field in the form on render, the header button would have the value from when the scenario was initially loaded. Since the dropdown menu was rendered every time, it would always have the most up to date values, but not the header. By forcing the header button set to re-render just before downloading the GMS file we ensure that the most recent changes are reflected in the download. This is ready for another look.

jfrankl:
Looks good. One edit: I'd like to move the "New scenario" button to the left, so it's next to the scenario dropdown.

rajadain:
> move the "New scenario" button to the left, so it's next to the scenario dropdown.

Pushed.

image

PR #2986, opened 2018-09-27
MapShed Manual Entry: BMP Efficiencies

rajadain:
## Overview

Adds a tab for editing BMP efficiencies. Most of them are simple Direct replacements. A few are hard-coded Static values that are [out of scope](https://github.com/WikiWatershed/model-my-watershed/issues/2962#issuecomment-424730883) to manage.

Includes a new UI for handling triplet fields.

Connects #2962 

### Demo

image

image

## Testing Instructions

* Open a MapShed project and add a scenario
* Edit the BMP efficiencies. Ensure saving and undo work as they do for other tabs.
* Change a few fields and export GMS and compare with Current Conditions. Ensure the right fields diff.

### DIFF
--- a/src/mmw/js/src/modeling/gwlfe/entry/templates/field.html
+++ b/src/mmw/js/src/modeling/gwlfe/entry/templates/field.html
@@ -14,6 +14,7 @@
             {% if maxValue != null %} max="{{ maxValue }}" {% endif %}
             placeholder="{{ autoValue | round(3) }}"
             value="{{ userValue }}"
+            {% if readOnly %} disabled {% endif %}
     >
     <button
             type="button"
--- a/src/mmw/js/src/modeling/gwlfe/entry/templates/field.html
+++ b/src/mmw/js/src/modeling/gwlfe/entry/templates/field.html
@@ -1,5 +1,5 @@
 <div class="col-sm-8">
-    <div class="manual-entry-label">
+    <div class="mapshed-manual-entry-label">
         {{ label }}
     </div>
 </div>
--- a/src/mmw/js/src/modeling/gwlfe/entry/templates/field.html
+++ b/src/mmw/js/src/modeling/gwlfe/entry/templates/field.html
@@ -1,10 +1,4 @@
-<div class="col-sm-8">
-    <div class="mapshed-manual-entry-label">
-        {{ label }}
-    </div>
-</div>
 {% if type == 'NUMERIC' %}
-<div class="col-sm-4">
     <input
             type="number"
             class="form-control"
@@ -27,9 +21,7 @@
     >
         <i class="fa fa-undo"></i>
     </button>
-</div>
 {% elif type == 'YESNO' %}
-<div class="col-sm-4">
     <select
             class="form-control {{ '-unmodified' if not userValue }}"
             name="{{ name }}"
@@ -62,9 +54,6 @@
     >
         <i class="fa fa-undo"></i>
     </button>
-</div>
 {% else %}
-<div class="col-sm-4">
     Missing Field Type
-</div>
 {% endif %}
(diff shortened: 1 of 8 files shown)


src/mmw/js/src/modeling/gwlfe/entry/templates/field.html
```diff
@@ -14,6 +8,7 @@
             {% if maxValue != null %} max="{{ maxValue }}" {% endif %}
             placeholder="{{ autoValue | round(3) }}"
             value="{{ userValue }}"
+            {% if readOnly %} disabled {% endif %}
```
kellyi: Should these disabled inputs have a title/popover that indicates that they're readonly and why?

rajadain: They are quite evidently not editable, not sure if a popup indicating they are "read only" would add anything. As to why, that's due to them being "virtual BMPs" and not true ones, i.e. those that differ from a true BMP only in terms of efficiency, so with the efficiency changed they mean nothing.

This is a pretty deep part of the app where we expect only expert users. I don't think we need to add any explanation here, but I'll re-evaluate this if the clients ask for that explicitly.

kellyi:
For this AOI:

screen shot 2018-09-28 at 3 43 28 pm

I adjusted the efficiency values from

screen shot 2018-09-28 at 3 44 46 pm

to

screen shot 2018-09-28 at 3 44 25 pm

And the resulting gms file diff is

```
diff --git a/unadjusted_efficiencies.gms b/efficiencies.gms
index 4da17b2..5fd230a 100644
--- a/unadjusted_efficiencies.gms
+++ b/efficiencies.gms
@@ -111,10 +111,10 @@ Sediment
 0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0
 0,0,0,0,9.698808972503617,51.133176555716354,0,0,0,0,0,0,0,0,0,0
 0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0
-0.29,0.41,0.08,0.66,0.05,0,0.95,0.3,0.56,0.2,0.95,0.29,0.25
-0.5,0.28,0.4,0.22,0.1,0.1,0,0.95,0.3,0.78,0.45,0.95,0.44,0.35
-0.35,0.44,0.63,0.53,0.3,0.17,0.16,0,0.95,0.38,0.76,0.6,0.55,0.95,0.02,0.0035,2.55
-0.75,0.75,0.14,0.14,0.15,0.15,0.21,0.7,1,0.85,0.14,0.15,0.71,0.82,0.71
+1,1,0.08,0.66,0.05,0,0.95,0.3,1,0.2,1,1,1
+1,1,1,0.22,0.1,0.1,0,0.95,0.3,1,0.45,1,1,1
+1,1,1,1,0.3,0.17,0.16,0,0.95,0.38,1,0.6,1,1,0.02,0.0035,2.55
+1,1,1,1,0.15,0.15,0.21,0.7,1,0.85,0.14,0.15,0.71,0.82,0.71
 0,0,0,0,0,0,0,0,0,0,0,0,0,0
 0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0
 0,0,0,0,0,0,0,0,0,0,0,0,0,None
```

kellyi:
This probably falls under an enhancement if it's even something we'd want to consider, but I wonder if it'd make sense to have a "Reset" button on these modals now that some of them have 20ish inputs:

screen shot 2018-09-28 at 3 44 25 pm

Conceivably you could reset all the values just by making another new scenario, but I could imagine a case in which a user might want to reset values on one of the tabs but keep the others.

rajadain:
re: the reset button, that's a good idea, but I don't know how warranted it is. I think we should wait and see if the clients bring it up, and make a card for it then.

rajadain:
Thanks for taking a look! Merging this now.

Issue #2983, opened 2018-09-25
Three Equal Stages: Workflow Redesign


rajadain:
Mockups: https://share.goabstract.com/2de55961-8d6e-4232-aa28-99f2564960e5

The current workflow is a remnant of a time when there was only one model, and there was a clear directed flow from Draw to Analyze to Model. Over time, we have added more models and more modes (like Monitor), and the current workflow is less linear. As these features were added over time to the existing workflow, it has stretched and reshaped, but has never been properly refactored.

This card is to record conversations with @jfrankl around what a potential other workflow could look like.

Issues with the current workflow are:

 - Analyze, Monitor, and Model aren't treated as equals
 - There is no "Project" for Analyze or Monitor, but selecting a Model automatically creates one. This can be unexpected and confusing.
 - There is no chance to review the shape before moving to Analyze. This is especially problematic when a shape is drawn computationally, such as with RWD or the 1 Sq Km tool.

In the proposed workflow, the users would go from the Draw stage:

![image](https://user-images.githubusercontent.com/1430060/46042638-e6f93100-c0e3-11e8-83d5-419956e07bdf.png)

to a Confirm stage:

![image](https://user-images.githubusercontent.com/1430060/46042649-eeb8d580-c0e3-11e8-8cdf-168eaef3ffca.png)

where they can see the shape and go back and change it cheaply, without incurring server resources. Once they hit confirm, we bring them to Analyze in "Project Mode":

![image](https://user-images.githubusercontent.com/1430060/46042934-a4842400-c0e4-11e8-802c-0b6594984952.png)

a model won't be created until the user selects it, similar to #2978:

![image](https://user-images.githubusercontent.com/1430060/46043081-0775bb00-c0e5-11e8-8e86-807fa7095716.png)

upon which we return to a familiar project screen:

![image](https://user-images.githubusercontent.com/1430060/46043142-2e33f180-c0e5-11e8-830c-92e3a975ded0.png)

There will be challenges around how saving works. Some initial brainstorming led to the following ideas:

- Allow projects to be created without a `model_package`
  - This could allow users to save projects at the Analyze or Monitor stage
  - But what would they be "saving" really? None of the information in those tabs is currently persisted to the database.
  - This would also require updating the many places in the front-end which currently always assume a `model_package` will _always_ exist in a project
- Make saving manual, so we don't automatically save when a model is created, but prompt a user to if they try to navigate away from the page
[...]

Issue #2978, opened 2018-09-20
Three Equal Stages: Move Model alongside Analyze, Monitor


rajadain:
We now have wireframes for this effort here: https://share.goabstract.com/12b172b2-4f9a-44ed-a2cf-50fa0dec92c1

![image](https://user-images.githubusercontent.com/1430060/46041138-1f970b80-c0e0-11e8-9692-85d6e7447e68.png)

![image](https://user-images.githubusercontent.com/1430060/46041239-4e14e680-c0e0-11e8-8c9c-61c565854344.png)

- [x] Remove the "Model" button from the footer of Analyze / Monitor, let "Change Area" take the full space
- [x] Add a new tab "Model" next to "Analyze" and "Monitor"
- [x] Inside the new tab, add the same description and buttons for selecting a model and creating a project as previously in the popup when clicking "Model":

    ![image](https://user-images.githubusercontent.com/1430060/46041356-8f0cfb00-c0e0-11e8-82d7-3e7f26d50aaf.png)

- [x] Clicking the buttons should do the same thing it does now
- [x] There should be no change to routing. Opening existing projects should work as before.

rajadain:
We now have wireframes for this effort here: https://share.goabstract.com/12b172b2-4f9a-44ed-a2cf-50fa0dec92c1

![image](https://user-images.githubusercontent.com/1430060/46041138-1f970b80-c0e0-11e8-9692-85d6e7447e68.png)

![image](https://user-images.githubusercontent.com/1430060/46041239-4e14e680-c0e0-11e8-8c9c-61c565854344.png)

- [ ] Remove the "Model" button from the footer of Analyze / Monitor, let "Change Area" take the full space
- [ ] Add a new tab "Model" next to "Analyze" and "Monitor"
- [ ] Inside the new tab, add the same description and buttons for selecting a model and creating a project as previously in the popup when clicking "Model":

    ![image](https://user-images.githubusercontent.com/1430060/46041356-8f0cfb00-c0e0-11e8-82d7-3e7f26d50aaf.png)

- [ ] Clicking the buttons should do the same thing it does now
- [ ] There should be no change to routing. Opening existing projects should work as before.

PR #2987, opened 2018-10-01
Three Equal Stages: Move Model alongside Analyze, Monitor

rajadain:
## Overview

Moves "Model" button and action to be another tab, like "Analyze" and "Monitor", to bring it on equal footing with the other two. The routing is not changed, and the behavior of the buttons is identical to before.

Connects #2978 

### Demo

screen shot 2018-10-01 at 1 15 48 pm

### Notes

A larger workflow revision is suggested in #2983, which may be attempted at a later date.

## Testing Instructions

* Open the site, select shape and proceed to Analyze
* Ensure there is no Model button in the sidebar footer
* Ensure the "Change Area" button takes full width, and clicking it works as before
* Go back to Analyze stage and select the Model tab. Ensure the text matches what was there before.
* Test creating projects using both models. Ensure they work same as before.
* Re-open saved projects. Ensure they open correctly.


kellyi:
Tested this out and it's working well! Going to check out the code changes now.

rajadain:
Thanks for taking a look!

Issue #2989, opened 2018-10-02
Manual Entry: Limit Land Cover Numbers to 1 Decimal Place


rajadain:
Currently the Land Cover modal shows area in hectares rounded up to 3 decimal places:

![image](https://user-images.githubusercontent.com/1430060/46375534-b1af8e80-c661-11e8-8081-2f2ed6cadd03.png)

Make it round up to 1.

PR #2992, opened 2018-10-09
Manual Entry: Round Land Cover values to 1 decimal place

rajadain:
## Overview

This PR makes it so that the Land Cover modal shows only 1 decimal place for values. MapShed is not designed to be run for any higher precision, and the extra decimal places gave the incorrect impression that it was.

Makes `FieldsView` parametrizable since it is used by multiple modals, most of which want 3 decimal places, with the Land Cover modal the only exception which prefers 1.

Connects #2989 

### Demo

image

## Testing Instructions

* Create / Open a MapShed project
* Add a scenario. Open the Land Cover modal. Ensure that all the auto-populated values are limited to 1 decimal place.
* Ensure as you type, the total and diff values are always limited to 1 decimal place.
* Ensure you can type in a higher decimal place value, but the total calculation is limited to 1 decimal place.
* Ensure that other settings modals still show higher decimal places when the auto value has it

### DIFF
--- a/src/mmw/js/src/modeling/gwlfe/entry/models.js
+++ b/src/mmw/js/src/modeling/gwlfe/entry/models.js
@@ -14,6 +14,7 @@ var EntryFieldModel = Backbone.Model.extend({
         label: '',
         name: '',
         type: ENTRY_FIELD_TYPES.NUMERIC,
+        decimalPlaces: 3,
         minValue: null,
         maxValue: null,
         autoValue: null,
--- a/src/mmw/js/src/modeling/gwlfe/entry/templates/field.html
+++ b/src/mmw/js/src/modeling/gwlfe/entry/templates/field.html
@@ -3,10 +3,10 @@
             type="number"
             class="form-control"
             name="{{ name }}"
-            step="0.000001"
+            step="{{ step }}"
             {% if minValue != null %} min="{{ minValue }}" {% endif %}
             {% if maxValue != null %} max="{{ maxValue }}" {% endif %}
-            placeholder="{{ autoValue | round(3) }}"
+            placeholder="{{ autoValue }}"
             value="{{ userValue }}"
             {% if readOnly %} disabled {% endif %}
     >
@@ -17,7 +17,7 @@
             data-placement="bottom"
             data-trigger="hover"
             data-html="true"
-            data-content="<strong>Reset to auto-estimated:</strong><br />{{ autoValue | round(3) }}"
+            data-content="<strong>Reset to auto-estimated:</strong><br />{{ autoValue }}"
     >
         <i class="fa fa-undo"></i>
     </button>
--- a/src/mmw/js/src/modeling/gwlfe/entry/templates/landCoverTotal.html
+++ b/src/mmw/js/src/modeling/gwlfe/entry/templates/landCoverTotal.html
@@ -1,13 +1,13 @@
 {% if is_modified %}
-    Total: <strong>{{ userTotal | toLocaleString(3) }} ha</strong>
+    Total: <strong>{{ userTotal | toLocaleString(1) }} ha</strong>
     {% if (userTotal) != (autoTotal) %}
     <small>
-        (<span class="error">{{ (userTotal - autoTotal) | abs | toLocaleString(3) }}
+        (<span class="error">{{ (userTotal - autoTotal) | abs | toLocaleString(1) }}
             {{ 'less' if userTotal < autoTotal else 'more' }} than
         </span>
-        required {{ autoTotal | toLocaleString(3) }})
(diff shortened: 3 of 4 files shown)


kellyi:
Taking a look at this!

kellyi:
> Ensure you can type in a higher decimal place value, but the total calculation is limited to 1 decimal place.

Would it make sense to disallow typing a higher number of decimal places? Right now I can enter this:

screen shot 2018-10-11 at 12 55 05 pm

Which gives me the impression that we're dealing with that level of precision but in practice tosses out everything after the tenths place.

kellyi:
At the very least I wonder if it makes sense to have the number be rounded visibly for the user when the input field with a long decimal loses focus -- say when the user tabs into another field.

kellyi:
Another small thing I discovered: you can cause an error in the modeling state by using a float value for a number that the model expects to be integral, like the numeric inputs on the Animals tab. The 1.5 cows here causes the model not to run, although I can set that back to an integer and subsequent runs of the model will work:

screen shot 2018-10-11 at 12 59 13 pm

I think it makes sense to have fields which have to be integer values be rounded to integers -- or at least to indicate what's wrong with them.

If that's out of scope for this issue, I think it's worth making another card for that.

kellyi:
Aside from the visual quirks I noted above this is working well so far!

rajadain:
I think letting users type in higher values if they want is alright ... validating against it may prove more complicated than the budget allows. The higher decimal places are not lost, but rounded, so they have some effect. I think we can expect users to take a hint from the other values and limit their expectations of precision.

The model not running for float inputs for Animal values is a good catch. I'll look into that.

rajadain:
Alright, went with a significantly different approach to fix this. This allows specifying allowed decimal places in the field configuration, so we can say that land cover takes 1, and animals take 0, with the rest defaulting to 3:

image

rajadain:
Additionally, higher precision user entries are now also highlighted as being problematic:

image

kellyi:
I like the validation changes; however it's still letting me save values that are longer than the specified number of digits:

screen shot 2018-10-11 at 5 05 07 pm

screen shot 2018-10-11 at 5 06 06 pm

It seems like we should either disable the save button when the fields are invalid OR automatically round the input to make it valid, so that it shows up as valid after saving and reopening the modal.

rajadain:
I think that's a good idea, and I have a rough idea of how to implement it:

```diff
diff --git a/src/mmw/js/src/modeling/gwlfe/entry/views.js b/src/mmw/js/src/modeling/gwlfe/entry/views.js
index 16e66535..4fa61b42 100644
--- a/src/mmw/js/src/modeling/gwlfe/entry/views.js
+++ b/src/mmw/js/src/modeling/gwlfe/entry/views.js
@@ -77,9 +77,11 @@ var LandCoverModal = modalViews.ModalBaseView.extend({
 
     validateModal: function() {
         var autoTotal = round(this.model.get('autoTotal'), 1),
-            userTotal = round(this.model.get('userTotal'), 1);
+            userTotal = round(this.model.get('userTotal'), 1),
+            hasInvalidFields = this.$el.find('input:invalid').length > 0;
 
-        this.ui.saveButton.prop('disabled', autoTotal !== userTotal);
+        this.ui.saveButton.prop('disabled', hasInvalidFields ||
+                                            autoTotal !== userTotal);
     },
 
     saveAndClose: function() {
@@ -132,6 +134,15 @@ var EntryModal = modalViews.ModalBaseView.extend({
 
     initialize: function(options) {
         this.mergeOptions(options, ['addModification']);
+
+        var listenTo = this.listenTo,
+            validateModal = _.bind(this.validateModal, this);
+
+        this.model.get('tabs').forEach(function(tab) {
+            tab.get('sections').forEach(function(section) {
+                listenTo(section.get('fields'), 'change', validateModal);
+            });
+        });
     },
 
     // Override to populate tabs
@@ -148,6 +159,12 @@ var EntryModal = modalViews.ModalBaseView.extend({
         this.$el.modal('show');
     },
 
+    validateModal: function() {
+        var hasInvalidFields = this.$el.find('input:invalid').length > 0;
+
+        this.ui.saveButton.prop('disabled', hasInvalidFields);
+    },
+
     saveAndClose: function() {
         var addModification = this.addModification;
 
```

but there's a number of edge cases (such as reverting a value), which will need to be thought out and addressed. I think that is out of scope of this card.

rajadain:
I'll make a separate card for the above.

kellyi:
Sounds good. Since we're spinning off strengthening the validation to another card, this is ready to go.

rajadain:
Made #2994 for disabling save for invalid fields. Marked it as an enhancement. Going to squash and merge this now. Thanks for the feedback!

Issue #2988, opened 2018-10-02
Protected Lands: Make Legend Clearer


rajadain:
In the Protected Lands layer, both Agricultural Easements `#8D6E63`  and Conservation Easements `#827717` are brown-ish. They are easily distinguished on the map, but not in the legend:

![image](https://user-images.githubusercontent.com/1430060/46375129-985a1280-c660-11e8-80cb-9c5f4cd4e542.png)

Make them easier to distinguish in the legend.

rajadain:
How does this look:

![image](https://user-images.githubusercontent.com/1430060/46829844-f24b8e00-cd6c-11e8-891a-ad163a00b1fa.png)

The Conservation Easements are west of West Chester and north of Wallace.

PR #2993, opened 2018-10-11
Make Protected Lands legend more distinguishable

rajadain:
## Overview

Adjusts legend color of Conservation Easement to make it easier to distinguish from Agricultural Easement, while still matching the tiles.

Connects #2988 

### Demo

image

In the above, the Conservation Easements are west of West Chester and north of Wallace.

Tagging @aufdenkampe 	for visual review.

## Testing Instructions

* Go to [:8000/](http://localhost:8000/) and turn on the Protected Lands layer. Open the legend. Ensure the Conservation Easement color is distinguishable from Agricultural Easement, and still matches those cases in the map.

### DIFF
--- a/src/mmw/mmw/settings/layer_settings.py
+++ b/src/mmw/mmw/settings/layer_settings.py
@@ -202,7 +202,7 @@ LAYER_GROUPS = {
             'code': 'protected-lands-30m',
             'display': 'Protected Lands',
             'short_display': 'Protected Lands',
-            'css_class_prefix': 'protected-lands-30m',
+            'css_class_prefix': 'protected-lands-30m ',
             'help_text': '',
             'url': 'https://{s}.tiles.azavea.com/protected-lands-30m/{z}/{x}/{y}.png',
             'maxNativeZoom': 13,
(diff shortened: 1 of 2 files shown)


src/mmw/mmw/settings/layer_settings.py
```diff
@@ -202,7 +202,7 @@
             'code': 'protected-lands-30m',
             'display': 'Protected Lands',
             'short_display': 'Protected Lands',
-            'css_class_prefix': 'protected-lands-30m',
+            'css_class_prefix': 'protected-lands-30m ',
```
kellyi: Just checking to make sure this space was intentional. I assume it's changed to match the CSS rule changes?

rajadain: Yes. This space allows us to have `protected-lands-30m` as a separate class.

kellyi:
Tried this out and it looks good to me; the colors are visually more distinct now that the swatches are larger:

screen shot 2018-10-11 at 5 14 29 pm

aufdenkampe:
Looks good to me. I think it is a big improvement. Let's move forward.

rajadain:
Thanks for reviewing! Merging this now.

Issue #2995, opened 2018-10-15
Manual Entry: Decimal Places not allowed in BMP Efficiencies


rajadain:
![image](https://user-images.githubusercontent.com/1430060/46961572-0c3be800-d06f-11e8-949b-d0569c198e96.png)

rajadain:
It's because the `step` parameter isn't being filled correctly, and defaults to `1`:

![image](https://user-images.githubusercontent.com/1430060/46961637-2f669780-d06f-11e8-96d9-a0af3bffc8d7.png)

PR #2996, opened 2018-10-15
Manual Entry: Fix Decimal Place Entry in BMP Efficiencies

rajadain:
## Overview

Previously the decimal place limitation was only applied to field with labels. This excluded the fields in the BMP Efficiencies modal, which because of their tight arrangement do not have labels. This was a mistake, since they need to follow the same decimal place behavior as others.

Connects #2995 

### Demo

image

## Testing Instructions

* Ensure each manual entry form follows the appropriate decimal place limitations:
  - Land Cover: 1
  - Animals: 0
  - Everything Else: 3

### DIFF
--- a/src/mmw/js/src/modeling/gwlfe/entry/views.js
+++ b/src/mmw/js/src/modeling/gwlfe/entry/views.js
@@ -682,24 +682,28 @@ function showSettingsModal(title, dataModel, modifications, addModification) {
                                 name: 'NumNormalSys',
                                 label: 'Normally Functioning Systems',
                                 calculator: calcs.Array12,
+                                decimalPlaces: 0,
                                 minValue: 0
                             },
                             {
                                 name: 'NumPondSys',
                                 label: 'Surface Failures',
                                 calculator: calcs.Array12,
+                                decimalPlaces: 0,
                                 minValue: 0
                             },
                             {
                                 name: 'NumShortSys',
                                 label: 'Subsurface Failures',
                                 calculator: calcs.Array12,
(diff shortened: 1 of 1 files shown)


src/mmw/js/src/modeling/gwlfe/entry/views.js
```diff
@@ -277,6 +277,28 @@ var FieldView = Marionette.ItemView.extend({
...
+        }
+
+        var autoValue = this.model.get('autoValue'),
+            decimalPlaces = this.model.get('decimalPlaces'),
+            // How many decimal places are allowed for the field in the UI
+            //   decimalPlaces = 0 => step = 1
+            //   decimalPlaces = 1 => step = 0.1
+            //   decimalPlaces = 3 => step = 0.001
+            // and so on.
+            step = Math.pow(10, -decimalPlaces);
```
kellyi: :+1:

kellyi:
Testing this out now!

kellyi:
Fix is working well! However, I noticed one additional set of fields that seem like they have to have integer values in order for the model to run: the "Number of Persons..." fields in the "Waste Water" tab:

screen shot 2018-10-15 at 12 55 33 pm

Assuming the fix is now pretty trivial to do, seems like it makes sense to fix that here too.

rajadain:
Done in 349318c

image

kellyi:
screen shot 2018-10-15 at 1 54 23 pm

rajadain:
Thanks for taking a look!

Issue #2999, opened 2018-10-17
HydroShare Copying: Generate Project Snapshot


rajadain:
In order to support importing projects from HydroShare resources that are not in the mapping table, we need to create and send a project snapshot which can be queried at import to find its provenance, as described in https://github.com/WikiWatershed/model-my-watershed/issues/2990#issuecomment-430025273.

This snapshot should be a complete JSON representation of a project, including all scenarios, all modifications, and all results.

- [x] Add a function to the project Backbone model to generate this snapshot
- [x] Add the snapshot to the list of files exported to HydroShare

PR #3004, opened 2018-10-18
HydroShare Copying: Generate project Snapshot and export to HydroShare

rajadain:
## Overview

Adds a method to the project model to generate a JSON snapshot of the entire model state, including a `snapshot_at` key with a timestamp of when the snapshot was made. This is included in the HydroShare Export as `mmw_project_snapshot.json`, to be read in at a later stage.

This snapshot is a complete representation of the project in MMW.

Connects #2999 

### Demo

https://beta.hydroshare.org/resource/e98d11dfc9a04b5689476665545fd3a6/

image

```bash
$ jq '.id, .name, .snapshot_at' < mmw_project_snapshot.json

37
"HydroShare Copying"
"2018-10-18T15:22:43.107Z"
```

### Notes

I considered excluding the Area of Interest from the snapshot since it is already exported in its own file, and we'll most likely only be reading the project id from the snapshot file for now. But this doesn't seem to add a significant delay to the export, and maintains a complete representation of an MMW project, so I've kept it in for now.

## Testing Instructions

* Find the HydroShare Beta Client Secret Key in LastPass and populate `/etc/mmw.d/env/MMW_HYDROSHARE_SECRET_KEY` in the `app` VM and restart `mmw-app` service
* Go to [:8000/](http://localhost:8000/) and log in. Go to your account and under linked accounts, unlink from HydroShare if you are already linked. Then re-link to HydroShare. (You'll need a HydroShare Beta account to do this, you can make one if you don't have it already).
* Make a project and export it to HydroShare. Ensure the HydroShare export contains the `mmw_project_snapshot.json` file and it is an accurate representation of the project (has the right ID and name).


src/mmw/js/src/modeling/models.js
```diff
@@ -502,6 +502,16 @@ var ProjectModel = Backbone.Model.extend({
         return url;
     },
 
+    getSnapshot: function() {
+        // Stringify and parse the project to convert all Backbone models and
+        // collections to JSON objects and arrays.
+        var snapshot = JSON.parse(JSON.stringify(this));
+
+        snapshot.snapshot_at = (new Date()).toISOString();
```
jwalgran: How much of a problem is it that the snapshot date is coming from the client? We can't 100% trust the user's clock.

rajadain: Hmm, good point. I'll look into doing this on the server side.

Issue #3014, opened 2018-10-31
Unit Conversion: Set User Preferred Units


rajadain:
Add a variable to the `user` object that is their preferred unit scheme, Metric or US Customary. Make it so that this is saved to the database, and read at load. Add a section in the User Account section to be able to set preferred units.

The default should be Metric. Users that are not logged in will not be able to change to US Customary.

PR #3019, opened 2018-11-13
Unit Conversion: User Preference

rajadain:
## Overview

Adds a user preference for favored unit scheme, which defaults to Metric, but can be set to US Customary. This preference is saved to the user's account. On load and on update, this changes `clientSettings.unit_scheme`, which will be read at run time to decide which units to show.

Connects #3014 

### Demo

image

## Testing Instructions

* Check out this branch and `bundle.sh`
* Run `manage.sh migrate` to run migrations
* Run `debugserver.sh`
* Go to [:8000/](http://localhost:8000/) and in the browser console, check for `clientSettings.unit_scheme`. Ensure it is 'METRIC'
* Create a new account in MMW. In the registration modal, ensure you see a dropdown for Unit Scheme, which should default to Metric.
* In the browser console, check that `clientSettings.unit_scheme` is still 'METRIC'
* Choose 'US Customary' as the scheme and save the registration. Ensure that `clientSettings.unit_scheme` is now set to `USCUSTOMARY`.
* Go to your account and change your preference there. Ensure it is reflected correctly in the browser console after saving.
* Refresh the page. Ensure your preference is saved correctly and that `clientSettings.unit_scheme` has the correct value.


kellyi:
After signing up:

screen shot 2018-11-13 at 5 40 58 pm

kellyi:
After changing it in the UI:

screen shot 2018-11-13 at 5 41 44 pm

Actual UI:
screen shot 2018-11-13 at 5 41 51 pm

kellyi:
Encountered an edge case which I think we should either fix or make a new issue to track:

If I sign in as one user, then refresh the page, that user's unit_scheme setting is in force.

If I sign out, then sign in as another user without refreshing the browser, the prior unit scheme is still in force.

To replicate: set up two users, one with metric and the other with customary.

Sign in as the metric user, then refresh the browser and check `clientSettings.unit_scheme`.

Without refreshing:
- sign out
- sign in as the customary user
- check `clientSettings.unit_scheme`

It looks like it'll still be `"METRIC"`.

kellyi:
(Everything else seems to be working well btw!)

rajadain:
Great catch. Added in 763f152.

PR #3033, opened 2018-11-27
Unit Conversion: MapShed

rajadain:
## Overview

Implements Unit Conversion for MapShed.

Connects #3024 

### Demo

image

image

image

image

image

image

image

image

image

image

### Notes

They had asked to use US short ton for Sediments, since its values are so much larger than Nitrogen and Phosphorus which are better represented in pounds. However, some of our tables (particularly the summary table in Water Quality result tab in MapShed) put the units in the first column for each row, which could not resolve different cells having different units, so I've kept them all as pounds for now. This can be addressed at a later stage.

## Testing Instructions

* Checkout this branch and `bundle --debug`
* Login and set your units as "US Customary"
* Make a MapShed project for a well known area. On Staging, also make a MapShed project using the same well known area.
* As you go through MapShed, ensure that US Customary units are used throughout: in the results, in the compare tab, in manual entry, in land cover entry, in conservation practices modifications.
* Compare with the project on staging and ensure the conversions make sense
* Go to settings and change your units to Metric. Re-open the project. Ensure all results and your modifications are appropriately converted.


rajadain:
Thanks for taking a look!

Issue #3028, opened 2018-11-16
Unit Conversion: HydroShare CSV Generation


rajadain:
HydroShare CSV generation is separate from the rendered tables as of https://github.com/WikiWatershed/model-my-watershed/commit/dbeff8cb97c30f32169e3568ad8430e73aa26d6c. These need the correct units too.

Issue #2575, opened 2017-12-28
Change filenames for CSVs downloaded from Analyze UI


rajadain:
Currently the CSVs for Analyze tabs do not match those downloaded from the UI. This is explained in the commit message here: https://github.com/WikiWatershed/model-my-watershed/commit/7b2b1125d158f7287c81ac6e497c5141f7aa1052

> Currently, the CSV Download in the UI is handled via the tableExport library which converts a rendered HTML table to CSV and downloads the file. Since when we're exporting from the MultiShareModal we will not have the renderings at hand to reference, we must work off the internal data each Analyze Task has.

~Make it so that the CSV exported to HydroShare matches the formatting of the downloaded one.~
See https://github.com/WikiWatershed/model-my-watershed/issues/2575#issuecomment-367437360

ajrobbins:
Stroud would like the HS-exported CSVs to exactly match the ones downloaded from the UI, in schema and filename.

kellyi:
For reference here are the downloadable values for this AOI:

![screen shot 2018-02-20 at 2 55 54 pm](https://user-images.githubusercontent.com/4165523/36446080-303969f2-164e-11e8-8472-dca5d4bbee8c.png)

Streams:

streams2018-02-20T19_55_22.470Z.csv
```csv
"Stream Order","Total Length (km)","Mean Channel Slope (%)"
1st,26.96,0.68%
2nd,24.06,0.10%
3rd,3.70,0.42%
4th,4.61,0.01%
5th,0.00,"No Data"
6th,45.06,0.01%
7th,0.00,"No Data"
8th,0.00,"No Data"
9th,0.00,"No Data"
10th,0.00,"No Data"
Other,7.51,"No Data"
Combined,111.90,0.20%
```

Land ->

nlcd_land_cover_2018-02-20T19_58_37.199Z.csv
```csv
Type,"NLCD Code","Area (km2)","Coverage (%)"
"Open Water",11,18.53,8.0
"Perennial Ice/Snow",12,0.00,0.0
"Developed, Open Space",21,28.25,12.2
"Developed, Low Intensity",22,30.03,13.0
"Developed, Medium Intensity",23,69.17,29.9
"Developed, High Intensity",24,69.68,30.1
"Barren Land (Rock/Sand/Clay)",31,0.12,0.1
"Deciduous Forest",41,7.86,3.4
"Evergreen Forest",42,0.06,0.0
"Mixed Forest",43,0.33,0.1
[...]

PR #3036, opened 2018-11-28
Unit Conversion: HydroShare Export

rajadain:
## Overview

Augments HydroShare CSV Generation code to respect user's preferred unit scheme.

Connects #3028 

## Testing Instructions

* Check out this branch
* Set your scheme to US Customary
* Export to HydroShare
* Download result CSVs from HydroShare
* Ensure they are in the correct units


rajadain:
Actually, there is a separate card for ensuring that the Analyze CSVs match the UI export: #2673. Some background discussion for this can be found in #2575. These are marked as `enhancement`s. Thus, I don't think that is required for this effort, and this is ready for another look.

Issue #3089, opened 2019-02-21
Investigate Celery SegFaults on Staging


rajadain:
On staging we're seeing segfaults like this:

```
[2019-02-20 13:58:35,634: ERROR/MainProcess] Task a8dc7519-21e4-42db-b3aa-33ff69240367 run from job 73134 raised exception: Worker exited prematurely: signal 11 (SIGSEGV).
```

particularly around GWLF-E execution for MapShed Stage 2. These errors are not happening in local or production. Investigate and resolve.

rajadain:
I downloaded the [Lower Schuylkill HUC-10](https://github.com/rajadain/mmw-gwlfe-comparison/blob/master/tests/huc10__1341.json) for testing, and then created this test script:

```python
# test.py

import os
import json

from cStringIO import StringIO

from gwlfe import gwlfe, Parser

filename = 'huc10__1341.json'
filepath = os.path.abspath(filename)

with open(filepath, 'r') as input_json:
    mapshed_data = json.load(input_json)

    # Round Areas
    mapshed_areas = [round(a, 1) for a in mapshed_data['Area']]
    mapshed_data['Area'] = mapshed_areas

    # Prepare Input GMS
    pre_z = Parser.DataModel(mapshed_data)
    output = StringIO()
    writer = Parser.GmsWriter(output)
    writer.write(pre_z)
    output.seek(0)

    # Read Input GMS
    reader = Parser.GmsReader(output)
    z = reader.read()

    # Run the Model
    result, _ = gwlfe.run(z)

    # Write to file
    outpath = os.path.abspath('output.json')

    with open(outpath, 'w') as outfile:
        json.dump(result, outfile)
```

which runs fine on my local Worker VM:

```
vagrant@worker:/vagrant/scratch/celery-segfault$ python test.py 
vagrant@worker:/vagrant/scratch/celery-segfault$ ll
total 296
drwxr-xr-x 1 vagrant vagrant    160 Feb 21 17:03 ./
drwxr-xr-x 1 vagrant vagrant   3104 Feb 21 16:54 ../
-rw-r--r-- 1 vagrant vagrant 290782 Feb 21 16:50 huc10__1341.json
-rw-r--r-- 1 vagrant vagrant   5506 Feb 21 17:03 output.json
-rw-r--r-- 1 vagrant vagrant    808 Feb 21 17:03 test.py
```

But when I run it on Staging I get this:

```
[...]

PR #3090, opened 2019-02-22
Circumvent GWLF-E SegFaults due to NumPy

rajadain:
## Overview

A recent release of NumPy introduced a new module `_multiarray_umath`. Unfortunately, there are some tools that depend on NumPy and pull in the latest version as a build dependency. This latest version is then replaced with the version specified in our `requirements.txt`, thus the module is not available at run-time causing the segfaults.

By installing NumPy separately before `requirements.txt` is gathered, we ensure that whatever was using the most recent version of NumPy in its installation now uses the pre-installed one.

We'll have to ensure that the version of NumPy in `requirements.txt` matches the one in Ansible.

For details see:

https://github.com/numpy/numpy/issues/11871
https://stackoverflow.com/q/54153886

Connects #3089 

### Demo

image

## Testing Instructions

* Download and extract this file: [celery-segfault.zip](https://github.com/WikiWatershed/model-my-watershed/files/2894563/celery-segfault.zip)
* Check out this branch and destroy and rebuild the Worker VM
* Transfer the files into the Worker VM, and execute `test.py`

    ```
    $ python test.py
    ```

  - [x] Ensure there are no segfaults
* Compare the generated `output.json` with the included `output.verify.json`:

    ```
    $ diff output.json output.verify.json
    ```
  - [x] Ensure they match
* Go to [:8000](http://localhost:8000) and create a MapShed project.
  - [ ] Ensure the modeling stages complete successfully

### DIFF
--- a/deployment/ansible/group_vars/all
+++ b/deployment/ansible/group_vars/all
@@ -60,5 +60,6 @@ observation_api_url: "http://www.wikiwatershed-vs.org/"
 enabled_features: ''
 
 numba_version: "0.38.1"
+numpy_version: "1.14.5"
 
 redis_version: "2:2.8.4-2ubuntu0.2"
--- a/deployment/ansible/roles/model-my-watershed.app/tasks/dependencies.yml
+++ b/deployment/ansible/roles/model-my-watershed.app/tasks/dependencies.yml
@@ -2,6 +2,9 @@
 - name: Install numba
   pip: name=numba version={{ numba_version }}
 
+- name: Install numpy
+  pip: name=numpy version={{ numpy_version }}
+
 - name: Install application Python dependencies for development and test
   pip: requirements="{{ app_home }}/requirements/{{ item }}.txt"
   with_items:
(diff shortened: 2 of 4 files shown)


rajadain:
> What is the numpy version where this module was introduced?

According to https://stackoverflow.com/a/54153887, it was `numpy-1.16.0rc2` in January. The current latest release of NumPy is [1.16.1](https://pypi.org/project/numpy/1.16.1/).

> I'm curious, if it within a compatible release range, that we should just upgrade? Are you concerned with introducing breaking changes?

I know that GWLF-E and Ulmo both require NumPy, and may transitively bring in their versions. I think the amount of testing required to ensure that the new version works would exhaust the available budget, and this seems the most light-weight solution.

> If we stick with this approach, rather than keep the numpy version in sync with `requirements.txt`, can numpy just be removed from that file altogether since ansible is now installing it?

Good observation. Currently Ansible only installs NumPy for the Worker. In App, it is installed via `requirements.txt`. I can update the App's role definition to install NumPy and remove it from `requirements.txt` so we don't have to keep track.

rajadain:
Applied the suggested fixes, confirmed that it still works:

image

PR #3113, opened 2019-07-01
Add support for Ubuntu Xenial (16.04) on staging

### LINKED ISSUE #3103: Upgrade Ubuntu operating system in staging
- [x] Upgrade deployment dependencies where possible
- [x] Ensure AMIs build properly for all machine types
- [x] Update staging with newly built AMIs
- [x] Replace the bastion with an smaller T3 instance type

rbreslow:
## Overview

- Configure Gunicorn to log errors to syslog
- Configure rsyslog to forward Celery logs to syslog
- Configure rsyslog to forward syslog to Papertrail
- Remove references to monitoring service
- Reconcile VPC and DataPlane stack drift
- Reconcile `/var/lib/jenkins/.aws/staging.yml`

Resolves #3103 

## Testing Instructions

See that the CloudFormation stacks are in-sync:

image

Verify `feature/jrb/xenial-staging` is on staging:

 ```bash
maiden:~ rbreslow$ http --body https://staging.modelmywatershed.org/version.txt
origin/feature/jrb/xenial-staging 1.24.2-54-g4691c0a833f5e4ed07e080fff4f6ac675ac7e9d8
```

Select a HUC-8 unit and view its analysis:

image

See the [Papertrail logs](https://papertrailapp.com/groups/14907082/events):

image

### CODE REVIEW

File: deployment/ansible/roles/model-my-watershed.app/tasks/main.yml [hunk 1]
```diff
@@ -1,6 +1,6 @@
 ---
-- { include: configuration.yml }
 - { include: app.yml, when: "['packer'] | is_in(group_names)" }
+- { include: configuration.yml }
```
rbreslow: This is so the working directory gets created before we try to configure `mmw.app.service`.

deployment/ansible/roles/model-my-watershed.app/templates/gunicorn.py.j2
```diff
@@ -12,7 +12,7 @@ preload_app = False
 reload = True
 {% else %}
 accesslog = None
-errorlog = "{{ app_gunicorn_log }}"
+errorlog = "-"
```
rbreslow: This is so we pick up Gunicorn error logs in syslog.

deployment/ansible/roles/model-my-watershed.celery-worker/templates/celery-syslog.j2
```diff
@@ -0,0 +1,8 @@
+module(load="imfile")
+
+{% for id in range(0,celery_number_of_workers) %}
+input(type="imfile"
+      File="/var/log/celery/*-worker{{ id }}.log"
+      stateFile="imfile-state:-var-log-celery-worker{{ id }}.log"
```
rbreslow: It's worth surfacing that rsyslog doesn't like that we're defining `stateFile` while we're using a wildcard in `File`:

```
Jul 01 12:22:51 ip-10-0-5-105 rsyslogd-2207:  error during parsing file /etc/rsyslog.d/23-celery.conf, on or before line 6: parameter 'statefile' deprecated but accepted, consider removing or replacing it [v8.16.0 try http://www.rsyslog.com/e/2207 ]
Jul 01 12:22:51 ip-10-0-5-105 rsyslogd-2207:  error during parsing file /etc/rsyslog.d/23-celery.conf, on or before line 10: parameter 'statefile' deprecated but accepted, consider removing or replacing it [v8.16.0 try http://www.rsyslog.com/e/2207 ]
Jul 01 12:22:51 ip-10-0-5-105 rsyslogd:  rsyslogd's groupid changed to 108
Jul 01 12:22:51 ip-10-0-5-105 rsyslogd:  rsyslogd's userid changed to 104
Jul 01 12:22:51 ip-10-0-5-105 rsyslogd-2420:  imfile: warning: it looks like to-be-monitored file "/var/log/celery/*-worker0.log" contains wildcards. This usually does not work well with specifying a state file. [v8.16.0 try http://www.rsyslog.com/e/2420 ]
Jul 01 12:22:51 ip-10-0-5-105 rsyslogd-2420:  imfile: warning: it looks like to-be-monitored file "/var/log/celery/*-worker1.log" contains wildcards. This usually does not work well with specifying a state file. [v8.16.0 try http://www.rsyslog.com/e/2420 ]
...
Jul 01 12:23:13 ip-10-0-5-105 rsyslogd-2361:  imfile: state file '/var/spool/rsyslog/imfile-state:-var-log-celery-worker0.log' contains file name '/var/log/celery/Black-worker0.log', but is used for file '/var/log/celery/Blue-worker0.log'. State file deleted, starting from begin of file. [v8.16.0 try http://www.rsyslog.com/e/2361 ]
Jul 01 12:23:15 ip-10-0-5-105 rsyslogd-2361:  imfile: state file '/var/spool/rsyslog/imfile-state:-var-log-celery-worker1.log' contains file name '/var/log/celery/Black-worker1.log', but is used for file '/var/log/celery/Blue-worker1.log'. State file deleted, starting from begin of file. [v8.16.0 try http://www.rsyslog.com/e/2361 ]
```

This came up towards the end of the effort to get Xenial working on staging. I did some research, and the rsyslog documentation is a bit heavy. It may be worth splitting this out.

deployment/cfn/application.py
```diff
@@ -94,10 +96,7 @@ class Application(StackNode):
     def set_up_stack(self):
         super(Application, self).set_up_stack()
 
-        tags = self.get_input('Tags').copy()
-        tags.update({'StackType': 'Application'})
-
-        self.default_tags = tags
+        self.default_tags = self.get_input('Tags').copy()
```
rbreslow: This is to reconcile CloudFormation drift.

deployment/cfn/data_plane.py
```diff
@@ -299,10 +299,10 @@ def create_rds_instance(self):
             DBParameterGroupName=Ref(rds_parameter_group),
             DBSubnetGroupName=Ref(rds_subnet_group),
             Engine='postgres',
-            EngineVersion='9.4.1',
+            EngineVersion='9.4.20',
             MasterUsername=Ref(self.rds_username),
             MasterUserPassword=Ref(self.rds_password),
-            MultiAZ=True,
+            MultiAZ=Ref(self.rds_multi_az),
```
rbreslow: I parameterized this to reconcile the DataPlane drift in staging, which had this set to `False`.

deployment/cfn/utils/cfn.py
```diff
@@ -42,30 +44,42 @@ def get_availability_zones(aws_profile):
     return [AvailabilityZone(az) for az in conn.get_all_zones()]
 
 
+def get_availability_zones(aws_profile, zones):
```
rbreslow: This is an overloaded method to allow us to specify specific AZs objects we want to retrieve to assist in reconciling the VPC stack drift.

deployment/cfn/utils/cfn.py
```diff
@@ -42,30 +44,42 @@ def get_availability_zones(aws_profile):
...
     while current <= high:
-        yield '10.0.%s.0/24' % current
+        yield "10.0.%s.0/24" % current
         current += 1
 
 
-def get_recent_ami(aws_profile, ami_name, owner="self", executable_by="self"):
-    conn = boto.connect_ec2(profile_name=aws_profile)
-    filters = {'name': ami_name}
+def get_recent_ami(
```
rbreslow: Pulling in updates to this method from Cicero.

deployment/cfn/vpc.py
```diff
@@ -49,20 +46,17 @@ class VPC(StackNode):
 
     ATTRIBUTES = {'StackType': 'StackType'}
 
-    PUBLIC_SUBNETS = []
```
rbreslow: I don't think this was being used for anything.

deployment/cfn/vpc.py
```diff
@@ -49,20 +46,17 @@ class VPC(StackNode):
...
         super(VPC, self).set_up_stack()
 
-        tags = self.get_input('Tags').copy()
-        tags.update({'StackType': 'VPC'})
-
-        self.default_tags = tags
+        self.default_tags = self.get_input('Tags').copy()
         self.region = self.get_input('Region')
-        self.availability_zones = get_availability_zones(self.aws_profile)
+        self.availability_zones = get_availability_zones(self.aws_profile,
```
rbreslow: Before applying these changes, the stack JSON would have random AZs and CIDR ranges every time. This allows us to pin to specific AZs and CIDR ranges, like we do in Terraform, so we can update the existing VPC stack (without replacing everything).

deployment/packer/driver.py
```diff
@@ -71,6 +71,16 @@ def get_git_branch():
 
     return subprocess.check_output(git_command).rstrip()
 
+def get_git_desc():
```
rbreslow: This is to differentiate between the string in `version.txt` vs the actual SHA.

deployment/ansible/roles/model-my-watershed.app/tasks/main.yml
hectcastro: Interesting. This makes sense because we enable and start the service via an Ansible task now.

Was this preventing the services from coming up on boot, or you just noticed the initial failures in the logs?

deployment/ansible/roles/model-my-watershed.celery-worker/templates/celery-syslog.j2
hectcastro: I think that we can just get rid of the deprecated `stateFile` parameter here. That appears to be suppressing the warnings, and logs are still flowing.

deployment/cfn/data_plane.py
```diff
@@ -512,14 +512,6 @@ def create_dns_records(self, bastion_host, rds_database,
             'dnsPrivateRecords',
             HostedZoneId=Ref(self.private_hosted_zone_id),
             RecordSets=[
-                r53.RecordSet(
```
hectcastro: Good job recognizing that this was specific to internal to the VPC metrics and log aggregation.

deployment/packer/mmw.json
```diff
@@ -81,28 +81,6 @@
                 "Environment": "{{user `stack_type`}}"
             },
             "associate_public_ip_address": true
-        },
```
hectcastro: I recall you mentioning that you wanted to convert the Packer builder to the T3 instance family.

deployment/ansible/roles/model-my-watershed.app/tasks/main.yml
rbreslow: This was preventing the services from coming up on boot, which caused instances to fail health checks.

deployment/ansible/roles/model-my-watershed.app/templates/nginx-app.conf.j2
```diff
@@ -5,15 +5,18 @@ resolver_timeout 10s;
 
 proxy_cache_path {{ nginx_cache_dir }} levels=1:2 keys_zone=OBSERVATION:10m max_size=100m;
 
+set_real_ip_from 10.0.0.0/8;
```
rbreslow: Nice catch.

deployment/cfn/utils/cfn.py
rbreslow: I'll go through and make corrections to things. I had been using https://github.com/python/black (without tuning any of its parameters), although I agree the method signature you posted is much more pretty.

deployment/cfn/utils/constants.py
```diff
@@ -1,4 +1,5 @@
 EC2_INSTANCE_TYPES = [
+    't3.micro',
```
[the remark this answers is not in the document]
rbreslow: It's probably worth remaining consistent as single quotes. Sorry for letting this stuff slip through.

deployment/packer/mmw.json
rbreslow: Thanks for surfacing this.

rbreslow:
> The tile deliver stack drift still appears to exist (S3 bucket reference was lost)

I had ignored this since we didn't need to make any modifications to that stack and it seems like a potential time sink with not a lot to gain from.

I spent an hour or so this afternoon trying to reconcile the drift and haven't been successful. Since the S3 bucket was deleted and a new one was created out-of-band, the CloudFormation stack update will fail because it tries to create a bucket that already exists. Same with the DNS record. I believe the only option is to delete the resource so CloudFormation can recreate it, which is something we wouldn't want to do in production.

deployment/cfn/worker.py
```diff
@@ -446,11 +457,34 @@ def get_cloud_config(self):
...
+                '  - $ActionQueueTimeoutEnqueue 2\n',
+                '  - $ActionQueueDiscardSeverity 0\n',
+                '  - "*.*  @@', Ref(self.papertrail_host), ':', Ref(
+                    self.papertrail_port), '"\n',
+                'rsyslog_filename: 22-mmw-papertrail.conf\n',
                 '\n',
                 'runcmd:\n',
                 '  - cat /etc/fstab.rwd-data >> /etc/fstab\n',
-                '  - mount -t ext4 /dev/xvdf /opt/rwd-data && initctl emit rwd-ready\n',  # NOQA
+                '  - mount -t ext4 /dev/xvdf /opt/rwd-data && docker restart $(docker ps -q)\n',  # NOQA
```
hectcastro: It looks like the RWD service has a name: `mmw_mmw-rwd_1`. I think that it would be more precise to restart just that vs. any container running on boot. We don't have any other running containers now, but if we ever add others, this approach could have unintended consequences.

deployment/cfn/utils/constants.py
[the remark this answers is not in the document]
hectcastro: I think it makes sense to keep `t3.micro` in the `EC2_INSTANCE_TYPES` list, but the bastion is currently still a `t3.micro`. We should convert it back to the T2 instance family because we have 12 RIs within the account (we're going to be over no matter what because the bastion used to be a larger instance type).

Issue #3124, opened 2019-07-17
Upgrade Django Rest Swagger


rajadain:
As of the Django Upgrade in #3123, the Swagger documentation is no longer working:

![image](https://user-images.githubusercontent.com/1430060/61403286-f92aa280-a8a2-11e9-9776-12e1953bf9b0.png)

It fails with this error:

```
Internal Server Error: /api/docs/
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/generic/base.py", line 68, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/generic/base.py", line 88, in dispatch
    return handler(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework_swagger/views.py", line 84, in get
    template_name, RequestContext(request, data))
  File "/usr/local/lib/python2.7/dist-packages/django/shortcuts.py", line 21, in render_to_response
    content = loader.render_to_string(template_name, context, using=using)
  File "/usr/local/lib/python2.7/dist-packages/django/template/loader.py", line 68, in render_to_string
    return template.render(context, request)
  File "/usr/local/lib/python2.7/dist-packages/django/template/backends/django.py", line 64, in render
    context = make_context(context, request, autoescape=self.backend.engine.autoescape)
  File "/usr/local/lib/python2.7/dist-packages/django/template/context.py", line 287, in make_context
    raise TypeError('context must be a dict rather than %s.' % context.__class__.__name__)
TypeError: context must be a dict rather than RequestContext.
```

This Django change from `RequestContext` to a dict was changed in our own codebase as part of the Django Upgrade in c264aa3d70d88bc75167caa314b57a6ad8a8f8eb.

[...]

PR #3123, opened 2019-07-15
Django Upgrade

rajadain:
## Overview

Upgrades Django to latest long term supported version for security patches. This is the last version of Django to support Python 2.

Connects #3118 
Connects #3105 

### Notes

There are two known issues with the upgrade that were not resolved, and have been extracted into their own cards:

- Upgrade Django Rest Swagger #3124 
- Verbose Logging for 404s #3125 
- Celery `chord_unlock` Loop #3126

The Swagger update is higher priority, since currently our API docs are broken. The verbose logging is a minor annoyance. The Celery Loop is hard to recreate, but could hold up workers in production if happening consistently.

## Testing Instructions

* Check out this branch and destroy and recreate `app` and `worker`.
  - Also destroy and recreate `tiler` because the tiler service doesn't work after the first time #3112.
* Run migrations `manage.sh migrate`.
* Run `debugserver` and `debugcelery`.
  - Also start `debugtiler` to ensure it is working correctly.
* Go to [:8000/](http://localhost:8000/).
* Ensure you are not logged in.
* Draw a shape and proceed to Analyze. Ensure the analysis completes successfully.
* Go to the Monitor tab and search for items in all four catalogs. Ensure they return results correctly. Spot check the detail pages of those results to ensure they work too.
* Go to the Model tab and create a TR-55 project. Ensure it is created correctly.
* Add a new scenario, and add some changes to it.
* Open the Compare View and ensure it works correctly.
* Go back to the home page.
* Sign up for a new account. Ensure you see the email output in the console correctly. Log in with it.
* Search for a location on the map. Ensure you get suggestions. Ensure you can select a suggestion.
* Select a HUC-10 or HUC-8. Proceed to Analyze.
* Go the Model tab and create a MapShed project. Ensure it is created correctly.
* Add a new scenario. Make some changes. Ensure it works correctly.
* Go to the Water Quality tab, and select Subbasin
* Ensure subbasin results come through correctly

## Checklist

- [x] All JavaScript tests pass `./scripts/testem.sh`

PR #3133, opened 2019-08-01
Switch API Documentation to DRF-YASG

rajadain:
## Overview

The Django Upgrade in #3123 broken Django Rest Swagger, a project that is no longer maintained. We switch to the recommended (by the author of Django Rest Swagger) alternative drf-yasg for the API generation instead.

This new tool doesn't use the YAML documentation previously there, instead relying on OpenAPI Schema decoration. The documentation is converted to this new format, and some supplemental instruction is added as well.

Connects #3124 

### Demo

Screen Shot 2019-07-30 at 18 30 51-fullpage

## Testing Instructions

* Check out this branch and reprovision your `app` VM to install the new dependencies

      $ vagrant reload --provision app

* Go to [:8000/api/docs](http://localhost:8000/api/docs) and ensure you see the Swagger UI, and not an error page
* Compare the values with production https://modelmywatershed.org/api/docs/ and ensure that the information there has been faithfully captured
* Ensure there are no typos or confusing sentences in the opening instructions near the header

### DIFF
--- a/src/mmw/apps/geoprocessing_api/schemas.py
+++ b/src/mmw/apps/geoprocessing_api/schemas.py
@@ -1,0 +1,130 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from drf_yasg.openapi import (
+    Parameter, Schema,
+    IN_QUERY,
+    FORMAT_DATETIME, FORMAT_UUID,
+    TYPE_ARRAY, TYPE_BOOLEAN, TYPE_NUMBER, TYPE_OBJECT, TYPE_STRING
+)
+
+WKAOI = Parameter(
+    'wkaoi',
+    IN_QUERY,
+    description='The table and ID for a well-known area of interest, '
+                'such as a HUC. '
+                'Format "table__id", eg. "huc12__55174" will analyze '
+                'the HUC-12 City of Philadelphia-Schuylkill River.',
+    type=TYPE_STRING,
+)
+
+MULTIPOLYGON = Schema(
+    title='Area of Interest',
+    description='A valid single-ringed Multipolygon GeoJSON '
+                'representation of the shape to analyze. '
+                'See the GeoJSON spec '
+                'https://tools.ietf.org/html/rfc7946#section-3.1.7',
+    type=TYPE_OBJECT,
+    properties={
+        'type': Schema(type=TYPE_STRING,
+                       example='MultiPolygon'),
+        'coordinates': Schema(type=TYPE_ARRAY,
+                              items=Schema(
+                                  type=TYPE_ARRAY,
+                                  items=Schema(
+                                      type=TYPE_ARRAY,
+                                      items=Schema(
+                                          type=TYPE_ARRAY,
+                                          items=Schema(
+                                              type=TYPE_NUMBER
+                                          )))),
+                              example=[[[
+                                  [-75.16468, 39.97054],
+                                  [-75.17086, 39.95001],
+                                  [-75.14888, 39.95975],
+                                  [-75.16468, 39.97054]]]])
+    }
+)
+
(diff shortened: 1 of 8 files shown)


mmcfarland:
I generated a Token and authorized the docs with `Token XXX-XXX` but I get:

```json
{
  "detail": "CSRF Failed: CSRF token missing or incorrect."
}
```

rajadain:
This is very odd. I was able to reproduce the above error in the morning, but after restarting the VM I can no longer reproduce it. Now, I am able to make API requests without CSRF tokens:

```http
$ echo '{"coordinates":[[[[-75.16468,39.97054],[-75.17086,39.95001],[-75.14888,39.95975],[-75.16468,39.97054]]]],"type":"MultiPolygon"}' | http --print HhBb POST :8000/api/analyze/land/ Accept:application/json Authorization:"Token d0faa35ccd57214e5496f7f50ce1fc4a07c00013"

POST /api/analyze/land/ HTTP/1.1
Accept: application/json
Authorization: Token d0faa35ccd57214e5496f7f50ce1fc4a07c00013
Connection: keep-alive
Content-Length: 128
Host: localhost:8000
User-Agent: HTTPie/1.0.2

{
    "coordinates": [
        [
            [
                [
                    -75.16468,
                    39.97054
                ],
                [
                    -75.17086,
                    39.95001
                ],
                [
                    -75.14888,
                    39.95975
                ],
                [
                    -75.16468,
                    39.97054
                ]
            ]
        ]
    ],
    "type": "MultiPolygon"
}

HTTP/1.1 200 OK
Allow: POST, OPTIONS
Connection: keep-alive
Content-Encoding: gzip
Date: Fri, 02 Aug 2019 19:19:07 GMT
Location: /api/jobs/c9643a13-3194-499c-9419-779fccfb58c6/
Server: nginx
Transfer-Encoding: chunked
Vary: Accept-Encoding
Vary: Accept, Cookie, Origin

{
    "job": "c9643a13-3194-499c-9419-779fccfb58c6",
    "status": "started"
}
```

Will try a couple of other workflows to see if I can reproduce it again.

rajadain:
According to https://drf-yasg.readthedocs.io/en/1.0.2/readme.html?highlight=csrf#c-swagger-settings-and-redoc-settings, the Swagger UI adds an `X-CSRFToken` header only when `USE_SESSION_AUTH` is enabled. This header is required for CSRF verification. We turned `USE_SESSION_AUTH` off because our API works off Token Authentication, not Session Authentication (which is in place only for the web app, not for API users).

With `USE_SESSION_AUTH` turned on, we do get the `X-CSRFToken` header (and successful requests), but also the Django Login / Logout button:

image

Our options now are:

1. Leave the Django Login / Logout button in the interface enabled
    - Optionally add some text in the intro instructing the user to ignore it
2. Override the core `swagger-ui.html` template in DRF-YASG to try and hide that button, even when `USE_SESSION_AUTH` is turned on https://drf-yasg.readthedocs.io/en/1.0.2/custom_ui.html
3. Add a `@csrf_exempt` decorator to our API views so they can be queried without CSRF protection. They will still need Session / Token Authentication to work, and will still be logged.

I think doing **1** is cheapest, but **3** is most graceful, _if_ we can accept those views not being CSRF protected. Thoughts?

mmcfarland:
I agree with your assessment. Given that we have the CSRF protection currently enabled, it doesn't seem like a great trade-off to lose that in order to get a marginally easier to work with API doc page. I suggest we go with 1 and add some small text to instruct users to it's use (or non-use, as it were), as you suggest.

rajadain:
Alright, ready for another look. Updated the settings and added a line to ignore Django Login in ad1a60f. Also added an empty space at the end of each sentence in the description so adding lines in the future doesn't affect existing ones.

rajadain:
Thanks for the detailed testing! Squashed the fixup. Merging now.

Issue #3100, opened 2019-03-19
Update Front-End Dependencies


rajadain:
- [x] bootstrap
- ~nunjucks~ See #3140 
- [x] lodash
- [x] moment
- [x] jquery

PR #3138, opened 2019-08-09
Upgrade Front-end Dependencies

rajadain:
## Overview

Upgrades a number of front-end dependencies for security purposes. See commit messages for details.

Notably, Nunjucks is not upgraded, because we use [nunjucksify](https://github.com/rotundasoftware/nunjucksify) to make it work with browserify. Unfortunately, nunjucksify hasn't been worked on in 3 years, and the latest version (which we are already on) pulls in the old version of Nunjucks we're trying to move away from. This adds bloat to the JS payload, without solving the original issues, so we're going to defer that for now.

An eventual solution may be to move away from Nunjucks, or move away from Browserify, for the front-end. This will be investigated in #3140.

Connects #3100 

## Testing Instructions

* Check out this branch
* Run `./scripts/npm.sh install` to update the dependencies
* Run `./scripts/bundle.sh --debug --vendor --tests` to compile the bundle
* Run `./scripts/testem.sh` to run JavaScript tests
  - [x] Ensure all tests pass
* Go to [:8000](http://localhost:8000/) and test the app
  - [x] Ensure there are no console errors (outside of the Google Maps API Key warning)


rajadain:
Thanks for taking a look! Merging now.

Issue #3083, opened 2019-02-05
Upgrade Ulmo, Switch to Native Timeout


rajadain:
In #2866 we used a [`timeout-decorator`](https://github.com/pnpnpn/timeout-decorator) library to add timeouts to Ulmo calls. This was [not without drawbacks](https://github.com/WikiWatershed/model-my-watershed/pull/2866#discussion_r192785997), but the best we could do at the time.

Today, Ulmo has [added the ability](https://github.com/ulmo-dev/ulmo/pull/165) to [pass in timeout values for suds directly](https://github.com/ulmo-dev/ulmo/issues/155), obviating the use of that external library.

Upgrade to a version of Ulmo that exposes that native support, remove the external library, and adjust our code to use native timeouts.

https://github.com/WikiWatershed/model-my-watershed/blob/05ad197bc2de43249d52e622c895e1977077d1bc/src/mmw/apps/bigcz/clients/cuahsi/details.py#L34-L44

rajadain:
After more than 2 years, Ulmo has a new release `0.8.5`: https://pypi.org/project/ulmo/#history. This includes all the latest features. We should switch to using that rather than the custom build we currently link to.

rajadain:
Blocked until we have the green light from @dboyer that this can be worked on within the NSF contract.

dboyer:
Unblocked as work has been approved.

PR #3143, opened 2019-08-12
Upgrade Ulmo to 0.8.5

rajadain:
## Overview

This version supports the native suds timeout parameter, which obviates the need for an external timeout library that we'd been using so far.

With this upgrade, the get_values call is updated to supply the timeout to suds, and the now extraneous library is removed.

Connects #3083 

### Demo

image

### Notes

There may still be some other errors you encounter. If so, use the same shape / test on staging, and ensure you see the same failures there. If you do, then the failures are not related to the changes in this PR, but instead come from the third-party services we're querying instead.

## Testing Instructions

* Check out this branch
* Destroy and recreate the `app` VM

      $ vagrant destroy app && vagrant up app

* Go to [:8000/](http://localhost:8000/) and pick a shape
* Go to the Monitor tab search for "water"
* Switch to the CUAHSI tab and click on any of the results in the sidebar
* Wait for the values to come in
    - [x] Ensure they do
      - If they fail, ensure they also fail on staging
* Pick a new result and open it
* Wait for the first `details` call to succeed (this brings in the units)
* Then turn off your local internet to simulate a timeout for the `values` endpoint (which brings in the numbers)
* Ensure that the error you see for the timeouts are different from a plain "Error 500 during fetch", as shown in the screenshot above

### DIFF
--- a/src/mmw/apps/bigcz/clients/cuahsi/details.py
+++ b/src/mmw/apps/bigcz/clients/cuahsi/details.py
@@ -4,7 +4,7 @@ from __future__ import unicode_literals
 from __future__ import division
 
 from datetime import date, timedelta
-from timeout_decorator import timeout
+from socket import timeout
 
 from rest_framework.exceptions import ValidationError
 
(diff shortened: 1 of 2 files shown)


rajadain:
Thanks for taking a look!

Issue #3168, opened 2019-09-19
ADR for Second Front-end Stack


rajadain:
PA DEP workflow will require a new front-end, as it requires two areas of interest. We are taking this opportunity to introduce some new technology for the front-end. We don't have the designs yet (which will come from #3166), so create an outline / ADR for how the new front-end will be created and included in the current codebase.

PR #3170, opened 2019-09-24
ADR for Second Front-end Stack

rajadain:
## Overview

Adds ADR outlining how the second stack will be integrated into the app.

Connects #3168

### DIFF
--- a/doc/arch/adr-007-pa-dep-client-app.md
+++ b/doc/arch/adr-007-pa-dep-client-app.md
@@ -1,0 +1,136 @@
+# Integrating a New Client App for PA DEP
+
+## Context
+
+The Pennsylvania Department of Environmental Protection (PA DEP) has been using desktop MapShed and MMW for their workflows for some time.
+This workflow involves running the MapShed model in MMW for a HUC-12,
+and also for an area of interest within the HUC-12,
+and copying the output tables into specific parts of the [MMW BMP Spreadsheet Tool][bmp-tool].
+This is cumbersome and error prone,
+and would be much better be done in an automated fashion,
+where when users download the spreadsheet it is pre-populated with the right values in the right places.
+
+Since this requires having two effective areas of interest at once,
+this workflow does not fit neatly in MMW,
+which always expects a single area of interest.
+Rather than change MMW to accomodate this unusual and rare use case,
+it would be much simpler to create a new client app that is focused on this workflow.
+It will use the MMW API for all calculations.
+
+Given that we'll have a new client app,
+this is an opportunity to introduce a more modern tech stack on the front-end.
+This was recently done for [a similar project][beekeepers],
+where a new front-end stack was added within a Django app to sandbox it from the rest of the project.
+This document aims to discuss the pros and cons of different arrangements for the new client app,
+and pick one that satisfies all project constraints.
+
+The one major directive from the clients is that,
+from a user's perspective,
+the site should be deeply integrated with MMW.
+That is to say,
+it should use the same colors and branding,
+and not feel like its own project.
+
+From the development perspective,
+we should aim for using the latest front-end stacks,
+clean communication with the back-end API,
+and having a fast deployment pipeline.
+
+### Considerations
+
+To find the right balance of coupling between the two apps,
+we must consider how the code is stored and written (same vs different repositories),
+how the apps are deployed (paired or independently),
+and how the apps are accessed by users (via subdomain or subdirectory).
+These considerations are not entirely independent of each other,
+e.g. paired deployments are easier to do with the same repository.
+
+We must also consider non-technical factors,
+such as the explicit directive to have a consistent user experience between the apps.
+This requirement rules out subdomain access,
+which would break the immersion of using the same app.
+
+#### Repository: Same vs Different
+
+Advantages of having the **same repository**:
+
+  * Familiarity, as it was done in the [Beekeepers][beekeepers] project
+  * All issues can be in one repo, simplifying project management
+  * Tasks that affect both front-end and back-end can be tracked in one commit or PR
+  * Simpler deployment, as front-end and back-end will always be in sync
+  * Styles can be easily reused between the two apps
+  * Easier to simulate production environments in development,
+    as we only need to start a single development environment.
+
+Advantages of having **different repositories**:
+
+  * Cleaner logical separation in the codebase
+  * Fewer cohabiting developer entry points, i.e. `package.json`, or multiple `node_modules`
+  * Does not require changing the `node` or `npm` version in MMW
+  * Does not require adding a Docker environment in MMW for building the front-end
+    - In case we do not change the `node` version in MMW,
+      we can isolate the new build toolchain in a Docker environment
+  * Faster deployment, as the client app can be deployed independently of the main app
+
+#### Deployment: Paired vs Independent
+
+Advantages of **paired deployment**:
+
+  * Front-end and back-end will always be in sync
+  * Does not require extra configuration of AWS
+  * Does not need extra steps during deployment
+  * Easier to write end-to-end tests
+
+Advantages of **independent deployment**:
+
+  * Front-end can be iterated much faster, without the heavy deployment cycle of MMW
+
+### Architecture Options
+
+#### Option A: Same Repository, Paired Deployment
+
+Similar to Beekeepers,
+we will create a Django app to house all files for the new client app.
+We will evaluate if the `node` version in the App VM can be upgraded,
+and if not we will use add a Docker container for building the front-end.
+Django will be configured to handle routing,
+which will render a different home page for the new routes.
+Deployments will not be affected at all and will proceed as usual.
+Some basic styles and colors will be shared between the two apps:
+they will have independent stylesheets, but some common resources,
+such as colors,
+may be shared between the two.
+
+#### Option B: Different Repositories, Independent Deployments
+
+The new app gets its own repository,
+which better isolates the new front-end code from legacy code.
+During development, the MMW VM would have to be running in addition to the client app.
+The app will be deployed to a static host like S3 or Netlify.
+nginx will be configured to proxy the static host at the subdirectory route.
+Deployments of cosmetic changes to the new client app can be made independently,
+but functional changes to how the API is consumed will have to be coordinated with MMW.
+Project management will be split across the two repos,
+with front-end tasks listed on the new repo and back-end tasks listed on MMW.
+
+## Decision
+
+We will go with Option A above,
+keeping one repository for the codebase and doing paired deployments.
+This has the advantages of being a familiar approach,
+keeping the back- and front-ends in sync,
+and not needing any changes to the deployment pipeline.
+
+## Consequences
+
+Since the more modern stack of the new app will require a more modern version of `node`,
+we will have to investigate if the one in the App VM can be safely upgraded without affecting the current `bundle` pipeline,
+or if it will require isolation within Docker as done for Beekeepers.
+
+As the new app will be deeply integrated with MMW,
+any changes,
+no matter how small,
+will require a full MMW deployment.
+
+[bmp-tool]: https://github.com/WikiWatershed/MMW-BMP-spreadsheet-tool
+[beekeepers]: https://github.com/project-icp/bee-pollinator-app/tree/develop/src/icp/apps/beekeepers


doc/arch/adr-007-pa-dep-client-app.md
```diff
@@ -0,0 +1,122 @@
...
+
+Advantages of **subdirectory access**:
+
+  * Maintains MMW brand identity
+  * Would not have any CORS issues
+    - This would not apply to the documented public API endpoints,
+      but there are some undocumented APIs such as MapShed which we'll need to run,
+      which may have CORS headers set on them.
+
+### Option A: Same Repository, Paired Deployment, Subdirectory Access
```
kellyi: We chatted about this in person but just wanted to record it here, too:

Since the prior test outlines three pairs of options, each of which seems relatively independent as presented above, it's not clear why there are only two options presented here rather than however many permutations would encompass all possible options.

I think that implies either that the prior pairs of options aren't really independent of one another or some of the un-mentioned groupings aren't really viable enough even to consider. I think it'd be helpful to add a little clarification here about why the other sets of options aren't viable -- and why these two options are.

doc/arch/adr-007-pa-dep-client-app.md
```diff
@@ -0,0 +1,122 @@
...
+The app will be deployed to a static host like S3 or Netlify.
+nginx will be configured to proxy its contents at the subdirectory route.
+Deployments of cosmetic changes to the new client app can be made independently,
+but those of functional changes to how the API is consumed will have to be coordinated with MMW.
+Project management will be split across the two repos,
+with front-end tasks listed on the new repo and back-end tasks listed on MMW.
+
+## Decision
+
+The main argument for having a separate repository is better code separation,
```
kellyi: Consider adding a single sentence at the beginning here, probably in its own paragraph, which states the decision that was reached. Consider also reframing this explanation to have a form like:

"doing X will make it easier to a, b, and c..."

doc/arch/adr-007-pa-dep-client-app.md
```diff
@@ -0,0 +1,122 @@
...
+### Repository: Same vs Different
+
+Advantages of having the **same repository**:
+
+  * Familiarity, as it was done in the [Beekeepers][beekeepers] project
+  * All issues can be in one repo, simplifying project management
+  * Tasks that affect both front-end and back-end can be tracked in one commit or PR
+  * Simpler deployment, as front-end and back-end will always be in sync
+  * Styles can be easily reused between the two apps
+  * Easier to simulate production environments in development,
```
hectcastro: Just wanted to add that there is precedent for adjusting the client application's backend configuration such that it targets staging vs. a local development environment. This works for a lot of cases, but falls apart a bit in the event that a set of changes span the frontend and backend.

doc/arch/adr-007-pa-dep-client-app.md
```diff
@@ -0,0 +1,122 @@
...
+  * Does not need extra steps during deployment
+  * Easier to write end-to-end tests
+
+Advantages of **independent deployment**:
+
+  * Front-end can be iterated much faster, without the heavy deployment cycle of MMW
+
+### Access: Subdomain vs Subdirectory
+
+Advantages of **subdomain access**:
```
hectcastro: Some of the detail behind this gets described under Option A, but to clarify, the approach here would be to:

- Have `foo.modelmywatershed.org` map to the existing MMW application instance
- Dispatch the assets associated with the app in question based on `Host` header matches via Django?

doc/arch/adr-007-pa-dep-client-app.md
```diff
@@ -0,0 +1,122 @@
...
+From the development perspective,
+we should aim for using the latest front-end stacks,
+clean communication with the back-end API,
+and having a fast deployment pipeline.
+
+### Repository: Same vs Different
+
+Advantages of having the **same repository**:
+
+  * Familiarity, as it was done in the [Beekeepers][beekeepers] project
```
hectcastro: Not to devalue the other points, but having done it one way before (and if folks feel like it was a successful approach) carries a bit of extra weight for me.

doc/arch/adr-007-pa-dep-client-app.md
```diff
@@ -0,0 +1,122 @@
...
+Advantages of **subdomain access**:
+
+  * More easily supports independent deployments
+
+Advantages of **subdirectory access**:
+
+  * Maintains MMW brand identity
+  * Would not have any CORS issues
+    - This would not apply to the documented public API endpoints,
+      but there are some undocumented APIs such as MapShed which we'll need to run,
```
fungjj92: So I have no idea how MMW plumbing works, but I do understand Bees. Is it possible at this time to write with more certainty about the development we expect to do? Model changes / migrations? Public GET requests? Protected POST requests? Any keys needed on requests? Anyway, some clarity here would help support the conclusion I think. 

I'm definitely biased from Beekeepers, whose dev env and arch pipeline was pretty seamless/avoided a lot of headaches. The conclusion makes sense to me.

doc/arch/adr-007-pa-dep-client-app.md
```diff
@@ -0,0 +1,122 @@
...
+Deployments will not be affected at all and will proceed as usual.
+Some basic styles and colors will be shared between the two apps.
+
+### Option B: Different Repositories, Independent Deployments, Subdirectory Access
+
+The new app gets its own repository,
+which makes the code much easier to navigate.
+During development, the MMW VM would have to be running in addition to the client app.
+The app will be deployed to a static host like S3 or Netlify.
+nginx will be configured to proxy its contents at the subdirectory route.
```
jwalgran: Does "its contents" in this sentence refer to the content of the S3 bucket or Netlify site?

doc/arch/adr-007-pa-dep-client-app.md
```diff
@@ -0,0 +1,122 @@
...
+Deployments of cosmetic changes to the new client app can be made independently,
+but those of functional changes to how the API is consumed will have to be coordinated with MMW.
```
jwalgran: Consider ~those of~

doc/arch/adr-007-pa-dep-client-app.md
```diff
@@ -0,0 +1,122 @@
...
+Project management will be split across the two repos,
+with front-end tasks listed on the new repo and back-end tasks listed on MMW.
+
+## Decision
+
+The main argument for having a separate repository is better code separation,
+and faster, leaner deployment cycles.
```
jwalgran: In the cases where front end changes depend on an API change, we would loose some of this improved velocity.

doc/arch/adr-007-pa-dep-client-app.md
```diff
@@ -0,0 +1,122 @@
...
+This is overshadowed by the simpler deployments and project management,
+easier ability to keep the front- and back-ends in sync,
+and ability to write end-to-end tests,
+if we keep the same repository.
+This also implies paired deployments and subdirectory access.
```
jwalgran: Related to [Kelly's comment above](https://github.com/WikiWatershed/model-my-watershed/pull/3170/files#r328138808) consider adding a some explanation of why it does not make sense to have a single repository and deploy the new app to a subdomain.

doc/arch/adr-007-pa-dep-client-app.md
```diff
@@ -0,0 +1,122 @@
...
+
+  * More easily supports independent deployments
+
+Advantages of **subdirectory access**:
+
+  * Maintains MMW brand identity
+  * Would not have any CORS issues
+    - This would not apply to the documented public API endpoints,
+      but there are some undocumented APIs such as MapShed which we'll need to run,
+      which may have CORS headers set on them.
```
kellyi: We could add MapShed endpoints to the public API as part of this work.

doc/arch/adr-007-pa-dep-client-app.md
```diff
@@ -0,0 +1,122 @@
...
+  * Styles can be easily reused between the two apps
+  * Easier to simulate production environments in development,
+    as we only need to start a single development environment.
+
+Advantages of having **different repositories**:
+
+  * Cleaner logical separation in the codebase
+  * Fewer cohabiting developer entry points, i.e. `package.json`, or multiple `node_modules`
+  * Does not require changing the `node` or `npm` version in MMW
+  * Does not require adding a Docker environment in MMW for building the front-end
```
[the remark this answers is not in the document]
rajadain: The purpose of containerization here would be to isolate the more modern version of `node` required for building the new client app, without changing the version of `node` in the App VM. I'll make that clearer. Ideally, we'd upgrade the version of the `node` in the App VM.

doc/arch/adr-007-pa-dep-client-app.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Access: Subdomain vs Subdirectory
+
+Advantages of **subdomain access**:
+
+  * More easily supports independent deployments
+
+Advantages of **subdirectory access**:
+
+  * Maintains MMW brand identity
```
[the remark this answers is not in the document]
rajadain: I will expand this section, but essentially, this is in response to the client constraint mentioned in the beginning:

> The one major directive from the clients is that, from a user's perspective, the site should be deeply integrated with MMW. That is to say, it should use the same colors and branding, and not feel like its own project.

Within MMW, we already have `/draw`, `/analyze`, `/monitor`, and `/model` routes. By keeping this app within something like `/bmp/`, we make it seem more a regular part of the app rather than something different.

Perhaps, given that the client constraint actively discourages subdomains, I should take them out of this document entirely.

doc/arch/adr-007-pa-dep-client-app.md
rajadain: Yes, just as done for BiG-CZ and Beescape.

doc/arch/adr-007-pa-dep-client-app.md
rajadain: That is not an explicit requirement of this contract, and adding the documentation needed for it may be a considerable effort. Furthermore, exposing those endpoints may have traffic implications. Furthermore, there may be business reasons to keep them undocumented at this time (potential monetization, for one).

These APIs are undocumented, but not inaccessible, so we the experts should be able to work with them without problems.

doc/arch/adr-007-pa-dep-client-app.md
```diff
@@ -0,0 +1,122 @@
...
+### Option A: Same Repository, Paired Deployment, Subdirectory Access
+
+Similar to Beekeepers,
+we will create a Django app to house all files for the new client app.
+We will evaluate if the `node` version in the App VM can be upgraded,
+and if not we will use add a Docker container for building the front-end.
+Django will be configured to handle routing,
+which will render a different home page for the new routes.
+Deployments will not be affected at all and will proceed as usual.
+Some basic styles and colors will be shared between the two apps.
```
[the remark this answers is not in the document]
rajadain: Option A would not use the same stylesheets. We should attempt to share color variables between the apps.

doc/arch/adr-007-pa-dep-client-app.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Option B: Different Repositories, Independent Deployments, Subdirectory Access
+
+The new app gets its own repository,
+which makes the code much easier to navigate.
```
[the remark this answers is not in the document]
rajadain: Having an app be its own repo feels easier to navigate for me rather than it being nestled inside a much larger app. But that perspective may be subjective and hard to defend. I'll evaluate and remove it, if needed.

doc/arch/adr-007-pa-dep-client-app.md
rajadain: Yes. I will make that clearer.

doc/arch/adr-007-pa-dep-client-app.md
rajadain: Good point. I will add that explicitly here.

doc/arch/adr-007-pa-dep-client-app.md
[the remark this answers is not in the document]
rajadain: Fixed in d98c84d9

doc/arch/adr-007-pa-dep-client-app.md
rajadain: Fixed in 3787e246.

rajadain:
This is ready for a final review.

doc/arch/adr-007-pa-dep-client-app.md
rajadain: I removed the Access section, and added a Considerations section that discusses how the different options play in to each other.

rajadain:
Thanks for all the reviews! I squashed this into a single commit, merging now.

Issue #3185, opened 2019-10-17
Make Excel Generation Endpoint


rajadain:
- [x] Import the [BMPxlsx](https://pypi.org/project/BMPxlsx/) library
- [x] Make a new endpoint under `/api/modeling/`
- [x] Have the endpoint take a shape
- [x] For that shape, find all HUC-12 watersheds it intersects
- [x] Clip the shape to each HUC-12 watershed it intersects
- [ ] Generate a sample spreadsheet using the library and return it

The actual calculation of values in the spreadsheet will be done in a follow up card.

rajadain:
I've gotten it to the point where the input shape is clipped to the HUC-12s it intersects:

![image](https://user-images.githubusercontent.com/1430060/67607643-ee44a100-f720-11e9-96c2-db138de8eaf8.png)

PR #3191, opened 2019-10-30
Scaffold Excel Generation Endpoint 

rajadain:
## Overview

This scaffolds two new endpoints:

- `/mmw/modeling/worksheet/`: This endpoint takes a shape, pairs it with HUC-12s, and in the future will start a Celery job that calculates a number of values for each pair, resulting in a JSON blob
- `/export/worksheet/`: This endpoint takes the JSON blob generated above and returns a ZIP file containing Excel worksheets corresponding to each pair

Currently, the first endpoint returns an empty job, and prints out the pairs in the debug log. The second endpoint takes an array of objects, and returns a ZIP of Excel worksheets that are hardcoded to have the same value, for as many objects were in the input array. This scaffolding will be expanded upon in future work.

Next steps:

- Create a Celery queue that generates the JSON blob
- Update the export endpoint to generate files corresponding to the HUC-12 names, rather than `0.xlsx` like it currently does
- Add input validation
- Add tests?

Connects #3185 

### Demo

Getting a Celery Job response shape from the modeling endpoint where a shape is submitted:

```http
$ http :8000/api/modeling/worksheet/ Authorization:"Token $MMW_API_TOKEN" < triple-hucs.geojson

HTTP/1.1 200 OK
Allow: POST, OPTIONS
Connection: keep-alive
Content-Encoding: gzip
Date: Tue, 29 Oct 2019 20:50:32 GMT
Server: nginx
Transfer-Encoding: chunked
Vary: Accept-Encoding
Vary: Accept, Cookie, Origin

{
    "job": "00000000-0000-0000-0000-000000000000",
    "status": "started"
}
```

The clipped shapes are currently printed in the `debugserver` output. Porting them to GeoJSON.io, we can visualize the results like this:

image

image

Getting a ZIP file containing Excel Spreadsheets back from the export endpoint:

```http
$ http --download :8000/export/worksheet/ < test.json

HTTP/1.1 200 OK
Allow: POST, OPTIONS
Connection: keep-alive
Content-Disposition: attachment; filename="MMW_BMP_Spreadsheets.zip"
Content-Length: 244138
Content-Type: application/zip
Date: Tue, 29 Oct 2019 20:31:05 GMT
Server: nginx
Vary: Accept, Cookie

Downloading 238.42 kB to "MMW_BMP_Spreadsheets.zip"
Done. 238.42 kB in 0.00400s (58.15 MB/s)
```

Excel file with Stream Lengths filled out:

image (2)

## Testing Instructions

* Checkout this branch and run the server `$ ./scripts/debugserver.sh`
* Go to [:8000/](http://localhost:8000/) and turn on the HUC-12 boundary layer in the layer selector
* Draw a shape that intersects 2 or more HUC-12s
* Download a GeoJSON of the shape
* Using your user's MMW API Token, fire off a request to the `/mmw/modeling/worksheet/` endpoint with the GeoJSON of the shape as the body
  - [x] Ensure you see a Celery-like output
  - [x] Ensure you see the cut shapes paired with their HUC-12s in the `debugserver` output
* Fire off a request to the `/export/worksheet/` endpoint, with an array of objects like `[{},{}]` as the body
  - [x] Ensure you get a ZIP file in response
  - [x] Ensure it has as many Excel worksheets as objects in the input array
  - [x] Ensure the Excel worksheets have values in the stream length section of the MMW Output tab


src/mmw/apps/export/views.py
```diff
@@ -33,6 +36,7 @@
 SHAPEFILE_EXTENSIONS = ['cpg', 'dbf', 'prj', 'shp', 'shx']
 DEFAULT_KEYWORDS = {'mmw', 'model-my-watershed'}
 MMW_APP_KEY_FLAG = '{"appkey": "model-my-watershed"}'
+EXCEL_TEMPLATE = '/opt/app/worksheet/MMW_BMP_Spreadsheet_Tool.xlsx'
```
mmcfarland: Is this path portable between environments (dev and aws)? My initial instinct would be to make this resource in a directory relative to the `views.py` file. Is there precedent for either approach in the codebase?

src/mmw/apps/export/views.py
```diff
@@ -159,3 +163,50 @@ def shapefile(request):
...
+
+    worksheets = glob.glob('{}/*.xlsx'.format(tempdir))
+
+    # Create a zip file in memory for all the worksheets
+    stream = StringIO.StringIO()
+    with zipfile.ZipFile(stream, 'w') as zf:
+        for fpath in worksheets:
+            _, fname = os.path.split(fpath)
+            zf.write(fpath, fname)
+            os.remove(fpath)
```
mmcfarland: Do you have a sense of the memory consumption of these files in a worst case scenario? Curious to see what kind of load we could expect that would put memory pressure on the app servers. I expect the worksheet is relatively small, and that the likely number of intersected HUCs wouldn't exceed single digits.

If we have access to a logger, it would be interesting to `DEBUG` the number files created in the zip so we can monitor use metrics.

src/mmw/apps/export/views.py
rajadain: This path is portable, but doing something relative is probably better, and has precedent: 

https://github.com/WikiWatershed/model-my-watershed/blob/51367b833b72717d48b394cfd431cf2931453d94/src/mmw/mmw/settings/layer_settings.py#L24-L27

Will change.

src/mmw/apps/export/views.py
rajadain: I'll make a card to add some monitoring. Also to see if we can limit the size of the smaller area of interest, which should protect from being too large and intersecting too many HUC-12s.

rajadain:
Posted fixups for:

1. Moving the Excel template from `/src/mmw/worksheet/` to `/src/mmw/apps/export/templates/`
2. Updating `export/views.py` to use a relative path instead
3. Updating `geoprocessing_api/views.py` to have a more accurate comment description

rajadain:
Squashed fixups. Merging now.

Issue #3196, opened 2019-11-05
Nginx Service is Failing in All Environments


rajadain:
```
● nginx.service - A high performance web server and a reverse proxy server
   Loaded: loaded (/lib/systemd/system/nginx.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Tue 2019-11-05 15:28:00 UTC; 33s ago
     Docs: man:nginx(8)
  Process: 30523 ExecStartPre=/usr/sbin/nginx -t -q -g daemon on; master_process on; (code=exited, status=1/FAILURE)

Nov 05 15:28:00 app systemd[1]: Starting A high performance web server and a reverse proxy server...
Nov 05 15:28:00 app nginx[30523]: nginx: [emerg] host not found in upstream "www.wikiwatershed-vs.org" in /etc/nginx/sites-enabled/mmw-app:99
Nov 05 15:28:00 app nginx[30523]: nginx: configuration file /etc/nginx/nginx.conf test failed
Nov 05 15:28:00 app systemd[1]: nginx.service: Control process exited, code=exited status=1
Nov 05 15:28:00 app systemd[1]: Failed to start A high performance web server and a reverse proxy server.
Nov 05 15:28:00 app systemd[1]: nginx.service: Unit entered failed state.
Nov 05 15:28:00 app systemd[1]: nginx.service: Failed with result 'exit-code'.
```

Issue #3183, opened 2019-10-16
Observations Tab Isn't Loading


rajadain:
![image](https://user-images.githubusercontent.com/1430060/66949922-5d394000-f025-11e9-8eec-e858c7f26687.png)

dboyer:
From conversation with Stroud:

Based on what used to be under "observations" tab, listed here (https://wikiwatershed.org/help/model-help/mmw-tech/#overlays-tab-observations), it would be best to try to keep the EPA Permitted Point Sources (would be best if we could expose this for the entire country on this tab (perhaps limiting the viewing to some zoom scale if this is too much to view at national scale), and the USGS NWIS sites. We should talk about also making that layer available nationwide, as there are REST services available...https://waterservices.usgs.gov/rest/. I think we can let the NOAA and DEOS site go at this point.

rajadain:
There's two datasets that are already in the database:

`ms_pointsource`:

![image](https://user-images.githubusercontent.com/1430060/71929747-50317600-3168-11ea-96ce-b314a695af5f.png)
![image](https://user-images.githubusercontent.com/1430060/71929988-c7ffa080-3168-11ea-9d9c-94a04a7f0a22.png)

There is a further set, `ms_pointsource_drb`, which has more detail for the Delaware River Basin:

![image](https://user-images.githubusercontent.com/1430060/71930071-eb2a5000-3168-11ea-87f3-57594aaa9356.png)
![image](https://user-images.githubusercontent.com/1430060/71930099-f7aea880-3168-11ea-94e4-5033df16dc52.png)

The available metadata is this:

```sql
SELECT * FROM ms_pointsource LIMIT 5;

 ogc_fid |                        geom                        | npdes_id  |     city     | state |  latitude  |  longitude   |   huc_12_cod    | avg_n_conc |   lbsn_yr   |  mgd   | avgpconc  |     lbsp_yr     |  kgn_yr  | kgp_yr 
---------+----------------------------------------------------+-----------+--------------+-------+------------+--------------+-----------------+------------+-------------+--------+-----------+-----------------+----------+--------
   23070 | 0101000020E610000015C5ABAC6D1454C038A0A52BD8A24340 | WV0005363 | CLARKSBURG   | WV    | 39.2722220 |  -80.3191940 |  50200020204.00 |       21.6 | 50005660.00 | 1100.0 |           |                 | 22702570 |       
[...]

PR #3197, opened 2019-11-05
Fix Nginx Service by Removing Obsolete Host

rajadain:
## Overview

The http://www.wikiwatershed-vs.org/ domain was recently shut down, which started causing provisioning errors in our Nginx service. By removing all references to it from our configuration, we ensure provisioning works correctly. A couple other items are also updated necessary for provisioning.

This furthers the non-working nature of the Observations tab, as noted in #3183. Going forward we will be replacing that with different behavior, but that work will be done at a later stage.

This immediate fix is to restore provisioning and ensure that the Nginx service works correctly in development, staging, and production environments.

Connects #3196 

## Testing Instructions

* Checkout this branch and reprovision the App VM

  ```console
  $ vagrant destroy -f app && vagrant up
  ```

  - [ ] Ensure it provisions
  - [ ] Ensure the site is up


rajadain:
Merging this to test provisioning in Jenkins / Staging.

Issue #3199, opened 2019-11-07
Add Celery Back-end for Analyzing Active River Areas


rajadain:
Add a new analyze endpoint that takes an area of interest and returns the total active river area within that area of interest.

PR #3200, opened 2019-11-07
Add Active River Area numbers to Land analysis

rajadain:
## Overview

We are incorporating Active River Area analysis into the Land tab. Currently, this is tightly coupled with the endpoint in the back-end, so we incorporate it in to the NLCD analyze end-point.

The input values do not change, but the output of the endpoint now includes an `active_river_area` key which has the area of land under each land use which qualifies as an active river area.

This uses the same Active River Area raster which is used to visualize the layer in the layer picker.

Since now the same tab is doing both calculations, we switch to using a double RasterGroupedCount operation, which pairs the NLCD results with the Active River Area results. To get just the NLCD values as before, we sum them up ignoring the Active River Area values. The Active River Area values will be summed up on the front-end.

Connects #3199 

### Demo

```json
{
  "active_river_area": 794069.7734965094,
  "area": 2862240.200512842,
  "code": "deciduous_forest",
  "coverage": 0.03351157147209295,
  "nlcd": 41,
  "type": "Deciduous Forest"
},
```

## Testing Instructions

* Check out this branch
* Go to [:8000/](http://localhost:8000/) and open the Network tab
* Draw a shape
  - [x] Ensure that Analyze still works, especially the Land tab
* Inspect the network tab for the Land result
  - [x] Ensure in the results you see `active_river_area` key

### DIFF
--- a/src/mmw/apps/geoprocessing_api/tasks.py
+++ b/src/mmw/apps/geoprocessing_api/tasks.py
@@ -116,18 +116,21 @@ def analyze_nlcd(result, area_of_interest=None):
 
     pixel_width = aoi_resolution(area_of_interest) if area_of_interest else 1
 
+    result = parse(result)
     histogram = {}
     total_count = 0
     categories = []
 
     # Convert results to histogram, calculate total
-    for key, count in result.iteritems():
+    for (nlcd, ara), count in result.iteritems():
         total_count += count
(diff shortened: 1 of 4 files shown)


rajadain:
Thanks, fixed that and force pushed an update.

rajadain:
Thanks for reviewing and pairing on this!

Issue #2833, opened 2018-05-16
HUC-10 search is slow


jfrankl:
Searching for "Lower Sch" takes ~15 seconds to complete. There is a spinner visible, but because there are already results listed under the HUC headings, it seems like the search has already completed and failed.

We should either speed up the search, if possible, or find a way to make it more clear that the results under HUC-8, HUC-10, and HUC-12 are in progress.

![xgpscdkmtk](https://user-images.githubusercontent.com/1809908/40136985-587ef45c-5917-11e8-8a24-3ca5c9080345.gif)

lossyrob:
I'm noticing that modelmywatershed.org is not exhibiting this behavior.

For the development environment, if I run the following SQL in the vagrant app through `psql`,

```
mmw=# CREATE EXTENSION pg_trgm;
CREATE EXTENSION
mmw=# CREATE INDEX trgm_idx_boundary_huc08_name ON boundary_huc08 USING gin (name gin_trgm_ops);
CREATE INDEX
mmw=# CREATE INDEX trgm_idx_boundary_huc10_name ON boundary_huc10 USING gin (name gin_trgm_ops);
CREATE INDEX
mmw=# CREATE INDEX trgm_idx_boundary_huc12_name ON boundary_huc12 USING gin (name gin_trgm_ops);
CREATE INDEX
```

The speed issue goes away. I'll look to where this index creation will best fit into the dev environment setup to solve on the dev side.

PR #3207, opened 2019-12-10
Create pg_trgm index on boundary_huc* tables to speed up search

lossyrob:
## Overview

This PR adds the pg_trgm extension and creates indexes on the `name` column of the `boundary_huc*` tables on database setup.

[See this blog post](https://niallburkley.com/blog/index-columns-for-like-in-postgres/) about speeding up the LIKE operator with the `pg_trgm` extension.

Connects #2833

### Demo

mmw-search
### Notes

With the index, I still have seen some slow-ish database query times - but this seems to be on startup, and after a bit of warm up time the queries come back quite fast.

## Testing Instructions

* Re-run the `setupdb.sh` script as explained in the README changes as part of this PR.
* Perform the duplication steps of the related issue
* Notice it's not as slow as it once was.

## Checklist

- [ ] ~~All JavaScript tests pass `./scripts/testem.sh`~~ (no javascript changes)

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,13 +50,15 @@ $ ./scripts/bundle.sh
 To load or reload boundary data, from an `app` server, run (`scripts` is not mounted by default to the VM, you may need to copy the file over):
 
 ```bash
-$ ./scripts/setupdb.sh -b
+$ vagrant upload ./scripts/ app
+$ vagrant ssh app
+$ ./scripts/aws/setupdb.sh -b
 ```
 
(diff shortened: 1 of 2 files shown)


lossyrob:
I was seeing longer query times on cold startup, so it seems like potentially an index caching issue? I don't have a recommendation for a workaround - we could perform a few queries against the database to warm it up on startup, but that would introduce some deployment and maintenance complexity that might not be worth it.

Issue #3204, opened 2019-11-27
Switch to MapBox Static Tiles API for basemap


caseycesari:
Documentation: https://docs.mapbox.com/help/glossary/static-tiles-api/

Sample code: https://codepen.io/ccazavea/full/XWWypPz

designmatty:
@rajadain could you take a look at my adapted [MapBox Studio Style](https://api.mapbox.com/styles/v1/mwilliamsazavea/ck41mg8zu1dcx1cnwxbng7ifl.html?fresh=true&title=view&access_token=pk.eyJ1IjoibXdpbGxpYW1zYXphdmVhIiwiYSI6ImNpZ21oMmMzZzAyMmV1bmx6a20zbjllOHMifQ.JsUVaCRdxK16APj3-M1xDQ#3.1/40.612372/-90.943141/0). Let me know what you think. It's not a perfect one to one representation but it's fairly close.

If you don't have any changes, I'll copy the styles over to the Stroudcenter mapbox account.

rajadain:
The new styles have really tiny labels, can we make them bigger? Also, the default zoom seems to be different. Haven't changed anything on the Leaflet end.

Before:

![image](https://user-images.githubusercontent.com/1430060/70650853-d87e3300-1c1d-11ea-84f3-db302c1d1c2d.png)

After:

![image](https://user-images.githubusercontent.com/1430060/70650866-dd42e700-1c1d-11ea-98df-1be2c5698d3d.png)

The labels remain tiny regardless of zoom level:

![2019-12-11 13 55 27](https://user-images.githubusercontent.com/1430060/70650920-f64b9800-1c1d-11ea-837b-7008d6a88b41.gif)

rajadain:
Once the styles were transferred to the `stroudcenter` Mapbox account, I had to go in to the Share and Develop dialog and get CARTO style URL to work with Leaflet:

![image](https://user-images.githubusercontent.com/1430060/70652164-2136eb80-1c20-11ea-8e38-080cd7274464.png)

rajadain:
It's much closer now:

![2019-12-11 14 12 20](https://user-images.githubusercontent.com/1430060/70652266-54797a80-1c20-11ea-91d0-25de0476c3b6.gif)

The one missing piece is state boundaries. How difficult do you think it would be to add those?

designmatty:
That's a small tweak. The data is there but gets hidden at that zoom level. I'll fix it.

[...]

PR #3208, opened 2019-12-11
Upgrade Basemap

rajadain:
## Overview

Mapbox is increasing its Raster Tiles API pricing, causing us to use one of the newer APIs which has newly styled tiles. The tiles have been styled to generally match the previous cartography, although minor differences will persist.

This new Static Tiles API will be cheaper and more sustainably priced than the increased Raster Tiles API.

Connects #3204 

## Testing Instructions

* Check out this branch
  - [x] Ensure you see the new basemap
  - [x] Ensure it maintains the look and feel of the previous basemap


rajadain:
Thanks for taking a look!

Issue #3188, opened 2019-10-29
Manual Entry: Minor Adjustments to Curve Number Calculation


rajadain:
Change the numbers here:

https://github.com/WikiWatershed/model-my-watershed/blob/7bc57ba51dcb090451f85d5cc4c0ee4b5bac53b2/src/mmw/js/src/modeling/gwlfeModificationConfig.js#L247-L256

to be half.

| Old Coefficients | New Coefficients |
|-----------------:|-----------------:|
|           1.7969 |           1.3985 |
|           71.966 |           35.983 |

PR #3210, opened 2019-12-12
Adjust curve number calculation.

lossyrob:
## Overview

Adjusts the curve number calculations as per #3188.

Connects #3188

## Checklist

- [x] All JavaScript tests pass `./scripts/testem.sh`


lossyrob:
@rajadain just to be sure - the issues says "Change the numbers here: ... to be half", but `1.7969 / 2 != 1.3985`.

Issue #3195, opened 2019-11-01
Implement Front-end for Generating Worksheets for Area of Interest


rajadain:
Once #3192 is in, add a front-end button in the Modeling tab to generate the PA DEP worksheets for the area of interest. This button should trigger the `/mmw/modeling/worksheet/` endpoint, poll for its results, receive the payload, then send it to `/export/worksheet/`, and save the resulting file.

The button should go here:

![image.png](https://images.zenhubusercontent.com/5c9cd6db2250f9495b572d5b/17c61cff-432d-4a72-969f-a8c833d39cb0)

with a separator to make it stand further away from the other buttons.

PR #3216, opened 2019-12-20
PA DEP: Add Front-end Button to Download Worksheet

rajadain:
## Overview

A button is added to the Model tab of the Analyze section. Visually, there is a dividing line between this new button and the two existing modeling buttons.

When the button is clicked, we send a request to the initial data gathering endpoint and poll for results. A "Gathering data" message is shown beneath the button with a spinner. The button itself is disabled, so it cannot be double clicked.

When the polling finishes successfully, a "Download worksheet" message is shown beneath the button with a check mark, which is clickable. Clicking this submits a hidden form which had been populated with the results of the first endpoint, which opens a new tab and downloads the file.

In case of error gathering data, an error message is shown.

This process is divided into these two steps on the front-end because if we open a new tab or window programmatically as the effect of a non-user action (such as the first promise finishing), browsers treat that as a pop-up and block it by default. Rather than having users deal with that and manually add an exception for MMW, we have the user explicitly click some text to start the download process.

Connects #3195 

### Notes

Notably, this process _does not work_ if the user selects a WKAOI. This is not effectively communicated in the front-end.

BME has given us some verbiage to add to the shape selection screen to that effect. I will add that as a separate commit shortly.

## Testing Instructions

- Check out this branch and `bundle --debug`
- Go to [:8000/](http://localhost:8000/) and draw a shape
- Proceed to Analyze. Click on the Model tab.
  - [x] Ensure you see the third model option for generating the worksheet
- Click that option
  - [x] Ensure you see a status message indicating that the data is being gathered
  - [x] Ensure you see a spinner and it is spinning
  - [x] Ensure you cannot submit another request for the worksheet while this is in process
  - [x] Ensure it turns in to a check mark message asking you to Download the worksheet
- Click the Download worksheet status
  - [x] Ensure you can download a ZIP file which contains a geojson and XLSX file for each HUC-12 your shape corresponded to
  - [x] Open the XLSX files and ensure the numbers look roughly correct

### DIFF
--- a/src/mmw/js/src/analyze/templates/worksheetExport.html
+++ b/src/mmw/js/src/analyze/templates/worksheetExport.html
@@ -1,0 +1,37 @@
+<a href="#"
+    class="analyze-worksheet-link {{ 'disabled' if started or complete or disabled }}"
+    {% if disabled %}
+        data-toggle="popover" tabindex="0"
+        data-html="true" data-container="body" role="button"
+        {# TODO Update link to technical documentation #}
+        data-content="Worksheets cannot be generated for boundary shapes. Please draw, delineate, or upload a shape. For more details, see <a href='https://wikiwatershed.org/help/model-help/mmw-tech/' target='_blank'>our technical documentation.</a>"
+        data-template="<div class='popover'><div class='pull-right' id='popover-close-button' onclick='closePopover()'><i class='fa fa-times' /></div><div class='popover-content'></div><div class='arrow'></div></div>"
+    {% endif %}
+>
+    <div class="analyze-model-name">
+        Watershed Multi-Year Worksheet
+    </div>
+    <p>
+        Generates an Excel worksheet using the given urban
+        area of interest, pairing it with the HUC-12s
+        it belongs to. It contains Land, Stream, and
+        Water Quality analyses generated using the
+        Watershed Multi-Year Model for the HUC-12.
+        The worksheet allows for more detailed BMP analyses.
+    </p>
+</a>
+<a href="#" class="analyze-worksheet-status status-started disabled {{ 'hidden' if not started }}">
+    <i class="fa fa-circle-o-notch fa-spin" aria-hidden="true"></i> Gathering data ...
+</a>
+<a href="#" class="analyze-worksheet-status status-complete {{ 'hidden' if not complete }}">
+    <i class="fa fa-check" aria-hidden="true"></i> Download worksheet
+</a>
+<a href="#" class="analyze-worksheet-status status-failed {{ 'hidden' if not failed }}"
+   data-toggle="popover" data-content="{{ error }}" data-container="body" data-placement="bottom">
+    <i class="fa fa-exclamation-triangle"></i> An error occurred.
+</a>
+<form id="worksheet-form" method="post" action="/export/worksheet/" target="_blank">
+    <input type="hidden" name="csrfmiddlewaretoken" value="{{ csrftoken }}">
+    <!-- Put `payload` in single quotes, to keep any unescaped chars from breaking the template -->
+    <input type="hidden" name="payload" value='{{ payload }}' />
+</form>
--- a/src/mmw/sass/pages/_analyze.scss
+++ b/src/mmw/sass/pages/_analyze.scss
@@ -216,7 +216,7 @@
       border-radius: $radius-sm;
       padding: 1rem;
 
-      > .analyze-model-package-link {
+      > a {
         white-space: normal;
 
         > .analyze-model-name {
@@ -228,6 +228,27 @@
         > p {
           font-size: $font-size-h5;
           color: $ui-grey;
+        }
+
+        &.analyze-worksheet-status {
+          display: block;
+          margin-top: 10px;
+
+          &.status-started {
+            color: $ui-primary;
+          }
+
+          &.status-complete {
+            color: $brand-primary;
+          }
+
+          &.status-failed {
+            color: $ui-danger;
+          }
+        }
+
+        &.disabled {
+          opacity: 0.75;
         }
       }
     }
--- a/src/mmw/apps/export/views.py
+++ b/src/mmw/apps/export/views.py
@@ -171,8 +171,11 @@ def shapefile(request):
 @decorators.api_view(['POST'])
 def worksheet(request):
     """Generate a ZIP of BMP Excel Worksheets prefilled with relevant data."""
+    params = request.data
+    payload = json.loads(params.get('payload', '{}'))
+
     # Extract list of items containing worksheet specifications and geojsons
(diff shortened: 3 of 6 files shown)


src/mmw/js/src/analyze/templates/worksheetExport.html
```diff
@@ -0,0 +1,26 @@
+<a href="#" class="analyze-worksheet-link {{ 'disabled' if started or complete }}">
+    <div class="analyze-model-name">
+        Watershed Multi-Year Worksheet
+    </div>
+    <p>
+        Generates a worksheet using a the given urban
+        area of interest, pairing it with the HUC-12s
+        it belongs to, containing Land, Stream, and
+        Water Quality analyses generated using the
+        Watershed Multi-Year Model for the HUC-12.
```
rajadain: Open to suggestions for improvement to this verbiage.

src/mmw/sass/pages/_analyze.scss
```diff
@@ -216,7 +216,7 @@
       border-radius: $radius-sm;
       padding: 1rem;
 
-      > .analyze-model-package-link {
+      > a {
```
rajadain: This was changed because each `analyze-model-package-link` is handled by the JavaScript to create a project with the given model package, but we want a button that looks like that but doesn't work like that. Hence the switch to a generic `a` tag.

src/mmw/js/src/analyze/templates/worksheetExport.html
mmcfarland: I'll withhold any suggestions as I believe you are returning to add some improved text. If that isn't the case, I'm happy to make suggestions.

rajadain:
Theoretically, the smaller urban area being analyzed is done so in the context of an encompassing HUC. It will almost always be smaller than a HUC-12. There may be some rare cases when a HUC-10 is needed, and BME conceded that those can be handled manually by the previously existing workflow.

I tried removing the WKAOI check:

```diff
diff --git a/src/mmw/js/src/analyze/models.js b/src/mmw/js/src/analyze/models.js
index b843a5a0..179da155 100644
--- a/src/mmw/js/src/analyze/models.js
+++ b/src/mmw/js/src/analyze/models.js
@@ -28,7 +28,6 @@ var WorksheetModel = coreModels.TaskModel.extend({
             name: 'worksheet',
             displayName: 'Worksheet',
             area_of_interest: null,
-            wkaoi: null,
             taskName: 'modeling/worksheet',
             taskType: 'api',
             token: settings.get('api_token'),
@@ -38,12 +37,10 @@ var WorksheetModel = coreModels.TaskModel.extend({
     runWorksheetAnalysis: function() {
         var self = this,
             aoi = self.get('area_of_interest'),
-            wkaoi = self.get('wkaoi'),
             result = self.get('result');
 
         if (aoi &&
             !result &&
-            !utils.isWKAoIValid(wkaoi) &&
             self.runWorksheetPromise === undefined) {
             var taskHelper = {
                     contentType: 'application/json',
```

and improved the error display in the UI in 0ec3024, which resulted in this error for a HUC-12:

image

I don't recall seeing that before, and don't have a clear understanding of what is a `Boxed Error` in a GeoTrellis context. However, if we continue to exclude HUC (and other boundaries) by design, we will likely not see these errors.

rajadain:
Just made an issue for updating the Select Area text: https://github.com/WikiWatershed/model-my-watershed/issues/3217

rajadain:
In 4dc40f0 I've added a check for WKAOIs that makes the Worksheet Export button only visible for non-WKAOI shapes:

image

mmcfarland:
> will almost always be smaller than a HUC-12

For this I was considering the other types of wkids, like County. I'm not sure if that is important or not, I was mostly just trying to surface whether it has been determined to be important or not yet.

rajadain:
How about something like this:

image

The box is slightly grayed out, and clicking it opens this popup instead.

> like County
> I was mostly just trying to surface whether it has been determined to be important or not yet.

It has not been determined. We have not talked to BME about non-HUC WKAOIs.

rajadain:
The above error message implementation is in b394717.

mmcfarland:
> The above error message implementation is in b394717.

I think that looks good :+1:

src/mmw/js/src/analyze/templates/worksheetExport.html
```diff
@@ -0,0 +1,36 @@
...
+        {# TODO Update link to technical documentation #}
+        data-content="Worksheets cannot be generated for boundary shapes. Please draw, delineate, or upload a shape. For more details, see <a href='https://wikiwatershed.org/help/model-help/mmw-tech/' target='_blank'>our technical documentation.</a>"
+        data-template="<div class='popover'><div class='pull-right' id='popover-close-button' onclick='closePopover()'><i class='fa fa-times' /></div><div class='popover-content'></div><div class='arrow'></div></div>"
+    {% endif %}
+>
+    <div class="analyze-model-name">
+        Watershed Multi-Year Worksheet
+    </div>
+    <p>
+        Generates a worksheet using a the given urban
```
mmcfarland:
```suggestion
        Generates an Excel Worksheet using the given urban
```

src/mmw/js/src/analyze/templates/worksheetExport.html
```diff
@@ -0,0 +1,36 @@
...
+        area of interest, pairing it with the HUC-12s
+        it belongs to, containing Land, Stream, and
```
mmcfarland:
```suggestion
        it belongs to. It contains Land, Stream, and
```

src/mmw/js/src/analyze/templates/worksheetExport.html
```diff
@@ -0,0 +1,36 @@
...
+        Water Quality analyses generated using the
+        Watershed Multi-Year Model for the HUC-12.
```
mmcfarland: Potentially at the end here, it should indicate what can be done with it? Or maybe just another link to the TBD-technical documentation. I was thinking something along the lines of:

> The worksheet will allow you to make more detailed BMP analyses

But truthfully, I don't have a great understanding of how the sheet is to be used.

rajadain: Let me see if I can include a link to the tech documentation somehow. Can't just pop it in because the whole box is clickable.

rajadain:
Added the proposed text changes:

image

src/mmw/js/src/analyze/templates/worksheetExport.html
rajadain: I did add that line. Forewent adding the link for now. Will confer with BME once this is up to fine tune it.

rajadain:
Thanks for the thorough review. Going to squash and merge now.

Issue #3202, opened 2019-11-08
Add Front-end for Analyzing Active River Areas


rajadain:
Similar to the length totals in the Streams tab:

<img width="1289" alt="image" src="https://user-images.githubusercontent.com/1430060/68495655-a5abde00-021e-11ea-9475-264cc62021fc.png">

Add the total "Active River Area" value after the Land Cover table. Do not show this if the area of interest falls outside the ARA layer's extent.

PR #3220, opened 2019-12-30
Add ARA output to Analyze Land tab

rajadain:
## Overview

The output, shown in square km, will be shown in the Analyze Land tab if the area of interest is within the ARA Perimeter.

In the code, we inspect all the values for ARA for all the land types. If some of them are NULL we treat it as 0, and still sum up all the values to a total. The value will be shown even if it is 0.

Only when all the values are NULL that we take it as a signal that the AoI must be outside the ARA Perimeter, in which case the entire line is hidden.

Connects #3202

### Demo

Within ARA Perimeter:

image

Within ARA Perimeter avoiding ARA values:

image

Within ARA Perimeter for small AoIs, using smaller units:

image

Outside ARA Perimeter:

image

### Notes

We are showing totals for data layer that has a corresponding visual layer. Usually, the entire Analyze tab is dedicated to a single layer, and has controls for turning on the related layers at the top.

The ARA line added to the bottom doesn't have the significance to deserve a top-level addition of the ARA layer for the chart and the table, yet it feels like something is missing. I'm not sure on the best way to proceed, and this may likely be a larger design conversation with Stroud and **#design**.

As such I'd rather defer this to a later card.

## Testing Instructions

* Check out this branch and `$ ./scripts/bundle.sh --debug`
* Analyze a shape within the ARA perimeter
  - [x] Ensure you see an "Active river area" line item at the bottom of the NLCD table
* Analyze a shape outside the ARA perimeter
  - [x] Ensure you see no such line

### DIFF
--- a/src/mmw/js/src/analyze/templates/table.html
+++ b/src/mmw/js/src/analyze/templates/table.html
@@ -12,3 +12,10 @@
     <tbody>
     </tbody>
 </table>
+{% if isLandTable and not araNull %}
+<div id="land-tab-ara-section">
+    <div class="land-tab-ara-line-item">
+        Active river area = {{ araTotal|round(2)|toLocaleString(2) }} {{ headerUnits }}
+    </div>
+</div>
+{% endif %}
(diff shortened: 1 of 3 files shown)


rajadain:
Thanks for reviewing!

Issue #2782, opened 2018-04-18
Use MultiOperation for Climate Analysis


rajadain:
The new MultiOperation endpoint accomplishes in one geoprocessing call what previously required a number of them. Since it caches all results at once, it can return cached results that much faster. Finally, it has significantly less (1/24th) the amount of Celery overhead.

Switch Climate Analysis to use the MultiOperation endpoint to lower resource utilization and improve front-end performance.

rajadain:
This can also solve some recent `chord_unlock` floodings we've seen.

PR #3221, opened 2019-12-31
Switch Climate Analysis to Multioperation

rajadain:
## Overview

To reduce Celery overhead and the possibility of unchecked `chord_unlock` calls, we switch Climate analysis to the multioperation geoprocessing call. This results in a more efficient, faster analysis that is easier to cache.

Connects #2782 

### Demo

image

## Testing Instructions

* Check out this branch and debug celery `$ ./scripts/debugcelery.sh`
* Select a WKAOI (preferably a fresh one that would not have been cached on your local machine) and analyze it
    - [x] Ensure all analyses complete successfully, including Climate
* Select the same WKAOI on staging
    - [x] Ensure the Climate analysis has the same numbers as your local
* Also try with drawing a shape, so it doesn't use the cache
    - [x] Ensure it completes successfully


rajadain:
Thanks for taking a look!

Issue #3198, opened 2019-11-05
Site Storm Model: Unable to view all modifications on a project with a very large number of modifications


SRGDamia1:
There's no scroll in the modifications drop down, so once the number of modifications exceeds the screen length they become un-viewable.

![image](https://user-images.githubusercontent.com/5374350/68241030-f41d6a80-ffdb-11e9-9cb1-18b46df3cdc3.png)

rajadain:
This was fixed in #3232 and is now available on production.

SRGDamia1:
Thank you!

PR #3232, opened 2020-01-20
Make modifications list pop-up scrollable

rajadain:
## Overview

Previously the modifications list pop-up would overflow beyond a short viewport and be cut off from view. By adding an explicit `max-height` we ensure that the pop-up always stays within the viewport, and with `overflow-y` we ensure that there is a scrollbar for vertical scrolling when required.

Connects #3198 

### Demo

On macOS, with an implicit scrollbar:

On Windows, with an explicit scrollbar:

## Testing Instructions

* Go to [:8000/](http://localhost:8000/) and create a new TR-55 project
* Add a large number of modifications, like so:

    image

* Open the modifications pop-up on a short screen, such that you cannot see all the modifications in one screen
    - [ ] Ensure you can scroll the pop-up to bring the lower ones up into view

### DIFF
--- a/src/mmw/sass/base/_header.scss
+++ b/src/mmw/sass/base/_header.scss
@@ -201,6 +201,9 @@ header {
     }
 
     #modifications {
+      max-height: calc(100vh - 152px);
+      overflow-y: scroll;
+
       .modification-type {
         min-width: 140px;
       }


rajadain:
Thanks for taking a look!

Issue #3233, opened 2020-01-22
Split Active River Area across NLCD Categories


rajadain:
Currently the Active River Area is shown only as a total at the bottom of the NLCD tab:

![image](https://user-images.githubusercontent.com/1430060/72912625-e9848e80-3d09-11ea-99eb-3fab99789e47.png)

It is also not included in the CSV export.

Split the Active River Area data across each NLCD type, by adding a column to the table showing the ARA number for the NLCD class.

Ensure the split happens across:

- [x] NLCD Table
- [x] CSV Download
- [x] HydroShare CSV Export
- [x] Tests

rajadain:
-1 to #3245

PR #3234, opened 2020-01-22
Split Active River Area across NLCD

rajadain:
## Overview

Currently the Active River Area total only is shown in the user interface. However, our users want to see the ARA split across each land use type. This adds a column to the Land Analysis table for the Active River Area value, allowing users to see how the ARA is divided between land uses.

For regions outside the Mid-Atlantic and North-East, where ARA data is not available, "No Data" is shown.

Also adds a Total row at the bottom of the Land, Protected Lands, and Soil tables to sum the columns.

Connects #3233 

### Demo

In Philadelphia, within the ARA region:

image

In Phoenix, outside the ARA region:

image

### Notes

For areas of interest outside the ARA region, I elected to have "No Data" appear in the ARA column. This has some precedent in the app, e.g. the Point Sources tab:

image

Another option is to hide the column completely. This would give the table more breathing room, and be identical in function to current behavior for areas of interest outside the ARA region. This is current behavior on staging:

image

However, this would lead to the output table having different structures in different areas of interest, which is arguably more egregious than having "No Data" in most of the country.

I can be convinced to go the other way if it is considered better though.

## Testing Instructions

* Go to [:8000/](http://localhost:8000/) and select an area within the Mid-Atlantic North-East region
* Switch to the Land Analysis tab
  - [x] Ensure you see a new column in the table
  - [x] Turn on the ARA layer and ensure the numbers roughly make sense
  - [x] Ensure you see a totals row at the bottom of the table
  - [x] Click "Download this data" and ensure the downloaded CSV has all the data shown
* Switch to Protected Lands
  - [x] Ensure table has a new totals row and its numbers are correct
* Switch to Soil tab
  - [ ] Ensure table has a new totals row and its numbers are correct

### DIFF
--- a/src/mmw/js/src/analyze/templates/table.html
+++ b/src/mmw/js/src/analyze/templates/table.html
@@ -14,11 +14,24 @@
     </thead>
     <tbody>
     </tbody>
+    <tfoot>
+    <tr>
+        <th>Total</th>
+        {% if isLandTable %}
+            <!-- Empty column for NLCD -->
+            <th class="hidden-analyze-table-column" data-tableexport-display="always"></th>
+        {% endif %}
+        <th class="text-right">{{ areaTotal|round(2)|toLocaleString(2) }}</th>
+        <th class="text-right">{{ coverageTotal|round(1)|toLocaleString(1) }}</th>
+        {% if isLandTable %}
+            <th class="text-right">
+                {% if araNull %}
+                    No Data
+                {% else %}
+                    {{ araTotal|round(2)|toLocaleString(2) }}
+                {% endif %}
+            </th>
+        {% endif %}
+    </tr>
+    </tfoot>
 </table>
-{% if isLandTable and not araNull %}
-<div id="land-tab-ara-section">
-    <div class="land-tab-ara-line-item">
-        Active river area = {{ araTotal|round(2)|toLocaleString(2) }} {{ headerUnits }}
-    </div>
-</div>
-{% endif %}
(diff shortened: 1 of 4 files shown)


lossyrob:
Question: The protected lands layer seems to have NoData values, that means each pixel is not accounted for in the area in the table. So there might be ARI where protected lands are NoData. Did the previous total take into account only pixels that were for Protected Lands, Soil or NLCD non-NoData values? Or should it show the total ARI area regardless of NLCD/PL values?

lossyrob:
This is looking good, but I noticed something while running Analysis on `Roanoke Rapids, HUC-8 Subbasin`: there seems to be ARAs in the visual layer, but all the table data is NoData. It's on the edge of the ARA layer. If I move north a couple HUC-8's to `Nottoway`, it does show data.

rajadain:
That's probably because we drew by hand a less than accurate, and conservative, outline of the area covered by the raster in #3218. If there's a better way to generate a polygon outline of a raster dataset than this, we can look in to it.

Would be open to pairing on this as you may have some new ideas.

lossyrob:
Ok, cool that makes sense - we'll tackle that next week!

PR looks good to go!

rajadain:
Thanks for taking a look! Before I merge this, I want to run it the _Notes_ section by @mmcfarland and @designmatty. The main question is:

When running this analysis in a part of the country where there is no data available for Active River Areas, should we:

- :one: Show the Active River Area column with "No Data" values, to keep the table consistent, although verbose with irrelevant information, or
- :two: Hide the Active River Area column, but leading to potential confusion about why is it sometimes there

Currently we're doing :one:.

mmcfarland:
> When running this analysis in a part of the country where there is no data available for Active River Areas, should we:

I agree with using :one: and keeping the column consistently available. `No Data` is used extensively in the output for other analysis areas, so I think it makes sense to keep that consistent, even if it glosses over some of the nuance between the different ways one could arrive at `No Data` based on the data set.  Looks good to me.

rajadain:
Thanks, merged this in. And I made #3239 for the ARA edge cases.

Issue #3242, opened 2020-01-28
Add Zoom and Center to AoI Button


rajadain:
Add a button on the map, visible only when there is an active Area of Interest, that zooms and pans the map so that the AoI is in the center. This button should be a Leaflet Map button, and accompany other buttons:

![image](https://user-images.githubusercontent.com/1430060/73310072-860acd00-41f1-11ea-8d50-b92f0c41ed92.png)

This functionality is already supported when readjusting the sidebar:

![2020-01-28 17 17 17](https://user-images.githubusercontent.com/1430060/73310405-4690b080-41f2-11ea-9170-54974464f439.gif)

Just need to add an explicit button for it.

rajadain:
Any recommendations on what icon we should use for the button, @designmatty?

PR #3247, opened 2020-01-30
Add Zoom to AoI Button

lossyrob:
## Overview

This PR adds a button to the map that zooms to an AOI if one exists.

Connects #3242

### Demo

If there is not an AOI set, the button is hidden on the map:

Screen Shot 2020-01-30 at 10 50 15 AM

Once an AOI is selected, the button shows up on the bottom right of the map:

Screen Shot 2020-01-30 at 10 51 01 AM

Here's what happens when you press it:

zoom-to-aoi

## Notes

I implemented analogous entries for the `SidebarToggleControl` control everywhere I found it, as I believe that gave the best coverage for where the new control needed to be referenced (tests, settings, compare map). However a second look on what other areas of the app should be considered from someone with broader knowledge would be great!

## Testing Instructions

 * Notice there is no button on the initial page
 * Select an AOI
 * Notice the button with the target icon; hover over to confirm correct tooltip
 * Zoom to somewhere that isn't centered on the AOI
 * Press the button: You're back!
 * "Change area" to clear the AOI. No more button!

### DIFF
--- a/src/mmw/js/src/core/fitToAoiControl.js
+++ b/src/mmw/js/src/core/fitToAoiControl.js
@@ -1,0 +1,50 @@
+'use strict';
+
+var L = require('leaflet'),
+    Marionette = require('../../shim/backbone.marionette'),
+    fitToAoiControlButtonTmpl = require('./templates/fitToAoiControlButton.html');
+
+module.exports = L.Control.extend({
+    options: {
+        position: 'bottomright'
+    },
+
+    initialize: function(options) {
+        this.mapModel = options.model;
+        this.fitToAoi = options.fitToAoi;
+    },
+
+    onAdd: function() {
+        return new FitToAoiControlButton({ model: this.mapModel,
+                                           fitToAoi: this.fitToAoi }).render().el;
+    }
+});
+
+var FitToAoiControlButton = Marionette.ItemView.extend({
+    // model: MapModel,
+    template: fitToAoiControlButtonTmpl,
+    className: 'leaflet-control leaflet-control-fit-to-aoi',
+
+    ui: {
+        button: '.leaflet-bar-button'
+    },
+
+    events: {
+        'click @ui.button': 'fitToAoi'
+    },
+
+    modelEvents: {
+        'change:size': 'render',
+        'change:areaOfInterest': 'render'
+    },
+
+    initialize: function(options) {
+        this.fitToAoi = options.fitToAoi;
+    },
+
+    templateHelpers: function() {
+        return {
+            hidden: this.model.get('areaOfInterest') === null
+        };
+    }
+});
--- a/src/mmw/js/src/core/views.js
+++ b/src/mmw/js/src/core/views.js
@@ -16,7 +16,8 @@ var L = require('leaflet'),
     modalModels = require('./modals/models'),
     modalViews = require('./modals/views'),
     settings = require('./settings'),
-    SidebarToggleControl = require('./sidebarToggleControl');
+    SidebarToggleControl = require('./sidebarToggleControl'),
+    FitToAoiControl = require('./fitToAoiControl');
 
 require('leaflet.locatecontrol');
 require('leaflet-plugins/layer/tile/Google');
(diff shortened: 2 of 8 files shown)


rajadain:
Tested it out, works great. Will take a look at the code shortly.

src/mmw/js/src/core/fitToAoiControl.js
```diff
@@ -0,0 +1,50 @@
...
+    },
+
+    initialize: function(options) {
+        this.mapModel = options.model;
+        this.fitToAoi = options.fitToAoi;
+    },
+
+    onAdd: function() {
+        return new FitToAoiControlButton({ model: this.mapModel,
+                                            fitToAoi: this.fitToAoi }).render().el;
```
rajadain: [style] A little over-indented here

src/mmw/js/src/core/fitToAoiControl.js
```diff
@@ -0,0 +1,50 @@
...
+        'change:areaOfInterest': 'render'
+    },
+
+    initialize: function(options) {
+        this.fitToAoi = options.fitToAoi;
+    },
+
+    templateHelpers: function() {
+        return {
+            disabled: this.model.get('areaOfInterest') === null
```
rajadain: [convention] Since the action here is hiding, not disabling, prefer `hidden` to `disabled`

src/mmw/js/src/core/views.js
```diff
@@ -745,11 +755,22 @@ var MapView = Marionette.ItemView.extend({
...
-    fitToAoi: function() {
-        var areaOfInterest = this.model.get('areaOfInterest');
+    /**
+    Fits the current map view to an AOI if one is set.
+    @param layer -- A leaflet layer of the AOI if one is available.
+                    If not provided, a new L.GeoJSON layer is created to
+                    obtain the bounds of the AOI.
+    **/
+    fitToAoi: function(layer) {
+        if(!layer) {
```
rajadain: [style] Prefer one space after `if` for consistency

src/mmw/js/src/core/views.js
```diff
@@ -393,6 +396,13 @@ var MapView = Marionette.ItemView.extend({
             map.addControl(new SidebarToggleControl({ model: options.model }));
         }
 
+        if (options.addFitToAoiControl) {
+            map.addControl(new FitToAoiControl({
+                model: options.model,
+                fitToAoi: function() { self.fitToAoi(); }
```
rajadain: [convention] Usually we do this with `_.bind`, something like:

```js
fitToAoi: _.bind(self.fitToAoi, self),
```

src/mmw/js/src/core/views.js
```diff
@@ -342,12 +343,14 @@ var MapView = Marionette.ItemView.extend({
     _googleMaps: (window.google ? window.google.maps : null),
 
     initialize: function(options) {
-        var map_controls = settings.get('map_controls');
+        var map_controls = settings.get('map_controls'),
+            self = this;
```
rajadain: [convention] In most cases, the `var self = this` declaration is the first in the function body. Prefer:

```js
var self = this,
    map_controls = settings.get('map_controls');
```

src/mmw/js/src/core/views.js
```diff
@@ -745,11 +755,22 @@ var MapView = Marionette.ItemView.extend({
...
+            var areaOfInterest = this.model.get('areaOfInterest');
+
+            if (areaOfInterest) {
+                layer = new L.GeoJSON(areaOfInterest);
+            }
+        }
 
-        if (areaOfInterest) {
-            var layer = new L.GeoJSON(areaOfInterest);
+        if(layer) {
```
rajadain: Does this `if` condition need to be here? Seems like `layer` will always be defined by this time.

lossyrob: If the layer is not passed in, but there is also no areaOfInterest defined, then layer here will be undefined.

src/mmw/js/src/core/views.js
lossyrob: If I do it this way, then the first argument because a jQuery event, which breaks the argument-less call. Is there a better way to state this that gets around this issue?

rajadain: Ah I see, let's keep it as is then. 👍

src/mmw/js/src/core/views.js
rajadain: Ah I see now, thanks for pointing that out. 👍

Issue #3243, opened 2020-01-28
Use Observations Tab to Show All Weather Stations


rajadain:
Since the Observations Tab [hasn't been working](https://github.com/WikiWatershed/model-my-watershed/issues/3183) for a while and we're [looking in to exposing MMW data](https://github.com/WikiWatershed/model-my-watershed/issues/3227), we should use it for showing the Weather Stations.

The structure of the table is like:

```sql
SELECT * FROM ms_weather_station LIMIT 5;

 ogc_fid | station |  location  | meanrh | meanwind | meanprecip | yearrange | begyear | endyear | eroscoeff | rain_cool | rain_warm | etadj | grw_start | grw_end |                        geom
---------+---------+------------+--------+----------+------------+-----------+---------+---------+-----------+-----------+-----------+-------+-----------+---------+----------------------------------------------------
       1 |   13876 | Birmingham |     70 |      3.1 |       1386 | 1961-1990 |    1961 |    1990 | N         |      0.15 |      0.34 |  1.24 | Mar       | Nov     | 0101000020E61000000000000000B055C0295C8FC2F5C84040
       2 |    3856 | Huntsville |     70 |      3.6 |       1452 | 1961-1990 |    1961 |    1990 | N         |      0.15 |      0.34 |  1.21 | Apr       | Oct     | 0101000020E6100000E17A14AE47B155C03333333333534140
       3 |   13894 | Mobile     |     73 |      3.9 |       1623 | 1961-1990 |    1961 |    1990 | Y         |      0.15 |      0.19 |  1.15 | Mar       | Nov     | 0101000020E610000000000000001056C0AE47E17A14AE3E40
       4 |   13895 | Montgomery |     71 |        3 |       1357 | 1961-1990 |    1961 |    1990 | N         |      0.15 |      0.19 |  1.16 | Mar       | Nov     | 0101000020E61000009A999999999955C06666666666264040
       5 |   13964 | Fort Smith |     69 |      3.4 |       1039 | 1961-1990 |    1961 |    1990 | N         |      0.21 |      0.27 |  1.02 | Apr       | Oct     | 0101000020E610000048E17A14AE9757C00AD7A3703DAA4140
(5 rows)
```

- [ ] Add an endpoint that returns all Weather Stations as GeoJSON with details in properties
- [ ] Add a layer to the Observations Tab in the Layer Selector that turns this on or off
[...]

PR #3253, opened 2020-02-03
Add weather stations to Observation tab.

lossyrob:
## Overview

This PR adds weather stations to the observation tab, which allows
the user to show all weather stations on the map. There is a new
endpoint that returns each weather station and properties stored in
the ms_weather_station table. The markers are based on the CUASHI
water level data in from Monitor. A future card will add any
additional design treatment.

Connects #3243

### Demo

observation-tab

### Notes

VizLayers is no longer used in the Observation tab, so a note is made
to clean this up in a subsiquent card.
## Testing Instructions

 * Navigate to Observation Tab
 * Notice it does not show an error anymore.
 * Notice it shows the correct count of weather stations
 * Click to turn on the weather station layers
 * Click on weather stations to show popups, notice the info is correct.
 * Turn off the weather stations and see them disappear.

### DIFF
--- a/src/mmw/js/src/core/models.js
+++ b/src/mmw/js/src/core/models.js
@@ -6,11 +6,10 @@ var Backbone = require('../../shim/backbone'),
     turfArea = require('turf-area'),
     L = require('leaflet'),
     utils = require('./utils'),
-    pointSourceLayer = require('../core/pointSourceLayer'),
+    weatherStationLayer = require('../core/weatherStationLayer'),
     drawUtils = require('../draw/utils'),
     settings = require('./settings'),
-    coreUnits = require('./units'),
-    VizerLayers = require('./vizerLayers');
+    coreUnits = require('./units');
 
 var MapModel = Backbone.Model.extend({
     defaults: {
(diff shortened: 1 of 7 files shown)


rajadain:
Taking a look now

src/mmw/js/src/core/weatherStationLayer.js
```diff
@@ -0,0 +1,47 @@
...
+    }
+};
+
+var WeatherStationPopupView = Marionette.ItemView.extend({
+    template: weatherStationPopupTmpl,
+    className: 'weather-station-popup',
+
+    templateHelpers: function() {
+        return {
+            station_id: this.model.get('station_id'),
```
rajadain: The attributes of `this.model` are available natively to the template, so this entire `templateHelpers` function can be elided.

src/mmw/js/src/core/weatherStationLayer.js
```diff
@@ -0,0 +1,47 @@
...
+            endYear: this.model.get('endYear'),
+            growStart: this.model.get('growStart'),
+            growEnd: this.model.get('growEnd'),
+            meanPrecip: this.model.get('meanPrecip')
+        };
+    }
+});
+
+module.exports.Layer = Layer;
+module.exports.WeatherStationPopupView = WeatherStationPopupView;
```
rajadain: `WeatherStationPopupView` is never used outside of this module, don't need to export it.

src/mmw/js/src/core/models.js
```diff
@@ -290,46 +289,44 @@ var ObservationsLayerGroupModel = LayerGroupModel.extend({
...
-                                leafletLayer: leafletLayer,
-                                display: display,
-                                active: false,
-                                layerType: 'observations'
-                            };
-                    }),
-                    observationLayersCollection = new Backbone.Collection(observationLayerObjects);
+        this.set('polling', true);
+
+        $.when($.ajax({ 'url': weatherStationAPIUrl, 'type': 'GET'}))
```
rajadain: We don't  need to wrap this in `$.when`. Since `$.ajax` returns a promise, we can call `$.ajax( ... ).done( ... )` directly.

src/mmw/js/src/core/templates/weatherStationPopup.html
```diff
@@ -0,0 +1,19 @@
+<div class="data-catalog-popover-result-region">
+    <div class="resource-wrapper">
+        <h3 class="resource-title">
+            Weather Station: {{ station_id }}</h2>
+        </h3>
+        <div class="resource-detail">
+            <i class="fa fa-star"></i> {{ location if location else '' }}
+        </div>
+        <div class="resource-detail">
+            <i class="fa fa-calendar"></i> {{ beginYear }} - {{ endYear|toDateWithoutTime }}
```
rajadain: Similarly, let's have a symmetrical treatment of dates here. Let's do just the year in both cases.

src/mmw/js/src/core/vizerLayers.js
```diff
@@ -14,6 +14,9 @@ var $ = require('jquery'),
     vizerIgnore = require('./settings').get('vizer_ignore'),
     vizerNames = require('./settings').get('vizer_names');
 
+// TODO: Fix or delete, as this is no longer used in the Observation Tab.
```
rajadain: I made #3255 for this. Can you add a link to that card in this comment?

lossyrob:
Removed the change around avoiding fetching on polling, as @rajadain explained that there's a mechanism for avoiding the actual network call.

rajadain:
I added a commit c23d4d46 for reducing the repetition in the weather station fetches. Previously, due to unintentional model updating effects, the fetch was called multiple times:

image

(ignore the basemap, something odd is happening on this branch but it's fine once merged to `develop`)

However, after the commit, we get only one call to fetch them:

image

rajadain:
Before this is merged, we should remove c3a0067c and b179d75f from this PR, which seem to have been accidentally included.

Issue #3245, opened 2020-01-28
Highlight Weather Stations used in MapShed Model


rajadain:
Subsequent to #3243, make it so that when a MapShed model is run, the two weather stations used in its modeling are highlighted in the Weather Station layer.

- [ ] Add the weather station IDs to the [MapShed output dictionary](https://github.com/WikiWatershed/model-my-watershed/blob/15dac2cd3921643045c1fd088ca3f2ea91014b39/src/mmw/apps/modeling/mapshed/tasks.py#L176)
- [ ] Add a line that turns on the Weather Stations to both the Hydrology and Water Quality tabs:
    ![image](https://user-images.githubusercontent.com/1430060/73311866-7f7e5480-41f5-11ea-9d82-fc33974a435b.png)
    similar to how there is a "related layer" text button in Analyze:
    ![image](https://user-images.githubusercontent.com/1430060/73311822-6a092a80-41f5-11ea-93f2-06da8fdf1482.png)
- [ ] Clicking the button should turn on the Weather Stations layer if not already visible
- [ ] The Weather Stations used in the model should be highlighted in a different color
- [ ] Turning the layer on should zoom the map out to all used weather stations
- [ ] Clicking one of these Weather Stations should show the same pop-up as in #3243, with the addition of:
  - Distance from the AoI
  - A text line that says something to the effect of: "This weather station was used to model this project."

PR #3257, opened 2020-02-20
Highlight Weather Stations used in MapShed Model

lossyrob:
## Overview

This PR adds functionality for highlighting weather stations in the
weather stations layer that are participating in a MapShed modeling
project. This includes the ability to set weather station layer
markers as highlighted and add their distance to the project AOI into
the popup. This functionality is used when the 'gis_data' is fetched
from the MapShed endpoint, which now includes the weather station IDs
used along with the distances. Also added in this PR is a toggle
button on the 'runoff' and 'quality' tabs of the MapShed model result
viewer that allows a user to switch on and off the weather
stations. This differs in the LayerPicker functionality only in that
when toggling on the weather station layer, the map will zoom to a
viewport that allows the user to see the active weather stations and
the AOI.

Connects #3245 

### Demo

mmw-weather-stations

### Notes

Optional. Ancillary topics, caveats, alternative strategies that didn't work out, anything else.

## Testing Instructions

 * [x] Select an AOI. Go to Model -> Select Watershed Multi-Year Model
 * [x] Notice the new toggle layer for both the 'Hydrology' and 'Water Quality' tabs
 * [x] Toggle the layer on. Notice the zoom to the stations with highlights.
 * [x] Check out the popups on the active weather stations. Notice the distance info and the info about how this is used in the model.
 * [x] Click the weather station layer on and off through the Layer Picker and the toggle buttons. Works in tandem!
 * [x] Keep the weather station layer on. Select 'New Project'. Notice the highlighting goes away.
 * [x] Try again, this time keep the weather station layer on. Run the mapshed model. Notice the active weather stations have the highlight, but aren't zoomed to.
 * [x] Notice the toggle says "turn off" as expected. Turn it off, and back on again. Zoom zoom!

### DIFF
--- a/src/mmw/apps/modeling/mapshed/calcs.py
+++ b/src/mmw/apps/modeling/mapshed/calcs.py
@@ -60,7 +60,11 @@ def nearest_weather_stations(shapes, n=NUM_WEATHER_STATIONS):
     subquery = '''
           (SELECT %s watershed_id, station, location, meanrh, meanwind,
                  meanprecip, begyear, endyear, eroscoeff, rain_cool,
-                 rain_warm, etadj, grw_start, grw_end
+                 rain_warm, etadj, grw_start, grw_end,
+                 ST_Distance(ST_Transform(geom, 5070),
+                             ST_Transform(
+                                ST_SetSRID(ST_GeomFromText(%s), 4326),
+                                           5070)) dist
           FROM ms_weather_station
           ORDER BY geom <-> ST_SetSRID(ST_GeomFromText(%s), 4326)
           LIMIT %s)
@@ -69,7 +73,7 @@ def nearest_weather_stations(shapes, n=NUM_WEATHER_STATIONS):
     for (_, watershed_id, aoi) in shapes:
         subqueries.append(subquery)
         geom = GEOSGeometry(aoi, srid=4326)
-        params.extend([watershed_id, geom.wkt, n])
+        params.extend([watershed_id, geom.wkt, geom.wkt, n])
 
     sql = ' UNION '.join(subqueries) + ';'
 
--- a/src/mmw/js/src/core/models.js
+++ b/src/mmw/js/src/core/models.js
@@ -6,7 +6,7 @@ var Backbone = require('../../shim/backbone'),
     turfArea = require('turf-area'),
     L = require('leaflet'),
     utils = require('./utils'),
-    weatherStationLayer = require('../core/weatherStationLayer'),
+    weatherStationLayer = require('./weatherStationLayer'),
     drawUtils = require('../draw/utils'),
     settings = require('./settings'),
     coreUnits = require('./units');
@@ -154,6 +154,7 @@ var LayerModel = Backbone.Model.extend({
         legendUnitsLabel: null,
         legendUnitBreaks: null,
         bigCZ: null,
+        bringToFront: false
     },
 
     buildLayer: function(layerSettings, layerType, initialActive) {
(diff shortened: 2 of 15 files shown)


rajadain:
Taking a look now

rajadain:
Seeing this error:

image

Steps to reproduce:

* Select a shape, go to Model
* Click "Related Layer: Weather Stations used in this model" link in the sidebar under Hydrology
* Notice the highlighted points show and the layer turns on correctly, but auto zoom out doesn't work

rajadain:
Moving this check within the condition prevents the console error, but autozoom still doesn't work:

```diff
diff --git a/src/mmw/js/src/core/models.js b/src/mmw/js/src/core/models.js
index 314e5c71..78aad971 100644
--- a/src/mmw/js/src/core/models.js
+++ b/src/mmw/js/src/core/models.js
@@ -287,11 +287,12 @@ var WeatherStationLayerModel = LayerModel.extend({
 
         if(this.get('active')) {
             // We toggled on; zoom to the active weather stations.
-            var geoms = this.activeWeatherStationPoints.slice(0);
-            if(areaOfInterest) {
-                geoms.push(areaOfInterest);
-            }
             if (this.activeWeatherStationPoints) {
+                var geoms = this.activeWeatherStationPoints.slice(0);
+                if(areaOfInterest) {
+                    geoms.push(areaOfInterest);
+                }
+
                 var geomCollection = {
                     'type': 'GeometryCollection',
                     'geometries': geoms
```

rajadain:
Another bug: I can click to see the popup of a weather station further away, but not for one in (or very near) the AoI —

rajadain:
I'll take another look when the above have been addressed.

src/mmw/js/src/core/models.js
```diff
@@ -154,6 +154,7 @@ var LayerModel = Backbone.Model.extend({
         legendUnitsLabel: null,
         legendUnitBreaks: null,
         bigCZ: null,
+        bringToFront: false
```
lossyrob: Added `bringToFront` option for layers in order to allow some layers to be promoted to the front of the map when added. This fixes the issue in the case where the project AOI was over the weather station marker and so the popup would not work.

src/mmw/js/src/core/models.js
```diff
@@ -238,6 +239,77 @@ var LayerModel = Backbone.Model.extend({
...
+    },
+
+    // Toggles the layer. If toggling on, zoom the map to fit both
+    // the AOI and the active weather stations.
+    toggleAndZoom: function(areaOfInterest, map, layerGroup) {
+        utils.toggleLayer(this, map, layerGroup);
+
+        if(this.get('active')) {
+            // We toggled on; zoom to the active weather stations.
+            var activeStationPoints = this.get('activeWeatherStationPoints'),
```
lossyrob: Moved this code to use the model `get` and `set`, which I wasn't doing beforehand mistakenly.

src/mmw/js/src/core/weatherStationLayer.js
```diff
@@ -5,30 +5,89 @@ var L = require('leaflet'),
...
+    // Highlights the active weather stations in this layer, and returns
+    // a list of point geometries that represent the active weather stations.
+    setActiveWeatherStations: function(layer, activeWeatherStations) {
+        var activePoints = [];
+        layer.eachLayer(function(point) {
+            var props = point.feature.properties;
+            if(props.station in activeWeatherStations) {
+                activePoints.push(point.feature.geometry);
+                var ws = activeWeatherStations[props.station];
+                props.distanceToProjectAoi = (ws.distance / 1000.0).toFixed(2); // Use KM.
```
lossyrob: Calling `toFixed` here instead of inside the template; in the case where the distance was 0 this was causing exceptions to be thrown during popup template rendering.

lossyrob:
@rajadain thanks for finding those issues - I hadn't tested the case where the weather station was contained within the project AOI, and that exposed some issues with the feature.

I've addressed and tested that case, and made some comments in the code (which is rebased) to indicate changes I've made.

rajadain:
Taking another look now.

src/mmw/js/src/core/weatherStationLayer.js
```diff
@@ -5,30 +5,89 @@ var L = require('leaflet'),
...
+                bindPopup(point, point.feature, true);
+            }
+        });
+        return activePoints;
+    },
+
+    clearActiveWeatherStations: function(layer, activeWeatherStations) {
+        layer.eachLayer(function(point) {
+            var props = point.feature.properties;
+            if(props.station in activeWeatherStations) {
```
rajadain: Seeing a failure on this line. To reproduce:

1. Select the City of Philadelphia-Schuylkill HUC-12
2. Go to MapShed
3. Turn on Weather Station layer
4. Click "New Project" to select another shape
5. See crash

rajadain:
Another bug, now when I turn on the related layer in MapShed, it turns on the basic weather stations layer. The active ones are not highlighted at all, and no zoom happens on turning the layer on:

image

rajadain:
I'll take another look when the above have been addressed.

src/mmw/js/src/core/models.js
```diff
@@ -238,6 +239,77 @@ var LayerModel = Backbone.Model.extend({
...
+    },
+
+    clearActiveWeatherStations: function() {
+        var leafletLayer = this.get('leafletLayer'),
+            prevActiveWeatherStations = this.get('activeWeatherStations');
+
+        this.set('activeWeatherStations', null);
+        this.set('activeWeatherStationPoints', null);
+
+        if(leafletLayer && prevActiveWeatherStations) {
```
lossyrob: New change - guard against the case where this method is called when `this.get('activeWeatherStations')` is null.

rajadain:
Alright, all the bugs seem to have been addressed and I can't find anymore. Going to take a deeper look at the code now.

src/mmw/js/src/core/weatherStationLayer.js
rajadain: Not directly reproducible, seems to have been fixed in 94f3542.

rajadain:
This is a minor issue, and can be deferred until later if too complicated, but if easy would nice to get in: if the weather stations are turned on before choosing a shape, the layer is under the AoI. If it is turned on after, it is over the AoI.

lossyrob:
> This is a minor issue, and can be deferred until later if too complicated, but if easy would nice to get in: if the weather stations are turned on before choosing a shape, the layer is under the AoI. If it is turned on after, it is over the AoI.
> 
> 

I'm not seeing this:

mmw weathe rlayer

This gif is short because of 10MB size limit, but replication steps I tried:
- Refresh state
- Turn on weather stations
- Choose the Schuylkill AOI
- Verify weather station popup works for weather station under AOI
- Run MapShed, see it becomes active - verify popups still work.

src/mmw/apps/modeling/mapshed/calcs.py
```diff
@@ -60,7 +60,11 @@ def nearest_weather_stations(shapes, n=NUM_WEATHER_STATIONS):
     subquery = '''
           (SELECT %s watershed_id, station, location, meanrh, meanwind,
                  meanprecip, begyear, endyear, eroscoeff, rain_cool,
-                 rain_warm, etadj, grw_start, grw_end
+                 rain_warm, etadj, grw_start, grw_end,
+                 ST_Distance(ST_Transform(geom, 5070),
+                             ST_Transform(
+                                ST_SetSRID(ST_GeomFromText(%s), 4326),
+                                           5070)) dist
```
lossyrob: Changed this to use `epsg:5070`

lossyrob:
To follow up on the above comment - the issue was not the popup, but the visual appearance of the layer under the AOI (which I am seeing). This will be addressed in a future issue.

rajadain:
@azavea-bot rebuild

rajadain:
CI is passing now, this is good to merge!

Issue #3254, opened 2020-02-04
Change percentage display for AoI with small percentages of protected areas


dboyer:
From MMW: Where area of interests have small proportion of protected area, significant decimals on the x axis for % cover probably needs to include 0.5% (one digit to the right of the decimal).

PR #3261, opened 2020-02-26
Alter precision for protected area graphs and tables

mmcfarland:
## Overview

When the horizontal analyze charts have a maximum coverage value of <
1%, allow the y-axis label to show 2 digits of precision. Previously,
these values would show only whole integers, labeling each tick as 0%.
Max values > 1% will retain the previous functionality of whole numbers.

This adjustment was due to adding a sparse coverage layer, Protected
Lands, to the Analyze Land tab. These values are often below 1% and
displayed this undesired behavior. 

Switching this from using 2 digits to a single digit of precision would allow
for more compact and evenly distributed ticks (it would use 4 character per
label item instead of 5), however, the number of areas that would still show
0.0% is higher. I am willing to alter this from 2 to 1 if that tradeoff works better.

Additionally, a commit is added that expands the Coverage column in the
Analyze tables since the same low values would slice off  precision for the
display. This more closely matches the behavior of the horizontal bar charts
associated with these tables.

One drawback is that some protected areas are so small that the
rounding used in the sum displayed in the total row may not always
match the sum of the values in the table column, as they be be even
below the 2-digit of precision threshold.

Since this associated fix is not mentioned specifically in the issue, it may be
better to drop it, but it does keep the values between the two visualizations
better, if not perfectly, aligned.

Connects #3254 

#### With default precision
image

#### With higher precision
image

## Testing Instructions

 * Visit an AoI with a very small proportion of Protected Areas, such as `Flatrock Creek-Auglaize River, HUC-10`
* Ensure the Land and Soil charts remain identical to staging, with whole number Coverage tick labels
* Check the Protected Area chart, and note that more significant digits of precision are used
* Visit AoI's of varying maximum (within a single class) Protected Area coverage and note the tradeoffs between display values and labels.

### DIFF
--- a/src/mmw/js/src/core/chart.js
+++ b/src/mmw/js/src/core/chart.js
@@ -46,7 +46,14 @@ function handleCommonOptions(chart, options) {
         chart.yAxis.axisLabel(options.yAxisLabel);
     }
     if (options.isPercentage) {
-        chart.yAxis.tickFormat(d3.format('.0%'));
+        // If the max proportion area is less than 1%, include
+        // 2 digits of precision on the y axis label, otherwise
+        // use whole integers
+        if (options.maxYValue < .01) {
+            chart.yAxis.tickFormat(d3.format('.02%'));
+        } else {
+            chart.yAxis.tickFormat(d3.format('.0%'));
+        }
     }
     if (options.abbreviateTicks) {
         chart.yAxis.tickFormat(d3.format('.2s'));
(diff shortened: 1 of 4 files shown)


src/mmw/js/src/core/chart.js
```diff
@@ -46,7 +46,14 @@ function handleCommonOptions(chart, options) {
         chart.yAxis.axisLabel(options.yAxisLabel);
     }
     if (options.isPercentage) {
-        chart.yAxis.tickFormat(d3.format('.0%'));
+        // If the max proportion area is less than 1%, include
+        // 2 digits of precision on the y axis label, otherwise
+        // use whole integers
+        if (options.maxYValue < .01) {
```
mmcfarland: This `if` statement could be further graduated to display a more granular precision based on max Y values (...`< .001`, etc), but I thought this was a reasonable middle ground.

mmcfarland:
The tests have been fixed and CI passed. This is ready for review.

rajadain:
Made #3262 for the follow-up bug mentioned above.

mmcfarland:
Thanks for the review and the follow up card :+1:

Issue #3263, opened 2020-02-28
Add button to support login with Concord Consortium Learn Portal


dboyer:
Note: Work should be associated with NSF ITEST billing.

This button behaves similarly to the existing Login with ITSI button. Text should read "Login to Learn Portal." 

Email with specific technical options forwarded to @rajadain

rajadain:
Outline:

1. Create an account on https://learn.concord.org, for testing SSO
2. Add a button in the login UI that says "Login with Learn Portal"
3. Add back-end to handle the request and perform the OAuth dance for logging in
    - Uses same OAuth credentials as ITSI
4. Cannot use `?next=/` parameter in the response anymore, since the OAuth response needs to be to the same URL
    - Consider saving it in the session

PR #3281, opened 2020-04-03
Add Concord Learn Portal Single Sign-On

rajadain:
## Overview

Adds a "Learn Portal" Single Sign-On option.

---

This is a newer version of the educational portal ITSI. ITSI may be phased out in the future, but must supported for now. The Learn Portal SSO (referred to as Concord SSO in the code) implements a stricter version of [the OAuth2 spec](https://tools.ietf.org/html/rfc6749#section-4.2.1). The differences from ITSI are:

1. The `redirect_uri` needs to be static, i.e. it cannot have a `?next=/` query parameter
2. MMW needs to include a `response_type=code` parameter for the redirect to work

For the first, we include the redirection in the `state` parameter. For the second, we specify it in the request.

---

Since we're only using Concord for authentication and not for project integration, we don't use a staging environment. All authentication is done from the live site, using the same URL in all environments. The CLIENT_SECRET_KEY is also the same in all environments. In this case, it is used as an environment variable for security, not dynamism.

Credentials for logging in to Learn Portal are in LastPass if you search for "Learn Portal Concord".

The CLIENT_SECRET_KEY can be found in LastPass if you search for MMW_CONCORD_SECRET_KEY.

---

Single Sign-On continues to be excluded from BiG-CZ.

---

Much of the work in this PR was making our SSO implementation more generic and extensible to support multiple SSO providers. The first few commits are to that end, while ensuring that ITSI continues to work as before. Learn Portal support is added in the final few commits.

See commit messages for more details.

Connects #3263 

### Notes

Classically it is [not recommended](https://auth0.com/docs/protocols/oauth2/redirect-users) to use plain-text URLs for redirection in the `state` parameter. The recommended approach is to send and receive a generated `nonce` that can be matched to ensure the validity of the response. However, that requires local storage somewhere. The obvious spot is the session, however anonymous users don't have a session, and thus that's not available.

More time could be spent devising a solution to this, but it did not seem like a large enough risk at this time.

#### Preparation

* Get the value of MMW_ITSI_SECRET_KEY and MMW_CONCORD_SECRET_KEY from LastPass
* Check out this branch and reprovision `app`:

    ```console
    $ MMW_ITSI_BASE_URL="https://itsi.portal.concord.org/" MMW_ITSI_SECRET_KEY="***" MMW_CONCORD_SECRET_KEY="***" vagrant reload app --provision
    ```

* Run migrations

    ```console
    $ ./scripts/manage.sh migrate
    ```

* Go to [:8000](http://localhost:8000/) and login using a traditional MMW user
    - Run `debugserver` if you don't already have an account to sign up for one

* Create a modeling project (TR-55 or MapShed, either works) and set it to be public. Copy it's public URL.

* Logout of that account. Go to the public project's URL to ensure you can still view it as an anonymous user.

#### Verifying ITSI Still Works

* While on the public project page, click Login, and choose ITSI
* When redirected to itsi.portal.concord.org, login with the `mmwteacher` account (credentials are in LastPass)
* [x] Ensure you are redirected back to the project's page
* [x] Ensure a new account is created for you using the first, last, and user names from ITSI
* Logout of this account
* Log back in using ITSI
* [x] Ensure you are logged in to the same account

#### Verifying Learn Portal SSO Works

* Logout of MMW
* Go back to the public project page, click Login, and choose Learn Portal
* When redirected to learn.concord.org, login with the `mmwteacher` account (credentials are in LastPass)
* [x] Ensure you are redirected back to the project's page
* [x] Ensure a new account is created for you using the first, last, and user names from the Learn Portal
* Logout of this account
* Log back in using the Learn Portal
* [x] Ensure you are logged in to the same account

#### Verifying SSO is not available for BiG-CZ

* Logout if logged in
* Open the login modal
* [x] Ensure you don't see any SSO options


mmcfarland:
I've been unable to complete the ITSI login test. I get an error, the following URL is what was generated:

image

```
https://learn.staging.concord.org/auth/concord_id/authorize?redirect_uri=http%3A%2F%2Fsantacruz.internal.azavea.com%3A8000%2Fuser%2Fitsi%2Fauthenticate%3Fnext%3D%2Fproject%2F10%2Fscenario%2F10&client_id=model-my-watershed
```

The two secret keys appear to be provisioned onto the app server correctly:
image

rajadain:
Ah sorry about that, I think there is an error in my provisioning instructions. The MMW_ITSI_BASE_URL is not updated automatically by providing it as an environment variable, must be edited manually. Please run:

```console
$ vagrant ssh app -c 'sudo vim /etc/mmw.d/env/MMW_ITSI_BASE_URL'
```

and set its value to `https://itsi.portal.concord.org/`, then run `debugserver` to pick up the new variable.

Not sure why Learn Portal is not working though, that should as long as the key is correct. You are accessing MMW at `localhost:8000`, right? We have `localhost:8000/user/concord/authenticate` (and the corresponding staging and production URLs) set in Learn Portal as the blessed redirect URLs for this app.

rajadain:
Thank you for taking a look!

Issue #3292, opened 2020-04-14
Deploy to Production


rajadain:
Made a [`release/1.28.0`](https://github.com/WikiWatershed/model-my-watershed/compare/master...release/1.28.0) branch that includes #3293, a copy of #3281 for release. Should get this out at ops convenience.

rajadain:
Reopening to count for Civic Apps.

PR #3300, opened 2020-04-17
Pin llvmlite to 0.31.0 to fix builds

rajadain:
## Overview

With the release of llvmlite [v0.32.0](https://github.com/numba/llvmlite/releases/tag/v0.32.0) 2 days ago being auto-picked up by numba, builds were breaking. By pinning llvmlite to 0.31.0, the latest working version, we protect ourselves from the breakage.

We've been using the same version of numba for a few years now, which started with llvmlite 0.23.0, so we're okay by not always using the latest version.

Connects #3292 

## Testing Instructions

Destroy and reprovision app. Ensure it builds successfully.

### DIFF
--- a/deployment/ansible/group_vars/all
+++ b/deployment/ansible/group_vars/all
@@ -51,6 +51,7 @@ nginx_cache_dir: "/var/cache/nginx"
 
 enabled_features: ''
 
+llvmlite_version: "0.31.0"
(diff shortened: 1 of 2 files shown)


rajadain:
Thanks for taking a look! Made the variable in 2b3a834 and merged.

Issue #3298, opened 2020-04-15
Custom Weather: Rewrite Button to Associate with Scenarios


rajadain:
Rewrite the work done for #3272 in #3291 to now show the Custom Weather button for editable scenarios.

We're switching custom weather to be uploadable per-scenario. This means that current conditions will _not_ have custom weather, but all editable scenarios will. This button is just another modification, and should live among the other BMPs:

![image](https://user-images.githubusercontent.com/1430060/79377522-d577b400-7f29-11ea-9cc7-667ecc23c050.png)

Run some names by Stroud to decide between calling it:

* Custom Weather
* Weather / Weather Data
* Climate / Climate Data

Issue #3287, opened 2020-04-12
Custom Weather: Add File Upload Dialog, Upload File


rajadain:
Add a dialog to upload custom weather data. This should trigger from the button added in #3298 .

This dialog should:

- [x] Allow all users to see if any custom weather data is currently applied, and download it
- [ ] Allow project owners to download a sample CSV of weather data
- [x] Allow project owners to upload new custom weather data
- [x] Allow project owners to switch the scenario to use default weather data
- [x] Allow project owners to permanently remove custom weather data
- ~~Allow project owners to switch the scenario to use future weather data if the AoI is within the DRB~~ Future Weather Data will be handled later.

PR #3303, opened 2020-05-01
Custom Weather: Rewrite Button for Scenarios

rajadain:
## Overview

Previously the button was associated with Current Conditions to affect the entire project. Now it is associated with editable scenarios to affect only that one.

Still uses the dummy modal. The actual modal is to be implemented in #3287.

Connects #3298 

## Testing Instructions

* Go to [:8000](http://localhost:8000/) and log in
* Create a MapShed project
    - [x] Ensure there's no Custom Weather button in Current Conditions
* Add a scenario
    - [x] Ensure you see the Custom Weather button


rajadain:
Thanks for taking a look!

Issue #3256, opened 2020-02-10
Switch to Different Basemap than Mapbox


rajadain:
Mapbox is deprecating Mapbox Classic styles in the spring:

> Ahead of deprecation, here are the two key dates that you should be aware of:
> 
> * March 31, 2020 - you will no longer be able to upload new Studio Classic styles or modify existing Studio Classic styles via the Uploads API
> * June 1, 2020 - you will no longer be able to make valid API requests using Studio Classic styles

They recommend switching to the Static Tiles API or Static Images API (both of which are **_significantly_**  more expensive than the Raster Tiles API we're currently using), or switching Leaflet out in favor of Mapbox GL JS, which is prohibitively complex and expensive given the extensive use of Leaflet in the app.

Given this, the only option we have is to switch away from Mapbox for the default basemap. Evaluate basemap alternatives and replace the current one with it. Try to find something visually similar, so as not to be too jarring a change.

Some initial places to look:

* https://www.arcgis.com/home/group.html?id=702026e41f6641fb85da88efe79dc166&view=list&categories=%5B%22%2FCategories%2FWeb%20Mercator%22%5D#content
* https://viewer.nationalmap.gov/help/3.0%20TNM%20Base%20Maps.htm
* https://leaflet-extras.github.io/leaflet-providers/preview/

rajadain:
The [USA Topo Map](https://www.arcgis.com/home/webmap/viewer.html?webmap=931d892ac7a843d7ba29d085e0433465) looks good from afar:

![image](https://user-images.githubusercontent.com/1430060/81005379-db233400-8e1b-11ea-81e5-e30aebd0012a.png)

But isn't really usable at higher zoom levels:

![image](https://user-images.githubusercontent.com/1430060/81005409-e9715000-8e1b-11ea-8e31-065caecb729d.png)

rajadain:
[Streets with Relief](https://www.arcgis.com/home/webmap/viewer.html?webmap=00f90f3f3c9141e4bea329679b257142) looks promising:

![image](https://user-images.githubusercontent.com/1430060/81005704-5c7ac680-8e1c-11ea-9c91-61539b91aa79.png)

I suspect it is using a combination of vector layer (for the streets) and raster (for the relief). Not sure if our Leaflet setup can handle vector layers.

The topographic features are not as prominent as USA Topo above. Furthermore, the map is less legible, both in terms of labels and topography, than our current MapBox:

[...]

PR #3304, opened 2020-05-06
Switch to ESRI World Topography as Default Basemap

rajadain:
## Overview

Since Mapbox is shutting down the classic tiles, and the new APIs are prohibitively expensive, we have to switch to a different default basemap provider. ESRI's World Topography Map is sophisticated, well maintained, and a worthy replacement.

Connects #3256 

### Demo

image

## Testing Instructions

    - [ ] Ensure you see the new basemap

### DIFF
--- a/src/mmw/mmw/settings/layer_settings.py
+++ b/src/mmw/mmw/settings/layer_settings.py
@@ -55,36 +55,14 @@ NHD_REGION2_PERIMETER = json.load(nhd_region2_simple_perimeter_file)
 LAYER_GROUPS = {
     'basemap': [
         {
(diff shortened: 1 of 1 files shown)


mmcfarland:
I'm getting that same behavior on `develop` so it doesn't appear to be related to this branch.

rajadain:
I think that's the case. I always get a Google Maps JavaScript API error in the console in local development:

image

rajadain:
Thanks for taking a look!

rajadain:
This has now been deployed to staging: https://staging.modelmywatershed.org/ and the Google layers are working correctly.

Issue #3315, opened 2020-05-16
HydroShare Export: Send Scenario-Specific GMS Files


rajadain:
Currently, we export the exact same GMS file for every scenario to HydroShare. This was introduced in 4c96a444960260b531940294ee18f1b37a575f60 as part of #2897 and has been in production since [1.24.0](https://github.com/WikiWatershed/model-my-watershed/releases/tag/1.24.0).

In that commit, we started sending the MapShed Job UUID from the client:

https://github.com/WikiWatershed/model-my-watershed/blob/c98fb88742928d035e6217db7c3e308646c05509/src/mmw/js/src/modeling/models.js#L769-L779

and then convert that UUID's job to GMS, before sending to HydroShare:

https://github.com/WikiWatershed/model-my-watershed/blob/c98fb88742928d035e6217db7c3e308646c05509/src/mmw/apps/export/tasks.py#L95-L108

However, this is the same for every scenario, as the UUID job is set on the project. The scenario's modifications are not accounted for.

---

Make it so that the GMS exported to HydroShare accounts for each scenario's modifications.

PR #3317, opened 2020-05-18
HydroShare Export: Fix GMS file export

rajadain:
## Overview

What 4c96a444 should have been.

Fixes GMS file exports so that each scenario's true GMS is exported to HydroShare. The `getModifiedGwlfeGisData` method was not available when the original commit was made, which does not excuse the mistake, but makes this fix much more straightforward.

Connects #3315

#### Setup HydroShare

* Populate `/etc/mmw.d/env/MMW_HYDROSHARE_SECRET_KEY` in your `app` and `worker` VMs from the "HydroShare OAuth MMW Beta" `client-secret` from LastPass
    - `vagrant ssh app -c 'sudo vim /etc/mmw.d/env/MMW_HYDROSHARE_SECRET_KEY'`
    - `vagrant ssh worker -c 'sudo vim /etc/mmw.d/env/MMW_HYDROSHARE_SECRET_KEY'`
* Restart the relevant services
    - `vagrant ssh app -c 'sudo service mmw-app restart'`
    - `vagrant ssh worker -c 'sudo service celeryd restart'`
* Go to [:8000/](http://localhost:8000/) and log in
* Go to your account and then Linked Accounts, unlink (if applicable) then relink to HydroShare, using the credentials in the same LastPass entry

#### Test the Export

* Create a MapShed project
* Add a new scenario and add some modifications (custom weather / land use changes / conservation practices / settings)
* Export the Current Conditions and New Scenario GMS files
    - [x] Ensure they are not identical
* Share the project to HydroShare. This can take some time, be patient.
* Once shared, go to the resource in HydroShare and download the GMS files
    - [x] Ensure they are identical to the MMW downloads

### DIFF
--- a/src/mmw/apps/export/tasks.py
+++ b/src/mmw/apps/export/tasks.py
@@ -14,7 +14,6 @@ from celery import shared_task
 from django.utils.timezone import now
 from django.contrib.gis.geos import GEOSGeometry
 
-from apps.core.models import Job
 from apps.modeling.models import Project
 from apps.modeling.tasks import to_gms_file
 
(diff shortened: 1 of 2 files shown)


rajadain:
Thank you for the review!

Issue #3312, opened 2020-05-14
Custom Weather: Bugfixes


rajadain:
* Seeing a strange `AWSAccessKey` suffix on staging:
  ![Screen Shot 2020-05-14 at 2 36 30 PM](https://user-images.githubusercontent.com/1430060/81973317-f9461c80-95f1-11ea-82cf-c462095f54cb.png)
  ![image](https://user-images.githubusercontent.com/1430060/82226561-a53e7f00-98f4-11ea-9015-7ec1fe6b727e.png)

mmcfarland:
This workflow can provide unexpected results:

1. Upload valid weather data file & save
2. Open CWD panel, delete/confirm the custom data
3. Without closing the panel, upload an invalid data file and receive error
4. Hit Cancel to close the panel
5. Open the panel again, the deleted data file is still listed 

Having confirmed delete, I would not expect the original file to still be associated with scenario.

![mmw-cdw-1](https://user-images.githubusercontent.com/1014341/82153452-7c0ae980-9835-11ea-84b8-5a0ad155cb03.gif)

rajadain:
The above is similar to the "cancel after delete" workflow error noted in #3309. My approach to fixing this is to remove the Save / Cancel button, replace it with a Done which simply closes the dialog, and always save on switching weather type and successful upload. Will implement for #3309.

rajadain:
All the above was handled in other PRs. No bugs were reported by the clients during testing on staging. Closing.

Issue #3309, opened 2020-05-13
Custom Weather: Usage Across Scenarios


rajadain:
When a scenario is duplicated, it's weather fields should be included in the duplication. This includes the `weather_custom` field. The duplicated scenario should have the exact same file used for its custom weather.

When deleting a custom weather file, check to see if it is also used in other scenarios. If so, then give the user a list of those scenarios and ask them to delete them before deleting the file.

When successfully deleting a custom weather file, switch back to default weather and save and close the dialog, so as not to leave the user in a state wherein they can "Cancel" after deleting custom weather and still hang on to the previous results.

PR #3319, opened 2020-05-18
Custom Weather: Usage Across Scenarios

rajadain:
## Overview

Adds support for duplicating scenarios with custom weather applied. The custom weather modification is copied over, and the same file is referenced.

During deletion, we check to see if the file is used by any other scenario. If so, we ask the user to either delete those scenarios or set them to use Available data and try again.

Also updates the UI to always save on valid states, and remove the Cancel / Save button in favor of a Done button.

Connects #3309

### Notes

This includes a fix for an error specified in #3312. I'm leaving that card open to gather feedback from clients, but wanted to get this in before giving it over to them for testing.

## Testing Instructions

* Go to [:8000/](http://localhost:8000) and create a new MapShed project
* Add changes to it. Rename "New Scenario" to "Original".
    image
* Open the Weather Data dialog
    image
  - [x] Ensure you see a Done button instead of Cancel / Save
* Add a valid custom weather file.
    image
  - [x] Ensure the scenario is updated immediately on successful save
* Click Done to dismiss the dialog. Duplicate the scenario.
    image
  - [x] Ensure there is a new scenario called "Copy of Original"
* Open its Weather Data
    image
  - [x] Ensure it is the same as Original
* Try to delete the weather file
    image
  - [ ] Ensure you are stopped from doing so
* Go back to Original and switch it to Available Data
    image
* Go back to Copy of Original and delete the file
    image
  - [x] Ensure there are no errors this time
* Go back to Original and open Weather Data
    image
  - [ ] Ensure no custom data file is shown


src/mmw/apps/modeling/views.py
```diff
@@ -231,6 +262,34 @@ def scenario_custom_weather_data(request, scen_id):
...
+                          .exclude(id=scenario.id)
+                          .filter(weather_type=WeatherType.CUSTOM,
+                                  weather_custom=scenario.weather_custom)
+                          .values_list('name', flat=True))
+
+                if others.count() > 0:
+                    errors.append('Cannot delete weather file.'
+                                  ' It is also used in: "{}".'
+                                  ' Either delete those scenarios, or set them'
+                                  ' to use available weather before deleting'
```
mmcfarland, line 276: Consider changing this to "to use Available Data" or "to use another available weather data set".  I think it may be helpful to use consistent  label terminology.

rajadain: Fixed in 6ceff3b

rajadain:
Thanks for reviewing!

Issue #3306, opened 2020-05-11
Shippensburg Land Projections: Update Layers in Selector


rajadain:
Once the new layers have been ingested (see #3305), update the Shippensburg layers in the Layer Selector:

![image](https://user-images.githubusercontent.com/1430060/81607328-82671480-93a2-11ea-8763-5c5ac10ee1d4.png)

- [ ] Remove DRB Land 2050 Business as Usual
- [ ] Replace DRB Land 2050 Corridors with DRB Land 2100 Corridors
- [ ] Replace DRB Land 2050 Centers with DRB Land 2100 Centers

rajadain:
Shares points with #3305.

Issue #3305, opened 2020-05-11
Shippensburg Land Projections: Ingest Layers for Visualization


rajadain:
Ingest `ProbabilityUrban_CentersOSI.tif` and `ProbabilityUrban_CorridorsOSI.tif` for visualization only. These layers look like this:

![image](https://user-images.githubusercontent.com/1430060/81607070-14225200-93a2-11ea-938c-71aa5ff3deaf.png)

(where the pink overlay is the DRB outline).

Use the same color ramp as used for past Shippensburg layers.

rajadain:
My current approach is to use GDAL instead of our usual GeoTrellis based pipeline for this conversion. After some trial and error, I've settled on this toolchain:

```bash
# Convert the input to 3857
gdalwarp -co TILED=YES -co COMPRESS=DEFLATE -t_srs EPSG:3857 ../"$NAME".tif "$NAME"_3857.tif
# Colorize with ramp
gdaldem color-relief -alpha -of PNG "$NAME"_3857.tif ../color-relief.txt "$NAME"_3857_colored.tif
# Create the tiles
gdal2tiles.py-3.8 -z 0-13 -r near -v -e --xyz "$NAME"_3857_colored.tif "$NAME-tiles"/
```

Where `color-relief.txt` is:

```
nv 255 255 255 0
0 255 255 255
0.13 246 226 226
0.26 237 196 196
0.39 227 167 167
0.52 218 137 137
0.65 209 107 107
0.78 199 78 78
0.9 190 50 50
1 183 28 28
```

This color ramp was made in QGIS using the same colors as used for the original Shippensburg layers. The first line makes NODATA values transparent.

On my local, this took 11m 11s to run this for one layer.

I'm going to make an instance in EC2 and do this there for the 6 layers.

rajadain:
All the layers have been ingested, and are now available on S3:

```shell
❯ aws s3 ls tiles.us-east-1.azavea.com/ | grep shippensburg
    PRE shippensburg-2050-baseline-30m/
    PRE shippensburg-2050-scenario1-30m/
    PRE shippensburg-2050-scenario2-30m/
    PRE shippensburg-2100-centers-30m/
    PRE shippensburg-2100-centers-np-30m/
    PRE shippensburg-2100-centers-osi-30m/
    PRE shippensburg-2100-corridors-30m/
    PRE shippensburg-2100-corridors-np-30m/
    PRE shippensburg-2100-corridors-osi-30m/
```

[...]

PR #3320, opened 2020-05-21
Shippensburg Land Projections: Update Layers

rajadain:
## Overview

Uses 4 of the new Shippensburg layers ingested in #3305 in the Layer Selector. Removes old layers.

Connects #3306

### Demo

Screen Shot 2020-05-29 at 11 55 55-fullpage

(the layer selector's height was increased for the purposes of this screenshot)

### Notes

There were a total of 6 layers of future land projections ingested:

```
❯ aws s3 ls tiles.us-east-1.azavea.com/ | grep shippensburg-2100
    PRE shippensburg-2100-centers-30m/
    PRE shippensburg-2100-centers-np-30m/
    PRE shippensburg-2100-centers-osi-30m/
    PRE shippensburg-2100-corridors-30m/
    PRE shippensburg-2100-corridors-np-30m/
    PRE shippensburg-2100-corridors-osi-30m/
```

We're using the vanilla and `-np-` ones here. Confirmed with Stroud that these are what they want to show.

## Testing Instructions

* Check out this branch
* Go to [:8000](http://localhost:8000/)
* Switch to the Coverage tab in the Layer Selector
    - [x] Ensure you don't see DRB Urban Land 2050 layers
    - [x] Ensure you do see DRB Urban Land 2011 Baseline layer
    - [x] Ensure you do see DRB Urban Land 2100 layers

### DIFF
--- a/src/mmw/mmw/settings/layer_settings.py
+++ b/src/mmw/mmw/settings/layer_settings.py
@@ -228,7 +228,7 @@ LAYER_GROUPS = {
             'big_cz': False,
         },
         {
-            'display': 'DRB Urban Land 2011 Baseline',
+            'display': 'DRB 2011 Urban Baseline',
             'code': 'urban-land-cover-2011-30m',
             'css_class_prefix': 'urban-land-cover-2011-30m',
             'short_display': '2011 Urban Land Cover',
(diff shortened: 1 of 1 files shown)


rajadain:
From Claire Jantz:

> I think it makes the most sense to put up these four:
> 
> centers
> centers-np
> corridors
> corridors-np
> 
> I’ll defer to Dave on if we want four up; if we only want to show two, then centers and corridors would be the ones.

From Dave Arscott:

> Please add the 4 layers that Claire mentioned. Sounds like you can easily turn the NP layers off if we decide at the last minute to exclude.

> here is my suggestion (trying to ensure no text wrapping in the Layers/CoverageGrid window - but probably just one character over wrap width on the longest version)
> * DRB 2011 Urban Baseline [yes, switching the year forward for emphasis, suggesting delete "land" to make room]
> * DRB 2100 Urban Center FX {abbreviation for forecast]
> * DRB 2100 Urban Center-NP FX
> * DRB 2100 Urban Corridor FX [perhaps abbreviate Corridor to Corr or Cor?]
> * DRB 2100 Urban Corridor-NP FX

rajadain:
This is now ready for review.

rajadain:
Thanks for taking a look!

Issue #3330, opened 2020-06-09
Future Weather: Include in HydroShare Export


rajadain:
Include the future weather dataset generated for a project in its HydroShare Export.

Issue #3329, opened 2020-06-09
Future Weather: Show Server Errors


rajadain:
Currently server side errors are sent back from the API but not shown in the UI. Make it so that server errors are shown in the UI.

Issue #3324, opened 2020-05-28
Future Weather: Incorporate 2080-2099 Projections


rajadain:
Following up from #3294, we now have a set of 22 CSVs for Future Weather projections for the 11 identified weather stations.

- [x] Find a place to store the information (S3 / Database / Somewhere Else)
- [x] Add the two scenarios (RCP 4.5 and RCP 8.5) to the Available Data section of the Weather Data dialog
- [x] Make the options selectable only if the area of interest is within the DRB
- [x] If the user selects it, find the two nearest weather stations and use their corresponding simulated data

rajadain:
I made a bucket in `mmw-prd` to host this data: `weather.modelmywatershed.org`. I chose to go there to indicate the runtime nature of this data, as opposed to `data.mmw.azavea.com` in `mmw-stg` which is used during setup. All its contents are public.

The URLs are so:

https://s3.amazonaws.com/weather.modelmywatershed.org/RCP45_2080_2100/13739.csv

The folder structure corresponds to the type of weather dataset to use. The file name corresponds to the ID of the weather station. The translation is such:

| Station ID | Station Name  | Source File Name                   |
|------------|---------------|------------------------------------|
| 14735      | Albany        | albanyRCP4520802100daily.csv       |
| 14737      | Allentown     | allentownRCP4520802100daily.csv    |
| 93730      | Atlantic City | atlanticCityRCP4520802100daily.csv |
| 4725       | Binghamton    | binghamtonRCP4520802100daily.csv   |
| 14751      | Harrisburg    | harrisburgRCP4520802100daily.csv   |
| 14734      | Newark        | newarkNJRCP4520802100daily.csv     |
| 94728      | New York City | NYCRCP4520802100daily.csv          |
| 13739      | Philadelphia  | phillyRCP4520802100daily.csv       |
| 14777      | WB/Scranton   | wilkesbarreRCP4520802100daily.csv  |
| 14778      | Williamsport  | williamsportRCP4520802100daily.csv |
| 13781      | Wilmington    | wilmingtonRCP4520802100daily.csv   |

rajadain:
The URLs were changed to reflect proper years, and the files were processed to have LF endings instead of CRLF, which caused problems with `requests.get().iter_lines()`

https://s3.amazonaws.com/weather.modelmywatershed.org/RCP45_2080_2099/13739.csv

PR #3331, opened 2020-06-09
Future Weather: Incorporate 2080-2099 Projections

rajadain:
## Overview

Adds two simulated future weather options to shapes within the DRB in the Weather Data dialog's Available Data section.

For each simulation, we have data for 11 stations used by shapes within the DRB. Outside the DRB this option is disabled. Inside the DRB we fetch the two nearest weather stations and fetch their simulations, then average them for the project. The generated data is sent back to the front-end, which then uses it as a modification, overriding the default values.

The simulation data is stored in a public S3 bucket and does not require AWS credentials to access.

There's a lot of changes in this PR, so review should focus more on functionality than code.

Connects #3324 

### Demo

With an inside-DRB shape:

With an outside-DRB shape:

image

### Notes

Although there are many places in the code where we report errors, these are not shown in the UI. We have #3329 for displaying them.

The generated weather data is not exported to HydroShare as a CSV file, although it is indirectly there within the GMS file. We have #3330 to generate a CSV file and export it to HydroShare.

## Testing Instructions

* Go to localhost:8000/ and log in
* Create a new project within the DRB
* Add a scenario
* Open the weather data dialog
  - [x] Ensure you see RCP 4.5 and RCP 8.5 options
  - [ ] Ensure selecting them updates the data accordingly
* Switch between Default and Simulated weather, between Custom and Simulated weather
  - [x] Ensure everything works as expected
* Set it to use Simulated Weather. Refresh the page to reload the project. Open the weather data dialog.
  - [x] Ensure it shows the correct selection
* Create a project outside the DRB
  - [ ] Ensure you cannot select a weather simulation for it

### DIFF
--- a/src/mmw/apps/modeling/models.py
+++ b/src/mmw/apps/modeling/models.py
@@ -101,6 +101,12 @@ class WeatherType:
     # e.g. within DRB.
     SIMULATION = 'SIMULATION'
 
+    # Valid Simulations currently supported in the app
+    simulations = [
+        'RCP45_2080_2099',
+        'RCP85_2080_2099',
+    ]
+
     # Custom, user uploaded, may be more accurate, more recent, or more
     # relevant than our default set. Will go through some validation.
     # Must be between 3 and 30 years of data.
--- a/src/mmw/apps/modeling/models.py
+++ b/src/mmw/apps/modeling/models.py
@@ -4,10 +4,14 @@ from __future__ import unicode_literals
 from __future__ import division
 
 from django.db.models import FileField
+from django.conf import settings
 from django.contrib.gis.db import models
 from django.contrib.auth.models import User
 
 from apps.core.models import Job
+
+
+DRB = settings.DRB_PERIMETER.buffer(0.1)
 
 
 def project_filename(project, filename):
@@ -77,6 +81,9 @@ class Project(models.Model):
 
     def __unicode__(self):
         return self.name
+
+    def in_drb(self):
+        return self.area_of_interest.within(DRB)
 
 
 class WeatherType:
--- a/src/mmw/apps/modeling/models.py
+++ b/src/mmw/apps/modeling/models.py
@@ -82,6 +82,7 @@ class Project(models.Model):
     def __unicode__(self):
         return self.name
 
+    @property
     def in_drb(self):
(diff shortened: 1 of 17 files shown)


src/mmw/apps/modeling/models.py
```diff
@@ -4,12 +4,16 @@
...
 
 from django.db.models import FileField
+from django.conf import settings
 from django.contrib.gis.db import models
 from django.contrib.auth.models import User
 
 from apps.core.models import Job
 
 
+DRB = settings.DRB_PERIMETER.buffer(0.1)
```
rajadain, line 14: I compared "Headwaters Pequea Creek", which is outside the DRB:

image

and the next-door "Upper West Branch Brandywine Creek", which is inside the DRB:

image

to get this number. I don't know if I should go lower.

src/mmw/apps/modeling/models.py
```diff
@@ -78,6 +82,10 @@ class Project(models.Model):
     def __unicode__(self):
         return self.name
 
+    @property
+    def in_drb(self):
+        return self.area_of_interest.within(DRB)
```
rajadain, line 87: At first I added this as a method, but then it became apparent that having this easy check in the front-end would be useful, so I made it into a `@property` and included it in the serializer. Not sure what performance impact this may have to calculate this for every serialization request. Maybe we can find a different way to do it if it would be severe.

src/mmw/apps/modeling/serializers.py
```diff
@@ -86,18 +86,20 @@ def validate_weather_type(self, value):
             raise ValidationError('Cannot use Custom Weather Data: '
                                   'none specified.')
 
-        if value == WeatherType.SIMULATION and \
-                not self.instance.weather_simulation:
-            raise ValidationError('Cannot use Simulation Weather Data: '
-                                  'none specified.')
```
rajadain, line 92: This validation check cannot be here because the user may set `weather_type = SIMULATION` and provide a value for `weather_simulation` in the same request. That would require payload-level validation rather than field-level validation. I have not looked in to how to do that.

rajadain:
Thanks for taking a look! Sorry about all the setup trouble!

Issue #3332, opened 2020-06-11
Shippensburg Land Projections: Add Front-end for Analysis


rajadain:
Following #3325, add more variants to the Land Analysis dropdown, as was done for Protected Lands in #3228. These dropdowns should be available only if the shape is within the DRB, and the analysis should happen only if the shape is within the DRB.

Of the six Shippensburg layers, use only `centers` and `corridors`. Not the `-osi-` or `-np-` variants.

Issue #3328, opened 2020-06-02
Shorten / Cleanup Layer Names


rajadain:
From Dave A:

> In an effort to "clean up" the naming conventions on the "Layers" tabs, can you change the naming convention for the following layers? Hoping this will reduce the need to scroll somewhat and provide a bit more simplicity....
> 
> - On Streams tab
> 	- Delaware River Basin TN Concentration from SRAT [shorten to: DRB TN conc. from SRAT]
> 	- Delaware River Basin TP Concentration From SRAT [shorten to: DRB TP conc. from SRAT]
> 	- Delaware River Basin TSS Concentration From SRAT  [shorten to: DRB TSS conc. from SRAT]
> - On Coverage Grid tab
> 	- DRB Catchment Water Quality Data TN Loading Rates from SRAT Catchments  [shorten to DRB TN Loading Rate from SRAT]
> 	- DRB Catchment Water Quality Data TP Loading Rates from SRAT Catchments  [shorten to: DRB TP Loading Rate from SRAT]
> 	- DRB Catchment Water Quality Data TSS Loading Rates from SRAT Catchments  [shorten to: DRB TSS Loading Rate from SRAT]

Also, remove the Shippensburg NP layers from the visualization.

rajadain:
Shares points with #3332

Issue #3325, opened 2020-05-28
Shippensburg Land Projections: Add Back-end for Analysis


rajadain:
Following #3323, write an Analyze back-end that queries the Drexel Fast Zonal API if the shape is within the DRB and sends back the results. The endpoint API should be identical to all the other analyze endpoints, so as to be easily used by the front-end.

rajadain:
Meeting with Lin Perez and the Drexel CCI tentatively on Friday 6/5 to discuss changes in API structure. We may investigate potentially getting one scenario back at a time, and see if that speeds things up a bit. Currently one request returns values for all 6 scenarios.

rajadain:
Further updates:

> This update is good to go. I’ve made it so that when you submit your payload, you must specify inside a JSON the geometry as well as a list of rasters. If you submit an empty list (i.e. []) then all six of the rasters will be run. Also, if you don’t submit a “rasters” parameter at all then all six will be returned. I can change this if you would prefer a different behavior. Below is a list of the acceptable raster names. If any name in what was submitted is not recognized, we will return an error.
> 
> Rasters = [“centers”, “corridors”, “centers_np”, “corridors_np”, “centers_osi”, “corridors_osi”]
> 
> If you include a raster name that is not within this list, you will just receive this error message and no other output:
> 
> “At least one of the rasters submitted is not supported. Please enter in any number of only these names ['centers', 'corridors', 'centers_np', 'corridors_np', 'centers_osi', 'corridors_osi']”
>  
> This does change the format of the request. Now, instead of just submitting a geojson you should submit a JSON with the geojson as well as the raster files of interest:
> 
> ```yaml
> {
>     “geom”: {“type”: ””, “coordinates”: []},
>     “rasters”: []
> }
> ```
> 
> EXAMPLE REQUEST:
> 
> ```yaml
> {
>     "geom": {  "type": "Polygon",  "coordinates": [[  [-75.328551114,42.0893891182],  [-75.2980940131,42.0922444714],  [-75.2809618939,42.0532213109],  [-75.3019011507,42.0437034669],  [-75.3209368388,42.0437034669],  [-75.3314064672,42.0484623889],  [-75.34282788,42.0579802329],  [-75.3485385865,42.0779677054],  [-75.3485385865,42.0874855494],  [-75.328551114,42.0893891182]  ]]}
>     ,"rasters": ["centers", "corridors"]
> }
> ```
> 
> ```
[...]

Issue #3323, opened 2020-05-27
Shippensburg Land Projections: Investigate Drexel / ANS API


rajadain:
From Mike Campagna at ANS:

> ### DOCUMENTATION:
> http://watersheds.cci.drexel.edu/docs#/Fast%20Zonal%20Statistics/post_api_fzs_buildout_
> 
> ### ENDPOINT / CURL REQUEST & EXAMPLE PAYLOAD:
> ```
> curl --location --request POST 'http://watersheds.cci.drexel.edu/api/fzs_buildout/' --data-raw '{ "type": "MultiPolygon", "coordinates": [ [ [ [ -75.32956412136285, 41.648657839645097 ], [ -75.328937630586921, 41.6485529894574 ], [ -75.328937630586921, 41.6485529894574 ], [ -75.329061897528447, 41.648098743510282 ], [ -75.329698780964492, 41.64820226267841 ], [ -75.32956412136285, 41.648657839645097 ] ] ] ] }'
> ```
> 
> ### INPUT PAYLOAD:
> A GeoJsonPolygon or GeoJsonMultiPolygon. Below is an example of a complex GeoJsonMultiPolygon with holes that will return a valid output if placed into the above curl request.
> 
> ```yaml
> {
>   "type": "MultiPolygon",
>       "coordinates": [
>           [ [ [ -75.180214049407581, 40.141181982438425 ],
>                   [ -75.180904553654415, 40.141301927039351 ],
>                       [ -75.180825285990366, 40.141552097757135 ],
>                       [ -75.180143885352564, 40.141437734484931 ],
>                       [ -75.180214049407581, 40.141181982438425 ] ] ],
>                   [ [ [ -75.179911175633862, 40.142225582917121 ],
>                   [ -75.179977358689214, 40.141965108996658 ],
>                       [ -75.18033371139191, 40.142033025151903 ],
>                       [ -75.180259564292101, 40.142284055491253 ],
>                       [ -75.179911175633862, 40.142225582917121 ] ] ],
>                   [ [ [ -75.18206769189301, 40.142315928881644 ],
>                   [ -75.182217693303457, 40.141789834122285 ],
>                       [ -75.182868360375181, 40.141899027545428 ],
>                       [ -75.182713240108242, 40.142424263325346 ],
>                       [ -75.18206769189301, 40.142315928881644 ] ] ],
>                   [ [ [ -75.183569012827562, 40.143386664471521 ],
[...]

PR #3334, opened 2020-06-21
Shippensburg DRB 2100 Land Analysis

rajadain:
## Overview

Adds analysis of Shippensburg 2100 Land Forecast layers. This is done by querying the Fast Zonal API developed by Drexel / ANS, investigated in #3323. The query is wrapped in the standard Celery API used for all other analyses, and is identical from the client perspective. This API is publicly available in the MMW Public API.

The analysis is DRB specific, and is only available in areas of interest within the DRB. This is the first analyze endpoint to have such a limitation, and is validated upon request, returning an HTTP 400 if the requirement is not met. The front-end only queries these endpoints for AOIs within the DRB. The DRB check is also standardized between the front- and back-ends.

Some layer name cleanup is also done, as requested.

### TODO

- [x] Figure out what to put in the **Source** line for the analysis (emailed Dave, waiting for response)
- [x] Ensure that the API description is correct (emailed Dave, waiting for response)
- [x] Ensure that, in the case of Drexel / ANS API failure, the other analyses (NLCD, Protected Lands) are not affected
- [x] Add all possible rasters to the allowed rasters list.

Connects #3325 
Connects #3332 
Connects #3328 

### Demo

image

image

## Testing Instructions

* Check out this branch and `bundle` and restart your Celery service to pick up new tasks `vagrant ssh worker -c 'sudo service celeryd restart'`
* Go to [:8000/](http://localhost:8000/) and pick a shape that is obviously in the DRB
  - [x] Ensure you see the two DRB analyses under the Land tab
* Change the shape, and choose something obviously not in the DRB
  - [x] Ensure you don't see those two options under Land analysis
* Change the shape, and choose something on the border, such as Upper Mahanoy Creek
  - [x] Ensure you don't see those two options under Land analysis
* Edit `geoprocessing_api/calcs.py` and add the following line:
    ```diff
    diff --git a/src/mmw/apps/geoprocessing_api/calcs.py b/src/mmw/apps/geoprocessing_api/calcs.py
    index c6163d3d..f88ed3a9 100644
    --- a/src/mmw/apps/geoprocessing_api/calcs.py
    +++ b/src/mmw/apps/geoprocessing_api/calcs.py
    @@ -340,6 +340,8 @@ def drexel_fast_zonal(geojson, dataset):
     
         res = requests.post(settings.DREXEL_FAST_ZONAL_API['url'], data=payload)
     
    +    raise Exception('boom')
    +
         # Raise any non-success responses as exceptions, indicating failure
         res.raise_for_status()
    ```
* Restart Celery to pick up the changes `vagrant ssh worker -c 'sudo service celeryd restart'`
* Pick a shape within the DRB
  - [ ] Ensure that while the DRB checks fail, the other two analyses render correctly


src/mmw/apps/geoprocessing_api/calcs.py
```diff
@@ -332,3 +333,17 @@ def streams_for_huc12s(huc12s, drb=False):
         cursor.execute(sql, [tuple(huc12s)])
 
         return [row[0] for row in cursor.fetchall()]  # List of GeoJSON strings
+
+
+def drexel_fast_zonal(geojson, dataset):
+    payload = '{{"geom": {}, "rasters": ["{}"]}}'.format(geojson, dataset)
```
rajadain, line 339: This looks a little awkward, but is done to avoid converting geojson (which is a stringified JSON object) to a dict and back to a string for the request payload.

src/mmw/apps/geoprocessing_api/schemas.py
```diff
@@ -69,6 +78,17 @@
     }
 )
 
+DRB_2100_LAND_ERROR_RESPONSE = Schema(
+    title='Future Land Error Response',
```
rajadain, line 82: Should say "DRB 2100 Land Error Response"

src/mmw/apps/geoprocessing_api/tasks.py
```diff
@@ -343,6 +344,35 @@ def analyze_protected_lands(result, area_of_interest=None):
     }
 
 
+@shared_task
+def analyze_drb_2100_land(area_of_interest, dataset):
```
rajadain, line 348: Implementation similar to `analyze_land` and `analyze_protected_lands`

src/mmw/apps/geoprocessing_api/views.py
```diff
@@ -1149,6 +1151,193 @@ def start_analyze_protected_lands(request, format=None):
...
+@decorators.permission_classes((IsAuthenticated, ))
+@decorators.throttle_classes([BurstRateThrottle, SustainedRateThrottle])
+@log_request
+def start_analyze_drb_2100_land(request, dataset=None, format=None):
+    """
+    Starts a job to get a land-use histogram for an area within DRB in 2100.
+
+    Uses simulations of land use in the year 2100 based on two scenarios of
+    growth: centers and corridors. Generated by Shippensburg University,
+    serviced via APIs by Drexel University and the Academy of Natural Sciences.
```
rajadain, line 1172: Need to confirm this text is okay for the Public MMW API.

src/mmw/apps/geoprocessing_api/views.py
```diff
@@ -1149,6 +1151,193 @@ def start_analyze_protected_lands(request, format=None):
...
+
+    errs = []
+    if not dataset:
+        errs.append('`dataset` must be specified')
+    if dataset not in ['centers', 'corridors']:
+        errs.append('`dataset` must be one of ["centers", "corridors"]')
+
+    # A little redundant since GeoJSON -> GEOSGeometry is already done once
+    # within _parse_input, but it is not returned from there and changing
+    # that API could break many things.
```
rajadain, line 1323: Three-line comments is probably a code smell for things being too complicated

src/mmw/apps/modeling/models.py
```diff
@@ -84,7 +81,7 @@ def __unicode__(self):
 
     @property
     def in_drb(self):
-        return self.area_of_interest.within(DRB)
+        return self.area_of_interest.within(settings.DRB_SIMPLE_PERIMETER)
```
rajadain, line 84: Went with this which is actually a superior approximation:

image

src/mmw/js/src/analyze/models.js
```diff
@@ -239,6 +239,20 @@ function createAnalyzeTaskGroupCollection(aoi, wkaoi) {
...
+                    area_of_interest: aoi,
+                    wkaoi: wkaoi,
+                    taskName: "analyze/drb-2100-land/centers"
+                },
+                {
+                    name: "drb_2100_land_corridors",
+                    displayName: "DRB 2100 land forecast (Corridors)",
+                    area_of_interest: aoi,
+                    wkaoi: wkaoi,
+                    taskName: "analyze/drb-2100-land/corridors"
```
rajadain, line 254: Pretty cool that path parameters can be added so easily

src/mmw/js/src/analyze/views.js
```diff
@@ -1252,7 +1252,10 @@ var ChartView = Marionette.ItemView.extend({
 
     getBarClass: function(item) {
         var name = this.model.get('name');
-        if (name === 'land') {
+        if (['land',
+             'drb_2100_land_centers',
+             'drb_2100_land_corridors'].indexOf(name) > -1)
```
rajadain, line 1257: Since we want to use the same color distribution in the chart for all three.

src/mmw/js/src/core/utils.js
```diff
@@ -563,13 +563,30 @@ var utils = {
...
+                return _.flatten(_.flatten(geom.coordinates));
+            case 'Polygon':
+                return _.flatten(geom.coordinates);
+            default:
+                throw new Error('Unsupported geometry');
+        }
     },
 
+    // Check if a geom is completely within DRB_SIMPLE_PERIMETER
+    isInDrb: _.memoize(function(geom) {
```
rajadain, line 579: Memoized for efficiency.

`isInDrb` was previously only used to send Google Analytics, so this slight change in the behavior (of not counting partial intersections) should not break anything.

src/mmw/apps/geoprocessing_api/views.py
rajadain: Sent an email to Dave and Claire, waiting to hear back.

src/mmw/apps/geoprocessing_api/calcs.py
```diff
@@ -332,3 +333,17 @@ def streams_for_huc12s(huc12s, drb=False):
         cursor.execute(sql, [tuple(huc12s)])
 
         return [row[0] for row in cursor.fetchall()]  # List of GeoJSON strings
+
+
+def drexel_fast_zonal(geojson, dataset):
```
mmcfarland, line 338: It may be valuable to add a small docstring comment here to indicate that `dataset` is a dataset name. You may also consider changing to `dataset_name`, but the variable is used extensively and can probably be cleared up by a comment.

mmcfarland:
Re: the Analyze API calls during modeling, that only happened when I forced the exception in `calcs`. This is potentially related to the memoizing function? Regardless, it may be a side effect that we can live with - it doesn't seem to have an impact and when the API is working, the extra calls are not made.

src/mmw/mmw/settings/layer_settings.py
```diff
@@ -259,22 +259,6 @@
             'legend_unit_breaks': [0, '', 20, '', 40, '', 60, '', 80, '', 100],
             'big_cz': False,
         },
-        {
-            'display': 'DRB 2100 Urban Centers-NP FX',
```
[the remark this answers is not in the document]
rajadain: Correct. This is stated (but not expounded upon) in #3328. Basically they wanted the NP layers for a specific audience, but worried that it could confuse the larger audience, so weren't sure if they were going to keep it. In one of our recent update calls, it seemed like the specific audience can use the Drexel API directly, and thus does not need this included in MMW.

rajadain:
Hmm the Analyze results should _not_ be called during modeling. I'll take a quick look.

rajadain:
Fixed a lint issue in 640af180

rajadain:
Text updated from Dave in d55f8090 and fb2e1c39

rajadain:
Changed `dataset` to `key` in 3db6a384

rajadain:
Fixed an issue in 51eb7cb6 where, when on a Land Analysis that was still fetching results, when the results would arrive, the dropdown text would update (say from "(Loading) DRB 2100 ..." to "DRB 2100 ...") but the body of the tab would remain stuck on "Analyzing" until the dropdown was switched to another option.

rajadain:
I was not able to replicate the analyze calls being fired during modeling issue.

rajadain:
This is ready for another quick look, if you have time.

rajadain:
Ah thanks for that detailed demo. It is probably a rogue `fetchAnalysisIfNeeded` call somewhere, which refetches those that did not complete successfully. They should not be called on every scenario creation though. I'll take a quick look to see if I can fix that, else make a follow up card for it.

rajadain:
The refetch is fixed in 8db64e8:



See commit message for details, but essentially now we don't refetch if something is in "failed" or "complete" with results states.

Also added a long personal wish of mine -- better messaging for polling results in 9c0f9f7e:

image

The `429` in the screenshot above are throttled requests. The burst limit is increased in b2643b17 to allow them.

rajadain:
Let me know if you have thoughts on the above, but I don't think it requires a formal re-review. Going to squash the fixups now, will merge once green.

Issue #3343, opened 2020-07-16
Manual Entry: Modifications Not Saved Correctly


rajadain:
## Overview

Open a Mapshed project. Add some Land Cover or Settings changes to it. Wait for the model to run and the results to be saved, so that there are no more spinners in the UI. Refresh the page.

## Expected Behavior

Saved results load once.

## Actual Behavior

Results are recalculated after reloading.

## Additional Context

This behavior does not affect numbers, just results in extra API calls.

This is possible because `inputmod_hash` is not being correctly updated when the results save the first time, although they are saved on reload, and this doesn't happen on further reloads.

Issue #3333, opened 2020-06-11
Shippensburg Land Projections: Add to Land Cover Modification Dialog


rajadain:
Add a dropdown to select a "preset" land distribution to the Land Cover Modification dialog:

<img width="1552" alt="image" src="https://user-images.githubusercontent.com/1430060/84425131-14a34800-abef-11ea-8298-bf86ff6ad941.png">

There should be two options:

* DRB 2100 Urban Centers FX
* DRB 2100 Urban Corridors FX

These options should only be available within the DRB.

Add a back-end to query this. It may be possible to reuse the Analyze back-end introduced in #3325, even possible to reuse those results without having to query the back-end.

Selecting a preset should fill in the text boxes. The dialog should remember what preset is being used. If the text boxes are changed manually, the preset should switch to "Custom" or something, away from the established list.

Discuss with @designmatty what's the best place to add that dropdown in the dialog.

rajadain:
Spoke with @designmatty, here are the main takeaways of that conversation:

The dropdown should go in the modal body, not in the header or footer. Being at the top, before all the fields start, is probably best.

Selecting an item from the dropdown should update the base / placeholder values in the text fields, and should still allow the user to manually update them.

The selected value from the dropdown should be saved and remembered.

rajadain:
The Land Cover data should be used as the "base layer" for all other modifications, such as Conservation Practices.

rajadain:
I've gotten this to the point where the UI is now setup. Next is the matter of how to actually implement the functionality. Here are some notes:

> The next step is to add a SPECIAL modification when this preset changes. That SPECIAL modification should be shifted upon add to ALWAYS be the FIRST one in the list.
> 
> Then, the Land Cover Modal should be updated to MERGE the SPECIAL modification on load, and on preset change.
> 
> Later on, this MERGING of the SPECIAL modification can be generalized, so that when other modifications such as Conservation Practices are being added, they respect the SPECIAL modification as well.
> 
> Furthermore, if modifications are applied in sequence, then the Land Cover modifications can be respected by others like the Conservation Practices as well.

rajadain:
Thinking about the SPECIAL modification, but trying to generalize it for a larger implementation, here is one design, inspired by how Operating Systems work:

![image](https://user-images.githubusercontent.com/1430060/87214925-0ab05a00-c2ff-11ea-9c94-9e09c937c7db.png)

[...]

Issue #3159, opened 2019-08-27
Manual Entry: Land Cover Changes Should Affect Entire MapShed Dictionary


rajadain:
As of #3151, we update certain keys in the MapShed dictionary which are derived from the `Area` list, whose values can be updated by users using the Land Cover modal. However, conceptually, there are a number of other fields that should also be affected:

> Despite being computed from `Area`, the value of `LS` is not recalculated, because it also relies on the `lu_stream_pct` geoprocessing result. This is not currently not saved, and thus cannot be used during recalculation. Thus, this field is left to its initially calculated value.
> 
> A number of other geoprocessing operations combine NLCD with another raster to get a grouped count, such as `nlcd_soil`, `nlcd_streams`, `nlcd_slope`, and `nlcd_kfactor`. These are in turn used to calculate a number of results, such as `cn`, `ag_stream_pct`, `low_urban_stream_pct`, `med_high_urban_stream_pct`, `lu_stream_pct`, `ag_slope_3_pct`, `ag_slope_3_8_pct`, `n41`, `kf`, and `avg_kf`. These are in turn used to calculate a number of MapShed fields, such as `AgLength`, `UrbLength`, `n42`, `n46e`, `n46f`, `CN`, `AgSlope3`, `AgSlope3To8`, `n41`, `AvKF`, `KF`, and `P`.
> 
> Recalculating these fields from the modifications is complicated, because the modifications are only changing the proportion of land use distribution, not the actual distribution in space. Without the spatial updates, we cannot run raster operations. This will need to be clarified and addressed at a later date.

Updating these fields may require a spatial representation of modifications. It may be possible to simply change the geoprocessing service results in proportion to the land use changes made in the front-end, but that remains to be seen.

In any case, we'll have to save the geoprocessing results to the project. Currently they are discarded after creating the initial MapShed dictionary, which is the only thing that is saved to the database. Having the source results available would allow us to modify them on a per-scenario basis.

It is possible that the increase in accuracy this would provide is diminishing in value, and may not be worth the effort of implementing this. However, in case it is, the context are recorded herein.

Issue #3158, opened 2019-08-27
Manual Entry: Update Modifications on Land Use Change


rajadain:
A follow up to #3151.

Consider the case when a user changes some Land Use values that affect Conservation Practices templates:

![localhost_8000_project_13_scenario_17_](https://user-images.githubusercontent.com/1430060/63364645-07f1f280-c344-11e9-8e64-e6e71affea01.png)

![localhost_8000_project_13_scenario_17_ (1)](https://user-images.githubusercontent.com/1430060/63364656-0cb6a680-c344-11e9-80ad-e760cad87760.png)

However, the user can go back and further change the land use value, even reset it to original, and the conservation practice modification persists:

![localhost_8000_project_13_scenario_17_ (2)](https://user-images.githubusercontent.com/1430060/63364708-22c46700-c344-11e9-87a1-7b7a381610a8.png)

In this case, it has an invalid value because it is trying to change 8 our of 4.1 hectares of row crops.

These are some ideas on how to handle this:

* Do nothing (current behavior)
  - Easiest to implement
  - Puts onus on user to remember which ones are valid and which are not
* Remove all conservation practices whenever Land Use is updated
  - Possibly straightforward thanks to being indiscriminate
  - May be too aggressive an approach
* Remove select conservation practices that are affected by land use changes
  - Still pretty straightforward, as the list of conservation practices is set at design time
  - May still be too aggressive
* Remove only those conservation practices which are now invalid
  - Difficult, as we'd have to come up with validation methods and string them through the system
  - Experience would be slightly better for the user, but deleting automatically may still be disorienting
* Highlight those conservation practices which are now invalid
  - Even more difficult, as we'd have to develop a visual system of indication
  - Best for users
* Automatically scale modifications to maintain proportions
  - If a certain modification was made to convert 80% of a certain land use, scale the absolute value so that it stays 80% of the new land use
  - May require some training to set user expectations

Consider the options and implement the appropriate one.

PR #3344, opened 2020-07-16
Shippensburg Land Projections: Add to Land Cover Modification Dialog

rajadain:
## Overview

Adds a dropdown to the Land Cover Modification Dialog to auto-populate the fields with values from the selected Land Cover analysis results. This takes advantage of the current implementation, with minor edits to populate the values and remember which preset was used.

The connecting issue mentions two implementation details that are deferred for later:

* Using the preset values to populate the gray "auto values" rather than the black "user values"
  - This layered modification will likely require something akin to the [ring approach](https://github.com/WikiWatershed/model-my-watershed/issues/3333#issuecomment-656975031), which was too complex to implement here
* Having these values affect other modifications, such as conservation practices
  - These are known drawbacks of the current implementation (see #3158 and #3159), and would be too complex to implement here

Connects #3333 

### Notes

Instead of adding another endpoint to fetch the values from the Drexel API again, I chose to reuse the values already fetched during analysis. This creates a new dependency on Analyze results, saves a lot of code.

The area total for the DRB numbers does not match exactly (because the NLCD numbers come from our own geoprocessing service and the DRB numbers from from Drexel's fast zonal API). I could not come up with a way to resolve this discrepancy, and is left to the user to manage.

Once you save Land Cover modifications, wait for the results to save, and refresh the page, it calculates the results again. This is existing behavior, logged in #3343.

The code for getting the appropriately modified GMS for a scenario is currently spread across many functions, and needs to be unified. I'll make a follow up `enhancement` card for that.

## Testing Instructions

* Check out this PR and `bundle`
* Go to [:8000/](http://localhost:8000/) and select a shape within the DRB
* Create a Mapshed project with it. Add changes to the area.
* Open the Land Cover Modification dialog
  - [x] Ensure you see a preset dropdown at the top
* Select a preset
  - [x] Ensure values are appropriately filled out for the preset, and that the total is updated
* Ensure the total matches, then click "Save"
  - [x] Ensure the scenario results are recalculated with the new updates
* Open the Land Cover Modification dialog again
  - [x] Ensure the land cover edits are persisted
  - [x] Ensure the preset selects the right dropdown
* Select the default preset
  - [x] Ensure all fields are emptied
* Save and open the Land Cover Modification dialog again
  - [x] Ensure the scenario results are recalculated with the new updates
  - [x] Ensure the land cover edits are persisted
  - [x] Ensure the preset selects the right dropdown

### DIFF
--- a/src/mmw/apps/modeling/mapshed/calcs.py
+++ b/src/mmw/apps/modeling/mapshed/calcs.py
@@ -562,6 +562,8 @@ def landuse_pcts(n_count):
     Given a dictionary mapping NLCD Codes to counts of cells, returns an
     array mapping MapShed Land Use Types (as the index) to percent
     area covered.
+
+    In sync with js/src/modeling/utils.js:nlcdToMapshedLandCover
     """
 
     total = sum(n_count.values())
--- a/src/mmw/js/src/modeling/controls.js
+++ b/src/mmw/js/src/modeling/controls.js
@@ -504,9 +504,9 @@ var GwlfeLandCoverView = ControlView.extend({
 
         entryViews.showLandCoverModal(
             this.model.get('dataModel'),
-            currentScenario.get('modifications'),
-            this.addModification,
-            App.currentProject.get('in_drb')
+            currentScenario,
+            App.currentProject.get('in_drb'),
+            App.getAnalyzeCollection()
         );
     },
 });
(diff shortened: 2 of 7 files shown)


rajadain:
Thanks for taking a look! I added #3345 to add the DRB test polygon as a visualizable layer, to help debug this both for us and the users, since there is now much DRB-specific stuff in the app.

rajadain:
Created #3346, #3347, and #3348 for further work along these lines.

Issue #3351, opened 2020-07-17
Add Custom / Future Weather to Compare View


rajadain:
Currently, a change in the weather data has the most impact on the curves in the Compare View:

![image](https://user-images.githubusercontent.com/1430060/87830255-ab969c00-c84e-11ea-8ea9-ed81d2685a5a.png)

However they are not shown in the list of modifications when hovering on a scenario title. Add it to the dropdown.

Issue #3350, opened 2020-07-17
Add Display Names for Settings Modifications for Compare View


rajadain:
Currently, for every modification added via the Settings menu, there is no name, and the underlying key is shown raw:

![image](https://user-images.githubusercontent.com/1430060/87830255-ab969c00-c84e-11ea-8ea9-ed81d2685a5a.png)

Add display names for each of these and handle them appropriately.

Issue #3349, opened 2020-07-17
Compare View Crashes on Scenarios with Land Cover Preset


rajadain:
https://staging.modelmywatershed.org/project/1484/scenario/2642

![image](https://user-images.githubusercontent.com/1430060/87808887-80995180-c828-11ea-9378-a954bca1ffbe.png)

PR #3352, opened 2020-07-17
Fix Compare View Crash

rajadain:
## Overview

Fixes Compare View crashes when a scenario has modifications that don't have display names. The previous implementation only considered the Conservation Practices in the Compare View, but since now we also have a number of Manual Entry modifications, they need to be handled explicitly.

If a display name is not found for a modification, we fallback to its key instead. This prevents crashes in the present. Names for those modifications will be added in the future #3350. Also, weather data is currently not represented in this popup, this will be handled in #3351.

Connects #3349 

### Demo

image

## Testing Instructions

* Create a Mapshed project
* To a new scenario, add conservation practices, land cover modifications, and some changes to the "Settings" menu
* Open the compare view
  - [x] Ensure it does not crash
  - [x] Ensure all modifications listed herein are shown correctly

### DIFF
--- a/src/mmw/js/src/compare/templates/compareModificationsPopover.html
+++ b/src/mmw/js/src/compare/templates/compareModificationsPopover.html
@@ -20,7 +20,7 @@
 
 {% macro gwlfeModificationsTable(caption, mods) %}
 <table class="compare-modifications-table -gwlfe">
-{% if modifications.length > 0 %}
+{% if mods.length > 0 %}
     <h5>{{ caption }}</h5>
 {% endif %}
     <tbody>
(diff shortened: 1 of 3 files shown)


mmcfarland:
Build failure is related to http://nodejs.org/ being down.

src/mmw/js/src/compare/templates/compareModificationsPopover.html
```diff
@@ -20,7 +20,7 @@ <h5>{{ caption }}</h5>
 
 {% macro gwlfeModificationsTable(caption, mods) %}
 <table class="compare-modifications-table -gwlfe">
-{% if modifications.length > 0 %}
```
mmcfarland, line 23: Curious, was this just an existing undiagnosed but?

rajadain: Yeah 😬

rajadain:
Thanks for taking a look! The site is back up and CI is passing now.

PR #3354, opened 2020-07-20
Shippensburg Land Projections: Proportion land cover distribution to match

rajadain:
## Overview

The raw land cover numbers from the Drexel API don't match our internal NLCD 2011 numbers, and were slightly off, resulting in the user having to manually reconcile in the front-end.

Since these are future projections, the individual raw numbers are not the main feature, rather the distribution as a whole. By attuning it to the existing total, we ensure that the new distribution still reflects the forecasts but adds up to the same total, resulting in a more fluid experience.

Connects #3333 

## Testing Instructions

* Check out this PR and `bundle`
* Go to [:8000/](http://localhost:8000/) and select a shape within the DRB
* Create a Mapshed project with it. Add changes to the area.
* Open the Land Cover Modification dialog
* Select a preset
  - [ ] Ensure you don't have to manually adjust the values to get the total to match

### DIFF
--- a/src/mmw/js/src/modeling/gwlfe/entry/views.js
+++ b/src/mmw/js/src/modeling/gwlfe/entry/views.js
@@ -135,7 +135,14 @@ var LandCoverModal = modalViews.ModalBaseView.extend({
                                 acc[category.nlcd] = category.area;
                                 return acc;
                             }, {}),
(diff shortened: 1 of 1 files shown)


rajadain:
Thanks for taking a look!

Issue #3358, opened 2020-07-23
SSO and cookie issue


dboyer:
Reported by Concord Consortium:

We discovered an issue logging in to MMW when it is embedded in LARA and Chrome's new SameSite cookie policy is enabled.

You can learn more about the SameSite cookie policy change here: https://www.chromium.org/updates/same-site
This article might be a better read: https://blog.heroku.com/chrome-changes-samesite-cookie

The gist is that Chrome is gradually rolling out a more strict cookie policy. By default this policy prevents cookies from "3rd party" sites.
When MMW is embedded in LARA, MMW is considered a 3rd party site because it is in an iframe of a different domain.
Because MMW doesn't include properties in it cookies to override this default, chrome will not store cookies sent by MMW when it is in an iframe.

rajadain:
Timeframe:

**July 14, 2020**: SameSite cookie enforcement has resumed, with a gradual rollout starting today (July 14) and ramping up over the next several weeks as we continue to monitor overall ecosystem readiness and engage with websites and services to ensure they are prepared for the SameSite labeling policy. The SameSite features are being enabled for Chrome Stable channel users on versions 80 and 81 (who should update Chrome!), 83, as well as the newly released 84.

So we may need to work on this sooner than later.

rajadain:
Use this site to test: http://concord-consortium.github.io/lara-interactive-api/

PR #3360, opened 2020-07-24
Add, configure django-cookies-samesite

rajadain:
## Overview

Recent versions of Chrome require cookies be set with explicit Cross Site policies. Since MMW is embedded within the LARA educational system, logging in requires cookies to be tagged with "SameSite=None; Secure".

This configuration is [available out of the box in Django 2.1+](https://docs.djangoproject.com/en/2.1/ref/settings/#std:setting-SESSION_COOKIE_SAMESITE), in order to use it in Django 1.11 we add the [`django-cookies-samesite`](https://pypi.org/project/django-cookies-samesite/) package.

This configuration can cause issues with some older browsers, a full list of which can be seen here:

https://www.chromium.org/updates/same-site/incompatible-clients

However, since our official policy is to support the two most recent versions of popular browsers, all of which implement this functionality correctly, we shouldn't have to worry about them.

Browser support data: https://caniuse.com/#feat=mdn-http_headers_set-cookie_samesite_none

image

Notably, IE11 is missing from this list.

Connects #3358 

### Demo

image

### Preparation

* Ensure you have an account and saved projects in both local and staging
* Using Chrome, go to chrome://flags, search for "same site", and enable both options:
    image

### Testing the Bug

* In Chrome, go to https://concord-consortium.github.io/lara-interactive-api/
* In the iframe box's URL, put in https://staging.modelmywatershed.org/
* Open the developer tool's network tab
* Log in to your account
* Inspect the network response for POST login
  - [x] Ensure it sets a cookie without "Same Site=None; Secure" tags
* Click on "Projects" to load your list of projects
  - [x] Ensure the list doesn't load, and returns a 403 in the network tab

### Testing the Fix

* Check out this branch
* Run `ngrok`
* In Chrome, go to https://concord-consortium.github.io/lara-interactive-api/
* In the iframe box's URL, put in the ngrok URL
* Open the developer tool's network tab
* Log in to your account
* Inspect the network response for POST login
  - [x] Ensure it sets a cookie with "Same Site=None; Secure" tags
* Click on "Projects" to load your list of projects
  - [x] Ensure the list does load, and returns a 200 in the network tab

### DIFF
--- a/src/mmw/mmw/settings/base.py
+++ b/src/mmw/mmw/settings/base.py
@@ -254,6 +254,8 @@ TEMPLATES = [
 # MIDDLEWARE CONFIGURATION
 # See: https://docs.djangoproject.com/en/1.11/topics/http/middleware/
 MIDDLEWARE = (
+    # Django Cookies Samesite Middleware must be first
+    'django_cookies_samesite.middleware.CookiesSameSite',
     # Default Django middleware.
     'corsheaders.middleware.CorsMiddleware',
     'django.middleware.common.CommonMiddleware',
@@ -298,6 +300,10 @@ THIRD_PARTY_APPS = (
 )
 
 # THIRD-PARTY CONFIGURATION
+
+# django-cookies-samesite
+SESSION_COOKIE_SAMESITE = 'None'  # Allows for cross site embedding into LARA
+SESSION_COOKIE_SECURE = True      # Only set cookies in HTTPS connections
 
 # rest_framework
 REST_FRAMEWORK = {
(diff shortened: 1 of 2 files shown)


rajadain:
Thanks for taking a look! Luckily I had just done this recently when trying to use MMW as a back-end for https://github.com/rajadain/research-react-hooks.

scytacki:
@rajadain thanks for fixing this.

Safari only supports this in version 13 and only with OS X 10.15 (most recent version).  So anyone with an older mac or iPad that hasn't upgraded and chooses to use Safari will have problems. The bug in Safari about this is here: https://bugs.webkit.org/show_bug.cgi?id=198181

It might be necessary to support these old versions of Safari by removing the SameSite=none in that case. I'll follow up via an email with the team.

Issue #3361, opened 2020-07-27
Disable Recent Weather in 1.30.0


rajadain:
And enable it in staging afterwards.

From the clients:

> * The NLDAS data may present some problems for users in comparing with other climate data (this is a modeled product that is very different from the 1960-1990 actual weather data
we may ask that this does not get deployed at this point...how difficult is it to turn that off/remove from what is deployed, but leave it on staging for us to further interrogate??

PR #3362, opened 2020-07-27
Disable Recent Weather

rajadain:
## Overview

The NLDAS dataset is modeled, not observed, which may need further explanation to differentiate from Default Weather.

Will be enabled on staging after this for further consideration.

Connects #3361 

### Demo

image

## Testing Instructions

* Open the weather data dialog
  - [ ] Ensure you don't see Recent Weather in the Available Data dropdown


rajadain:
Thanks for taking a look, and catching the other issue!

Issue #3359, opened 2020-07-23
Update TR-55 Colors to be more Accessible


rajadain:
Using color recommendations decided in https://github.com/WikiWatershed/mmw-micro/issues/18

designmatty:
Infiltration: #f2cf46;
    Evapotranspiration: #2d8bff;
    Runoff: #ed3b3b;

rajadain:
Should be noted that the colors look different between Firefox and Chrome:

![image](https://user-images.githubusercontent.com/1430060/88565404-8bab6900-d002-11ea-8d78-2c514299b1c2.png)

This is a well known issue in the industry.

PR #3366, opened 2020-07-30
Replace TR-55 Colors with Accessible Variants

rajadain:
## Overview

These colors work better under color accessibility tests, and can serve a more inclusive set of users. The change is required primarily for the runoff site. MMW colors are updated correspondingly for consistency.

Connects #3359
Connects https://github.com/WikiWatershed/mmw-micro/issues/18

### Demo

Screen Shot 2020-07-27 at 11 33 42 AM

Screen Shot 2020-07-27 at 11 33 54 AM

Screen Shot 2020-07-27 at 11 34 10 AM

### Notes

The above screenshots are taken in Firefox, which has notably more vibrant colors than Chrome on macOS. This is a [known issue](https://stackoverflow.com/a/6338124/6995854). The new colors are accessible on both browsers.

## Testing Instructions

* Create a TR-55 project
  - [ ] Ensure the colors are different, and match the work done in https://github.com/WikiWatershed/mmw-micro/pull/27
* Open the compare view
  - [ ] Ensure colors are updated there too


rajadain:
> I'm assuming the Water Quality charts don't need to be changed because it is a single, non stacked color?

Correct! Thanks for taking a look. Merging this now.

Issue #3068, opened 2018-12-27
Follow up on Empty AoI Projects


rajadain:
From #3064

TODO: Add recommendations for how to proceed.

rajadain:
This list of projects with no area of interest, taken from a backup before the last production deployment on 12/18/2018, indicates that the only two projects affected were those addressed in #3064:

```sql
SELECT id, name, area_of_interest_name, modified_at
FROM modeling_project
WHERE area_of_interest IS NULL
ORDER BY modified_at DESC;

  id   |                       name                       |                     area_of_interest_name                      |          modified_at
-------+--------------------------------------------------+----------------------------------------------------------------+-------------------------------
 25753 | Untitled Project                                 | Selected Area                                                  | 2018-11-14 18:45:09.520279+00
 25660 | WEST GROVE BOROUGH                               | Selected Area                                                  | 2018-11-07 19:09:05.256515+00
 25588 | modle my water shed                              | 1 Square Km                                                    | 2018-11-06 14:27:20.506396+00
 25556 | Earth Space project                              | 1 Square Km                                                    | 2018-11-04 21:11:43.079216+00
 24780 | FSNHP_RockyRunLower_SiteStorm                    | Delaware High Resolution                                       | 2018-10-03 14:17:32.920611+00
 25362 | sandy run site storm                             | Delaware High Resolution                                       | 2018-10-02 15:02:55.558739+00
 25204 | Honeybrook Phase I and I planning                | Continental US Medium Resolution                               | 2018-09-07 16:29:43.962262+00
 24789 | New Britain HS demo                              | Selected Area                                                  | 2018-07-20 17:35:44.870013+00
 21885 | Untitled Project                                 | Lower Conestoga River, HUC-12 Subwatershed                     | 2018-07-12 21:07:57.368595+00
 18602 | FiveMile Creek                                   | Fivemile Creek, HUC-12 Subwatershed                            | 2018-06-22 18:38:26.805671+00
 18309 | Spanish Needle                                   | Spanish Needle Creek-Upper Macoupin Creek, HUC-12 Subwatershed | 2018-06-22 18:37:12.351288+00
 18337 | Bullard Lake                                     | Bullard Lake-Middle Macoupin Creek, HUC-12 Subwatershed        | 2018-06-22 18:35:40.461002+00
[...]

PR #3376, opened 2021-01-06
Prevent Empty AoIs

rajadain:
## Overview

For some time there has been a mysterious behavior that empties out the areas of interest in projects. We have been unable to pinpoint the cause of this. The work in this PR attempts at adding guards to prevent this from happening in the future.

The manner in which most people interact with projects is via the MMW UI, since project manipulation is not available in the public API. Given the size and complexity of MMW's UI, it is possible there are some edge cases which empty out the area of interest, which then gets saved, causing this. To prevent this, we add a validator to the Project Update Serializer, which is the only way an existing project is updated via the UI, to validate that the AoI is not empty.

In case this is happening via other means, we also add a database constraint to ensure that the area of interest cannot be NULL. This will ensure that no matter what updates the project: Django or direct SQL, this will always be respected. In order for this constraint to be applied, any existing projects with NULL areas of interest will have to be deleted. We will export those projects from production before deploying this update.

An exemption is made for Activities, which are Projects created via ITSI mode. These Projects, marked by `is_activity=True`, are allowed to have empty AoIs. As far as we know, these Activities were never used in production, but they did weaken the protections around Area of Interest in b81edefa, laying the ground for this bug.

Connects #3068 

### Preparation

Before you check out this PR, ensure some behaviors on `develop`:

#### Empty AoI Project

* Open MMW on `develop` and log in
* Create a new project (TR-55 may be simpler, but this would apply to both TR-55 and GWLF-E projects)
* Open the network tab of developer tools
* Rename the project to be "Empty AoI"
* Find that PUT request in the network tab, and copy it as a `curl` command
* Paste that in a text editor, find `area_of_interest` key in the request body, and set its value to `null`
* Run that curl command with `area_of_interest: null` set
    - [x] Ensure the command runs successfully, and the project's AoI is deleted

#### Activity Project

* Create another new project, and note its id (referenced later as `ACTIVITY_PROJECT_ID`)
* In your terminal, run `./scripts/manage.sh shell_plus`
* In the Django shell, run the following:
    ```python
    a = Project.objects.get(pk=ACTIVITY_PROJECT_ID)
    a.is_activity = True
    a.area_of_interest = None
    a.save()
    ```

### Testing

* Check out this branch
* Run the `curl` command with the empty AoI again
  - [x] Ensure you get a 400 with a validation error
* Run migrations `./scripts/manage.sh migrate`
  - [x] Ensure the Empty AoI Project has been deleted
  - [x] Ensure the Activity Project is still there
* Run `shell_plus` and select any project other than the Activity Project
    ```python
    p = Project.objects.get(pk=NON_ACTIVITY_PROJECT_ID)
    ```
* Try and set its AoI to NULL
    ```python
    p.area_of_interest = None
    p.save()
    ```
    - [x] Ensure you get an error
* Select the activity project, and try to mark it as a non-activity
    ```python
    a = Project.objects.get(pk=ACTIVITY_PROJECT_ID)
    a.is_activity = False
    a.save()
    ```
    - [x] Ensure you get an error


rajadain:
Ah requiring non-empty AoIs are breaking a lot of tests, which use mock projects with no AoIs. They'll have to be updated for this to pass. Updates coming soon.

src/mmw/apps/modeling/serializers.py
```diff
@@ -180,6 +180,19 @@ def create(self, validated_data):
...
+    def validate(self, data):
+        """
+        Ensure that all non-activity projects have a legitimate AoI.
+        """
+        is_activity = data.get('is_activity', False)
+
+        if not is_activity:
+            # Validate that either AoI or WKAoI is specified correctly
+            serializer = AoiSerializer(data=data)
+            serializer.is_valid(raise_exception=True)
```
mmcfarland, line 192: Will this exception be logged by default with Rollbar? If not, could it be caught and reported? It seems likely, but not certain, that the rogue `null` AoIs are coming in through a malformed request, and it would be convenient to eventually track that down as well.

mmcfarland (commented):
Tested the functionality, which behaves correctly with regard to empty AoIs, respecting `Activity` types. However, for a traditional TR55 Square Km project, it fails to save when creating. I receive a `400 {"non_field_errors":["Area of interest must be valid GeoJSON, of type Feature, Polygon or MultiPolygon"]}`  from the POST to `/projects`.

The AOI appears to be a valid MultiPolygon in this case. I get the same error for a hand drawn shape and the MapShed model.

rajadain:
The TR-55 project breaking should be fixed in a9a03e8.

src/mmw/apps/modeling/serializers.py
rajadain: I'm going to try and get this on staging and then send the malformed request to see if it comes up on staging. If not, I'll add an explicit log.

rajadain:
Just tried this on staging. Did get a 400 with:

```
{"non_field_errors":["Must supply either the area of interest (GeoJSON), or a WKAoI ID."]}
```


But no Rollbar error, likely because this was caught by the 400 handler. I'll add an explicit Rollbar line.

rajadain:
Added Rollbar reporting in 8c738b7. Deploying that to staging right now, will report once tested there.

rajadain:
Tested on staging, this is generating the right Rollbar message: https://rollbar.com/WikiWatershed/ModelMyWatershed/items/117/

And the request details are available in Rollbar for inspection:

image

rajadain:
Rerunning the PR job, it failed on an email step, not related to the changes in this PR.

rajadain:
Alright the re-check succeeded, this is ready for a final review! 🙏

rajadain:
Mdae #3377 for releasing this, noting that data needs to be exported before the migrations are applied.

mmcfarland (approved):
Looks good :+1: , project saving is restored and empty AoIs error with reporting. Nice job.

rajadain:

Issue #3381, opened 2021-02-20
Disable Recent Weather


rajadain:
Exactly as #3362. We do not want the NLDAS work on production.

PR #3382, opened 2021-02-22
Disable Recent Weather

rajadain:
## Overview

Functionally identical to #3362.

This is a dataset that should not be visible on production. Once merged, we'll do another deploy to production.

Connects #3381


fungjj92 (approved):
Confirmed: diff exactly matches that in the referenced PR

rajadain:
Thanks for taking look!

PR #3384, opened 2021-02-25
Update Drexel Fast Zonal API URL to HTTPS

rajadain:
## Overview

Drexel has added HTTPS functionality to their endpoint. While the existing HTTP endpoint currently forwards to the HTTPS endpoint, we change it officially here.

### Demo

image

### Note

Even though the code is setup to use `MMW_DREXEL_FAST_ZONAL_API_URL` from the URL, that environment variable is not defined in `default.yml`, thus the fallback value in the code is what ends up being used.

## Testing Instructions

* Check out this branch
* Go to [:8000/](http://localhost:8000/) and pick / draw any shape
* In the Analyze view, go to the Land tab, then from the dropdown select "DRB 2100 land forecasts (Centers)"
  - [x] Ensure it finishes successfully and has results

### DIFF
--- a/src/mmw/mmw/settings/base.py
+++ b/src/mmw/mmw/settings/base.py
@@ -459,7 +459,7 @@ SRAT_CATCHMENT_API = {
 # Drexel Fast Zonal API Settings
 DREXEL_FAST_ZONAL_API = {
     'url': environ.get('MMW_DREXEL_FAST_ZONAL_API_URL',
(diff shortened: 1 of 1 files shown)


jwalgran:
Looking at this now.

jwalgran (approved):
Confirmed that the data loaded correctly after the switch to https

Screen Shot 2021-03-01 at 5 14 02 PM

rajadain:
Thanks for reviewing!

Issue #3398, opened 2021-07-01
Don't Run All Analyses By Default


rajadain:
We have a number of analyze options available now, and not all of these are required for all users. To lessen the server burden and optimize the usage for all users, only run a subset of analyses by default, and run the rest when they are selected / focused by the user.

Issue #3397, opened 2021-06-30
Ingest All NLCD 2019 Products


rajadain:
These were launched in June, after we had ingested the 2016 products. There is a new version of all years in that ZIP, so re-ingest 2001, 2006, 2011, 2016, and the new 2019.

rajadain:
I've made a `provision.sh` file to encapsulate all the tools needed to setup a c4.2xl EC2 instance with 200GB EBS volume for ingesting:

```bash
#!/bin/bash

set -ex

# Install dependencies and setup Docker
sudo apt update
sudo apt install -y docker.io jq awscli unzip make openjdk-8-jdk-headless
sudo usermod -aG docker ubuntu

# Setup sbt
mkdir -p ~/.sbt/launchers/0.13.15
pushd .sbt/launchers/0.13.15
wget https://scala.jfrog.io/artifactory/ivy-releases/org.scala-sbt/sbt-launch/0.13.15/sbt-launch.jar
popd

# Setup Scala and Spark
mkdir ~/tools/
pushd tools
wget https://downloads.lightbend.com/scala/2.11.12/scala-2.11.12.tgz
tar xvzf scala-2.11.12.tgz
wget https://apache.osuosl.org/spark/spark-2.4.8/spark-2.4.8-bin-hadoop2.7.tgz
tar xvzf spark-2.4.8-bin-hadoop2.7.tgz
popd

cat << EOF > .profile
if [ -d "$HOME/tools/scala-2.11.12/bin" ] ; then
    PATH="$HOME/tools/scala-2.11.12/bin:$HOME/tools/spark-2.4.8-bin-hadoop2.7/bin:$PATH"
fi
EOF

source ~/.profile

# Setup ETL
mkdir ~/dev/
git clone git@github.com:azavea/civic-apps-etl.git ~/dev/ca-etl
pushd dev/ca-etl
git checkout tt/nlcd-2016-ingest
make mmw-create-jar
popd

# Download source file
wget https://s3-us-west-2.amazonaws.com/mrlc/NLCD_landcover_2019_release_all_files_20210604.zip

# Configure AWS
aws --profile=azavea-datahub configure
```

After this, one must log out and log back in for the Docker group change to take affect.

Then, in a tmux session:

```console
$ tmux new -s ingest
```

Run `ingest-all.sh`:

```bash
#!/bin/bash

set -ex

for YEAR in 2019 2016 2011 2006 2001
do
	./ingest.sh $YEAR
done
[...]

Issue #3395, opened 2021-05-12
Add New Analyze Views for New NLCD Layers


rajadain:
- [x] Add endpoints for new NLCD 2016 and new NLCD 2011 endpoints
- [x] Add analyze views in the sidebar that use those endpoints, and reference appropriate visual layers
- [x] Make the new NLCD ~2016~ 2019 the default one
- [x] Rename all NLCD options together in the layer selection dropdown according to https://github.com/WikiWatershed/model-my-watershed/pull/3396#issuecomment-870886945

Issue #2745, opened 2018-03-29
Save analyze results with projects & don't re-run


kellyi:
Currently when loading a saved project the saved modeling results are displayed and therefor the modeling job request isn't rerun. However, the analysis results don't seem to be saved -- even if they are, it doesn't seem to prevent the analysis job network requests from rerunning.

To replicate this: save a mapshed project, then refresh it and verify that the saved modeling results show up without having to fetch them again from the API and also that the app makes requests to `/api/analyze` for...

- climate
- streams
- land
- soil
- terrain
- point source
- animals

Unless we intentionally don't save these results for some reason, we should save the analyze results with projects and make it so we don't poll for analyze results if they are already saved.

PR #3399, opened 2021-07-01
Add Analyze Views for 2019 NLCD Products

rajadain:
## Overview

Adds Analyze Views for 2019, 2016, 2011, 2006, and 2001 years of NLCD from the NLCD 2019 product ingested in #3397. Also renames the old NLCD 2011 layer to also be available.

Connects #3395 

### Notes

This adds 5 additional analyses that are kicked off on _every_ shape drawn / selected, which could increase our CPU and Network usage significantly. This may exacerbate the need for optimizations to the analyze view, such as #2745 and #3398. Those efforts are noted in separate cards for future prioritization.

This only changes the Analyze views. All other instances that use old NLCD 2011 (such as TR-55 Modeling, etc) still use it, and will be switched in future PRs.

## Testing Instructions

* Check out this branch and `bundle --debug` and `vagrant ssh worker -c 'sudo service celeryd restart'`
* Go to [:8000/](http://localhost:8000/) and pick / draw a shape
* Select the Land tab in the Analyze sidebar
  - [x] Ensure you default to 2019 NLCD
  - [x] Ensure there are options available for 2016, 2011, 2006, 2001, and old 2011
  - [x] Ensure each is appropriately described and connects to the correct visual layer
* Make a Site Storm Model project
  - [x] Ensure it is created successfully
* Make a Watershed Multi Year Model project
  - [x] Ensure it is created successfully
* Go to [:8000/api/docs](http://localhost:8000/api/docs)
  - [x] Ensure the Analyze Land API is described appropriately

### DIFF
--- a/src/mmw/apps/geoprocessing_api/urls.py
+++ b/src/mmw/apps/geoprocessing_api/urls.py
@@ -14,7 +14,7 @@ app_name = 'geoprocessing_api'
 urlpatterns = [
     url(r'^token/', views.get_auth_token,
         name="authtoken"),
-    url(r'analyze/land/(?P<nlcd_year>\w*)/?$', views.start_analyze_land,
+    url(r'analyze/land/(?P<nlcd_year>\w+)/?$', views.start_analyze_land,
         name='start_analyze_land'),
     url(r'analyze/soil/$', views.start_analyze_soil,
         name='start_analyze_soil'),
--- a/src/mmw/apps/geoprocessing_api/urls.py
+++ b/src/mmw/apps/geoprocessing_api/urls.py
@@ -14,7 +14,7 @@ app_name = 'geoprocessing_api'
 urlpatterns = [
     url(r'^token/', views.get_auth_token,
         name="authtoken"),
-    url(r'analyze/land/$', views.start_analyze_land,
+    url(r'analyze/land/(?P<nlcd_year>\w*)/?$', views.start_analyze_land,
         name='start_analyze_land'),
     url(r'analyze/soil/$', views.start_analyze_soil,
         name='start_analyze_soil'),
--- a/doc/MMW_API_landproperties_demo.ipynb
+++ b/doc/MMW_API_landproperties_demo.ipynb
@@ -260,7 +260,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "#### Issue job request: **analyze/land/**"
+    "#### Issue job request: **analyze/land/2011_2011/**"
    ]
   },
   {
@@ -269,7 +269,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "result = analyze_api_request('land', s, api_url, json_payload)"
+    "result = analyze_api_request('land/2011_2011', s, api_url, json_payload)"
    ]
   },
   {
--- a/src/mmw/apps/geoprocessing_api/schemas.py
+++ b/src/mmw/apps/geoprocessing_api/schemas.py
@@ -11,6 +11,21 @@ from drf_yasg.openapi import (
 
 from django.conf import settings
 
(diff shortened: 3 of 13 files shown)


src/mmw/apps/geoprocessing_api/urls.py
```diff
@@ -14,7 +14,7 @@
 urlpatterns = [
     url(r'^token/', views.get_auth_token,
         name="authtoken"),
-    url(r'analyze/land/$', views.start_analyze_land,
+    url(r'analyze/land/(?P<nlcd_year>\w*)/?$', views.start_analyze_land,
```
jwalgran: I see in commit bb107ae that we documented this breaking change to the API. I do not know much about how API consumers we have. Is is worth considering setting up a 302 from `analyze/land/` to `analyze/land/2011_2011/`? Alternatively, is there a design reason that is making us choose a path parameter over a backwards compatible optional query string parameter (for example `analyze/land/?nlcd_year=2019_2016`)?

jwalgran (approved):
I confirmed all the functionality. Looks good. The small, separate commits made the stack of changes easy to follow.

Changing years does update the graphs properly but I noticed two workflow issues that we may want to consider addressing

The NLCD years are not sorted chronologically



When you turn on the layer for a year, then change years, the table shows data from the new year, but the map shows the tiles from the previous year, which has some potential for confusion. 



In addition I saw that there was a broken markdown link in the new documentation.

Screen Shot 2021-07-06 at 2 28 13 PM

rajadain:
Broken links fixed in d854ca9. Added #3401 to polish that further.

rajadain:
> The NLCD years are not sorted chronologically

They are sorted in the order specified in https://github.com/WikiWatershed/model-my-watershed/pull/3396#issuecomment-870886945

image

All the NLCD19 items are together, sorted chronologically, then we the old NLCD11 item.

rajadain:
> When you turn on the layer for a year, then change years, the table shows data from the new year, but the map shows the tiles from the previous year, which has some potential for confusion.

Nice catch. Made #3402 to work on that.

src/mmw/apps/geoprocessing_api/urls.py
rajadain: I went with the path parameter to ensure it is explicitly specified every time, and because there's precedent for this with the DRB 2100 endpoint:

https://github.com/WikiWatershed/model-my-watershed/blob/ae94ac6020b08d6163e43f99cfde35b9b413346c/src/mmw/apps/geoprocessing_api/urls.py#L36-L38

While falling back to 2011 would preserve backwards compatibility, as of this work we'll be switching to 2019 as the "default" dataset for all future work. We _could_ have the default be 2019, and then while no API users will have their systems break, but their values will be different. This ensures that any existing users will get a breaking notification, causing them to update their endpoints mindfully.

Looking at API usage:

```sql
SELECT num_api_calls, username, first_name, last_name, email, organization, user_type, country, postal_code
FROM auth_user u
    INNER JOIN user_userprofile p ON u.id = p.user_id
    INNER JOIN (
        SELECT user_id, COUNT(id) AS num_api_calls
        FROM core_requestlog
        WHERE user_id IS NOT NULL
          AND user_id <> 7305 -- MMW Web App User
          AND api = 't'
          AND requested_at > (CURRENT_DATE - INTERVAL '12 months')
        GROUP BY user_id
    ) x ON u.id = x.user_id
ORDER BY num_api_calls DESC;
```

I see only one API user in the last year, and they just explored for a week in June (not sharing further output of the query for privacy reasons). Given the low API usage and lack of clarity on what would be a good default, I decided to introduce this breaking change without a fallback.

rajadain:
Thanks for the thoughtful feedback! I've created follow up issues for some of these tasks. Going to merge this now.

Issue #3400, opened 2021-07-06
Add Multi Layer Support to Projects


rajadain:
## Overview

* [x] Add a `layers` JSON field to `modeling_project` which will contain a dictionary of layer names to be overwritten
* [x] Initialize the above field to `{"__LAND__": "nlcd-2011-30m-epsg5070-512-int8"}` for all existing rows, but use `{"__LAND__": "nlcd-2019-30m-epsg5070-512-byte"}` as the default value going forward
* [x] Add this to the project serializer and send it to the front-end
* [x] In the front-end, include this in all `model_input` parameters given to TR-55 and MapShed jobs
* [x] Modify [TR-55](https://github.com/WikiWatershed/model-my-watershed/blob/d78510a7f31c71389b8bea6aad502f8ce5d3622b/src/mmw/apps/modeling/views.py#L599) and [MapShed](https://github.com/WikiWatershed/model-my-watershed/blob/d78510a7f31c71389b8bea6aad502f8ce5d3622b/src/mmw/apps/modeling/views.py#L539) to read this dictionary from the `model_input`

## Previously

Currently, the data layers queried for a project are hardcoded into the system. As of #3397 we now have multiple NLCD layers to pick from. We must now add a mechanism to specify which layers a project is using, and adjust the geoprocessing to get the layer name from the project before querying it. We must also set all existing projects to use the 2011 (NLCD11) layer, and all new projects to default to 2019 (NLCD19) layer.

When there is only one layer for a project, the design is simple. If folks want to compare different layers per scenario, then it gets more complicated:

- We could have scenario level layers, that would have to be sandwiched between project level base data, and the scenario-specific modifications
- We could handle different layers as modifications
  - This is easier in Mapshed, where we already have a Land Cover Modification dialog
  - Harder in TR-55, where modifications are spatially relevant, can't just use aggregations

rajadain:
At first I thought that TR-55 would be the exceptional case, but since we run the entire TR-55 model on every run, it's pretty simple to provide it with a different input layer.

The harder question may be how this is handled for MapShed. Currently, Land Cover changes are applied as modifications, but this is not perfect, as it leads to issues like #3346 and #3347. If we were to change the land cover changes from being applied as modifications, that may require a larger refactor.

rajadain:
Since the current contract has been budgeted and scoped with the simplest, most immediate solution in mind, that is all we can implement at this time. We will have to defer addressing the related issues above to a later time.

[...]

PR #3396, opened 2021-06-28
Add NLCD 2016 Edition Visual Layers

rajadain:
## Overview

Add the four new NLCD 2016 version visual layers to the Coverage Tab of the Layer Selector.

These were ingested via GDAL on EC2, using the script specified here: https://github.com/WikiWatershed/model-my-watershed/issues/3393#issuecomment-866179490

The old layer was renamed "National Land Cover Database Old 2011". /cc @darscott @SRGDamia1 @aufdenkampe for confirmation if this is appropriate, or if we should call it something else.

The layer selector was widened a bit to accommodate the lengthier layer names.

Connects #3394 

### Demo

The NLCD Analyze connects to the Old 2011 layer, this will be amended in #3395 

image

## Testing Instructions

* Check out this branch and `./scripts/bundle.sh --debug --vendor`
* Go to [:8000/](http://localhost:8000/)
* Switch the layer selector to the Coverage (second) tab
  - [x] Ensure you see 4 new NLCD layers, and they work correctly
  - [x] Ensure the old NLCD layer is renamed
* Choose a shape to analyze
  - [x] Ensure the Land Analysis still uses the Old 2011 layer, and it turns on from the sidebar correctly

PR #3411, opened 2021-07-29
Allow Projects to Override Layers for Geoprocessing

rajadain:
## Overview

Previously, every layer class (land, soil, etc.) had a single raster layer for it. As of recent work done in #3396 and #3399, we now have allow multiple raster layers for ever class of raster. The actual raster used is specified in the settings, can be overridden, as of #3406.

This PR adds override capability to Projects, in the form of a `layer_overrides` JSON field. The value of this field will be used to override the default layer configuration on every geoprocessing run. For all existing projects, this is set to override the land layer with NLCD 2011. For all new projects, no overrides are set. The front-end is updated to read this value from the project and use it when firing model runs, and the back-end is updated to expect this value in the model runs and use it. Finally, the default land layer is set to NLCD 2019.

This allows us to preserve the work done in all existing projects, thus not changing results from under our users without their awareness, while still using the latest data for all work going forward.

Connects #3400 

## Testing Instructions

Much of the work in this PR is to preserve existing data. Thus, in order to test this, you must create local saved work on `develop`, then checkout this branch and perform the upgrades, then create some new work, and ensure that the old work is preserved.

To compare old and new calculations, we'll use the same Area of Interest in both cases.

* Start on the latest `develop`
* [x] Ensure you locally have a user and are logged in
  - In case registering through the app, follow the Django log to see the confirmation link in the email:
    ```console
    $ vagrant ssh app -c 'sudo journalctl -fu mmw-app'
    ```
* [x] Find an Area of Interest that has noticeable changes between 2011 and 2019
  - I used the CONUS Land Cover Change Index layer in https://www.mrlc.gov/viewer/ to find Flat Rock Creek near Kenbridge, VA that works well. You can search for Kenbridge in the geocoder, go there, then click "Select Boundary", "HUC-12", then "Flat Rock Creek":
    
* Once analysis is done, go to the Model tab and select "Site Storm Model". Since you are logged in, this project will be saved to the database. Rename the project to be "TR55 Old 2011"
* Under "Details", click "Edit" next to Model to go back one step to Analyze
  image
* Next, select the "Watershed Multi-Year Model". Rename this project to be "MapShed Old 2011"

Now that our reference projects are setup, we can start testing the work in this branch.

* Checkout this branch and run `./scripts/manage.sh migrate` and `./scripts/bundle.sh --debug`
* Clear your geoprocessing cache `vagrant ssh services -c 'redis-cli -n 1 --raw KEYS ":1:geop_*" | xargs redis-cli -n 1 DEL'`
* Restart Celery `vagrant ssh worker -c 'sudo service celeryd restart'`
* Go to http://localhost:8000/ and open the Network tab of developer tools
* Select the same Area of Interest and create a Site Storm Model project named "TR55 New 2019"
  - [x] Notice that the payload for POST http://localhost:8000/mmw/modeling/tr55/ contains a `layer_overrides: {}` key
    image
* With the same Area of Interest, create a Watershed Multi-Year Model project named "MapShed New 2019"
  - [x] Ensure the payload for POST http://localhost:8000/mmw/modeling/mapshed/ contains a `layer_overrides: {}` key
* Go to your projects (from the top-right), and open the old projects (perhaps in new tabs)
* Compare the output numbers in the tables between the old and new projects
  - [x] Ensure they are different
* In the TR55 Old 2011 project, with the Network tab open, create a new Scenario by clicking "Add changes to this project"
* In the New Scenario, change the Precipitation slider, to re-run the TR-55 model
* In the Network tab, notice that the payload for POST http://localhost:8000/mmw/modeling/tr55/ contains a `layer_overrides: {"__LAND__":"nlcd-2011-30m-epsg5070-512-int8"}` key
  image
* Inspect the database with `./scripts/manage.sh dbshell`
  - [x] Ensure that old projects have the 2011 override set, and the new ones don't:
    ```sql
    SELECT id, name, layer_overrides FROM modeling_project;

     id |       name       |                 layer_overrides                 
    ----+------------------+-------------------------------------------------
      1 | TR55 Old 2011    | {"__LAND__": "nlcd-2011-30m-epsg5070-512-int8"}
      2 | MapShed Old 2011 | {"__LAND__": "nlcd-2011-30m-epsg5070-512-int8"}
      9 | TR55 New 2019    | {}
     10 | MapShed New 2019 | {}
    (4 rows)
    ```


src/mmw/apps/modeling/migrations/0035_project_layer_overrides.py
```diff
@@ -0,0 +1,22 @@
...
+    dependencies = [
+        ('modeling', '0034_forbid_null_aoi_on_non_activity_projects'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='project',
+            name='layer_overrides',
+            field=django.contrib.postgres.fields.jsonb.JSONField(default={'__LAND__': 'nlcd-2011-30m-epsg5070-512-int8'}, help_text='JSON object of layers to override defaults with'),
+            preserve_default=False,
```
rajadain, line 20: This ensures that the 2011 override is used for all existing projects, but no longer.

src/mmw/apps/modeling/migrations/0036_no_overrides_by_default.py
```diff
@@ -0,0 +1,21 @@
...
+
+    dependencies = [
+        ('modeling', '0035_project_layer_overrides'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='project',
+            name='layer_overrides',
+            field=django.contrib.postgres.fields.jsonb.JSONField(default=dict, help_text='JSON object of layers to override defaults with'),
```
rajadain, line 19: This ensures that the default is an empty dictionary, effectively no overrides, for all new projects.

rajadain (commented):
Left some comments. The rest of the diff is just wiring `layer_overrides` through all the necessary parts.

jwalgran:
I have started reviewing this, but am switching to pair on another project.

jwalgran:
When setting up the test on the `develop` branch I saw some 500 errors in the network tab related to weather stations.

Screen Shot 2021-07-30 at 2 02 41 PM

Because the error is a missing Postgres function I am assuming that my database VM may be outdated. I don't know if having a broken weather data endpoint will affect the testing of this feature.

rajadain:
Hmm, `jsonb` was added in Postgres 9.4, and `to_jsonb` was added in 9.5.

It's possible that reprovisioning your `services` VM is enough:

```console
$ vagrant reload services --provision
```

But in case that doesn't work, you'll have to recreate it, which is a bit more time taking:

```console
$ vagrant destroy services
$ vagrant up services
```

Then, edit your `setupdb.sh` to remove the tile purging bits (which shouldn't run locally):

```diff
diff --git a/scripts/aws/setupdb.sh b/scripts/aws/setupdb.sh
index 0d5b8443..4644c371 100755
--- a/scripts/aws/setupdb.sh
+++ b/scripts/aws/setupdb.sh
@@ -81,9 +81,7 @@ function download_and_load {
 }
 
 function purge_tile_cache {
-    for path in "${PATHS[@]}"; do
-        aws s3 rm --recursive "s3://tile-cache.${PUBLIC_HOSTED_ZONE_NAME}/${path}/"
-    done
+    echo "Skipping purging tile cache locally"
 }
 
 function create_trgm_indexes {
```

and then run:

```console
$ vagrant ssh app -c 'cd /vagrant && ./scripts/aws/setupdb.sh -bsdpmqc'
```

to load all the necessary dev data. Then you'll have to create another account for yourself in the app, and then should be ready to go.

---

The DRB 2011 analyses may still fail because they need an API key set. But they are not necessary for testing this work.

jwalgran:
I've been unable to complete testing and will need to schedule some pairing time to get this wrapped up. While retesting the process I noticed that the initial POST to create the streams job was timing out and returning 504.

Screen Shot 2021-08-03 at 7 53 41 PM

rajadain:
Let's pair on this today.

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -494,19 +494,21 @@ def nlcd_kfactor(result):
     return output
 
 
-def multi_mapshed(aoi, wkaoi):
+def multi_mapshed(aoi, wkaoi, layer_overrides={}):
```
rajadain, line 497: Perhaps I should replace this with `dict`

```suggestion
def multi_mapshed(aoi, wkaoi, layer_overrides=dict):
```

src/mmw/apps/modeling/mapshed/tasks.py
```diff
@@ -494,19 +494,21 @@ def nlcd_kfactor(result):
...
     shape = [{'id': wkaoi or NOCACHE, 'shape': aoi}]
     stream_lines = streams(aoi)[0]
 
-    return multi.s('mapshed', shape, stream_lines)
+    return multi.s('mapshed', shape, stream_lines,
+                   layer_overrides=layer_overrides)
 
 
-def multi_subbasin(parent_aoi, child_shapes):
+def multi_subbasin(parent_aoi, child_shapes, layer_overrides={}):
```
rajadain, line 505: Here too

```suggestion
def multi_subbasin(parent_aoi, child_shapes, layer_overrides=dict):
```

src/mmw/apps/modeling/mapshed/tasks.py
jwalgran: Yes!

src/mmw/apps/modeling/mapshed/tasks.py
jwalgran: Yes!

jwalgran (approved):
The changes look good to me. 

I spent so much time working on my dev environment I did not go back and rerun the entire test process after fixing my analysis/data issues. I compared new projects in development with staging and verified that they did produce different results.

Screen Shot 2021-08-04 at 12 14 23 PM


Screen Shot 2021-08-04 at 12 16 01 PM

jwalgran:
We traced the 5xx timeout error I was seeing to the fact that my nhdflowlines table had no indexes. We got analysis working on my dev VM by loading/reloading layers.

Issue #3415, opened 2021-08-05
Ingest High Resolution NHD Streams


rajadain:
Evaluate how they fit in, and import them.

We would want all the high resolution streams for the Continental United States.

Look through the values here: https://www.usgs.gov/core-science-systems/ngp/national-hydrography/access-national-hydrography-products

Consider this dataset: https://prd-tnm.s3.amazonaws.com/index.html?prefix=StagedProducts/Hydrography/NHD/National/HighResolution/GDB/

Inspect the current `nhdflowline` table and see if this dataset has equivalent columns. Find equivalents if not.

Evaluate the size of the database and report the impact this will have.

We only need the LineStrings, so consider dropping any other geometries in the dataset.

rajadain:
Here's what the existing table looks like:

```sql
modelmywatershed=> \d nhdflowline;
              Table "public.nhdflowline"
    Column    |            Type            | Modifiers
--------------+----------------------------+-----------
 ogc_fid      | integer                    | not null
 geom         | geometry(LineStringZ,4326) |
 comid        | numeric(9,0)               |
 fdate        | date                       |
 gnis_id      | character varying(10)      |
 gnis_name    | character varying(65)      |
 lengthkm     | numeric(19,11)             |
 reachcode    | character varying(14)      |
 flowdir      | character varying(15)      |
 ftype        | character varying(24)      |
 fcode        | numeric(9,0)               |
 shape_leng   | numeric(19,11)             |
 gnis_nbr     | numeric(9,0)               |
 stream_order | numeric(2,0)               |
 slope        | numeric(14,8)              |
Indexes:
    "nhdflowline_pkey" PRIMARY KEY, btree (ogc_fid)
    "nhdflowline_stream_order_idx" btree (stream_order)
    "nhdflowline_wkb_geometry_geom_idx" gist (geom)
```

The dataset from the link has many layers in it:

```console
❯ ogrinfo NHD_H_National_GDB.zip
INFO: Open of `NHD_H_National_GDB.zip'
      using driver `OpenFileGDB' successful.
1: NHDPointEventFC (Point)
2: NHDLineEventFC (Multi Line String)
3: NHDAreaEventFC (Multi Polygon)
4: NHDFCODE (None)
5: ExternalCrosswalk (None)
6: NHDFlow (None)
7: NHDSourceCitation (None)
8: NHDMetadata (None)
9: NHDPoint (3D Point)
10: NHDReachCodeMaintenance (None)
[...]

Issue #3416, opened 2021-08-05
Upgrade base box to at least Ubuntu 18.04


colekettler:
We are currently on Ubuntu 16.04 which recently transitioned to Extended Security Maintenance (ESM). This requires an Ubuntu Advantage subscription and valid token to continue to access its apt repositories. This is currently causing CI to fail.

We should target an upgrade to at least Ubuntu 18.04 to get back onto an LTS release under active support.

Ideally we could target an upgrade to 20.04, but it's worth noting that this is the first release to use Python 3 as the default interpreter. We would need to install Python 2 and switch to it as the default interpreter ourselves. This may be prohibitive until we complete #3165. If this does prove to be prohibitive, we should target 18.04 and create an issue to upgrade to 20.04 after #3165 is completed.

We have previously completed a similar upgrade in https://github.com/azavea/cicero/pull/2206. This could serve as an example for implementation.

rajadain:
(I'm going to keep a running log of notes as I work through this here.)

I've been trying to upgrade to Ubuntu 20.04 and have a branch [`tt/upgrade-ubuntu-20.04`](https://github.com/WikiWatershed/model-my-watershed/compare/tt/upgrade-ubuntu-20.04) where I'm adding commits, very rough, just to get things in a working state.

I've been able to get through most things, but this seems like a blocker: https://code.djangoproject.com/ticket/31838

Seemingly Django 1.11 is not supported in Ubuntu 20.04?

A Django upgrade can be evaluated, but I'm hesitant to entangle that with this effort. Will revisit this tomorrow.

rajadain:
Setting points from 5️⃣ to 0️⃣ for tracking purposes.

PR #3417, opened 2021-08-18
Ingest NHD High Resolution Streams, Add Tile Layer

rajadain:
## Overview

The full ingest process is documented in https://github.com/WikiWatershed/model-my-watershed/issues/3415#issuecomment-900309456. This process took several days, not just to figure out the data shape, but to process it, because of the volume. This is by far the largest volume of data we've added to MMW, and may have non-trivial consequences for performance and hosting costs that will reveal themselves as this moves to staging and production.

The work here adds the NHD High Resolution stream data to the MMW database, and wires up the tiler to render it on the map. It does not switch our analyses and models to use the new High Resolution data, which still use Medium Resolution. That will be done in future cards.

I was unable to provision the tiler VM on my local, which could only be resolved by upgrading the NPM version on that VM. It seemingly works well enough, and is isolated from the app VM, so should be fine.

Connects #3415 

### Demo

image

### Notes

 The compressed data is ~17GB, and in the database it is ~26GB:

```sql
SELECT pg_size_pretty(pg_total_relation_size('nhdflowlinehr'));

 pg_size_pretty 
----------------
 26 GB
(1 row)
```

There's a total of ~24.5M rows:

```sql
SELECT COUNT(*) FROM nhdflowlinehr;

  count   
----------
 24517604
(1 row)
```

I attempted to drop the extraneous columns in the table to see if that would reduce the size much, but it didn't. I imagine that most of the data is the geometry of the streams, which we cannot elide.

About 2% of the streams do not have a value for `stream_order` or `slope`, which were obtained by joining with the Value Added Attributes table:

```sql
SELECT COUNT(*) FROM nhdflowlinehr WHERE stream_order IS NULL;

 count  
--------
 493431
(1 row)
```

## Testing Instructions

Before you begin, ensure you have ~80+ GB of free space on your host computer.

* Import the new high resolution streams with:
    ```console
    $ vagrant ssh app -c 'cd /vagrant && ./scripts/aws/setupdb.sh -S'
    ```
    This might take 30-60 minutes.
  - If this fails, try giving your services VM more resources:
      ```diff
      diff --git a/Vagrantfile b/Vagrantfile
      index 600fc14d..561d833c 100644
      --- a/Vagrantfile
      +++ b/Vagrantfile
      @@ -52,7 +52,8 @@ Vagrant.configure("2") do |config|
 
           services.vm.provider "virtualbox" do |v|
             v.customize ["guestproperty", "set", :id, "/VirtualBox/GuestAdd/VBoxService/--timesync-set-threshold", 10000 ]
      -      v.memory = 2048
      +      v.memory = 6144
      +      v.cpus = 4
           end
 
           services.vm.provision "ansible" do |ansible|
      ```
  - [ ] Ensure it succeeds
* While the above is happening, reprovision your tiler:
    ```console
    $ vagrant reload --provision tiler
    ```
  - [ ] Ensure it succeeds
* Once the import is complete, go to http://localhost:8000/
* Turn on the Continental US High Resolution Stream Network layer
  - [ ] Ensure it renders as expected
* Turn on the Continental US Medium Resolution Stream Network layer
  - [ ] Ensure that still works as before

### DIFF
--- a/src/tiler/styles/streams.mss
+++ b/src/tiler/styles/streams.mss
@@ -92,6 +92,90 @@
   }
 }
 
+#nhdflowlinehr {
+  line-color: @streamColor;
+  line-join: round;
+  line-cap: round;
+}
+
+#nhdflowlinehr[zoom<=4] {
+  [stream_order=10] {
+    line-width: 5.0 * @zoomBase;
+  }
+  [stream_order=9] {
+    line-width: 3.0 * @zoomBase;
+  }
+  [stream_order<=8] {
+    line-width: 2.0 * @zoomBase;
+  }
+}
+
+#nhdflowlinehr[zoom>=5][zoom<=6] {
+  [stream_order=10] {
+    line-width: 7.0 * @zoomBase;
+  }
+  [stream_order=9] {
+    line-width: 4.0 * @zoomBase;
+  }
+  [stream_order<=8] {
+    line-width: 3.0 * @zoomBase;
+  }
+}
+
+#nhdflowlinehr[zoom>=9][zoom<=10] {
+  [stream_order>=9] {
+    line-width: 14.0 * @zoomBase;
+  }
+  [stream_order<=8][stream_order>=6] {
+    line-width: 10.0 * @zoomBase;
+  }
+  [stream_order<=5][stream_order>=4] {
+    line-width: 5.0 * @zoomBase;
+  }
+  [stream_order=3] {
+    line-width: 3.0 * @zoomBase;
+  }
+}
+
+#nhdflowlinehr[zoom>=11][zoom<=12] {
+  [stream_order>=9] {
+    line-width: 18.0 * @zoomBase;
+  }
+  [stream_order<=8][stream_order>=6] {
+    line-width: 12.0 * @zoomBase;
+  }
+  [stream_order<=5][stream_order>=4] {
+    line-width:  9.0 * @zoomBase;
+  }
+  [stream_order=3] {
+    line-width:  6.0 * @zoomBase;
+  }
+  [stream_order=2] {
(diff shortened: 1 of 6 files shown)


src/tiler/styles/streams.mss
```diff
@@ -92,6 +92,90 @@
   }
 }
 
+#nhdflowlinehr {
```
rajadain, line 95: These styles are identical to those for `#nhdflowline`

jwalgran:
An update. I set up a new set of vagrant VMs on an external drive and increased the memory and CPU as suggested. Loading the high res data has been chugging on ALTER TABLE for a while, a lot more than the suggested "30-60 minutes"

```
+ curl -s https://s3.amazonaws.com/data.mmw.azavea.com/nhdflowlinehr.sql.gz
+ gunzip -q
+ psql --single-transaction
SET
SET
SET
SET
SET
 set_config
------------

(1 row)

SET
SET
SET
SET
SET
CREATE TABLE
ALTER TABLE
CREATE SEQUENCE
ALTER TABLE
ALTER SEQUENCE
ALTER TABLE
COPY 24517604
  setval
----------
 24517604
(1 row)

ALTER TABLE

````

The disk on my services VM is up to about 33GB. 

Screen Shot 2021-08-22 at 8 30 24 PM

I will let it go overnight.

jwalgran (approved):
I was unable to see the data load complete in a reasonable amount of time when attempting to run the VM from my external spinning hard disk. The process did appear to be working, and the changes in the PR are mostly a reuse of existing functionality. 👍

rajadain:
Thanks for taking a look. I'll merge this and we can get this on staging and evaluate there. (Although we may have to wait for #3416 before staging deployments work again.)

Issue #3418, opened 2021-08-19
Analyze High Resolution NHD Streams


rajadain:
- [ ] Add an Analyze endpoint for the high resolution NHD streams added in #3417 
- [ ] Update the Analyze Streams tab with a dropdown selector, similar to the Land Analyze tab. This should have two options: NHD Medium Resolution and NHD High Resolution. NHD High Resolution should be the default.
- [ ] Query the NHD High Resolution analyze endpoint from the sidebar and display the results in the table

PR #3428, opened 2021-10-21
Analyze High Resolution NHD Streams

rajadain:
## Overview

Add analysis of NHD High Resolution Streams to the Analyze sidebar.

The existing NHD Medium Resolution Streams are still analyzed, but are relegated to a non-default dropdown selection.

The stream selection query was modified to work with `LineString`s from NHD Medium Resolution table `nhdflowline` and the `MultiLineString`s from NHD High Resolution table `nhdflowlinehr`. This change will actually make the processing faster, since we're doing fewer operations, `ST_Multi` instead of `ST_Collect` and `ST_Force2D`. `ST_Multi` would be a no-op for `nhflowlinehr`, make it even faster.

Also change current analyze streams endpoint to require an explicit stream datasource parameter. This is a breaking change, similar to the change made for Land analysis in #3399.

Connects #3418 

## Testing Instructions

* Ensure you have both NHD Streams datasets installed locally. Comment out [these lines](https://github.com/WikiWatershed/model-my-watershed/blob/aaaace876aaf53a8ca3e536e9103b42501f3142a/scripts/aws/setupdb.sh#L87-L89), then run:
    ```
    vagrant ssh app -c 'cd /vagrant && ./scripts/aws/setupdb.sh -sS
    ```
* Check out this branch and restart Celery to load the new tasks
    ```
    ```
* Go to http://localhost:8000/ and choose an area of interest
* In the analyze stage:
  - [x] Ensure you see a dropdown in the Stream tab
  - [x] Ensure both analyses complete successfully
  - [x] Ensure the values are different, and make sense (High Resolution having more stream length than Medium Resolution)

### DIFF
--- a/scripts/aws/setupdb.sh
+++ b/scripts/aws/setupdb.sh
@@ -3,14 +3,15 @@
 set -e
 set -x
 
-usage="$(basename "$0") [-h] [-b] [-s] \n
+usage="$(basename "$0") [-h] [options] \n
 --Sets up a postgresql database for MMW \n
 \n
-where: \n
+where options are one or more of: \n
     -h  show this help text\n
     -b  load/reload boundary data\n
     -f  load a named boundary sql.gz\n
     -s  load/reload stream data\n
+    -S  load/reload Hi Res stream data (very large)\n
     -d  load/reload DRB stream data\n
     -m  load/reload mapshed data\n
     -p  load/reload DEP data\n
--- a/src/mmw/apps/geoprocessing_api/calcs.py
+++ b/src/mmw/apps/geoprocessing_api/calcs.py
@@ -52,9 +52,9 @@ def animal_population(geojson):
     }
 
 
-def stream_data(results, geojson):
+def stream_data(results, geojson, datasource='nhd'):
     """
-    Given a GeoJSON shape, retreive stream data from the `nhdflowline` table
+    Given a GeoJSON shape, retreive stream data from the specified table
     to display in the Analyze tab
 
     Returns a dictionary to append to outgoing JSON for analysis results.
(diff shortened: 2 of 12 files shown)


emilyhu0106 (commented):
The new implementation works very well in the pre-defined boundaries, but sometimes the hand-drawn areas (both from free hand and square) yields unexpected results. Below is an example of stream length getting shorter in High Resolution, even though it captured one additional stream inside that area:


I am not sure if this is related to `streams_for_huc12s` at line 325 in `src/mmw/apps/geoprocessing_api/calcs.py`?

> def streams_for_huc12s(huc12s, drb=False):
    """
    Get MultiLineString of all streams in the given HUC-12s
    """
    sql = '''
          SELECT ST_AsGeoJSON(ST_Collect(ST_Force2D(s.geom)))
          FROM {datasource} s INNER JOIN boundary_huc12 b
            ON ST_Intersects(s.geom, b.geom_detailed)
          WHERE b.huc12 IN %s
          '''.format(datasource='drb_streams_50' if drb else 'nhdflowline')

    with connection.cursor() as cursor:
        cursor.execute(sql, [tuple(huc12s)])

        return [row[0] for row in cursor.fetchall()]  # List of GeoJSON strings

rajadain:
Ooh that is an excellent catch Emily!

Can you export that shape as GeoJSON and attach here? I'd like to test it out myself. You can export a shape using the Download button in the top right:

image

> I am not sure if this is related to `streams_for_huc12s` at line 325 in `src/mmw/apps/geoprocessing_api/calcs.py`?

It is not, but that code should also be updated, and was missed. That is used when we're running subbasin analysis, which I haven't demoed for you yet. Much appreciated!

emilyhu0106:
> Can you export that shape as GeoJSON and attach here? I'd like to test it out myself. You can export a shape using the Download button in the top right:

Please find the GeoJson file attached: [1 Square Km.geojson.zip](https://github.com/WikiWatershed/model-my-watershed/files/7470701/1.Square.Km.geojson.zip)

I found this problem more common in smaller areas, especially when high-resolution layer captures one more stream. Hopefully it is helpful!

rajadain:
I'm going to fix that issue as part of this PR, but wrote up my investigation in #3430 for the record.

rajadain:
That was a pretty amazing catch, exposed a 4 year old bug in our analysis! Fixed in 5b1f624. Here's a demo:



The values are going to be lower lengths than before, since previously we weren't clipping the streams to the area of interest, and now we are.

rajadain:
The `streams_for_huc12s` bit and nearby entities are related to modeling, not analyzing, and will be updated as part of #3429 which is upcoming.

rajadain:
Tests are passing locally, seem to be failing sporadically on CI. Continuing to rebuild.

rajadain:
Build is finally green again ✅ . This is ready for another look.

emilyhu0106 (approved):
Great implementation! That also explained why the previous problem only exists when the use define the area, but not in pre-defined hydrologic units. 
I tested several areas and they all look nice!

Issue #3419, opened 2021-08-26
Upgrade Django to 3.2


rajadain:
1.11 LTS ended in April 2020. Once the Python 3 upgrade (#3165) is done, upgrade Django to latest LTS, which as of now is 3.2: https://endoflife.date/django

Also unskip tests that were skipped in 6489389b.

PR #3443, opened 2021-12-16
Upgrade to Django 3.2

rajadain:
## Overview

Upgrades to the most recent LTS version of Django.

The approach was to increment from 1.11 to 2.0, then to 2.2, then to 3.2.

At each stage, I ran the test suite with all Python warnings turned on, using the `-Wa` switch:

```bash
vagrant ssh app -c 'cd /opt/app && PYTHONTRACEMALLOC=1 envdir /etc/mmw.d/env python3 -Wa manage.py test --no-input --exclude-tag=mapshed'
```

And performed upgrades / fixes as it indicated.

Finally, I ran the server with warnings turned on:

```bash
vagrant ssh app -c 'cd /opt/app && PYTHONTRACEMALLOC=1 envdir /etc/mmw.d/env python3 -Wa manage.py runserver'
```

And interacted with the app as much as possible, looking for edge cases not covered by the tests, and fixed those as they came up.

Connects #3419 

### Notes

Given the large number of dependency changes, we may see some fallout on staging and production.

## Testing Instructions

* Check out this branch
* Destroy your `app` and `worker` VMs, and recreate them:
    ```bash
    vagrant destroy -f app worker && vagrant up
    ```
* In two separate tabs, run the `debugserver` and `debugcelery` scripts
* Go to http://localhost:8000
  - [x] Ensure it loads
  - [x] Ensure there's no errors in `debugserver` output
* Select a shape and analyze
  - [x] Ensure analysis works, and that `debugcelery` output looks correct
* Model a shape with Watershed Multi-Year
  - [x] Ensure that works correctly
* Run tests with all warnings turned on:
    ```bash
    vagrant ssh app -c 'cd /opt/app && PYTHONTRACEMALLOC=1 envdir /etc/mmw.d/env python3 -Wa manage.py test --no-input --exclude-tag=mapshed'
    ```
  - [x] Ensure there are no warnings in our code
    - We do see two `RemovedInDjango41Warning`, one each in `rest_framework_gis` and `django_celery_results`, even though we are running the latest versions of both. These should be fixed in future updates to those libraries. Nothing to worry about until we upgrade to Django 4.2
* Pick a standard shape and run Watershed Multi-Year Model in production: https://modelmywatershed.org/. Note the value of Mean Flow under Water Quality:
    image
* Compare the same value on staging, and see how it is incorrectly lower than the production value:
    image
* Run the same shape on your local:
    image
  - [x] Ensure the value is larger than production (since we're now using the high resolution streams which have longer length)


emilyhu0106 (approved):
Nice work for this big upgrade project and thank you for fixing the pip problem timely yesterday! Everything works well!

rajadain:
Thank you for reviewing!

Issue #3446, opened 2021-12-27
Subbasin Timing Out for Schuylkill HUC-8


rajadain:
Subbasin is timing out for the Schuylkill HUC-8 on staging, which has the NLCD2019 and Hi Res Streams:

![image](https://user-images.githubusercontent.com/1430060/147492208-21e22bfa-a389-4be3-ae28-e74c158758d5.png)

It works correctly on production, and works for smaller shapes like the Lower Schuylkill HUC-10 on staging.

See #2835 for similar work in the past.

rajadain:
While the increased timeout in https://github.com/WikiWatershed/mmw-geoprocessing/pull/98 gets us part of the way, we're not done yet. Now seeing `chord_unlock` loops when running GWLF-E:

```
[2022-01-04 14:47:42,228: INFO/MainProcess] Task apps.modeling.geoprocessing.multi[57048cbc-41c0-4fd3-8542-1aced8880850] received
[2022-01-04 14:47:42,631: INFO/MainProcess] Task apps.modeling.mapshed.tasks.collect_subbasin[06fd646f-e3f6-43a1-966f-eebd741b4630] received
[2022-01-04 14:47:54,519: INFO/MainProcess] Task apps.modeling.tasks.subbasin_results_to_dict[3f8585e1-339e-4884-874b-b31dc0762a2f] received
[2022-01-04 14:47:57,573: INFO/MainProcess] Task apps.core.tasks.save_job_result[a1440582-b1ac-4495-a695-6e67f44a0e01] received
[2022-01-04 14:48:03,618: INFO/MainProcess] Task celery.chord_unlock[cffdccbf-1a47-4686-839e-e3c49a503e27] received
[2022-01-04 14:48:03,621: INFO/MainProcess] Task apps.modeling.tasks.run_subbasin_gwlfe_chunks[f46c977c-19b6-4a5a-a12e-1e458118cf03] received
[2022-01-04 14:48:03,641: INFO/MainProcess] Task apps.modeling.tasks.run_subbasin_gwlfe_chunks[98658202-179b-458b-a119-c255c91ed30f] received
[2022-01-04 14:48:04,552: DEBUG/ForkPoolWorker-2] Running model...
[2022-01-04 14:48:04,556: DEBUG/ForkPoolWorker-1] Running model...
[2022-01-04 14:48:04,941: DEBUG/ForkPoolWorker-2] Model run complete for 30 years of data.
[...]

Issue #3447, opened 2021-12-27
Boundary Search Broken post Django Upgrade


rajadain:
https://rollbar.com/WikiWatershed/ModelMyWatershed/items/179/

```
Internal Server Error: /mmw/modeling/boundary-layers-search/
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.8/dist-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.8/dist-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python3.8/dist-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/local/lib/python3.8/dist-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "/usr/local/lib/python3.8/dist-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/rest_framework/decorators.py", line 50, in handler
    return func(*args, **kwargs)
  File "/opt/app/apps/modeling/views.py", line 702, in boundary_layer_search
    result = boundary_search_context(search_term)
  File "/opt/app/apps/modeling/calcs.py", line 396, in boundary_search_context
    _do_boundary_search(search_term)
  File "/opt/app/apps/modeling/calcs.py", line 458, in _do_boundary_search
    point = wkb_r.read(row[4])
  File "/usr/local/lib/python3.8/dist-packages/django/contrib/gis/geos/io.py", line 18, in read
    return GEOSGeometry(super().read(wkb))
  File "/usr/local/lib/python3.8/dist-packages/django/contrib/gis/geos/prototypes/io.py", line 153, in read
    return wkb_reader_read_hex(self.ptr, wkb, len(wkb))
[...]

Issue #3434, opened 2021-11-11
Incorporate Hi Res Stream Adjustments


rajadain:
Once #3432 is in, and BME takes a look on staging, incorporate adjustments to SedAAdjust and other factors on his recommendation.

rajadain:
BME recommends a SedAAdjust value of 1.25.

rajadain:
Shares points with #3446

rajadain:
BME recommended a value of 0.8.

PR #3432, opened 2021-11-10
Model NHD Hi Res Streams

rajadain:
## Overview

Switch to using NHD Hi Res streams for modeling by default. A migration is added to declare all existing projects to NHD Medium Res, so their values should not change. Also switch the Subbasin processing to use them.

Bring in a new version of the geoprocessing service that supports multiple MultiLineStrings for the `/multi` endpoint. See https://github.com/WikiWatershed/mmw-geoprocessing/pull/94

Connects #3429 

### Notes

Discovered #3431 while working on this, will fix that next.

Since the PA DEP Worksheet Export is run _before_ there is a project, it uses the default values, which going forward will be NHD Hi Res streams. There is currently no way to run the Worksheet Export with Medium Res streams. This may be added as a query parameter set by a modal in the future if necessary.

### Preparation

* Start on latest `develop`, then turn on the VMs
    ```console
    vagrant up
    ```
* From LastPass, copy the "New key" value of "SRAT Catchment API MMW (Staging)"
* Use that value as an environment variable to provision the `app` and `worker` VMs
    ```console
    MMW_SRAT_CATCHMENT_API_KEY=*** vagrant provision app worker
    ```

### Baseline

* Go to https://staging.modelmywatershed.org/
* Select a HUC-10 Boundary
* Once Analysis completes, select the Watershed Multi-Year Model
    image
* Once the Watershed Multi-Year Model completes, export the GMS file
    image
* Rename the file to be `staging-huc10.gms`

### Pre-existing NHD Project

* Go to http://localhost:8000/
* Ensure you are logged in
* Select the same HUC-10 Boundary
* Run the Watershed Multi-Year Model completes, export the GMS file and save it as `local-huc10-nhd-old.gms`
* [x] Diff the two files, ensure there is no difference
    ```console
    diff staging-huc10.gms local-huc10-nhd-old.gms
    ```

### Test NHD Upgrade

* Checkout this branch
* Reprovision worker 
    ```console
    MMW_SRAT_CATCHMENT_API_KEY=*** vagrant provision app worker
    ```
* Run migrations
    ```console
    ./scripts/manage.sh migrate
    ```
* Go to http://localhost:8000/
* Open the project you made in the previous step
* Create a new scenario
    image
* Export the GMS file for the new scenario, save it as `local-huc10-nhd-new.gms`
* [x] Diff the old and new files, ensure there is no difference
    ```console
    diff local-huc10-nhd-old.gms local-huc10-nhd-new.gms
    ```

[...]

PR #3448, opened 2021-12-28
Update SedAAdjust Value, Fix Boundary Search Queries

rajadain:
## Overview

Fixes boundary search queries which were broken since the Python upgrade, as the new WKBReader expects bytes as input, not strings, which are returned from the query. That is handled by removing WKBReader altogether, and using PostGIS functions instead.

Also updates the value of `SedAAdjust` to ~1.25~ 0.8, which accounts for the higher resolution streams now available. The hi res streams are used by default for all projects going forward. Older projects, which use medium resolution streams, are updated to use the old value of 1.5, unless they were using something else already.

Also fixes subbasin timeouts for HUC-8s.

Connects #3447 
Connects #3434 
Connects #3446 

### Demo

image

image

### Boundary Search

- Check out this branch
- Reprovision the `app` VM
    ```
    vagrant provision app
    ```
- In the search bar, type in the name of a HUC, such as Schuylkill
  - [x] Ensure the HUCs show up in the search results
  - [x] Ensure you can select one and it brings up the right shape

### SedAAdjust

The testing instructions validate how the migration works. For old projects, if they are marked as using NHD Streams (as done in c4c5350fd8614d8c9ee6fbc2c74ceaf902e0d6b7 for #3432), and they don't already override the Streambank Erosion SedAAdjust value, then they are set to use the old 1.5 value. This means:

- Old projects that previously used 1.5 by default will now use 1.5 by override, whether they overrode other values or not
- Old projects that previously overrode the default 1.5 will continue to use their override
- New projects will not be touched

To prepare for this:

- Ensure you are logged in
- Pick a shape, and with the same shape, create 4 different Watershed Multi-Year Projects with it
- Rename them to be "Project 1", "Project 2", "Project 3", and "Project 4"
- For Project 1, click "Add changes to this area", but don't make any changes to New Scenario
- For Project 2, click "Add changes to this area", and for the New Scenario, open Settings, go to Other Model Data, and add a Streambank Erosion Adjustment Factor of 1.7:
    image
- For Project 3, go to the same spot, and leave Streambank Erosion blank, but set Sediment Delivery Ratio to 0.2:
    image
- For Project 4, click "Add changes to this area", but don't make any changes to New Scenario
- Open the Python shell with:
    ```
    ./scripts/manage.sh shell_plus
    ```
- Mark the first three projects as old, with this Python snippet pasted into the shell:
    ```python
    for p in Project.objects.filter(name__in=['Project 1', 'Project 2', 'Project 3']):
        p.layer_overrides['__STREAMS__'] = 'nhd'
        p.save()
    ```
- Exit the shell with `Ctrl+D`
- Run migrations
    ```
    ./scripts/manage.sh migrate
    ```

Finally, let's test our four projects:

- Open Project 1 → New Scenario → Settings → Other Model Data
  - [x] Ensure you see Streambank Erosion set to 1.5 (the default case)
- Open Project 2 → New Scenario → Settings → Other Model Data
  - [x] Ensure you see Streambank Erosion set to 1.7 (so the previous override was preserved)
- Open Project 3 → New Scenario → Settings → Other Model Data
  - [x] Ensure you see Streambank Erosion set to 1.5 and Sediment Delivery Ratio to 0.2 (so other previous overrides are preserved)
- Open Project 4 → New Scenario → Settings → Other Model Data
  - [x] Ensure you see no overridden value of Streambank Erosion (so non-NHD projects are not affected)

### Subbasin HUC-8

- Copy the value of "New Key" in "SRAT Catchment API MMW (Staging)" in LastPass
- Using that as an environment variable, reprovision `app` and `worker`:
    ```
    MMW_SRAT_CATCHMENT_API_KEY="***" vagrant provision app worker
    ```
- Search for "Schuylkill" and select the HUC-8 shape
- Run the Watershed Multi-Year Model on this
- Switch to the Water Quality tab, and then click "View subbasin attenuated results"

    image

  - [x] Ensure it finishes successfully


rajadain:
Just added a few commits to fix #3446 to this PR. Updated Overview, Demo, Testing Instructions to match.

emilyhu0106 (approved):
Great work! Everything works well! It is interesting to review one big PR with multiple parts inside.

rajadain:
Thanks for reviewing! I have a call with Barry today, he may recommend a different value of SedAAdjust. I'll amend this PR after conferring with him with the right new value. It will not need re-review, since all the mechanisms we tested will remain the same.

Once the new value has been set, I'll merge this.

PR #3459, opened 2022-01-13
Add Timeout Environment Variable, Use Throughout, Up to 200s

rajadain:
## Overview

We noticed that despite #3448 and https://github.com/WikiWatershed/mmw-geoprocessing/pull/98, which fixed timeouts for the Schuylkill HUC-8 in development, we were still seeing timeouts on Staging. This could be because of multiple users requesting runs at the same time. This PR further ups that to 200s.

Updates to mmw-geoprocessing 5.2.0, which exposes environment variable overrides for certain settings. See https://github.com/WikiWatershed/mmw-geoprocessing/pull/100.

Adds an `MMW_GEOPROCESSING_TIMEOUT` environment variable that is used in the `mmw-geoprocessing` service definition to set a timeout value of 200 seconds, up from the 120 baked in.

Also wires in that environment variable to be used for Celery, and the front-end, up from the 120 seconds as previously configured. Also updates messaging to be more appropriate.

Connects #3446 

## Testing Instructions

Since this was already working in development and not staging, the true test is on staging.

* Checkout this PR run:
    ```
    MMW_SRAT_CATCHMENT_API_KEY=*** vagrant provision app worker
    ```
* Run sub-basin on the Schuylkill HUC-8
  - [x] Ensure it succeeds

### DIFF
--- a/deployment/ansible/group_vars/all
+++ b/deployment/ansible/group_vars/all
@@ -44,8 +44,9 @@ docker_compose_version: "1.26.*"
 geop_host: "localhost"
 geop_port: 8090
 
-geop_version: "5.1.0"
+geop_version: "5.2.0"
 geop_cache_enabled: 1
+geop_timeout: 200
 
 nginx_cache_dir: "/var/cache/nginx"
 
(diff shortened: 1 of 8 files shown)


rajadain:
With CI passing, in the interest of time, I'm going to merge this in.

Issue #3465, opened 2022-01-20
Investigate High RDS IOPS Use, Production Service Interruption


rajadain:
Since the deployment of 1.33, we're observing very high IOPS use of the RDS instance in production:

![image](https://user-images.githubusercontent.com/1430060/150265222-ca37b54c-1582-4d02-b58a-9339b7003f05.png)

This is coincident with the site going down a lot:

![image](https://user-images.githubusercontent.com/1430060/150265508-db93e89d-f05b-4511-8142-df85c0dff361.png)

And we see these messages in RDS:

![image](https://user-images.githubusercontent.com/1430060/150265594-43deb40e-05b9-4a6a-b636-9526e190d4eb.png)

Since this is happening after 1.33, it is highly likely it has something to do with the High Res Streams, which is a very large table introduced in that release. Also, the high IOPS use may be caused by a very high request rate, possibly caused by a large number of requests from the tiler.

It is possible that #3424 may help. Or batch-generation of tiles. Or transferring the tiles already made for staging to production. Or it may be something else entirely.

Investigate, and if possible, fix.

rajadain:
To reduce the load of NHD hi res tile requests, I wrote this script:

```python
import mercantile
import requests

for zoom in range(19):
    # bbox via https://gist.github.com/jsundram/1251783
    for tile in mercantile.tiles(-124.7844079,24.7433195,-66.9513812,49.3457868, zoom):
        url = f'http://33.33.34.35/nhd_streams_hr_v1/{zoom}/{tile.x}/{tile.y}.png'
        requests.get(url)
        print(url)
```

and then configured the `tiler` VM to point to (and have access to) the Production tile cache bucket (by configuring the `MMW_TILECACHE_BUCKET` environment variable, and adding credentials for `mmw-prd` as the `default` in `/var/lib/mmw/.aws/credentials` inside the `tiler`). Then I ran the script, which ran for 1d 29m 11s before I canceled it, with the final tile getting pre-cached being `http://33.33.34.35/nhd_streams_hr_v1/15/7385/11671.png`.

This means that all of zoom levels 0-14 for the entire CONUS have tiles pre-cached, and a partial cache for zoom 15 was also added. In most cases, zooming beyond level 15 doesn't result in as many stream reads, so the usage of the burst balance should be reduced.

The tile cache for this layer on production is currently ~30 GB in size:

[...]

PR #3468, opened 2022-01-25
Hide hi res streams layer below zoom 8

rajadain:
## Overview

This is the min zoom used for HUC-12 and other layers as well.
This will reduce the loads on the database as there can be a very high number of streams at lower zoom levels.

Connects #3465


caseycesari (approved):
Looks good! Nice quick fix.

rajadain:
Thanks for reviewing!

Issue #3457, opened 2022-01-12
Investigate Sub-basin Errors


rajadain:
In this spreadsheet: [mapshed_water_quality_loads_Wissahickon.xlsx](https://github.com/WikiWatershed/model-my-watershed/files/7851500/mapshed_water_quality_loads_Wissahickon.xlsx), BME discovered that the numbers being sent to SRAT for component HUC-12s are larger than those shown in MMW when modeling just the HUC-12s themselves.

Investigate and correct.

From BME:

> Per our call yesterday, attached is a marked-up Excel file that shows results for the Wissahickon HUC10 basin near Philadelphia. To the left are results for the HUC10 without any sub-basin modeling, and to the right are results for the same basin after sub-basin modeling (and attenuation). As you can see, after this run several categories show an increase in loads instead of a decrease. This suggests to me that maybe estimated load reductions due to attenuation are being added to the "non-attenuated" loads rather than being subtracted. I haven't heard back from Mike yet about what he may have found, so there's no need for you to dig too deeply into this until I hear his response since we don't know whether the error is occurring on the API side before the results are sent back to MMW, or after. However, maybe what you could do in the meantime is too verify what data are being returned to MMW immediately after the Drexel sub-basin API is executed so that we can better understand what may be happening. Thanks.

> I've put the load results shown in the Manatawny JSON "request" file that you gave me into another Excel file (see attached) and compared them with the results given by MMW for each of the HUC12s comprising the Manatawny HUC10, and the results are different than what we got for the Wissahickon HUC10. As you recall, with the Wissahickon, the JSON request results for one HUC12 were OK for all values, but higher than expected for Hay/Pasture and Cropland for the other HUC12. However, for the Manatawny (which has 3 HUC12s), the results for one HUC12 were OK, but the Hay/Pasture loads for the other two HUC12s were lower than expected (and the Cropland results were OK). I'm not sure what to make of this, but hopefully you can sort it out. Let me know if we need to discuss this further. Thanks.

rajadain:
Also check if we're using different point sources for sub-basin vs normal MapShed modeling.

rajadain:
I added the following print statements and turned off geoprocessing caching:

```diff
diff --git a/src/mmw/apps/modeling/tasks.py b/src/mmw/apps/modeling/tasks.py
index 8fa654e4..877f7180 100644
--- a/src/mmw/apps/modeling/tasks.py
[...]

PR #3469, opened 2022-01-29
Fix Subbasin Errors by Deepcopying Default Dictionary

rajadain:
## Overview

Previously we were using `dict.copy()` to make a copy of the default GWLFE dictionary before augmenting it with real values for calculated fields. This caused unexpected runtime errors, as `dict.copy()` is only one level deep, i.e. while the first level of values are copied, the values of values are still references. Changing those affected the default dictionary, which affected subsequent calculations.

By using `copy.deepcopy`, we ensure that we always start with a clean copy of all the default values, and can change them in isolation without affecting the global defaults.

This default dictionary corruption / bleeding was observed by Barry when running subbasin, but it could potentially have interfered with other operations as well. The true impact of this can not be calculated.

For details see:

- https://docs.python.org/3/library/stdtypes.html#dict.copy
- https://docs.python.org/3/library/copy.html#copy.deepcopy
- https://github.com/WikiWatershed/model-my-watershed/issues/3457#issuecomment-1024624915

Connects #3457 

### Demo

See here: https://github.com/WikiWatershed/model-my-watershed/issues/3457#issuecomment-1024627107

## Testing Instructions

* Check out this branch
* Add the following line for observation:
    ```diff
    diff --git a/src/mmw/apps/modeling/tasks.py b/src/mmw/apps/modeling/tasks.py
    index 8fa654e4..bdd2481a 100644
    --- a/src/mmw/apps/modeling/tasks.py
    +++ b/src/mmw/apps/modeling/tasks.py
    @@ -347,6 +347,8 @@ def run_gwlfe(model_input, inputmod_hash, watershed_id=None):
         result['inputmod_hash'] = inputmod_hash
         result['watershed_id'] = watershed_id

    +    print(f'==> Total Sediment: {result["SummaryLoads"][0]["Sediment"]}')
         return result
    ```
* Go to http://localhost:8000/
* Search for and select the **Upper Wissahickon HUC-12** and run the Watershed Multi-Year Model
* In the debugcelery output, note the value of "Total Subbasin" shown in yellow
* Search for and select the **Lower Wissahickon HUC-12** and run the Watershed Multi-Year Model
* In the debugcelery output, note the value of "Total Subbasin" shown in yellow
* Search for and select the **Wissahickon HUC-10** and run the Watershed Multi-Year Model
* Click the Water Quality tab and select "Subbasin Attenuated Results"
* In the debugcelery output, note the values of "Total Subbasin" shown in yellow
  - [x] Ensure they match those of the solo HUC-12 runs

### DIFF
--- a/src/mmw/apps/modeling/mapshed/tasks.py
+++ b/src/mmw/apps/modeling/mapshed/tasks.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from copy import deepcopy
 from celery import shared_task
 from django.conf import settings
 
@@ -50,7 +51,7 @@ def collect_data(geop_results, geojson, watershed_id=None, weather=None,
     area = geom.transform(5070, clone=True).area  # Square Meters
 
     # Data Model is called z by convention
-    z = settings.GWLFE_DEFAULTS.copy()
+    z = deepcopy(settings.GWLFE_DEFAULTS)
 
     z['watershed_id'] = watershed_id
 
--- a/src/mmw/apps/user/views.py
+++ b/src/mmw/apps/user/views.py
@@ -1,5 +1,6 @@
 import rollbar
 
+from copy import deepcopy
 from uuid import uuid1
 
 from django.contrib.auth import (authenticate,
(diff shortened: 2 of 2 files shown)


emilyhu0106 (changes requested):
The number matches well! Thank you for the clear explanation about `dict.copy` and `deepcopy()`

emilyhu0106 (approved):

rajadain:
Thanks for reviewing! Small changes that will make a big impact!

Issue #3424, opened 2021-10-10
Optmize NHD High Resolution Querying


rajadain:
Currently the NHD HR layer uses up a lot of RDS CPU and Read Operations: https://github.com/WikiWatershed/model-my-watershed/issues/3415#issuecomment-939494702

This can potentially lock up the database for other transactional queries (loading / saving projects, etc), and a single user can bring down the entire system.

This can be especially bad when the users zoom a lot, and can lead to tile requests timing out.

---

Find a way to optimize the tile requests to make them faster and lighter. Consider:

- Adding the right index
- Adjusting which stream orders are shown at which level
- Batch generating the tile cache

rajadain:
This may help https://postgis.net/2014/03/14/tip_intersection_faster/

rajadain:
The tile cache was batch generated in https://github.com/WikiWatershed/model-my-watershed/issues/3465#issuecomment-1020416107

rajadain:
This is also a good resource on how to read query plans in general: https://thoughtbot.com/blog/reading-an-explain-analyze-query-plan, such as the one in https://github.com/WikiWatershed/model-my-watershed/issues/3465#issuecomment-1020451039

PR #3476, opened 2022-02-02
Cache and reuse Stream Analyses for WKAoIs

rajadain:
## Overview

Previously we would only cache geoprocessing results, since they were the most expensive (by time, compute resources) operations. Database operations so far have been pretty instantaneous.

However, with the high resolution streams, some database queries have become very expensive. To ameliorate this, we now cache the results of those queries where possible.

Just as for geoprocessing, if operating on a WKAoI, the results will be cached, for instantaneous reuse. As this cache builds up, the apps performance will improve, and the load on the database will significantly decrease.

This work follows the caching pattern used in geoprocessing:

https://github.com/WikiWatershed/model-my-watershed/blob/7a906dce56c5084b91dcb2e6f9c5f96f152e1af3/src/mmw/apps/modeling/geoprocessing.py#L67-L72

https://github.com/WikiWatershed/model-my-watershed/blob/7a906dce56c5084b91dcb2e6f9c5f96f152e1af3/src/mmw/apps/modeling/geoprocessing.py#L101-L104

Connects #3424 

## Testing Instructions

* Check out this branch and make the following local changes:
    ```diff
    diff --git a/src/mmw/apps/geoprocessing_api/tasks.py b/src/mmw/apps/geoprocessing_api/tasks.py
    index 67f7b69d..b726e7d8 100644
    --- a/src/mmw/apps/geoprocessing_api/tasks.py
    +++ b/src/mmw/apps/geoprocessing_api/tasks.py
    @@ -100,12 +100,15 @@ def analyze_streams(results, area_of_interest, datasource='nhdhr', wkaoi=None):
             key = f'db_{wkaoi}__{datasource}__stream_data'
             cached = cache.get(key)
             if cached:
    +            print(f'==> CACHE HIT {key}')
                 return {'survey': cached}
    
    +    print('==> CACHE MISS')
         survey = stream_data(results, area_of_interest, datasource)
    
         if key:
             cache.set(key, survey, None)
    +        print(f'==> CACHE SAVE {key}')
    
         return {'survey': survey}
    ```
* Go to http://localhost:8000/ and select a large shape, such as a HUC-8
* Observe the Celery output
  - [x] Ensure you see CACHE MISS and CACHE SAVE messages in yellow
* Click "Change Area" and reselect the same shape as before
  - [x] Ensure you see CACHE HIT message in yellow
  - [x] Ensure the stream analysis completes faster this time
* Click "Change Area" and draw an area, instead of selecting one
  - [x] Ensure you see CACHE MISS message in yellow, but no CACHE SAVE
  - [x] Ensure the results still complete

### DIFF
--- a/src/mmw/apps/geoprocessing_api/tasks.py
+++ b/src/mmw/apps/geoprocessing_api/tasks.py
@@ -11,6 +11,7 @@ from celery import shared_task
 import requests
 
 from django.conf import settings
+from django.core.cache import cache
 
 from mmw.settings import layer_classmaps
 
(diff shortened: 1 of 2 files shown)


emilyhu0106 (approved):
The result is significantly faster after implementing cache. Awesome way to improve the speed and concise implementation!

rajadain:
Thanks for reviewing!

Issue #3478, opened 2022-02-03
Increase HTTP Max Content Length of Geoprocessing Service


rajadain:
We're seeing errors like this:

```
Feb 03 21:16:26 ip-10-0-1-142 geoprocessing[479]: [ERROR] [02/03/2022 21:16:26.192] [WebServer-akka.actor.default-dispatcher-6171] [akka://WebServer/user/StreamSupervisor-0/flow-4323-0-unknown-operation] Error during preStart in [akka.http.scaladsl.model.HttpEntity$Limitable@321f922d]
Feb 03 21:16:26 ip-10-0-1-142 geoprocessing[479]: EntityStreamSizeException: actual entity size (Some(85990207)) exceeded content length limit (52428800 bytes)! You can configure this by setting `akka.http.[server|client].parsing.max-content-length` or calling `HttpEntity.withSizeLimit` before materializing the dataBytes stream.
Feb 03 21:16:26 ip-10-0-1-142 geoprocessing[479]:         at akka.http.scaladsl.model.HttpEntity$Limitable$$anon$1.preStart(HttpEntity.scala:609)
Feb 03 21:16:26 ip-10-0-1-142 geoprocessing[479]:         at akka.stream.impl.fusing.GraphInterpreter.init(GraphInterpreter.scala:520)
Feb 03 21:16:26 ip-10-0-1-142 geoprocessing[479]:         at akka.stream.impl.fusing.GraphInterpreterShell.init(ActorGraphInterpreter.scala:380)
Feb 03 21:16:26 ip-10-0-1-142 geoprocessing[479]:         at akka.stream.impl.fusing.ActorGraphInterpreter.tryInit(ActorGraphInterpreter.scala:538)
Feb 03 21:16:26 ip-10-0-1-142 geoprocessing[479]:         at akka.stream.impl.fusing.ActorGraphInterpreter.preStart(ActorGraphInterpreter.scala:586)
Feb 03 21:16:26 ip-10-0-1-142 geoprocessing[479]:         at akka.actor.Actor$class.aroundPreStart(Actor.scala:510)
Feb 03 21:16:26 ip-10-0-1-142 geoprocessing[479]:         at akka.stream.impl.fusing.ActorGraphInterpreter.aroundPreStart(ActorGraphInterpreter.scala:529)
Feb 03 21:16:26 ip-10-0-1-142 geoprocessing[479]:         at akka.actor.ActorCell.create(ActorCell.scala:590)
[...]

PR #3479, opened 2022-02-07
Increase geoprocessing max length for large AoIs

rajadain:
## Overview

In certain cases, the number of high resolution streams in the area of interest can exceed the 50m default, so we increase it to 150m.

The default value is set here:

https://github.com/WikiWatershed/mmw-geoprocessing/blob/412565a58d26ecfe141706faae55aba1648579de/api/src/main/resources/application.conf#L18

Connects #3478 

## Testing Instructions

* On `develop`, watch the geoprocessing logs:
    ```
    vagrant ssh worker -c 'sudo journalctl -u mmw-geoprocessing.service -f'
    ```
* Go to http://localhost:8000/ and select Lower White HUC-8 shape in Indiana:
    image
    - [x] Ensure you see the error in the geoprocessing logs
* Check out this branch and reprovision the worker:
    ```
    vagrant provision worker
    ```
* Watch the geoprocessing logs again:
    ```
    vagrant ssh worker -c 'sudo journalctl -u mmw-geoprocessing.service -f'
    ```
* Run the same shape again. It will still exceed the timeout and not finish correctly, BUT there should not be an `EntityStreamSizeException` anymore
    - [x] Ensure you don't see that error in the logs

### DIFF
--- a/deployment/ansible/group_vars/all
+++ b/deployment/ansible/group_vars/all
@@ -47,6 +47,7 @@ geop_port: 8090
 geop_version: "5.2.0"
 geop_cache_enabled: 1
 geop_timeout: 200
+geop_maxlen: "150m"
 
 nginx_cache_dir: "/var/cache/nginx"
 
--- a/deployment/ansible/roles/model-my-watershed.base/defaults/main.yml
+++ b/deployment/ansible/roles/model-my-watershed.base/defaults/main.yml
@@ -14,6 +14,7 @@ envdir_config:
   MMW_GEOPROCESSING_PORT: "{{ geop_port }}"
   MMW_GEOPROCESSING_VERSION: "{{ geop_version }}"
   MMW_GEOPROCESSING_TIMEOUT: "{{ geop_timeout }}"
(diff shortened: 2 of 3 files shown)


emilyhu0106 (approved):
Great fix! Looking forward to seeing the next step.

I found the running time to get an error is a lot longer when the max length is 150m, it took me ~1hr to receive the error message. Is this because it takes longer time to geo-process the longer streams?

rajadain:
> it took me ~1hr to receive the error message.

😬 

Hmm, good observation. At that length, the Celery / front-end will timeout way before the geoprocessing finishes. Perhaps we are better off denying those requests outright, than letting them tie up our geoprocessing service for an extended period.

However, given how often this error is seen:

image

It may be worthwhile trying to see if this makes an improvement.

I'll think if this should be merged. Thank you for the review!

rajadain:
After thinking about this a bit more, I'm deciding to not merge this for now. The increased content allowance will tie up our geoprocessing workers for longer, but still not produce results within our timeout, so not improve the user experience at all. Perhaps a different solution is required.

Thank you for reviewing and posting your observations!

Issue #3467, opened 2022-01-25
Cannot Create Users Via Concord


rajadain:
See https://rollbar.com/WikiWatershed/ModelMyWatershed/items/221/

rajadain:
This is caused by this commit: https://github.com/django/django/commit/4b9330ccc04575f9e5126529ec355a450d12e77c

which makes it necessary for custom authentication backends to explicitly take a `request` parameter in their `authenticate` method, here:

https://github.com/WikiWatershed/model-my-watershed/blob/79cd1bd485c41654f468db69504041ccecd5f19c/src/mmw/apps/user/backends.py#L23-L31

rajadain:
This bug causes new users to be created successfully, but they are unable to be authenticated afterwards. This leads to multiple users being created:

```sql
SELECT c.concord_id, u.id, u.username, COUNT(p.id) AS projects
FROM user_concorduser c
	INNER JOIN auth_user u ON c.user_id = u.id
	LEFT JOIN modeling_project p ON u.id = p.user_id
WHERE c.concord_id IN (
	SELECT concord_id
	FROM user_concorduser
	GROUP BY concord_id
	HAVING COUNT(user_id) > 1
)
GROUP BY c.concord_id, u.id, u.username
ORDER BY concord_id, projects DESC;
```

concord_id | id | username | projects
-- | -- | -- | --
**397987** | **11005** | **ebrownson.concord** | **8**
397987 | 12722 | ebrownson.concord1 | 0
397987 | 12725 | ebrownson.concord4 | 0
397987 | 12721 | ebrownson.concord0 | 0
397987 | 12723 | ebrownson.concord2 | 0
397987 | 12724 | ebrownson.concord3 | 0
**526224** | **10653** | **hhood.concord** | **4**
526224 | 12738 | hhood.concord0 | 0
588729 | 12743 | WSTEMKline.concord2 | 0
588729 | 10926 | WSTEMKline.concord | 0
588729 | 12742 | WSTEMKline.concord1 | 0
588729 | 12739 | WSTEMKline.concord0 | 0

As can be seen above, some users that have tried to login have been unable to, and have had multiple accounts made for them.

I'm going to delete the ones with 0 projects when this is deployed, so that those users can login correctly.

PR #3486, opened 2022-02-22
Fix Concord Login

rajadain:
## Overview

Users of the educational platform Concord Learn Portal have been unable to login since the Django upgrade. This happened because the signature that Django looks for in custom backends changed. By changing the signature of our custom backends to bring them back in line with what Django expects, we ensure that Concord users can log in again.

Note that this targets `release/1.33.3`, not `develop`.

Connects #3467 

### Setup

* Copy the secret password from LastPass for `oAuth (MMW_ITSI_SECRET_KEY / MMW_CONCORD_SECRET_KEY)`
* Go into the `app` VM and populate the `MMW_CONCORD_SECRET_KEY` variable:
    ```bash
    vagrant ssh app -c 'echo *** | sudo tee /etc/mmw.d/env/MMW_CONCORD_SECRET_KEY'
    ```
* Restart the `mmw-app` service so it picks up the new variable:
    ```bash
    vagrant ssh app -c 'sudo service mmw-app restart'
    ```
* Go to http://localhost:8000/ and ensure you are not logged in
* Try to log in using the "Learn Portal" link
  - In the Learn Portal, use the credentials from the `MMW Teacher Learn Portal Concord` entry in LastPass
  - [x] Ensure a new account is created correctly for you
* Log out of the account, and try to log back in using Lean Portal again
  - [x] Ensure you are able to log back in to the same account

### DIFF
--- a/src/mmw/apps/user/backends.py
+++ b/src/mmw/apps/user/backends.py
@@ -1,11 +1,12 @@
 # -*- coding: utf-8 -*-
 from django.core.exceptions import ObjectDoesNotExist
+from django.contrib.auth.backends import BaseBackend
 from django.contrib.auth.models import User
 
 from apps.user.models import ItsiUser, ConcordUser
 
 
-class SSOAuthenticationBackend(object):
+class SSOAuthenticationBackend(BaseBackend):
     """
     A custom authentication back-end for Single Sign On providers.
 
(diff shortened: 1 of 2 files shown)


emilyhu0106 (approved):
Very excited to see this special part of MMW and learn about custom Django backend module

Issue #3490, opened 2022-03-02
Ensure we use ms_pointsource_drb when in DRB


rajadain:
BME has noted that when running MapShed in DRB-specific HUC-8s, we gets results from `ms_pointsource`, not from `ms_pointsource_drb`. `ms_pointsource_drb` seems to be used in both Analyze and Subbasin for those shapes. Notably, Cohaney-Maurice HUC-08 has 0 point source loads in MapShed, but significant ones in Subbasin.

Investigate and ensure we use `ms_pointsource_drb` at all times for all shapes within the DRB.

rajadain:
I have deployed https://github.com/WikiWatershed/model-my-watershed/tree/tt/ensure-correct-pointsources to staging and have asked BME to test there. If it is successful, then will open a PR.

PR #3501, opened 2022-03-15
Ensure Correct Point Sources Are Used For MapShed

rajadain:
## Overview

We use `DRB_SIMPLE_PERIMITER` in most cases, and had missed this one. Because of this incongruity, certain HUC-8s had significantly lower Point Source Pollution values, since they were using the `ms_pointsource` dataset instead of the `ms_pointsource_drb`.

By fixing this, we ensure that the correct dataset is used when needed.

Connects #3490 

### Demo

image

### Notes

The second commit is to re-enable Recent Weather on staging, after it was disabled for production in the 1.33.6 release.

## Testing Instructions

* Check out this branch and
    ```shell
    ./scripts/bundle.sh --debug
    ```
* Go to http://localhost:8000/ and select the Cohansey-Maurice HUC-08.
* Run the Watershed Multi-Year Model on it
  - [ ] In the Water Quality tab, ensure there's non-zero values for Nitrogen and Phosphorus under Point Sources
* Compare the same shape on production https://modelmywatershed.org
  - [ ] Ensure that has all 0s


rajadain:
Thanks for reviewing!

Issue #3484, opened 2022-02-18
Add Input Validation, Tests for GWLF-E Endpoints


rajadain:
The inputs for [`gwlf-e/prepare`](https://github.com/WikiWatershed/model-my-watershed/blob/4c8714ba7acdcd91118c1a04711a7a5115e26ec1/src/mmw/apps/geoprocessing_api/views.py#L1383-L1420) and [`gwlf-e/run`](https://github.com/WikiWatershed/model-my-watershed/blob/4c8714ba7acdcd91118c1a04711a7a5115e26ec1/src/mmw/apps/geoprocessing_api/views.py#L1423-L1454) endpoints should be validated.

- [x] Validate the inputs for `gwlf-e/prepare`, ensuring that `area_of_interest` and `wkaoi` and `huc` are not specified together, and that `layer_overrides` has valid values
- [x] Validate the inputs of `gwlf-e/run`, ensuring that `input` and `job_uuid` are not specified together, and that `input` has valid keys
- [ ] Add tests for these endpoints encoding correct behavior

Validating the values of `input` may be too complicated and extraneous, since crashes further down will report any errors there.

rajadain:
`layer_overrides` keys should be validated against:

https://github.com/WikiWatershed/model-my-watershed/blob/4c8714ba7acdcd91118c1a04711a7a5115e26ec1/src/mmw/mmw/settings/base.py#L474-L488

and the values should be validated against:

https://github.com/WikiWatershed/model-my-watershed/blob/4c8714ba7acdcd91118c1a04711a7a5115e26ec1/src/mmw/apps/geoprocessing_api/schemas.py#L194-L201

https://github.com/WikiWatershed/model-my-watershed/blob/4c8714ba7acdcd91118c1a04711a7a5115e26ec1/src/mmw/apps/geoprocessing_api/schemas.py#L227

rajadain:
Keys for the `gwlf-e/run` endpoint:

```json
[
  "AEU",
  "AWMSGrPct",
  "AWMSNgPct",
  "Acoef",
  "AgLength",
  "AgSlope3",
  "AgSlope3To8",
  "AnimalDailyN",
  "AnimalDailyP",
  "AnimalFlag",
  "AnimalName",
  "AntMoist",
  "Area",
  "AttenFlowDist",
  "AttenFlowVel",
  "AttenLossRateN",
  "AttenLossRateP",
  "AttenLossRatePath",
  "AttenLossRateTSS",
[...]

PR #3504, opened 2022-03-28
Add Validation to GWLF-E Endpoints

emilyhu0106:
## Overview
We have created the [`gwlf-e/prepare`](https://github.com/WikiWatershed/model-my-watershed/blob/4c8714ba7acdcd91118c1a04711a7a5115e26ec1/src/mmw/apps/geoprocessing_api/views.py#L1383-L1420) and [`gwlf-e/run`](https://github.com/WikiWatershed/model-my-watershed/blob/4c8714ba7acdcd91118c1a04711a7a5115e26ec1/src/mmw/apps/geoprocessing_api/views.py#L1423-L1454) endpoints to make previously internal APIs public, making running GWLF-E models easier. In order to help our users using the correct/recommended form of input, we add the following validations to the `gwlf-e/prepare/` and `gwlf-e/run/` endpoints:

For `/prepare/`
1. `area_of_interest`, `wkaoi`, and `huc` are not specified together
2. The keys in `layer_overrides` are within the standard layer overrides
options
3. The value of `__LAND__` and `__STREAMS__` layer overrides are within
its valid range

For `/run/`
1. `input` and `job_uuid` are not specified together
2. `input` has all the required valid keys to run the modeling

Connects #3484

### Demo

Examples of validation:

1. `/gwlf-e/prepare/`

Data object
```
{
"wkaoi":"huc12_9466",  
  "layer_overrides": {
    "__LAND__": "nlcd-2019-30m-epsg5070-512-byte",
    "__STREAMS__": "nhdh"
  }
}
```
Response
```
[
  "Invalid __STREAMS__ param: Must be `drb`, `nhdhr`, or `nhd`"
]
```
Data object
```
{
"wkaoi":"huc12_9466",  
  "layer_overrides": {
    "__LAND": "nlcd-2019-30m-epsg5070-512-byte",
    "__STREAMS": "nhdh"
  }
}
```
 Response
```
[
  "These layers are not standard layers for layer overrides: __LAND __STREAMS "
]
```

2. `/gwlf-e/run/`

Data object
```
{
  "input": {
    "NRur": 10
   }
}
```
Response
```
[
  "Invalid input: Please use the full result of gwlf-e/prepare endpoint's result object"
]
```

### Notes

 - In the original scope of #3484, tests for these two endpoints were included. We [decide](https://github.com/WikiWatershed/model-my-watershed/pull/3504#issuecomment-1082376223) to add tests in a separate issue since some tables related to modeling are not in the test database yet. 
 - We are not validating the value of the `input` in `/run/` since there are more than 400 keys in the `input` object

## Testing Instructions

 * Check out this branch and go to http://localhost:8000/api/docs/
 * Start a `gwlf-e/prepare` job with all kinds of parameters, including but not limit to:
    * Having different combinations of geographic inputs (could be zero, one, or more types)
    * In `"layer_overrides"`, with correct and wrong types of keys and values
    - [ ] Make sure all error codes and error messages make sense and provide useful information for users to correct the input
 * Start a `gwlf-e/run` job with all kinds of parameters, including but not limit to:
    * With or without an `job_uuid`
    * Different forms of `input` (missing some default keys, adding keys beside the default ones, wrong spelling of the default keys, empty object, etc )
    - [ ] Make sure all error codes and error messages make sense and provide useful information for users to correct the input

### DIFF
--- a/src/mmw/apps/geoprocessing_api/validation.py
+++ b/src/mmw/apps/geoprocessing_api/validation.py
@@ -102,6 +102,18 @@ def validate_gwlfe_prepare(data):
         raise ValidationError(error)
 
 
+def validate_gwlfe_run(input, job_uuid):
+    if not check_gwlfe_only_one([input, job_uuid]):
+        error = ('Invalid parameter: Only one type of prepared input'
+                 '(input JSON or job_uuid) is allowed')
+        raise ValidationError(error)
+
+    if not check_gwlfe_run_input(input):
+        error = ("Invalid input: Please use the full result "
+                 "of gwlf-e/prepare endpoint's result object")
+        raise ValidationError(error)
+
+
 def check_gwlfe_only_one(params):
     if sum(map(check_is_none, params)) == 1:
         return True
@@ -130,6 +142,11 @@ def check_land_layer_overrides(layers):
 
 def check_streams_layer_overrides(layers):
     return layers['__STREAMS__'] in STREAM_LAYER_OVERRIDES
+
+
+def check_gwlfe_run_input(input):
+    result = all(el in input for el in settings.GWLFE_DEFAULTS.keys())
+    return result
 
 
 def create_layer_overrides_keys_not_valid_msg(layers):
--- a/src/mmw/apps/geoprocessing_api/views.py
+++ b/src/mmw/apps/geoprocessing_api/views.py
@@ -45,7 +45,8 @@ from apps.geoprocessing_api.throttling import (BurstRateThrottle,
 
 from apps.geoprocessing_api.validation import (validate_rwd,
                                                validate_uuid,
-                                               validate_gwlfe_prepare)
+                                               validate_gwlfe_prepare,
+                                               validate_gwlfe_run)
 
 
 @swagger_auto_schema(method='post',
@@ -1707,11 +1708,12 @@ def _parse_gwlfe_input(request, raw_input=True):
         model_input = request.data.get('input')
 
         if model_input:
-            # TODO #3484 Validate model_input
+            validate_gwlfe_run(model_input, job_uuid)
             return model_input, job_uuid, mods, hash
 
     if not job_uuid:
-        raise ValidationError('`job_uuid` must be specified')
+        raise ValidationError('Either `input` or `job_uuid` '
+                              'must be specified.')
 
     if not validate_uuid(job_uuid):
         raise ValidationError(f'Invalid `job_uuid`: {job_uuid}')
@@ -1727,5 +1729,6 @@ def _parse_gwlfe_input(request, raw_input=True):
 
     model_input = json.loads(input_job.result)
 
-    # TODO #3484 Validate model_input
+    validate_gwlfe_run(model_input, job_uuid)
+
     return model_input, job_uuid, mods, hash


src/mmw/apps/geoprocessing_api/views.py
```diff
@@ -1399,8 +1402,13 @@ def start_modeling_gwlfe_prepare(request, format=None):
     user = request.user if request.user.is_authenticated else None
     area_of_interest, wkaoi = _parse_modeling_input(request)
 
+    aoi_input = request.data.get('area_of_interest')
+    wkaoi_input = request.data.get('wkaoi')
+    huc_input = request.data.get('huc')
     layer_overrides = request.data.get('layer_overrides', {})
 
+    validate_gwlfe_prepare(aoi_input, wkaoi_input, huc_input, layer_overrides)
```
rajadain, line 1410: Consider moving this to be within `_parse_modeling_input`

rajadain (commented):
Left some early comments.

src/mmw/apps/geoprocessing_api/views.py
```diff
@@ -1671,8 +1671,9 @@ def _parse_aoi(data):
     serializer.is_valid(raise_exception=True)
     area_of_interest = serializer.validated_data.get('area_of_interest')
     wkaoi = serializer.validated_data.get('wkaoi')
+    layer_overrides = data.get('layer_overrides', {})
```
emilyhu0106, line 1674: It is a bit redundant since line 1662 has already got this object. I didn't pass it as a parameter to `_parse_aoi` since there are other places (e.g. analyze) that uses this function. Not very sure what the best practice is here

emilyhu0106:
I wrote the following test for the `gwlf-e/prepare` endpoint
```
class ExerciseModelingGwlfeApi(TestCase):
    PREPARE_URL = reverse('geoprocessing_api:start_modeling_gwlfe_prepare')
    RUN_URL = reverse('geoprocessing_api:start_modeling_gwlfe_run')

    def setUp(self):
        self.c = APIClient()
        self.test_user = User.objects.create_user(username='test',
                                                  email='test@azavea.com',
                                                  password='test')
        self.c.login(username='test', password='test')

    def get_logged_in_session(self, username, password):
        c = Client()
        c.login(username=username,
                password=password)
        return c

    def test_gwlfe_prepare(self):
        print(self.PREPARE_URL)

        response = self.c.post(self.PREPARE_URL, { 'huc': '020402031008' })

        print(response)

        self.assertEqual(response.status_code, 200)
```
and got error message
```
File "/usr/local/lib/python3.8/dist-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
django.db.utils.ProgrammingError: relation "boundary_huc12" does not exist
LINE 3:           FROM boundary_huc12
                       ^


----------------------------------------------------------------------
Ran 1 test in 0.441s
```
stating the `huc12` table is not included in the test database at this moment. Adding the table to test database or having integrated tests are beyond the scope for this issue. We are going to further explore adding tests to GWLF-E endpoints in a separate issue.

emilyhu0106:
I'm planning to squash the `fixup` commits and add more context to the commit messages when the code itself are ready to go. Please feel free to let me know if squashing is preferred to be done before reviewing.

src/mmw/apps/geoprocessing_api/views.py
```diff
@@ -1671,8 +1671,9 @@ def _parse_aoi(data):
...
 
-    return area_of_interest, wkaoi
+    return area_of_interest, wkaoi, layer_overrides
```
rajadain, line 1676: Adding a third return value here breaks all the `analyze` endpoints, which only expect two values:

https://github.com/WikiWatershed/model-my-watershed/blob/9b1aebada97f14480a3fdbab70491ab24e8d8993/src/mmw/apps/geoprocessing_api/views.py#L524

Since `layer_overrides` only makes sense for modeling endpoints and not analyze ones (where analyzing a different layer is a path parameter, e.g. land and streams), we should return it from `_parse_modeling_input`.

rajadain (changes requested):
Consider a refactor where the entire `request.data` object is given to the validation function:

```python
def validate_gwlfe_prepare(data):
    area_of_interest = request.data.get('area_of_interest')
    wkaoi = request.data.get('wkaoi')
    huc = request.data.get('huc')
    layer_overrides = request.data.get('layer_overrides', {})

    if not check_gwlfe_only_one([area_of_interest, wkaoi, huc]):
        ...
```

Then `_parse_modeling_input` can be something like:

```python
def _parse_modeling_input(request):
    validate_gwlfe_prepare(request.data)
    layer_overrides = request.data.get('layer_overrides', {})
    area_of_interest, wkaoi = _parse_aoi(request.data)

    return area_of_interest, wkaoi, layer_overrides
```

The `layer_overrides` bit is a little repetitive, but don't think there's an elegant way around that, so let's do what's simplest, and get it again, as above.

rajadain:
I think we should also validate the input here:

https://github.com/WikiWatershed/model-my-watershed/blob/9b1aebada97f14480a3fdbab70491ab24e8d8993/src/mmw/apps/geoprocessing_api/views.py#L1719-L1722

Even though this is generated by MMW, we should still validate it just the same, since arbitrary errors can happen at any point in the chain. Best to treat all code paths the same.

rajadain:
Will take another look once the above has been resolved.

emilyhu0106:
Ready for another look!

rajadain:
Taking another look

src/mmw/apps/geoprocessing_api/validation.py
```diff
@@ -69,3 +71,97 @@ def validate_uuid(uuid):
...
+
+    not_valid_layers = check_layer_overrides_keys(layer_overrides)
+
+    if not_valid_layers:
+        error = create_layer_overrides_keys_not_valid(not_valid_layers)
+        raise ValidationError(error)
+
+    if ('__LAND__' in layer_overrides.keys() and
+            not check_land_layer_overrides(layer_overrides)):
+        error = 'Invalid __LAND__ param: Must be one of the NLCD values'
```
rajadain, line 95: Instead of saying "one of the NLCD values", consider specifying the allowed values in a list, as is done here:

https://github.com/WikiWatershed/model-my-watershed/blob/9b1aebada97f14480a3fdbab70491ab24e8d8993/src/mmw/apps/geoprocessing_api/schemas.py#L197

src/mmw/apps/geoprocessing_api/validation.py
```diff
@@ -69,3 +71,97 @@ def validate_uuid(uuid):
...
+    return layers['__STREAMS__'] in STREAM_LAYER_OVERRIDES
+
+
+def check_gwlfe_run_input(input):
+    result = all(el in input for el in settings.GWLFE_DEFAULTS.keys())
+    return result
+
+
+def create_layer_overrides_keys_not_valid(layers):
+    error = 'These layers are not standard layers for layer overrides: '
```
rajadain, line 152: Consider adding the valid values to this response as well:

```js
'These layers are not standard layers for layer overrides: "_LAND_". Must be one of: "__LAND__", "__STREAMS__"'
```

rajadain (commented):
Nice edits! Still seeing some edge cases:

```shell
http --verbose :8000/api/modeling/gwlf-e/run/ Authorization:"Token b0c671e1424f37e58c81670ce8118e8eff0f0c14" bad=key
```
```http
POST /api/modeling/gwlf-e/run/ HTTP/1.1
Accept: application/json, */*;q=0.5
Authorization: Token b0c671e1424f37e58c81670ce8118e8eff0f0c14
Connection: keep-alive
Content-Length: 14
Host: localhost:8000
User-Agent: HTTPie/3.1.0

{
    "bad": "key"
}


HTTP/1.1 400 Bad Request
Allow: OPTIONS, POST
Connection: keep-alive
Content-Length: 32
Date: Wed, 30 Mar 2022 20:24:54 GMT
Server: nginx
Vary: Accept, Cookie, Origin

[
    "`job_uuid` must be specified"
]
```

src/mmw/apps/geoprocessing_api/validation.py
```diff
@@ -69,3 +71,104 @@ def validate_uuid(uuid):
...
+
+
+LAND_LAYER_OVERRIDES = [
+     'nlcd-2019-30m-epsg5070-512-byte',
+     'nlcd-2016-30m-epsg5070-512-byte',
+     'nlcd-2011-30m-epsg5070-512-byte',
+     'nlcd-2006-30m-epsg5070-512-byte',
+     'nlcd-2001-30m-epsg5070-512-byte',
+     'nlcd-2011-30m-epsg5070-512-int8',
+ ]
```
rajadain, line 172:
```suggestion
]
```

src/mmw/apps/geoprocessing_api/validation.py
```diff
@@ -69,3 +71,104 @@ def validate_uuid(uuid):
...
+        raise ValidationError(error)
+
+    if ('__LAND__' in layer_overrides.keys() and
+            not check_land_layer_overrides(layer_overrides)):
+        error = create_layer_overrides_land_not_valid_msg(LAND_LAYER_OVERRIDES)
+        raise ValidationError(error)
+
+    if ('__STREAMS__' in layer_overrides.keys() is not None and
+            not check_streams_layer_overrides(layer_overrides)):
+        error = 'Invalid __STREAMS__ param: Must be `drb`, `nhdhr`, or `nhd`'
```
rajadain, line 100: Making this consistent with the other one:

```suggestion
        error = f'Invalid __STREAMS__ param: Must be one of {STREAM_LAYER_OVERRIDES}'
```

src/mmw/apps/geoprocessing_api/validation.py
```diff
@@ -69,3 +71,104 @@ def validate_uuid(uuid):
...
+
+def validate_gwlfe_run(input, job_uuid):
+    if not check_gwlfe_only_one([input, job_uuid]):
+        error = ('Invalid parameter: Only one type of prepared input'
+                 '(input JSON or job_uuid) is allowed')
+        raise ValidationError(error)
+
+    if not check_gwlfe_run_input(input):
+        error = ("Invalid input: Please use the full result "
+                 "of gwlf-e/prepare endpoint's result object")
```
rajadain, line 112: Let's make a follow up issue to call out specific keys that are missing from the payload.

rajadain (commented):
Left some edits for cleanup. With these, I think we can merge this, and make follow up cards for the rest.

rajadain (approved):
Alright this is in good shape. Nice job following all the different threads and getting this in!

src/mmw/apps/geoprocessing_api/validation.py
rajadain: Made #3509 for this.

Issue #3475, opened 2022-02-02
Switch MMW Front-End to Use Geoprocessing API for Mapshed, GWLF-E


rajadain:
Once #3472 and #3473 are in:

- [x] Switch the front-end to use the new public endpoints
- [x] Ensure that the modeling works correctly for anonymous and logged in users
- [ ] Remove the old internal endpoints
- [ ] Update any tests (consider moving them to the `geoprocessing_api` app if needed)

rajadain:
Need to move these endpoints to the public API under `subbasin` as well:

https://github.com/WikiWatershed/model-my-watershed/blob/9b1aebada97f14480a3fdbab70491ab24e8d8993/src/mmw/apps/modeling/views.py#L660-L682

and to switch the front-end to use the new APIs.

PR #3507, opened 2022-03-30
Switch to New APIs for Front-End

rajadain:
## Overview

Switch the front-end to use the new public APIs.

Connects #3475 

### Demo

image

### Notes

There's still a few endpoints left, will address them in a future PR.

## Testing Instructions

* Check out this branch and `bundle.sh --debug`
* Choose a HUC-8 or HUC-10 shape in the Eastern US
* Open the Network tab of Developer Tools
* Run the Watershed Multi-Year model, and from the Water Quality tab, run Subbasin modeling
* Observe the requests in the Network tab
  - [x] Ensure they go to the /api/modeling/ endpoints

### DIFF
--- a/src/mmw/apps/geoprocessing_api/views.py
+++ b/src/mmw/apps/geoprocessing_api/views.py
@@ -1542,7 +1542,13 @@ def start_modeling_subbasin_run(request, format=None):
     return Response(
         {
             'job': task_chain.id,
+            'job_uuid': task_chain.id,
             'status': JobStatus.STARTED,
+            # TODO Remove this message when `job` is deprecated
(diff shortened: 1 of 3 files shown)


emilyhu0106 (approved):
The implementation is great! Glad to see all the pieces coming together. I encountered the same SRAT issue when I run subbasin results on HUCs in California, but things works well in the Delaware River region. I'm wondering if it is better to include something like a tooltip to inform the users that getting subbasin attenuated results is for certain regions only?

rajadain:
Thanks for reviewing! I did mention the non-California support and we are working on getting those in.

Issue #3510, opened 2022-04-04
Switch Front-end Analyze to New API Format


rajadain:
The /analyze/ back-end was switched to a different format in #3503. While the changes are backwards compatible, our front-end should use the new style API.

Switch the front-end to use the new style API.

PR #3503, opened 2022-03-23
Align Analyze and Modeling APIs

rajadain:
## Overview

The `/modeling/` APIs take a JSON object as input, while `/analyze/` took a combination of GeoJSON and query parameters. This makes it so that `/analyze/` supports the `/modeling/` style API, while maintaining support for the old `/analyze/` style as well. Users of the old style are given warnings about changes in a new `messages` key. The Job Started Response has also been upgraded to include a `job_uuid`, which is more standard.

Connects #3497 

### Demo

New style:

```shell
xh --verbose :8000/api/analyze/soil/ Authorization:"Token b0c671e1424f37e58c81670ce8118e8eff0f0c14" huc=020402031001
```
```http
POST /api/analyze/soil/ HTTP/1.1
Accept: application/json, */*;q=0.5
Accept-Encoding: gzip, deflate, br
Authorization: Token b0c671e1424f37e58c81670ce8118e8eff0f0c14
Connection: keep-alive
Content-Length: 22
Content-Type: application/json
Host: localhost:8000
User-Agent: xh/0.13.0

{
    "huc": "020402031001"
}



HTTP/1.1 200 OK
Allow: POST, OPTIONS
Connection: keep-alive
Content-Type: application/json
Date: Wed, 23 Mar 2022 13:56:53 GMT
Location: /api/jobs/fbd6e474-cc20-4ff0-831a-c8e1a9c2aa36/
Server: nginx
Transfer-Encoding: chunked
Vary: Accept-Encoding
Vary: Accept, Cookie, Origin

{
    "job": "fbd6e474-cc20-4ff0-831a-c8e1a9c2aa36",
    "job_uuid": "fbd6e474-cc20-4ff0-831a-c8e1a9c2aa36",
    "status": "started",
    "messages": [
        "The `job` field will be deprecated in an upcoming release. Please switch to using `job_uuid` instead."
    ]
}
```

Old style:

```shell
xh --verbose POST :8000/api/analyze/soil/ Authorization:"Token b0c671e1424f37e58c81670ce8118e8eff0f0c14" huc==020402031001
```
```http
POST /api/analyze/soil/?huc=020402031001 HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate, br
Authorization: Token b0c671e1424f37e58c81670ce8118e8eff0f0c14
Connection: keep-alive
Host: localhost:8000
User-Agent: xh/0.13.0

HTTP/1.1 200 OK
Allow: POST, OPTIONS
[...]

PR #3516, opened 2022-04-06
Use new API format in Front-end Analyze

emilyhu0106:
## Overview

We switched the `/analyze/` endpoints to take a JSON object as input in #3503, while the old query param style is still compatible. This change makes the front-end Analyze feature uses the new JSON object style to keep them consistent. Instead of GeoJSON of the area of interest in the request body, and wkaoi and huc as query parameters, the new style send all in the request body, as `{ area_of_interest, wkaoi }`, which is consistent with the `/modeling/` endpoints.

Connects #3510

### Demo

image

## Testing Instructions

**Result with the old style as baseline**
 * Check out `develop`, go to http://localhost:8000/, select any HUC, and run Analyze
 * Download the result and name it `analyze-result-huc-old`
 * Click "Change Area " to start a new session. Use "Draw Area" to create a new area and run Analyze
 * Download the shape in geojson and name it `analyze-draw-shape`; Download the result and name it `analyze-result-draw-old`

**Result with the new style**
 * Check out this branch and run 
   ``` 
   ./scripts/bundle.sh --debug
    
   ```
 * Go to http://localhost:8000/, clear the cache if needed, select the same HUC, and then run Analyze
 * Open the Network tab, filter the requests by `analyze/` 
     - [ ] The payload in all filtered requests should be in JSON object style
image
     - [ ] There should be only one message in the response
 * Download the result and name it `analyze-result-huc-new`
     - [ ] Make sure the results -`analyze-result-huc-old.json` and `analyze-result-huc-new.json` - are the same
     - [ ] Check the tabs of different analyze result and all of the graphs and tables are loaded properly
 * Start a new session. Upload `analyze-draw-shape.geojson` 
 * Download the result and name it `analyze-result-draw-new`
     - [ ] Make sure the results -`analyze-result-draw-old.json` and `analyze-result-draw-new.json` - are the same
     - [ ] Check the tabs of different analyze result and all of the graphs and tables are loaded properly

### DIFF
--- a/src/mmw/js/src/analyze/models.js
+++ b/src/mmw/js/src/analyze/models.js
@@ -102,8 +102,9 @@ var AnalyzeTaskModel = coreModels.TaskModel.extend({
             var isWkaoi = utils.isWKAoIValid(wkaoi),
                 taskHelper = {
                     contentType: 'application/json',
-                    queryParams: isWkaoi ? { wkaoi: wkaoi } : null,
-                    postData: isWkaoi ? null : JSON.stringify(aoi)
+                    postData: isWkaoi ? 
+                        JSON.stringify({ wkaoi : wkaoi }) : 
+                        JSON.stringify({ area_of_interest : aoi })
                 },
                 promises = self.start(taskHelper);
 


rajadain:
I'm sorry I completely missed this PR! Will take a look today!

rajadain (approved):
Works well, nice job! Good detail in the testing instructions!

rajadain:
I will note that this PR branched off of another PR rather than from `develop`. Unless the work you're doing is directly building off of another PR and cannot be done without it, prefer always branching from the latest `develop`.

emilyhu0106:
> I will note that this PR branched off of another PR rather than from `develop`. Unless the work you're doing is directly building off of another PR and cannot be done without it, prefer always branching from the latest `develop`.

Sounds good! I'll pay attention to this. I think it was an accident this time. Is it still safe to merge to develop?

rajadain:
Yes, merging still works as is in this case, because it is a simple change and only one commit. With more commits and more complex changes, it may be more problematic.

You are good to merge this!

emilyhu0106:
Thank you for the review!

Issue #3511, opened 2022-04-04
Add GWLF-E endpoint tests


rajadain:
Test the behavior of the endpoint sans geoprocessing:

- [x] Test that bad inputs are handled as expected
- [ ] Test that modifications are applied as expected

Using CELERY_TEST_OVERRIDES may help:

https://github.com/WikiWatershed/model-my-watershed/blob/917b5c0f8ec63a1bdc456e38365fcb1a5eff332c/src/mmw/apps/modeling/tests.py#L135-L139

https://github.com/WikiWatershed/model-my-watershed/blob/917b5c0f8ec63a1bdc456e38365fcb1a5eff332c/src/mmw/apps/modeling/tests.py#L212-L213

PR #3518, opened 2022-04-11
Add Tests for New Geoprocessing API Endpoints

rajadain:
## Overview

Adds test for the new Geoprocessing API endpoints.

Most of these tests run the unsuccessful cases, because success runs require a large amount of mocking, from Celery to Geoprocessing Service to Data Tables that are not available in the test database.

There is also a test for applying modifications to existing GMS payloads. This is also not exhaustive, but demonstrative, and covers basic use.

Also improves the performance of some tests by switching from `LiveServerTestCase` to `TestCase`.

Connects #3511 

## Testing Instructions

* Check out this PR
* Run `./scripts/test.sh`
  - [x] Ensure it finishes successfully


emilyhu0106 (approved):
Tests are well designed and cover a wide variety of bad inputs. The change from `LiveServerTestCase` to `TestCase` to save time is also nice.

rajadain:
Thanks for reviewing!

Issue #3546, opened 2022-06-09
PA DEP Drainage Area Black box 1️⃣


rajadain:
Outline document: https://docs.google.com/document/d/1-O56QcVxKMqlW-26MnDm5HkYI71yMzQ80GQiZVV6oTE/

- [x] Create dummy endpoints for drainage area
- [x] Add Point-based drainage area draw tool
- [x] Add Stream-segment based drainage area draw tool
- [ ] Fix bugs

rajadain:
Repointing to :one:

PR #3551, opened 2022-06-23
Add Drainage Area Draw Tools

rajadain:
## Overview

Adds two new draw tools for Drainage Area, with dummy implementations. The first takes a point, and requests an area that drains to it. The second takes a polygon, drawn around streams, and requests and area that drains to those streams. The dummy implementation simply buffers the point or stream by 0.1 and returns the shape.

In upcoming changes, the source points / streams will be sent to the Drexel / ANS API, and we'll get the correct shapes from them.

The returned shape also has `drainage_area: true` property, which will be used to show special drainage area analysis.

Connects #3546 

### Notes

In the current implementation the new draw endpoints are synchronous. However, we may need to make them Celery task based, as are all the other analyze and model endpoints, since the Drexel API may take a while to run.

There is also more validation that needs to be done, and error handling added, so that validation errors are communicated clearly to the user. For example, if in your stream selection shape, there are no streams, there is currently no graceful error handling for that.

## Testing Instructions

* Go to http://localhost:8000/ and click Get Started
  - [x] Ensure you see a new Drainage Area set of draw tools
* Click the point-based drainage area draw tool
  - [x] Ensure it draws an oval around the point you clicked
  - [x] Ensure Analyze completes as usual
* Reset, and click the stream-based drainage area draw tool
  - [x] Ensure it turns on the streams layer for you to see where you're drawing
  - [x] Ensure after you draw the shape, it draws ovals around the streams in your selection
  - [x] Ensure Analyze completes as usual


emilyhu0106 (commented):
The new drainage area draw tools work well in general. Good job on this black box!

I noticed the app will return errors like this in these two scenarios:
image
 1. using point-based tool when not signed in  -> `403`
 2. using stream based tool before the stream layer is fully loaded, especially when zoomed out a bit -> polygon payload to endpoint causing `500`.

rajadain:
[01ae5a6](https://github.com/WikiWatershed/model-my-watershed/pull/3551/commits/01ae5a67e627d9b630ed8cc37e0e6a52dbd22892) should cover the first point above.

I'll take a look at the second one. The streams being displayed are just for the user's benefit, the intersection happens on the server side and should be unrelated. But perhaps when zoomed out the stream intersection is too big.

rajadain:
I was going to defer fixing the error box until later, but will see if I can easily capture it in this PR as well.

emilyhu0106 (approved):
All works well! I agree with deferring the error box fix and focus this PR in the drainage draw tool

emilyhu0106 (approved):
I tested the point-based and the stream-based tool - both works well! Nice job in refactoring it to a task. The points and segments are helpful for users.

src/mmw/js/src/draw/views.js
```diff
@@ -1050,6 +1064,185 @@ var WatershedDelineationView = DrawToolBaseView.extend({
...
+                    deferred.resolve(response.result);
+                },
+
+                pollFailure: function(response) {
+                    self.model.set({
+                        pollError: true,
+                        polling: false,
+                    });
+                    console.error(response.error);
+                    deferred.reject('Drainage area could not be calculated');
```
emilyhu0106, line 1222: I noticed that when the polygon is not intersecting with any streams, the app will return the error here, as shown below:


I'm thinking would it be helpful if we add something like "the drainage area must intersect with streams" to help first-time users?

rajadain: Thanks for prompting this. Fixed in eb2c372:

image

rajadain:
Thanks for the detailed review and continued encouragement for better user experience!

PR #3555, opened 2022-07-22
Add analysis for drainage areas

rajadain:
## Overview

This takes an input AoI, finds its centroid, finds a HUC-12 that contains the centroid, and runs GWLF-E on it. Ultimately, we will scale the GWLF-E results to the input AoI, but for that we need the Drexel / ANS API. So this is not complete until that API has been integrated.

Connects #3553 

## Testing Instructions

* Check out this branch
* Go to http://localhost:8000/ and draw any drainage area
* Download the resulting AoI as GeoJSON
    image
* Copy your API token
* Exercise the new API, like so:
    ```console
    http :8000/api/analyze/drainage-area/ Authorization:"Token $YOUR_TOKEN" < $DOWNLOADED_FILE.geojson
    ```
  - [x] Ensure the call succeeds
  - [x] Ensure the job results are GWLF-E outputs of Loads and monthlies

PR #3559, opened 2022-08-02
Add Drainage Area Analyze View

### LINKED ISSUE #3557: Add Drainage Area Analyze View
- [x] This should only be visible if the area of interest is a Drainage Area
- [x] This should trigger the `/analyze/drainage-area/` endpoint made above
- [x] This should show the results of the drainage area in Hydrology and Water Quality tabs, a-la GWLF-E
    - Verify this with BME

rajadain:
## Overview

When the area of interest is a Drainage Area, we show this additional analysis that shows the GWLF-E output of the HUC-12 to which the area of interest's centroid belongs. For other areas of interest, this analysis is not shown.

Once the Drexel / ANS API integration is in place, the numbers will be scaled to the land use distribution of the area of interest. See #3558 for details.

Because of the number of analyses, the Drainage Area addition forces the tabs to overflow to a new line. This is less than ideal. However, this is likely the best we can do at this point, for a larger redesign is out of scope. Thankfully, this only happens for Drainage Areas, and other areas of interest are not affected.

Closes #3557 

### Demo

image

image

## Testing Instructions

* Check out this branch
* Bundle the scripts
    ```
    ./scripts/bundle.sh --debug
    ```
* Restart celery
    ```
    ```
* Go to http://localhost:8000/
* Draw a non-drainage area shape
  - [ ] Ensure the analysis does not include a Drainage Area result
* Go back. Draw a Point-based Drainage Area shape
  - [ ] Ensure the analysis does include a Drainage Area result
  - [ ] Ensure it has Hydrology and Water Quality options in the dropdown with reasonable (not null) values
  - [ ] Ensure the CSV export works for the Hydrology table and both Water Quality tables
* Go back. Draw a Stream-based Drainage Area shape
  - [ ] Ensure it works the same as above
* Go to Model, select Watershed Multi-Year Model
* Check the Hydrology and Water Quality tabs
  - [ ] Ensure they work the same as before

### DIFF
--- a/src/mmw/apps/geoprocessing_api/tasks.py
+++ b/src/mmw/apps/geoprocessing_api/tasks.py
@@ -584,3 +584,18 @@ def draw_drainage_area_stream(input):
         'area_of_interest': area_of_interest,
         'stream_segment': json.loads(segment.geojson),
     }
+
+
+@shared_task
+def wrap_in_survey(result):
+    """
+    Takes a result, and wraps it in a dict with a `survey` key
+
+    This is used to make results compatible with /analyze/ API, which always
+    wraps results in a `survey`.
+
+    Example:
+
+    wrap_in_survey({'a': 1, 'b': 2}) => {'survey': {'a': 1, 'b': 2}}
+    """
+    return {'survey': result}
--- a/src/mmw/apps/geoprocessing_api/views.py
+++ b/src/mmw/apps/geoprocessing_api/views.py
@@ -1654,6 +1654,7 @@ def start_analyze_drainage_area(request, format=None):
         collect_data.s(huc12_geojson),
         run_gwlfe.s(inputmod_hash=''),
         # TODO Scale results to Drainage Area
+        tasks.wrap_in_survey.s(),
     ], area_of_interest, user)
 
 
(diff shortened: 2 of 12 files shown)


rajadain:
- Fixed a comment in [3068c1d](https://github.com/WikiWatershed/model-my-watershed/pull/3559/commits/3068c1d3c9bfda4d1aa5b6512029f79953b8ffe4)
- Removed unused chart region in [5d224b4](https://github.com/WikiWatershed/model-my-watershed/pull/3559/commits/5d224b4f5e979e76780db2034212a0c5cc599567)
- Added reference to TODO item in [0eda2dd](https://github.com/WikiWatershed/model-my-watershed/pull/3559/commits/0eda2dd182cfc47606f2c969ea41689cde05cbfe)

emilyhu0106 (approved):
All the analysis and download features work well for both point-based and stream-based analysis! Very nice job for finishing the front-end

rajadain:
Thanks for the review! After talking with the clients, they no longer want to have a special analysis for Drainage Areas. It is too disruptive a flow, and creates special cases for certain AoIs, which doesn't fit the established patterns of the app. It also runs a model in the Analyze stage which is confusing.

As a result, this feature is being scrapped. We will still keep the Drainage Area drawing based on the Drexel / ANS API (so far), and I will remove the Analyze endpoint added in #3555. Will confirm with clients early next week.

PR #3567, opened 2022-10-04
Update Provisioning, README

rajadain:
## Overview

We use Firefox for JS testing. Every 10 versions we have to update the version string. Firefox has been stable for the last few years, and the tests have not broken. By unversioning it we will get the latest version every time, which should be fine given the stability over the last few years.

PostGIS 3.3 is needed because a step in provisioning requests `ST_TriangulatePolygon`, which is only available in PostGIS 3.3. It is unclear what is requesting that. This does not have an impact on us, because we're not using any feature that is deprecated in PostGIS 3.3.

Also instruct to upgrade base box.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,9 +52,10 @@ echo "
 * 33.33.0.0/16" | sudo tee /etc/vbox/networks.conf
 ```
 
-Next, use the following command to bring up a local development environment:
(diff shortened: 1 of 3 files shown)


rachelekm (approved):
Looks great 👍 tracks with my notes and succeeds when I run `vagrant reload --provision `

rajadain:
Thanks for taking a look!

PR #3572, opened 2022-10-14
Make NHD Medium Resolution Default, Don't Analyze NHD High Resolution Until Selected

### LINKED ISSUE #3568: Make NHD Hi Res Analysis Optional
Currently, all users of MMW run NHD Hi Res analysis for every shape they select. This is a very expensive operation that leads to very high database usage, for something that not every user wants.

Add the ability for some analyses to be optional, or lazy loaded, so they are not run until a user explicitly selects them.

Make NHD Medium Resolution the default analysis in the Streams tab. Make NHD Hi Res an optional analysis, that should only run if a user selects it from the dropdown.

rajadain:
## Overview

To lower usage loads, we're going to make NHD HR analysis lazy, so it will only be triggered when users explicitly select its tab. By default only the Medium Resolution will be shown.

Adds the concept of "lazy" analyses that are not fetched by default. When they are brought in to focus, the fetch is triggered. If the analysis has already been fetched, it will not be fetched again.

Closes #3568 

### Demo

https://user-images.githubusercontent.com/1430060/195915125-8d48d182-3d7d-4917-88e4-a1a0836bd292.mp4

## Testing Instructions

- Check out this branch and `./scripts/bundle.sh --debug`
- Draw a shape, move to Analyze
  - [x] Ensure that NHD Medium Resolution is the default analysis under Streams
  - [x] Ensure that NHD High Resolution is not analyzed until selected from the dropdown
  - [x] Ensure that you if you select another analysis, and come back NHD Hi Res, the results are already there, and no re-fetch is triggered

### DIFF
--- a/src/mmw/js/src/analyze/models.js
+++ b/src/mmw/js/src/analyze/models.js
@@ -220,18 +220,18 @@ function createAnalyzeTaskGroupCollection(aoi, wkaoi) {
             displayName: "Streams",
             tasks: [
                 {
-                    name: "streams_nhdhr",
-                    displayName: "NHD High Resolution Streams",
-                    area_of_interest: aoi,
-                    wkaoi: wkaoi,
-                    taskName: "analyze/streams/nhdhr"
-                },
-                {
                     name: "streams_nhd",
                     displayName: "NHD Medium Resolution Streams",
                     area_of_interest: aoi,
                     wkaoi: wkaoi,
-                    taskName: "analyze/streams/nhd"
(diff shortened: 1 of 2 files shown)


rachelekm (approved):
Works great! All the testing instruction expectations were met. The Networks tab of my browser dev tools is showing 95.9 kB transferred over network and 950 kB resources loaded by the page with the medium resolution as the default, compared to the 398 kB transferred and 1.4 MB resources on staging 👍

rajadain:
Thanks for reviewing!

PR #3573, opened 2022-10-15
Unversion Firefox, Upgrade PostGIS to 3.3

rajadain:
## Overview

We use Firefox for JS testing. Every 10 versions we have to update the version string. Firefox has been stable for the last few years, and the tests have not broken. By unversioning it we will get the latest version every time, which should be fine given the stability over the last few years.

PostGIS 3.3 is needed because a step in provisioning requests ST_TriangulatePolygon, which is only available in PostGIS 3.3. It is unclear what is requesting that. This does not have an impact on us, because we're not using any feature that is deprecated in PostGIS 3.3.

This is a copy of #3567, targeting the `hotfix/1.33.8` branch, to get CI to run correctly.

### Demo

http://civicci01.internal.azavea.com/view/mmw/job/model-my-watershed-pull-requests/4468/console

### DIFF
--- a/deployment/ansible/group_vars/all
+++ b/deployment/ansible/group_vars/all
@@ -20,7 +20,7 @@ postgresql_support_repository_channel: "main"
 postgresql_support_libpq_version: "13.*.pgdg20.04+1"
 postgresql_support_psycopg2_version: "2.8.*"
 postgis_version: "3"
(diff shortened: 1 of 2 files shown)


rajadain:
Since the build is passing, as linked to in the demo above, and since the original PR was already reviewed and approved, I'm going to merge this in.

PR #3575, opened 2022-10-17
Switch to using correctly aligned NLCD19 layers, wipe cached results

### LINKED ISSUE #3571: Wipe Incorrect Geoprocessing Results for NLCD19
In the initial release of NLCD19 data, the layers were not properly aligned https://github.com/WikiWatershed/mmw-geoprocessing/issues/102. As a result, all calculations done using it are slightly incorrect.

This alignment was fixed in #3549, but the geoprocessing results made using incorrect alignment are still in the database.

Write a migration similar to https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/migrations/0040_clear_nlcd2019_tr55_results.py that clears out the stored results for Mapshed projects after `2022-01-17`. These results should be recalculated when they are reopened.

rajadain:
## Overview

Copy of #3549 for the hotfix.

Also adds a migration to clear out existing results for GWLF-E projects that used the unaligned 2019 layers since 1.33.0 was released in January 2022. All these projects will have their values recalculated when they are accessed again, allowing us to rebuild the project results over time rather than all at once and overloading the system.

Closes #3571 

## Testing Instructions

* On `hotfix/1.33.8`, go to http://localhost:8000/ and login with an account
  * Make one if you don't have one already
* Select / draw a shape, move to Analyze, move to Model, and select Watershed Multi-Year Model
* Once the calculation completes, create an additional scenario
* Refresh the page
  * [x] Ensure the results load quickly, without being recalculated
* Go to the database console
    ```
    ./scripts/manage.sh db_shell
    ```
* Run this query to check that your project has a true value for `gis_data`:
    ```sql
    SELECT gis_data FROM modeling_project WHERE id = XXX;
    ```
  where XXX is your project's id
* Check out this branch and run migrations
    ```
    ./scripts/manage.sh migrate
    ```
* Run this query to check that `gis_data` is now NULL:
    ```sql
    SELECT gis_data FROM modeling_project WHERE id = XXX;
    ```
  where XXX is your project's id
* Go back to your project in the UI
  * [x] Ensure the results are recalculated, evidenced by the loading spinner
  * [x] Ensure that the calculations finish correctly, with values for all tables
* Refresh the page
  * [x] Ensure the results load quickly, without being recalculated
* Run this query to check that your project has a true value for `gis_data`:
    ```sql
    SELECT gis_data FROM modeling_project WHERE id = XXX;
    ```
  where XXX is your project's id


rachelekm (approved):
Thanks for such thorough testing instructions! This looks great. I was able to get the new recalculated results after applying the migration and confirming each query result and UI functions were as expected.

rajadain:
Thanks for reviewing!

Issue #3515, opened 2022-04-05
Increase Worker VM EBS Volume Size on EC2


rajadain:
The Worker VMs on EC2 are full as soon as they start:

```
System load:  0.0               Users logged in:                  0
Usage of /:   99.2% of 7.69GB   IPv4 address for br-68637d7cab34: 172.18.0.1
Memory usage: 29%               IPv4 address for docker0:         172.17.0.1
Swap usage:   0%                IPv4 address for ens3:            10.0.1.229
Processes:    138
```

This leads to runtime issues, including extraneous SIGKILLs to `snapd.service`.

Most of this is Docker. But as logs get bigger, they also take up room.

Make it so the Worker VMs are provisioned with 10GB of default space.

rajadain:
For more context see: https://azavea.slack.com/archives/C044CG212/p1649176958355409

rajadain:
For an example of where this was done previously, see:

https://github.com/azavea/cac-tripplanner/blob/20d0f66ad2b6af23e1f08cc900de5b003c098db5/deployment/packer/cac.json#L46-L70

rajadain:
Getting these messages again, going to remake them.

rajadain:
Also consider adding log rotation so that the disk does not fill up as fast.

JN-Hernandez:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Apps-62c5ba9ee84ced001f5b5d00/issues/WikiWatershed/model-my-watershed/3515?planning-poker) with Zenhub @JN-Hernandez

JN-Hernandez:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Apps-62c5ba9ee84ced001f5b5d00/issues/WikiWatershed/model-my-watershed/3515?planning-poker) with Zenhub @caseycesari

rajadain:
This is now happening roughly weekly. Going to address this in this sprint.

PR #3585, opened 2022-12-26
Prevent Worker Crashes

rajadain:
## Overview

Configures `logrotate` to compress Celery logs, and delete those older than 3 days, to prevent the worker instance running out of space.

Also increases the EBS Volume size on workers make sure they have more space.

Connects #3515 

### Demo

image

Worker system before the change:

```console
$ df -h .

Filesystem      Size  Used Avail Use% Mounted on
/dev/root       7.6G  7.5G  141M  99% /
```

After the change:

```console
$ df -h .

Filesystem      Size  Used Avail Use% Mounted on
/dev/root        12G  7.3G  4.3G  64% /
```

## Testing Instructions

* Use the `mmw-stg` credentials in BitWarden to log in to AWS website
* Inspect the **Storage** of the [running worker instances](https://us-east-1.console.aws.amazon.com/ec2/home?region=us-east-1#Instances:instanceState=running;tag:Name=Worker;v=3;$case=tags:true%5C,client:false;$regex=tags:false%5C,client:false)
  * [x] Ensure it says 12G instead of 8

### DIFF
--- a/deployment/ansible/roles/model-my-watershed.celery-worker/tasks/celery-service.yml
+++ b/deployment/ansible/roles/model-my-watershed.celery-worker/tasks/celery-service.yml
@@ -15,3 +15,8 @@
     special_time: daily
     job: /usr/sbin/service celeryd restart
     state: present
+
+- name: Configure Celery log rotation
+  template: src=celery-logrotate.j2
+            dest=/etc/logrotate.d/celery
+            mode=0755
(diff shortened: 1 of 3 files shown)


JN-Hernandez (approved):
Looks good - checked staging and see the changes were applied successfully.
My only note: since this is a stop gap, can you either link the issues that address finding/fixing root cause, or spin up an issue to track that?  I just don't want it to fall off our radar.  Thanks!

rajadain:
My best guess for the root cause is increased size of dependencies gradually over the years. The detail on this PR and on the related issue #3515 should serve as good clues for if this starts to happen again. I don't expect it to, at least for a while.

PR #3603, opened 2023-11-27
Configurability Improvements

rajadain:
## Overview

Makes MMW more configurable by:

1. Explicitly specifying which S3 bucket to use for raster data
2. Initializing the geoprocessing service with the same environment variables as used for the main application so they have the same settings
3. Ensuring that the geoprocessing service and the Celery task management service use the same timeout value

Also adds a developer-experience improvement by making the tile cache purging explicit, which would otherwise always need to be manually removed in development.

### Notes

Similar to https://github.com/WikiWatershed/mmw-geoprocessing/pull/104, this does not change any actual runtime behavior, but allows for easier configuration in the event of using a different S3 bucket.

## Testing Instructions

- Checkout this branch
- Setup this project
- Go to https://localhost:8000/
  - [ ] Ensure it works correctly

### DIFF
--- a/deployment/ansible/group_vars/all
+++ b/deployment/ansible/group_vars/all
@@ -44,7 +44,7 @@ docker_compose_version: "1.26.*"
 geop_host: "localhost"
 geop_port: 8090
 
-geop_version: "5.3.0"
+geop_version: "5.4.0"
 geop_cache_enabled: 1
 geop_timeout: 200s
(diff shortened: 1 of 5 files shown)


jpolchlo (approved):
Looks ... plausible.  For expediency's sake, giving a +1.  Caveat emptor.

rajadain:
Thanks for taking a look!

PR #3633, opened 2024-07-15
AWS 2-4: Add Celery Task for Global Land Cover Data Fetching

rajadain:
## Overview

Adds an endpoint allowing for analysis of Global Land Use / Land Cover data, using the Impact Observatory Annual LULC dataset available on AWS Open Registry.

The new endpoint works identically to the existing `analyze/land` endpoint, except it doesn't include Active River Area.

This implementation uses the Python toolchain for querying and fetching data from STAC, as done in https://github.com/rajadain/mmw-io-10m-lulc-summary. This could potentially be switched for an MMW-Geoprocessing version, if https://github.com/WikiWatershed/mmw-geoprocessing/pull/117 ever becomes more performant.

Input validation and caching are deferred to a later card, with this PR focusing on core implementation.

Closes #3628 

### Demo

```bash
xh --verbose :8000/api/analyze/global-land/2019/ huc=0204020310 Authorization:"Token 4b91f3a5b5a635cb2aa6cefa5000ffdd28828b31"
```
```http
POST /api/analyze/global-land/2019/ HTTP/1.1
Accept: application/json, */*;q=0.5
Accept-Encoding: gzip, deflate, br
Authorization: Token 4b91f3a5b5a635cb2aa6cefa5000ffdd28828b31
Connection: keep-alive
Content-Length: 20
Content-Type: application/json
Host: localhost:8000
User-Agent: xh/0.22.0

{
    "huc": "0204020310"
}
```
```http
HTTP/1.1 200 OK
Allow: POST, OPTIONS
Connection: keep-alive
Content-Encoding: gzip
Content-Type: application/json
Date: Mon, 15 Jul 2024 15:35:39 GMT
Location: /api/jobs/caf54a2a-5f89-47dd-b5cd-903b40d74934/
Server: nginx
Transfer-Encoding: chunked
Vary: Accept-Encoding
Vary: Accept, Cookie, Origin

{
    "job": "caf54a2a-5f89-47dd-b5cd-903b40d74934",
    "job_uuid": "caf54a2a-5f89-47dd-b5cd-903b40d74934",
    "status": "started",
    "messages": [
        "The `job` field will be deprecated in an upcoming release. Please switch to using `job_uuid` instead."
    ]
}
```
```bash
xh --verbose :8000/api/jobs/caf54a2a-5f89-47dd-b5cd-903b40d74934/ Authorization:"Token 4b91f3a5b5a6
35cb2aa6cefa5000ffdd28828b31"
```
```http
[...]

PR #3635, opened 2024-07-16
Add layer config for preliminary IO LULC static tileset

KlaasH:
## Overview

Adds the 2023 IO LULC layer to the layer settings, pointing at the static tiles that were generated and uploaded to the Datahub S3 bucket. This is just a preview of the layer, since the tiles were only generated through z8 and this static source will ultimately be replaced by a hybrid static/dynamic TiTiler setup (see
https://github.com/WikiWatershed/mmw-tiler/issues/10 and https://github.com/WikiWatershed/mmw-tiler/issues/11).

Connects https://github.com/WikiWatershed/mmw-tiler/issues/9

### Demo

image

### Notes

- ~At the moment, the tile for zoomlevel zero is missing. The tiles for z3 and below have to be generated locally because they take longer than the maximum Lambda timeout, and z0 still timed out when I set the limit to 90 minutes.  I'll try again overnight, but if that doesn't work I'll give up, since this is just a preview.~ Update: the z0 tile is up now. It took 4.5 hours to generate, per the `time` utility, though my computer had a charger issue and went to sleep for a while during the process, so that might be inflated.
- More details about the tile generation process will be coming soon in a PR on the [mmw-tiler](https://github.com/WikiWatershed/mmw-tiler/) repo.

## Testing Instructions

- Load the site, select the "Coverage Grid" tab from the map layers list, and activate the new "IO Global LULC 2023" layer at the top.
- If you're on the default view, the layer will show up but look very fuzzy since it's over-zoomed from z8 up to z12.
- Open the dev tools network tab and confirm that you get tiles from the new layer (they'll be the ones with a crazy hash for a filename, you'll have to hover to see the z/x/y values) for z8 through z1.
- Confirm that the layer shows up everywhere in the world (though the tiles are empty/invisible in the ocean)

### DIFF
--- a/src/mmw/mmw/settings/layer_settings.py
+++ b/src/mmw/mmw/settings/layer_settings.py
@@ -96,6 +96,22 @@ LAYER_GROUPS = {
     ],
     'coverage': [
         {
+            'display': 'IO Global LULC 2023',
+            'code': 'io-lulc-2023',
+            'css_class_prefix': 'io-lulc-2023 nlcd',
+            'short_display': 'IO LULC 2023',
+            'helptext': 'Global land use/land cover dataset produced by '
+                        'Impact Observatory, Microsoft, and Esri, derived from '
+                        'ESA Sentinel-2 imagery at 10 meter resolution.',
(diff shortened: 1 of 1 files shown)


rajadain (approved):
+1 tested, looks amazing. Will rebase my analyze PR #3633 over this and change the legend keys there, since that has the IO_LULC class names.

Great work!

PR #3652, opened 2024-09-10
AWS 2-5: Analyze Global Streams

### LINKED ISSUE #3645: AWS 2-5: Analyze Global Streams
Similar to the medium resolution streams:

image

With the exception of "Length in Agricultural / Non-Agricultural Areas", which requires intersecting with the land layer. That will be implemented later.

Aggregate results by Stream Order, using the following operations:

- SUM length
- AVG slope
- SUM (dscontarea - uscontarea)

@aufdenkampe what should that final field be called in the UI?

rajadain:
## Overview

Adds analysis for global streams. The analysis is lazy, so it doesn't run until someone explicitly selects it from the dropdown. It is also cached when using WKAoIs.

The analysis doesn't include Agricultural Length, because we're not intersecting it with land use data. That is deferred until later.

The Contributing Area calculation needs feedback from the clients before finalizing.

This also adds Vagrant scripts to increase the disk size of the `services` VM to 128GB, to accommodate the new streams dataset.

Closes #3645 

### Demo

image

https://github.com/user-attachments/assets/0e505c02-4d31-46c9-97db-323ecbcabf6a

## Testing Instructions

Testing this locally requires the addition of the global streams dataset, which is a multi-hour process. To make it easier, I recommend testing on staging, which has this branch deployed:

- Go to https://staging.modelmywatershed.org
- Select or draw an area of interest
- Select "Global Medium Resolution Streams" from the streams analysis tab's dropdown
  - [ ] Ensure it completes


rajadain:
Talked to clients, they recommend removing Contributing Area for now.

rajadain:
Alright removed the Contributing Area column. This is now ready for review.

rajadain:
Reminder to myself for making sure we use the proper projection based on where the AoI is

rachelekm (approved):
This looks good! Noting that I did not test locally so I could avoid loading the global streams dataset, but staging works as expected and confirming the Contributing Area column is also not displayed.

rajadain:
Alright, the code that gets the correct Albers projection based on AoI location is now implemented in [a122968](https://github.com/WikiWatershed/model-my-watershed/pull/3652/commits/a122968977363304b2769ef20cd1302810feddfd). This is now ready for another look on staging. Please run the same shape, Schuylkill HUC-8 (I've cleared the cache) and check that you get the same numbers as in the screenshot.

rachelekm:
> #3645

Sorry for the delay in re-reviewing! Confirming that I get the same numbers in the analysis on staging as in the screenshot above 👍

rajadain:
Thanks for reviewing!

Issue #3704, opened 2025-09-08
PA DEP: Add UI for Point Source Analysis


rajadain:
Currently, we show the underlying selected point sources in the Analyze tab without revealing whether we are using the National `ms_pointsource` or DRB `ms_pointsource_drb` dataset:

<img width="1392" height="938" alt="Image" src="https://github.com/user-attachments/assets/74541924-182e-4798-8e26-67ee0b7588ed" />

We should add a dropdown, similar to the land and streams tabs, that lets the user pick which dataset they are using:

<img width="1392" height="938" alt="Image" src="https://github.com/user-attachments/assets/39df4f4e-7839-4090-a636-5232321a6808" />

Once the back-end implementation is complete #3693:

- [ ] Add a drop-down to the Point Sources Analyze tab, using the same components and patterns as the Land and Streams tabs
- [ ] Ensure selecting a different value runs a different analysis
- [ ] Make this a `lazy` analysis, i.e. the non-selected dropdowns will only run on demand, and not by default
- [ ] Add PA to this list of options
- [ ] Ensure that DRB and PA are grayed out if the AoI is outside of those areas

Issue #3702, opened 2025-08-25
PA DEP: Update Recent Weather Eligibility Check


rajadain:
Currently, we use an AoI containment check as a proxy to decide whether or not to allow Recent Weather:

https://github.com/WikiWatershed/model-my-watershed/blob/81c6cae41c59329f60b738a7d967208279066fc1/src/mmw/apps/modeling/models.py#L86-L94

https://github.com/WikiWatershed/model-my-watershed/blob/81c6cae41c59329f60b738a7d967208279066fc1/src/mmw/apps/modeling/views.py#L144-L149

But this is not exact, leading to cases like this, where some natural watersheds that should be modelable are not, not because they don't have recent weather available, but because they are not contained within our boundary checks: https://github.com/WikiWatershed/model-my-watershed/pull/3701#pullrequestreview-3133631664

To alleviate this, we should switch from using an area proxy to testing if recent weather is available for the weather stations selected for a project.

The steps to do this may look like:

- Add a `project` field called `has_recent_weather` 
- Check to see if `project.model_package` is `gwlfe`, and if so,
- Find the nearest weather stations using [`nearest_weather_stations`](https://github.com/WikiWatershed/model-my-watershed/blob/develop/src/mmw/apps/modeling/mapshed/calcs.py#L50),
- Then check the NLDAS bucket to see if there's files there corresponding to the weather station id
- If so, set it to `true`, else `false`
- Remove all references to `in_drwi` and `in_pa` first

Since this check involves database calls, geometry operations, and S3 calls, try to find the most efficient place to do this, so it happens once, not every time.

PR #3708, opened 2025-09-12
Add Point Source Analysis Selector

rachelekm:
## Overview

Updates the Point Sources Analyze tab UI to display a dropdown enabling users to conditionally select different point source analyses to run. Dropdown follows UX of other analysis tab selectors:

- If a user selects an area of interest within the continental US but outside the DRB and PA, the user does not see a dropdown selector the CONUS analysis runs by default
- If user selects an area of interest within the continental US and PA but outside the DRB, the user will see the dropdown selector options for PA and CONUS. The PA analysis will run by default and the CONUS analysis is lazy loaded and will run if selected.
- If user selects an area of interest within the continental US, PA and DRB, the user will see the dropdown selector options for all task options. The DRB analysis will run by default and the PA and CONUS analyses are lazy loaded and will run if selected.

Connects #3704

### Demo
All below screenshots show an area of interest's datasource options dropdown as well as the initial call for analysis path, which includes default datasource option.

AoI within CONUS only:
Screenshot 2025-09-14 at 12 59 35 PM

AoI within PA and CONUS:
Screenshot 2025-09-14 at 12 54 35 PM

AoI within DRB, PA, and CONUS:

Screenshot 2025-09-14 at 12 53 21 PM

AoI within DRB, CONUS, and intersecting PA but not within:
Screenshot 2025-09-14 at 1 51 54 PM

AoI within CONUS, intersecting PA but not within:

Screenshot 2025-09-14 at 1 49 21 PM

### Notes
I noticed in some scenarios the frontend check for perimeters was still allowing PA as an option when there was overlap, causing the request to fail as the backend serializer has a more strict containment method. Commit 5a9a57bc46b3793fceb4fd771033e5b96462ee60 adds a new PA containment check on the frontend that better mirrors the logic of the backend validation. This utility function is separate from `isInPA` so as not to affect weather stations work, although we may want to similarly evaluate how to handle intersecting areas for analysis as as part of #3702.
Screenshot 2025-09-12 at 6 35 58 PM

## Testing Instructions

 * `./scripts/bundle.sh`
 * Spin up app and open dev tools to view network requests
 * Select an area of interest outside of PA and DRB
 * Confirm there is only one point source analysis request called with the path param includes `conus`
 * Confirm point source analysis is successful
 * Confirm the point source analysis dropdown selector is hidden as there is only one option
 * Change area and select an area of interest fully within PA
 * Confirm there is only one point source analysis request called with the path param includes `pa`
 * Confirm point source analysis is successful
 * Confirm the point source analysis dropdown selector is displaying with the options for PA and CONUS
 * Select CONUS option and confirm point source analysis request called with the path param `conus`
 * Confirm point source analysis is successful
 * Change area and select an area of interest fully within DRB
 * Confirm there is only one point source analysis request called with the path param includes `drb`
 * Confirm point source analysis is successful
 * Confirm the point source analysis dropdown selector is displaying with the options for CONUS and PA
 * Select each option and confirm point source analysis request called with their respective path params
 * Confirm each point source analysis is successful
 * Change area and select an area of interest that intersects the PA buffered boundary
 * Confirm PA as an option is ignored and initial point source analysis called with appropriate path param
 * Confirm PA datasource is not an option in the analysis dropdown


src/mmw/js/src/analyze/models.js
```diff
@@ -436,11 +436,31 @@ function createAnalyzeTaskGroupCollection(aoi, wkaoi) {
...
+                    displayName: "Pennsylvania Point Source Data (NPDES)",
+                    area_of_interest: aoi,
+                    wkaoi: wkaoi,
+                    taskName: "analyze/pointsource/pa",
+                    conusOnly: true,
+                    lazy: true
+                },
                 {
                     name: "pointsource",
+                    displayName: "Continental US Point Source Data (NPDES)",
```
rachelekm, line 458: I added these display names but let me know if there's specific text I should use for these labels instead

src/mmw/js/src/analyze/models.js
```diff
@@ -472,15 +492,48 @@ function createAnalyzeTaskGroupCollection(aoi, wkaoi) {
...
+
+                // DRB is the point source analysis default task.
+                // Mark new first task in point source task group
+                // as not lazy so the UI has something to show.
+                if(isPointSource){
+                    if (tg.tasks.length > 0) {
+                        tg.tasks[0].lazy = false;
+                    }
+                }
+
```
rachelekm: Instead of greying out invalid options in an analysis selector dropdown the existing setup is to remove options from the dropdown entirely. I decided to continue with that convention for consistency.

rajadain:
Taking a look now

src/mmw/bundle.sh
```diff
@@ -145,6 +145,7 @@ JS_DEPS=(backbone
          turf-erase
          turf-intersect
          turf-kinks
+         turf-boolean-contains
```
rajadain:
```suggestion
         @turf/boolean-contains
```

src/mmw/js/src/analyze/models.js
rajadain: Let's get these on staging for now, will ask Barry if he'd like to change them.

src/mmw/js/src/analyze/models.js
```diff
@@ -472,15 +492,48 @@ function createAnalyzeTaskGroupCollection(aoi, wkaoi) {
...
+            // Remove task groups with no tasks
+            .filter(function(tg) {
+                return !_.isEmpty(tg.tasks);
+            })
+            .value();
+    }
+
     if (!utils.isInDrb(aoi)) {
         var isDrbTask = function(task) {
             return task.name.startsWith('drb');
```
rajadain: Let's do this, and rename the tasks to be `pointsource_conus`, `pointsource_pa`, and `pointsource_drb`, to follow the pattern of `streams_nhd`, etc.

```suggestion
            return task.name.startsWith('drb') || task.name.endsWith('_drb');
```

rajadain (changes requested):
Great work! This is almost there, and looks pretty good.

The one thing we need changed: PA should never be the default analysis. It should either be DRB if the AoI is within DRB, else CONUS. PA should be available in the dropdown if AoI is within PA, but never the default. This is to maintain parity with Modeling, where the old behavior was DRB or CONUS only, which we will continue for backwards compatibility.

src/mmw/js/src/analyze/models.js
rachelekm: Updated in e5ec7dcf235727f76c53060e2ebf69a7c3f57281

src/mmw/bundle.sh
rachelekm: Fixed in 063d353c173829d42004285d499bafc3a0509681

rachelekm:
Thanks, @rajadain! I implemented the above fixes including ensuring that PA is never the default analysis (the work is in the last commit b5cfbbad3b45f25c206a4285a44ff0b0e465556c). This is all set for another review!
Screenshot 2025-09-16 at 3 52 06 PM
Screenshot 2025-09-16 at 3 59 32 PM

rajadain:
Taking another look

rajadain (approved):
This looks good to me! Nice work getting geo check to match on the front-end. Let's squash the fixups, wait for CI, and merge. The CI failed on a flaky test, should pass next time with no changes.

PR #3709, opened 2025-09-16
Add backend support for mapshed modeling pointsource modifications

rachelekm:
## Overview

Updates the backend to support point source modifications in gwlfe modeling scenarios. Follows existing scenario modification conventions setup in weather data.

As part of adding a new `pointsource_type` field to `Scenario`, conditionally initializes the existing scenario objects with their respective pointsource types ('conus' or 'drb').

Connects #3707 

### Demo

Existing scenario response with `pointsource_type` of `drb`
Screenshot 2025-09-16 at 2 05 49 PM

Newly created scenarios currently default to NULL (must update request data with project's existing `pointsource_type` as part of next Task 3 issue)

Screenshot 2025-09-16 at 1 23 28 PM

Request for pointsource data modifications given valid type:
```
$ curl 'http://localhost:8000/mmw/modeling/projects/2/pointsource/drb' \  ...
{
    "output": {
        "PointNitr": [
            2.7983277227869334,
            2.7983277227869334,
            2.7983277227869334,
            2.7983277227869334,
            2.7983277227869334,
            2.7983277227869334,
            2.7983277227869334,
            2.7983277227869334,
            2.7983277227869334,
            2.7983277227869334,
            2.7983277227869334,
            2.7983277227869334
        ],
        "PointPhos": [
            0.2399995188096,
            0.2399995188096,
            0.2399995188096,
            0.2399995188096,
            0.2399995188096,
            0.2399995188096,
            0.2399995188096,
            0.2399995188096,
            0.2399995188096,
            0.2399995188096,
            0.2399995188096,
            0.2399995188096
        ],
        "PointFlow": [
            0.00021866740944337425,
            0.00019750604723917672,
            0.00021866740944337425,
            0.0002116136220419751,
            0.00021866740944337425,
            0.0002116136220419751,
            0.00021866740944337425,
            0.00021866740944337425,
            0.0002116136220419751,
            0.00021866740944337425,
            0.0002116136220419751,
            0.00021866740944337425
        ]
    }
} 
```

Request for pointsource data modifications given invalid type (area outside of DRB):
```
$ curl 'http://localhost:8000/mmw/modeling/projects/14/pointsource/drb' \  ...
{"errors":["Only supported within Delware River Basin."]}  
```

## Testing Instructions

 * `./scripts/manage.sh migrate && ./scripts/bundle.sh`
 * Confirm successful data migration
 * Open app in browser, login, and open up the networks tab of browser developer tools
 * Navigate to projects list and launch any existing project
[...]

PR #3711, opened 2025-09-17
PA DEP: Update Weather Simulation Eligibility Check

### LINKED ISSUE #3702: PA DEP: Update Recent Weather Eligibility Check

rajadain:
## Overview

Previously we would use an area proxy to gate weather simulations: Future Weather was DRB only, and Recent Weather was DRWI and PA. However, this proxy is inaccurate, leading to certain eligible AoIs being excluded.

This new check sees if weather stations selected for a project have Future or Recent Weather data available for them. If so, the simulations are marked as enabled. Otherwise, they are disabled.

This is checked once, when a new project is created, since it is a function of the area of interest, and is saved to the project's model. This work will have to be rebased atop #3709 since that also touches the project model.

A migration is added to backfill all existing projects with the correct simulations. This is done in direct SQL for performance, necessary in production which currently has ~95K projects.

The front-end is updated to read the simulation configuration from the project, rather than constants.

This allows us to get rid of the `in_drwi` and `in_pa` fields. `in_drb` is used in the Land Cover modal, so that is kept.

Closes #3702 

### Demo

TODO

## Testing Instructions

- On `develop`, create and save 3 projects: one in DRB (Philadelphia), one in PA (Pittsburgh), and one outside (e.g. Chicago)
- Check out this branch
- Run `./scripts/manage.sh migrate` to install the new field and backfill the existing projects
- Run `./scripts/bundle.sh --debug` to update the front-end scripts
- Open the Philadelphia project, and look at its available weather options
  - [x] Ensure both Recent and Future Weather are available
- Open the Pittsburgh project, and look at its available weather options
  - [x] Ensure both Recent Weather is available
- Open the Chicago project, and look at its available weather options
  - [x] Ensure neither is available
- Create a new project
  - [ ] Ensure its weather simulations are correctly set based on where you made it


src/mmw/apps/modeling/migrations/0044_project_weather_simulations.py
```diff
@@ -0,0 +1,80 @@
...
+        '''),
+        # Then set those eligible for future weather
+        migrations.RunSQL(f'''
+            UPDATE modeling_project
+            SET weather_simulations = '{WEATHER_SIMULATIONS["future"]}'::jsonb
+            WHERE model_package = 'gwlfe'
+              AND (
+	            SELECT COUNT(station_id)
+	            FROM (
+		            SELECT (station_data ->> 'station')::int AS station_id
```
rajadain: Should ensure we use spaces for indentation, no tabs

src/mmw/apps/modeling/migrations/0044_project_weather_simulations.py
```diff
@@ -0,0 +1,80 @@
...
+            SET weather_simulations = '{WEATHER_SIMULATIONS["both"]}'::jsonb
+            WHERE model_package = 'gwlfe'
+              AND (
+	            SELECT COUNT(station_id)
+	            FROM (
+		            SELECT (station_data ->> 'station')::int AS station_id
+		            FROM jsonb_array_elements(gis_data::jsonb -> 'WeatherStations') as station_data
+                ) AS stations_in_project
+	            WHERE station_id = ANY(ARRAY[{",".join(common_weather_stations)}])
+              ) = 2;
```
rajadain: Should probably explain this magic query more. `jsonb_array_elements` converts an array of JSON elements into individual rows. When we `COUNT(station_id)` that are in the given array, and total to 2, that means both weather stations are in that array, which is what we need to test before assigning a simulation to a project.

src/mmw/apps/modeling/calcs.py
```diff
@@ -191,6 +192,41 @@ def get_weather_simulation_for_project(project, category):
...
+    # Remove any simulation that does not have values for selected stations
+    for sim in WeatherType.simulations:
+        if sim in available_sims:
+            for ws in wss:
+                url = settings.WEATHER_DATA_BUCKET_URL.format(
+                          category=sim, station=ws.station)
+                res = requests.head(url)
+                if not res.ok:
+                    available_sims.remove(sim)
+                    break
```
rajadain, line 218: Haven't written a `break` statement in a while. We're trying to disqualify simulations as soon as possible, so break out of the loop once that is established.

src/mmw/apps/modeling/serializers.py
```diff
@@ -195,6 +198,25 @@ def create(self, validated_data):
...
             validated_data['user'] = self.context['request'].user
 
+        if validated_data['model_package'] == Project.GWLFE and (
+           'weather_simulations' not in validated_data or
+           validated_data['weather_simulations'] == []):
+            # Initialize weather simulations for this project
+            try:
+                validated_data['weather_simulations'] = \
+                    get_available_simulations_for_aoi(
+                        validated_data['area_of_interest'])
```
rajadain: Since this makes a call to S3 to check eligibility, that's an external dependency that could potentially fail. Currently we wrap this in a `try` and `except` to the default value (all simulations disabled), to allow for a project to still be created. However, this behavior may be mysterious to users, who won't know why their project isn't eligible for certain simulations.

src/mmw/apps/modeling/views.py
```diff
@@ -141,22 +141,10 @@ def project_weather(request, proj_id, category):
...
-    if category in ['RCP45_2080_2099', 'RCP85_2080_2099'] \
-            and not project.in_drb:
-        return Response({'errors': ['Only supported within'
-                                    ' Delware River Basin.']},
-                        status=status.HTTP_400_BAD_REQUEST)
-
     # Get weather simulation data for this project
     mods, errs = get_weather_simulation_for_project(project, category)
 
+    # TODO We should handle 404s here
```
rajadain: Currently `get_weather_simulation_for_project` returns a list of text errors. Will have to devise a way to extract the status code from that, and return a 400 or 404 as appropriate.

src/mmw/js/src/modeling/gwlfe/quality/views.js
```diff
@@ -1,7 +1,7 @@
 "use strict";
 
 var $ = require('jquery'),
-    _ = require('underscore'),
+    _ = require('lodash'),
```
rajadain, line 4: This is required for `_.chain` below

src/mmw/js/src/modeling/gwlfe/runoff/views.js
```diff
@@ -3,11 +3,11 @@
 var $ = require('jquery'),
     _ = require('lodash'),
     Marionette = require('../../../../shim/backbone.marionette'),
+    App = require('../../../app.js'),
```
rajadain, line 6: This is a poor design choice that we've never been able to get away from. Since you often need a reference to the current project, or the current AoI, we put that in the top level `App`, but then you often end up importing it all over the project, which can create circular dependencies if you're not careful. Thankfully, that's not the case for this file.

src/mmw/js/src/modeling/gwlfe/weather/templates/modal.html
```diff
@@ -31,10 +31,10 @@ <h2 class="weather-col-title">
                     <optgroup label="Default">
                         <option value="DEFAULT" {{ 'selected' if available_data == 'DEFAULT' }}>USEPA National Climate 1960&ndash;1990</option>
                     </optgroup>
-                    {% for s in Simulations %}
-                        <optgroup label="{{ s.group }} {{ '(DRB Only)' if s.in_drb and not in_drb }} {{ '(DRWI or PA Only)' if (s.in_drwi or s.in_pa) and not (in_drwi or in_pa) }}">
+                    {% for s in weather_simulations %}
+                        <optgroup label="{{ s.group }} {{ '(' + s.eligible_area + ' only)' if s.disabled }}">
                             {% for i in s.items %}
-                                <option value="{{ i.name }}" {{ 'selected' if available_data == i.name }} {{ 'disabled' if (s.in_drb and not in_drb) or ((s.in_drwi or s.in_pa) and not (in_drwi or in_pa)) }}>{{ i.label }}</option>
+                                <option value="{{ i.name }}" {{ 'selected' if available_data == i.name }} {{ 'disabled' if s.disabled }}>{{ i.label }}</option>
```
rajadain, line 37: Much simpler and cleaner

rajadain (commented):
Left some comments

rachelekm (approved):
Great job with this refactor 👏 This feels much more clean and the weather station options are working as expected for all testing scenarios and well as some of the border PA areas we tested originally. For the sake of time I didn't get too into the weeds with the SQL query or calc, but from a cursory look these seem efficient and robust

We discussed how the point source analysis/modeling scenario modifications continue to use PA state constraints so we can't get rid of those quite yet. Just caught that you reverted this so this looks good, thank you! There should still be an opportunity for cleanup, at least with old simulation constants and dwri constraint checks, but will be easier to go back and cleanup once point sources work is complete.

rajadain:
May need to change how the default value is filled out. Putting that logic in the serializer is probably not best practice, and it seems to break a number of tests.

rajadain:
Thanks for the review!

Issue #3710, opened 2025-09-16
PA DEP: Add UI for Point Source modifications in MapShed modeling


rachelekm:
Following work in #3709, add frontend UX to support selecting a point source dataset for MapShed modeling. Dropdown for point source datasets should mirror UI of existing modifiable data, specifically dropdown of Conservation Practice
<img width="576" height="175" alt="Image" src="https://github.com/user-attachments/assets/4263da9c-408e-41a5-aed7-b0bfed387572" />

- [ ] On the creation of a new scenario, ensure request data sets the `pointsource_type` value with the current conditions of project
- [ ] Button to modify point sources available in button group following "Add changes to area"
- [ ] Point source type options in dropdown are constrained by geography following same conventions in backend work #3709
- [ ] On selection of new point source, call for project point source type data
- [ ] Following successful response of point source data modifications, update the `pointsource_type` value of scenario and respective data values from response

rachelekm:
@rajadain documenting our check-in here, I just pushed up the recently rebased branch of this in-progress work to remote. That is available at https://github.com/WikiWatershed/model-my-watershed/commit/092e07cbddc3c85aad919dad4deb92422348aaf0#diff-20cbff3e8a54f385cb242ea727508e47f9dc6442ba3a36022c1fb59f61794bca.

The above commit creates a very basic setup for adding point source modifications to a scenario, including: point source button to match others in scenario control toolbar, a modal dropdown on select with all point source type options in dropdown, and very basic scaffolding for fetching point source data on selection of option. At this moment, the fetched data gets mapped to individual GwlfeModificationModel objects and passed into addModifications() but this isn't worked as expected so it may be they're supposed to be passed in as one object.

Following successful response of point source data modifications the pointsource_type value of scenario is updated.

However, noting that initially setting project id on completely new projects as well as setting pointsource_type defaults on new scenarios is not setup in the above commit. Similarly, the point source options in the dropdown are not geographically constrained and will always show all 3 options (plus, needs lots of design polishing).

Thank you for picking this work up!

rajadain:
After looking into this a bit more, it seems like we already have UI for providing manual point source values:

<img width="957" height="971" alt="Image" src="https://github.com/user-attachments/assets/421298a7-3358-41a0-ab6d-71a526f73d70" />

It would make sense to add the new selector somewhere there, but the Entry Modal's construction is quite programmatic, will make adding something there difficult. However, if something is added elsewhere, it will also need to update these values.

PR #3712, opened 2025-09-19
Add UI for Point Source Preset Selection for GWLFE Modeling

rajadain:
## Overview

Adds a UI for selecting a preset for Point Sources / Waste Water. Opposed to the original instructions in #3710, we don't need a top level dropdown for this, because it is already covered in Settings modal. We add a preset dropdown atop the relevant Waste Water fields, which can still be overwritten by users.

We show DRB and PA options when available. This only works for saved projects. 

Connects #3710 

### Demo

[Screencast from 2025-09-19 16-22-07.webm](https://github.com/user-attachments/assets/d9697ef6-2153-414b-9353-b0d4f4622374)


rachelekm (approved):
Tested this locally following our pairing and this is working well for a variety of scenarios 👏 great job finding the existing inputs for a really clean solution

